PR #2743, opened 2019-03-12
tkt-79851: Ensure webdav share directories exist

sonicaj:
This commit makes sure webdav share directories exist when we create/update or start webdav service.
Ticket: #79851

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/apache24/Includes/webdav.conf
+++ b/src/middlewared/middlewared/etc_files/local/apache24/Includes/webdav.conf
@@ -62,7 +62,7 @@
 	def change_perms(path):
 		subprocess.Popen(f'chown -R webdav:webdav {path}', stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True, close_fds=True).communicate()
 
-%>
+%>\
 Listen ${webdav_config['tcpport']}
 	<VirtualHost *:${webdav_config['tcpport']}>
 		DavLockDB "/etc/local/apache24/var/DavLock"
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/webdav.py
```diff
@@ -24,6 +25,8 @@ class Config:
             )
         else:
             await check_path_resides_within_volume(verrors, self.middleware, f'{schema}.path', data['path'])
+            if not verrors and not os.path.exists(path):
```
william-gr: There could be errors after this and the directory will be create.

It seems to make more sense to create only after, if there are no errors.

src/middlewared/middlewared/etc_files/local/apache24/Includes/webdav.conf
```diff
@@ -97,9 +97,13 @@ Listen ${webdav_config['tcpport']}
 			AllowMethods GET OPTIONS PROPFIND
 		</Location>
 	% endif
-<% if share['perm']:
-	change_perms(share['path'])	
-%>
+<%
+	if not os.path.exists(share['path']):
+		os.makedirs(share['path'])
```
william-gr: Not sure if we want to create the dir in the config file gen, wouldnt it be enough in share create?

sonicaj: Actually i was confused about that. So that will only work if the user is updating/creating the share - this is for compatibility for old shares. It can be removed if you think this doesn't make much sense here. Thoughts ?

william-gr: I am not sure other shares do it that way?

We either mkdir on create/update or config gen, not both. I vote for create/update.

sonicaj: I agree. Will have the PR updated asap ; )

PR #2750, opened 2019-03-13
tkt-74442: feat(middleware/disk): add disk type, rotation rate and model

william-gr:
Ticket:	#74442


src/middlewared/middlewared/plugins/disk.py
```diff
@@ -884,20 +901,10 @@ def check_clean(self, disk):
...
                 disk['disk_name'] = name
 
-            reg = RE_DSKNAME.search(name)
-            if reg:
-                disk['disk_subsystem'] = reg.group(1)
-                disk['disk_number'] = int(reg.group(2))
-            serial = ''
-            g = geom.geom_by_name('DISK', name)
+            g = await self.__disk_data(disk, name)
+            serial = disk.get('disk_serial') or ''
```
skarekrow:
```suggestion
            serial = disk.get('disk_serial', '')
```

william-gr: Thats not the same, or '' is safer

PR #2755, opened 2019-03-14
feat(rc): move ix-collectd and ix-system to pool.import_on_boot

william-gr:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2379,6 +2379,14 @@ def import_on_boot(self, job):
                     'aclmode': {'value': 'restricted'},
                 }})
 
+        try:
+            # Now that pools have been imported we are ready to configure system dataset
+            # and collectd which may depend on them.
+            self.middleware.call('etc.generate', 'system_dataset')
+            self.middleware.call('etc.generate', 'collectd')
```
themylogin: Is this intended to be skipped if `system_dataset` fails?

william-gr: Yes, there is no point in raising the exception and not continue on remaining tasks

themylogin: I mean, skip setting up collectd if system_dataset setup failed

themylogin: We probably need to prevent starting collectd if it is not set up

william-gr: If system dataset fails to the point of generating an exception we are probably already dead in the water and the system will malfunction

PR #2758, opened 2019-03-14
Document websocket events

william-gr:


src/middlewared/middlewared/main.py
```diff
@@ -1191,13 +1192,29 @@ def done(_):
...
+        and can be browsed in documentation page without source code inspection.
+        """
+        self.__events.register(name, description)
+
     def send_event(self, name, event_type, **kwargs):
+
+        if name not in self.__events:
+            # We should eventually deny events that are not registered to ensure every event is
+            # documented but for backward-compability and safety just log it for now.
+            self.logger.debug(f'Event {name} not registered.')
```
skarekrow: Perhaps a higher level since we want to deny this eventually?

PR #2759, opened 2019-03-14
tkt-76621: Implement shadow_copy_zfs as default

anodos325:
- Add checkbox to GUI "enable snapshots"
- Remove snapshot task selection dropdown

### DIFF
--- a/docs/api/resources/sharing.rst
+++ b/docs/api/resources/sharing.rst
@@ -46,7 +46,6 @@ List resource
                 "cifs_path": "/mnt/tank/MyShare",
                 "cifs_ro": false,
                 "cifs_guestonly": true,
-                "cifs_storage_task": null,
                 "id": 1,
                 "cifs_browsable": true
         }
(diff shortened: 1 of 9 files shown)


gui/sharing/migrations/0015_add_shadow_copy_checkbox.py
```diff
@@ -0,0 +1,26 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name='cifs_share',
+            name='cifs_shadowcopy',
+            field=models.BooleanField(default=False, verbose_name='Enable Shadow Copies'),
+        ),
+        migrations.RunPython(
+            migrate_shadowcopies
+        ),
```
william-gr: Dont you have to remove the old field?

gui/api/resources.py
```diff
@@ -1774,10 +1774,6 @@ class Meta:
 
     def dehydrate(self, bundle):
         bundle = super().dehydrate(bundle)
```
william-gr: Might as well remove the whole method

docs/api/resources/sharing.rst
```diff
@@ -46,7 +46,6 @@ List resource
                 "cifs_path": "/mnt/tank/MyShare",
                 "cifs_ro": false,
                 "cifs_guestonly": true,
-                "cifs_storage_task": null,
```
william-gr: We should probably add cifs_shadowcopy in these places?

PR #2770, opened 2019-03-19
Volume check in migration

sonicaj:
This commit checks if a volume exists before selecting the first volume as in the case where we don't have any volumes i.e build, this will result in a failure.


william-gr:
Please squash and merge

sonicaj:
There only is one commit ? :p

william-gr:
I am on mobile :)

sonicaj:
Already done, no worries ;)

PR #2772, opened 2019-03-19
Initial NIS middleware plugin

anodos325:


src/middlewared/middlewared/plugins/nis.py
```diff
@@ -0,0 +1,189 @@
...
+        new.update(data)
+        await self.nis_compress(new)
+        await self.middleware.call(
+            'datastore.update',
+            'directoryservice.nis',
+            '1',
+            new,
+            {'prefix': 'nis_'}
+        )
+        await self.nis_extend(new)
```
william-gr: `return await self.config()` instead of the extend call

src/middlewared/middlewared/plugins/nis.py
```diff
@@ -0,0 +1,189 @@
...
+
+        await self.middleware.call('cache.pop', 'NIS_State')
+        await self.middleware.call('etc.generate', 'rc')
+        await self.middleware.call('etc.generate', 'pam')
+        await self.middleware.call('etc.generate', 'hostname')
+        await self.middleware.call('etc.generate', 'nss')
+        await self.middleware.call('nis.cache_fill')
+        return True
+
+    @job(lock=lambda args: 'nis_cache_fill')
```
william-gr: Should this be private?

gui/directoryservice/forms.py
```diff
@@ -712,48 +713,53 @@ def save(self):
...
-        # XXX: We need to have a method to test server connection.
-        try:
-            started = notifier().started("nis")
-        except Exception:
-            raise MiddlewareError(_("Failed to check NIS status."))
-        finally:
-            super(NISForm, self).save()
-
+        with client as c:
+            state = c.call('nis.get_state')
```
william-gr: I am thinking all this logic about start and start/restart service should be bundled into nis.update method?

Otherwise this will be left to the new UI team to figure out and more logic in the client side.

src/middlewared/middlewared/plugins/nis.py
```diff
@@ -0,0 +1,189 @@
...
+    ))
+    async def do_update(self, data):
+        old = await self.config()
+        new = old.copy()
+        new.update(data)
+        await self.nis_compress(new)
+        await self.middleware.call(
+            'datastore.update',
+            'directoryservice.nis',
+            '1',
```
william-gr: Why no old['id'] ?

src/middlewared/middlewared/plugins/nis.py
```diff
@@ -0,0 +1,189 @@
...
+    @private
+    async def nis_extend(self, nis):
+        nis['servers'] = nis['servers'].split(',') if nis['servers'] else []
+        nis.pop('id')
+        return nis
+
+    @private
+    async def nis_compress(self, nis):
+        nis['servers'] = ','.join(nis['servers'])
+        nis.update({'id': 1})
```
william-gr: Why forcing id 1?

src/middlewared/middlewared/plugins/nis.py
```diff
@@ -0,0 +1,189 @@
...
+
+    @private
+    async def started(self):
+        """
+        If for some reason, the status check returns false, but NIS is enabled
+        set the NIS_state value to FAULTED.
+        The return code from ypwhich is not a reliable health indicator.
+        """
+        nis = await self.middleware.call('nis.config')
+        ypwhich = await run(['/usr/bin/ypwhich'], check=False)
```
william-gr: Didnt you mention ypwhich could block?

subprocess has a timeout param, if that doesnt work we can use asyncio.wait_for

src/middlewared/middlewared/plugins/nis.py
anodos325: Yes

gui/directoryservice/forms.py
anodos325: Yes. I agree.

gui/directoryservice/forms.py
```diff
@@ -712,48 +713,27 @@ def save(self):
...
+    middleware_plugin = 'nis'
+    is_singletone = True
+
     class Meta:
         fields = '__all__'
         model = models.NIS
 
     def __init__(self, *args, **kwargs):
         super(NISForm, self).__init__(*args, **kwargs)
+        self.original_instance = dict(self.instance.__dict__)
```
william-gr: Is this being used anywhere?

src/middlewared/middlewared/plugins/nis.py
```diff
@@ -0,0 +1,211 @@
...
+            errmsg = ypbind.stderr.decode().strip()
+            if 'ypbind not running' not in errmsg:
+                raise CallError(f'ypbind failed to stop: [{ypbind.stderr.decode().strip()}]')
+
+        await self.middleware.call('cache.pop', 'NIS_State')
+        await self.middleware.call('etc.generate', 'rc')
+        await self.middleware.call('etc.generate', 'pam')
+        await self.middleware.call('etc.generate', 'hostname')
+        await self.middleware.call('etc.generate', 'nss')
+        await self.middleware.call('nis.cache_fill')
```
william-gr: Should we be calling cache_fill on stop? Wouldnt it be cache expire?

PR #2792, opened 2019-03-24
tkt-69562: Fix "unscheduled system reboot" alert after update

themylogin:


themylogin:
When update process creates a clone of current boot env, sentinel file is still present in `/data`, so new boot env inherits it there.

themylogin:
backport

PR #2793, opened 2019-03-25
tkt-64392: Crash reporting opt-out

themylogin:
@william-gr I've made two assumptions while working on this, please agree or disagree:

1. Crash Reporting is undecided (which means off) for both existing and newly installed systems
2. In new UI we'll require to agree or disagree to enable crash reporting on first login. In legacy UI we don't want to spend our time on this, it'll be undecided until user opens System → General


gui/system/forms.py
```diff
@@ -1029,6 +1030,9 @@ def middleware_clean(self, update):
         for key in ('ui_address', 'ui_v6address'):
             if not update.get(key):
                 update.pop(key, None)
+
+        update['crash_reporting'] = update['crash_reporting'] == 'True'
```
william-gr: Why True as string?

gui/system/migrations/0039_crash_reporting.py
```diff
@@ -0,0 +1,22 @@
...
+
+    dependencies = [
+        ('system', '0038_merge_20190226_1008'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='settings',
+            name='stg_crash_reporting',
+            field=models.NullBooleanField(help_text='Enable sending anonymous crash reports to iXsystems', verbose_name='Crash reporting'),
```
william-gr: It should default to True for freenas, False for truenas.

gui/system/forms.py
themylogin: `CheckboxInput` for `NullBooleanField` returns it this way for some reason

gui/system/migrations/0039_crash_reporting.py
themylogin: Fixed

PR #2795, opened 2019-03-25
tkt-83491: Correctly display post install messages

sonicaj:
This commit fixes a bug where we did not display the post install messages for a plugin correctly.
Ticket: #83491


sonicaj:
backport

PR #2800, opened 2019-03-25
tkt-82218: Proactive Support to middlewared

themylogin:


src/middlewared/middlewared/plugins/support.py
```diff
@@ -28,7 +28,71 @@
...
+        await self.middleware.call(
+            'datastore.update',
+            self._config.datastore,
+            config_data['id'],
+            config_data,
+        )
+
+        return await self.config()
+
+    async def is_available(self):
```
william-gr: `@accepts()` ?

Also add a docstring please, same for next 3 methods

PR #2801, opened 2019-03-26
tkt-77916: Use nfsd command-line option instead of /etc/nfsd.virtualhost

themylogin:
@william-gr should this be backported to 11.2? Changes in `network.py` should be backported anyway because original code was not functional.


yocalebo:
Would it make sense to raise a ValidationError in nfs.py in the do_update method for the following conditions?

1. is truenas_ha
2. nfsv4 is enabled
3. kerberos authentication is enabled
4. no virtual hostname is set

PR #2811, opened 2019-03-28
Normalize NFSv4 and TN HA

themylogin:
@yocalebo is this what you wanted yesterday?

### DIFF
--- a/src/middlewared/middlewared/plugins/nfs.py
+++ b/src/middlewared/middlewared/plugins/nfs.py
@@ -54,6 +54,16 @@ class NFSService(SystemServiceService):
 
         verrors = ValidationErrors()
 
+        if new["v4"] and new["v4_krb"] and not await self.middleware.call("system.is_freenas"):
+            if await self.middleware.call("failover.licensed"):
+                gc = await self.middleware.call("datastore.config", "network.globalconfiguration")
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -54,6 +54,15 @@ def nfs_compress(self, nfs):
 
         verrors = ValidationErrors()
 
+        if new["v4"] and not await self.middleware.call("system.is_freenas"):
+            if await self.middleware.call("failover.status") != "SINGLE":
```
william-gr: I think we could use failover.licensed here instead

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -54,6 +54,15 @@ def nfs_compress(self, nfs):
 
         verrors = ValidationErrors()
 
+        if new["v4"] and not await self.middleware.call("system.is_freenas"):
```
yocalebo: @themylogin Looks great! I only have one problem.

We should only require this check if the user has enabled NFSv4 and checks the "Require Kerberos for NFSv4:" in the services section of the webui which is the `nfs['v4_krb']` in db.

We should change the ValidationError verbiage to say "Enabling kerberos authentication on TrueNAS HA requires setting the virtual hostname and domain.

PR #2816, opened 2019-03-28
tkt-76971: Run zfs services in the process poll

william-gr:


src/middlewared/middlewared/main.py
```diff
@@ -738,20 +738,75 @@ def __init__(self, middleware):
         await self.middleware.run_in_thread(t_worker.join)
 
 
-class Middleware(object):
+class MiddlewarePluginsMixin(object):
+
+    def _load_plugins(self, on_module_begin=None, on_module_end=None, on_modules_loaded=None):
+        from middlewared.service import Service, CRUDService, ConfigService, SystemServiceService
```
skarekrow: Why not import these above?

src/middlewared/middlewared/utils/__init__.py
```diff
@@ -381,3 +402,65 @@ def load_classes(module, base, blacklist):
...
+class LoadPluginsMixin(object):
+
+    def __init__(self, overlay_dirs):
+        self.overlay_dirs = overlay_dirs or []
+        self._schemas = Schemas()
+        self._services = {}
+        self._services_aliases = {}
+
+    def _load_plugins(self, on_module_begin=None, on_module_end=None, on_modules_loaded=None):
+        from middlewared.service import Service, CRUDService, ConfigService, SystemServiceService
```
skarekrow: Same here

william-gr: I didnt change that, just moved code. I think it was because circular dependency

PR #2817, opened 2019-03-28
tkt-72659: vmware.match_datastores_with_datasets

themylogin:


src/middlewared/middlewared/plugins/vmware.py
```diff
@@ -188,17 +320,10 @@ def get_datastores(self, data):
...
         connect.Disconnect(server_instance)
 
-        # Datastore names are unique among different esxi host under a single vcenter server. As we only require
-        # datastore names and additional information is not needed for datastore choices, we refine the datastores
-        # dict and send back only the datastore names
-
-        return sorted({
-            name for host in datastores.values() for name in host
-        })
+        return datastores
```
william-gr: My only concern is that we are changing the return type of this method, arent we?

Isnt there anything using it?

themylogin: `get_datastores` was returning `[str]` and now it is the same, nothing has changed, hasn't it?

william-gr: Isnt datastores a dict? Before it was returning a list?

sonicaj: @william-gr, `get_datastores` method's implementation has been redefined above ;)

william-gr: Ah, I was looking at the wrong place ;)

PR #2825, opened 2019-04-02
tkt-39653: fix(jail/fetch): Remove plugin index workaround

skarekrow:
Ticket: #39653


skarekrow:
NOTE: Not merging until iocage side is merged

PR #2826, opened 2019-04-02
tkt-84268: Create a snapshot task when creating volumes

sonicaj:
This commit exposes a snapshot attribute which when supplied will create a periodic snapshot task for the newly created pool.
Ticket: #84268

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -469,6 +469,7 @@ class PoolService(CRUDService):
         'pool_create',
         Str('name', required=True),
         Bool('encryption', default=False),
+        Bool('snapshot_task', default=False),
         Str('deduplication', enum=[None, 'ON', 'VERIFY', 'OFF'], default=None, null=True),
         Dict(
             'topology',
@@ -502,6 +503,9 @@ class PoolService(CRUDService):
     async def do_create(self, job, data):
         """
         Create a new ZFS Pool.
+
+        `snapshot_task` if set, creates a periodic snapshot task which takes snapshots of the created pool every day
+        at 09:15 am with a lifetime of 1 week.
 
         `topology` is a object which requires at least one `data` entry.
         All of `data` entries (vdevs) require to be of the same type.
@@ -659,6 +663,29 @@ class PoolService(CRUDService):
         asyncio.ensure_future(restart_services())
 
         pool = await self._get_instance(pool_id)
+
+        # Pool has been created
+        if data.get('snapshot_task'):
+            await self.middleware.call(
+                'pool.snapshottask.create', {
+                    'dataset': pool['name'],
+                    'recursive': True,
+                    'lifetime_value': 1,
+                    'lifetime_unit': 'WEEK',
+                    'naming_schema': 'auto_%Y-%m-%d_%H-%M',
+                    'allow_empty': False,
+                    'schedule': {
+                        'minute': '15',
+                        'hour': '09',
+                        'dom': '*',
+                        'month': '*',
+                        'dow': '*',
+                        'begin': '09:00',
+                        'end': '18:00'
+                    }
+                }
+            )
+
         await self.middleware.call_hook('pool.post_create_or_update', pool=pool)
         return pool
 


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -503,6 +504,9 @@ def pool_extend(self, pool):
         """
         Create a new ZFS Pool.
 
+        `snapshot_task` if set, creates a periodic snapshot task which takes snapshots of the created pool every day
+        at 09:15 am with a lifetime of 1 week.
```
william-gr: Now that I think about it, I am not sure what I think about 09:15 am.

e.g. What if FreeNAS is turned off that period?

Any better suggestions? Perhaps a snapshot every couple hours?

themylogin: Seems very strange to hardcode this value. No choice will be good enough. I think we should require passing a schedule and ask for it in UI. And UI can use some lightweight widget that by default only asks for hour, minute and day of week. Not doing snapshots at weekends is also an interesting consideration for default value to match default when creating periodic snapshot task. Also "begin" and "end" do not have any sense here, they are nullable and null by default.

sonicaj: @william-gr i like @themylogin's suggestion. We can allow this option to be configurable but we can default to doing this every 4 hours daily and not on weekends . Thoughts ?

william-gr: I have to disagree with you both.

We do the same for scrub, we create a scrub task with a reasonable default because good part of the people do not know what it is and never set it up. They are free to change the scrub attributes at any time.

Same situation with snapshot task. Good part of the people never set up snapshots, which means they lose their data if something bad happens. Why cant we set up a simple sane snapshot task for them? They are free to change it at any time, its not like its hard coded?

We dont want to make it configurable in the pool creation to keep it simple. We do not want to make the task of creating a pool over complicated (which is already is not an easy task) with yet another set of options to configure. And by the same token, we should let them configure scrub attributes as well?

yocalebo: This looks to be like a recursive snapshot from the root of the zpool. This means that all future dataset/zvols will have a snapshot created for them. This will cause duplicate snapshots to be created which isn't an issue but can cause confusion. Can this snapshot task have a unique name so that it can be quickly identified instead of using the standard "auto_.." naming scheme?

william-gr: > This looks to be like a recursive snapshot from the root of the zpool. This means that all future dataset/zvols will have a snapshot created for them. This will cause duplicate snapshots to be created which isn't an issue but can cause confusion. Can this snapshot task have a unique name so that it can be quickly identified instead of using the standard "auto_.." naming scheme?

Yes, very good point!

themylogin: We run default scrub at Sunday 00:00 and we don't care about users who power off their NAS at the end of workday of for entire weekend.

If we want the opposite for snapshots, we should probably set time to 12:00 to maximize the probability of NAS being powered on.

And who disagrees with adding this option? I only think IMPLICIT auto-snapshotting is not a good idea. I have deleted the file, but did not get my disk space back, and when FBI came to my office, they were able to restore it with one mouse click, this was frustrating. On the other hand, tiny "at which hour do you want the daily snapshot, or don't you want it at all" checkbox + dropdown will do both jobs: remove one magic constant that only has 1/24 chance to be the right choice and inform user that he will get snapshots. After all, what if the point of snapshots if user does not know that they exist?

themylogin: > This will cause duplicate snapshots to be created which isn't an issue but can cause confusion

No, they won't

If we have recursive snapshot task for `tank` and another one for `tank/data` and they are scheduled at the same time, `tank/data` will have only one snapshot named `auto-2019-04-03-10-00`. This is one of the features of new replication.

themylogin: > Can this snapshot task have a unique name so that it can be quickly identified instead of using the standard "auto_.." naming scheme?

One the other hand, if we do this, there will be two identical snapshots: `auto-2019-04-03-10-00` and `other-naming-schema-2019-04-03-10-00`

yocalebo: Another potential issue I see with this is will the end user be able to choose which dataset/zvols are snapshotted on pool creation? What happens if a user has a dataset that is used for build environments or "throw-away" data that has 100's of GB of changed data every night? This means the snapshots could potentially take up a lot of space. I only ask this because we have TrueNAS customers that have this exact scenario.

yocalebo: I know what this ticket is trying to address but maybe it would make more sense to prompt the end-user to take a snapshot of the zpool before they upgrade? That way, it's a 1-time operation and we don't have to worry about scheduling or management of space etc etc? Just an idea.

themylogin: The issue with build environments can be solved like this too: allow disabling snapshots and let the user create snapshot task manually with proper "exclude" field

william-gr: > We run default scrub at Sunday 00:00 and we don't care about users who power off their NAS at the end of workday of for entire weekend.
> 
> If we want the opposite for snapshots, we should probably set time to 12:00 to maximize the probability or NAS being powered on.
> 
> And who disagrees with adding this option? I only think IMPLICIT auto-snapshotting is not a good idea. I have deleted the file, but did not get my disk space back, and when FBI came to my office, they were able to restore it with one mouse click, this was frustrating. On the other hand, tiny "at which hour do you want the daily snapshot, or don't you want it at all" checkbox + dropdown will do both jobs: remove one magic constant that only has 1/24 chance to be the right choice and inform user that he will get snapshots. After all, what if the point of snapshots if user does not know that they exist?

If we user will have a checkbox its not implicit, he will know. And it could be very easily changed after pool creation.

However it seems there are enough people against this feature that we should reconsider.

yocalebo: > > This will cause duplicate snapshots to be created which isn't an issue but can cause confusion
> 
> No, they won't
> 
> If we have recursive snapshot task for `tank` and another one for `tank/data` and they are scheduled at the same time, `tank/data` will have only one snapshot named `auto-2019-04-03-10-00`. This is one of the features of new replication.

That's great! I did not know that was the case. However, what happens when snapshot is enabled for tank at `auto-2019-04-03-10-00` and they create another snapshot task for tank/data at 'auto-2019-04-03-11-00` I'm assuming it will operate as intended and have 2 snapshots for the same dataset that is taken 1 hour apart from each other?

sonicaj:
Closing this as discussed with @william-gr, we think it's best to not indulge in this right now. We can look to having this after more discussions with how to move forward. Thank you all for your input

PR #2827, opened 2019-04-02
tkt-84596: Allow remote_ssh_semiautomatic_setup using password instead of token

themylogin:


src/middlewared/middlewared/plugins/keychain.py
```diff
@@ -398,7 +398,8 @@ def generate_ssh_key_pair(self):
         "keychain_remote_ssh_semiautomatic_setup",
         Str("name", required=True),
         Str("url", required=True, validators=[URL()]),
-        Str("token", required=True),
+        Str("token"),
+        Str("password"),
```
william-gr: Is there any benefit in using private=True in these?

PR #2830, opened 2019-04-03
fix(middleward/network): fake all state entries for inexistent interface

william-gr:


src/middlewared/middlewared/plugins/network.py
```diff
@@ -314,6 +314,17 @@ def query(self, filters, options):
                 'link_address': '',
                 'cloned': True,
                 'mtu': 1500,
+                'flags': [].
```
sonicaj: This looks like a typo `.` ?

william-gr: Hah, you got me :P

PR #2833, opened 2019-04-03
tkt-79239: Ensure UPS is not ONLINE at shutdown event

sonicaj:
This commit ensures UPS is not ONLINE when shutdown event is passed by upssched. There are cases where battery/charger issues can result in ups.status being 'OL LB' at the same time. This will ensure that we don't initiate a shutdown if ups is OL.
Ticket: #79239

### DIFF
--- a/src/middlewared/middlewared/plugins/ups.py
+++ b/src/middlewared/middlewared/plugins/ups.py
@@ -206,7 +206,7 @@ class UPSService(SystemServiceService):
                     f'{config["identifier"]} is ONLINE (OL).'
                 )
             else:
-                syslog.syslog(syslog.LOG_INFO, 'upssched-cmd "issuing shutdown"')
+                syslog.syslog(syslog.LOG_NOTICE, 'upssched-cmd "issuing shutdown"')
                 await run('/usr/local/sbin/upsmon', '-c', 'fsd', check=False)
         elif notify_type.lower() in ('email', 'commbad', 'commok'):
             if config['emailnotify']:
--- a/src/middlewared/middlewared/plugins/ups.py
+++ b/src/middlewared/middlewared/plugins/ups.py
@@ -178,11 +178,37 @@ class UPSService(SystemServiceService):
         Str('notify_type')
     )
     async def upssched_event(self, notify_type):
+        config = await self.config()
+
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/ups.py
```diff
@@ -178,11 +178,32 @@ def driver_choices(self):
...
+                fr'ups.status: (.*)',
+                '' if not stats_output else stats_output.decode()
+            )
+            if ups_status and 'ol' in ups_status[0].lower():
+                self.middleware.logger.debug(
+                    f'Shutdown not initiated as ups.status ({ups_status[0]}) indicates '
+                    f'{config["identifier"]} is ONLINE (OL).'
+                )
+            else:
+                syslog.syslog(syslog.LOG_INFO, 'upssched-cmd "issuing shutdown"')
```
themylogin: @william-gr we do use syslog in middlewared, maybe we can start getting rid of it now?

src/middlewared/middlewared/plugins/ups.py
```diff
@@ -178,11 +178,32 @@ def driver_choices(self):
...
+
         if notify_type.lower() == 'shutdown':
-            syslog.syslog(syslog.LOG_INFO, 'upssched-cmd "issuing shutdown"')
-            await run('/usr/local/sbin/upsmon', '-c', 'fsd', check=False)
+            # Before we start FSD with upsmon, lets ensure that ups is not ONLINE (OL).
+            # There are cases where battery/charger issues can result in ups.status being "OL LB" at the
+            # same time. This will ensure that we don't initiate a shutdown if ups is OL.
+            stats_output = (
+                await run(
+                    '/usr/local/bin/upsc', f'{config["identifier"]}@localhost:{config["remoteport"]}',
```
themylogin: Does this work when we are in the slave mode? (I don't know)

src/middlewared/middlewared/plugins/ups.py
william-gr: We do not, we only use file logging.

If I am not mistaken that syslog was supposed to log to console. Weird that I am not seeing an `openlog`

sonicaj:
backport

PR #2836, opened 2019-04-03
Fix 82915

anodos325:
Create middleware plugins for kerberos, idmap, and active directory.
idmap:
- idmap backends for multiple trusted domains is now mostly there (additional work needed on smb4.conf generation and UI).

kerberos:
- Validation and debugging will probably be needed for edge cases with LDAP.

active directory:
- removed unecessary paramters.
- added autodetection for AD site, workgroup, removed dependencies on a single domain controller.
- added autoconfiguration of krb5.conf with multiple servers in AD site (so that we don't talk to kerberos servers on other side of the world).
- added ability to specify the OU where the AD computer object will be created.
- Introduced new default behavior to generate a machine account keytab file and clear the service account password once we successfully join AD. This is to reduce reliance on a privileged user account.


gui/directoryservice/migrations/0011_add_new_idmap_model.py
```diff
@@ -0,0 +1,259 @@
...
+            model_name='idmap_tdb2',
+            name='idmap_tdb2_domain',
+            field=models.OneToOneField(
+                on_delete=django.db.models.deletion.CASCADE, 
+                to='directoryservice.idmap_domain', 
+                to_field='idmap_domain_name', 
+                null=True, verbose_name='pre-Windows 2000 Domain Name'
+            ),
+        ), 
+        migrations.RunPython(migrate_to_new_idmap, reverse_code=revert_idmap_changes),
```
william-gr: You can use `migrations.RunPython.noop` instead of an empty method.

gui/directoryservice/views.py
```diff
@@ -205,33 +205,13 @@ def directoryservice_idmap_ad(request, id):
             return JsonResp(request, form=form)
     else:
         form = forms.idmap_ad_Form(instance=idmap_ad)
+        log.debug(form)
```
william-gr: Leftover debug

gui/directoryservice/views.py
```diff
@@ -360,6 +319,7 @@ def directoryservice_idmap_rfc2307(request, id):
 
 def directoryservice_idmap_rid(request, id):
     idmap_rid = models.idmap_rid.objects.get(id=id)
+    log.debug('got here')
```
william-gr: Leftover debug

src/middlewared/middlewared/alert/source/active_directory.py
```diff
@@ -1,13 +1,38 @@
-from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, FilePresenceAlertSource
+from datetime import timedelta
+import os
+import logging
+from middlewared.alert.base import Alert, AlertLevel, ThreadedAlertSource
+from middlewared.alert.schedule import IntervalSchedule
 
+log = logging.getLogger("activedirectory_check_alertmod")
 
-class ActiveDirectoryDomainBindAlertClass(AlertClass):
```
william-gr: @themylogin can you verify this file please? It seems its reverting your alert class required changes?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+        """
+        servers = []
+        if not self.ad['domainname']:
+            return servers 
+
+        if self.ad['site'] and self.ad['site'] != 'Default-First-Site-Name':
+            if 'msdcs' in srv_prefix.value: 
+                parts = srv_prefix.value.split('.')
+                srv = '.'.join([parts[0],parts[1]])
+                msdcs = '.'.join([parts[2], parts[3]])
```
william-gr: Is there a guarantee we always get 4 parts?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+                    server.port = 636
+
+        return servers
+        
+    def get_n_working_servers(self, srv=SRV['DOMAINCONTROLLER'], number=1):
+        servers = self._get_servers(srv)
+        found_servers = []
+        for server in servers:
+            if len(found_servers) == number:
+                continue
```
william-gr: It seems we should `break` here, not continue

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+
+            host = server.target.to_text(True)
+            port = int(server.port)
+            if self.port_is_listening(host, port, timeout=1):
+                server_info = {'host': host, 'port': port}
+                found_servers.append(server_info)
+
+        if self.ad['verbose_logging']:
+            self.logger.debug(f'Request for [{number}] of server type [{srv.name}] returned: {found_servers}')
```
william-gr: For future reference, we should use % formatting in logger because it will be useful to aggregate the same kind of message but with different args.

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+
+class SSL(enum.Enum):
+    NOSSL = 'off'
+    USESSL = 'on'
+    USETLS = 'start_tls'
+
+
+class ActiveDirectory_DNS(object):
+    """
+    :get_n_working_servers: often only a few working servers are needed and not the whole
```
william-gr: We usually put the method comments within the method, not in the class, unless I am missing some intent of putting it here

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+        self.logger = kwargs.get('logger')
+        self.pagesize = 1024
+        self._isopen = False
+        self._handle = None
+        self._rootDSE = None
+        self._rootDomainNamingContext = None
+        self._configurationNamingContext = None
+        self._defaultNamingContext = None
+        
+        return
```
william-gr: Return is not necessary

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+                        ldap.OPT_X_TLS_ALLOW
+                    )
+
+                if SSL(self.ad['ssl']) == SSL.USESSL:
+                    try:
+                        self._handle.start_tls_s()
+
+                    except ldap.LDAPError as e:
+                        raise CallError(e)
+                        continue
```
william-gr: If we are raising the continue will be useless here

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+        self._isopen = False
+        if self._handle:
+            self._handle.unbind()
+            self._handle = None
+
+    def _search(self, basedn='', scope=ldap.SCOPE_SUBTREE, filter='', timeout=-1, sizelimit=0):
+        if not self._handle:
+            self._open
+
+
```
william-gr: Double line break

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+                except ldap.LDAPError as e:
+                    self._logex(e)
+                    break
+
+                results.append(data)
+
+            for i in range(len(results)):
+                for entry in results[i]:
+                    result.append(entry)
```
william-gr: is results[i] a list? Perhaps it could be `result += results[i]` or `result += list(results[i])`

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+        if ad['kerberos_realm']:
+            ad['kerberos_realm'] = ad['kerberos_realm']['id']
+
+        return ad 
+
+    @accepts(Dict(
+        'activedirectory_update',
+        Str('domainname', required=True),
+        Str('bindname'),
+        Str('bindpw'),
```
william-gr: private=True so password is not leaked in logs

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+        Str('ssl', default='off', enum=['off', 'on', 'start_tls']),
```
william-gr: We have an unspoken rule of using uppercase for all enums in middleware, would it be hard to switch here?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+        Dict('certificate'),
```
william-gr: Why certificate is a dict? Do you mean Int?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+
+        """
+        'workgroup' is the 'pre-Windows 2000 domain name'. It must be set to the nETBIOSName value in Active Directory.
+        This must be properly configured in order for Samba to work correctly as an AD member server.
+        'site' is the ad site of which the NAS is a member. If sites and subnets are unconfigured this will
+        default to 'Default-First-Site-Name'.
+        """
+
+        if not ad['site']:
+            new_site = await asyncio.wait_for(self.get_site(), 10)
```
william-gr: This may raise an exception on timeout, is this handled/expected?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+        await self.middleware.call('service.restart', 'cifs')
+        await self.middleware.call('etc.generate', 'pam')
+        await self.middleware.call('etc.generate', 'nss')
+        await self.middleware.call('cache.pop', 'AD_State')
+
+    @private
+    async def validate_credentials(self):
+        ret = False
+        ad = await self.config()
+        with ActiveDirectory_DNS(conf=ad, logger=self.logger) as AD_DNS:
```
william-gr: This dos synchronous calls, correct? If so `validate_credentials` should be a method, not a coroutine

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+        if not dcs:
+            raise CallError('Failed to open LDAP socket to any DC in domain.')
+
+        with ActiveDirectory_LDAP(ad_conf=ad, logger=self.logger, hosts=dcs) as AD_LDAP:
+            ret = AD_LDAP.validate_credentials()
+
+        return ret
+
+    @accepts()
+    async def check_clockskew(self):
```
william-gr: This should be a method because it dos synchronous calls.

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+            self.logger.debug(f'net ads testjoin failed with error: [{errout}]')
+            if '0xfffffff6' in errout: 
+                return neterr.NOTJOINED
+            else:
+                return neterr.FAULT
+
+        return neterr.JOINED
+
+    @private
+    async def get_netbios_domain_name(self):
```
william-gr: I think this should also be a method

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+            ret = AD_LDAP.get_netbios_name()
+
+        if ret and smb['workgroup'] != ret:
+            self.logger.debug(f'Updating SMB workgroup to match the short form of the AD domain [{ret}]')
+            await self.middleware.call('datastore.update', 'services.cifs', smb['id'], {'cifs_srv_workgroup': ret})
+
+        return ret
+
+    @private
+    async def get_kerberos_servers(self):
```
william-gr: Same here, a method

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+    async def get_site(self):
+        """
+        First, use DNS to identify domain controllers
+        Then, find a domain controller that is listening for LDAP connection if this information is not cached.
+        Then, perform an LDAP query to determine our AD site
+        """
+        ad = await self.middleware.call('activedirectory.config')
+        i = await self.middleware.call('interfaces.query')
+        dcs = await self._get_cached_srv_records(SRV['DOMAINCONTROLLER'])
+        self.logger.debug(dcs)
```
william-gr: Leftover debug?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+        if await self.middleware.call('cache.has_key', 'ad_cache') and not force:
+            raise CallError('AD cache already exists. Refusing to generate cache.')
+        ad = await self.config()
+        smb = await self.middleware.call('smb.config')
+        if not ad['disable_freenas_cache']:
+            """
+            These calls populate the winbindd cache
+            """
+            pwd.getpwall()
+            grp.getgrall()
```
william-gr: We should run these in a thread

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1147 @@
...
+            return []
+
+        ret = []
+        for key, val in ad_cache.items():
+            ret.extend(val[entry_type])
+
+        return ret
+
+    @private
+    async def get_uncached_userorgroup_legacy(self, entry_type='users', obj=None):
```
william-gr: Should be a method

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1103 @@
...
+        """
+        sorted_idmaps = await self.get_configured_idmap_domains()
+        low_range = sorted_idmaps[-1]['backend_data']['range_high'] + 1
+        high_range = sorted_idmaps[-1]['backend_data']['range_high'] + 100000000
+        return (low_range, high_range)
+
+    @accepts()
+    async def clear_idmap_cache(self):
+        await self.middleware.call('service.stop', 'smb')
+        os.remove('/var/db/samba4/winbindd_cache.tdb')
```
william-gr: Perhaps its safer to have a try/except here?

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1103 @@
...
+            if len(c) == 6 and c[0] != smb['workgroup']:
+                await self.idmap.domain.create({'name': c[0], 'dns_domain_name': c[1]})
+
+class IdmapDomainService(CRUDService):
+    class Config:
+        datastore = 'directoryservice.idmap_domain'
+        datastore_prefix = 'idmap_domain_'
+        namespace = 'idmap.domain'
+
+
```
william-gr: Double linebreak

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1103 @@
...
+    @accepts(Int('id'))
+    async def do_delete(self, id):
+        if id <= 5:
+            entry = await self._get_instance(id)
+            raise CallError(f'Deletion of system idmap domain [{entry["name"]}] is not permitted.', errno.EPERM)
+        await self.middleware.call("datastore.delete", self._config.datastore, id)
+
+    @accepts(Int("id"))
+    async def run(self, id):
```
william-gr: What is this for?

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1103 @@
...
+        """
+        Enforce referential integrity during domain-to-backend creation
+        by also creating a corresponding entry in the correct idmap backend table.
+        """
+        verrors = ValidationErrors()
+        if verrors:
+            raise verrors
+
+        data = await self.middleware.call('idmap.common_backend_compress', data)
+        self.logger.debug(data)
```
william-gr: Leftover debug

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1103 @@
...
+            "idmap_domaintobackend_update",
+            ("attr", {"update": True})
+        )
+    )
+    async def do_update(self, id, data):
+        old = await self._get_instance(id)
+        new = old.copy()
+        new.update(data)
+        verrors = ValidationErrors()
+
```
william-gr: Is there any validation missing? if not we should remove verrors

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1103 @@
...
+            Str('certificate'),
+            register=True
+        )
+    )
+    async def do_create(self, data):
+        verrors = ValidationErrors()
+        if verrors:
+            raise verrors
+
+        self.logger.debug(data)
```
william-gr: Leftover debug, missing validation?

src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -0,0 +1,613 @@
...
+
+class KerberosService(ConfigService):
+    """
+    :start:  - configures kerberos and performs kinit if needed
+    :stop:   - performs kdestroy and clears cached klist output
+    :status: - returns false if there is no kerberos tgt or if it is expired 'klist -t'
+    :renew:  - compares current time with expiration timestamp in tgt. Issues 'kinit -R' if needed.
+               uses cached klist output if available.
+    :update: - this modifies the krb5.conf. There is a rudamentary parser for the auxiliary
+               parameters.
```
william-gr: These need to be documented in their respective methods to be part of the API documentation. In fact we need every public method to be documented.

src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -0,0 +1,613 @@
...
+            await self.middleware.call('cache.put', 'KRB_TGT_INFO', {'ad_TGT': ad_TGT, 'ldap_TGT': ldap_TGT})
+        return {'ad_TGT': ad_TGT, 'ldap_TGT': ldap_TGT}
+        
+    @private
+    async def renew(self):
+        """
+        Compare timestamp of cached TGT info with current timestamp. If we're within 5 minutes
+        of expire time, renew the TGT via 'kinit -R'.
+        """
+        self.logger.debug('entered kinit renew')
```
william-gr: Leftover debugs

src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -0,0 +1,613 @@
...
+
+    @private
+    async def kerberos_keytab_compress(self, data):
+        data['file'] = await self.middleware.call('pwenc.encrypt', data['file'])
+        return data 
+
+    @accepts(
+        Dict(
+            'kerberos_keytab_create',
+            Any('file'),
```
william-gr: Should be Str then if its b64encoded?

gui/directoryservice/forms.py
```diff
@@ -243,6 +217,21 @@ class ActiveDirectoryForm(ModelForm):
...
+    ad_kerberos_principal = forms.ChoiceField(
+        label=models.ActiveDirectory._meta.get_field('ad_kerberos_principal').verbose_name,
+        required=False,
+        choices=choices.KERBEROS_PRINCIPAL_CHOICES(),
+        help_text=_(
+            "Kerberos principal to use for AD-related UI and middleware operations "
+            "Field is populated with principals present in the system keytab. "
+            "During the domain join process a keytab entry is generated for the "
+            "AD Machine Account associated with the NAS. The name for this account "
+            "is the netbios name of the server with a '$' appended to it. Once "
```
Mrt134: "The name for this account is the netbios name of the server with a '$' appended to it." -->
"The account name is the server netbios name appended with '$'.

gui/directoryservice/forms.py
```diff
@@ -243,6 +217,21 @@ class ActiveDirectoryForm(ModelForm):
...
+            "the NAS is joined to active directory, the bind credentials will be "
+            "automatically cleared and all future operations carried out by the AD "
+            "machine account, which has a restricted set of privileges in the AD domain."),
```
Mrt134: "Once the NAS is joined to active directory, the bind credentials will be automatically cleared and all future operations carried out by the AD machine account, which has a restricted set of privileges in the AD domain." --> 

"Bind credentials are automatically cleared after the NAS is joined to active directory. Later operations are done by the AD machine account, which has restricted privileges in the AD domain."

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1064 @@
...
+
+    @accepts(Int("id"))
+    async def run(self, id):
+        data = await self._get_instance(id)
+
+    @private
+    async def _validate(self, data):
+        verrors = ValidationErrors()
+        if data['id'] <= dstype['DS_TYPE_DEFAULT_DOMAIN'].value:
+            verrors.append(f'Modification of system idmap domain [{data["name"]}] is not permitted.')
```
a-stjohn:
```suggestion
            verrors.append(f'Modifying system idmap domain [{data["name"]}] is not permitted.')
```

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1064 @@
...
+            new,
+            {'prefix': self._config.datastore_prefix}
+        )
+        return await self._get_instance(id)
+
+    @accepts(Int('id'))
+    async def do_delete(self, id):
+        entry = await self._get_instance(id)
+        if entry['domain']['id'] <= dstype['DS_TYPE_DEFAULT_DOMAIN'].value: 
+            raise CallError(f'Deletion of mapping for [{entry["domain"]["idmap_domain_name"]}] is not permitted.', errno.EPERM)
```
a-stjohn:
```suggestion
            raise CallError(f'Deleting mapping for [{entry["domain"]["idmap_domain_name"]}] is not permitted.', errno.EPERM)
```

gui/directoryservice/migrations/0011_add_new_idmap_model.py
```diff
@@ -0,0 +1,259 @@
...
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('idmap_domain_name', models.CharField(
+                    max_length=120,
+                    unique=True,
+                    help_text=(
+                        'Short form of domain name as represented by the nETBIOSName '
+                        'LDAP entry in an Active Directory domain (commonly indicated as the '
+                        '"pre-Windows 2000" domain name). This must not be confused with the '
+                        'netbios host name of the server.'
```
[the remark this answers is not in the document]
anodos325: Host name and netbios name are not the same thing.

gui/api/resources.py
```diff
@@ -4066,12 +4057,7 @@ def get_list(self, request, **kwargs):
         wizard_ds = request.session.get('wizard_ds')
         if request.GET.get('wizard') == '1' and wizard_ds:
             if wizard_ds.get('ds_type') == 'ad':
-                wizard_users = FreeNAS_ActiveDirectory_Users(
-                    domainname=wizard_ds.get('ds_ad_domainname'),
-                    bindname=wizard_ds.get('ds_ad_bindname'),
-                    bindpw=wizard_ds.get('ds_ad_bindpw'),
-                    flags=FLAGS_DBINIT,
-                )
+                wizard_users = None
```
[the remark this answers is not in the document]
anodos325: I initially disabled this with intention of re-enabling. The user caching is now mostly passive (we rely on actually being joined to the domain and winbind's internal cache), which means that the wizard won't show AD users until after the domain join is complete, which reduces the usefulness of building out the list of available users in this way. I waffled about it for a while then decided to re-add the AD user support.

gui/common/freenasusers.py
```diff
@@ -303,15 +358,20 @@ def __init__(self, **kwargs):
...
+
         elif dflags & U_NIS_ENABLED:
             dir = FreeNAS_NIS_Users
         elif dflags & U_LDAP_ENABLED:
             dir = FreeNAS_LDAP_Users
 
         if dir is not None:
             try:
-                self.__users = dir(**kwargs)
+                self.__users = dir
```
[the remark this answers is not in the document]
anodos325: Me too? Should be added back now.

gui/directoryservice/forms.py
```diff
@@ -350,367 +283,43 @@ def __init__(self, *args, **kwargs):
...
+        smb_update = {}
+        with client as c:
+            smb = c.call('smb.config')
+            old = c.call('activedirectory.config')
+            if smb['netbiosname'] != data['netbiosname_a']:
+                smb_update.update({'netbiosname': data['netbiosname_a']})
+            if 'netbiosname_b' in data and smb['nebiosname_b'] != data['netbiosname_b']:
+                smb_update.update({'netbiosname_b': data['netbiosname_b']})
+            if smb['netbiosalias'] != data['netbiosalias'].split():
+                smb_update.update({'netbiosalias': data['netbiosalias']})
```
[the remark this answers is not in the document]
anodos325: This is now moved to the middleware plugin.

gui/directoryservice/forms.py
```diff
@@ -350,367 +283,43 @@ def __init__(self, *args, **kwargs):
...
-            if started is True:
-                timeout = _fs().directoryservice.activedirectory.timeout.stop
+        if data['kerberos_principal'] == '----------':
+            data['kerberos_principal'] = ''
+        if data['enable'] and not old['enable']:
+            with client as c:
                 try:
-                    started = notifier().stop("activedirectory", timeout=timeout)
+                    c.call('activedirectory.validate_credentials')
+                    c.call('activedirectory.validate_domain')
```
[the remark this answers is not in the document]
anodos325: Concur . Moved.

gui/directoryservice/forms.py
```diff
@@ -1036,61 +643,8 @@ def clean_keytab_file(self):
...
-
-        return ret
-
     def save(self):
         obj = super(KerberosKeytabCreateForm, self).save()
-        if not self.save_principals(obj):
-            obj.delete()
-            log.debug("save(): unable to save principals")
-
         notifier().start("ix-kerberos")
```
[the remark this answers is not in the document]
anodos325: Done. Removed other instances where we were using ix-kerberos. Remaining instances on LDAP side will be removed when I work on that plugin.

gui/directoryservice/forms.py
```diff
@@ -1114,19 +668,19 @@ def save(self):
...
```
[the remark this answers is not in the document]
anodos325: Done.

gui/directoryservice/forms.py
```diff
@@ -1114,19 +668,19 @@ def save(self):
...
         model = models.KerberosSettings
 
-    def save(self):
-        super(KerberosSettingsForm, self).save()
-        notifier().start("ix-kerberos")
+    def __init__(self, *args, **kwargs):
+        super(KerberosSettingsForm, self).__init__(*args, **kwargs)
+
+    def middleware_clean(self, data):
+        return data
```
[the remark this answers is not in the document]
anodos325: Done.

gui/directoryservice/migrations/0011_add_new_idmap_model.py
anodos325: Good to know. Thanks!

gui/directoryservice/views.py
anodos325: Removed

gui/directoryservice/views.py
anodos325: Removed

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: Yes.

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: Agreed

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: Okay. Consolidated the comments to the method itself.

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: Removed

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: Done. Good point.

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: Yes.

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: Added some handling here.

src/middlewared/middlewared/plugins/idmap.py
anodos325: Removed various unnecessary validation stubs. I will add validation on an as-needed basis.

src/middlewared/middlewared/plugins/kerberos.py
anodos325: Okay. Done.

src/middlewared/middlewared/plugins/kerberos.py
anodos325: Fixed

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: Yeah, not sure what I was thinking here. I've decided to come down on one side of the fence instead of trying to lie down on it.

gui/directoryservice/views.py
william-gr: Hmm, it seems its still there

gui/directoryservice/views.py
william-gr: Seems debug still there

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1292 @@
...
+        if timeout:
+            s.settimeout(timeout)
+
+        try:
+            s.connect((host, port))
+            ret = True
+
+        except Exception as e:
+            s.close()
+            raise CallError(e)
```
william-gr: No necessary to change, but I think e could have a finally:

```
try:
    ...
except:
    ...
finally:
    s.close()
```

and remove the other closes

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1292 @@
...
+
+    def _close(self):
+        self._isopen = False
+        if self._handle:
+            self._handle.unbind()
+            self._handle = None
+
+    def _search(self, basedn='', scope=ldap.SCOPE_SUBTREE, filter='', timeout=-1, sizelimit=0):
+        if not self._handle:
+            self._open
```
william-gr: You meant `self._open()` ?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1292 @@
...
+
+        if verrors:
+            raise verrors
+
+        start = False
+        stop = False
+
+        if old['idmap_backend'] != new['idmap_backend']:
+            idmap = await self.middleware.call('idmap.domaintobackend.query', [('domain', '=', 'DS_TYPE_ACTIVEDIRECTORY')])
+            self.logger.debug(idmap[0]['id'])
```
william-gr: Leftover debug?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1292 @@
...
+            await self.middleware.call('service.restart', 'cifs')
+
+        return await self.config()
+
+    @private
+    async def _set_state(self, state):
+        await self.middleware.call('cache.put', 'AD_State', state.name)
+
+    @private
+    async def get_state(self):
```
william-gr: Its weird that _set_state starts with underscode but get_state doesnt? : P

src/middlewared/middlewared/plugins/activedirectory.py
william-gr: You may have missed to turn this into a method? Since there are sync calls being done here

src/middlewared/middlewared/plugins/activedirectory.py
william-gr: And here

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1292 @@
...
+        if len(ntp_servers) != 3 and len(default_ntp_servers) != 3:
+            return
+
+        ad = self.middleware.call_sync('activedirectory.config')
+        with ActiveDirectory_DNS(conf=ad, logger=self.logger) as AD_DNS:
+            pdc = AD_DNS.get_n_working_servers(SRV['PDC'], 1)
+        self.middleware.call_sync('system.ntpserver.create', {'address': pdc[0]['host'], 'prefer': True})
+
+    @private
+    async def get_site(self):
```
william-gr: And here, ActiveDirectory_DNS being synchronous

src/middlewared/middlewared/plugins/activedirectory.py
william-gr: Or likely turn this into a method, because there are other pwd/grp calls below

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1128 @@
...
+    that exists in the domaintobackend table and has a corresponding backend configured in the respective
+    idmap_{backend} table. List is sorted based in ascending order based on the id range.
+
+    :clear_idmap_cache: - removes samba's idmap cache. This may be required if idmap settings are changed
+    after the server has entered production. This will cause a service disruption.
+
+    There are default system domains 'DS_TYPE_ACTIVEDIRECTORY', 'DS_TYPE_LDAP', 'DS_TYPE_DEFAULT_DOMAIN',
+    which exist for compatibility reasons and correspond with the idmap backend that is configured
+    under 'Active Directory', 'LDAP', and 'SMB' respectively.
+    """
```
william-gr: Same here about having the method comments within their down docstring

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1128 @@
...
+                Str('idmap_domain_name'),
+                Str('idmap_domain_dns_domain_name'),
+            ),
+            Int('range_low', required=True, validators=[Range(min=1000, max=2147483647)]),
+            Int('range_high', required=True, validators=[Range(min=1000, max=2147483647)]),
+            Str('base_dn'),
+            Str('user_dn'),
+            Str('url'),
+            Str('ssl', default='off', enum=['off', 'on', 'start_tls']),
+            Str('certificate'),
```
william-gr: Should certificate be an Int instead?

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1128 @@
...
+            Str('bind_path_group'),
+            Str('user_cn'),
+            Str('cn_realm'),
+            Str('ldap_domain'),
+            Str('ldap_url'),
+            Str('ldap_user_dn'),
+            Str('ldap_user_dn_password'),
+            Str('ldap_realm'),
+            Str('ssl', default='off', enum=['off', 'on', 'start_tls']),
+            Str('certificate'),
```
william-gr: certificate as Int?

src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -0,0 +1,639 @@
...
+        try:
+            klist = await asyncio.wait_for(
+                run(['/usr/bin/klist', '-v'], check=False, stdout=subprocess.PIPE),
+                timeout=10.0
+            )
+            if klist.returncode != 0:
+                raise CallError(f'klist failed with error: {klist.stderr.decode()}')
+        except asyncio.TimeoutError:
+            self.logger.debug('klist attempt failed after 10 seconds.')
+            await self._kdestroy
```
william-gr: I dont see a _kdestroy defined?

william-gr:
@themylogin would you care to review this one too? Its huge and very easy to miss things, so the more eyes the better ;)

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: get_state() shouldn't be private. So I'll fix that.

src/middlewared/middlewared/plugins/kerberos.py
anodos325: It used to be. :) switched to self.stop()

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1128 @@
...
+    async def autodiscover_trusted_domains(self):
+        smb = await self.middleware.call('smb.config')
+        wbinfo = await run(['/usr/local/bin/wbinfo', '-m', '--verbose'], check=False)
+        if wbinfo.returncode != 0:
+            raise CallError(f'wbinfo -m failed with error: {wbinfo.stderr.decode().strip()}')
+
+        for entry in wbinfo.stdout.decode().splitlines():
+            c = entry.split()
+            if len(c) == 6 and c[0] != smb['workgroup']:
+                await self.idmap.domain.create({'name': c[0], 'dns_domain_name': c[1]})
```
themylogin: Where is `self.idmap` defined?

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1128 @@
...
+        if id <= 5:
+            entry = await self._get_instance(id)
+            raise CallError(f'Deleting system idmap domain [{entry["name"]}] is not permitted.', errno.EPERM)
+        await self.middleware.call("datastore.delete", self._config.datastore, id)
+
+    @private
+    async def _validate(self, data):
+        verrors = ValidationErrors()
+        if data['id'] <= dstype['DS_TYPE_DEFAULT_DOMAIN'].value:
+            verrors.append(f'Modifying system idmap domain [{data["name"]}] is not permitted.')
```
themylogin: `verrors.add`?

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1128 @@
...
+        if verrors:
+            raise verrors
+
+        backend_entry_is_present = False
+        idmap_data = await self.middleware.call(f'idmap.{data["idmap_backend"]}.query')
+        for i in idmap_data:
+            if not i['domain']:
+                continue
+            if i['domain']['idmap_domain_name'] == data['domain']['idmap_domain_name']:
+                backend_entry_is_present
```
themylogin: Did you want to say `backend_entry_is_present = True`?

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1128 @@
...
+                break
+
+        if not backend_entry_is_present:
+            next_idmap_range = await self.get_next_idmap_range()
```
themylogin: This method is not defined on this class

src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -0,0 +1,639 @@
...
+                if ad_kinit.returncode != 0:
+                    raise CallError(f"kinit for domain [{ad['domainname']}] with principal [{ad['kerberos_principal']}] failed: {ad_kinit.stderr.decode()}")
+            else:
+                principal = f'{ad["bindname"]}@{ad["domainname"]}'
+                ad_kinit = await Popen(
+                    ['/usr/bin/kinit', '--renewable', '--password-file=STDIN', principal],
+                    stdout=subprocess.PIPE, stderr=subprocess.PIPE, stdin=subprocess.PIPE
+                )
+                output = await ad_kinit.communicate(input=ad['bindpw'].encode())
+                if ad_kinit.returncode != 0:
```
themylogin: This will never be reached, you've probably forgot to add `check=False`

src/middlewared/middlewared/plugins/kerberos.py
themylogin: You probably also need to `return` here because you are using `klist` which may not be present if exception happens. Also asyncio timeout exception is not handled.

src/middlewared/middlewared/plugins/service_monitor.py
```diff
@@ -1,315 +0,0 @@
...
-            #
-            # We should probably have a configurable threshold for number of
-            # failures before starting or stopping the service
-            #
-            if self.finished.is_set():
-                # Thread.cancel() takes a while to propagate here
-                ServiceMonitorThread.reset_alerts(service)
-                return
-
-            if os.path.exists('/tmp/.ad_start'):
```
themylogin: So we can also drop `ServiceMonitorAlertSource` that checks for these files. But will we have a replacement?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1292 @@
...
+        foreign entries.
+        kinit will fail if domain name is lower-case.
+        """
+        if ad['kerberos_realm']:
+            ad['kerberos_realm'] = ad['kerberos_realm']['id']
+        ad['domainname'] = ad['domainname'].upper()
+
+        smb_ha_mode = await self.middleware.call('smb.get_smb_ha_mode')
+        if smb_ha_mode == 'STANDALONE':
+            await self.middleware.call(
```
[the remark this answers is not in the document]
anodos325: In the big picture, the GUI needs to display the netbios name to the user in the AD form. We can do it in middleware or let the UI team handle it. I'm rather ambivalent toward it and happy to go whichever way you think is best.

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1292 @@
...
+        if not new["bindname"] and not new["kerberos_principal"]:
+            verrors.add(f"activedirectory_update.bindname", "Bind credentials or kerberos keytab are required to join an AD domain.")
+
+        if data['bindpw'] and data['enable'] and not old['enable']:
+            try:
+                await self.middleware.run_in_thread(self.validate_credentials)
+            except Exception as e:
+                verrors.add(f"activedirectory_update.bindpw", f"Failed to validate bind credentials: {e}")
+
+            await self.middleware.run_in_thread(self.validate_domain)
```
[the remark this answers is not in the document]
anodos325: No. That's an oversight.

src/middlewared/middlewared/plugins/activedirectory.py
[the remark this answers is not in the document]
themylogin: I am not sure I catch how displaying something in the GUI and this method are related?

src/middlewared/middlewared/plugins/service_monitor.py
anodos325: I was planning to rip out the service monitor. It seems redundant if there is an adequate level of alerting.

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: These have now been changed to methods

src/middlewared/middlewared/alert/source/active_directory.py
```diff
@@ -1,13 +1,54 @@
...
+class ActiveDirectoryBindAlertClass(AlertClass):
     category = AlertCategory.DIRECTORY_SERVICE
     level = AlertLevel.WARNING
-    title = "ActiveDirectory Did Not Bind to the Domain"
-    text = "ActiveDirectory Did Not bind to the domain."
+    title = "ActiveDirectory Bind is Not Healthy"
```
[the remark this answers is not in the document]
anodos325: Yes.

src/middlewared/middlewared/alert/source/active_directory.py
```diff
@@ -1,13 +1,54 @@
...
+        ad = self.middleware.call_sync("activedirectory.config")
+        if not ad['enable']:
+            return
+
+        try:
+            self.middleware.call_sync("activedirectory.validate_domain")
+        except Exception as e:
+            return Alert(
+                ActiveDirectoryDomainHealthAlertClass,
+                {'verrs': e}
```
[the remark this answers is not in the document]
anodos325: Domain Health Check now contains synchronous code. Domain Bind Check is now async.

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1292 @@
...
+                    self.logger.debug(f'Successfully initialized LDAP server: [{uri}]')
+
+                res = None
+                ldap.protocol_version = ldap.VERSION3
+                ldap.set_option(ldap.OPT_REFERRALS, 0)
+                ldap.set_option(ldap.OPT_NETWORK_TIMEOUT, 10.0)
+
+                if SSL(self.ad['ssl']) != SSL.NOSSL:
+                    ldap.set_option(ldap.OPT_X_TLS_ALLOW, 1)
+                    if self.certfile:
```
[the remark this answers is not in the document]
anodos325: Yeah, whoops. I will mark this one as resolved once I have an opportunity to validate LDAPs.

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1292 @@
...
+                sizelimit=sizelimit
+            )
+
+            type = ldap.RES_SEARCH_ENTRY
+            while type != ldap.RES_SEARCH_RESULT:
+                try:
+                    type, data = self._handle.result(id, 0)
+
+                except ldap.LDAPError as e:
+                    self._logex(e)
```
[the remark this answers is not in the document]
anodos325: In a file that doesn't exist anymore. Fixed. :)

src/middlewared/middlewared/plugins/activedirectory.py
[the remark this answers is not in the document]
anodos325: Ah, I had a bad case of the Fridays when I wrote that. I will move this to its own method, do validation, and follow standards for ad_compress.

src/middlewared/middlewared/etc_files/smb_configure.py
```diff
@@ -56,6 +56,11 @@
     for parm in parm_to_test:
         conf[parm] = await get_smb4conf_parm(parm)
 
+    if not conf['privatedir']:
+        conf['privatedir'] = '/var/db/system/samba4/private'
+    if not conf['state directory']:
+        conf['state directory'] = '/var/db/system/samba4'
+    logger.debug(conf)
```
themylogin: Leftover debug?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -698,17 +709,18 @@ class Config:
             new,
             {'prefix': 'ad_'}
         )
-        verrors = ValidationErrors()
-        if not new["bindname"] and not new["kerberos_principal"]:
-            verrors.add(f"activedirectory_update.bindname", "Bind credentials or kerberos keytab are required to join an AD domain.")
+        if not new["bindpw"] and not new["kerberos_principal"]:
+            raise ValidationError("activedirectory_update.bindname", "Bind credentials or kerberos keytab are required to join an AD domain.")
```
themylogin: Why do you raise this after update?

src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -568,7 +570,7 @@ class Config:
             if entry['principal'] not in kerberos_principals:
                 kerberos_principals.append(entry['principal'])
 
-        return kerberos_principals
+        return kerberos_principals.sort()
```
themylogin: `sort()` method sorts in-place and returns `None`. You want `return sorted(kerberos_principals)`

src/middlewared/middlewared/plugins/account.py
```diff
@@ -570,22 +570,24 @@ def shell_choices(self):
...
+                data['smbhash'] = f'{data["username"]}:{data["uid"]}:{"X" * 32}:{nt_password(password)}:[DU         ]:LCT-{int(time.time()):X}:'
+            else:
+                data['smbhash'] = f'{data["username"]}:{data["uid"]}:{"X" * 32}:{nt_password(password)}:[U          ]:LCT-{int(time.time()):X}:'
         else:
             data['unixhash'] = '*'
             data['smbhash'] = '*'
         return password
 
-    async def __set_smbpasswd(self, username, password):
+    def __set_smbpasswd(self, username):
```
themylogin: Why is this method synchronous? It does not do anything synchronous and it is always called from asynchronous context.

src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -0,0 +1,643 @@
...
+            klist = await asyncio.wait_for(
+                run(['/usr/bin/klist', '-v'], check=False, stdout=subprocess.PIPE),
+                timeout=10.0
+            )
+        except Exception as e:
+            await self.stop()
+            raise CallError("Attempt to list kerberos tickets failed with error: %s", e)
+
+        if klist.returncode != 0:
+            raise CallError(f'klist failed with error: {klist.stderr.decode()}')
```
themylogin: Don't you also need `await self.stop()` here?

src/middlewared/middlewared/alert/source/active_directory.py
```diff
@@ -1,13 +1,53 @@
...
-    path = "/tmp/.adalert"
-    klass = ActiveDirectoryDomainBindAlertClass
+class ActiveDirectoryDomainHealthAlertClass(AlertClass):
+    category = AlertCategory.DIRECTORY_SERVICE
+    level = AlertLevel.WARNING
+    title = "ActiveDirectory Domain Validation Failed"
+    text = "Domain validation failed with error: %(verrs)s."
+
+
+class ActiveDirectoryDomainHealthAlertSource(ThreadedAlertSource):
```
themylogin: It should be just `AlertSource` if it does not do anything asynchronous and method should be just `check`

src/middlewared/middlewared/alert/source/active_directory.py
```diff
@@ -1,13 +1,53 @@
...
+    schedule = IntervalSchedule(timedelta(hours=24))
```
themylogin: If we run this every 24 hours, can'it it be better idea to use `CronSchedule` instead and run at the same time every day?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1314 @@
...
+            old['id'],
+            new,
+            {'prefix': 'ad_'}
+        )
+
+        if data['bindpw'] and data['enable'] and not old['enable']:
+            try:
+                await self.middleware.run_in_thread(self.validate_credentials)
+            except Exception as e:
+                verrors.add("activedirectory_update.bindpw", f"Failed to validate bind credentials: {e}")
```
themylogin: Raising `ValidationError` will make an impression that update was not successful when, in fact, `directoryservice.activedirectory` table row was changed. I've checked your code and have not found any obstacles to passing AD config to `self.validate_credentials` and `self.validate_domain` rather than reading it there from database directly. Am I missing something?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1292 @@
...
+        super(ActiveDirectory_DNS, self).__init__()
+        self.ad = kwargs.get('conf')
+        self.logger = kwargs.get('logger')
+        return
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, typ, value, traceback):
+        if typ is not None:
```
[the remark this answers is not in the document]
themylogin: @anodos325 sorry, this is **not** required and, in fact, only works by accident

src/middlewared/middlewared/alert/source/active_directory.py
anodos325: Good idea.

src/middlewared/middlewared/plugins/account.py
anodos325: Fixed.

src/middlewared/middlewared/plugins/kerberos.py
anodos325: Fixed.

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: No. You're not.

src/middlewared/middlewared/alert/source/active_directory.py
themylogin: @anodos325 you've forgot to change this

themylogin: Also another class in this file too

anodos325: Thanks for your help. I think this is now fixed.

src/middlewared/middlewared/etc_files/smb_configure.py
```diff
@@ -376,43 +270,41 @@ def setup_samba_dirs(middleware, conf):
...
+def render(service, middleware):
     conf = {}
-    conf = await get_config(middleware)
+    conf = get_config(middleware)
     if conf['systemdataset']['path'] is None:
         logger.debug("systemdataset.config returned 'None' as dataset path. Possible zpool import in progress. Exiting configure.")
         return
 
-    ret = await middleware.run_in_thread(setup_samba_dirs, middleware, conf)
+    ret = middleware.run_in_thread(setup_samba_dirs, middleware, conf)
```
william-gr: render is not a coroutine anymore so this wont work anymore, you should just run `setup_samba_dirs(middleware, conf)`

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1315 @@
...
+        super(ActiveDirectory_DNS, self).__init__()
+        self.ad = kwargs.get('conf')
+        self.logger = kwargs.get('logger')
+        return
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, typ, value, traceback):
+        return
```
william-gr: I am confused why we are making ActiveDirectory_DNS a context manager. Any future plans?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1315 @@
...
+                        ldap.OPT_X_TLS_REQUIRE_CERT,
+                        ldap.OPT_X_TLS_ALLOW
+                    )
+
+                if SSL(self.ad['ssl']) == SSL.USESSL:
+                    try:
+                        self._handle.start_tls_s()
+
+                    except ldap.LDAPError as e:
+                        self.logger.debug('%s', e)
```
william-gr: If start TLS failed, should we keep going, or short circuit the loop with `continue`?

src/middlewared/middlewared/plugins/idmap.py
anodos325: I agree.

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1315 @@
...
+        command will be issued.
+        """
+        verrors = ValidationErrors()
+        old = await self.config()
+        new = old.copy()
+        new.update(data)
+        try:
+            await self.update_netbios_data(old, new)
+        except Exception as e:
+            raise ValidationError('netbiosname', str(e))
```
william-gr: `activedirectory_update.netbiosname`

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1315 @@
...
+            except KeyError:
+                try:
+                    await self.started()
+                except Exception:
+                    pass
+
+            return (await self.middleware.call('cache.get', 'AD_State'))
+
+    @private
+    async def start(self):
```
william-gr: How long can start take?

I am thinking if perhaps we should let it be a job so we can report progress?

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1120 @@
...
+            Str('idmap_backend', enum=['ad', 'autorid', 'fruit', 'ldap', 'nss', 'rfc2307', 'rid', 'script', 'tdb', 'tdb2']),
+            register=True
+        )
+    )
+    async def do_create(self, data):
+        """
+        Set an idmap backend for a domain.
+        :domain: - dictionary containing domain information. Has one-to-one relationship with idmap_domain entries. id value
+        in dictionary is not required.
+        :idmap_backed: - type of idmap backend to use for the domain.
```
william-gr: We should check how these look in /api/docs/

Noting the attributes with `:attribute:` would probably have no effect on the markup and they will all be concatenated in the same line unless they is a double line break.

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1120 @@
...
+                'domain',
+                Int('id'),
+                Str('idmap_domain_name'),
+                Str('idmap_domain_dns_domain_name'),
+            ),
+            Int('range_low', required=True, validators=[Range(min=1000, max=2147483647)]),
+            Int('range_high', required=True, validators=[Range(min=1000, max=2147483647)]),
+            Str('schema_mode'),
+            Bool('unix_primary_group'),
+            Bool('unix_nss_info'),
```
william-gr: We should have default values set for all `create` in the schema, otherwise it will use whatever is the database default and the API user would have no clue about that.

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -0,0 +1,1120 @@
...
+            'idmap_ad_create',
+            Dict(
+                'domain',
+                Int('id'),
+                Str('idmap_domain_name'),
+                Str('idmap_domain_dns_domain_name'),
+            ),
+            Int('range_low', required=True, validators=[Range(min=1000, max=2147483647)]),
+            Int('range_high', required=True, validators=[Range(min=1000, max=2147483647)]),
+            Str('schema_mode'),
```
william-gr: What are the possible modes? Would we need an `enum` for this?

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -22,6 +27,63 @@
 RE_NETBIOSNAME = re.compile(r"^[a-zA-Z0-9\.\-_!@#\$%^&\(\)'\{\}~]{1,15}$")
 
 
+class smbhamode(enum.Enum):
```
william-gr: In python its sort of an unspoken rule to use starting uppercase for class names to not confuse things.

Could we perhaps use: SmbHAMode or something?
Also, should this be an enum.IntEnum?

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: It won't be too hard, but I should probably postpone until I get the LDAP plugin completed and in 11.3.

src/middlewared/middlewared/plugins/activedirectory.py
william-gr: Let me elaborate a bit more, we should only consider that if joining AD can take a lot of time so the user can have an idea what its doing. Either way, not urgent and can be addressed in another PR. Just wanted to keep that in mind.

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: Even better. I realized that this block of code is unnecessary so I just removed it.

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: That's a good thought. It doesn't take as long as it used to (the largest culprit before was filling the AD cache and initializing everything through freenasldap.py). I think we're at less than 10 seconds assuming a healthy AD environment. Throwing it in a job might be a good idea for the future if this proves to be problematic.

src/middlewared/middlewared/plugins/idmap.py
anodos325: Yes. We do.

src/middlewared/middlewared/plugins/activedirectory.py
william-gr: Fair enough, ~10 seconds is fast enough

gui/directoryservice/views.py
```diff
@@ -205,33 +206,13 @@ def directoryservice_idmap_ad(request, id):
             return JsonResp(request, form=form)
     else:
         form = forms.idmap_ad_Form(instance=idmap_ad)
+        log.debug(form)
```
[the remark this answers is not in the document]
anodos325: Removed now.

gui/services/migrations/0030_remove_service_monitor.py
```diff
@@ -0,0 +1,14 @@
+import django.core.validators
```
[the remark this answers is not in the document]
anodos325: Incremented to 0031 and changed dependencies.

gui/system/forms.py
```diff
@@ -1845,32 +1845,9 @@ def clean(self):
...
-                        error = ', '.join(error)
-                    else:
-                        error = str(e)
-
-                    raise forms.ValidationError("{0}".format(error))
-                except Exception as e:
-                    raise forms.ValidationError("{0}".format(e))
+                with client as c:
+                    if not c.call('activedirectory.validate_credentials'):
+                        raise forms.ValidationError("Failed to validate AD bind credentials")
```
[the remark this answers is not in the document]
anodos325: Good point. I'm now passing the configuration from the wizard to 'validate_credentials'.

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: Stopped making it a context manager.

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -0,0 +1,1278 @@
...
+
+class ActiveDirectory_DNS(object):
+    def __init__(self, **kwargs):
+        super(ActiveDirectory_DNS, self).__init__()
+        self.ad = kwargs.get('conf')
+        self.logger = kwargs.get('logger')
+        return
+
+    def __enter__(self):
+        return self
```
william-gr: Forgot to also remove `__enter__`

PR #2839, opened 2019-04-04
tkt-85019: Save SSH keys in database

sonicaj:
This commit fixes a bug where we referenced freenas_sqlite_cmd variable from a file which no longer can give us access to it ( as that portion was removed in migration of rc.conf to middlewared ).


src/freenas/etc/ix.rc.d/ix_sshd_save_keys
```diff
@@ -10,6 +10,7 @@
 # Save any generated keys by /usr/local/etc/rc.d/sshd into the config db.
 #
 
+. /etc/rc.freenas
 . /etc/rc.subr
```
william-gr: We dont need both, you can remove rc.subr

PR #2841, opened 2019-04-04
Allow decimal input for serialport field

sonicaj:
This commit fixes a bug where we expected serialport to be in hexadecimal always. Now if it is provided in base 10, we normalize it and convert to hexadecimal.

### DIFF
--- a/src/middlewared/middlewared/plugins/system.py
+++ b/src/middlewared/middlewared/plugins/system.py
@@ -102,19 +102,15 @@ class SytemAdvancedService(ConfigService):
                     f'{schema}.serialport',
                     'Please specify a serial port when serial console option is checked'
                 )
-            elif serial_choice not in await self.serial_port_choices():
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/system.py
```diff
@@ -102,19 +102,15 @@ class Config:
...
-            # TODO: THIS CHECK CAN BE REMOVED WHEN WE DISALLOW NONE VALUES IN THE SCHEMA LAYER
-
-            verrors.add(
-                f'{schema}.serialport',
-                'Empty serial port is not allowed'
-            )
+            else:
+                data['serialport'] = serial_choice = hex(
+                    int(serial_choice)
+                ) if serial_choice.isnumeric() else serial_choice
```
william-gr: You want isigit, not isnumeric. The latter will allow unicode numeric chars.

PR #2842, opened 2019-04-04
tkt-85012: Correctly write multiple isns servers

sonicaj:
This commit fixes a bug where when we wrote isns-servers, we split on spaces only.
Ticket: #85012


sonicaj:
backport

PR #2844, opened 2019-04-05
tkt-83924: Normalize/Validate Crontab Commands

sonicaj:
This PR aims to validate cron commands making sure they donot exceed 1000 characters as FreeBSD enforces a hard limit. It also normalizes how we write commands in crontab ensuring that we donot insert unnecessary spaces which consume 1000characters buffer.

Ticket: #83924

### DIFF
--- a/src/middlewared/middlewared/plugins/cron.py
+++ b/src/middlewared/middlewared/plugins/cron.py
@@ -17,6 +17,20 @@ class CronJobService(CRUDService):
         return data
 
     @private
+    async def construct_cron_command(self, schedule, user, command, stdout=True, stderr=True):
+        return list(
+            filter(
+                bool, (
+                    schedule['minute'], schedule['hour'], schedule['dom'], schedule['month'],
+                    schedule['dow'], user,
+                    'PATH="/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin:/root/bin"',
+                    command.replace('\n', '').replace('%', r'\%'),
+                    '> /dev/null' if stdout else '', '2> /dev/null' if stderr else ''
+                )
+            )
+        )
+
+    @private
     async def validate_data(self, data, schema):
         verrors = ValidationErrors()
 
@@ -41,14 +55,37 @@ class CronJobService(CRUDService):
                     'Specified user does not exist'
                 )
 
+        command = data.get('command')
+        if not command:
+            verrors.add(
+                f'{schema}.command',
+                'Please specify a command for cronjob task.'
+            )
+        else:
+            crontab_cmd = (await self.construct_cron_command(
+                data['schedule'], user, command, data['stdout'], data['stderr']
+            ))[6:]
+            command = crontab_cmd.pop(1)
+
+            # When cron(8) reads an entry from a crontab, it keeps a buffer of 1000 characters and anything more then
+            # that is truncated. We validate that the user supplied command is not more than 1000 characters.
+            allowed_length = 1000 - len(' '.join(crontab_cmd))
+
+            if len(command) > allowed_length:
+                verrors.add(
+                    f'{schema}.command',
+                    f'Command must be less than or equal to {allowed_length} characters. '
+                    'Newline characters are automatically removed and "%" characters escaped with a backslash.'
+                )
+
         return verrors, data
 
     @accepts(
         Dict(
             'cron_job_create',
             Bool('enabled'),
-            Bool('stderr'),
-            Bool('stdout'),
+            Bool('stderr', default=False),
+            Bool('stdout', default=True),
             Cron('schedule'),
             Str('command', required=True),
             Str('description'),
@@ -104,7 +141,7 @@ class CronJobService(CRUDService):
 
         await self.middleware.call('service.restart', 'cron')
 
-        return data
+        return await self._get_instance(data['id'])
 
     @accepts(
         Int('id', validators=[Range(min=1)]),
@@ -137,7 +174,7 @@ class CronJobService(CRUDService):
 
             await self.middleware.call('service.restart', 'cron')
 
-        return await self.query(filters=[('id', '=', id)], options={'get': True})
+        return await self._get_instance(id)
 
     @accepts(
         Int('id')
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/cron.py
```diff
@@ -41,14 +48,38 @@ def cron_extend(self, data):
...
+        else:
+            allowed_length = 1000 - len(' '.join(
+                filter(
+                    bool, (
+                        f'PATH="{await self.cronjob_path()}"',
+                        '> /dev/null' if data['stdout'] else '',
+                        '2> /dev/null' if data['stderr'] else ''
+                    )
+                )
+            ))
```
william-gr: Seems we are repeating ourselves here, arent we?
Why not make a private method to get the full command, including PATH and use it in cron generation?

Then `cronjob_path` can also go away as i dont see that being useful for public API.

src/middlewared/middlewared/plugins/cron.py
```diff
@@ -41,14 +48,38 @@ def cron_extend(self, data):
...
                 )
 
+        command = data.get('command')
+        if not command:
+            verrors.add(
+                f'{schema}.command',
+                'Please specify a command for cronjob task.'
+            )
+        else:
+            allowed_length = 1000 - len(' '.join(
```
william-gr: Ah, also a comment about why the magical 1000, please. The next developer working here will appreciate ;)

src/middlewared/middlewared/plugins/cron.py
```diff
@@ -16,6 +16,20 @@ def cron_extend(self, data):
         Cron.convert_db_format_to_schedule(data)
         return data
 
+    @private
+    async def construct_cron_command(self, schedule, user, command, stdout=True, stderr=True):
+        return list(
+            filter(
```
william-gr: What is this filter for?

sonicaj: So when we join this list on space, if `stdout` is `''`, it adds an extra space in the final command. This filter flushes `stdout/stderr` if they are emtpy

src/middlewared/middlewared/plugins/cron.py
```diff
@@ -41,14 +55,38 @@ def cron_extend(self, data):
...
+                'Please specify a command for cronjob task.'
+            )
+        else:
+            crontab_cmd = (await self.construct_cron_command(
+                data['schedule'], user, command, data['stdout'], data['stderr']
+            ))[6:]
+            command = crontab_cmd.pop(1)
+
+            # cron(8) when reads an entry from a crontab, it keeps a buffer of 1000 characters and anything more then
+            # that is truncated. So we validate that the user supplied command is not of more then 1000 characters.
```
Mrt134:
```suggestion
            # that is truncated. We validate that the user supplied command is not more than 1000 characters.
```

src/middlewared/middlewared/plugins/cron.py
```diff
@@ -41,14 +55,38 @@ def cron_extend(self, data):
...
+                f'{schema}.command',
+                'Please specify a command for cronjob task.'
+            )
+        else:
+            crontab_cmd = (await self.construct_cron_command(
+                data['schedule'], user, command, data['stdout'], data['stderr']
+            ))[6:]
+            command = crontab_cmd.pop(1)
+
+            # cron(8) when reads an entry from a crontab, it keeps a buffer of 1000 characters and anything more then
```
Mrt134:
```suggestion
            # When cron(8) reads an entry from a crontab, it keeps a buffer of 1000 characters and anything more then
```

src/middlewared/middlewared/plugins/cron.py
```diff
@@ -41,14 +55,38 @@ def cron_extend(self, data):
...
+            # that is truncated. So we validate that the user supplied command is not of more then 1000 characters.
+            allowed_length = 1000 - len(' '.join(crontab_cmd))
+
+            if len(command) > allowed_length:
+                verrors.add(
+                    f'{schema}.command',
+                    f'Command must be less then or equal to {allowed_length} characters. '
+                    'It should be noted that in the command provided, all newline characters are removed and '
```
Mrt134:
```suggestion
                    'Any newline or "%" characters in the command are automatically removed or escaped with '
```

src/middlewared/middlewared/plugins/cron.py
```diff
@@ -41,14 +55,38 @@ def cron_extend(self, data):
...
+                    '"%" characters are escaped with backslash.'
```
Mrt134:
```suggestion
                    'backslash.'
```

src/middlewared/middlewared/plugins/cron.py
```diff
@@ -41,14 +55,38 @@ def cron_extend(self, data):
...
+            command = crontab_cmd.pop(1)
+
+            # cron(8) when reads an entry from a crontab, it keeps a buffer of 1000 characters and anything more then
+            # that is truncated. So we validate that the user supplied command is not of more then 1000 characters.
+            allowed_length = 1000 - len(' '.join(crontab_cmd))
+
+            if len(command) > allowed_length:
+                verrors.add(
+                    f'{schema}.command',
+                    f'Command must be less then or equal to {allowed_length} characters. '
```
Mrt134:
```suggestion
                    f'Command must be less than or equal to {allowed_length} characters. '
```

src/middlewared/middlewared/plugins/cron.py
sonicaj: @Mrt134 newline characters are not escaped with a backslash ;) How about we make that explicit here ? Or here implies that both cases can happen ( removed/escaped ) which is of course not the case ?

sonicaj: @Mrt134 ping

sonicaj: @freenas/docs-team ping

Mrt134: There's a couple different ways. It's currently suggested as an understood implication that first item correlates to first item, but you can add `, respectively` to the end of the sentence to clarify or rewrite to break up the paired pairs:
"Any newline or "%" characters in the command are automatically removed or escaped with backslash, respectively."
or
"Newline characters are automatically removed and "%" characters escaped with a backslash."

sonicaj: Perfect, thank you :) I'll have this covered.

Mrt134: No problem! Sorry it took a bit to get back to you.

PR #2850, opened 2019-04-07
tkt-101233: Correct brand/logo for new boot loader in FN 12

sonicaj:
We have switched boot loaders from fourth to lua in FN 12. This PR updates FN 12 to show FN brand/logos for installer and FN iso.
Ticket: #101233


sonicaj:
We should be able to manage this via ports. Thank you for pointing it out :) I'll have this accommodated.

PR #2856, opened 2019-04-08
Improve UPS Validation

sonicaj:
This PR improves UPS validation making sure certain fields are always required by UPS Service else it will fail to start.
Ticket: NAS-101320

### DIFF
--- a/src/middlewared/middlewared/plugins/ups.py
+++ b/src/middlewared/middlewared/plugins/ups.py
@@ -106,19 +106,21 @@ class UPSService(SystemServiceService):
                 )
 
         mode = data.get('mode')
-        if mode:
-            if mode == 'MASTER':
-                if not data.get('port'):
-                    verrors.add(
-                        f'{schema}.port',
-                        'This field is required'
-                    )
-            else:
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/ups.py
```diff
@@ -98,6 +98,15 @@ def driver_choices(self):
                     'Use alphanumeric characters, ".", "-" and "_"'
                 )
 
+        for field in filter(
+            lambda f: not data[f],
+            ['monpwd', 'monuser', 'identifier', 'shutdowncmd']
+        ):
+            verrors.add(
```
themylogin: Why didn't you just mark the field as required?

sonicaj: @themylogin because this is a config service and we have `update=True` here. Also required doesn't check for empty string iirc

themylogin: `empty=False` does

sonicaj: This simplifies a lot of things, will update the PR. Thank you

PR #2859, opened 2019-04-08
fix(jail): Make start, stop, restart and exec jobs

skarekrow:
Update tests to reflect this

NAS-101362


skarekrow:
@erinix GUI will need to adapt to this

PR #2861, opened 2019-04-09
Defaults for Cron Schema

sonicaj:
This PR adds defaults for services using Cron Schema to ensure that we specify the defaults in the service's Cron Schema.
Ticket: NAS-101376

### DIFF
--- a/src/middlewared/middlewared/plugins/cloud_sync.py
+++ b/src/middlewared/middlewared/plugins/cloud_sync.py
@@ -638,7 +638,11 @@ class CloudSyncService(CRUDService):
         Bool("filename_encryption", default=False),
         Str("encryption_password", default=""),
         Str("encryption_salt", default=""),
-        Cron("schedule", required=True),
+        Cron(
+            "schedule",
+            defaults={"minute": "00"},
+            required=True
+        ),
         Bool("follow_symlinks", default=False),
         Int("transfers", null=True, default=None, validators=[Range(min=1)]),
         List("bwlimit", default=[], items=[Dict("cloud_sync_bwlimit",
--- a/src/middlewared/middlewared/plugins/replication.py
+++ b/src/middlewared/middlewared/plugins/replication.py
@@ -125,8 +125,20 @@ class ReplicationService(CRUDService):
             List("also_include_naming_schema", items=[
                 Str("naming_schema", validators=[ReplicationSnapshotNamingSchema()])], default=[]),
             Bool("auto", required=True),
-            Cron("schedule", begin_end=True, null=True, default=None),
-            Cron("restrict_schedule", begin_end=True, null=True, default=None),
+            Cron(
+                "schedule",
+                defaults={"minute": "00"},
+                begin_end=True,
+                null=True,
+                default=None
+            ),
+            Cron(
+                "restrict_schedule",
+                defaults={"minute": "00"},
+                begin_end=True,
+                null=True,
+                default=None
+            ),
             Bool("only_matching_schedule", default=False),
             Bool("allow_from_scratch", default=False),
             Bool("hold_pending_snapshots", default=False),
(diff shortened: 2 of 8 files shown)


src/middlewared/middlewared/plugins/replication.py
```diff
@@ -125,8 +125,24 @@ def _legacy_extend_context(self):
             List("also_include_naming_schema", items=[
                 Str("naming_schema", validators=[ReplicationSnapshotNamingSchema()])], default=[]),
             Bool("auto", required=True),
-            Cron("schedule", begin_end=True, null=True, default=None),
-            Cron("restrict_schedule", begin_end=True, null=True, default=None),
+            Cron(
+                "schedule",
+                Str('minute', default='00'),
+                Time('begin', default='00:00'),
+                Time('end', default='23:45'),
```
william-gr: I think this could be 23:59, it was 23:45 because how dojo worked.

william-gr:
Is that because we accept them in the schema anyway?

src/middlewared/middlewared/plugins/snapshot.py
```diff
@@ -50,7 +50,14 @@ class Config:
             Int('lifetime_value', required=True),
             Str('lifetime_unit', enum=['HOUR', 'DAY', 'WEEK', 'MONTH', 'YEAR'], required=True),
             Str('naming_schema', required=True, validators=[ReplicationSnapshotNamingSchema()]),
-            Cron('schedule', required=True, begin_end=True),
+            Cron(
+                'schedule',
+                Str('minute', default='00'),
+                Time('begin', default='09:00'),
+                Time('begin', default='18:00'),
```
william-gr: Should be `end`

src/middlewared/middlewared/plugins/cloud_sync.py
```diff
@@ -638,7 +638,11 @@ class Config:
         Bool("filename_encryption", default=False),
         Str("encryption_password", default=""),
         Str("encryption_salt", default=""),
-        Cron("schedule", required=True),
+        Cron(
+            "schedule",
+            Str('minute', default='00'),
```
william-gr: Everything here uses double quote, i will nitpick and ask for double quotes too :P

src/middlewared/middlewared/plugins/replication.py
```diff
@@ -125,8 +125,20 @@ def _legacy_extend_context(self):
             List("also_include_naming_schema", items=[
                 Str("naming_schema", validators=[ReplicationSnapshotNamingSchema()])], default=[]),
             Bool("auto", required=True),
-            Cron("schedule", begin_end=True, null=True, default=None),
-            Cron("restrict_schedule", begin_end=True, null=True, default=None),
+            Cron(
+                "schedule",
+                Str('minute', default='00'),
```
william-gr: Same here about quotes

src/middlewared/middlewared/plugins/cloud_sync.py
```diff
@@ -638,7 +638,11 @@ class Config:
         Bool("filename_encryption", default=False),
         Str("encryption_password", default=""),
         Str("encryption_salt", default=""),
-        Cron("schedule", required=True),
+        Cron(
+            "schedule",
+            Str("minute", default="00"),
```
themylogin: Wouldn'it it be better to pass defaults as dictionary instead of this. What will happen if I say `Int("minute", default=0),`? Or `Str("minuete", default="00"),`?

sonicaj: @themylogin This will certainly cover mistakes made on our part, but it would differ from the schema standard we have laid out. Do you think i should proceed with this ? If yes, I'll move ahead and take care of this :)
If not the former case can be taken care of quite easily by only taking defaults into consideration but then again it depends on how we choose to implement it. Please let me know what you think. Thank you

themylogin: I don't think we have any schema standard we have laid out related to this. Your proposed `Cron` definition looks like `Dict` definition, that's fancy, but IMO strange. @william-gr what do you think?

william-gr: I agree with @themylogin here, redefining whole Schema type within Cron seems strange, simply overriding defaults seems saner.

sonicaj: Perfect. Thank you @themylogin @william-gr. I'll have this covered asap

PR #2865, opened 2019-04-11
Two-Factor Authentication

sonicaj:
This PR introduces Two Factor Authentication with following changes:
1) Introduces a service to manage 2FA.
2) Supports TOTP ( time based OTP ) only for right now as it is generally considered more secure then standard HOTP ( counter based ) which can be valid for unknown period of time.
3) Ability to use google authenticator or other 2FA apps.

Ticket: NAS-100123

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/ssh/sshd_config
+++ b/src/middlewared/middlewared/etc_files/local/ssh/sshd_config
@@ -20,6 +20,10 @@
 
 	if bind_ifaces:
 		bind_ifaces.insert(0, '127.0.0.1')
+
+	twofactor_auth = middleware.call_sync('auth.twofactor.config')
+	twofactor_enabled = twofactor_auth['enabled'] and twofactor_auth['services']['ssh']
+
 %>\
 Subsystem	sftp	/usr/libexec/sftp-server -l ${ssh_config['sftp_log_level']} -f ${ssh_config['sftp_log_facility']}
 % if 'Protocol' not in ssh_config['options']:
@@ -28,7 +32,7 @@ Protocol 2
 % if 'UseDNS' not in ssh_config['options']:
 UseDNS no
 % endif
-% if 'ChallengeResponseAuthentication' not in ssh_config['options']:
+% if 'ChallengeResponseAuthentication' not in ssh_config['options'] and not twofactor_enabled:
 ChallengeResponseAuthentication no
 % endif
 % if 'ClientAliveCountMax' not in ssh_config['options']:
@@ -76,3 +80,8 @@ GSSAPIAuthentication yes
 % endif
 PubkeyAuthentication yes
 ${ssh_config['options']}
+% if twofactor_enabled:
+# These are forced to be enabled with 2FA
+UsePAM yes
+ChallengeResponseAuthentication yes
+% endif
--- a/src/middlewared/middlewared/etc_files/local/users.oath
+++ b/src/middlewared/middlewared/etc_files/local/users.oath
@@ -1,0 +1,10 @@
+<%
+	import base64
+
+	twofactor_auth = middleware.call_sync('auth.twofactor.config')
+	if twofactor_auth['enabled'] and twofactor_auth['services']['ssh']:
+		twofactor_auth['secret_hex'] = base64.b16encode(base64.b32decode(twofactor_auth['secret'])).decode()
+	else:
+		raise FileShouldNotExist()
+%>\
+${f'HOTP/T{twofactor_auth["interval"]}/{twofactor_auth["otp_digits"]}'}	root	-	${twofactor_auth['secret_hex']}
--- a/src/middlewared/middlewared/etc_files/pam.d/sshd
+++ b/src/middlewared/middlewared/etc_files/pam.d/sshd
@@ -3,10 +3,12 @@
 #
 # PAM configuration for the "sshd" service
 #
-<%namespace name="pam" file="pam.inc" />
+<%namespace name="pam" file="pam.inc" />\
 <%
-    dsp = pam.getDirectoryServicePam(middleware=middleware, file='sshd')
-%>
+	dsp = pam.getDirectoryServicePam(middleware=middleware, file='sshd')
+	twofactor_auth = middleware.call_sync('auth.twofactor.config')
+	twofactor_enabled = twofactor_auth['enabled'] and twofactor_auth['services']['ssh']
+%>\
 # auth
 auth		sufficient	pam_opie.so		no_warn no_fake_prompts
 auth		requisite	pam_opieaccess.so	no_warn allow_local
@@ -16,6 +18,9 @@ ${dsp.pam_auth()}
 #auth		sufficient	pam_krb5.so		no_warn try_first_pass
 #auth		sufficient	pam_ssh.so		no_warn try_first_pass
 auth		required	pam_unix.so		no_warn try_first_pass
+% if twofactor_enabled:
+auth		required	/usr/local/lib/security/pam_oath.so	usersfile=/usr/local/etc/users.oath	window=${twofactor_auth['window']}
+% endif
 
 # account
 account		required	pam_nologin.so
--- a/src/middlewared/middlewared/plugins/auth.py
+++ b/src/middlewared/middlewared/plugins/auth.py
@@ -295,6 +295,11 @@ class AuthService(Service):
             return None
 
     @no_auth_required
+    @accepts()
+    async def two_factor_auth(self):
+        return (await self.middleware.call('auth.twofactor.config'))['enabled']
+
+    @no_auth_required
     @accepts(Str('username'), Str('password'), Str('otp_token', null=True, default=None))
     @pass_app
     async def login(self, app, username, password, otp_token=None):
--- a/src/middlewared/middlewared/plugins/auth.py
+++ b/src/middlewared/middlewared/plugins/auth.py
@@ -427,6 +427,17 @@ class TwoFactorAuthService(ConfigService):
 
         return True
 
+    @accepts()
+    async def provisioning_uri(self):
+        config = await self.middleware.call(f'{self._config.namespace}.config')
+        return pyotp.totp.TOTP(
+            config['secret'], interval=config['interval'], digits=config['otp_digits']
+        ).provisioning_uri(
+            f'{(await self.middleware.call("system.info"))["hostname"]}@'
+            f'{await self.middleware.call("system.product_name")}',
+            'iXsystems'
+        )
+
     @private
     def generate_base32_secret(self):
         return pyotp.random_base32()
--- a/src/middlewared/middlewared/plugins/auth.py
+++ b/src/middlewared/middlewared/plugins/auth.py
@@ -295,13 +295,23 @@ class AuthService(Service):
             return None
 
     @no_auth_required
-    @accepts(Str('username'), Str('password'))
+    @accepts(Str('username'), Str('password'), Str('otp_token', null=True, default=None))
     @pass_app
-    async def login(self, app, username, password):
+    async def login(self, app, username, password, otp_token=None):
         """Authenticate session using username and password.
         Currently only root user is allowed.
         """
         valid = await self.check_user(username, password)
+        twofactor_auth = await self.middleware.call('auth.twofactor.config')
+
+        if twofactor_auth['enabled']:
+            # We should run auth.twofactor.verify nevertheless of check_user result to prevent guessing
+            # passwords with a timing attack
+            valid &= await self.middleware.call(
+                'auth.twofactor.verify',
+                otp_token
+            )
+
         if valid:
             self.session_manager.login(app, LoginPasswordSessionManagerCredentials())
         return valid
@@ -375,6 +385,19 @@ class TwoFactorAuthService(ConfigService):
         )
 
         return await self.config()
+
+    @accepts(
+        Str('token', null=True)
+    )
+    def verify(self, token):
+        config = self.middleware.call_sync(f'{self._config.namespace}.config')
+        if not config['enabled']:
+            raise CallError('Please enable Two Factor Authentication first.')
+
+        totp = pyotp.totp.TOTP(
+            config['secret'], interval=config['interval'], digits=config['otp_digits']
+        )
+        return totp.verify(token, valid_window=config['window'])
 
     @accepts()
     def renew_secret(self):
--- a/src/middlewared/middlewared/plugins/auth.py
+++ b/src/middlewared/middlewared/plugins/auth.py
@@ -422,6 +422,9 @@ class TwoFactorAuthService(ConfigService):
             }
         )
 
+        if config['services']['ssh']:
+            self.middleware.call_sync('service.reload', 'ssh')
+
         return True
 
     @private
--- a/src/middlewared/middlewared/plugins/auth.py
+++ b/src/middlewared/middlewared/plugins/auth.py
@@ -7,9 +7,10 @@ import string
 import subprocess
 import time
 
-from middlewared.schema import Dict, Int, Str, accepts
-from middlewared.service import Service, filterable, filter_list, no_auth_required, pass_app, private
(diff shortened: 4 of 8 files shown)


src/middlewared/middlewared/etc_files/local/ssh/config.py
```diff
@@ -24,6 +24,10 @@ def generate_ssh_config(middleware):
         if os.path.exists(os.path.join(SSH_CONFIG_PATH, f)):
             os.chmod(os.path.join(SSH_CONFIG_PATH, f), 0o600)
 
+    twofactor_auth = middleware.call_sync('auth.twofactor.config')
+    if any(not v for v in (twofactor_auth['enabled'], twofactor_auth['services']['ssh'])):
```
william-gr: Thats a over-engineered construct :P

`if not twofactor_auth['enabled'] or not twofactor_auth['services'].get('ssh'):`

looks much more direct, no? :P

src/middlewared/middlewared/etc_files/local/ssh/sshd_config
```diff
@@ -20,6 +20,10 @@
 
 	if bind_ifaces:
 		bind_ifaces.insert(0, '127.0.0.1')
+
+	twofactor_auth = middleware.call_sync('auth.twofactor.config')
+	twofactor_enabled = all(v for v in (twofactor_auth['enabled'], twofactor_auth['services']['ssh']))
```
william-gr: `(twofactor_auth['enabled'] and twofactor_auth['services'].get('ssh'))` ? :)

src/middlewared/middlewared/etc_files/pam.d/sshd
```diff
@@ -3,10 +3,12 @@
...
 # PAM configuration for the "sshd" service
 #
-<%namespace name="pam" file="pam.inc" />
+<%namespace name="pam" file="pam.inc" />\
 <%
-    dsp = pam.getDirectoryServicePam(middleware=middleware, file='sshd')
-%>
+	dsp = pam.getDirectoryServicePam(middleware=middleware, file='sshd')
+	twofactor_auth = middleware.call_sync('auth.twofactor.config')
+	twofactor_enabled = all(v for v in (twofactor_auth['enabled'], twofactor_auth['services']['ssh']))
```
william-gr: Again :P

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -1,15 +1,19 @@
+import base64
 import crypt
 from datetime import datetime, timedelta
+import pyotp
```
william-gr: Gonna need a PR adding it as a dependency in ports

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -1,15 +1,19 @@
...
 import socket
 import string
 import subprocess
 import time
 
-from middlewared.schema import Dict, Int, Str, accepts
-from middlewared.service import Service, filterable, filter_list, no_auth_required, pass_app, private
+from middlewared.schema import Dict, Int, Str, accepts, Bool
+from middlewared.service import ConfigService, Service, filterable, filter_list, \
+    no_auth_required, pass_app, private, CallError
```
william-gr: Lets use parenthesis to break lines as we do in the others plugins, please

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -325,9 +339,130 @@ def token(self, app, token):
...
+    @accepts
+    def provisioning_uri(self):
+        """
+        Returns the provisioning URI for the OTP. This can then be encoded in a QR Code and used to
+        provision an OTP app like Google Authenticator.
+        """
+        config = self.middleware.call_sync(f'{self._config.namespace}.config')
+        return pyotp.totp.TOTP(
+            config['secret'], interval=config['interval'], digits=config['otp_digits']
+        ).provisioning_uri('root@freenas', 'iXsystems')
```
william-gr: How does this work and why are we hardcoding these values?

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -325,9 +339,130 @@ def token(self, app, token):
...
+
+    @private
+    def two_factor_extend(self, data):
+        data['secret'] = self.middleware.call_sync('pwenc.decrypt', data['secret'])
+
+        for srv in ['ssh']:
+            if not data['services'].get(srv):
+                data['services'][srv] = False
+
+        data['secret_hex'] = base64.b16encode(base64.b32decode(data['secret'])).decode()
```
william-gr: We already have secret, why do we need secret_hex in the service? If its only used in the config file, perhaps it can be encoded there?

src/middlewared/middlewared/plugins/etc.py
```diff
@@ -195,6 +195,8 @@ class EtcService(Service):
         ],
         'ssh': [
             {'type': 'mako', 'path': 'local/ssh/sshd_config'},
+            {'type': 'mako', 'path': 'local/users.oath'},
```
william-gr: I think you dont want users.oath file to have group/others permissions?

src/middlewared/middlewared/etc_files/local/users.oath
```diff
@@ -0,0 +1,6 @@
+<%
+	twofactor_auth = middleware.call_sync('auth.twofactor.config')
+%>\
+% for user in middleware.call_sync('user.query'):
+${f'HOTP/T{twofactor_auth["interval"]}/{twofactor_auth["otp_digits"]}'}	${user['username']}	-	${twofactor_auth['secret_hex']}
```
william-gr: So all users will have the same two factor? I wonder if we shouldnt restrict this only for uid 0

src/middlewared/middlewared/plugins/etc.py
sonicaj: By default it has 0600 perms and those are exactly what we require but i should make it explicit here.

william-gr: It should not be 0600 by default, there is likely a bug somewhere changing umask

src/middlewared/middlewared/plugins/auth.py
sonicaj: So first param is `name` which is of the user account which is being secured, second is the name of the provider/issuer name which will show up in the authenticator app as the organization which i defaulted to `iXsystems`. These values don't have any role to play in the authentication itself but are used by the 2FA apps to help users know which code is being used for which service etc.
We can perhaps allow this to be configurable - the account part ?
https://github.com/google/google-authenticator/wiki/Key-Uri-Format

william-gr: I think we should at least use the hostname of freenas to make it more identifiable then

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -325,9 +339,128 @@ def token(self, app, token):
...
+    @accepts
+    async def provisioning_uri(self):
+        """
+        Returns the provisioning URI for the OTP. This can then be encoded in a QR Code and used to
+        provision an OTP app like Google Authenticator.
+        """
+        config = self.middleware.call_sync(f'{self._config.namespace}.config')
+        return pyotp.totp.TOTP(
+            config['secret'], interval=config['interval'], digits=config['otp_digits']
+        ).provisioning_uri((await self.middleware.call('system.info'))['hostname'], 'iXsystems')
```
william-gr: It will show just the hostname in the app?

Giving it a second thought, perhaps $hostname@$product_name?
What is iXsytems, hoe does that show?

william-gr:
@themylogin care to review this too? maybe i am missing something more fundamental

src/middlewared/middlewared/etc_files/local/ssh/config.py
```diff
@@ -24,6 +24,10 @@ def generate_ssh_config(middleware):
         if os.path.exists(os.path.join(SSH_CONFIG_PATH, f)):
             os.chmod(os.path.join(SSH_CONFIG_PATH, f), 0o600)
 
+    twofactor_auth = middleware.call_sync('auth.twofactor.config')
+    if not twofactor_auth['enabled'] or not twofactor_auth['services']['ssh']:
+        os.remove('/usr/local/etc/users.oath')
```
themylogin: Can't we invent some mechanism that will prevent this file from being created instead?

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -291,13 +295,23 @@ def get_token(self, token_id):
...
+        """
+        Authenticate session using username and password.
         Currently only root user is allowed.
+        `otp_token` must be specified if two factor authentication is enabled.
         """
         valid = await self.check_user(username, password)
+        twofactor_auth = await self.middleware.call('auth.twofactor.config')
+
+        if twofactor_auth['enabled']:
+            valid &= await self.middleware.call(
```
themylogin: Please leave a comment here about why we are doing this even if `check_user` returns false so it won't get "optimized" by someone one day.

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -325,9 +339,128 @@ def token(self, app, token):
...
+        datastore = 'system.twofactorauthentication'
+        datastore_extend = 'auth.twofactor.two_factor_extend'
+        namespace = 'auth.twofactor'
+
+    @private
+    async def two_factor_extend(self, data):
+        data['secret'] = await self.middleware.call('pwenc.decrypt', data['secret'])
+
+        for srv in ['ssh']:
+            if not data['services'].get(srv):
```
themylogin: Lifehack: `data['services'].setdefault(srv, False)` (not asking you to change it this time, just for the future)

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -325,9 +339,128 @@ def token(self, app, token):
...
+
+        if config['enabled'] and not config['secret']:
+            # Only generate a new secret on `enabled` when `secret` is not already set.
+            # This will aid users not setting secret up again on their mobiles.
+            config['secret'] = await self.middleware.run_in_thread(
+                self.generate_base32_secret
+            )
+
+        config['secret'] = await self.middleware.call('pwenc.encrypt', config['secret'])
+        config.pop('secret_hex', None)
```
themylogin: What's this?

src/middlewared/middlewared/plugins/auth.py
sonicaj: Added ;)

src/middlewared/middlewared/etc_files/local/ssh/config.py
```diff
@@ -24,6 +25,16 @@ def generate_ssh_config(middleware):
         if os.path.exists(os.path.join(SSH_CONFIG_PATH, f)):
             os.chmod(os.path.join(SSH_CONFIG_PATH, f), 0o600)
 
+    twofactor_auth = middleware.call_sync('auth.twofactor.config')
+    if twofactor_auth['enabled'] and twofactor_auth['services']['ssh']:
+        with open(USERS_OATH_FILE, 'w') as f:
```
themylogin: You've forgot to delete this file in `else` branch

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -291,13 +295,25 @@ def get_token(self, token_id):
...
+        """
+        Authenticate session using username and password.
         Currently only root user is allowed.
+        `otp_token` must be specified if two factor authentication is enabled.
         """
         valid = await self.check_user(username, password)
+        twofactor_auth = await self.middleware.call('auth.twofactor.config')
+
+        if twofactor_auth['enabled']:
+            # We should validate that BOTH `check_user` AND `auth.twofactor.verify` are true and only then
```
themylogin: That's not the real reason, isn't it? The real reason is to prevent guessing passwords with timing attack.

src/middlewared/middlewared/etc_files/local/ssh/config.py
william-gr: So, there is a small security breach here, you create and write the file, only then you chmod it.
That leaves an opportunity for an attacker to read the file before you chmod it.

We should do `with open(os.open(USERS_OATH_FILE, os.O_CREAT | os.O_WRONLY, 0o600), 'w') as f:`

src/middlewared/middlewared/etc_files/local/ssh/config.py
```diff
@@ -24,6 +25,16 @@ def generate_ssh_config(middleware):
...
+            f.write(
+                f'HOTP/T{twofactor_auth["interval"]}/{twofactor_auth["otp_digits"]}\troot\t-\t'
+                f'{base64.b16encode(base64.b32decode(twofactor_auth["secret"])).decode()}\n'
+            )
+
+        os.chmod(USERS_OATH_FILE, 0o600)
+
```
william-gr: What if it gets disabled? I think we should remove the file, so we are back at square one :P

src/middlewared/middlewared/etc_files/local/ssh/config.py
themylogin: I was thinking that we can throw something from `mako` template that will make mako stop generating the file and delete it if it already exists.

william-gr: Good idea

william-gr: raise DeprecatedFile ? : P

src/middlewared/middlewared/plugins/auth.py
sonicaj: Yes indeed. That's what the two factor auth enabled check represents ;) I'll update the comment :)

src/middlewared/middlewared/utils/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class DeprecatedFile(Exception):
```
themylogin: Bad name, because in our case it's not deprecated, it just should not exist. Maybe `ShouldNotExist` or something like that?

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -325,9 +350,130 @@ def token(self, app, token):
...
+        """
+        Returns the provisioning URI for the OTP. This can then be encoded in a QR Code and used to
+        provision an OTP app like Google Authenticator.
+        """
+        config = self.middleware.call_sync(f'{self._config.namespace}.config')
+        return pyotp.totp.TOTP(
+            config['secret'], interval=config['interval'], digits=config['otp_digits']
+        ).provisioning_uri(
+            f'{(await self.middleware.call("system.info"))["hostname"]}@'
+            f'{"freenas" if await self.middleware.call("system.is_freenas") else "truenas"}',
```
themylogin: `system.product_name`

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -291,13 +295,34 @@ def get_token(self, token_id):
...
+        """
+        Authenticate session using username and password.
         Currently only root user is allowed.
+        `otp_token` must be specified if two factor authentication is enabled.
         """
         valid = await self.check_user(username, password)
+        twofactor_auth = await self.middleware.call('auth.twofactor.config')
+
+        if twofactor_auth['enabled']:
+            # We should validate that BOTH `check_user` AND `auth.twofactor.verify` are true and only then
```
themylogin: "We should run `auth.twofactor.verify` nevertheless of `check_user` result to prevent guessing passwords with a timing attack."

src/middlewared/middlewared/etc_files/local/users.oath
```diff
@@ -0,0 +1,11 @@
+<%
+	from middlewared.utils.exceptions import DeprecatedFile
```
themylogin: Can we add this exception to the context of all templates like we do with `middleware` object?

sonicaj:
@themylogin thank you for raising my level on these too ;)

PR #2868, opened 2019-04-11
NAS-101386 / 11.2 /  Do not run alerts if failover is not complete yet

themylogin:


src/middlewared/middlewared/plugins/alert.py
```diff
@@ -279,6 +275,17 @@ def restore(self, id):
...
+    async def __should_run_or_send_alerts(self):
+        if (
+            not await self.middleware.call('system.is_freenas') and
+            await self.middleware.call('notifier.failover_licensed') and
+            (
+                await self.middleware.call('notifier.failover_status') == 'BACKUP' or
+                await self.middleware.call('notifier.failover_in_progress')
+            )
+        ):
+            return False
```
william-gr: Shouldnt we return True on the else?

PR #2871, opened 2019-04-11
NAS-101074 / 11.2 /  Block VolumeStatus alert source while creating pool

themylogin:


gui/middleware/notifier.py
```diff
@@ -774,30 +774,35 @@ def stripe_first(a, b):
...
-        # We've our pool, lets retrieve the GUID
-        p1 = self._pipeopen("zpool get guid %s" % z_name)
-        if p1.wait() == 0:
-            line = p1.communicate()[0].split('\n')[1].strip()
-            volume.vol_guid = re.sub('\s+', ' ', line).split(' ')[2]
-            volume.save()
-        else:
-            log.warn("The guid of the pool %s could not be retrieved", z_name)
+        with client as c:
+            lock = c.call("alert.block_source", "VolumeStatus")
```
william-gr: I think the Volume object will be create long before this for 11.2, which is enough for the pool.query to return it as an "existing" pool. Perhaps we should do that on the places calling this method, before creating Volume entry?

src/middlewared/middlewared/plugins/alert.py
```diff
@@ -377,6 +389,21 @@ def restore(self, id):
         except UnavailableException:
             raise CallError("This alert checker is unavailable", CallError.EALERTCHECKERUNAVAILABLE)
 
+    @private
+    async def block_source(self, source_name):
+        if source_name not in ALERT_SOURCES:
+            raise CallError("Invalid alert source")
+
+        lock = str(uuid.uuid4())
+        self.blocked_sources[source_name].add(lock)
```
william-gr: A thought that just crossed my mind, what if for whatever reason the client simply dies before it can unlock?
Should we have some sort of timer to autolock after a huge amount of time has passed? e.g. 5 or 10 min?

themylogin: Pool creation take take half an hour on one of our test systems in the lab, that's why I've set default timeout to 3600.

PR #2878, opened 2019-04-12
NAS-101226: Clarify "reset interface" message

wblock:
https://jira.ixsystems.com/browse/NAS-101226

### DIFF
--- a/src/freenas/etc/netcli
+++ b/src/freenas/etc/netcli
@@ -250,9 +250,9 @@ def configure_interface():
             continue
 
     while True:
-        _input = input(_("Reset network configuration? (y/n) ")).lower()
+        _input = input(_("Restore this interface to saved configuration? (This causes a momentary disconnection of the network.) (y/n) ")).lower()
         if _input.startswith("y"):
-            print(_("Resetting interface configuration:"), end=' ')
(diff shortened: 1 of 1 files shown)


src/freenas/etc/netcli
```diff
@@ -250,9 +250,9 @@ def configure_interface():
             continue
 
     while True:
-        _input = input(_("Reset network configuration? (y/n) ")).lower()
+        _input = input(_("Restore this interface to saved configuration? (This causes a momentary disconnection of the network.) (y/n) ")).lower()
```
william-gr: What saved configuration are you referring to?

I think this is a step back since its actually resetting, all interface values are being blanked.

william-gr:
Also the referenced ticket NAS-100144 seems wrong

src/freenas/etc/netcli
wblock: This was from input from other people.  The problem is that "reset" is very ambiguous.
But I see what you mean, will update.

wblock:
@william-gr Can this be cherry-picked to master or does the network interface change mean it needs to be worded differently or not at all?

william-gr:
It can, netcli is still the same there.

PR #2903, opened 2019-04-17
NAS-101178 / 11.2 / Nas 101178

themylogin:


src/freenas/etc/ix.rc.d/ix-shutdown
```diff
@@ -36,6 +36,8 @@ do_shutdown()
 
 	execute_task "command"
 	execute_task "script"
+
+	/usr/local/bin/midclt call core.event_send system ADDED '{"id": "shutting-down"}' > /dev/null
```
william-gr: Could as well make it the first command to run, before the scripts

PR #2908, opened 2019-04-17
NAS-101466 / 11.3 / Failover Upgrade [FreeNAS side]

william-gr:


src/middlewared/middlewared/client/client.py
```diff
@@ -232,6 +232,39 @@ def __init__(self, method, params):
...
+class Job(object):
+
+    def __init__(self, client, job_id, callback=None):
+        self.client = client
+        self.job_id = job_id
+        # If a job event has been received already then we must set an Event
+        # to wait for this job to finish.
+        # Otherwise we create a new stub for the job with the Event for when
+        # the job event arrives to use existing event.
+        with client._jobs_lock:
```
themylogin: IMO as this code uses `Client` internals, it should have stayed in client.

PR #2909, opened 2019-04-17
NAS-100431 / 11.3 / Use alerts framework instead of periodic scripts

themylogin:


src/freenas/etc/periodic.conf
```diff
@@ -27,3 +27,9 @@ security_show_success="NO"
 weekly_locate_enable="NO"
 weekly_whatis_enable="NO"
 monthly_show_success="NO"
+daily_output="/dev/null"
+daily_status_security_output="/dev/null"
+weekly_output="/dev/null"
+weekly_status_security_output="/dev/null"
+monthly_output="/dev/null"
+monthly_status_security_output="/dev/null"
```
william-gr: That means periodic will still run?

Do we need that? Seems it would be more clear if we remove this failed and stop calling periodic in crontab?

src/freenas/etc/periodic/security/800.loginfail-freenas
```diff
@@ -83,7 +83,7 @@ def get_login_failures(now, messages):
     for message in messages:
         if message.strip():
             if message.startswith(yesterday):
-                if re.search(rb"\b(fail(ures?|ed)?|invalid|bad|illegal|auth.*error)\b", message):
+                if re.search(rb"\b(fail(ures?|ed)?|invalid|bad|illegal|auth.*error)\b", message, re.I):
```
william-gr: This was made an alert source, no? Why still have this around?

PR #2916, opened 2019-04-18
NAS-101478 / 11.3 / fix(vm/device_create): Referencing this schema without an id failed

skarekrow:
There is no way to supply this id to build the enum for vm.create, so if the id is not supplied, we simply do nothing.

NAS-101478

### DIFF
--- a/src/middlewared/middlewared/plugins/vm.py
+++ b/src/middlewared/middlewared/plugins/vm.py
@@ -3,6 +3,7 @@ from middlewared.async_validators import check_path_resides_within_volume
 from middlewared.schema import accepts, Error, Int, Str, Dict, List, Bool, Patch, Ref
 from middlewared.service import (
     item_method, pass_app, private, CRUDService, CallError, ValidationErrors,
+    ValidationError
 )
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1259,8 +1260,14 @@ def nic_attach_choices(self):
         """
         Create a new device for the VM of id `vm`.
         """
+        if not data['vm']:
+            raise ValidationError('id', '"id" is required')
```
william-gr: `vmdevice_create.vm`

"vm" is required

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1259,8 +1260,14 @@ def nic_attach_choices(self):
         """
         Create a new device for the VM of id `vm`.
         """
+        if not data.get('vm'):
+            raise ValidationError('vm', '"vm" is required')
```
william-gr: `vmdevice_create.vm`

PR #2917, opened 2019-04-18
NAS-101482 / 11.3 / fix(vm): Correctly overcommit memory

skarekrow:
NAS-101482


src/middlewared/middlewared/plugins/vm.py
```diff
@@ -752,7 +752,7 @@ def get_available_memory(self, overcommit):
         memory_available = await self.middleware.call('vm.get_available_memory', overcommit)
         memory_bytes = memory * 1024 * 1024
         if memory_bytes > memory_available:
-            return False
+            return False if not overcommit else True
```
william-gr: I dont think thats quite right, you're just telling it succeeded without possibly changing ARC.

Please overcommit means using more memory than availalbe to support all guests, not allowing it to launch 10 VMs of 4GB each on a 6gb system.

PR #2923, opened 2019-04-19
NAS-100740 / 11.3 / Cloud Sync: SFTP: Use private key from keychain instead of requiring …

themylogin:
…to specify FS path


src/middlewared/middlewared/plugins/cloud_sync.py
```diff
@@ -45,27 +45,26 @@ def __init__(self, cloud_sync):
...
         self.tmp_file = tempfile.NamedTemporaryFile(mode="w+")
 
         # Make sure only root can read it as there is sensitive data
         os.chmod(self.tmp_file.name, 0o600)
 
         config = dict(self.cloud_sync["credentials"]["attributes"], type=self.provider.rclone_type)
-        config = dict(config, **self.provider.get_credentials_extra(self.cloud_sync["credentials"]))
+        config = dict(config, **await self.provider.get_credentials_extra(self.cloud_sync["credentials"]))
         if "pass" in config:
             config["pass"] = rclone_encrypt_password(config["pass"])
```
william-gr: Are other methods in here safe to run in event loop? e.g. rclone_encrypt_password

src/middlewared/middlewared/plugins/cloud_sync.py
```diff
@@ -465,6 +468,38 @@ class Config:
         if verrors:
             raise verrors
 
+    @private
+    def migrate_sftp_key_file(self):
```
william-gr: I am not sure I like the idea of running that every time system is ready or pool is imported.

It sounds like we need a one-time scripts run for upgrades?

src/middlewared/middlewared/plugins/cloud_sync.py
themylogin: `rclone_encrypt_password` yes, writing temporary files in ramfs - subject to discuss

src/middlewared/middlewared/plugins/cloud_sync.py
```diff
@@ -113,14 +116,14 @@ def get_remote_path(provider, attributes):
 
 
 async def rclone(middleware, job, cloud_sync):
-    if not os.path.exists(cloud_sync["path"]):
+    if not middleware.run_in_thread(os.path.exists, cloud_sync["path"]):
```
william-gr: You have to await in these run_in_thread, no?

src/middlewared/middlewared/plugins/cloud_sync.py
```diff
@@ -925,9 +960,25 @@ def common_task_schema(self, provider):
...
+
+
 async def setup(middleware):
     for module in load_modules(os.path.join(os.path.dirname(os.path.realpath(__file__)), os.path.pardir,
                                             "rclone", "remote")):
         for cls in load_classes(module, BaseRcloneRemote, []):
             remote = cls(middleware)
             REMOTES[remote.name] = remote
+
+    for credential in await middleware.call("cloudsync.credentials.query", [["provider", "=", "SFTP"]]):
```
william-gr: I am still puzzled why we cant have a one time migration instead of having to support this code forever in middlewared code?

themylogin: Because these keys are stored on datasets and datasets are not mounted yet on migrations stage.

william-gr: But I didnt say database migration, but one time script migrations.

Its not the first time we have needed this, and I doubt its gonna be the last

themylogin: Is there any example of such script already present? Couldn't find any.

william-gr: No, but thats my point.

We have done many ad-hocs solutions like this for a long time, varying from ix rc.d scripts that would set sentinel files in /data to not run again and permanent code to handle migration.

Perhaps its time we come up with a better solution for such cases

src/middlewared/middlewared/plugins/migration.py
```diff
@@ -0,0 +1,52 @@
...
+
+    class Config:
+        private = True
+
+    async def run(self):
+        if await self.middleware.call("keyvalue.get", "fake_migration", False):
+            for module in load_migrations(self.middleware):
+                await self.middleware.call("datastore.insert", "system.migration", {"name": module.__name__})
+
+            await self.middleware.call("keyvalue.set", "fake_migration", False)
```
william-gr: I thought you said something yesterday about filling these during build time, did you change your mind?

src/middlewared/middlewared/migration/0001_cloudsync_sftp_private_key_file.py
```diff
@@ -0,0 +1,34 @@
...
+
+                del credential["attributes"]["key_file"]
+                credential["attributes"]["private_key"] = keypair["id"]
+
+                middleware.call_sync("datastore.update", "system.cloudcredentials", credential["id"], {
+                    "attributes": credential["attributes"],
+                })
+            except Exception as e:
+                middleware.logger.warning(f"Error migrating SFTP cloud credential %d to keychain: %r",
+                                          credential['id'], e)
```
william-gr: should we raise here to log the traceback?

src/middlewared/middlewared/plugins/migration.py
themylogin: After you said that we are going to do things differently (hopely the proper way) in 12.0, I've changed my mind.

src/middlewared/middlewared/migration/0001_cloudsync_sftp_private_key_file.py
themylogin: Traceback won't be interesting in this case, so did not add exc_info for sake of more compact logs.

PR #2924, opened 2019-04-20
NAS-101496 / 11.2 / fix legacy replication to allow spaces in dataset names

yocalebo:
Ticket: NAS-101496

After investigation, creating datasets with spaces in names doesn't work with legacy replication and hasn't worked for quite some time. (https://redmine.ixsystems.com/issues/24310) The work-around listed in that ticket does not work on TrueNAS 11.1-U7.1.

I've tried to keep the commit as trivial as possible since this is legacy replication and has been replaced in 11.3+.

So this commit fixes 2 primary issues. 

1. properly escape the volume and filesystems appropriately (i.e. "zpool/dataset with space")
2. fix the regex that is replacing tabs with spaces which breaks the ability to compare the remote side with the local side.

### DIFF
--- a/gui/tools/autorepl.py
+++ b/gui/tools/autorepl.py
@@ -355,7 +355,7 @@ for replication in replication_tasks:
         # We expect to see "on" in the output, or cannot open '%s': dataset does not exist
         # in the error.  To be safe, also check for children's readonly state.
         may_proceed = False
-        rzfscmd = "'zfs list -H -o readonly -t filesystem,volume -r \"%s\"'" % (remotefs_final)
+        rzfscmd = '"zfs list -H -o readonly -t filesystem,volume -r %s"' % (remotefs_final)
         sshproc = pipeopen('%s %s' % (sshcmd, rzfscmd))
         output, error = sshproc.communicate()
         error = error.strip('\n').strip('\r').replace('WARNING: ENABLED NONE CIPHER', '')
@@ -413,9 +413,9 @@ Hello,
 
     # Grab map from remote system
     if recursive:
-        rzfscmd = "'zfs list -H -t snapshot -p -o name,creation -r \"%s\"'" % (remotefs_final)
+        rzfscmd = '"zfs list -H -t snapshot -p -o name,creation -r \'%s\'"' % (remotefs_final)
     else:
-        rzfscmd = "'zfs list -H -t snapshot -p -o name,creation -d 1 -r \"%s\"'" % (remotefs_final)
+        rzfscmd = '"zfs list -H -t snapshot -p -o name,creation -d 1 -r \'%s\'"' % (remotefs_final)
     sshproc = pipeopen('%s %s' % (sshcmd, rzfscmd), debug)
     output, error = sshproc.communicate()
     error = error.strip('\n').strip('\r').replace('WARNING: ENABLED NONE CIPHER', '')
--- a/gui/tools/autorepl.py
+++ b/gui/tools/autorepl.py
@@ -320,7 +320,7 @@ for replication in replication_tasks:
         remote_zfslist[data[0]] = {'readonly': data[1] == 'on'}
 
     # Attempt to create the remote dataset.  If it fails, we don't care at this point.
-    rzfscmd = "'zfs create -o readonly=on "
+    rzfscmd = "zfs create -o readonly=on "
     ds = ''
     if "/" not in localfs:
         localfs_tmp = "%s/%s" % (localfs, localfs)
@@ -337,7 +337,7 @@ for replication in replication_tasks:
             if ds_full in remote_zfslist:
                 continue
             log.debug("ds = %s, remotefs = %s" % (ds, remotefs))
-            sshproc = pipeopen('%s %s \"%s\"' % (sshcmd, rzfscmd, ds_full) + "\'", quiet=True)
+            sshproc = pipeopen('%s %s \"%s\"' % (sshcmd, rzfscmd, ds_full), quiet=True)
             output, error = sshproc.communicate()
             error = error.strip('\n').strip('\r').replace('WARNING: ENABLED NONE CIPHER', '')
             # Debugging code
--- a/gui/tools/autorepl.py
+++ b/gui/tools/autorepl.py
@@ -127,7 +127,7 @@ def sendzfs(fromsnap, tosnap, dataset, localfs, remotefs, followdelete, throttle
         os.close(writefd)
 
     compress, decompress = compress_pipecmds(compression)
-    replcmd = '%s%s/usr/local/bin/pipewatcher $$ | %s "%s/sbin/zfs receive -F -d \'%s\' && echo Succeeded"' % (compress, throttle, sshcmd, decompress, remotefs)
+    replcmd = '%s%s/usr/local/bin/pipewatcher $$ | %s \"%s/sbin/zfs receive -F -d \'%s\' && echo Succeeded\"' % (compress, throttle, sshcmd, decompress, remotefs)
     log.debug('Sending zfs snapshot: %s | %s', ' '.join(cmd), replcmd)
     with open(templog, 'w+') as f:
         readobj = os.fdopen(readfd, 'rb', 0)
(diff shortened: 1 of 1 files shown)


yocalebo:
I agree that this change is gross but it was the way to fix this particular problem. I don't have a customer that is threatening to return a unit or anything but I do have a TrueNAS customer that can't use our replication because they have datasets with spaces in the name.

I did test this quite a bit and made sure, as best as I could, to not break anything else. I understand your skepticism but I can't tell a customer that they can't use our replication until TrueNAS 11.3 because that's a very long way away. So what would be our fix for them because the work-around listed in that original redmine ticket (https://redmine.ixsystems.com/issues/24310) does not work.

yocalebo:
As far as I can tell, this was the bare minimum to make it work. I tested the following filesystem layout between 2 VMs.

`sending                                                        2.93G  6.66G   117K  /mnt/sending
sending/more spaces                                            1.95G  6.66G  1.22G  /mnt/sending/more spaces
sending/more spaces/some more spaces                            750M  6.66G   750M  /mnt/sending/more spaces/some more spaces
sending/no-spaces                                              1000M  6.66G   500M  /mnt/sending/no-spaces
sending/no-spaces/nested spaces                                 500M  6.66G   500M  /mnt/sending/no-spaces/nested spaces`

Also, I'm quoting the zpool/dataset entirely instead of just the dataset.

So it will be "zpool/your dataset/another one" instead of zpool/"your dataset",

gui/tools/autorepl.py
```diff
@@ -285,16 +285,16 @@ def write_results():
...
-    remotefs_final = "%s%s%s" % (remotefs, localfs.partition('/')[1], localfs.partition('/')[2])
+    remotefs_final = '%s%s%s' % (remotefs, localfs.partition('/')[1], localfs.partition('/')[2])
 
     # Examine local list of snapshots, then remote snapshots, and determine if there is any work to do.
     log.debug("Checking dataset %s" % (localfs))
 
     # Grab map from local system.
     if recursive:
-        zfsproc = pipeopen('/sbin/zfs list -H -t snapshot -p -o name,creation -r "%s"' % (localfs), debug)
+        zfsproc = pipeopen('/sbin/zfs list -H -t snapshot -p -o name,creation -r \"%s\"' % (localfs), debug)
```
themylogin: Does this change intend to change anything?

gui/tools/autorepl.py
```diff
@@ -314,12 +314,13 @@ def write_results():
...
-    for i in re.sub(r'[ \t]+', ' ', output, flags=re.M).splitlines():
-        data = i.split()
+    for i in re.sub(r'\t+', ' ', output, flags=re.M).splitlines():
+        readonly_value = i.split()[-1]
+        data = [i.split(readonly_value)[0].rstrip(), readonly_value]
         remote_zfslist[data[0]] = {'readonly': data[1] == 'on'}
 
     # Attempt to create the remote dataset.  If it fails, we don't care at this point.
-    rzfscmd = "zfs create -o readonly=on "
+    rzfscmd = "\'" + "zfs create -o readonly=on "
```
themylogin: Do you want `"'zfs create -o readonly=on "`?

gui/tools/autorepl.py
```diff
@@ -314,12 +314,13 @@ def write_results():
     sshproc = pipeopen('%s %s' % (sshcmd, rzfscmd))
     output, error = sshproc.communicate()
     remote_zfslist = {}
-    for i in re.sub(r'[ \t]+', ' ', output, flags=re.M).splitlines():
-        data = i.split()
+    for i in re.sub(r'\t+', ' ', output, flags=re.M).splitlines():
```
themylogin:
```
output = 'beverages/vodka smirnoff/0.5l\toff\n'

for i in re.sub(r'\t+', ' ', output, flags=re.M).splitlines():
    readonly_value = i.split()[-1]
    data = [i.split(readonly_value)[0].rstrip(), readonly_value]
    print(data)

['beverages/vodka smirn', 'off']
# But I expected it to be ['beverages/vodka smirnoff/0.5l', 'off']
```

gui/tools/autorepl.py
```diff
@@ -412,9 +413,9 @@ def write_results():
 
     # Grab map from remote system
     if recursive:
-        rzfscmd = '"zfs list -H -t snapshot -p -o name,creation -r \'%s\'"' % (remotefs_final)
+        rzfscmd = "'zfs list -H -t snapshot -p -o name,creation -r \"%s\"'" % (remotefs_final)
```
themylogin: What do you want to change by swapping ' and " here?

gui/tools/autorepl.py
yocalebo: No, I guess it doesn't change anything. The idea was to be consistent by escaping double-quotes. Do you want me to remove it?

themylogin: Let's make this diff as small as possible please

gui/tools/autorepl.py
yocalebo: Correct, the zfs command needs to be prefixed with a single-quote so that I can escape the zpool/dataset with double-quotes on line 340.

gui/tools/autorepl.py
yocalebo: This works as expected for me.

Apr 22 08:33:04 truenas /autorepl.py: [tools.autorepl:318] data = ['receiving', 'on']
Apr 22 08:33:04 truenas /autorepl.py: [tools.autorepl:318] data = ['receiving/DatasetWithUpperCase', 'on']
Apr 22 08:33:04 truenas /autorepl.py: [tools.autorepl:318] data = ['receiving/dataset_with_underscores', 'on']
Apr 22 08:33:04 truenas /autorepl.py: [tools.autorepl:318] data = ['receiving/more spaces', 'on']
Apr 22 08:33:04 truenas /autorepl.py: [tools.autorepl:318] data = ['receiving/more spaces/0.5l', 'on']

gui/tools/autorepl.py
yocalebo: Being consistent by using single-quotes and escaping double-quotes with other parts of the code.

gui/tools/autorepl.py
themylogin: @yocalebo try dataset that contains `on` in the middle of it's name for readonly=on dataset or `off` in the middle of it's name for readonly=off

gui/tools/autorepl.py
themylogin: So it changes nothing, then we should avoid it.

gui/tools/autorepl.py
themylogin: Why so cryptic then?

gui/tools/autorepl.py
yocalebo: Understood. WIll do!

gui/tools/autorepl.py
yocalebo: Understood. Will revert this.

gui/tools/autorepl.py
yocalebo: I'm not sure I understand? Would you like me to add the single-quote inside the double-quote instead of using the "+" operator?

themylogin: `rzfscmd = "'zfs create -o readonly=on "`

yocalebo: Okay, done.

gui/tools/autorepl.py
```diff
@@ -336,7 +337,7 @@ def write_results():
             if ds_full in remote_zfslist:
                 continue
             log.debug("ds = %s, remotefs = %s" % (ds, remotefs))
-            sshproc = pipeopen('%s %s %s' % (sshcmd, rzfscmd, ds_full), quiet=True)
+            sshproc = pipeopen('%s %s \"%s\"' % (sshcmd, rzfscmd, ds_full) + "\'", quiet=True)
```
william-gr: What if we didnt change rzfscmd above and just did:

`sshproc = pipeopen('%s \'%s \"%s\"\'' % (sshcmd, rzfscmd, ds_full), quiet=True)`

wouldnt it make ir more clear? or is it used anywhere else?

william-gr: In fact, it seems we are breaking the bidirectional replication down below (if even used!?) by changing rzfscmd to have a single quote in beginning ?

yocalebo: The bi-directional replication isn't used. I'll escape in the `pipeopen` function.

gui/tools/autorepl.py
yocalebo: @themylogin sorry for delay. Please let me know if final commit makes sense.

gui/tools/autorepl.py
```diff
@@ -314,8 +314,9 @@ def write_results():
     sshproc = pipeopen('%s %s' % (sshcmd, rzfscmd))
     output, error = sshproc.communicate()
     remote_zfslist = {}
-    for i in re.sub(r'[ \t]+', ' ', output, flags=re.M).splitlines():
-        data = i.split()
+    for i in output.rstrip().split("\n"):
+        readonly_value = i.split()[-1]
+        data = [i.split("\t")[0].rstrip(), readonly_value]
```
themylogin: Why `rstrip`, dataset can have space as it's last name character e.g. `"data/work "`. 

What you need is
```
>>> "data/work \tyes".rsplit("\t",1)
['data/work ', 'yes']
```

yocalebo: I used rstrip to remove the trailing newline that comes back from the remote ssh command.

william-gr: That is handled with the first rstrip in `output.rstrip()`

yocalebo: Right, the 1st rstrip will strip the trailing newline but the second rstrip was to actually remove the trailing whitespace at the end of a dataset name. I did not realize that was even allowed or recognized in zfs....

yocalebo: @themylogin okay, I've made the commit to account for trailing whitespace in the dataset name. Let me know if that works.

william-gr:
Lets get this in, we need to freeze branches. Vladimir is off for the next hours

PR #2929, opened 2019-04-22
NAS-100016 / 11.3 / Support Amazon Glacier storage classes

themylogin:


src/middlewared/middlewared/rclone/remote/s3.py
```diff
@@ -24,7 +24,10 @@ class S3RcloneRemote(BaseRcloneRemote):
     ]
 
     task_schema = [
-        Str("encryption", enum=[None, "AES256"], default=None, null=True),
+        Str("encryption", title="Server-Side Encryption", enum=[None, "AES256"], default=None, null=True),
+        Str("storage_class", title="The storage class to use", enum=["", "STANDARD", "REDUCED_REDUNDANCY",
+                                                                     "STANDARD_IA", "ONEZONE_IA", "GLACIER",
+                                                                     "DEEP_ARCHIVE"]),
```
william-gr: How does this work with non-AWS? Shoudnt we have a default here?

themylogin:
@william-gr no, we don't, this value is just passed through to AWS API, rclone does not do anything about it.

I've tested it, it sets storage class on object (it's visible in S3 Web UI). @kmoore134 what are issued?

src/middlewared/middlewared/rclone/remote/s3.py
themylogin: It's not required, so we don't need a default. `""` is an rclone default and it works with non-AWS.

PR #2932, opened 2019-04-22
NAS-100100 / 11.3 / Allow disabling zilstat for SNMP

themylogin:
I've set `snmp_zilstat` to `False` by default because, if running it has performance implications, it's better that way.

Also migration is ready for backporting into 11.2, I think it may have sense there.


gui/services/migrations/0025_snmp_zilstat.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.8 on 2019-04-22 04:56
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import freenasUI.freeadmin.models.fields
+
+
+class Migration(migrations.Migration):
```
william-gr: I think we should be smarter about this field when upgrading.

Defaulting it to true if the service is marked to start on boot.

gui/services/models.py
```diff
@@ -1164,6 +1164,10 @@ class SNMP(Model):
         blank=True,
         help_text=_("These parameters will be added to /etc/snmpd.config."),
     )
+    snmp_zilstat = models.BooleanField(
+        verbose_name=_("Expose zilstat via SNMP"),
+        default=False,
```
william-gr: We should have a help text warning about performance implications of having this enabled.

themylogin:
@william-gr won't have time by 11.2-U4 or never?

PR #2947, opened 2019-04-24
NAS-101349 / 11.3 / Use Alerts for UPS events

sonicaj:
This PR introduces changes where we don't spam terminals on each UPS events and instead use Alerts system to notify user of different UPS events.
Ticket: #NAS-101349


src/middlewared/middlewared/alert/source/ups.py
```diff
@@ -0,0 +1,43 @@
...
+    category = AlertCategory.UPS
+    level = AlertLevel.INFO
+    title = 'UPS Communications established'
+    text = 'Communications with UPS %(ups)s established.'
+
+
+class UPSReplbattAlertClass(AlertClass, SimpleOneShotAlertClass):
+    category = AlertCategory.UPS
+    level = AlertLevel.CRITICAL
+    title = 'UPS Battery needs to be replaced'
```
william-gr: We should be using title case for every word
USB Battery Needs To Be Replace

Thats the unspoken rule we have been using, iirc

wblock: Yes, these should be title case.  But title case is not every word, there are odd rules.
There are online things that will do it correctly, mostly.

src/middlewared/middlewared/alert/source/ups.py
```diff
@@ -0,0 +1,43 @@
+from middlewared.alert.base import SimpleOneShotAlertClass, AlertClass, AlertCategory, AlertLevel
+
+
+class UPSBatteryLowAlertClass(AlertClass, SimpleOneShotAlertClass):
+    category = AlertCategory.UPS
+    level = AlertLevel.ALERT
+    title = 'UPS Battery LOW'
+    text = 'UPS %(ups)s Battery level low.'
```
wblock:
```suggestion
    text = 'UPS %(ups)s battery level low.'
```

src/middlewared/middlewared/alert/source/ups.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class UPSOnlineAlertClass(AlertClass, SimpleOneShotAlertClass):
+    category = AlertCategory.UPS
+    level = AlertLevel.INFO
+    title = 'UPS ONLINE'
```
wblock:
```suggestion
    title = 'UPS On Line Power'
```

src/middlewared/middlewared/alert/source/ups.py
```diff
@@ -0,0 +1,43 @@
...
+    level = AlertLevel.CRITICAL
+    title = 'UPS On Battery'
+    text = 'UPS %(ups)s is on battery power.'
+
+
+class UPSCommbadAlertClass(AlertClass, SimpleOneShotAlertClass):
+    category = AlertCategory.UPS
+    level = AlertLevel.CRITICAL
+    title = 'UPS Communication Lost'
+    text = 'Communications with UPS %(ups)s lost.'
```
wblock:
```suggestion
    text = 'Communication with UPS %(ups)s lost.'
```

src/middlewared/middlewared/alert/source/ups.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class UPSCommokAlertClass(AlertClass, SimpleOneShotAlertClass):
+    category = AlertCategory.UPS
+    level = AlertLevel.INFO
+    title = 'UPS Communications Established'
```
wblock:
```suggestion
    title = 'UPS Communication Established'
```

src/middlewared/middlewared/alert/source/ups.py
```diff
@@ -0,0 +1,43 @@
...
+    text = 'Communications with UPS %(ups)s established.'
```
wblock:
```suggestion
    text = 'Communication with UPS %(ups)s established.'
```

src/middlewared/middlewared/alert/source/ups.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class UPSReplbattAlertClass(AlertClass, SimpleOneShotAlertClass):
+    category = AlertCategory.UPS
+    level = AlertLevel.CRITICAL
+    title = 'UPS Battery Needs To Be Replaced'
```
wblock:
```suggestion
    title = 'UPS Battery Needs Replacement'
```

src/middlewared/middlewared/alert/source/ups.py
```diff
@@ -0,0 +1,43 @@
...
+    text = 'UPS %(ups)s Battery needs to be replaced.'
```
wblock:
```suggestion
    text = 'UPS %(ups)s battery needs replacement.'
```

src/middlewared/middlewared/plugins/ups.py
```diff
@@ -191,10 +191,31 @@ def driver_choices(self):
         old_config['toemail'] = ';'.join(old_config['toemail']) if old_config['toemail'] else ''
 
         if len(set(old_config.items()) ^ set(config.items())) > 0:
+            if config['identifier'] != old_config['identifier']:
```
themylogin: But you are going to restart UPS service anyway and they will be dismissed there.

src/middlewared/middlewared/plugins/ups.py
```diff
@@ -191,10 +191,31 @@ def driver_choices(self):
...
+            'ONBATT': 'UPSOnBattery',
+            'ONLINE': 'UPSOnline',
+            'REPLBATT': 'UPSReplbatt'
+        }
+
+    @private
+    async def dismiss_alerts(self):
+        config = await self.config()
+
+        for alert in (await self.alerts_mapping).values():
```
themylogin: `await self.alerts_mapping()`

src/middlewared/middlewared/plugins/ups.py
```diff
@@ -191,10 +191,31 @@ def driver_choices(self):
...
+            await self.middleware.call('alert.oneshot_delete', alert, {'ups': config['identifier']})
```
themylogin: We need to dismiss them for all UPSes.
Probably it's not possible with current alerts implementation and you might need to extend it a little.

src/middlewared/middlewared/plugins/ups.py
sonicaj: I am not so sure about that - if the identifier changes - new identifier will be used to dismiss them and that wouldn't work ?

PR #2954, opened 2019-04-25
NAS-101506 / 11.3 / Update system dataset error messages

sonicaj:
This commit updates error messages for pool which hosts system dataset if locking or a passphrase action is performed on the respected pool.
Ticket: #NAS-101506

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -1351,8 +1351,8 @@ class PoolService(CRUDService):
             # Only allow removing passphrase for pools being used by system dataset service
             verrors.add(
                 'id',
-                'A passphrase on a pool containing the system dataset can only be removed. '
-                'It cannot be set or changed.'
+                f'Pool {pool["name"]} contains the system dataset. Passphrases are not allowed on the '
+                'system dataset pool.'
             )
 
         # For historical reasons (API v1.0 compatibility) we only require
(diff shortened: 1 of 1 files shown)


sonicaj:
@wblock i have updated the error messages as you suggested but there is a point to note. If a pool was being used which already had a passphrase set and it hosted system dataset, we only allow removal of that passphrase from the given pool and i think it should be mentioned.

wblock:
> @wblock i have updated the error messages as you suggested but there is a point to note. If a pool was being used which already had a passphrase set and it hosted system dataset, we only allow removal of that passphrase from the given pool and i think it should be mentioned.

Good point, but when should we tell them that?  If the user has the system dataset on a pool with passphrase and upgrades to a current version, it ought to immediately alert them that they must remove the passphrase or move the system dataset or the system will not work.  It should not just show that message when they go to change the passphrase.

sonicaj:
@wblock so now this is how it is going to work - when the user has a system dataset configured on a pool which has passphrase set and upgrades to the latest version where we don't want system dataset configured on such pools, that pool would be locked because of how upgrades work. So in order to retrieve their system dataset contents, they will have to unlock it and then we only allow removal of passphrase from that pool. I am afraid, there is not much we can do about the versions they are already in ( i mean we can't introduce changes their and whatever we do, when user upgrades, that pool will automatically be locked ).
So i think it is reasonable to tell them about the removal part when they try to change the passphrase

wblock:
Can we warn the user before the upgrade begins, or at least before it reboots and then fails to come back up?  I'm thinking it should say warn them ("System dataset is on a pool with a passphrase.") and then give them the option to remove the passphrase or cancel the upgrade.

sonicaj:
Just to clarify @wblock, even if system dataset is on a pool which has passphrase set and that pool is locked on upgrades, the new version will boot just fine. 
Secondly what you are suggesting is not possible unfortunately so our options are limited :)

sonicaj:
ping @wblock

PR #2956, opened 2019-04-25
fix(middleward/zfs): typo in service name

william-gr:


william-gr:
That cant happen, method is returned earlier if pool is None

PR #2959, opened 2019-04-26
NAS-101572 / 11.3 / Allow more ip formats for jails

sonicaj:
This commit inreases the ip formats we currently support to create jails reflecting iocage. This is particularly needed for new UI so they can just pass ip for vnet+ip jails.


william-gr:
I would rather not review this one ; )

PR #2960, opened 2019-04-26
NAS-101544 / 11.3 / move LDAP to middleware plugin

anodos325:


src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -0,0 +1,639 @@
...
+        self._handle = None
+        self._rootDSE = None
+        return
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, typ, value, traceback):
+        if self._isopen:
+            self._close()
```
william-gr: Dont we want to reraise the exception if it occurs, after closing connection?

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -0,0 +1,639 @@
...
+    def __init__(self, **kwargs):
+        super(LDAPQuery, self).__init__()
+        self.ldap = kwargs.get('conf')
+        self.logger = kwargs.get('logger')
+        self.hosts = kwargs.get('hosts')
+        self.pagesize = 1024
+        self._isopen = False
+        self._handle = None
+        self._rootDSE = None
+        return
```
william-gr: This return can be removed, it doesnt have any effect for init

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -0,0 +1,639 @@
...
+                        ldap.OPT_X_TLS_REQUIRE_CERT,
+                        ldap.OPT_X_TLS_ALLOW
+                    )
+
+                if SSL(self.ldap['ssl']) == SSL.USESSL:
+                    try:
+                        self._handle.start_tls_s()
+
+                    except ldap.LDAPError as e:
+                        self.logger.debug('%s', e)
```
william-gr: Should you `continue` if we hit this error?

A more descriptive debug error message would be useful as well

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -0,0 +1,639 @@
...
+        await self.middleware.call('etc.generate', 'pam')
+        if ldap['has_samba_schema']:
+            await self.middleware.call('etc.generate', 'smb')
+            await self.middleware.call('service.restart', 'smb')
+        await self.middleware.call('cache.pop', 'LDAP_State')
+        await self.nslcd_cmd('onestop')
+
+    @private
+    @job(lock='fill_ldap_cache')
+    def fill_ldap_cache(self, force=False):
```
william-gr: Second argument will be `job` because of the decorator

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -0,0 +1,639 @@
...
+            })
+
+        self.middleware.call_sync('cache.put', 'LDAP_cache', cache_data, 86400)
+
+    @private
+    async def get_ldap_cache(self):
+        """
+        """
+        if not await self.middleware.call('cache.has_key', 'LDAP_cache'):
+            await self.middleware.run_in_thread(self.fill_ldap_cache)
```
william-gr: await self.middleware.call('ldap.fill_ldap_cache') is supposed to return a job
which we can then do:
await job.wait()

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -0,0 +1,641 @@
...
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, typ, value, traceback):
+        if self._isopen:
+            try:
+                self._close()
+            except Exception as e:
+                raise CallError(f'Failed to close connection to LDAP server: {e}')
```
william-gr: You may have misunderstood me.
If an exception occurs within this context it will be silently skipped.
You need to re-raise it:

if typ is not None:
    raise

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -580,10 +583,9 @@ def fill_ldap_cache(self, force=False):
 
     @private
     async def get_ldap_cache(self):
-        """
-        """
         if not await self.middleware.call('cache.has_key', 'LDAP_cache'):
-            await self.middleware.run_in_thread(self.fill_ldap_cache)
+            cache_job = await self.middleware.run_in_thread(self.fill_ldap_cache)
```
william-gr: I am not sure this works as expected, proper way would be `await self.middleware.call('ldap.fill_ldap_cache')`

william-gr:
@themylogin can you also review here, please? Need another pair of eyes ; ) Big PR

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -0,0 +1,641 @@
...
+        return data
+
+    @private
+    async def ldap_compress(self, data):
+        data['hostname'] = ','.join(data['hostname'])
+        return data
+
+    @private
+    async def ldap_validate(self, ldap):
+        port = 636 if SSL(self.ldap['ssl']) == SSL.USESSL else 389
```
themylogin: `self.ldap` -> `ldap`?

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -0,0 +1,641 @@
...
+        for h in ldap['hostname']:
+            self.middleware.call('ldap.port_is_listening', h, port, ldap['dns_timeout'])
```
themylogin: Probably should catch `CallError` and re-raise it as validation error?

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -0,0 +1,641 @@
...
+        Str('basedn', required=True),
+        Str('binddn'),
+        Str('bindpw', private=True),
+        Bool('anonbind', default=False),
+        Str('usersuffix'),
+        Str('groupsuffix'),
+        Str('passwordsuffix'),
+        Str('machinesuffix'),
+        Str('sudosuffix'),
+        Str('ssl', default='off', enum=['off', 'on', 'start_tls']),
```
themylogin: @william-gr is it ok that we use lowercase constants here?

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -0,0 +1,641 @@
...
+                return (await self.middleware.call('cache.get', 'LDAP_State'))
+            except KeyError:
+                await self.started()
+                return (await self.middleware.call('cache.get', 'LDAP_State'))
+
+    @private
+    async def nslcd_cmd(self, cmd):
+        nslcd = await run(['service', 'nslcd', cmd], check=False)
+        if nslcd.returncode != 0:
+            raise CallError(f'nslcd failed to {cmd} with errror: {nscld.stderr.decode()}', errno.EFAULT)
```
themylogin: typo: `nscld`

src/middlewared/middlewared/plugins/ldap.py
themylogin: We don't need to re-raise anything in `__exit__`, it is done automatically

src/middlewared/middlewared/plugins/ldap.py
william-gr: Its not mandatory but I have asked if it could be easily changed to uppercase before (for all enums), seems its not trivial?

What do you think @anodos325 ? Too much effort?

src/middlewared/middlewared/plugins/ldap.py
william-gr: Does it? I had a completely wrong idea how that worked, then.

Sorry @anodos325 :D

src/middlewared/middlewared/plugins/ldap.py
anodos325: For the life of me I can't type those letters in the correct order twice in a row.

src/middlewared/middlewared/plugins/ldap.py
anodos325: Not too much. I'll make the change here and in the AD plugin for consiste

src/middlewared/middlewared/plugins/ldap.py
anodos325: Indeed.

src/middlewared/middlewared/etc_files/local/smb4.conf
```diff
@@ -158,32 +158,32 @@
...
                 pc.update({'passdb backend': ldap_pdb_backend})
                 pc.update({'ldap admin dn': db['ldap']['ldap_binddn']})
-                pc.update({'ldap suffix': db['ldap']['ldap_basedn']}) 
+                pc.update({'ldap suffix': db['ldap']['ldap_basedn']})
                 pc.update({'ldap user suffix': db['ldap']['ldap_usersuffix']})
                 pc.update({'ldap group suffix': db['ldap']['ldap_groupsuffix']})
                 pc.update({'ldap machine suffix': db['ldap']['ldap_machinesuffix']})
                 pc.update({'workgroup': db['cifs']['workgroup'].upper()})
-                if db['ldap']['ldap_ssl'] is not "start_tls":
+                if db['ldap']['ldap_ssl'] is not "START_TLS":
```
william-gr: != ?

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -334,6 +340,11 @@ class Config:
 
         if must_reload:
             if new['enable']:
+                try:
+                    await self.middleware.call('ldap.ldap_validate', new)
+                except Exception as e:
+                    ValidationError('ldap_update', str(e))
```
william-gr: Missing raise

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -0,0 +1,654 @@
...
+                    break
+                except Exception as e:
+                    self.logger.debug(f'Failed to bind to [{uri}] using [{self.ldap["binddn"]}]: {e}')
+                    saved_simple_error = e
+                    continue
+
+            if res:
+                self._isopen = True
+            elif saved_gssapi_error:
+                raise CallError(saved_gssapi_error)
```
william-gr: Since these are exceptions objects we need CallError(str(saved_gssapi_error))

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -0,0 +1,654 @@
...
+            data.pop('bindpw')
+
+        return data
+
+    @private
+    async def ldap_validate(self, ldap):
+        port = 636 if SSL(ldap['ssl']) == SSL.USESSL else 389
+        for h in ldap['hostname']:
+            self.middleware.call('ldap.port_is_listening', h, port, ldap['dns_timeout'])
+        self.middleware.call('ldap.validate_credentials')
```
william-gr: Missing `await` here and call above

src/middlewared/middlewared/plugins/ldap.py
anodos325: Sounds good.

PR #2962, opened 2019-04-28
NAS-101569 / 11.3 / Send emails for manual cron runs

sonicaj:

1) Use middlewared for manual execution of a cron job task.
2) Remove runnow script.
3) Ability to send emails of manually executed cron job tasks.

Ticket: #101569


src/middlewared/middlewared/plugins/cron.py
```diff
@@ -195,3 +198,68 @@ def cron_extend(self, data):
...
+                await mail_job.wait()
+                if mail_job.error:
+                    self.middleware.logger.debug(f'Failed to send email for CronTask manual run: {mail_job.error}')
+            else:
+                job.set_progress(
+                    95,
+                    'Email for root user not configured. Skipping sending mail.'
+                )
+
+            self.middleware.logger.debug(f'Executed CronTask - {cron_cmd}: {stdout}')
```
william-gr: This is a job, if we want we can put logs to it, using logs_excerpt, lets not pollute middleware log with this

src/middlewared/middlewared/plugins/cron.py
```diff
@@ -195,3 +198,68 @@ def cron_extend(self, data):
...
+        )
+        stdout, stderr = await p.communicate()
+
+        job.set_progress(
+            85,
+            'Executed Cron Task'
+        )
+
+        if stdout:
+            email = (await self.middleware.call('user.query', [['username', '=', 'root']], {'get': True}))['email']
```
william-gr: Lets uid = 0 to be safer

src/middlewared/middlewared/plugins/cron.py
```diff
@@ -195,3 +198,68 @@ def cron_extend(self, data):
...
+        job.set_progress(
+            10,
+            'Executing Cron Task'
+        )
+
+        p = await Popen(
+            cron_cmd, stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT, shell=True
+        )
+        stdout, stderr = await p.communicate()
```
william-gr: You're ignoring the whole block of running the cron with the specified user and setting context.

See how it was done for rsync, we need to not duplicate code while doing it.

src/middlewared/middlewared/plugins/cron.py
william-gr: Actually, doesnt cron send email to the specified user?

src/middlewared/middlewared/utils/__init__.py
```diff
@@ -134,6 +137,60 @@ def setusercontext(user):
...
+    os.environ['PATH'] = (
+        '/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin:/root/bin'
+    )
+    proc = subprocess.Popen(
+        commandline, shell=True, stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT
+    )
+
+    # logger(1) does not honor original exit value so we use syslog module instead
+    syslog.openlog(ident=log_ident)
```
william-gr: We used to use syslog because there was no way to retrieve a specific job output.
I am thinking it should be OK to just drop syslog now in favor of job log? Thoughts?

src/middlewared/middlewared/plugins/cron.py
```diff
@@ -195,3 +198,76 @@ def cron_extend(self, data):
...
+                cron_task['command'], cron_task['stdout'], cron_task['stderr']
+            )[7:]
+        )
+
+        job.set_progress(
+            10,
+            'Executing Cron Task'
+        )
+
+        syslog.openlog('cron', facility=syslog.LOG_CRON)
```
william-gr: You're opening/closing log in middlewared process but rsync/cron commands run in another process, isnt that wrong/weird?

sonicaj: So for rsync, syslog is not happening at all - we are only logging to the jail log which i think is sufficient. For cron, i don't think it's wrong. Callback is being run in middlewared process and syslog is also open in that one. Please let me know if i am missing something 🤔

PR #2963, opened 2019-04-29
Bring up vlan interfaces on creation

freqlabs:
Issue: NAS-100773


freqlabs:
When an interface is created in the UI, middleware does bring the interface up when creating it. However, middleware doesn't create the interfaces during system startup, netif does.

william-gr:
You didnt understand my question. Middleware should create the interface with "up" in int_options database field, which would put it in rc.conf already.

Let me rephrase. Why do you need to `up` the interface at creation time, and not on configure time?

freqlabs:
Oh got it, yes "up" is not in the interface options in the config database, you're right.

freqlabs:
This should be fixed in middleware instead.

PR #2965, opened 2019-04-29
NAS-101303 / 11.3 / fix(middlewared/zfs): handle ENOENT on snapshot query

william-gr:
Ticket:	NAS-101303

### DIFF
--- a/src/middlewared/middlewared/plugins/zfs.py
+++ b/src/middlewared/middlewared/plugins/zfs.py
@@ -619,7 +619,12 @@ class ZFSSnapshot(CRUDService):
         Query all ZFS Snapshots with `query-filters` and `query-options`.
         """
         # Special case for faster listing of snapshot names (#53149)
-        if options and options.get('select') == ['name']:
+        if (
+            options and options.get('select') == ['name'] and (
+                not filters or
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -634,21 +639,30 @@ def query(self, filters=None, options=None):
...
+                    snapshots.append(zfs.get_snapshot(filters[0][2]).__getstate__())
                 except libzfs.ZFSException as e:
                     if e.code != libzfs.Error.NOENT:
                         raise
-                    snapshots = []
             else:
-                snapshots = [i.__getstate__() for i in list(zfs.snapshots)]
+                for i in zfs.snapshots:
+                    try:
+                        snapshots.append(i.__getstate__())
```
sonicaj: How about we retrieve a serialized version directly from py-libzfs ? It gives us better performance.
I'll double check to be sure but i think the format of getstate is respected thus avoiding any unintended consequences/surprises.

william-gr: This needs to be backported for 11.2, so not yet.

PR #2967, opened 2019-04-29
NAS-101574 / 11.3 / feat(middlewared/network): do not allow interfaces changes with HA enabled

william-gr:
Ticket:	NAS-101574

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -565,6 +565,14 @@ class InterfaceService(CRUDService):
         for i in self._original_datastores['laggmembers']:
             await self.middleware.call('datastore.insert', 'network.lagginterfacemembers', i)
 
+    async def __check_failover_disabled(self):
+        if await self.middleware.call('system.is_freenas'):
+            return
+        if not await self.middleware.call('failover.licensed'):
+            return
+        if not (await self.middleware.call('failover.config'))['disabled']:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -565,6 +565,14 @@ def iface_extend(self, iface_state, configs, is_freenas, fake=False):
         for i in self._original_datastores['laggmembers']:
             await self.middleware.call('datastore.insert', 'network.lagginterfacemembers', i)
 
+    async def __check_failover_disabled(self):
+        if await self.middleware.call('system.is_freenas'):
+            return
+        if not await self.middleware.call('failover.licensed'):
+            return
+        if not (await self.middleware.call('failover.config'))['disabled']:
+            raise CallError('Disable failover before performing interfaces changes.')
```
wblock: 'Failover must be disabled before making network interface changes.'

src/middlewared/middlewared/plugins/network.py
```diff
@@ -615,6 +624,7 @@ def iface_extend(self, iface_state, configs, is_freenas, fake=False):
         the interfaces changes happened as planned from the user. If checkin does not happen
```
wblock: With `rollback` true (default), changes will be rolled back if they have not been applied.
Network changes are rolled back after `checkin_timeout` (time in seconds) passes without
the user committing them.

PR #2970, opened 2019-04-29
NAS-100168 / 11.3 / disk.sata_dom_lifetime_left

themylogin:


src/middlewared/middlewared/plugins/disk.py
```diff
@@ -1722,6 +1722,18 @@ def unlabel(self, disk):
...
             asyncio.ensure_future(camcontrol_idle())
 
+    @private
+    async def sata_dom_lifetime_left(self, devname):
+        try:
+            smartctl = (await run('smartctl', '-A', devname, encoding="utf8", errors="ignore")).stdout
+        except subprocess.CalledProcessError:
+            return None
+
+        m = re.search(r'^164\s+.*\s+([0-9]+)$', smartctl, re.MULTILINE)
```
william-gr: We could have a global var for this regex so its not compiled every time this method is called.

PR #2977, opened 2019-05-01
NAS-101566 / 11.2 / Failover alerts fixes

themylogin:
So after failover the first e-mail will be:

```
New alerts:
* Active Controller - Enclosure 0 (iX 4024Sp c205) is HEALTHY
* Active Controller - Failed to check failover status with the other node: [EHOSTDOWN] Standby node is down

Gone alerts:
* Standby Controller - Enclosure 0 (iX 4024Ss c205) is HEALTHY
```

We get an alert about enclosure because it has different name on different systems, how do we want to handle that?

Also I've left one failover-related alert because how else will we know that other system is unavailable?

After other system gets ready, we get another e-mail:

```
Gone alerts:

* Active Controller - Failed to check failover status with the other node: [EHOSTDOWN] Standby node is down
```

Meaning that everything is back to normal

### DIFF
--- a/src/middlewared/middlewared/alert/base.py
+++ b/src/middlewared/middlewared/alert/base.py
@@ -127,7 +127,11 @@ class AlertService:
     async def _format_alerts(self, alerts, gone_alerts, new_alerts):
         product_name = await self.middleware.call("system.product_name")
         hostname = (await self.middleware.call("system.info"))["hostname"]
-        return format_alerts(product_name, hostname, alerts, gone_alerts, new_alerts)
+        if not await self.middleware.call("system.is_freenas"):
+            node_map = await self.middleware.call("alert.node_map")
+        else:
+            node_map = None
+        return format_alerts(product_name, hostname, node_map, alerts, gone_alerts, new_alerts)
 
 
 class ThreadedAlertService(AlertService):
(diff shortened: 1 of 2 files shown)


william-gr:
Perhaps have a timer to enable that alert again? Or get rid of the failover alert in favor of the alert system?

yocalebo:
> I am not sure we should have a `Active Controller - Failed to check failover status with the other node: [EHOSTDOWN] Standby node is down` alert.
> 
> The user already gets an email when a failover happens. Thoughts @yocalebo ?
> 
> About the enclosure, caleb says that should not normally happens, it does because they were flashed with wrong firmware.

Yeah, it shouldn't be saying Standby node is down.

Would logic like this be simpler to implement?

1. if failover occurs, have `carp-state-change-hook.py` call the necessary alert framework to send an email stating "failed over from one controller to the other blah blah"
2. don't generate any alerts in the webUI
3. if after 15mins (this is how long an m-series can take to reboot) we can't ```core.ping``` the remote controller, then generate alert in the webUI stating "standy node is down"?

@themylogin 
The difference with the enclosure names should never happen in the field. As a matter of fact, that should be an alert because that means the /dev/ses0 is flashed with different firmwares or is different hardware or something is wrong.

PR #2978, opened 2019-05-01
NAS-101608 / 11.3 / Properly restart UPS

sonicaj:
When UPS service is restarted, upsmon daemon in some cases has not stopped it's multiple processes entirely and is in the process of doing so when we initiate a restart. Restart fails in this case as old process is still cleaning itself. This commit makes sure we properly wait for upsmon to die and then initiate a restart.

### DIFF
--- a/src/middlewared/middlewared/plugins/service.py
+++ b/src/middlewared/middlewared/plugins/service.py
@@ -11,7 +11,7 @@ import time
 from subprocess import DEVNULL, PIPE
 
 from middlewared.schema import accepts, Bool, Dict, Ref, Str
-from middlewared.service import filterable, CallError, CRUDService
+from middlewared.service import filterable, CallError, CRUDService, job
 from middlewared.utils import Popen, filter_list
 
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/service.py
```diff
@@ -806,15 +806,41 @@ def _initializing_smartd_pid(self):
...
+        # Also we retrieve upsmon processes before executing a stop in case any process
+        # becomes a zombie and psutil fails to retrieve the name for it
+        upsmon_processes = set()
+        for proc in psutil.process_iter():
+            with contextlib.suppress(psutil.ZombieProcess):
+                if proc.name() == 'upsmon':
+                    upsmon_processes.add(proc)
+
+        await self._service("nut_upsmon", "stop", force=True, onetime=True)
+        await self.middleware.call('service._wait_for_upsmon', upsmon_processes)
```
themylogin: No need for separate method, you can do `self.middleware.run_in_thread(psutil.wait_procs, processes, timeout=30)`

sonicaj: Done :)

src/middlewared/middlewared/plugins/service.py
```diff
@@ -807,14 +807,34 @@ def _initializing_smartd_pid(self):
             asyncio.ensure_future(self.restart('collectd'))
 
     async def _restart_ups(self, **kwargs):
+        await self.middleware.call('service._restart_ups_job')
+
+    @job(lock='restart_upsmon')
+    async def _restart_ups_job(self, job, **kwargs):
```
william-gr: service.restart is supposed to return a boolean whether the service started or not.

Now it wont be possible since we wont wait for it?

src/middlewared/middlewared/plugins/service.py
```diff
@@ -807,14 +807,34 @@ def _initializing_smartd_pid(self):
...
-        await self._service("nut_upsmon", "restart", **kwargs)
-        await self._service("nut_upslog", "restart", **kwargs)
+        await self._service("nut", "stop", force=True, onetime=True)
+        # We need to wait on upsmon service to die properly as multiple processes are
+        # associated with it and in most cases they haven't exited when a restart is initiated
+        # for upsmon which fails as the older process is still running.
+        # Also we retrieve upsmon processes before executing a stop in case any process
+        # becomes a zombie and psutil fails to retrieve the name for it
+        upsmon_processes = set()
+        for proc in psutil.process_iter():
```
william-gr: I am not comfortable using this in the event loop. How about `pgrep`?

src/middlewared/middlewared/plugins/service.py
sonicaj: Yes, i am afraid that is a drawback using this approach. We can't trust service.restart anymore for ups.restart. After discussing with @themylogin we decided job is the best way to accomplish this because of the varying time upsmon might take to fully stop. Please let me know if you think we should handle this in another way

src/middlewared/middlewared/plugins/service.py
sonicaj: Sure, i can have it updated asap

src/middlewared/middlewared/plugins/service.py
william-gr: Hm, why? How about it kick it with a big hammer? Wait a few seconds, if it didnt stop, kill -9 everything.

sonicaj: Perfect :D It raises a upssched event by spawning a new process for it, so i guess we can wait for 15 seconds and if still it is going on, kick it with a big hammer :P ?

william-gr: I will say 10, unless you can think anything important its doing that grants a clean shutdown of the process? I cant

sonicaj: Well the only thing i have observed is the upssched event in which we also send a mail. So i think we can have it 10 for now, and if there are any issues later on, we can revisit this. Will have the changes up for your review soon. Thank you

PR #2983, opened 2019-05-01
QE tkt-86739 Adding FreeNAS API test to start and stop ups service

ericbsd:
Related to NAS-101608

### DIFF
--- a/tests/api2/ups.py
+++ b/tests/api2/ups.py
@@ -82,7 +82,7 @@ def test_06_starting_ups_service():
     }
     results = POST('/service/start/', payload)
     assert results.status_code == 200, results.text
-    sleep(2)
+    sleep(1)
 
 
 def test_07_look_UPS_service_status_is_running():
@@ -111,13 +111,17 @@ def test_10_stop_ups_service():
     }
     results = POST('/service/stop/', payload)
     assert results.status_code == 200, results.text
-    sleep(2)
+    sleep(1)
 
 
 def test_11_look_UPS_service_status_is_stopped():
     results = GET(f'/service/id/{ups_id}/')
     assert results.status_code == 200, results.text
     assert results.json()['state'] == 'STOPPED', results.text
+
+
+def test_12_wait_30_second():
+    sleep(30)
 
 
 def test_12_Change_UPS_options():
@@ -132,7 +136,7 @@ def test_12_Change_UPS_options():
     }
     results = PUT('/ups/', payload)
     assert results.status_code == 200, results.text
-    sleep(2)
+    sleep(1)
 
 
 @pytest.mark.parametrize('data', second_ups_list)
@@ -149,7 +153,7 @@ def test_14_starting_ups_service():
     }
     results = POST('/service/start/', payload)
     assert results.status_code == 200, results.text
-    sleep(2)
+    sleep(1)
 
 
 def test_15_look_UPS_service_status_is_running():
--- a/tests/api2/ups.py
+++ b/tests/api2/ups.py
@@ -6,7 +6,7 @@ import os
 import sys
 apifolder = os.getcwd()
 sys.path.append(apifolder)
-from functions import GET, PUT
+from functions import GET, PUT, POST
 
 ups_dc_list = list(GET('/ups/driver_choices/').json().keys())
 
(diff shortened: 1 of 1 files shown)


tests/api2/ups.py
```diff
@@ -53,22 +69,62 @@ def test_02_Set_UPS_options():
...
+def test_06_starting_ups_service():
+    payload = {
+        "service": "ups",
+        "service-control": {
+            "onetime": True,
+        }
+    }
+    results = POST('/service/start/', payload)
+    assert results.status_code == 200, results.text
+    sleep(1)
```
sonicaj: This should be good, but a delay of perhaps 5 seconds would be better ?

tests/api2/ups.py
```diff
@@ -53,22 +69,62 @@ def test_02_Set_UPS_options():
...
+def test_10_stop_ups_service():
+    payload = {
+        "service": "ups",
+        "service-control": {
+            "onetime": True,
+        }
+    }
+    results = POST('/service/stop/', payload)
+    assert results.status_code == 200, results.text
+    sleep(1)
```
sonicaj: Let's increase this to 30 please. upsmon daemon can take some time to fully exit even if the call returns back. Also we could skip the `sleep(30)` below then

tests/api2/ups.py
```diff
@@ -80,25 +136,44 @@ def test_06_Change_UPS_options():
     }
     results = PUT('/ups/', payload)
     assert results.status_code == 200, results.text
+    sleep(1)
```
sonicaj: We should be good to have this removed

tests/api2/ups.py
```diff
@@ -80,25 +136,44 @@ def test_06_Change_UPS_options():
...
+def test_14_starting_ups_service():
+    payload = {
+        "service": "ups",
+        "service-control": {
+            "onetime": True,
+        }
+    }
+    results = POST('/service/start/', payload)
+    assert results.status_code == 200, results.text
+    sleep(1)
```
sonicaj: Same as above comment, let's for safety increase this to 5 seconds please.

tests/api2/ups.py
```diff
@@ -107,19 +182,42 @@ def test_10_get_ups_driver_choice():
 
 
 @pytest.mark.parametrize('dkey', ups_dc_list)
-def test_11_check_ups_driver_choice_info_(dkey):
+def test_19_check_ups_driver_choice_info_(dkey):
     driver_choices = dkey.partition('$')[2]
```
sonicaj: While we are making a few changes, let's please name this to `driver_choice`, it can confuse a reader.

tests/api2/ups.py
```diff
@@ -107,19 +182,42 @@ def test_10_get_ups_driver_choice():
...
+
+def test_23_stop_ups_service():
+    payload = {
+        "service": "ups",
+        "service-control": {
+            "onetime": True,
+        }
+    }
+    results = POST('/service/stop/', payload)
+    assert results.status_code == 200, results.text
```
sonicaj: Some delay here please as above.

tests/api2/ups.py
ericbsd: test_12 has a sleep 30 second.

sonicaj: Yes but we check if the service has stopped before that sleep. So we can sleep after stopping and then check if the service has stopped and then start it / make changes etc. Consolidating the sleep ; )

sonicaj: Also you have 2 `test_12` this way : D

sonicaj:
Actually @ericbsd can you hold off any changes for a few please ? I am discussing with William to see if we could perhaps force kill it and then huge 30 sec sleeps wouldn't be necessary. I can update you once we have decided that and we can have this finalized ? Thank you

tests/api2/ups.py
ericbsd: Yeah I forgot thanks

ericbsd:
> Actually @ericbsd can you hold off any changes for a few please ? I am discussing with William to see if we could perhaps force kill it and then huge 30 sec sleeps wouldn't be necessary. I can update you once we have decided that and we can have this finalized ? Thank you

Oh, I just saw your post, so I will wait for your feedback.

sonicaj:
Thank you @ericbsd

PR #2986, opened 2019-05-03
Don't make restarting ups a job

sonicaj:
This commit removes restarting ups service as a job and we wait for upsmon for 10 seconds to cleanly exit, if it does not, we kill the processes.

### DIFF
--- a/src/middlewared/middlewared/plugins/service.py
+++ b/src/middlewared/middlewared/plugins/service.py
@@ -11,8 +11,8 @@ import time
 from subprocess import DEVNULL, PIPE
 
 from middlewared.schema import accepts, Bool, Dict, Ref, Str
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/service.py
```diff
@@ -807,28 +807,27 @@ def _initializing_smartd_pid(self):
...
+            gone, alive = await self.middleware.run_in_thread(
+                psutil.wait_procs,
+                map(
+                    lambda v: psutil.Process(int(v)),
+                    upsmon_processes.stdout.split()
+                ),
+                timeout=10
+            )
+            if alive:
+                p = await run(['pkill', '-x', 'upsmon'], check=False)
```
william-gr: Is SIGTERM enough? Why the regular stop doesnt work then?

PR #2992, opened 2019-05-05
NAS-100104 / 12 / Run manual smartctl tests

sonicaj:
This commit adds support in middlewared so a user can trigger manual smartctl tests for disks specified.
Ticket: NAS-100104

### DIFF
--- a/src/middlewared/middlewared/plugins/smart.py
+++ b/src/middlewared/middlewared/plugins/smart.py
@@ -313,6 +313,11 @@ class SMARTTestService(CRUDService):
         )
     )
     async def manual_test(self, disks):
+        """
+        Run manual SMART tests for `disks`.
+
+        `type` indicates what type of SMART test will be ran and must be specified.
+        """
         verrors = ValidationErrors()
         if not disks:
             verrors.add(
--- a/src/middlewared/middlewared/plugins/smart.py
+++ b/src/middlewared/middlewared/plugins/smart.py
@@ -1,4 +1,5 @@
 import functools
+import re
 from itertools import chain
 
 from middlewared.common.camcontrol import camcontrol_list
@@ -8,6 +9,9 @@ from middlewared.validators import Email, Range, Unique
 from middlewared.service import CRUDService, filterable, filter_list, private, SystemServiceService, ValidationErrors
 from middlewared.utils import run
 from middlewared.utils.asyncio_ import asyncio_map
+
+
+RE_TIME_DETAILS = re.compile(r'test will complete after(.*)', re.IGNORECASE)
 
 
 async def annotate_disk_smart_tests(devices, disk):
@@ -295,6 +299,102 @@ class SMARTTestService(CRUDService):
         await self._service_change('smartd', 'restart')
 
         return response
+
+    @accepts(
+        List(
+            'disks', items=[
+                Dict(
+                    'disk_run',
+                    Str('identifier', required=True),
+                    Str('mode', enum=['FOREGROUND', 'BACKGROUND'], default='BACKGROUND'),
+                    Str('type', enum=['LONG', 'SHORT', 'CONVEYANCE', 'OFFLINE'], required=True),
+                )
+            ]
+        )
+    )
+    async def manual_test(self, disks):
+        verrors = ValidationErrors()
+        if not disks:
+            verrors.add(
+                'disks',
+                'Please specify at least one disk.'
+            )
+        else:
+            test_disks_list = []
+            disks_data = await self.middleware.call('disk.query')
+            devices = await camcontrol_list()
+
+            for index, disk in enumerate(disks):
+                for d in disks_data:
+                    if disk['identifier'] == d['identifier']:
+                        current_disk = d
+                        test_disks_list.append({
+                            'disk': current_disk['name'],
+                            **disk
+                        })
+                        break
+                else:
+                    verrors.add(
+                        f'disks.{index}.identifier',
+                        f'{disk["identifier"]} is not valid. Please provide a valid disk identifier.'
+                    )
+                    continue
+
+                if current_disk['name'] is None:
+                    verrors.add(
+                        f'disks.{index}.identifier',
+                        f'Test cannot be performed for {disk["identifier"]} disk. Failed to retrieve name.'
+                    )
+
+                if current_disk['name'].startswith('nvd'):
+                    verrors.add(
+                        f'disks.{index}.identifier',
+                        f'Test cannot be performed for {disk["identifier"]} disk. NVMe devices cannot be mapped yet.'
+                    )
+
+                device = devices.get(current_disk['name'])
+                if not device:
+                    verrors.add(
+                        f'disks.{index}.identifier',
+                        f'Test cannot be performed for {disk["identifier"]}. Unable to retrieve disk details.'
+                    )
+
+        verrors.check()
+
+        return list(
+            await asyncio_map(functools.partial(self.__manual_test, devices), test_disks_list, 16)
+        )
+
+    async def __manual_test(self, devices, disk):
+        device = devices.get(disk['disk'])
+        args = await get_smartctl_args(disk['disk'], device)
+
+        proc = await run(
+            list(
+                filter(bool, ['smartctl', '-t', disk['type'].lower(), '-C' if disk['mode'] == 'FOREGROUND' else None])
+            ) + args,
+            check=False, encoding='utf8'
+        )
+
+        output = {}
+        if proc.returncode:
+            output['error'] = proc.stderr
+            self.middleware.logger.debug(
+                f'Self test for {disk["disk"]} failed with {proc.returncode} return code.'
+            )
+        else:
+            time_details = re.findall(RE_TIME_DETAILS, proc.stdout)
+            if not time_details:
+                output['error'] = f'Failed to parse smartctl self test details for {disk["identifier"]}.'
+            else:
+                output['expected_result_time'] = time_details[0].strip()
+                # TODO: Please setup alerts
+
+        return {
+            'disk': disk['disk'],
+            'identifier': disk['identifier'],
+            **output
+        }
 
     @filterable
     async def results(self, filters, options):


src/middlewared/middlewared/plugins/smart.py
```diff
@@ -296,6 +299,157 @@ class Config:
...
+            list(
+                filter(bool, ['smartctl', '-t', disk['type'].lower(), '-C' if disk['mode'] == 'FOREGROUND' else None])
+            ) + args,
+            check=False, encoding='utf8'
+        )
+
+        output = {}
+        if proc.returncode:
+            output['error'] = proc.stderr
+            syslog.openlog('smartctl', facility=syslog.LOG_DAEMON)
```
themylogin: Why do we use syslog here?

sonicaj: To log `smartctl` command status to `daemon.log` ?
```
root@freenas[~]# grep -rin "smartctl" /var/log/daemon.log 
168:May  9 14:14:27 freenas smartctl: Self test run for /dev/ada0 failed with 4 return code
```

themylogin: Why do we need it there?

sonicaj: We don't specifically need it. My thought was that if a user initiates this, he probably wouldn't be checking middlewared logs but daemon log to see what happened with it. Do you think we should perhaps move this to middlewared ? I can have that adjusted

src/middlewared/middlewared/plugins/smart.py
```diff
@@ -296,6 +299,162 @@ class Config:
...
+                syslog.LOG_WARNING,
+                f'Self test run for /dev/{disk["disk"]} failed with {proc.returncode} return code'
+            )
+            syslog.closelog()
+        else:
+            time_details = re.findall('test will complete after(.*)', proc.stdout, re.IGNORECASE)
+            if not time_details:
+                output['error'] = f'Failed to parse smartctl self test details for {disk["identifier"]}.'
+            else:
+                output['Expected Result Time'] = time_details[0].strip()
```
themylogin: Does it always produce the right time? Can't it say that it will be complete in 2 hours and finish in 5 minutes instead?

src/middlewared/middlewared/plugins/smart.py
themylogin: User has launched this via UI, why would he want to check system logs for result? Also this brings us to this: I can't find the code that gives us the list of currently running manual tests (and recently finished), is this considered non-necessary?

src/middlewared/middlewared/plugins/smart.py
sonicaj: I have been doing this for a while now and have tested this on 11.3/12, it doesn't change as far as i have seen. I can't find a reference which would explicitly say that indeed it doesn't. Please advise if you think there can be a fluctuation in the format ( It is dependent on the locale i think ).

themylogin: It's not the format that bothers me, it's the value, is it right to check only in 2 hours if it says that test will complete in 2 hours? What if it fails in 2 minutes instead?

src/middlewared/middlewared/plugins/smart.py
sonicaj: As far as i am aware, the same command `smartctl -l selftest disk` is good for this ? The `smartctl` docs reflect the same. Please let me know if there is something which i am overlooking ?

sonicaj: Also, i'll have the logging to `daemon.log` disabled, i think we can just keep that bit to ourselves in middlewared.log then

src/middlewared/middlewared/plugins/smart.py
```diff
@@ -296,6 +298,159 @@ class Config:
...
+                    Str('type', enum=['LONG', 'SHORT', 'CONVEYANCE', 'OFFLINE'], required=True),
+                )
+            ]
+        )
+    )
+    async def manual_test(self, disks):
+        """
+        Run manual SMART tests for `disks`.
+
+        `type` must be specified which indicates what type of SMART test should be run.
```
a-stjohn:
```suggestion
        `type` indicates what type of SMART test will be ran and must be specified.
```

src/middlewared/middlewared/plugins/smart.py
```diff
@@ -296,6 +298,159 @@ class Config:
...
+                filter(bool, ['smartctl', '-t', disk['type'].lower(), '-C' if disk['mode'] == 'FOREGROUND' else None])
+            ) + args,
+            check=False, encoding='utf8'
+        )
+
+        output = {}
+        if proc.returncode:
+            output['error'] = proc.stderr
+            self.middleware.logger.debug(
+                f'Self test run for /dev/{disk["disk"]} failed with {proc.returncode} return code'
```
a-stjohn:
```suggestion
                f'Self test for /dev/{disk["disk"]} failed with {proc.returncode} return code'
```

william-gr: You can also drop /dev/

src/middlewared/middlewared/plugins/alert.py
```diff
@@ -793,6 +793,49 @@ class Config:
...
+            Str('title', required=True),
+            Str('category', required=True, enum=list(AlertCategory.__members__)),
+            Str('level', required=True, enum=list(AlertLevel.__members__)),
+            Str('text'),
+            Str('node', default='A'),
+        )
+    )
+    async def send_alert(self, id, alert_data):
+        """
+        Send a manual alert using Alert Service `id`.
```
william-gr: Does that mean we can send alerts for unconfigured alert classes? Do we really want that?

themylogin: It was me who offered this solution. I think it's OK because user explicitly consents to receive this alert when he starts SMART test. On the other hand, if we list "SMART test finished successfully" as an alert class, user sets it's level to INFO (as it should be) and alert service level to WARNING, he will never receive this alert, which is undesirable.

src/middlewared/middlewared/plugins/smart.py
```diff
@@ -296,6 +298,159 @@ class Config:
...
+            output['error'] = proc.stderr
+            self.middleware.logger.debug(
+                f'Self test for {disk["disk"]} failed with {proc.returncode} return code.'
+            )
+        else:
+            time_details = re.findall('test will complete after(.*)', proc.stdout, re.IGNORECASE)
+            if not time_details:
+                output['error'] = f'Failed to parse smartctl self test details for {disk["identifier"]}.'
+            else:
+                output['Expected Result Time'] = time_details[0].strip()
```
themylogin: Why is this key named like that?

sonicaj: We can have it changed, there's no specific reason, just to let the user know what it holds

src/middlewared/middlewared/plugins/alert.py
william-gr: On the other hand you have to select an alert service while scheduling a manual test which kind of goes against how we do everything else, so why is this different?

While your example is undesirable it works as designed. Would probably be less confusing if we didnt have to choose which service to send to.

themylogin: > On the other hand you have to select an alert service while scheduling a manual test which kind of goes against how we do everything else, so why is this different?

If we make this UI widget collapsed dropdown with all options checked by default, it won't take much space and won't distract, nevertheless providing option to disable specific alert channels if user does not want to send "SMART test is done" alert there.

(This implies that we should change `def send_alert(self, id, alert_data):` to `def send_alert(self, ids, alert_data):` to provide ability to send to multiple sources)

99% of the users won't change any alert settings, leaving default service to "send e-mail to root when warning". With this settings they will just never be able to use this new feature.

Force button is essentially the same, just with less flexibility.

src/middlewared/middlewared/plugins/alert.py
```diff
@@ -793,6 +793,68 @@ class Config:
...
+                'SendAlertClass',
+                (AlertClass,), {
+                    'title': alert_data['title'],
+                    'category': AlertCategory[alert_data['category']],
+                    'level': AlertLevel[alert_data['level']],
+                    'text': alert_data['text']
+                }
+            ),
+            datetime=datetime.now(),
+            node=alert_data['node']
```
themylogin: With recent changes (not merged yet) you'll also need to add `uuid` here (random one)

src/middlewared/middlewared/plugins/alert.py
william-gr: That doesnt answer the question why this is so special.

We are going to do the same for every other thing now too? Run Cloud Sync, Run Replication, Run Cronjob, Run Rsync?

Lets discuss on Ryver. I will ping you both when I have time.

src/middlewared/middlewared/plugins/smart.py
themylogin: I still think that we need `underscore_case` name for this key

src/middlewared/middlewared/plugins/smart.py
```diff
@@ -296,6 +297,101 @@ class Config:
...
+    async def manual_test(self, disks):
+        """
+        Run manual SMART tests for `disks`.
+
+        `type` indicates what type of SMART test will be ran and must be specified.
+        """
+        verrors = ValidationErrors()
+        if not disks:
+            verrors.add(
+                'manual_test.disks',
```
william-gr: First part should be schema name, I dont see manual_test schema name?

src/middlewared/middlewared/utils/asyncio_.py
```diff
@@ -14,3 +15,13 @@
 
     futures = [func(arg) for arg in arguments]
     return await asyncio.gather(*futures)
+
+
+def call_later(delay, func, arguments):
```
william-gr: We are not using it anywhere so we probably can remove this until there is an actual need.

src/middlewared/middlewared/plugins/smart.py
```diff
@@ -296,6 +297,101 @@ class Config:
...
+        )
+
+        output = {}
+        if proc.returncode:
+            output['error'] = proc.stderr
+            self.middleware.logger.debug(
+                f'Self test for {disk["disk"]} failed with {proc.returncode} return code.'
+            )
+        else:
+            time_details = re.findall('test will complete after(.*)', proc.stdout, re.IGNORECASE)
```
william-gr: Lets compile the PR at module level please, something like RE_TIME_DETAILS = re.compile(r'test will complete after(.*)', re.IGNORECASE)

So its not compiled on every disk/usage.

src/middlewared/middlewared/plugins/smart.py
```diff
@@ -296,6 +297,101 @@ class Config:
...
+                        })
+                        break
+                else:
+                    verrors.add(
+                        'manual_test.disks.identifier',
+                        f'{disk["identifier"]} is not valid. Please provide a valid disk identifier.'
+                    )
+                    continue
+
+                if current_disk['name'] is None or current_disk['name'].startswith('nvd'):
```
william-gr: Whats the story about nvd?

sonicaj: I am not sure, this condition exists in `annotate_disk_smart_tests` and my usage for running smartctl tests is almost the same so i copied it over keeping the usage consistent. I was thinking maybe `camcontrol_list` function has something to do with it and is unable to retrieve disks starting with that ?
Anyways, the idea was it would be brought up in the PR and if it's out of scope, we can have it removed.
@themylogin can you please confirm if this check is requried ?

william-gr: Yeah well, there isnt a way to map nvd to nvme devices yet, so I believe thats why.

But lets make the error message more clear if thats really the case. NVMe users will be clueless.

sonicaj: Perfect, i'll have it updated

PR #2998, opened 2019-05-06
QE tkt-86793 Skip API jail testing if not FreeNAS

ericbsd:


pkgdemon:
FYI QA tickets are still in redmine

PR #3009, opened 2019-05-08
NAS-101542 / 11.3 / Fix issues with smb plugin and system dataset changes

anodos325:
Samba's python bindings proved to be somewhat unstable in long-running applications.
Switch back to using shell commands. Stop tracking 'disabled' status in the smbhash
because this only changes during password reset.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -22,6 +22,7 @@ LOGLEVEL_MAP = {
     '10': 'DEBUG',
 }
 RE_NETBIOSNAME = re.compile(r"^[a-zA-Z0-9\.\-_!@#\$%^&\(\)'\{\}~]{1,15}$")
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -299,19 +258,17 @@ def check_codec(encoding):
...
-                'sid_type': LSA_sidType(g.sid_name_use).name,
-            })
-        return groupmap_list
+    async def groupmap_list(self):
+        groupmap = []
+        out = await run([SMBCmd.NET.value, 'groupmap', 'list'], check=False)
+        if out.returncode != 0:
+            raise CallError(f'groupmap list failed with error {out.stderr.decode()}')
+        for line in (out.stdout.decode()).splitlines():
+            m = re.match(r'^(?P<ntgroup>.+) \((?P<SID>S-[0-9\-]+)\) -> (?P<unixgroup>.+)$', line)
```
william-gr: Just a minor nit but may be worth doing, we could make the regex a global var and compile it, so its not compiled every iteration and everytime this method is called

RE_NETGROUPMAP = re.compile(r'...')
or something, then RE_NTHASH.match(line)

PR #3010, opened 2019-05-08
NAS-101035 / 11.3 / Rework pool attachments

themylogin:
It's WIP, but can you please pre-review this now to ensure I am working in the right direction?


src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -145,6 +147,7 @@ class Config:
             default=[],
             items=[Str("provider", enum=["SYS", "KRB5", "KRB5I", "KRB5P"])],
         ),
+        Bool("enabled"),
```
william-gr: We prob want default=true here as well

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -448,5 +451,40 @@ def validate_hosts_and_networks(self, other_shares, data, schema_name, verrors,
...
+    async def get_attachment_name(self, attachment):
+        return ', '.join(attachment['paths'])
+
+    # NFS share can only contain paths from single dataset, that's why we can delete/disable entire share
+    # if even one path matches
+    async def delete(self, attachments):
+        for attachment in attachments:
+            await self.middleware.call(
+                'datastore.delete', 'sharing.nfs_share_path', [['share_id', '=', attachment['id']]]
+            )
```
william-gr: Do we really need to delete nfs_share_path first?
I always thought django ORM would delete it because its a FK with default action to delete related entries.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2950,6 +2907,38 @@ def add(attr):
...
+                    attachments["attachments"].append(await delegate.get_attachment_name(attachment))
+                if attachments["attachments"]:
+                    result.append(attachments)
+        return result
+
+    def __attachments_path(self, dataset):
+        if dataset['type'] == 'FILESYSTEM':
+            return dataset['mountpoint']
+
+        if dataset['type'] == 'VOLUME':
```
william-gr: If its a VOLUME its a zvol, I am confused why you're returning it as a pool?

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -951,5 +954,39 @@ def vfsobjects_choices(self):
         asyncio.ensure_future(middleware.call('service.reload', 'cifs'))
 
 
+class SMBFSAttachmentDelegate(FSAttachmentDelegate):
+    name = 'smb'
+    title = 'CIFS Share'
```
william-gr: SMB Share

CIFS is the legacy term we used

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1428,8 +1394,53 @@ def nic_attach_choices(self):
...
+            except Exception:
+                self.middleware.logger.warning('Unable to vm.stop %r', attachment['id'])
+
+    async def toggle(self, attachments, enabled):
+        for attachment in attachments:
+            action = 'vm.start' if enabled else 'vm.stop'
+            try:
+                await self.middleware.call(action, attachment['id'])
+            except Exception:
+                self.middleware.logger.warning('Unable to %s %r: %r', action, attachment['id'])
```
william-gr: 3 formats and 2 args

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -674,6 +676,7 @@ class Config:
         Bool('shadowcopy', default=False),
         Str('auxsmbconf'),
         Bool('default_permissions'),
+        Bool('enabled'),
```
william-gr: Defalt value here as well

gui/storage/forms.py
```diff
@@ -2397,11 +2397,11 @@ class VolumeExport(Form):
...
     def __init__(self, *args, **kwargs):
         self.instance = kwargs.pop('instance', None)
-        services = kwargs.pop('services', {})
+        attachments = kwargs.pop('attachments', [])
         super(VolumeExport, self).__init__(*args, **kwargs)
-        if list(services.keys()):
+        if attachments:
             self.fields['cascade'] = forms.BooleanField(
-                initial=True,
+                initial=False,
```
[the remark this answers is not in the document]
themylogin: Because it's more destructive now. Also new False is old True.

src/middlewared/middlewared/plugins/pool.py
themylogin: To make datasets and zvols share hierarchy.

This way, if I delete dataset `tank/ds1`, iSCSI extent `tank/ds1/ds2/zvol1` will also be deleted.

Otherwise I would have to check if we are deleting dataset or zvol in ISCSIFSAttachmentDelegate. And other delegates will be called with paths like `zvol/tank/ds1/ds2/zvol1` which does not have any sense.

gui/storage/forms.py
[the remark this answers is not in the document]
william-gr: More destructive how? We already did delete the shares before. Or do you mean because we also do for periodic snapshot task?

If thats the only case I think we are fine keeping the True or people will remain with the old shares.
Note that I see exporting a pool without destroying as an edge case, most people will destroy if they are at that point.

themylogin:
@william-gr proposed pool export/dataset delete flow is following:

When showing "Export pool" / "Delete dataset" dialog, run `pool.attachments` and `pool.processes` (or `pool.dataset.attachments` and `pool.dataset.processes`) and display results like this:

```
Currently the following services depend on this volume:
 * NFS Share:
    * /mnt/tank
 * Samba Share:
    * tank
Following running services are using this volume:
 * S3 [STOP]
 * WebDAV [STOP]
Following unknown processes are using this pool:
 * 12648 /usr/local/bin/zsh [TERMINATE]
```

At this step we should exclude service from the list of running services if it has any active attachments, because removing/disabling this attachment will (in theory) make this service stop using it.

User needs to click STOP (`service.stop(name)`) / TERMINATE (`service.terminate_process(pid)`) on all listed services and processes. If he does not, we show a warning: "Some services/processes are still using this dataset/pool, export/delete will probably fail, are you sure want to continue?" when user clicks "OK" in dialog.

If exporting pool / deleting dataset fails with code `EBUSY`, we run `pool.processes` again and, if there are any, show these services and processes again, allow to terminate them and retry delete operation..

themylogin:
This way we never implicitly kill anything, user has to do it himself, while we provide convenient way to do it.

william-gr:
I am not sure about giving the choice to stop services and terminate PIDs.

This is not a question "IF" it will fail, because it definitely will.

Catching EBUSY is not simple, it may happen after many filesystems have already been unmounted, entries deleted in database. What if user decides to give up at that point?

I am thinking we simply just go ahead and stop/terminate them with a big scary warning.

themylogin:
@william-gr imagine if we started exporting pool: we've disabled all shares, but samba service for some reason is still holding open files there, the only way to make it close them is apparently full service restart. What do we do now, we need to confirm this (service restart) with user, right? He is not expecting service interruption, otherwise what's the point for all this?

william-gr:
@themylogin in that case yes, but I see that as a different issue then what was proposed. We have already warned which services are going to be stopped/restarted and which process killed.

If exports fails after that user should be warned and not kill whatever else is still holding the pool.

themylogin:
@william-gr we have not warned user that samba will be restarted because we support removing shares and closing connections without restarting samba. However, under certain circumstances, restarting samba is the only way, that's why we need to warn user that, sadly, samba has to be restarted, so he can prepare for it (or delay dataset deletion).

themylogin:
The point in time to warn is between deleting/disabling attachments and doing everything else.

william-gr:
@themylogin I was under the impression that if we couldnt make close the connection and release the pool we would not go ahead and raise an error so the export screen could be updated to retry?

Or do you mean you want to warn the user in advance and just do it?

themylogin:
>  I was under the impression that if we couldnt make close the connection and release the pool we would not go ahead and raise an error so the export screen could be updated to retry?

Yes, that's what I was thinking too

themylogin:
@william-gr updated flow:

When showing "Export pool", run `pool.attachments` and `pool.processes` and display results like this:

```
Currently the following services depend on this pool:
 * NFS Share:
    * /mnt/tank
 * Samba Share:
    * tank
Following running services are using this pool:
 * S3
 * WebDAV
Following unknown processes are using this pool:
 * 12648 /usr/local/bin/zsh [TERMINATE]
WARNING: these processes will be terminated while exporting the pool
(we need to make clear that this warning only relates to unknown processes and services will continue functioning without interruption)
```

At this step we should exclude service from the list of running services if it has any active attachments, because removing/disabling this attachment will (in theory) make this service stop using it.

If exporting pool fails with code `EBUSY` and following `extra`:
```
{
    "code": "services_restart",
    "services": ["cifs"],
}
```

We show the following warning:
```
Warning: these services have to be restarted in order to export pool:
 * Samba
```

If user agrees, we run `pool.export` again, but now with `{"restart_services": true}`

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2959,6 +2932,139 @@ def add(attr):
...
+                             '-x', 'f',     # Cross filesystem boundaries
+                             '+D', path,    # Include specified directory and all its descendants
+                             stdout=subprocess.PIPE, stderr=subprocess.DEVNULL, check=False, encoding='utf8')
+            for pid, name in parse_lsof(lsof.stdout):
+                service = await self.middleware.call('service.identify_process', name)
+                if service:
+                    result.append({
+                        "pid": pid,
+                        "name": name,
+
```
william-gr: Why the extra line break?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2959,6 +2932,139 @@ def add(attr):
...
+                        cmdline = await self.middleware.run_in_thread(
+                            lambda: psutil.Process(pid).cmdline()
+                        )
+                    except psutil.NoSuchProcess:
+                        pass
+                    else:
+                        result.append({
+                            "pid": pid,
+                            "name": name,
+
```
william-gr: Why the extra line break?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2959,6 +2932,139 @@ def add(attr):
...
+          {
+            "pid": 97778,
+            "name": "minio",
+            "cmdline": "/usr/local/bin/minio -C /usr/local/etc/minio server --address=0.0.0.0:9000 --quiet /mnt/tank/wk"
+          }
+        ]
+        """
+        result = []
+        dataset = await self._get_instance(oid)
+        path = self.__attachments_path(dataset)
```
william-gr: Will this work with a zvol? Being used by iscsi

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2959,6 +2932,139 @@ def add(attr):
...
+        if path:
+            lsof = await run('lsof',
```
william-gr: I am not sure how practical this is. I have tried this in a pool with 100 datasets, lsof takes over 2 minutes to run and uses a lot of CPU.
Can we investigate some way to make it faster and/or alternatives?

src/middlewared/middlewared/plugins/pool.py
themylogin: Yes, that's why I prepend `/mnt` to zvol path

william-gr: I am confused, then. How can you verify if a block device is in use if you are checking with /mnt path?

gui/templates/storage/dataset_confirm_delete.html
```diff
@@ -10,6 +10,23 @@
         {% plural %}
         This will delete all files within the dataset and child datasets.
         {% endblocktrans %}
+        {% if attachments %}
+            <br />
+            <br />
+            The following resources depending on this dataset will also be deleted:
```
Mrt134:
```suggestion
            These dataset-dependent resources will also be deleted:
```

gui/templates/storage/volume_detach.html
```diff
@@ -22,10 +22,22 @@
         </td>
     </tr>
     {% endif %}
-    {% if services %}
+    {% if attachments %}
     <tr>
         <td colspan="2">
-        <p>{% trans "ATTENTION - Currently the following services depend on this volume: " %}<strong>{{ services.keys|join:", " }}</strong></p>
+            {% trans "ATTENTION - Currently the following services depend on this volume: " %}
```
Mrt134:
```suggestion
            {% trans "WARNING: The following services depend on this volume and could be disrupted when the volume is detached: " %}
```
Could "this volume" be reworked to show the name of the volume?

PR #3013, opened 2019-05-09
NAS-101633 / 11.3 / Set default ACL on dataset creation and fix misc issues with 'apply d…

anodos325:
…efault permissions'

- Add more intelligent 'default' permissions
- Remove default 'everyone@' entry in most situations
- Apply ACL on dataset creation for "windows" permissions type (prep for GUI ACL editor).

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2643,12 +2643,14 @@ class PoolDatasetService(CRUDService):
                 'notifier.change_dataset_share_type', data['name'], data.get('share_type', 'UNIX').lower()
             )
             if data.get('share_type', 'UNIX') == 'WINDOWS':
-                dataset = await self.middleware.call('zfs.dataset.query', [('name', '=', data['name'])])
+                dataset = await self.middleware.call('zfs.dataset.query', [('id', '=', data['id'])])
                 setacl_job = await self.middleware.call('filesystem.setacl', dataset[0]['mountpoint'], [
                     {"tag": "owner@", "id": None, "type": "ALLOW", "perms": {"BASIC": "FULL_CONTROL"}, "flags": {"BASIC": "INHERIT"}},
                     {"tag": "group@", "id": None, "type": "ALLOW", "perms": {"BASIC": "FULL_CONTROL"}, "flags": {"BASIC": "INHERIT"}}
                 ])
                 await setacl_job.wait()
+                if setacl_job.error:
+                    raise CallError(setacl_job.error)
 
         return await self._get_instance(data['id'])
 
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2642,6 +2642,13 @@ def transform(dataset):
             await self.middleware.call(
                 'notifier.change_dataset_share_type', data['name'], data.get('share_type', 'UNIX').lower()
             )
+            if data.get('share_type', 'UNIX') == 'WINDOWS':
+                dataset = await self.middleware.call('zfs.dataset.query', [('name', '=', data['name'])])
```
william-gr: Lets filter using `id`, we have a performance improvement for that one.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2642,6 +2642,13 @@ def transform(dataset):
...
+                setacl_job = await self.middleware.call('filesystem.setacl', dataset[0]['mountpoint'], [
+                    {"tag": "owner@", "id": None, "type": "ALLOW", "perms": {"BASIC": "FULL_CONTROL"}, "flags": {"BASIC": "INHERIT"}},
+                    {"tag": "group@", "id": None, "type": "ALLOW", "perms": {"BASIC": "FULL_CONTROL"}, "flags": {"BASIC": "INHERIT"}}
+                ])
+                await setacl_job.wait()
```
william-gr: We should probably either log or raise if setacl fails?
if setacl_job.error:
    # something

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -809,17 +809,103 @@ class Config:
...
+                    "tag": "everyone@",
+                    "id": None,
+                    "type": "ALLOW",
+                    "perms": {"BASIC": "TRAVERSE"},
+                    "flags": {"BASIC": "NOINHERIT"}
+                },
+            ])
+
+        job = await self.middleware.call('filesystem.setacl', path, acl, {'recursive': True, 'traverse': True})
+        await job.wait()
```
william-gr: Not sure we want to wait here since apply_default_perms is not a job and it can timeout? Perhaps just let it do its thing in background?

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -809,17 +810,104 @@ class Config:
...
+                    "id": None,
+                    "type": "ALLOW",
+                    "perms": {"BASIC": "TRAVERSE"},
+                    "flags": {"BASIC": "NOINHERIT"}
+                },
+            ])
+
+        job = await self.middleware.call('filesystem.setacl', path, acl, {'recursive': True, 'traverse': True})
+        if job.error:
+            raise CallError(job.error)
```
william-gr: This is pretty much a no-op if we are not waiting, so this should be removed

PR #3024, opened 2019-05-10
NAS-101725 / 11.2 / Reboot passive storage controller after system dataset changes

anodos325:


gui/system/forms.py
```diff
@@ -1705,9 +1705,15 @@ def __init__(self, *args, **kwargs):
...
-            "systemDatasetMigration();"
-        )
+        with client as c:
+            if c.call('system.is_freenas'):
+                self.fields['sys_pool'].widget.attrs['onChange'] = (
+                    "systemDatasetMigration();"
+                )
+            else:
+                self.fields['sys_pool'].widget.attrs['onChange'] = (
+                    "systemDatasetMigration_TN();"
```
william-gr: You probably also want to check for failover licensed so it doesnt show for single nodes?

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -112,6 +112,14 @@ class Config:
         if config['rrd'] != new['rrd']:
             await self.rrd_toggle()
             await self.middleware.call('service.restart', 'collectd')
+
+        if not await self.middleware.call('system.is_freenas') and await self.middleware.call('failover.licensed'):
+            if await self.middleware.call('notifier.failover_status') == 'MASTER':
+                try:
+                    await self.middleware.call('failover.call_remote', 'system.reboot')
+                except Exception as e:
+                    self.logger.debug('Failed to reboot passive storage controller after system dataset change: %s' % e)
```
william-gr: `self.logger.debug('Failed to reboot passive storage controller after system dataset change: %s', e)`

For logging they are arguments, we shouldnt replace the string before it goes to logging, although we do it wrong in many places.

william-gr:
backport

PR #3029, opened 2019-05-12
NAS-101562 / 11.3 / Register upgraded disks capacity

sonicaj:
This PR fixes a bug where when a user upgraded his boot pool with new disks to increase it's size, the pool didn't register the new size.
Ticket: #101562

### DIFF
--- a/src/middlewared/middlewared/plugins/boot.py
+++ b/src/middlewared/middlewared/plugins/boot.py
@@ -202,6 +202,10 @@ class BootService(Service):
 
         await job.wrap(extend_pool_job)
 
+        # If the user is upgrading his disks, let's set expand to True to make sure that we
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/boot.py
```diff
@@ -202,6 +202,10 @@ def get_swap_size(self, disk):
 
         await job.wrap(extend_pool_job)
 
+        # If the user is upgrading his disks, let's set expand to True to make sure that we
+        # register the new disks capacity which increase the size of the pool
+        await self.middleware.call('zfs.pool.online', 'freenas-boot', f'{dev}p2', True)
```
william-gr: Just make sure this doesnt break when no online -e is needed.

PR #3034, opened 2019-05-12
User alert.uuid everywhere

themylogin:


themylogin:
@william-gr it's `str(uuid.uuid4())`, we reuse existing ID when we get alert on the same node with the same class and key

PR #3042, opened 2019-05-13
NAS-100735 / 11.2 / Fix HDDs not entering standby

themylogin:


themylogin:
@william-gr user confirms that this was the issue (polling for temperature was preventing drives from sleep). Do you think we need to change API to allow fetching disk temperature from `smartctl` command directly instead of collectd? This way UI will be able to show disk temperature on user's demand.

william-gr:
@themylogin It could be a future improvement but we dont need it now, the current behavior is fine for short/mid term.

themylogin:
backport

PR #3045, opened 2019-05-14
NAS-100114 / 12 / OpenVPN as a Service

sonicaj:
This PR aims to introduces the following changes:

1) Ability to use FN/TN as an OpenVPN Server.
2) Ability to use FN/TN as an OpenVPN Client.
3) Ability to revoke certificates which have been compromised.
4) Manage extensions for certificates.

Ticket: NAS-100114

### DIFF
--- a/src/freenas/etc/ix.rc.d/ix-etc
+++ b/src/freenas/etc/ix.rc.d/ix-etc
@@ -31,7 +31,8 @@ generate_etc()
 {
 	copy_templates
 
-	[ -s /data/dhparam.pem ] || openssl dhparam -rand /dev/random 2048 > /data/dhparam.pem
+	dhparam_path="$(LD_LIBRARY_PATH=/usr/local/lib /usr/local/bin/midclt call certificate.dhparam 2> /dev/null)"
+	[ -s "$dhparam_path" ] || openssl dhparam -rand /dev/random 2048 > "$dhparam_path"
 
 	LD_LIBRARY_PATH=/usr/local/lib /usr/local/bin/midclt call etc.generate_all > /dev/null
 
--- a/src/middlewared/middlewared/alert/source/certificates.py
+++ b/src/middlewared/middlewared/alert/source/certificates.py
@@ -42,6 +42,56 @@ class CertificateIsExpiringAlertSource(AlertSource):
         return alerts
 
 
+class CertificateRevokedAlertClass(AlertClass):
+    category = AlertCategory.CERTIFICATES
+    level = AlertLevel.CRITICAL
+    title = 'Certificate Revoked'
+    text = '%(service)s %(type)s has been revoked. Please replace the certificate immediately.'
+
+
+class CertificateRevokedAlertSource(AlertSource):
+    async def check(self):
+        alerts = []
+
+        for cert_id, service, type_c, datastore in (
+            ((await self.middleware.call('ftp.config'))['ssltls_certificate'], 'FTP', 'certificate', 'certificate'),
+            ((await self.middleware.call('s3.config'))['certificate'], 'S3', 'certificate', 'certificate'),
+            ((await self.middleware.call('webdav.config'))['certssl'], 'Webdav', 'certificate', 'certificate'),
+            (
+                (await self.middleware.call('openvpn.server.config'))['server_certificate'],
+                'OpenVPN server', 'certificate', 'certificate'
+            ),
+            (
+                (await self.middleware.call('openvpn.client.config'))['client_certificate'],
+                'OpenVPN client', 'certificate', 'certificate'
+            ),
+            (
+                (await self.middleware.call('system.general.config'))['ui_certificate']['id'],
+                'Web UI', 'certificate', 'certificate'
+            ),
+            (
+                (await self.middleware.call('openvpn.server.config'))['root_ca'],
+                'OpenVPN server', 'root certificate authority', 'certificateauthority'
+            ),
+            (
+                (await self.middleware.call('openvpn.client.config'))['root_ca'],
+                'OpenVPN client', 'root certificate authority', 'certificateauthority'
+            )
+        ):
+            if (
+                cert_id and (
+                    await self.middleware.call(
+                        f'{datastore}.query', [
+                            ['id', '=', cert_id]
+                        ], {'get': True}
+                    )
+                )['revoked']
+            ):
+                alerts.append(Alert(CertificateRevokedAlertClass, {'service': service, 'type': type_c}))
+
+        return alerts
+
+
 class CertificateParsingFailedAlertClass(AlertClass):
     category = AlertCategory.CERTIFICATES
     level = AlertLevel.WARNING
--- a/src/middlewared/middlewared/etc_files/local/openvpn/client/openvpn_client.conf
+++ b/src/middlewared/middlewared/etc_files/local/openvpn/client/openvpn_client.conf
@@ -1,0 +1,38 @@
+<%
+	try:
+		middleware.call_sync('openvpn.client.config_valid')
+	except Exception as e:
+		raise FileShouldNotExist()
+
+	config = middleware.call_sync('openvpn.client.config')
+	root_ca = middleware.call_sync('certificateauthority.query', [['id', '=', config['root_ca']]], {'get': True})
+	client_cert = middleware.call_sync('certificate.query', [['id', '=', config['client_certificate']]], {'get': True})
+%>\
+client
+dev ${config['device_type'].lower()}
+#dev-type ${config['device_type'].lower()} -FIXME: This does not work, it is an openvpn issue in FreeBSD
+proto ${config['protocol'].lower()}
+port ${config['port']}
+remote ${config['remote']}
+user nobody
+group nobody
+persist-key
+persist-tun
+ca ${root_ca['certificate_path']}
+cert ${client_cert['certificate_path']}
+key ${client_cert['privatekey_path']}
+verb 3
+remote-cert-tls server
+% if config['compression']:
+compress ${config['compression'].lower()}
+% endif
+% if config['authentication_algorithm']:
+auth ${config['authentication_algorithm']}
+% endif
+% if config['cipher']:
+cipher ${config['cipher']}
+% endif
+% if config['nobind']:
+nobind
+% endif
+${config['additional_parameters']}
--- a/src/middlewared/middlewared/etc_files/local/openvpn/client/openvpn_client.conf
+++ b/src/middlewared/middlewared/etc_files/local/openvpn/client/openvpn_client.conf
@@ -35,4 +35,9 @@ cipher ${config['cipher']}
 % if config['nobind']:
 nobind
 % endif
+% if config['tls_crypt_auth_enabled']:
+<tls-crypt>
+${config['tls_crypt_auth']}
+</tls-crypt>
+% endif
 ${config['additional_parameters']}
--- a/src/middlewared/middlewared/etc_files/rc.conf.py
+++ b/src/middlewared/middlewared/etc_files/rc.conf.py
@@ -127,6 +127,7 @@ def services_config(middleware, context):
         'lldp': ['ladvd'],
         's3': ['minio'],
         'nfs': ['nfs_server', 'rpc_lockd', 'rpc_statd', 'mountd', 'nfsd', 'rpcbind'],
+        'openvpn_server': ['openvpn_server'],
         'rsync': ['rsyncd'],
         'snmp': ['snmpd', 'snmp_agent'],
         'ssh': ['openssh'],
@@ -287,6 +288,11 @@ def nut_config(middleware, context):
     yield 'nut_upsmon_enable="YES"'
 
 
+def openvpn_server_config(middleware, context):
+    yield 'openvpn_server_configfile="/usr/local/etc/openvpn/server/openvpn_server.conf"'
+    yield 'openvpn_server_dir="/usr/local/etc/openvpn/server"'
+
+
 def powerd_config(middleware, context):
     value = 'YES' if middleware.call_sync('system.advanced.config')['powerdaemon'] else 'NO'
     yield f'powerd_enable="{value}"'
@@ -435,6 +441,7 @@ def render(service, middleware):
         nfs_config,
         nis_config,
         nut_config,
+        openvpn_server_config,
         powerd_config,
         s3_config,
         smart_config,
--- a/src/middlewared/middlewared/etc_files/rc.conf.py
+++ b/src/middlewared/middlewared/etc_files/rc.conf.py
@@ -127,6 +127,7 @@ def services_config(middleware, context):
         'lldp': ['ladvd'],
         's3': ['minio'],
         'nfs': ['nfs_server', 'rpc_lockd', 'rpc_statd', 'mountd', 'nfsd', 'rpcbind'],
+        'openvpn_client': ['openvpn_client'],
         'openvpn_server': ['openvpn_server'],
         'rsync': ['rsyncd'],
         'snmp': ['snmpd', 'snmp_agent'],
@@ -293,6 +294,11 @@ def openvpn_server_config(middleware, context):
     yield 'openvpn_server_dir="/usr/local/etc/openvpn/server"'
 
 
+def openvpn_client_config(middleware, context):
+    yield 'openvpn_client_configfile="/usr/local/etc/openvpn/client/openvpn_client.conf"'
+    yield 'openvpn_client_dir="/usr/local/etc/openvpn/client"'
+
+
 def powerd_config(middleware, context):
     value = 'YES' if middleware.call_sync('system.advanced.config')['powerdaemon'] else 'NO'
     yield f'powerd_enable="{value}"'
@@ -441,6 +447,7 @@ def render(service, middleware):
         nfs_config,
         nis_config,
         nut_config,
+        openvpn_client_config,
         openvpn_server_config,
         powerd_config,
         s3_config,
--- a/src/middlewared/middlewared/plugins/crypto.py
+++ b/src/middlewared/middlewared/plugins/crypto.py
@@ -382,6 +382,7 @@ class CryptoKeyService(Service):
             'san': [],
             'email': obj.get_subject().emailAddress,
             'DN': '',
+            'extensions': {}
         }
 
         for ext in (
@@ -392,6 +393,10 @@ class CryptoKeyService(Service):
         ):
             if 'subjectAltName' == ext.get_short_name().decode():
                 cert_info['san'] = [s.strip() for s in ext.__str__().split(',') if s]
+
+            cert_info['extensions'][
+                re.sub(r"^(\S)", lambda m: m.group(1).upper(), ext.get_short_name().decode())
+            ] = ext.__str__()
 
         dn = []
         for k, v in obj.get_subject().get_components():
@@ -1111,7 +1116,8 @@ class CertificateService(CRUDService):
             cert.update({
                 key: None for key in [
                     'digest_algorithm', 'lifetime', 'country', 'state', 'city', 'from', 'until',
-                    'organization', 'organizational_unit', 'email', 'common', 'san', 'serial', 'fingerprint'
+                    'organization', 'organizational_unit', 'email', 'common', 'san', 'serial',
+                    'fingerprint', 'extensions'
                 ]
             })
 
--- a/src/middlewared/middlewared/plugins/crypto.py
+++ b/src/middlewared/middlewared/plugins/crypto.py
@@ -813,6 +813,70 @@ class CertificateService(CRUDService):
         datastore_extend = 'certificate.cert_extend'
         datastore_prefix = 'cert_'
 
+    PROFILES = {
+        'openvpn_server_certificate': {
+            'cert_extensions': {
+                'BasicConstraints': {
+                    'enabled': True,
+                    'ca': False,
+                    'extension_critical': True
+                },
+                'AuthorityKeyIdentifier': {
+                    'enabled': True,
+                    'authority_cert_issuer': True,
+                    'extension_critical': False
+                },
+                'ExtendedKeyUsage': {
+                    'enabled': True,
+                    'extension_critical': True,
+                    'usages': [
+                        'SERVER_AUTH',
+                    ]
+                },
+                'KeyUsage': {
+                    'enabled': True,
+                    'extension_critical': True,
+                    'digital_signature': True,
+                    'key_encipherment': True
+                }
+            },
+            'key_length': 2048,
+            'key_type': 'RSA',
+            'lifetime': 3650,
+            'digest_algorithm': 'SHA256'
+        },
+        'openvpn_client_certificate': {
+            'cert_extensions': {
+                'BasicConstraints': {
+                    'enabled': True,
+                    'ca': False,
+                    'extension_critical': True
+                },
+                'AuthorityKeyIdentifier': {
+                    'enabled': True,
+                    'authority_cert_issuer': True,
+                    'extension_critical': False
+                },
+                'ExtendedKeyUsage': {
+                    'enabled': True,
+                    'extension_critical': True,
+                    'usages': [
+                        'CLIENT_AUTH',
+                    ]
+                },
+                'KeyUsage': {
+                    'enabled': True,
+                    'extension_critical': True,
+                    'digital_signature': True
+                }
+            },
+            'key_length': 2048,
+            'key_type': 'RSA',
+            'lifetime': 3650,
+            'digest_algorithm': 'SHA256'
+        }
+    }
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.map_functions = {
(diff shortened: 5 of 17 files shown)


src/middlewared/middlewared/etc_files/rc.conf.py
```diff
@@ -287,6 +289,16 @@ def nut_config(middleware, context):
     yield 'nut_upsmon_enable="YES"'
 
 
+def openvpn_server_config(middleware, context):
+    yield 'openvpn_server_configfile="/etc/local/openvpn/server/openvpn_server.conf"'
```
william-gr: /etc/local is a freenas thing, we should probably use /usr/local/etc/ to make it clear the real location it means.

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -164,9 +170,129 @@ class CryptoKeyService(Service):
...
+            # create a certificate/ca as most wouldn't even want to know what extension is or does.
+            # Apart from this we also add subjectKeyIdentifier automatically
+            supported = [
+                'BasicConstraints', 'AuthorityKeyIdentifier',
+                'ExtendedKeyUsage', 'KeyUsage'
+            ]
+
+            for attr in filter(
+                lambda attr: attr in supported, dir(x509.extensions)
+            ):
```
william-gr: This seems a bit backwards to me, why loop through all x509 extensions if we already know the extensions we support and could just do `hasattr(attr, x509.extensions) for attr in supported`?

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -669,6 +983,10 @@ def __init__(self, *args, **kwargs):
             'CERTIFICATE_CREATE_ACME': self.__create_acme_certificate,
         }
 
+    @accepts()
+    async def profiles(self):
```
william-gr: Documentation?

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -1086,6 +1416,21 @@ def renew_certs(self, job):
...
+
+    @accepts()
+    async def extended_key_usage_choices(self):
+        return {
+            k: k for k in
+            dir(x509.oid.ExtendedKeyUsageOID) if not k.startswith('__')
+        }
+
+    @accepts()
+    async def dhparam(self):
```
william-gr: Should be private?

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -1086,6 +1416,21 @@ def renew_certs(self, job):
         """
         return {k: k for k in ['RSA', 'EC']}
 
+    @accepts()
+    async def certificate_extensions(self):
```
william-gr: Docs for these two?

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -1450,7 +1804,8 @@ def __create_acme_certificate(self, job, data):
         cert_info.update({
             'ca_privatekey': signing_cert['privatekey'],
             'ca_certificate': signing_cert['certificate'],
-            'serial': cert_serial
+            'serial': cert_serial,
+            'cert_extensions': data['cert_extensions']
```
william-gr: We pop cert_extensions for some certificates, are we sure cert_extensions key will always be available in this point?

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -1638,14 +2054,93 @@ def __init__(self, *args, **kwargs):
...
+        certs.append(ca)
+        return certs
+
     @private
     async def validate_common_attributes(self, data, schema_name):
         verrors = ValidationErrors()
 
         await _validate_common_attributes(self.middleware, data, verrors, schema_name)
 
+        if not data['cert_extensions']['BasicConstraints']['enabled']:
```
william-gr: How can you be sure these cert_extensions have these keys?

cert_extensions does not seem to be marked as required, neither each of its keys.

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -2018,10 +2529,11 @@ def set_enum(attr):
         Int('id', required=True),
         Dict(
             'ca_update',
+            Bool('revoked'),
             Int('ca_id'),
             Int('csr_cert_id'),
             Str('create_type', enum=['CA_SIGN_CSR']),
-            Str('name'),
+            Str('name')
```
william-gr: Nitpick, in our codebase we prefer to end things with a comma, that way adding a new entry does not require 2 line changes.

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -2105,7 +2628,20 @@ def set_enum(attr):
                 ]
             }
         """
-        ca = await self._get_instance(id)
```
william-gr: Removing this does not guarantee the `id` exists.

src/middlewared/middlewared/plugins/vpn.py
```diff
@@ -0,0 +1,532 @@
...
+                    'Please ensure provided client certificate exists in Root CA chain '
+                    'and has necessary extensions set.'
+                )
+
+        client_config = [
+            'client',
+            f'dev {config["device_type"].lower()}',
+            f'proto {config["protocol"].lower()}',
+            f'port {config["port"]}',
+            'remote "PLEASE FILL OUT SERVER DOMAIN/IP HERE"',
```
william-gr: That could be an optional argument to the method

src/middlewared/middlewared/plugins/vpn.py
```diff
@@ -0,0 +1,532 @@
...
+            'openvpn_server_update',
+            Bool('tls_crypt_auth_enabled'),
+            Int('netmask', validators=[Range(min=0, max=128)]),
+            Int('server_certificate'),
+            Int('port', validators=[Port()]),
+            Int('root_ca'),
+            IPAddr('server'),
+            Str('additional_parameters'),
+            Str('authentication_algorithm', enum=OpenVPN.digests(), null=True),
+            Str('cipher', null=True, enum=OpenVPN.ciphers()),
```
william-gr: These both seem to have tons of options.
It seems to make more sense to have `cipher_choices` and `authentication_algorithm_choices` instead.

src/middlewared/middlewared/schema.py
```diff
@@ -437,9 +437,9 @@ def clean(self, value):
                     try:
                         value[index] = i.clean(v)
                         found = True
+                        break
```
william-gr: What are you trying to achieve here? Allow different types in the list?
What if none of them match?

src/middlewared/middlewared/plugins/crypto.py
sonicaj: Yes for internal certificates they will always be available.

src/middlewared/middlewared/plugins/crypto.py
sonicaj: But we do set their defaults so even if they aren't marked as required, we have the default dictionary with specified defaults

src/middlewared/middlewared/schema.py
sonicaj: Then we raise an exception. I initially did this because i was thinking of handling extensions differently which included passing a list of dictionaries which obviously failed validation because right now we only support one type of item in list schema. I kept this around if you think we should allow this behaviour. I can have it removed and if none of them match, we raise an exception

src/middlewared/middlewared/plugins/vpn.py
sonicaj: To be sure i understand correctly, are you referring to changing `OpenVPN.ciphers` and `OpenVPN.digests` names to the ones you mentioned ?

william-gr: No, remove the enums and have the choices methods in each service.

src/middlewared/middlewared/etc_files/local/openvpn/client/openvpn_client.conf
```diff
@@ -0,0 +1,43 @@
...
+	except Exception as e:
+		raise FileShouldNotExist()
+
+	config = middleware.call_sync('openvpn.client.config')
+	root_ca = middleware.call_sync('certificateauthority.query', [['id', '=', config['root_ca']]], {'get': True})
+	client_cert = middleware.call_sync('certificate.query', [['id', '=', config['client_certificate']]], {'get': True})
+%>\
+client
+dev ${config['device_type'].lower()}
+#dev-type ${config['device_type'].lower()} -FIXME: This does not work, is it a openvpn freebsd bug ? Please check
```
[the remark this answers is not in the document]
sonicaj: Yes i am already doing it. The benefit of this working is that we can have a tun/tap interface named `openvpn_server`/`openvpn_client` instead of `tun0` etc. But openvpn is not liking it right now, i am investigating what is triggering that behaviour

src/middlewared/middlewared/plugins/crypto.py
william-gr: How about API calls, where user sents a payload with only cert_extensions without the needed keys?

src/middlewared/middlewared/schema.py
william-gr: Ok, understood. No I think that is fine. We should probably add these cases to the unittests though, as we are changing the behavior

src/middlewared/middlewared/plugins/vpn.py
```diff
@@ -0,0 +1,532 @@
...
+
+        await self._update_service(old_config, config)
+
+        return await self.config()
+
+
+def setup(middleware):
+    for srv in ('openvpn_client', 'openvpn_server'):
+        if not os.path.exists(f'/etc/local/rc.d/{srv}'):
+            os.symlink('/etc/local/rc.d/openvpn', f'/etc/local/rc.d/{srv}')
```
themylogin: Same note about `/usr/local/etc`

src/freenas/etc/ix.rc.d/ix-etc
```diff
@@ -31,7 +31,8 @@ generate_etc()
 {
 	copy_templates
 
-	[ -s /data/dhparam.pem ] || openssl dhparam -rand /dev/random 2048 > /data/dhparam.pem
+	dhparam_path="$(LD_LIBRARY_PATH=/usr/local/lib /usr/local/bin/midclt call certificate.dhparam 2> /dev/null || echo "/data/dhparam.pem")"
```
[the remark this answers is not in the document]
sonicaj: It's just a fall back if middleware call fails for some reason ( we obviously have bigger problems if middleware fails to respond at this point ). If you think this does not bring us much value, i'll have it reverted to just retrieving the name from middleware

src/middlewared/middlewared/plugins/etc.py
```diff
@@ -266,6 +272,9 @@ def __init__(self, *args, **kwargs):
             if rendered is None:
                 continue
 
+            if not os.path.exists(outfile.rsplit('/', 1)[0]):
```
william-gr: I just now saw this, I think we can use os.path.dirname instead, which is an "official" way to do what you did

src/middlewared/middlewared/plugins/vpn.py
```diff
@@ -0,0 +1,582 @@
+import ipaddr
```
william-gr: Same thing about ipaddr, we should use ipaddress

src/middlewared/middlewared/plugins/etc.py
```diff
@@ -266,6 +272,9 @@ def __init__(self, *args, **kwargs):
             if rendered is None:
                 continue
 
+            if not os.path.exists(os.path.dirname(outfile)):
+                os.makedirs(outfile.rsplit('/', 1)[0])
```
william-gr: You replaced only one one them, you're still doing the same thing on makedirs
Should save dirname and reuse it ;)

sonicaj: Uh oh, thank you for catching that :)

src/freenas/etc/ix.rc.d/ix-etc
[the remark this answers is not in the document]
themylogin: Yes, I think that's really strange thing to have fallback for.

src/middlewared/middlewared/alert/source/certificates.py
```diff
@@ -42,6 +42,56 @@ class CertificateIsExpiringAlertSource(AlertSource):
         return alerts
 
 
+class CertificateRevokedAlertClass(AlertClass):
+    category = AlertCategory.CERTIFICATES
+    level = AlertLevel.CRITICAL
+    title = 'Certificate Has Been Revoked'
```
Mrt134:
```suggestion
    title = 'Certificate Revoked'
```

src/middlewared/middlewared/alert/source/certificates.py
```diff
@@ -42,6 +42,56 @@ class CertificateIsExpiringAlertSource(AlertSource):
...
+    text = '%(service)s %(type)s has been revoked.'
```
Mrt134:
```suggestion
    text = '%(service)s %(type)s has been revoked.'
```
What should the user do to fix the issue? Edit the revoked certificate, create a new one, etc ?

src/middlewared/middlewared/alert/source/update.py
```diff
@@ -33,46 +26,27 @@ def check_sync(self):
...
             return Alert(HasUpdateAlertClass)
 
 
 class UpdateNotAppliedAlertClass(AlertClass):
     category = AlertCategory.SYSTEM
     level = AlertLevel.WARNING
-    title = "Update Not Applied"
-    text = "%s"
+    title = "Update Applied Pending Reboot"
+    text = "Update has been applied but is pending a reboot."
```
Mrt134:
```suggestion
    text = "Update is installed but cannot be applied until the system is rebooted."
```

src/middlewared/middlewared/alert/source/update.py
```diff
@@ -33,46 +26,27 @@ def check_sync(self):
...
+        if self.middleware.call_sync("update.get_pending"):
             return Alert(HasUpdateAlertClass)
 
 
 class UpdateNotAppliedAlertClass(AlertClass):
     category = AlertCategory.SYSTEM
     level = AlertLevel.WARNING
-    title = "Update Not Applied"
-    text = "%s"
+    title = "Update Applied Pending Reboot"
```
Mrt134:
```suggestion
    title = "Update Requires Reboot"
```

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -325,9 +349,130 @@ def token(self, app, token):
...
+        return totp.verify(token, valid_window=config['window'])
+
+    @accepts()
+    def renew_secret(self):
+        """
+        Generates a new secret for Two Factor Authentication. Returns boolean true on success.
+        """
+        config = self.middleware.call_sync(f'{self._config.namespace}.config')
+        if not config['enabled']:
+            raise CallError('Please enable Two Factor Authentication first.')
```
Mrt134:
```suggestion
            raise CallError('Please enable Two-Factor Authentication first.')
```
Would it be useful to the user to point them to where they can enable two-factor auth?

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -164,9 +170,127 @@ class CryptoKeyService(Service):
...
+            lambda v: v[1]['enabled'] and v[0] not in skip_extension,
+            extensions_data.items()
+        ):
+            klass = getattr(x509.extensions, extension[0])
+            try:
+                klass(*self.get_extension_params(extension))
+            except Exception as e:
+                verrors.add(
+                    f'{schema}.{extension[0]}',
+                    f'Please provide valid values for {extension[0]}: {e}'
```
Mrt134:
```suggestion
                    f'Please provide valid values for {extension[0]}: {e}'
```
This could use an example so the user will have a reference to what values are acceptable.

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -1477,7 +1837,10 @@ def __create_acme_certificate(self, job, data):
         """
         Update certificate of `id`
 
-        Only name attribute can be updated
+        Only name and revoked attribute can be updated.
+
+        When `revoked` is enabled, the specified cert `id` is revoked and if it belongs to a CA chain which
+        exists on this system, it's serial number is added to the CA's certificate revocation list.
```
Mrt134:
```suggestion
        exists on this system, its serial number is added to the CA's certificate revocation list.
```

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -1638,14 +2061,96 @@ def __init__(self, *args, **kwargs):
...
     @private
     async def validate_common_attributes(self, data, schema_name):
         verrors = ValidationErrors()
 
         await _validate_common_attributes(self.middleware, data, verrors, schema_name)
 
+        if not data['cert_extensions']['BasicConstraints']['enabled']:
+            verrors.add(
+                f'{schema_name}.cert_extensions.BasicConstraints.enabled',
+                'For Certificate Authority, this should be enabled.'
```
Mrt134: Is this error a suggestion to improve performance or is it notifying the user the CA doesn't work without this being enabled?

Mrt134:
```suggestion
                'This must be enabled for a Certificate Authority'
```
Suggesting reversing the sentence order to remove the pause.

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -1638,14 +2061,96 @@ def __init__(self, *args, **kwargs):
...
+            )
+        elif not data['cert_extensions']['BasicConstraints']['ca']:
+            verrors.add(
+                f'{schema_name}.cert_extensions.BasicConstraints.ca',
+                'For Certificate Authority, "ca" must be enabled.'
```
Mrt134:
```suggestion
                '"ca" must be enabled for a Certificate Authority.'
```

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -1638,14 +2061,96 @@ def __init__(self, *args, **kwargs):
...
+            )
+
+        if not data['cert_extensions']['KeyUsage']['enabled']:
+            verrors.add(
+                f'{schema_name}.cert_extensions.KeyUsage.enabled',
+                'For Certificate Authority, this should be enabled.'
```
Mrt134:
```suggestion
                'This must be enabled for a Certificate Authority.'
```

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -1638,14 +2061,96 @@ def __init__(self, *args, **kwargs):
...
+            )
+        elif not data['cert_extensions']['KeyUsage']['key_cert_sign']:
+            verrors.add(
+                f'{schema_name}.cert_extensions.KeyUsage.key_cert_sign',
+                'For Certificate Authority, "key_cert_sign" must be enabled.'
```
Mrt134:
```suggestion
                '"key_cert_sign" must be enabled for a Certificate Authority.'
```

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -2028,7 +2554,10 @@ def set_enum(attr):
         """
         Update Certificate Authority of `id`
 
-        Only name attribute can be updated
+        Only `name` and `revoked` attribute can be updated.
+
+        If `revoked` is enabled, the CA and it's complete chain is marked as revoked and added to the CA's
```
Mrt134:
```suggestion
        If `revoked` is enabled, the CA and its complete chain is marked as revoked and added to the CA's
```

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -2064,10 +2593,18 @@ def set_enum(attr):
...
+            if new['name'] != old['name']:
+                await validate_cert_name(
+                    self.middleware, new['name'], self._config.datastore,
+                    verrors, 'certificate_authority_update.name'
+                )
+
+            if old['revoked'] != new['revoked'] and new['revoked'] and not new['privatekey']:
+                verrors.add(
+                    'certificate_authority_update.revoked',
+                    'Only CA\'s with privatekey can be marked as revoked.'
```
Mrt134:
```suggestion
                    'Only Certificate Authorities with a privatekey can be marked as revoked.'
```

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -2106,6 +2646,20 @@ def set_enum(attr):
...
+
+        # Let's make sure we don't delete a ca which is being used by any service in the system
+        for service_cert_id, text in [
+            ((await self.middleware.call('openvpn.server.config'))['root_ca'], 'OpenVPN Server'),
+            ((await self.middleware.call('openvpn.client.config'))['root_ca'], 'OpenVPN Client')
+        ]:
+            if service_cert_id == id:
+                verrors.add(
+                    'certificateauthority_delete.id',
+                    f'Selected certificate authority is being used by {text} service, please select another one.'
```
Mrt134:
```suggestion
                    f'This certificate authority is in use by {text} service and cannot be deleted.'
```

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -325,9 +349,130 @@ def token(self, app, token):
...
+                Bool('ssh', default=False)
+            ),
+            update=True
+        )
+    )
+    async def do_update(self, data):
+        """
+        `otp_digits` represents number of allowed digits in the OTP.
+
+        `window` extends the validity to `window` many counter ticks before and after the current one.
```
a-stjohn: I'm confused about this sentence. Is the second "\`window\`" a verb in this case?

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -325,9 +349,130 @@ def token(self, app, token):
...
+
+        `interval` is time duration in seconds specifying OTP expiration time from it's creation time.
```
a-stjohn:
```suggestion
        `interval` is the time duration in seconds that specifies OTP expiration time from the creation time.
```

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -325,9 +349,130 @@ def token(self, app, token):
...
+
+        if config['services']['ssh']:
+            self.middleware.call_sync('service.reload', 'ssh')
+
+        return True
+
+    @accepts()
+    async def provisioning_uri(self):
+        """
+        Returns the provisioning URI for the OTP. This can then be encoded in a QR Code and used to
```
a-stjohn:
```suggestion
        Returns the provisioning URI for the OTP. This can be encoded in a QR Code and used to
```

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -325,9 +349,130 @@ def token(self, app, token):
...
+
+    @private
+    def generate_base32_secret(self):
+        return pyotp.random_base32()
+
+
 async def check_permission(middleware, app):
     """
-    Authenticates connections comming from loopback and from
+    Authenticates connections coming from loopback and from
```
a-stjohn:
```suggestion
    Authenticate connections coming from loopback and
```

src/middlewared/middlewared/alert/source/certificates.py
sonicaj: This shows that the certificate is in use by one of the system's services and it should be replaced as soon as possible as it's been compromised and revoked.

src/middlewared/middlewared/plugins/crypto.py
sonicaj: It is notifying the user CA doesn't work without this being enabled.

src/middlewared/middlewared/plugins/crypto.py
sonicaj: @william-gr thoughts on this one ? Reason this has been done this way @Mrt134 is that we can in future support more certificate extensions and it doesn't seem right to validate all of them on our own and especially when it is dynamic for some of them ( values differing based on other values provided to the same extension ), so if we handle each extension validation ourselves apart from schema, that would be a lot of unnecessary code to maintain in my opinion. That however would be the only way to make the error message here more elaborate wrt valid values for each extension.

src/freenas/etc/ix.rc.d/ix-etc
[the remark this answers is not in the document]
sonicaj: @themylogin updated.

src/middlewared/middlewared/plugins/crypto.py
william-gr: If I understand correctly this will hardly be ever seen by the UI users, this is advanced usage of the API, people using these manually should now before hand what they are supposed to be.
IMO its fine to leave it like that.

src/middlewared/middlewared/alert/source/certificates.py
Mrt134: Updated suggestion to communicate the next step to the user.

PR #3048, opened 2019-05-14
NAS-101762 / 11.2 / Move cachetool to middleware job

anodos325:
Cachetool fill can very long-running and cause middleware timeouts. Change this to a job.

### DIFF
--- a/src/middlewared/middlewared/plugins/notifier.py
+++ b/src/middlewared/middlewared/plugins/notifier.py
@@ -1,4 +1,4 @@
-from middlewared.service import CallError, Service
+from middlewared.service import CallError, Service, job
 
 import errno
 import os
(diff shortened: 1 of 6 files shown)


src/middlewared/middlewared/plugins/notifier.py
```diff
@@ -241,11 +241,12 @@ def ds_idmap_type_code_to_string(self, code):
...
-            '/usr/local/bin/python /usr/local/www/freenasUI/tools/cachetool.py expire && '
-            '/usr/local/bin/python /usr/local/www/freenasUI/tools/cachetool.py fill',
-            stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, shell=True
-        )
+        cachetool_job = await self.middleware.call('notifier.cachetool', 'expire')
+        cachetool_job.wait()
+        if cachetool_job.error:
+            raise CallError(f'cachetool expire failed with error {cachetool_job.error}')
+
+        self.middleware.call('notifier.cachetool', 'fill')
```
william-gr: I think we need an `await` here to actually schedule the coroutine to run

PR #3049, opened 2019-05-14
NAS-101763 / 11.3 / filesystem.getacl should return advanced flags and permset instead of…

anodos325:
… 'OTHER'


src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -364,6 +364,11 @@ def getacl(self, path, simplified=True):
                 }
                 if ace['tag'] == 'everyone@' and ace['perms']['BASIC'] == 'NOPERMS':
                     continue
+
+                for key in ['perms', 'flags']:
+                    if ace[key]['BASIC'] == 'OTHER':
```
skarekrow: Where is the `key` key?

PR #3054, opened 2019-05-15
NAS-101778 / 11.2 / Allow getting disktemp for non-fatal smartctl exit statuses

tejp:
Fixes a problem where Freenas reports disk temperature "N/A" for one of my disks. The disk in question has smart errors which it got from a failed sata port but is otherwise fully functioning. smartctl reports an exit status of 64 for the disk and therefore the script doesn't fetch the disk temperature.

https://www.smartmontools.org/browser/trunk/smartmontools/smartctl.8.in#lbAH

### DIFF
--- a/src/freenas/usr/local/lib/collectd_pyplugins/disktemp.py
+++ b/src/freenas/usr/local/lib/collectd_pyplugins/disktemp.py
@@ -183,7 +183,7 @@ class DiskTemp(object):
             except Exception:
                 pass
         cp = subprocess.run(['/usr/local/sbin/smartctl', '-a', '-n', 'standby', f'/dev/{disk}'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-        if (cp.returncode & 7) != 0:
+        if (cp.returncode & 0b11) != 0:
             collectd.info(f'Failed to run smartctl for {disk}: {cp.stdout.decode("utf8", "ignore")}')
             return None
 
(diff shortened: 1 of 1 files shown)


src/freenas/usr/local/lib/collectd_pyplugins/disktemp.py
```diff
@@ -183,7 +183,7 @@ def get_temperature(self, disk):
             except Exception:
                 pass
         cp = subprocess.run(['/usr/local/sbin/smartctl', '-a', '-n', 'standby', f'/dev/{disk}'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-        if cp.returncode != 0:
+        if (cp.returncode & 7) != 0:
```
william-gr: Hi, thanks for your PR but it doesnt seem quite right.

If smartctl is returning 64 it means it is the bit 6

```
>>> 0b01000000
64
```

Looking at the exit return code bitmap it seems we should take into account only bits 0-3 so that would read
`if (cp.returncode & 0b1111) != 0:`

Thoughts @themylogin

themylogin: @william-gr these
```
Bit 2:
Some SMART or other ATA command to the disk failed, or there was a checksum error in a SMART data structure (see '-b' option above).
Bit 3:
SMART status check returned "DISK FAILING".
```
Also seem OK to me for case of reporting temperature. Probably we can do
`if (cp.returncode & 0b11) != 0:`

tejp: Thank you, I wasn't sure of what the appropriate bitmask would be, shall I change it to `0b11`?

william-gr: Yes, please

william-gr:
Thanks! We will merge it once the 11.2 branch is unfrozen, probably sometime next week.

william-gr:
backport

PR #3056, opened 2019-05-15
NAS-101750 / 11.2 / Properly reraise Client init exception

themylogin:


themylogin:
@william-gr probably these minor fixes are everything I can offer at this moment, was not able to repeat this weird behavior. It's not possible to somehow "eject" plugin, it's not possible to start middleware without TN overlay, worker processes signals are handled correctly.

william-gr:
Thats fine, is a good improvement nevertheless

PR #3057, opened 2019-05-15
NAS-101789 / 11.2 / Disable netdata on passive controller

themylogin:


src/freenas/etc/rc.conf.local
```diff
@@ -637,7 +637,13 @@ _gen_conf()
 	_bool_svcconfig ladvd_enable lldp
 	_bool_svcconfig ctld_enable iscsitarget
 	_bool_svcconfig apache24_enable webdav
-	_bool_svcconfig netdata_enable netdata
+
+	if ! is_freenas && [ "${failover_status}" = "BACKUP" ]; then
```
william-gr: During boot failover_status is pretty much useless since CARP is not initalized until boot completes.

Lets do how we did below with smartd and start netdata in the carp hook script.

PR #3058, opened 2019-05-16
NAS-101799 / 11.3 / feat(middlewared/network): return number of seconds remaining

william-gr:
Ticket:	NAS-101799

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -18,7 +18,6 @@ import shlex
 import signal
 import socket
 import subprocess
-import time
 import urllib.request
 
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -607,8 +608,12 @@ def iface_extend(self, iface_state, configs, is_freenas, fake=False):
         """
         Returns wether or not we are waiting user to checkin the applied network changes
         before they are rolled back.
+        Value is in number of seconds or null.
         """
-        return self._rollback_timer is not None
+        if self._rollback_timer:
+            remaining = self._rollback_timer.when() - time.monotonic()
```
themylogin: We should use `loop.time()` here:

```
when()
Return a scheduled callback time as float seconds.

The time is an absolute timestamp, using the same time reference as loop.time().
```

william-gr: Which is just a wrapper for time.monotonic ?
Or you just want to be safe for future changes to another event loop?

themylogin: Why depend on implementation if we can avoid it easily?

I did not know that `loop.time()` is a wrapper for `time.monotonic()`, my first thought was "it this a bug?"

PR #3061, opened 2019-05-16
NAS-101785 / 11.3 / Provide API for querying users and groups when directory services enabled

anodos325:
Make all directory services plugins provide consistent API for getting / setting user and group caches. Each directory service cache is a dictionary with two keys 'users' and 'groups', containing lists of basic user / group info.

Basic user: {'pw_name': <username>, 'pw_uid': <uid>, 'local', <boolean>}
Basic group: {'gr_name': <groupname>, 'gr_gid': <gid>, 'local', <boolean>}

'local' is provided so that UI team can easily differentiate between local and remote users/groups.


src/middlewared/middlewared/utils/__init__.py
```diff
@@ -322,6 +322,9 @@ def filterop(f):
     if options.get('get') is True:
         return rv[0]
 
+    if options.get('limit'):
+        return rv[:options['limit']]
```
william-gr: We should probably implement the same thing in datastore.query, to keep when equal.

src/middlewared/middlewared/plugins/account.py
```diff
@@ -88,6 +88,33 @@ class Config:
                 pass
         return user
 
+    @accepts(Ref('query-filters'), Ref('query-options'), Bool('dssearch', default=False))
```
william-gr: To not change .query behavior we would need to somehow use the same parameters, only filters and options, using `@filterable` here instead of accepts, meaning we would have to drop dssearch.

src/middlewared/middlewared/plugins/account.py
```diff
@@ -88,6 +88,33 @@ class Config:
...
+        options = options or {}
+        options['extend'] = self._config.datastore_extend
+        options['extend_context'] = self._config.datastore_extend_context
+        options['prefix'] = self._config.datastore_prefix
+
+        datastore_options = options.copy()
+        datastore_options.pop('count', None)
+        datastore_options.pop('get', None)
+
+        if dssearch:
```
william-gr: The above being said, I am not entirely sure how this block can be accomplished.

Perhaps we can look at the filters for `local = False` or `local != None` ?

PR #3062, opened 2019-05-16
NAS-101809 / 11.3 / fix(jail/list_resource): Return a list of dicts for BRANCHES

skarekrow:
NAS-101809


skarekrow:
@sonicaj and myself discussed this, and I looped in all relevant parties. The BRANCH call returns a dict, while all others return a list. This is to avoid POLA, and make it function like other parts of the middleware where this coexistence can happen.

PR #3071, opened 2019-05-19
NAS-101675 / 11.2 / Generate an alert when creating a new ticket for TrueNAS proactive al…

themylogin:
…ert fails

### DIFF
--- a/src/middlewared/middlewared/plugins/alert.py
+++ b/src/middlewared/middlewared/plugins/alert.py
@@ -299,11 +299,10 @@ class AlertService(Service):
                                 except Exception:
                                     self.logger.error(f"Failed to create a support ticket", exc_info=True)
                                     await self.middleware.call("mail.send", {
-                                        "cc": ["auto-support@ixsystems.com"],
                                         "subject": "Failed to notify iXsystems about alert (%s)" % serial,
                                         "text": f"""
                                             Notifying iXsystems about an alert on system {serial} failed.
-                                            Please contact iXsystems Support.
+                                            Please contact iXsystems Support: https://www.ixsystems.com/support/
 
                                             Alert:
 
--- a/src/middlewared/middlewared/plugins/alert.py
+++ b/src/middlewared/middlewared/plugins/alert.py
@@ -300,12 +300,12 @@ class AlertService(Service):
                                     self.logger.error(f"Failed to create a support ticket", exc_info=True)
                                     await self.middleware.call("mail.send", {
                                         "cc": ["auto-support@ixsystems.com"],
-                                        "subject": "Failed creating automatic alert (%s)" % serial,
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/alert.py
```diff
@@ -296,6 +298,18 @@ def restore(self, id):
                                     })
                                 except Exception:
                                     self.logger.error(f"Failed to create a support ticket", exc_info=True)
+                                    await self.middleware.call("mail.send", {
+                                        "cc": ["auto-support@ixsystems.com"],
```
themylogin: @william-gr which e-mail we should use here?

src/middlewared/middlewared/plugins/alert.py
```diff
@@ -296,6 +298,18 @@ def restore(self, id):
...
+                                        "subject": "Failed creating automatic alert (%s)" % serial,
```
Mrt134:
```suggestion
                                        "subject": "Failed to notify iXsystems about alert (%s)" % serial,
```

src/middlewared/middlewared/plugins/alert.py
```diff
@@ -296,6 +298,18 @@ def restore(self, id):
...
+                                        "text": f"""
+                                            Creating automatic alert for iXsystems for your system {serial} failed
```
Mrt134:
```suggestion
                                            Notifying iXsystems about an alert on system {serial} failed.
```

src/middlewared/middlewared/plugins/alert.py
```diff
@@ -296,6 +298,18 @@ def restore(self, id):
...
+                                            for some reason. Please, contact us.
```
Mrt134:
```suggestion
                                            Please contact iXsystems Support.
```

src/middlewared/middlewared/plugins/alert.py
```diff
@@ -296,6 +298,18 @@ def restore(self, id):
...
+
+                                            Below is the text of automatic alert:
```
Mrt134:
```suggestion
                                            Alert:
```

src/middlewared/middlewared/plugins/alert.py
Mrt134: Maybe add email for iX support here too?

src/middlewared/middlewared/plugins/alert.py
william-gr: Waiting on feedback from support people, will get back to you when I have an answer

yocalebo: @themylogin please remove the cc line from the `except Exception` block. We will let the customer manually open a ticket in this scenario.

PR #3079, opened 2019-05-21
fix(middlewared/etc): truncate ctl.conf

william-gr:


src/middlewared/middlewared/etc_files/ctld.py
```diff
@@ -384,12 +384,12 @@ def main(middleware):
         addline('}\n\n')
 
     # Write out the CTL config file
-    with open(os.open(ctl_config, os.O_CREAT | os.O_WRONLY, 0o600), 'w') as fh:
+    with open(os.open(ctl_config, os.O_CREAT | os.O_WRONLY | os.TRUNC, 0o600), 'w') as fh:
```
themylogin: How is it different from `open(ctl_config, 'w')` now?

william-gr: The mode, 0o600, which was the entire point to being with

PR #3085, opened 2019-05-22
NAS-101863 / 11.2 / Set up NFSv4 properly

themylogin:


src/middlewared/middlewared/etc_files/nfsd.py
```diff
@@ -88,6 +90,24 @@ def build_share_targets(share):
     except Exception:
         pass
 
+    # Workaround to work with "onetime", since the rc scripts depend on rc flags.
```
william-gr: Not sure what I feel about mixing config file generation with calling service start/stop.

Also NFS is picky about services order and how its done at boot (which we will be doing with this change).

I imagine its now broken because we do reload and these are not called? How about make it a common method in service.py and also do it on reload?

themylogin: > I imagine its now broken because we do reload and these are not called?

We call `etc.generate nfsd` on reload, so this is called

themylogin: > Not sure what I feel about mixing config file generation with calling service start/stop.

We already do this in other files, that's why we have a design proposal to fix at least some of this in 12

william-gr: Yes, i know this is called on reload but it werent before, we did a full restart. So why not just call these on service reload instead too? I worry about regressions on startup.

Also, which other etc plugins we are calling service?

themylogin: > I worry about regressions on startup.

The ticket actually fixes this regression. I am struggling to find what was calling `service.start nfs` on startup when it worked..

william-gr: service.start is not called on startup at all. Its plain and simple rc.d with rc.conf.local

themylogin: So what was setting this sysctl then?

william-gr: rc.conf.local

william-gr: Or it was supposed to, and that may be the real issue. Its where other sysctls are supposed to be set in 11.2

william-gr: Taking a deeper look: /etc/rc.d/nfsd is supposed to set it on boot

yocalebo: Changing this particular parameter doesn't actually require a reload or restart of the service. Just FYI

william-gr: We are not talking about full services reload/restart, speaking on terms of our shim in the service layer

themylogin: Yes, and `vfs.nfsd.enable_stringtouid` is set by `rc.conf`, so the only problem left is that neither `/etc/rc.d/nfsd` nor `rc.conf` are not called on `service.reload nfs`, which is fixed now.

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -71,6 +74,27 @@ def nfs_compress(self, nfs):
...
+        config = self.middleware.call_sync("nfs.config")
+
+        if config["v4"]:
+            sysctl.filter("vfs.nfsd.server_max_nfsvers")[0].value = 4
+            if config["v4_v3owner"]:
+                # Per RFC7530, sending NFSv3 style UID/GIDs across the wire is now allowed
+                # You must have both of these sysctl"s set to allow the desired functionality
+                sysctl.filter("vfs.nfsd.enable_stringtouid")[0].value = 1
+                sysctl.filter("vfs.nfs.enable_uidtostring")[0].value = 1
+                subprocess.run(["service", "nfsuserd", "forcestop"])
```
william-gr: Isnt this going to print to stdout/stderr? Do we want it?

themylogin:
@william-gr I have not seen any issues, I just did not understand how does this work.

PR #3091, opened 2019-05-23
feat(usage): Add usage collector plugin

skarekrow:
This will collect anonymous data that can be aggregated.

NAS-101420

### DIFF
--- a/src/middlewared/middlewared/plugins/usage.py
+++ b/src/middlewared/middlewared/plugins/usage.py
@@ -1,22 +1,29 @@
 from middlewared.schema import accepts
-from middlewared.service import Service, private
+from middlewared.service import Service
 from datetime import datetime
 
 import json
 import asyncio
 import random
-import requests
+import aiohttp
 
 
 class UsageService(Service):
-    @private
+    class Config:
+        private = True
+
     async def start(self):
         try:
             gather = await self.gather()
-            requests.post(url='https://usage.freenas.org/submit', data=gather)
-        except Exception:
+            async with aiohttp.ClientSession(raise_for_status=True) as session:
+                await session.post(
+                    'https://usage.freenas.org/submit',
+                    data=gather,
+                    headers={"Content-type": "application/json"}
+                )
+        except Exception as e:
             # We still want to schedule the next call
-            pass
+            self.logger.error(e)
 
         event_loop = asyncio.get_event_loop()
         now = datetime.utcnow()
@@ -32,7 +39,6 @@ class UsageService(Service):
 
         return True
 
-    @accepts()
     async def gather(self):
         network = await self.middleware.call('interfaces.query')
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/usage.py
```diff
@@ -0,0 +1,416 @@
...
+                if not d.get('path'):
+                    vdevs += 1
+                    type = d['type']
+                    disks += len(d['children'])
+                else:
+                    disks += 1
+                    used = pd['usedbysnapshots']['parsed'] + \
+                        pd['usedbydataset']['parsed'] + \
+                        pd['usedbychildren']['parsed'] + \
+                        pd['usedbyrefreservation']['parsed']
```
william-gr: I think I was not clear here,

this is the same as just `pd['usedbydataset']['parsed']`

Thought we could perhaps also send other usedby* to usage stats.

william-gr:
@themylogin would love to know your opinion on all of this too ; )

src/middlewared/middlewared/plugins/usage.py
```diff
@@ -0,0 +1,416 @@
...
+import random
+import requests
+
+
+class UsageService(Service):
+    @private
+    async def start(self):
+        try:
+            gather = await self.gather()
+            requests.post(url='https://usage.freenas.org/submit', data=gather)
```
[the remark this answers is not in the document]
themylogin: Also need a timeout here, and `headers={"Content-type": "application/json"}`

src/middlewared/middlewared/plugins/usage.py
```diff
@@ -0,0 +1,416 @@
...
+import asyncio
+import random
+import requests
+
+
+class UsageService(Service):
+    @private
+    async def start(self):
+        try:
+            gather = await self.gather()
```
themylogin: Probably,. if `system.general` -> `crash_reporting` is false, we need to skip gathering and sending data. @william-gr what do you think?

src/middlewared/middlewared/plugins/usage.py
```diff
@@ -0,0 +1,416 @@
...
+            requests.post(url='https://usage.freenas.org/submit', data=gather)
+        except Exception:
+            # We still want to schedule the next call
```
themylogin: Log the exception please

src/middlewared/middlewared/plugins/usage.py
william-gr: Yeah, we need a way to opt-out for this but I am not sure reusing crash_reporting is our best option. Will need to think about it.

themylogin: So we are going to ship 11.3 with this always enabled?

william-gr: No, we are still long ways to 11.3. We will figure something out until then.

PR #3095, opened 2019-05-24
NAS-101887 / 11.2 / Add nut_upsshut regardless of ups service status

sonicaj:
This commit adds changes to add nut_upsshut to rc.conf even if ups service is not enabled to start at boot to suppress nut warnings when the ups service is started.


sonicaj:
We donot want this change in 11.2, closing the ticket

PR #3102, opened 2019-05-26
NAS-101525 / 11.3 / Improve fstab validation

sonicaj:
This commit introduces following changes:
1) Removes 'edit' option from fstab api because 'edit' action in iocage refers to an interactive session.
2) Make sure right params exist for add/replace/remove actions.
3) Always set up reasonable defaults for iocage api which fixes a bug where replace failed as we forced src/destination to be empty strings.

Ticket: #101525

### DIFF
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -682,7 +682,7 @@ class JailService(CRUDService):
         Dict(
             "options",
             Str(
-                "action", enum=["ADD", "EDIT", "REMOVE", "REPLACE", "LIST"],
+                "action", enum=["ADD", "REMOVE", "REPLACE", "LIST"],
                 required=True
             ),
             Str("source"),
@@ -706,17 +706,30 @@ class JailService(CRUDService):
 
         verrors = ValidationErrors()
 
-        source = options.get('source')
-        if source:
-            if not os.path.exists(source):
+        if action in ('add', 'replace', 'remove'):
+            if action != 'remove' or index is None:
+                # For remove we allow removing by index or mount, so if index is not specified
+                # we should validate that rest of the fields exist.
+                for f in ('source', 'destination', 'fstype', 'fsoptions', 'dump', 'pass'):
+                    if not options.get(f):
+                        verrors.add(
+                            f'options.{f}',
+                            f'This field is required with "{action}" action.'
+                        )
+
+            if action == 'replace' and index is None:
                 verrors.add(
-                    'options.source',
-                    'Provided path for source does not exist'
+                    'options.index',
+                    'Index cannot be "None" when replacing an fstab entry.'
                 )
-        elif (not source and index is None) and action != 'list':
+
+        verrors.check()
+
+        source = options.get('source')
+        if action in ('add', 'replace') and not os.path.exists(source):
             verrors.add(
                 'options.source',
-                'Provide a source path'
+                'The provided path for the source does not exist.'
             )
 
         destination = options.get('destination')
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/jail.py
```diff
@@ -706,17 +706,30 @@ def fstab(self, jail, options):
...
+                            f'options.{f}',
+                            f'This field is required with "{action}" action.'
+                        )
+
+            if action == 'replace' and index is None:
                 verrors.add(
-                    'options.source',
-                    'Provided path for source does not exist'
+                    'options.index',
+                    'Index must not be None when replacing fstab entry.'
```
Mrt134:
```suggestion
                    'Index cannot be "None" when replacing an fstab entry.'
```

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -706,17 +706,30 @@ def fstab(self, jail, options):
...
-        elif (not source and index is None) and action != 'list':
+
+        verrors.check()
+
+        source = options.get('source')
+        if action in ('add', 'replace') and not os.path.exists(source):
             verrors.add(
                 'options.source',
-                'Provide a source path'
+                'Provided path for source does not exist.'
```
Mrt134:
```suggestion
                'The provided path for the source does not exist.'
```

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -732,45 +745,24 @@ def fstab(self, jail, options):
                     verrors.add(
                         'options.destination',
                         'Destination is not a directory, please provide a'
```
Mrt134:
```suggestion
                        'Destination is not a directory. Please provide a'
```

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -732,45 +745,24 @@ def fstab(self, jail, options):
...
-                        ' valid destination'
+                        ' valid destination.'
```
Mrt134:
```suggestion
                        ' empty directory for the destination.'
```

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -732,45 +745,24 @@ def fstab(self, jail, options):
...
                     )
                 elif os.listdir(destination):
                     verrors.add(
                         'options.destination',
-                        'Destination directory should be empty'
+                        'Destination directory should be empty.'
```
Mrt134:
```suggestion
                        'Destination directory should be empty.'
```
Is this a suggestion or requirement? If it is a suggestion, why? Are there system problems to avoid or performance improvements to gain by choosing an empty directory?

src/middlewared/middlewared/plugins/jail.py
wblock: If the user is ever going to see this message, it has to tell them what we mean by "index". Out of context, I don't see if index is a string or a line number, but either way, this needs to be clearer.

src/middlewared/middlewared/plugins/jail.py
sonicaj: @Mrt134 this is considered a requirement from our end as the operation of mounting on non empty directories can cause issues with the existing data they have if the mount succeeds.

Mrt134: Ok, thanks for the clarification. Updated the suggestion.

src/middlewared/middlewared/plugins/jail.py
sonicaj: @wblock @Mrt134 `index` here refers to an integer which points to the line no in fstab. Please advise how should we proceed with this. Thank you

PR #3125, opened 2019-05-31
NAS-101944 / 11.2 / New VMware snapshot flags

themylogin:


yocalebo:
> To add to what caleb said, we should prob not be adding fields, at least not for 11.2.
> 
> Unless @yocalebo can think of any reason a customer would want to have quiesce=False?

No, I don't see a reason why the customer would want `quiesce=False` because it won't guarantee that the VM can be restored successfully.

william-gr:
I would say just drop the fields that and wait until someone comes up with a good reason to change those, for the sake of simplicity.

themylogin:
@william-gr @yocalebo quiesce does not always work. Actually the guy in the ticket has changed False to True, and some snapshots started working, while others broke completely. I’ll remove memory option, but I think quiesce should stay (maybe true by default)

PR #3127, opened 2019-05-31
NAS-101724 / 11.3 / Move license logic to middleware

william-gr:


src/middlewared/middlewared/plugins/system.py
```diff
@@ -277,15 +272,38 @@ def version(self):
         return "BOOTING"
 
     async def __get_license(self):
-        licenseobj = get_license()[0]
-        if not licenseobj:
+        if not os.path.exists(LICENSE_FILE):
+            return
+
+        with open(LICENSE_FILE, 'r') as f:
```
sonicaj: William do we want to run this in a coroutine ?

william-gr: Good point, its probably safer to turn this into a method

PR #3133, opened 2019-06-03
NAS-101261 / 11.3 / fix(middlewared/vm): add 1400x1050 to possible choices

william-gr:
Ticket:	NAS-101261


william-gr:
@skarekrow its not so non-standard, it seems to be somewhat common from googling.
Also this was already made availalbe in new UI.

PR #3134, opened 2019-06-03
NAS-101877 / 11.3 / feat(middlewared/update): implement `update.set_train`

william-gr:
Ticket:	NAS-101877

### DIFF
--- a/src/middlewared/middlewared/plugins/update.py
+++ b/src/middlewared/middlewared/plugins/update.py
@@ -268,6 +268,58 @@ class UpdateService(Service):
             'selected': selected,
         }
 
+    @accepts(Str('train', empty=False))
+    def set_train(self, train):
+        """
+        Set an update train to be used by default in updates.
+        """
+        return self.__set_train(train)
+
+    def __set_train(self, train, trains=None):
+        """
+        Wrapper so we dont call get_trains twice on update method.
+        """
+        if trains is None:
+            trains = self.get_trains()
+        if train != trains['selected']:
+            if train not in trains['trains']:
+                raise CallError('Invalid train name.', errno.ENOENT)
+
+            try:
+                result = compare_trains(trains['current'], train)
+            except Exception:
+                self.logger.warning(
+                    "Failed to compare trains %r and %r", trains['current'], train, exc_info=True
+                )
+            else:
+                errors = {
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/update.py
```diff
@@ -268,6 +268,23 @@ def get_trains(self):
             'selected': selected,
         }
 
+    @accepts(Str('train'))
+    def set_train(self, train):
+        """
```
themylogin: Don't we want any check here? Like in `update`

william-gr: We probably do, good thought

themylogin: What is a point of `call("update.set_train", null)`?

themylogin:
What is a point of `call("update.set_train", null)`?

william-gr:
@themylogin return get_trains, otherwise we would need to make two calls of these, which depend on network

themylogin:
@william-gr have not thought of that. But things like these tend to complicate over time, and next time we'll need to pack more actions into single call will lead to even more strange code. I don't insist on doing anything, but in cases like this I would prefer to have simple and clean public API and hide ugliness in a wrapper method:
```
@accepts(Str("train", required=True, empty=False))
def set_train(self, train):
    ...

@private
def handle_train_change(self, new_train):
    if new_train:
        self.set_train()
    return self.get_trains()
```
Also is number of calls (furthermore, non-repeated) an important thing in TN HA? I've thought we have some kind of fiber optic between them, not 3G :)

themylogin:
If I (as e.g. a new UI developer) will start calling `set_train(null)` by mistake and it won't raise any error and just silently don't do anything, I will spend more time trying to investigate what's goind on.

william-gr:
@themylogin that doesnt quite work, set_train needs to call get_trains, and update needs to check for the selected train get_trains returns.

As far as number of calls, this has nothing to do with TN HA, its network call, external, to the internet.

william-gr:
But yes, I get your point, I will rework on that.

src/middlewared/middlewared/plugins/update.py
```diff
@@ -268,6 +268,58 @@ def get_trains(self):
             'selected': selected,
         }
 
+    @accepts(Str('train'))
```
themylogin: `empty=False`?

PR #3138, opened 2019-06-04
NAS-100512 / 11.2 / fix(jail/update): Update plugins properly in 11.2

skarekrow:
Before they would simply return.

NAS-100512

### DIFF
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -357,10 +357,14 @@ class JailService(CRUDService):
             raise CallError(e, errno=errno.ENOENT)
 
     @private
-    def check_jail_existence(self, jail, skip=True):
+    def check_jail_existence(self, jail, skip=True, callback=None):
         """Wrapper for iocage's API, as a few commands aren't ported to it"""
         try:
-            iocage = ioc.IOCage(skip_jails=skip, jail=jail)
+            if callback is not None:
+                iocage = ioc.IOCage(callback=callback,
+                                    skip_jails=skip, jail=jail)
+            else:
+                iocage = ioc.IOCage(skip_jails=skip, jail=jail)
             jail, path = iocage.__check_jail_existence__()
         except (SystemExit, RuntimeError):
             raise CallError(f"jail '{jail}' not found!")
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/jail.py
```diff
@@ -357,10 +357,14 @@ def check_dataset_existence(self):
             raise CallError(e, errno=errno.ENOENT)
 
     @private
-    def check_jail_existence(self, jail, skip=True):
+    def check_jail_existence(self, jail, skip=True, callback=None):
         """Wrapper for iocage's API, as a few commands aren't ported to it"""
         try:
-            iocage = ioc.IOCage(skip_jails=skip, jail=jail)
+            if callback is not None:
```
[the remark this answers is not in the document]
skarekrow: This is just a backport ;)

skarekrow:
> @skarekrow if i am not wrong, we want to bring the mechanism of updating the jail itself as well in case of plugins, no ? If this is correct, then i think we need iocage bits as well to make it happen. Right now it only updates the software inside the plugins.

That's way too invasive unfortunately, this unblocks updating the software in the plugins which is where most of the complaints are. 11.3 will have complete updating.

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -822,33 +826,38 @@ def exec(self, jail, command, options):
...
+            elif 'Preparing to download files... done.' in msg:
+                job.set_progress(50)
+            elif 'Applying patches... done.' in msg:
+                job.set_progress(75)
+            elif 'Installing updates... done.' in msg:
+                job.set_progress(90)
+            elif f'{jail} has been updated successfully' in msg:
+                job.set_progress(100)
+
+            job.set_progress(None, description=final_msg)
```
sonicaj: Do we want this in an `else` ?

skarekrow: For what reason?

sonicaj: We don't :D

sonicaj: My bad ;)

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -822,33 +826,38 @@ def exec(self, jail, command, options):
...
+        _, path, iocage = self.check_jail_existence(
+            jail,
+            callback=progress_callback
+        )
+        jail_type = IOCJson(path).json_load()['type']
+        iocage.update()
 
-        if started:
-            self.stop(jail)
+        if jail_type == 'pluginv2':
```
sonicaj: Also is this true for `plugin` jail_type ?

skarekrow: No, those are old and don't have many of the things we expect.

skarekrow:
> I think in 11.3 we have a similar issue as well. We are not able to update the a plugin's jail.

Yeah right now the update mechanism in iocage only updates the pkgs inside the jail. Upgrade is used to move the jail to a new world. We will need to extend iocage to also run an update on update. Good catch!

PR #3140, opened 2019-06-04
NAS-101997 / 11.2 / Allow integer input for serialport field

sonicaj:
This commit fixes a bug where we expected serialport to be in hexadecimal always. Now if it is provided in base 10, we normalize it and convert to hexadecimal.
Ticket: #NAS-101997

### DIFF
--- a/src/middlewared/middlewared/plugins/system.py
+++ b/src/middlewared/middlewared/plugins/system.py
@@ -96,11 +96,16 @@ class SytemAdvancedService(ConfigService):
                     f'{schema}.serialport',
                     'Please specify a serial port when serial console option is checked'
                 )
-            elif serial_choice not in await self.serial_port_choices():
-                verrors.add(
-                    f'{schema}.serialport',
-                    'Serial port specified has not been identified by the system'
-                )
+            else:
+                data['serialport'] = serial_choice = hex(
+                    int(serial_choice)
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/system.py
```diff
@@ -96,11 +96,16 @@ class Config:
...
+            else:
+                data['serialport'] = serial_choice = hex(
+                    int(serial_choice)
+                ) if serial_choice.isdigit() else serial_choice
+
+                if serial_choice not in await self.serial_port_choices():
+                    verrors.add(
+                        f'{schema}.serialport',
+                        'Serial port specified has not been identified by the system'
+                    )
```
william-gr: What is an example of decimal input port?

sonicaj: `1016`

william-gr: How is that a decimal?

william-gr: You mean integer?

sonicaj: Yup, base 10

william-gr: Ok, lets reword the ticket/commit please, decimal is very confusing

PR #3141, opened 2019-06-05
NAS-101827 / 11.3 / Download/Upload config with encrypted pool keys

sonicaj:
This PR adds the ability to be able to download/upload config with encrypted pool keys.
Ticket: #NAS-101827

### DIFF
--- a/src/middlewared/middlewared/plugins/config.py
+++ b/src/middlewared/middlewared/plugins/config.py
@@ -13,7 +13,10 @@ from datetime import datetime
 
 from middlewared.schema import Bool, Dict, accepts
 from middlewared.service import CallError, Service, job, private
+from middlewared.plugins.pwenc import PWENC_FILE_SECRET
+from middlewared.plugins.pool import GELI_KEYPATH
 
+CONFIG_FILES = {'pwenc_secret': PWENC_FILE_SECRET, 'geli': GELI_KEYPATH}
 FREENAS_DATABASE = '/data/freenas-v1.db'
 NEED_UPDATE_SENTINEL = '/data/need-update'
 RE_CONFIG_BACKUP = re.compile(r'.*(\d{4}-\d{2}-\d{2})-(\d+)\.db$')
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/config.py
```diff
@@ -13,38 +13,50 @@
 
 from middlewared.schema import Bool, Dict, accepts
 from middlewared.service import CallError, Service, job, private
+from middlewared.plugins.pwenc import PWENC_FILE_SECRET
+from middlewared.plugins.pool import GELI_KEYPATH
 
 FREENAS_DATABASE = '/data/freenas-v1.db'
 NEED_UPDATE_SENTINEL = '/data/need-update'
 RE_CONFIG_BACKUP = re.compile(r'.*(\d{4}-\d{2}-\d{2})-(\d+)\.db$')
+CONFIG_FILES = {'pwenc_secret': PWENC_FILE_SECRET, 'geli': GELI_KEYPATH}
```
william-gr: Let me be pedantic here and ask you to put it as the first global var, to be alphabetically ordered :P

src/middlewared/middlewared/plugins/config.py
```diff
@@ -13,38 +13,50 @@
...
+        Bool('pool_keys', default=False),
     ))
     @job(pipes=["output"])
-    async def save(self, job, options=None):
+    async def save(self, job, options):
         """
         Provide configuration file.
 
-        secretseed - will include the password secret seed in the bundle.
+        `secretseed` - will include the password secret seed in the bundle.
```
william-gr: And remove the dash (-)

src/middlewared/middlewared/plugins/config.py
```diff
@@ -13,38 +13,50 @@
...
-        if not options.get('secretseed'):
+        if not options['secretseed'] and not options['pool_keys']:
             bundle = False
             filename = FREENAS_DATABASE
         else:
             bundle = True
+            files = CONFIG_FILES.copy()
+            if not options['secretseed']:
+                files['secretseed'] = None
+            if not options['pool_keys'] or not os.path.exists(files['geli']):
```
william-gr: The dir could exist and not have anything within it, will that create an empty fir in the tar?

sonicaj: That will create an empty directory when the `tar` is extracted. I'll update this to also not include this if the directory is empty.

PR #3142, opened 2019-06-05
Install kernel-debug with other OS bits, and make sure a failure

kmoore134:
here is fatal to installer

### DIFF
--- a/src/freenas-installer/etc/installer12.sh
+++ b/src/freenas-installer/etc/installer12.sh
@@ -1122,11 +1122,12 @@ menu_install()
     mkdir -p /tmp/data/var/db/pkg
     export PKG_DBDIR="/tmp/data/var/db/pkg"
 
-    for inspkg in os/userland os/kernel ports-mgmt/pkg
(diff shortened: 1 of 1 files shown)


src/freenas-installer/etc/installer12.sh
```diff
@@ -1122,11 +1122,12 @@ menu_install()
     mkdir -p /tmp/data/var/db/pkg
     export PKG_DBDIR="/tmp/data/var/db/pkg"
 
-    for inspkg in os/userland os/kernel ports-mgmt/pkg
+    for inspkg in os/userland os/kernel os/kernel-debug ports-mgmt/pkg
     do
         env ASSUME_ALWAYS_YES=YES pkg -r /tmp/data install -yf $inspkg
         if [ $? -ne 0 ] ; then
             echo "Failed installing ${inspkg}"
+	    exit 1
```
sonicaj: Just a cosmetic note, i think we have a clash of spaces/tabs here.

PR #3144, opened 2019-06-05
NAS-101886 / 11.3 / Prevent cutting long lines pasted into terminal

themylogin:


src/middlewared/middlewared/main.py
```diff
@@ -550,12 +551,11 @@ def run(self):
         self.shell_pid, master_fd = os.forkpty()
         if self.shell_pid == 0:
             for i in range(3, 1024):
-                if i == master_fd:
-                    continue
```
william-gr: Why this change?

themylogin: It was useless, in child process `master_fd` is `-1`

william-gr: In that case we should use bsd.closefrom, which is much better than iterating 1021 times

william-gr:
@themylogin are you comfortable with having this merged into 11.2? what are the chances something with closefrom breaks or the terminal that does work quite right?

I would be fine with only 11.3 fix

themylogin:
I am, but since no one really complained about it, let's not tempt our fate.

PR #3145, opened 2019-06-05
NAS-101739 / 11.3 / fix(vm): We also need to respect arc_meta_limit

skarekrow:
Without doing so, if arc_max goes below arc_meta_limit, we will get an invalid argument from sysctl.

NAS-101739


skarekrow:
backport

PR #3149, opened 2019-06-05
NAS-101906 / 11.3 / feat(usage): Allow opting out of usage collection

skarekrow:
This will be in System - General, same as crash reporting.

NAS-101906


src/middlewared/middlewared/plugins/usage.py
```diff
@@ -399,14 +399,15 @@ class Config:
 
 
 async def setup(middleware):
-    now = datetime.utcnow()
-    event_loop = asyncio.get_event_loop()
+    if (await middleware.call('system.general.config'))['usage_collection']:
```
themylogin: Need a code which will stop/start this when user changes value.

src/middlewared/middlewared/plugins/usage.py
```diff
@@ -12,6 +12,11 @@ class Config:
         private = True
 
     async def start(self):
+        if not (
+            await self.middleware.call('system.general.config')
+        )['usage_collection']:
+            return False
```
themylogin: You only need to skip gather & send section, not the entire block, otherwise if data collection is re-enabled, loop won't be resumed.

PR #3151, opened 2019-06-06
NAS-102024 / 11.3 / Replication jobs

themylogin:


src/middlewared/middlewared/plugins/zettarepl.py
```diff
@@ -511,22 +572,69 @@ def _observer_queue_reader(self):
...
+                    if channel:
+                        channel.put(message)
+                    else:
+                        self.logger.warning("Don't have job channel for task %r", message.task_id)
+
             except Exception:
                 self.logger.warning("Unhandled exception in observer_queue_reader", exc_info=True)
 
+    @job(logs=True)
+    def replication_job(self, job, task_id):
```
william-gr: Thats awesome!

One thing though, zettarepl is a private service.
What are your thoughts of reusing `replication.run` to do it, turning it into a job?

That way we could call it just like any other method and if there isnt a repl job channel we start one. Possible?

PR #3156, opened 2019-06-07
NAS-101995 / 11.3 / Keychain Credentials: used_by and allow cascade delete

themylogin:


src/middlewared/middlewared/plugins/keychain.py
```diff
@@ -23,13 +24,84 @@ class KeychainCredentialType:
...
+    async def unbind(self, row):
+        raise NotImplementedError
+
+
+class KeychainCredentialUsedByDelegateUnbindMethod(enum.Enum):
+    DELETE = "delete"
+    DISABLE = "disable"
+
+
+class OtherKeychainCredentialKeychainCredentialUsedByDelegate(KeychainCredentialUsedByDelegate):
```
william-gr: KeychainCredentialKeychainCredential ?

src/middlewared/middlewared/plugins/keychain.py
```diff
@@ -277,16 +356,39 @@ class Config:
...
         await self.middleware.call(
             "datastore.delete",
             self._config.datastore,
             id,
         )
 
+    @accepts(Int("id"))
+    async def used_by(self, id):
+        """
+        Returns list of objects that use this credential
```
william-gr: dot at the end of phrase

src/middlewared/middlewared/plugins/keychain.py
themylogin: KeychainCredentialUsedByDelegate for keychain credential that is used by Other Keychain Credential

PR #3160, opened 2019-06-07
NAS-101742 / 11.3 / Adapt subprocess.CalledProcessError to CallError

themylogin:


themylogin:
> This is going to change how these exceptions are raised even for internal calls.

@william-gr yes, I've missed that. Updated my PR. Now exceptions are only changed when sending them over WebSocket or HTTP. The only problem left is jobs.

themylogin:
> run() could be similarly adapted to give more meaningful output for check() ?

That won't be satisfying because synchronous methods use `subprocess` directly

PR #3162, opened 2019-06-07
NAS-102043 / 11.2 / Prevent timeouts on post pool creation tasks by making them jobs

themylogin:


gui/storage/forms.py
```diff
@@ -329,10 +329,11 @@ def save(self):
 
         # This must be outside transaction block to make sure the changes
         # are committed before the call of ix-fstab
-        notifier().reload("disk")
-        if not add:
-            notifier().start("ix-syslogd")
-            notifier().restart("system_datasets")
+        with client as c:
+            c.call("core.bulk", "service.reload", [["disk"]], job=True)
```
william-gr: We dont think we need to wait for it to finish, it can run in background

PR #3163, opened 2019-06-09
NAS-101447 / 11.3 / Swagger definition issues

sonicaj:
This PR aims to fixes errors outlined by swagger in our schema layer.

Ticket: #NAS-101447


src/middlewared/middlewared/plugins/jail.py
```diff
@@ -103,7 +103,7 @@ def query(self, filters=None, options=None):
             self.logger.debug('Failed to get list of jails', exc_info=True)
 
         return filter_list(jails, filters, options)
-    query._fiterable = True
+    query._filterable = True
```
william-gr: I think `@filterable` has been fixed since and it can be used in jail plugin. Can you test, please?

PR #3168, opened 2019-06-10
NAS-101175 / 12 / Graphite SeparateInstances

themylogin:


themylogin:
> it will break the reporting plugin events if its checked

We only add new option to `<Node "graphite">`

`<Node "middleware">` continues working as before

Or do you mean something else?

PR #3173, opened 2019-06-10
NAS-101761 / 11.2 / Improve pool.filesystem_choices performance

themylogin:


william-gr:
Thats not necessary for 11.2 I'd think, we already made improvements with py-libzfs by other means since this report.

themylogin:
@william-gr you mean performance improvements since 11.2-U4.1? Because that's what Affected Version is. In test system it takes 3 seconds to list empty 1400 datasets using standard method, I suppose it might be worse if pool structure is more sophisticated.

william-gr:
3 seconds is not bad, I will take it. That many datasets is very unusual.

themylogin:
> I will take it

I am not sure I catch that, you want to work on this yourself from now on?

william-gr:
No, I mean 3 seconds is good enough for 11.2.

themylogin:
Yes, but looks like this user had > 60, because his pool has more data in it, metadata scattered all across hundreds of terabytes, takes a long time to read it all?.

william-gr:
How long is it taking, then? If its timing out (60 seconds) to list > 60 datasets seems like we have bigger problems than giving him filesystem choices.

themylogin:
I suppose he manages these using ZFS CLI, What he fails to do is to manage replication task, and this fix will probably give him that option because I don't see other obstacles. He was able to successfully use replication by manually inserting rows into SQLite DB.

william-gr:
Many other things rely on that. Do you think the periodic task to verify dataset quota wont fall over if it takes minutes?
Or even listing the datasets on the UI which should use that call under the hood?

themylogin:
They do, but he only had problem with this.

I asked him in the ticket to check if the problem is surely caused by `pool.filesystem_choices` and nothing else

themylogin:
Yes, it's that, and he's using that system successfully. Why do you think it would hurt us to merge this?

william-gr:
It _may_ not hurt, every change we merge from now on must be done with extreme care.

More importantly, you will be tackling a side effect, not the cause. I can doubt he is using the system successfully, clearly listing the datasets in UI will also take a long time.

themylogin:
backport

PR #3176, opened 2019-06-11
NAS-102062 / 11.3 / Prevent users from naming SMB shares 'printers' and 'homes'

anodos325:
Expand validation to include case insensitive check for samba's special reserved sections "global", "printers", and "homes". "global" is always present, "printers" isn't used in FreeNAS, and "homes" is controlled by flagging a share

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -760,10 +760,10 @@ class SharingSMBService(CRUDService):
                     f'The "noacl" VFS module is incompatible with the extended ACL on {data["path"]}.'
                 )
 
-        if data.get('name') and data['name'] == 'global':
+        if data.get('name') and data['name'].lower() in ['global', 'homes', 'printers']:
             verrors.add(
                 f'{schema_name}.name',
-                'Global is a reserved section name, please select another one'
+                f'{data["name"]} is a reserved section name, please select another one'
             )
 
     @private


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -760,10 +760,10 @@ class Config:
                     f'The "noacl" VFS module is incompatible with the extended ACL on {data["path"]}.'
                 )
 
-        if data.get('name') and data['name'] == 'global':
+        if data.get('name') and data['name'].lower() in ['global', 'homes', 'printers']:
             verrors.add(
                 f'{schema_name}.name',
-                'Global is a reserved section name, please select another one'
+                f'{data["name"]} is a reserved section name, please select another one'
```
wblock: @anodos325 
Suggestions:
1. Put the reserved word in quotes to delimit it from the rest of the message:
``` print(f'"{data["name"]}" is a ... ```
2.  "reserved section name" assumes Samba knowledge, doesn't really help the user.  Better to just say "is a reserved word and cannot be used for share names".
3. Since the entire list of reserved names is so short, it could actually be included in the error message in another sentence: 'Reserved names that cannot be used for shares include "global", "homes", "printers".' (Of course that should be defined as a global somewhere and reused:
```
samba_reserved_names = ['global', 'homes', 'printers']
if data.get('name') and data['name'].lower() in samba_reserved_names:
    verrors.add(
    f'{schema_name}.name',
    f'"{data["name"]}" is a reserved word and cannot be used for share names. ' +
    f'Reserved names that cannot be used for shares include {", ".join(samba_reserved_names)}.'
```

PR #3177, opened 2019-06-11
NAS-100137 / 12 / NFSv4 domain

themylogin:


src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -111,6 +114,9 @@ def nfs_compress(self, nfs):
             verrors.add(
                 "nfs_update.userd_manage_gids", "This option is incompatible with NFSv3 ownership model for NFSv4")
 
+        if not new["v4"] and new["v4_domain"]:
+            verrors.add("nfs_update.v4_domain", "This option has no sense for NFSv3")
```
yocalebo: Let's change verbiage to "This option does not apply to NFSv3"

PR #3178, opened 2019-06-11
Fix NFSv4 not working when "Start on boot" is not checked

themylogin:


themylogin:
@yocalebo they all are disabled by default so `gssd_enable=NO` has no sense.

Also `gssd` and `nfsuserd` are always started regarding of NFSv4 option because `service.start` is `onestart` by default. Does it cause any trouble, should I also fix this as well?

PR #3182, opened 2019-06-11
NAS-102021 / 12 / Accurately Retrieve Boot Environment space

sonicaj:
This PR aims to introduce changes to make sure we accurately retrieve boot environment space with focus being on displaying space such as how much space would be freed if the respective BE is deleted.

Ticket: #NAS-102021

### DIFF
--- a/src/middlewared/middlewared/plugins/zfs.py
+++ b/src/middlewared/middlewared/plugins/zfs.py
@@ -441,15 +441,21 @@ class ZFSDatasetService(CRUDService):
                     'type': type_.upper(),
                 })
         else:
+            extra = options.get('extra', {})
+            state_options = {
+                'snapshots': extra.get('snapshots', False),
+                'recursive': extra.get('recursive', True),
+                'snapshots_recursive': extra.get('snapshots_recursive', False)
+            }
             with libzfs.ZFS() as zfs:
                 # Handle `id` filter specially to avoiding getting all datasets
                 if filters and len(filters) == 1 and list(filters[0][:2]) == ['id', '=']:
                     try:
-                        datasets = [zfs.get_dataset(filters[0][2]).__getstate__()]
+                        datasets = [zfs.get_dataset(filters[0][2]).__getstate__(**state_options)]
                     except libzfs.ZFSException:
                         datasets = []
                 else:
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -424,8 +424,9 @@ class Config:
     def query(self, filters=None, options=None):
         # If we are only filtering by name, pool and type we can use
         # zfs(8) which is much faster than py-libzfs
+        options = options or {}
```
william-gr: Is this really necessary? options should default to `{}`

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -424,8 +424,9 @@ class Config:
...
         if (
-            options and set(options['select']).issubset({'name', 'pool', 'type'}) and
+            options and set(options.get('select', [])).issubset({'name', 'pool', 'type'}) and
```
william-gr: also here, select should default to [], no?

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -441,15 +442,21 @@ def query(self, filters=None, options=None):
                     'type': type_.upper(),
                 })
         else:
+            extra = options.pop('extra', {})
+            state_options = {
+                'snapshots': extra.pop('snapshots', False),
+                'recursive': extra.pop('recursive', True),
+                'snapshots_recursive': extra.pop('snapshots_recursive', False)
+            }
```
william-gr: I dont think we should `pop`, that will change the object from the caller. Possibly irrelevant in this case since its a new object, but still.

src/middlewared/middlewared/plugins/zfs.py
sonicaj: `options` by default is `None` as set by `query-options` schema

william-gr: Please elaborate, every Dict is special and should default to {}

william-gr: I see, you said query-options, not in this query. I think we should remove default=None there

sonicaj: That makes more sense ;)

PR #3186, opened 2019-06-11
NAS-100329 / 11.3 / feat(vm/logging): Keep individual logs for each VM

skarekrow:
They will remain in the main middlewared.log, but we also should break them out to their own log file.  This aids in debugging, and sorts out some noise.

NAS-100329

### DIFF
--- a/src/middlewared/middlewared/plugins/vm.py
+++ b/src/middlewared/middlewared/plugins/vm.py
@@ -89,9 +89,7 @@ class VMSupervisor(object):
     def __init__(self, manager, vm):
         self.manager = manager
 
-        if not os.path.exists('/var/log/vm'):
-            os.makedirs('/var/log/vm', exist_ok=True)
-
+        os.makedirs('/var/log/vm', exist_ok=True)
         self.logger = self.manager.logger.getChild(f'vm_{vm["id"]}')
 
         handler = middlewared.logger.ErrorProneRotatingFileHandler(
--- a/src/middlewared/middlewared/plugins/vm.py
+++ b/src/middlewared/middlewared/plugins/vm.py
@@ -97,12 +97,10 @@ class VMSupervisor(object):
             maxBytes=10485760,
             backupCount=5,
         )
-        handler.setFormatter(logging.Formatter(
-            '[%(asctime)s] (%(levelname)s) %(name)s.%(funcName)s():%(lineno)d'
-            ' - %(message)s'
-        ))
-        self.logger.addHandler(handler)  # main log + vm specific log
         self.middleware = self.manager.service.middleware
+
+        handler.setFormatter(logging.Formatter(self.middleware.log_format))
+        self.logger.addHandler(handler)  # main log + vm specific log
         self.logger.setLevel(self.middleware.debug_level)
 
         self.vm = vm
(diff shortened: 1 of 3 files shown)


william-gr:
Looks reasonable but I want @themylogin opinion too :P

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -87,7 +88,26 @@ class VMSupervisor(object):
 
     def __init__(self, manager, vm):
         self.manager = manager
-        self.logger = self.manager.logger
+
+        if not os.path.exists('/var/log/vm'):
+            os.makedirs('/var/log/vm')
+
+        self.logger = middlewared.logger.Logger(
```
themylogin: To ensure loggers hierarchy we usually do it like this `self.manager.logger.getChild(f'vm_{vm["id"]}')`. Also we usually don't put human-readable things like VM names (that can contain whitespaces, punctuation and other stuff) to logger names because they are identificators.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -87,7 +88,26 @@ class VMSupervisor(object):
...
+            f'VM: {vm["name"]}_{vm["id"]}'
+        ).getLogger()
+
+        handler = middlewared.logger.ErrorProneRotatingFileHandler(
+            f'/var/log/vm/{vm["name"]}_{vm["id"]}',
+            maxBytes=10485760,
+            backupCount=5,
+        )
+        handler.setFormatter(logging.Formatter(
+            '[%(asctime)s] (%(levelname)s) %(name)s.%(funcName)s():%(lineno)d'
```
themylogin: Maybe we should make this a constant somewhere?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -87,7 +88,26 @@ class VMSupervisor(object):
...
+            ' - %(message)s'
+        ))
+        self.logger.addHandler(handler)  # main log + vm specific log
+        self.logger.setLevel(logging.DEBUG)
```
themylogin: Is it possible to inherit logging level passed to middlewared via CLI?

src/middlewared/middlewared/plugins/vm.py
skarekrow: It's only used once, so I'm not sure if that's a major benefit.

themylogin: @skarekrow 
https://github.com/freenas/freenas/blob/master/src/middlewared/middlewared/logger.py#L228
https://github.com/freenas/freenas/blob/master/src/middlewared/middlewared/logger.py#L278

skarekrow: Sorry I should have been more clear, in the VM plugin.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -87,8 +88,25 @@ class VMSupervisor(object):
 
     def __init__(self, manager, vm):
         self.manager = manager
-        self.logger = self.manager.logger
+
+        if not os.path.exists('/var/log/vm'):
```
themylogin: This check is not required now

skarekrow: Fair point! One moment.

src/middlewared/middlewared/plugins/vm.py
themylogin: And it should be global

PR #3187, opened 2019-06-12
NAS-100209 / 12 / Add comment field for  initshutdown model

sonicaj:
This commit adds a migration and updates the initshutdown model to include a comment field.

### DIFF
--- a/gui/tasks/migrations/0014_ini_comment_field.py
+++ b/gui/tasks/migrations/0014_ini_comment_field.py
@@ -1,0 +1,20 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('tasks', '0013_init_tasks_timeout'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='initshutdown',
+            name='ini_comment',
+            field=models.CharField(
+                verbose_name='Comment',
+                blank=True,
+                max_length=255,
+            ),
+        ),
+    ]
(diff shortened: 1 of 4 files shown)


gui/tasks/migrations/0014_ini_comment_field.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('tasks', '0013_init_tasks_timeout'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='initshutdown',
+            name='ini_comment',
+            field=models.TextField(
```
william-gr: Not sure we want a textfield here, it seems to make sense to restrict the size of that field? Thoughts?

sonicaj: I did think of that initially, limiting to 100 chars perhaps as in tunables but then thought there might be some users who type longer comments ? But if you think that is better, updating the PR with that.

william-gr: Yes, a limit makes sense to me, it doesnt have to be as low as 100, could be 200

sonicaj: Perfect, i'll do that

src/middlewared/middlewared/plugins/init_shutdown_script.py
```diff
@@ -22,6 +24,7 @@ class Config:
         Str('when', enum=['PREINIT', 'POSTINIT', 'SHUTDOWN'], required=True),
         Bool('enabled', default=True),
         Int('timeout', default=10),
+        Str('comment', default=''),
```
william-gr: Range validator instead of that and COMMENT_LENGTH?

william-gr: Ah, that doesnt work for Str, I guess?

sonicaj: Good point, it doesn't but we should adapt it to work for string values as well checking their length if the value is str else integer it already is

sonicaj:
I'll update it to 255 as well, let's save the extra byte ;)

PR #3189, opened 2019-06-12
Make swap mirrors as many disks as amount of max acceptable failed disks

danny8376:
If we mirroring with more than 2 disks or using raid-z2/3, we assume that there'll be more than one disk failing at same time in the worst scenario.
However, the current swap mirror setup may crash the system in the worst case.
Try to fix this by checking type of vdevs of all pools when generating swap.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk.py
+++ b/src/middlewared/middlewared/plugins/disk.py
@@ -1306,6 +1306,30 @@ class DiskService(CRUDService):
         """
         await self.middleware.run_in_thread(geom.scan)
 
+        group_size = 2
+        pools = await self.middleware.call('pool.query')
+        for pool in pools:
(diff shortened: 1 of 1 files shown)


william-gr:
Thanks for the contribution, its very much appreciated.

I can think of one downside of this change, adding a new pool with higher redundancy will likely make it remove all swaps and start over? (If there isnt enough space in RAM it will fail, also).
Do you think thats an acceptable trade off?

danny8376:
It's reasonable that swap redundancy grow higher with new pool with higher redundancy.
However, i think it's not that suitable to remove all swap and start over.
Maybe need a further rewritten for swaps_configure (try to maintain mirror member instead of destroying then create)

themylogin:
Continuing this work here: https://github.com/freenas/freenas/pull/3462

PR #3192, opened 2019-06-12
NAS-100722 / 11.3 / Enforce string limits for services

sonicaj:
This PR aims to fixes the following:
1) Charfield max limits are not being imposed by sqlite. This is to enforce those limits at schema level in middlewared. We do so by defining a default limit of 1024 characters and for cases we would like to mimic textfield, we use none which copies maximum value sqlite accepts.
2) Fix a django form validation error where it validates disk serial of more then 30 characters.

Ticket: #NAS-100722


sonicaj:
@themylogin @anodos325 @william-gr please have a look at the changes. I have mostly reflected the limits enforced by database in schema, however i am sure we can increase/decrease these limits more for certain options. Please have a look and let me know what you think. Thank you

themylogin:
So, from now on, when introducing new field to our models, how do I choose, should it be 120 characters, 255 characters or no limit?

sonicaj:
That limit would be now enforced at schema level instead of the model as it's pointless doing there. We can work on that bit too fixing this at model level but as discussed with @william-gr, that poses other issues. Now say we have a disk with very long serial number, that would never have been inserted in the db in the first place.

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -475,22 +476,22 @@ def getparm(self, parm, section):
...
+        Str('netbiosname', validators=[Range(max=120)]),
+        Str('netbiosname_b', validators=[Range(max=120)]),
         List('netbiosalias', default=[]),
-        Str('workgroup'),
-        Str('description'),
+        Str('workgroup', validators=[Range(max=120)]),
+        Str('description', validators=[Range(max=120)]),
         Bool('enable_smb1'),
-        Str('unixcharset'),
+        Str('unixcharset', validators=[Range(max=120)]),
```
themylogin: Am I the only one, who is annoyed with this noise?
`netbiosname` is up to 120 characters, but `netbiosalias` can be of any length, is this what this code says?

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -475,22 +476,22 @@ def getparm(self, parm, section):
 
     @accepts(Dict(
         'smb_update',
-        Str('netbiosname'),
-        Str('netbiosname_b'),
+        Str('netbiosname', validators=[Range(max=120)]),
+        Str('netbiosname_b', validators=[Range(max=120)]),
         List('netbiosalias', default=[]),
```
themylogin: Am I the only one who is annoyed with this newly introduced noise?
`netbiosname` is up to 120 characters, but `netbiosalias` can be of any length, is this what this code says?

sonicaj: The model does enforce a restriction on this ( or tries to being more accurate ), but as we have it as a list, i wasn't sure what limit to impose here.

yocalebo: @sonicaj Microsoft doesn't allow more than 15 characters for NetBIOS names and 255 characters for the FQDN.

gui/storage/models.py
```diff
@@ -322,8 +322,7 @@ class Disk(Model):
         editable=False,
         default=1,
     )
-    disk_serial = models.CharField(
-        max_length=30,
+    disk_serial = models.TextField(
```
william-gr: I was thinking of ignoring the value sent on POST so we dont have to change the model which will grant a migration.

src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -437,11 +438,10 @@ class Config:
 
     @accepts(
         Int('id', required=True),
-        Dict(
+        Patch(
+            'kerberos_keytab_create',
```
william-gr: Why are you changing this in this PR?

sonicaj: I changed this in this PR because i added length validation in `kerberos_keytab_create` schema and would have to do that for update as well, so i instead used a `Patch` as we do elsewhere

src/middlewared/middlewared/schema.py
```diff
@@ -155,6 +155,10 @@ def to_json_schema(self, parent=None):
 
 class Str(EnumMixin, Attribute):
 
+    def __init__(self, *args, **kwargs):
+        self.max_length = kwargs.pop('max_length', 1024) or (2 ** 31) - 1
```
william-gr: A comment explaining why 2 ** 31 - 1 might come handy in the future

src/middlewared/middlewared/schema.py
```diff
@@ -184,6 +188,19 @@ def to_json_schema(self, parent=None):
...
         return schema
 
+    def validate(self, value):
+        if value is None:
+            return value
+
+        verrors = ValidationErrors()
+
+        if value and len(value) > self.max_length:
+            verrors.add(self.name, f'Value greater then {self.max_length} not allowed')
```
william-gr: than

src/middlewared/middlewared/schema.py
```diff
@@ -155,6 +155,10 @@ def to_json_schema(self, parent=None):
 
 class Str(EnumMixin, Attribute):
 
+    def __init__(self, *args, **kwargs):
+        self.max_length = kwargs.pop('max_length', 1024) or (2 ** 31) - 1
+        super(Str, self).__init__(*args, **kwargs)
```
william-gr: super(), class and self not required, like you did below.

PR #3193, opened 2019-06-12
NAS-100818 / 11.3 / feat(jail/list_resource): Return PLUGIN_INFO

skarekrow:
This will append the PLUGIN_INFO to a plugin for the UI to display in a web element, otherwise null.

NAS-100818

### DIFF
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -571,26 +571,26 @@ class JailService(CRUDService):
                     # iocage returns - for None
                     plugin[i] = elem if elem != "-" else None
 
-                # plugin[1] is the UUID
-                plugin_output = pathlib.Path(
-                    f'{iocroot}/jails/{plugin[1]}/root/root/PLUGIN_INFO'
-                )
-
-                if plugin_output.is_file():
-                    plugin_info = [[
-                        x for x in plugin_output.read_text().split('\n') if x
-                    ]]
-                else:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/jail.py
```diff
@@ -571,13 +571,26 @@ def list_resource(self, resource, remote, want_cache, branch):
                     # iocage returns - for None
                     plugin[i] = elem if elem != "-" else None
 
+                # plugin[1] is the UUID
+                plugin_output = pathlib.Path(
+                    f'{iocroot}/jails/{plugin[1]}/root/root/PLUGIN_INFO'
```
sonicaj: if `remote` is `True`, `iocroot` won't exist.

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -571,13 +571,26 @@ def list_resource(self, resource, remote, want_cache, branch):
                     # iocage returns - for None
                     plugin[i] = elem if elem != "-" else None
 
+                # plugin[1] is the UUID
```
sonicaj: Also i believe `0` index contains the plugin name, not sure but just did `iocage.fetch` locally and it seems the plugins description is at `1`.

skarekrow: iocage.fetch isn't the same, this iterates on all the elements. 1 is the name :)

PR #3196, opened 2019-06-13
NAS-101171 / 11.3 / Add parameter to disable DS cache for LDAP

anodos325:
Mirror what we have in AD plugin. If cache is disabled,
then don't fill the cache with ldap users. Also disable
enumeration in the nslcd.conf file.

### DIFF
--- a/gui/directoryservice/migrations/0013_add_disable_freenas_cache_to_ldap.py
+++ b/gui/directoryservice/migrations/0013_add_disable_freenas_cache_to_ldap.py
@@ -17,7 +17,7 @@ class Migration(migrations.Migration):
                 verbose_name='Disable LDAP user/group cache',
                 default=False,
                 help_text=(
-                    "Set this if you want to disable caching LDAP users "
+                    "Set to disable caching LDAP users "
                     "and groups. This is an optimization for large LDAP "
                     "Environments. When caching is disabled, LDAP users "
                     "and groups do not appear in dropdown menus, but are "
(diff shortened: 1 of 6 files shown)


gui/directoryservice/migrations/0013_add_disable_freenas_cache_to_ldap.py
```diff
@@ -0,0 +1,28 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name='LDAP',
+            name='ldap_disable_freenas_cache',
+            field=models.BooleanField(
+                verbose_name='Disable LDAP user/group cache',
+                default=False,
+                help_text=(
+                    "Set this if you want to disable caching LDAP users "
```
a-stjohn:
```suggestion
                    "Set to disable caching LDAP users "
```

gui/directoryservice/models.py
```diff
@@ -1099,6 +1099,17 @@ class LDAP(DirectoryServiceBase):
         null=True,
         limit_choices_to={'cert_certificate__isnull': False, 'cert_privatekey__isnull': False}
     )
+    ldap_disable_freenas_cache = models.BooleanField(
+        verbose_name=_("Disable LDAP user/group cache"),
+        help_text=_(
+            "Set this if you want to disable caching LDAP users "
```
a-stjohn:
```suggestion
            "Set to disable caching LDAP users "
```

PR #3199, opened 2019-06-13
NAS-100704 / 11.3 / Retrieve plugin versions

sonicaj:
This PR introduces changes to reliably retrieve plugin versions using `packagesite` defined for each plugin.

### DIFF
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -383,7 +383,7 @@ class JailService(CRUDService):
         """Returns the activated pool if there is one, or None"""
         try:
             pool = ioc.IOCage(skip_jails=True).get('', pool=True)
-        except RuntimeError as e:
+        except (RuntimeError, SystemExit) as e:
             raise CallError(f'Error occurred getting activated pool: {e}')
         except (ioc_exceptions.PoolNotActivated, FileNotFoundError):
             self.check_dataset_existence()
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -4,6 +4,7 @@ import subprocess as su
 import libzfs
 import requests
 import itertools
+import tempfile
 import pathlib
 import json
 import sqlite3
@@ -631,38 +632,33 @@ class JailService(CRUDService):
 
         return resource_list
 
+    @periodic(interval=86400)
     @private
+    @accepts(Str('branch', null=True, default=None))
     @job(lock='retrieve_plugin_versions')
-    def retrieve_plugin_versions(self, job):
-        branch = self.get_version()
-        ioc_plugin = IOCPlugin(branch=branch)
-        if self.get_activated_pool():
-            return ioc_plugin.fetch_plugin_versions()
+    def retrieve_plugin_versions(self, job, branch=None):
+        branch = branch or self.get_version()
+        try:
+            pool = self.get_activated_pool()
+        except CallError:
+            pool = None
+
+        if pool:
+            plugins = IOCPlugin(branch=branch).fetch_plugin_versions()
         else:
-            # iocage hasn't been activated yet, we should use requests to retrieve plugin json files
-            session = requests.Session()
-            url = 'https://raw.githubusercontent.com/freenas/iocage-ix-plugins/{branch}/{file_name}'
-            try:
-                index = session.get(url.format(branch=branch, file_name='INDEX'), timeout=20)
-                index.raise_for_status()
-            except Exception:
-                return {}
-
-            plugins_index_data = {}
-            index = index.json()
-            for plugin in index:
+            with tempfile.TemporaryDirectory() as td:
+                github_repo = 'https://github.com/freenas/iocage-ix-plugins.git'
                 try:
-                    plugin_data = session.get(url.format(branch=branch, file_name=f'{plugin}.json'), timeout=10)
-                    plugin_data.raise_for_status()
+                    IOCPlugin._clone_repo(branch, github_repo, td, depth=1)
                 except Exception:
-                    pass
+                    self.middleware.logger.error('Failed to clone iocage-ix-plugins repository.', exc_info=True)
+                    return {}
                 else:
-                    plugins_index_data[plugin] = {
-                        'primary_pkg': index[plugin].get('primary_pkg'),
-                        **(plugin_data.json())
-                    }
+                    plugins_index_data = IOCPlugin.retrieve_plugin_index_data(td)
+                    plugins = IOCPlugin.fetch_plugin_versions_from_plugin_index(plugins_index_data)
 
-            return ioc_plugin.fetch_plugin_versions_from_plugin_index(plugins_index_data)
+        self.middleware.call_sync('cache.put', 'iocage_plugin_versions', plugins)
+        return plugins
 
     @accepts(Str("action", enum=["START", "STOP", "RESTART"]))
     def rc_action(self, action):
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -1152,7 +1152,7 @@ class JailService(CRUDService):
 
         try:
             base_plugin = plugin.rsplit('_', 1)[0]  # May have multiple
-            primary_pkg = index_json[base_plugin]['primary_pkg']
+            primary_pkg = index_json[base_plugin].get('primary_pkg') or plugin
             version = ['N/A', 'N/A']
 
             # Since these are plugins, we don't want to spin them up just to
@@ -1162,6 +1162,11 @@ class JailService(CRUDService):
                 primary_pkg)
 
             for row in db_rows:
+                row = list(row)
+                if '/' not in primary_pkg:
+                    row[1] = row[1].split('/', 1)[-1]
+                    row[2] = row[2].split('/', 1)[-1]
+
                 if primary_pkg == row[1] or primary_pkg == row[2]:
                     version = [row[3], '1']
                     break
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -1083,65 +1083,6 @@ class JailService(CRUDService):
         return True
 
     @private
-    def get_plugin_version(self, pkg):
-        """
-        Fetches a list of pkg's from the http://pkg.cdn.trueos.org/iocage/
-        repo and returns a list with the pkg version and plugin revision
-        """
-        try:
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/jail.py
```diff
@@ -572,7 +579,8 @@ def list_resource(self, resource, remote, want_cache, branch):
                     plugin[i] = elem if elem != "-" else None
 
                 if remote:
-                    pv = self.get_plugin_version(plugin[2])
+                    plugin_version = plugins_versions_data.get(plugin[2], {}).get('version', 'N/A')
+                    pv = [plugin_version, '1' if plugin_version != 'N/A' else plugin_version]
```
sonicaj: @skarekrow does the last entry here represent patch number ? In the old implementation, we have hardcoded it to `1` and `N/A` if version can't be determined. For now i've copied the same logic

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -1094,7 +1062,7 @@ def get_local_plugin_version(self, plugin, index_json, iocroot):
 
         try:
             base_plugin = plugin.rsplit('_', 1)[0]  # May have multiple
-            primary_pkg = index_json[base_plugin]['primary_pkg']
+            primary_pkg = index_json[base_plugin].get('primary_pkg') or plugin
```
william-gr: Perhaps use the first pkg in the pkgs list?

src/middlewared/middlewared/service.py
```diff
@@ -399,6 +399,13 @@ def get_jobs(self, filters=None, options=None):
         ], filters, options)
         return jobs
 
+    @accepts(Int('id'))
+    @job()
+    def job_wait(self, job, id):
+        target_job = self.middleware.jobs.get(id)
+        target_job.wait_sync()
+        return target_job.result
```
william-gr: What if the job fails? How will you catch that?

src/middlewared/middlewared/plugins/jail.py
sonicaj: That causes issues for plugins which don't have `primary_pkg` key defined. Let's take asigra, it will start showing up version number for asigra based on `ca_root_nss` pkg which is the first. Reliability is hurt this way

src/middlewared/middlewared/plugins/jail.py
skarekrow: The last entry? Are you talking about index 2?

sonicaj: Nope, i am referring to the last entry of `pv`

sonicaj: Also @skarekrow, @william-gr i would urge we change how we show output right now, it is something along these lines
```
[
        "MineOS ",
        " MineOS is a web interface to set up and manage Minecraft servers with a few mouse clicks.",
        "mineos",
        "",
        "False",
        "https://www.trueos.org/iocage-icons/mineos.png",
        "N/A",
        "N/A"
      ],
```
UI will have to adapt to api changes for this one as it's a job now, they can perhaps take care of this too in the same way. I propose we convert this to a dictionary for readability.

william-gr: Completely agree

skarekrow: > Nope, i am referring to the last entry of `pv`

Yes that is the version. Though in recent versions the very last index shows the PLUGIN_INFO file.

sonicaj: I'll change the output format in a new ticket and have it up asap

sonicaj: > > Nope, i am referring to the last entry of `pv`
> 
> Yes that is the version. Though in recent versions the very last index shows the PLUGIN_INFO file.

I am confused @skarekrow. First entry is the actual plugin version and the second entry is a hardcoded default of `1`. What does `1` represent ?

skarekrow: IIRC it's the plugin schema, which most are 1. Unless they are a 2. In which case they specify it.

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -631,6 +664,55 @@ def list_resource(self, resource, remote, want_cache, branch):
...
+    @job(lock='retrieve_plugin_versions')
+    def retrieve_plugin_versions(self, job, branch=None):
+        branch = branch or self.get_version()
+        ioc_plugin = IOCPlugin(branch=branch)
+        if self.get_activated_pool():
+            plugins = ioc_plugin.fetch_plugin_versions()
+        else:
+            # iocage hasn't been activated yet, we should use requests to retrieve plugin json files
+            session = requests.Session()
+            url = 'https://raw.githubusercontent.com/freenas/iocage-ix-plugins/{branch}/{file_name}'
```
william-gr: This is the same problem as before, one request for each plugin.

Why is this not in iocage?

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -631,6 +664,29 @@ def list_resource(self, resource, remote, want_cache, branch):
...
+        branch = branch or self.get_version()
+        ioc_plugin = IOCPlugin(branch=branch)
+        if self.get_activated_pool():
+            plugins = ioc_plugin.fetch_plugin_versions()
+        else:
+            with tempfile.TemporaryDirectory() as td:
+                github_repo = 'https://github.com/freenas/iocage-ix-plugins.git'
+                try:
+                    IOCPlugin._clone_repo(branch, github_repo, td, depth=1)
+                except Exception:
```
william-gr: Should probably log the exception?

src/middlewared/middlewared/service.py
```diff
@@ -399,6 +399,13 @@ def get_jobs(self, filters=None, options=None):
         ], filters, options)
         return jobs
 
+    @accepts(Int('id'))
+    @job()
+    def job_wait(self, job, id):
```
william-gr: What I meant here earlier is that on error we should probably raise it, otherwise return the result? Thoughts?

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -631,6 +664,29 @@ def list_resource(self, resource, remote, want_cache, branch):
...
+                    return {}
```
skarekrow: Shouldn't we raise something here instead? @william-gr?

william-gr: I am on the fence between logging and raise

src/middlewared/middlewared/service.py
sonicaj: I don’t think that is a good idea. The caller will know if the job failed or not via the return value and it can work it out accordingly. Thoughts ?

src/middlewared/middlewared/plugins/jail.py
sonicaj: @skarekrow logging seems better in my opinion.

src/middlewared/middlewared/plugins/jail.py
sonicaj: Agreed

skarekrow: Agreed

src/middlewared/middlewared/plugins/jail.py
skarekrow: Agreed as well

src/middlewared/middlewared/service.py
skarekrow: I think explicit failure is always better than implicit

sonicaj: So if another job is waiting for for a different job, i don’t think it is a good idea to let the exception raise this way. Thoughts @skarekrow @william-gr  ?

src/middlewared/middlewared/plugins/jail.py
sonicaj: Perfect

src/middlewared/middlewared/service.py
skarekrow: I think the caller should be expecting and catching in an exception in that case. Silently discarding them or relying on return values seems odd to me.

sonicaj: We are not discarding ;) everything is available when we query for the job :p But please let me know what would be best and i can move ahead then

sonicaj: Idea is we wait for the job to complete. Not interpret how it ends, that’s my argument

william-gr: But you're not getting the Job object, just an error attribute which doesnt tell much.

So it either returns the whole job or we follow the standard practice of raising an error if something bad happened or the final result otherwise.

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -631,6 +661,34 @@ def list_resource(self, resource, remote, want_cache, branch):
 
         return resource_list
 
+    @periodic(interval=86400)
+    @private
+    @accepts(Str('branch', null=True, default=None))
```
themylogin: `@private` and `@accepts` at the same time?

sonicaj: Yup, we do that on other places as well, it's just to control the schema value being passed.

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -383,7 +384,7 @@ def get_activated_pool(self):
         """Returns the activated pool if there is one, or None"""
         try:
             pool = ioc.IOCage(skip_jails=True).get('', pool=True)
-        except RuntimeError as e:
+        except (RuntimeError, SystemExit) as e:
```
skarekrow: In this case iocage will raise a different exception now.  It raises RuntimeError (SystemExit if in foreground) if there are more than one, otherwise a PoolNotActivatedError

sonicaj: @skarekrow i don't think `PoolNotActivated` is raised because of this
```
if force_raise:
     raise callback_exception(message)
else:
     raise SystemExit(1)
```
And `force_raise` is always `False`,
```
ocage_lib.ioc_common.logit(
                                {
                                    'level': 'EXCEPTION',
                                    'message': 'Please specify a pool to'
                                    ' activate with iocage activate POOL'
                                },
                                _callback=self.callback,
                                silent=self.silent,
                                exception=ioc_exceptions.PoolNotActivated)
```

skarekrow: We should actually use force raise in the middleware. Will allow us to stop parsing for SystemExit and catch the actual exceptions iocage is raising. Gives us more insight ;)

PR #3210, opened 2019-06-17
QE NAS-101907 Adding activedirectory API v2 testing

ericbsd:


tests/api2/activedirectory.py
```diff
@@ -0,0 +1,563 @@
...
+def test_11_verify_activedirectory_data_of_(data):
+    if data == 'domainname':
+        assert results.json()[data].lower() == payload[data], results.text
+    else:
+        assert results.json()[data] == payload[data], results.text
+
+
+def test_12_setting_up_smb():
+    global payload, results
+    payload = {
```
anodos325: If you're using mount_smbfs on FreeBSD, then you may need to set `enable_smb1=True`

PR #3226, opened 2019-06-19
Bug fix for str schema

sonicaj:
This commit fixes an issue where time schema's validation would raise an exception as we would be checking length of datetime.time object. We convert the type to string before checking for length in str schema validation now.

### DIFF
--- a/src/middlewared/middlewared/schema.py
+++ b/src/middlewared/middlewared/schema.py
@@ -343,6 +343,9 @@ class Time(Str):
             except TypeError:
                 raise ValueError('Time should be in 24 hour format like "18:00"')
 
+    def validate(self, value):
+        return super().validate(str(value))
+
 
 class UnixPerm(Str):
 


src/middlewared/middlewared/schema.py
```diff
@@ -195,7 +195,7 @@ def validate(self, value):
 
         verrors = ValidationErrors()
 
-        if value and len(value) > self.max_length:
+        if value and len(str(value)) > self.max_length:
```
william-gr: What would this be, if not str?

sonicaj: `datetime.time` object in case of `Time` schema

william-gr: Almost seems like it should be handled in Time schema `validate` to keep things separate?

sonicaj: I am on fence wrt doing this. I'll proceed with doing this in Time schema per your recommendation :)

PR #3227, opened 2019-06-19
NAS-102191 / 11.2 / Do not start LDAP or AD on passive controller if system dataset not o…

anodos325:
…n boot device

When system dataset is on data pool in 11.2, we generate a stub config.
In this situation, the AD ctl script will fail to start and generate a UI alert about the passive controller.
Modify ctl scripts to check for this situation and return early.

### DIFF
--- a/src/freenas/etc/directoryservice/ActiveDirectory/ctl
+++ b/src/freenas/etc/directoryservice/ActiveDirectory/ctl
@@ -77,6 +77,17 @@ adctl_start()
 {
 	local cifs_started=0	
 	local ad_started=0
+	local syspool=$(eval echo $(${notifier} call systemdataset.config | /usr/local/bin/jq .pool))
+	local is_freenas=$(${notifier} call system.is_freenas)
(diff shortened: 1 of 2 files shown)


src/freenas/etc/directoryservice/ActiveDirectory/ctl
```diff
@@ -77,6 +77,17 @@ adctl_start()
...
 	local cifs_started=0	
 	local ad_started=0
+	local syspool=$(eval echo $(${notifier} call systemdataset.config | /usr/local/bin/jq .pool))
+	local is_freenas=$(${notifier} call system.is_freenas)
+	local failover_status='SINGLE'
+	if [ ${is_freenas} == 'False' ]; then
+		failover_status=$(${notifier} call notifier.failover_status)
+	fi
+
+	if [ \(${syspool} != "freenas-boot"\) -a \( ${failover_status} == "BACKUP" \) ]; then
```
william-gr: My shell script is weak, why the escaped parenthesis?

PR #3231, opened 2019-06-19
NAS-102203 / 11.3 / fix(restful): Catch IndexErrors

skarekrow:
We want these to be a 404 instead of a 500

NAS-102203

### DIFF
--- a/src/middlewared/middlewared/restful.py
+++ b/src/middlewared/middlewared/restful.py
@@ -12,7 +12,7 @@ from .client import ejson as json
 from .job import Job
 from .schema import Error as SchemaError
 from .service import CallError, ValidationError, ValidationErrors
-from .service_exception import adapt_exception
+from .service_exception import adapt_exception, MatchNotFound
 
 
 async def authenticate(middleware, req):
@@ -535,6 +535,12 @@ class Resource(object):
                     'errno': errno,
                 })
             resp = web.Response(status=422)
+        except MatchNotFound as e:
+            resp = web.Response(status=404)
+            result = {
+                'message': str(e),
+                'traceback': ''.join(traceback.format_exc()),
+            }
         except Exception as e:
             adapted = adapt_exception(e)
             if adapted:
--- a/src/middlewared/middlewared/restful.py
+++ b/src/middlewared/middlewared/restful.py
@@ -535,12 +535,7 @@ class Resource(object):
                     'errno': errno,
                 })
             resp = web.Response(status=422)
-        except MatchNotFound as e:
-            resp = web.Response(status=404)
-            result = {
-                'message': str(e),
-                'traceback': ''.join(traceback.format_exc()),
-            }
+
         except Exception as e:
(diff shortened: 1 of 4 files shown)


themylogin:
@william-gr it was my idea so I don't feel I can be reliable, can you look on this please too?

src/middlewared/middlewared/restful.py
```diff
@@ -535,6 +535,12 @@ def get_post_args(self, req, resp, **kwargs):
                     'errno': errno,
                 })
             resp = web.Response(status=422)
+        except MatchNotFound as e:
+            resp = web.Response(status=404)
+            result = {
+                'message': str(e),
+                'traceback': ''.join(traceback.format_exc()),
```
william-gr: We shouldnt put a traceback output for 404, the message is fine.

src/middlewared/middlewared/restful.py
```diff
@@ -535,6 +535,12 @@ def get_post_args(self, req, resp, **kwargs):
                     'errno': errno,
                 })
             resp = web.Response(status=422)
+        except MatchNotFound as e:
```
william-gr: I dont think MatchNotFound necessarily means the object was not found, we use get=True for other parts of the code, not only for the lookup of the specific object we are looking. e.g. we dont want 404 if for some reason MatchNotFound is raised on user.query (/api/v2.0/user/) so I think we still need a flag for the id lookup and handle this exception within Exception (`isinstance(e, MatchNotFound)`)

src/middlewared/middlewared/service_exception.py
```diff
@@ -101,3 +101,8 @@ def adapt_exception(e):
         output = ''.join([stdout, stderr]).rstrip()
 
         return CallError(f'Command {cmd} failed (code {e.returncode}):\n{output}')
+
+
+class MatchNotFound(IndexError):
+    """We be raised when a GET call is issued and there is no matching id"""
```
william-gr: That docstring may be misleanding, that is one of the use cases of that exception, but its generally used in filter list /datastore now.

src/middlewared/middlewared/utils/__init__.py
```diff
@@ -320,7 +321,10 @@ def filterop(f):
             rv = sorted(rv, key=lambda x: x[o], reverse=reverse)
 
     if options.get('get') is True:
-        return rv[0]
+        try:
+            return rv[0]
+        except IndexError:
+            raise MatchNotFound()
```
william-gr: filter_utils is not the only place it can raise IndexError, it can also happen in datastore.query

src/middlewared/middlewared/restful.py
themylogin: Good catch, have not thought about it. Probably rest controller should pass new option `get_or_404` and we should raise this new exception if only we got `get_or_404`

skarekrow:
Waiting to have a bigger discussion with William next week on the rest of this.

PR #3235, opened 2019-06-20
NAS-102190 / 11.3 / feat(middlewared/pool): translate a few zfs error codes and document it

william-gr:
Ticket:	NAS-102190

### DIFF
--- a/src/middlewared/middlewared/client/client.py
+++ b/src/middlewared/middlewared/client/client.py
@@ -19,6 +19,13 @@ import threading
 import time
 import uuid
 
+try:
+    from libzfs import Error as ZFSError
+except ImportError:
+    LIBZFS = False
+else:
+    LIBZFS = True
+
 
 class Event(TEvent):
 
@@ -273,6 +280,8 @@ class ErrnoMixin:
 
     @classmethod
     def _get_errname(cls, code):
+        if LIBZFS and 2000 <= code <= 2100:
+            return 'EZFS_' + ZFSError(code).name
         for k, v in cls.__dict__.items():
             if k.startswith("E") and v == code:
                 return k
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -1280,6 +1280,12 @@ class PoolService(CRUDService):
 
         `label` is the vdev guid or device name.
 
+        Error codes:
+
+            EZFS_NOSPC(2032): out of space to remove a device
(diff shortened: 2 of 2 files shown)


william-gr:
@themylogin i am not too attached to this solution but seemed a fair compromised into translating a few error codes and have them showing up ok without too much work. Let me know what you think please

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -64,6 +65,14 @@ def find_vdev(pool, vname):
         children += list(child.children)
 
 
+def libzfs_errcode_mapping(e):
+    return {
+        ZFSError.NOSPC: errno.ENOSPC,
+        ZFSError.NOREPLICAS: errno.ENOTRECOVERABLE,
+        ZFSError.NODEVICE: errno.ENOENT,
+    }.get(e.code) or e.code
```
themylogin: I don't like that we are mixing `errno` and `zfs_error` (which AFAIK do not match `errno`s), can it be `errno.EFAULT` by default instead?

william-gr: Its either that or have all errors specified in middleware (or they shown as EUNKNOWN) which I think its worse.

Defaulting to e.code at least let us identify what the original error code was and fix it up.

Thoughts?

themylogin: Can we make it `300 + e.code` for example? And unparse it back in `ErrnoMixin`

william-gr: I dont understand how that helps? e.code for ZFS are already in the 2000+ range.

themylogin: I want to isolate them from errno codes so they don't intersect. If they are already that way, fine.
But `ErrnoMixin` part seems useful to me nevertheless.

william-gr: Still not sure what you mean, are you talking about doing the inverse of `ZFSError.NODEVICE: errno.ENOENT` ?

And if so, why?

themylogin: I want to do this in `ErrnoMixin`:
```

    @classmethod
    def _get_errname(cls, code):
        if 2000 <= code <= 2100:
            return 'EZFS' + ZFSError(code).name    
```

william-gr: Ok, thats much smarter ;)

PR #3241, opened 2019-06-20
NAS-100161 / 11.3 / Default for UPS Shutdown command

sonicaj:
This PR sets a default for UPS shutdown command in case user hasn't supplied and system is not a truenas x series model, in other case where the system is a truenas x series model and database has no shutdown command, we let the command be empty.

### DIFF
--- a/src/middlewared/middlewared/plugins/ups.py
+++ b/src/middlewared/middlewared/plugins/ups.py
@@ -155,7 +155,7 @@ class UPSService(SystemServiceService):
             Str('port'),
             Str('remotehost'),
             Str('shutdown', enum=['LOWBATT', 'BATT']),
-            Str('shutdowncmd', empty=False),
+            Str('shutdowncmd', null=True),
             Str('subject'),
             List('toemail', items=[Str('email', validators=[Email()])]),
             update=True
--- a/gui/services/migrations/0033_ups_shutdowncmd.py
+++ b/gui/services/migrations/0033_ups_shutdowncmd.py
@@ -1,0 +1,33 @@
+from django.db import migrations, models
+
+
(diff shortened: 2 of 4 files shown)


src/middlewared/middlewared/plugins/ups.py
```diff
@@ -31,6 +31,11 @@ class Config:
         data['mode'] = data['mode'].upper()
         data['shutdown'] = data['shutdown'].upper()
         data['toemail'] = [v for v in data['toemail'].split(';') if v]
+        if not data['shutdowncmd'] and not (
+            not (await self.middleware.call('system.is_freenas')) and (
+                await self.middleware.call('truenas.get_chassis_hardware')).startswith('TRUENAS-X')
+        ):
+            data['shutdowncmd'] = '/sbin/shutdown -p now'
```
william-gr: That may give the impression shutdown -p was entered by the user, and the UI will use it to populate the form, this block should be moved to the upsmon.conf instead, I think.

sonicaj: But wouldn't that be kind of implicit ? User sees in the UI that there is no shutdown cmd default ( there is but he just doesn't know it yet ).

william-gr: We can have a tooltip in the UI or a placeholder, saying empty means shutdown -p.

If the user changes another value in the form it will also save the shutdown -p now command, which means if we ever want to change the default we wont be able to differentiate if it was entered by the user or its the system default.

sonicaj: Makes sense, i'll have it updated asap :)

PR #3243, opened 2019-06-20
NAS-101743 / 11.3 / fix(jail/setup): Supply interactive as False

skarekrow:
This will give us the correct exceptions instead of SystemExit.

Also remove the old checks for that.

NAS-101743


src/middlewared/middlewared/plugins/jail.py
```diff
@@ -20,6 +20,8 @@
 # iocage's imports are per command, these are just general facilities
 from iocage_lib.ioc_list import IOCList
 from iocage_lib.ioc_plugin import IOCPlugin
+from iocage_lib.ioc_common import set_interactive
```
sonicaj: We already have `ioc_common` imported, we can use that ;)

PR #3244, opened 2019-06-20
NAS-100292 / 11.3 / feat(vm/delete): Allow deleting DISKs

skarekrow:
This adds a bool to allow the UI to implement a checkbox for zvol cleanup during vm deletion.

NAS-100292

### DIFF
--- a/src/middlewared/middlewared/plugins/vm.py
+++ b/src/middlewared/middlewared/plugins/vm.py
@@ -944,23 +944,26 @@ class VMService(CRUDService):
         else:
             return update_devices
 
-    @accepts(Int('id'), Bool('disks', default=False))
-    async def do_delete(self, id, disks):
+    @accepts(Int('id'), Bool('zvols', default=False))
+    async def do_delete(self, id, zvols):
         """Delete a VM."""
         status = await self.status(id)
         if isinstance(status, dict):
             if status.get('state') == 'RUNNING':
                 await self.stop(id)
-        if disks:
+        if zvols:
             devices = await self.middleware.call('vm.device.query', [
                 ('vm', '=', id), ('dtype', '=', 'DISK')
             ])
 
-            for device in devices:
+            for zvol in devices:
+                if not device['attributes']['path'].startswith('/dev/zvol/'):
+                    continue
+
                 disk_name = device['attributes']['path'].rsplit(
                     '/dev/zvol/'
                 )[-1]
-                await self.middleware.call('zfs.dataset.delete', disk_name)
+                await self.middleware.call('pool.dataset.delete', disk_name)
         return await self.middleware.call('datastore.delete', 'vm.vm', id)
 
     @item_method
--- a/src/middlewared/middlewared/plugins/vm.py
+++ b/src/middlewared/middlewared/plugins/vm.py
@@ -944,14 +944,20 @@ class VMService(CRUDService):
         else:
             return update_devices
 
-    @accepts(Int('id'), Bool('zvols', default=False))
-    async def do_delete(self, id, zvols):
+    @accepts(
+        Int('id'),
+        Dict(
+            'vm_delete',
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/vm.py
```diff
@@ -929,13 +929,24 @@ def random_mac(self):
         else:
             return update_devices
 
-    @accepts(Int('id'))
-    async def do_delete(self, id):
+    @accepts(Int('id'), Bool('disks'))
+    async def do_delete(self, id, disks):
```
sonicaj: We should have a reasonable default value here

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -929,13 +929,24 @@ def random_mac(self):
...
         """Delete a VM."""
         status = await self.status(id)
         if isinstance(status, dict):
             if status.get('state') == 'RUNNING':
                 await self.stop(id)
+        if disks:
+            devices = await self.middleware.call('vm.device.query', [
+                ('vm', '=', id)
```
sonicaj: Also we can use a filter for `dtype` ;)

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -929,13 +929,23 @@ def random_mac(self):
...
             if status.get('state') == 'RUNNING':
                 await self.stop(id)
+        if disks:
+            devices = await self.middleware.call('vm.device.query', [
+                ('vm', '=', id), ('dtype', '=', 'DISK')
+            ])
+
+            for device in devices:
+                disk_name = device['attributes']['path'].rsplit(
+                    '/dev/zvol/'
```
william-gr: There is no guarantee it is a zvol, I think. In theory it could be a disk device. Need to make sure its a zvol first.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -929,13 +929,23 @@ def random_mac(self):
...
+                )[-1]
+                await self.middleware.call('zfs.dataset.delete', disk_name)
```
themylogin: `pool.dataset.delete` should be used here.

themylogin: `pool.dataset.delete` should be used here.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -929,23 +929,26 @@ def random_mac(self):
...
-            for device in devices:
+            for zvol in devices:
+                if not device['attributes']['path'].startswith('/dev/zvol/'):
+                    continue
+
                 disk_name = device['attributes']['path'].rsplit(
                     '/dev/zvol/'
                 )[-1]
-                await self.middleware.call('zfs.dataset.delete', disk_name)
+                await self.middleware.call('pool.dataset.delete', disk_name)
```
themylogin: It won't work, this method checks for dataset dependencies and it has at least one (VM). You need to remove from vm.device first, then call this method, if it raises CallError, then it means other attachments exist (e.g. iSCSI) and what to do in this case is undefined.

william-gr: How about we just use zfs.dataset.delete for now?
If the user is using the same zvol for another VM or iSCSI (both highly unadvised), we ignored it.

May revisit this if it becomes an issue in the future?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -929,13 +929,26 @@ def random_mac(self):
         else:
             return update_devices
 
-    @accepts(Int('id'))
-    async def do_delete(self, id):
+    @accepts(Int('id'), Bool('zvols', default=False))
```
william-gr: I would turn this into a Dict with zvols inside, so we can easily add new options later.

PR #3251, opened 2019-06-21
NAS-101249 / 11.3 / Bug fix for spaces in path names for fstab

sonicaj:
This commit fixes an issue where if spaces were defined in src/destination paths, we could not cleanly interpret where to split.

### DIFF
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -949,7 +949,7 @@ class JailService(CRUDService):
             )
 
             for i in _list:
-                fstab_entry = i[1].split()
+                fstab_entry = i[1]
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/jail.py
```diff
@@ -949,7 +949,7 @@ def fstab(self, jail, options):
             )
 
             for i in _list:
-                fstab_entry = i[1].split()
+                fstab_entry = i[1]
```
themylogin: `fstab_entry` changes type from `list` to `str`???

sonicaj: So with latest `iocage` changes `i[1]` is a list, and in older version, `fstab_entry` is changing `str` to list

PR #3254, opened 2019-06-21
NAS-102235 / 11.3 / Make smb.getparm public and standardize boolean formatting in smb4.conf

anodos325:
- if boolean value is in base smbd True or False
- if boolean valie is in a vfs module Yes or No

smb.getparm will return boolean type for the former and string type for the latter.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -442,12 +442,17 @@ class SMBService(SystemServiceService):
                     if deluser.returncode != 0:
                         raise CallError(f'Failed to delete user {entry["username"]}: {deluser.stderr.decode()}')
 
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -442,12 +442,17 @@ def check_codec(encoding):
                     if deluser.returncode != 0:
                         raise CallError(f'Failed to delete user {entry["username"]}: {deluser.stderr.decode()}')
 
-    @private
+    @accepts(
+        Str('parm', required=True),
+        Str('section', default='GLOBAL'),
+    )
     def getparm(self, parm, section):
```
william-gr: What is the use-case for having this API being public?

PR #3255, opened 2019-06-21
NAS-102238 / 11.2 / OneDrive 11.2 Workaround Tooltip

themylogin:


gui/freeadmin/static/lib/js/freeadmin.js
```diff
@@ -1683,6 +1683,10 @@ require([
                     }
                     input += '</select>';
                 }
+                if (provider == "ONEDRIVE" && property.property == "drive_id")
+                {
+                    input += '<br>Copy this value from address bar:<br><a href="https://onedrive.live.com" target="_blank">https://onedrive.live.com/</a>?id=root&cid=<b>XXX</b>';
```
Mrt134: If I understand this correctly, we want users to click the link and log in to their onedrive account, then copy the "XXX" value that appears after `?id=root&cid=` from the address bar. Writing a rough suggestion to clarify the whole procedure:

`Log in to the <a href="https://onedrive.live.com" target="_blank">OneDrive account</a>, copy the string after <b>cid=</b> from the browser address bar, and enter that value as the Drive ID.`

PR #3259, opened 2019-06-24
NAS-102237 / 11.3 / Expose method to restart UI

sonicaj:
This commit exposes a method in system.general service which can be used to restart UI after changes in UI have been made like cert changes etc. The method restarts nginx and django as soon as it is called.

### DIFF
--- a/src/middlewared/middlewared/plugins/system.py
+++ b/src/middlewared/middlewared/plugins/system.py
@@ -1003,6 +1003,14 @@ class SystemGeneralService(ConfigService):
         return await self.config()
 
     @accepts()
+    async def ui_restart(self):
+        """
+        Restart HTTP server to use latest UI settings.
+        """
+        await self.middleware.call('service.restart', 'http')
+        await self.middleware.call('service.restart', 'django')
+
+    @accepts()
     async def local_url(self):
         """
         Returns configured local url in the format of protocol://host:port


src/middlewared/middlewared/plugins/system.py
```diff
@@ -1002,6 +1002,14 @@ def _get_index(country_columns, column):
 
         return await self.config()
 
+    @accepts()
+    async def ui_restart(self):
+        """
+        Restart nginx server and django to reflect latest UI settings.
```
william-gr: Lets not mention technologies. Just say HTTP server.

a-stjohn:
```suggestion
        Restart nginx server and django to show latest UI settings.
```

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1002,6 +1002,14 @@ def _get_index(country_columns, column):
 
         return await self.config()
 
+    @accepts()
+    async def ui_restart(self):
+        """
+        Restart HTTP server to show latest UI settings.
```
william-gr: to show seems wrong here. `to use` maybe? There is nothing to show

sonicaj: @freenas/docs-team please advise

sonicaj: And i agree @william-gr

a-stjohn: `to use` works.

wblock:
It really means "So UI settings changes take effect."

PR #3260, opened 2019-06-24
NAS-102256 / 11.3 / Only expose static addresses as choices in SMB Bind IP Addresses

anodos325:


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -99,6 +99,14 @@ def key_cp(encoding):
             ] + initial
         }
 
+    @private
+    async def interface_choices(self):
```
william-gr: If this is to be used in smb.update we should name it `bindip_choices` and make it public, with a docstring.

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -549,6 +561,12 @@ def getparm(self, parm, section):
         if new['netbiosname'] and new['netbiosname'].lower() == new['workgroup'].lower():
             verrors.add('smb_update.netbiosname', 'NetBIOS and Workgroup must be unique')
 
+        if data['bindip']:
+            bindip_choices = [x[0] for x in (await self.bindip_choices()).items()]
```
william-gr: bindip_choices = list((await self.bindip_choices()).keys())

is simpler ; )

anodos325: Yeah, I'm having a Monday. :)

PR #3264, opened 2019-06-24
NAS-102258 / 11.3 / Update 'guest' help text for SMB shares

anodos325:
- Recent versions of windows may not allow guest SMB2 sessions by
  default.

### DIFF
--- a/gui/sharing/models.py
+++ b/gui/sharing/models.py
@@ -85,7 +85,10 @@ class CIFS_Share(Model):
         verbose_name=_('Allow Guest Access'),
         help_text=_(
             'If true then no password is required to connect to the share. '
-            'Privileges will be those of the guest account.'
+            'Privileges will be those of the guest account. Guest access may '
+            'be disabled by default in Windows 10 and Windows Server starting with '
+            'versions 1709 and 1903. Additional client-side configuration may be '
+            'required to provide guest access to these clients. '
         ),
         default=False,
     )
--- a/gui/sharing/models.py
+++ b/gui/sharing/models.py
@@ -86,7 +86,7 @@ class CIFS_Share(Model):
         help_text=_(
             'If true then no password is required to connect to the share. '
             'Privileges are the same as the guest account. Guest access is '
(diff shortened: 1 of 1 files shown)


gui/sharing/models.py
```diff
@@ -85,7 +85,10 @@ class CIFS_Share(Model):
         verbose_name=_('Allow Guest Access'),
         help_text=_(
             'If true then no password is required to connect to the share. '
-            'Privileges will be those of the guest account.'
+            'Privileges will be those of the guest account. Guest access may '
```
a-stjohn: This particular sentence might be worded strangely to a user. Perhaps:

```suggestion
            'Privileges belong to the guest account. Guest access may '
```

wblock: I think it means that privileges will be the same as the guest account.
"may" is the wrong word to use here, it implies permission.  "might" is clearer.  But if it happens every time with those Windows versions, it should just be "is":
```suggestion
            'Privileges are the same as the guest account. Guest access is '
```

gui/sharing/models.py
```diff
@@ -85,7 +85,10 @@ class CIFS_Share(Model):
...
+            'be disabled by default in Windows 10 and Windows Server starting with '
```
wblock:
```suggestion
            'disabled by default in Windows 10 version 1709 and Windows Server version 1903. '
```

gui/sharing/models.py
```diff
@@ -85,7 +85,10 @@ class CIFS_Share(Model):
...
+            'versions 1709 and 1903. Additional client-side configuration may be '
```
wblock:
```suggestion
            'Additional client-side configuration is required to provide '
```

gui/sharing/models.py
```diff
@@ -85,7 +85,10 @@ class CIFS_Share(Model):
...
+            'required to provide guest access to these clients. '
```
wblock:
```suggestion
            'guest access to these clients. See <url with instructions>.'
```

PR #3265, opened 2019-06-24
QE NAS-102255 Fix API v2 jail testing and update jail/list_resource testing

ericbsd:


tests/api2/jail.py
```diff
@@ -30,20 +30,45 @@ def test_02_verify_iocage_pool():
...
+
+
+def test_05_verify_available_bsd_release():
+    global freeze, RELEASE
+    freeze = False
+    global freeze_msg
+    stop_time = time.time() + 600
+    while True:
+        get_job = GET(f'/core/get_jobs/?id={JOB_ID}')
+        job_status = get_job.json()[0]
```
sonicaj: I think it might be better to assert the status code before grabbing `json`

tests/api2/jail.py
```diff
@@ -30,20 +30,45 @@ def test_02_verify_iocage_pool():
     assert results.json() == IOCAGE_POOL, results.text
 
 
-def test_03_verify_list_resources_endpoint():
+def test_03_verify_list_resources_is_and_integer():
```
sonicaj: `is_an_integer` perhaps ?

tests/api2/jail.py
```diff
@@ -30,20 +30,45 @@ def test_02_verify_iocage_pool():
...
+
+def test_05_verify_available_bsd_release():
+    global freeze, RELEASE
+    freeze = False
+    global freeze_msg
+    stop_time = time.time() + 600
+    while True:
+        get_job = GET(f'/core/get_jobs/?id={JOB_ID}')
+        job_status = get_job.json()[0]
+        if job_status['state'] in ('RUNNING', 'WAITING'):
```
sonicaj: Not necessary but do you think it might be better to have waiting for job as a common method somewhere and use it ? So if in future if any changes are introduced, we wouldn't be changing this in every tests ;)

ericbsd:
cool I will do the change

PR #3273, opened 2019-06-26
NAS-102281 / 11.3 / Expose destroying jail force flag

sonicaj:
This commit exposes a force flag for destroying a jail in jail plugin.

### DIFF
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -405,13 +405,19 @@ class JailService(CRUDService):
 
         return verrors
 
-    @accepts(Str("jail"))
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/jail.py
```diff
@@ -405,13 +405,13 @@ def common_validation(self, verrors, options, update=False, jail=None):
 
         return verrors
 
-    @accepts(Str("jail"))
-    def do_delete(self, jail):
+    @accepts(Str('jail'), Bool('force', default=False))
```
william-gr: I would rather we have a Dict() here with force in it.

Best for future backward compatibility.

PR #3278, opened 2019-06-26
NAS-100718 / 11.3 / Misc improvements to mDNS plugin

anodos325:
- Move service mDNS advertisement to middleware
- Properly configure mixed AFP and SMB Time Machine shares. This will allow protocol failover from SMB to AFP.
- Bring in afp, ftp, iscsi, smb advertisement to middleware
- Persistently store VUID for Time Machine over AFP.
- Add ability to selectively restart mDNS advertisement for services (less disruptive than restarting everything).
- Add ability to which services we're currently advertising.
- Only advertise on Active Controller, and use virtual hostname on HA servers

### DIFF
--- a/src/middlewared/middlewared/plugins/afp.py
+++ b/src/middlewared/middlewared/plugins/afp.py
@@ -1,4 +1,5 @@
 import asyncio
+import uuid
 
 from middlewared.async_validators import check_path_resides_within_volume
 from middlewared.common.attachment import FSAttachmentDelegate
@@ -109,6 +110,7 @@ class SharingAFPService(CRUDService):
         UnixPerm('umask', default='000'),
         List('hostsallow', items=[], default=[]),
         List('hostsdeny', items=[], default=[]),
+        Str('vuid', null=True, default=''),
         Str('auxparams', max_length=None),
         Bool('enabled', default=True),
         register=True
@@ -214,6 +216,11 @@ class SharingAFPService(CRUDService):
     @private
     async def validate(self, data, schema_name, verrors, old=None):
         await self.home_exists(data['home'], schema_name, verrors, old)
+        if data['vuid']:
+            try:
+                uuid.UUID(data['vuid'], version=4)
+            except ValueError:
+                verrors.add(f'{schema_name}.vuid', 'vuid must be a valid UUID.')
 
     @private
     async def home_exists(self, home, schema_name, verrors, old=None):
@@ -291,7 +298,8 @@ class SharingAFPService(CRUDService):
         data['rw'] = ' '.join(data['rw'])
         data['hostsallow'] = ' '.join(data['hostsallow'])
         data['hostsdeny'] = ' '.join(data['hostsdeny'])
-
+        if not data['vuid'] and data['timemachine']:
+            data['vuid'] = str(uuid.uuid4())
         return data
 
 
--- a/src/middlewared/middlewared/plugins/mdns.py
+++ b/src/middlewared/middlewared/plugins/mdns.py
@@ -120,31 +120,6 @@ class mDNSDaemonMonitor(threading.Thread):
         return p.returncode == 0
 
 
-class mDNSServiceObject(object):
-    def __init__(self, **kwargs):
-        self.sdRef = kwargs.get('sdRef')
-        self.txtrecord = kwargs.get('txtrecord')
-        self.port = kwargs.get('port')
-        self.regtype = kwargs.get('regtype')
-        self.name = kwargs.get('name')
-
-    @property
-    def fullname(self):
-        return "%s.%s.%s" % (
-            self.name.strip('.'),
-            self.regtype.strip('.'),
-        )
-
-    def to_dict(self):
-        return {
-            'type': 'mDNSSerivceObject',
-            'regtype': self.regtype,
-            'domain': self.domain,
-            'txtrecord': self.txtrecord,
-            'port': self.port
-        }
-
-
 class mDNSServiceThread(threading.Thread):
     def __init__(self, **kwargs):
         super(mDNSServiceThread, self).__init__()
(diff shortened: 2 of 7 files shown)


src/middlewared/middlewared/plugins/afp.py
```diff
@@ -109,6 +110,7 @@ class Config:
         UnixPerm('umask', default='000'),
         List('hostsallow', items=[], default=[]),
         List('hostsdeny', items=[], default=[]),
+        Str('vuid', null=True, default=''),
```
william-gr: Do we need to expose vuid?

src/middlewared/middlewared/plugins/mdns.py
```diff
@@ -1,14 +1,62 @@
...
+    FTP = 'mDNSServiceFTPThread'
+    HTTP = 'mDNSServiceHTTPThread'
+    HTTPS = 'mDNSServiceHTTPSThread'
+    ISCSI = 'mDNSServiceISCSIThread'
+    MIDDLEWARE = 'mDNSServiceMiddlewareThread'
+    MIDDLEWARE_SSL = 'mDNSServiceMiddlewareSSLThread'
+    NFS = 'mDNSServiceNFSThread'
+    SSH = 'mDNSServiceSSHThread'
+    SFTP_SSH = 'mDNSServiceSFTP_SSHThread'
+    SMB = 'mDNSServiceSMBThread'
```
william-gr: Is it a requirement to have a thread for each service?
If not, would it be too much work to have them in the same?

src/middlewared/middlewared/plugins/mdns.py
```diff
@@ -231,6 +392,24 @@ def __init__(self, *args, **kwargs):
...
+    @private
+    async def get_hostname(self):
+        """
+        Return virtual hostname if the server is HA and this is the active controller.
+        Return None if this is the passive controller.
+        In all other cases return the hostname_local.
+        This is to ensure that the correct hostname is used for mdns advertisements.
+        """
+        ngc = await self.middleware.call('network.configuration.config')
+        if not await self.middleware.call('system.is_freenas') and await self.middleware.call('notfier.failover_licensed'):
```
william-gr: If think this check can be removed if we check for the existence of `hostname_virtual` in ngc?

src/middlewared/middlewared/plugins/mdns.py
```diff
@@ -231,6 +392,24 @@ def __init__(self, *args, **kwargs):
...
+            failover_status = await self.middleware.call('notifier.failover_status')
```
william-gr: In 11.3 we use failover.status

src/middlewared/middlewared/plugins/mdns.py
anodos325: Yes. Good point

src/middlewared/middlewared/plugins/afp.py
anodos325: Probably not, but we do it for SMB. It's nice to have the consistency. An advanced user may wish to convert a time machine share from AFP to SMB. In this case being able to transfer the VUID is useful.

src/middlewared/middlewared/plugins/mdns.py
```diff
@@ -89,49 +120,201 @@ def start_mdnsd(self):
...
+        self.txtrecord = kwargs.get('txtrecord')
+        self.port = kwargs.get('port')
+        self.regtype = kwargs.get('regtype')
+        self.name = kwargs.get('name')
 
-    def active(self, sdRef):
-        return (bool(sdRef) and sdRef.fileno() != -1)
+    @property
+    def fullname(self):
+        return "%s.%s.%s" % (
```
william-gr: There are three %s and only two args.
Also lets use .format() or f-string please

src/middlewared/middlewared/plugins/mdns.py
```diff
@@ -89,49 +120,201 @@ def start_mdnsd(self):
...
+            mDNSServices[srv.name].sdRef = pybonjour.DNSServiceRegister(
+                name=self.hostname,
+                regtype=self.regtype,
+                port=port,
+                txtRecord=txtrecord,
+                callBack=None
+            )
+
+        self.finished.wait()
+        self.logger.trace("we're finished")
```
william-gr: Is this a debug left-over or intended?

src/middlewared/middlewared/plugins/mdns.py
```diff
@@ -240,36 +363,36 @@ def start(self):
...
-        for thread in self.threads.copy():
-            thread = self.threads.get(thread)
-            thread.cancel()
-            del self.threads[thread.service]
+        thread = self.threads.get('mDNSThread')
+        if thread is None:
+            return
+
+        thread.cancel()
+        del self.threads['mDNSThread']
```
william-gr: That seems unnecessary if we are setting threads below. The object will be garbage collected anyway.

src/middlewared/middlewared/plugins/service.py
```diff
@@ -836,14 +837,17 @@ def _initializing_smartd_pid(self):
         # restarting netatalk.
         await self._system("pkill -9 afpd")
         await self._system("pkill -9 cnid_metad")
+        await self.middleware.call('mdnsadvertise.restart')
 
     async def _restart_afp(self, **kwargs):
         await self._stop_afp()
         await self._start_afp()
+        await self.middleware.call('mdnsadvertise.restart')
```
william-gr: this is already called on _start_afp

src/middlewared/middlewared/plugins/mdns.py
anodos325: I'll do better and replace where I was using this with a simple dictionary.

src/middlewared/middlewared/plugins/mdns.py
anodos325: Removed.

src/middlewared/middlewared/plugins/service.py
anodos325: Good observation. Removed.

PR #3280, opened 2019-06-26
NAS-100224 / 11.3 / Update legacy UI to use "TrueNAS Controller 1/2" instead of "Node A/B"

themylogin:


src/middlewared/middlewared/plugins/alert.py
```diff
@@ -53,8 +53,8 @@ class AlertSourceRunFailedAlertClass(AlertClass):
 class AlertSourceRunFailedOnBackupNodeAlertClass(AlertClass):
     category = AlertCategory.SYSTEM
     level = AlertLevel.CRITICAL
-    title = "Alert Check Failed (Backup Node)"
-    text = "Failed to check for alert %(source_name)s on backup node:\n%(traceback)s"
+    title = "Alert Check Failed (Backup Controller)"
```
william-gr: The word is that we will also call it Passive Controller, not Backup

william-gr: Is that right, @wblock ?

william-gr: @freenas/docs-team ?

wblock: Unfortunately, no.  After the UI meeting, I brought this up to the Product Management meeting, and was told that Active/Passive is wrong, and we have always been Active/Standby.  I am supposed to research this, but that will have to be a separate set of tickets.

PR #3281, opened 2019-06-27
NAS-102308 / 11.3 / Only start nut if ups is configured in master mode

sonicaj:
When UPS is configured in slave mode, we should not start nut service at all as it is not used and results in erroneous messages saying relevant conf files don't exist.

### DIFF
--- a/src/middlewared/middlewared/plugins/service.py
+++ b/src/middlewared/middlewared/plugins/service.py
@@ -777,7 +777,8 @@ class ServiceService(CRUDService):
     async def _start_ups(self, **kwargs):
         await self.middleware.call('ups.dismiss_alerts')
         await self.middleware.call('etc.generate', 'ups')
-        await self._service("nut", "start", **kwargs)
+        if (await self.middleware.call('ups.config'))['mode'] == 'MASTER':
+            await self._service("nut", "start", **kwargs)
         await self._service("nut_upsmon", "start", **kwargs)
         await self._service("nut_upslog", "start", **kwargs)
         if await self.started('collectd'):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/service.py
```diff
@@ -787,14 +788,17 @@ def _initializing_smartd_pid(self):
         await self.middleware.call('ups.dismiss_alerts')
         await self._service("nut_upslog", "stop", force=True, **kwargs)
         await self._service("nut_upsmon", "stop", force=True, **kwargs)
-        await self._service("nut", "stop", force=True, **kwargs)
+        if (await self.middleware.call('ups.config'))['mode'] == 'MASTER':
```
william-gr: We may want to stop nut regardless? Perhaps it was MASTER and it isnt anymore?

src/middlewared/middlewared/plugins/service.py
```diff
@@ -787,14 +788,17 @@ def _initializing_smartd_pid(self):
...
+            await self._service("nut", "stop", force=True, **kwargs)
         if await self.started('collectd'):
             asyncio.ensure_future(self.restart('collectd'))
 
     async def _restart_ups(self, **kwargs):
+        master_mode = (await self.middleware.call('ups.config'))['mode'] == 'MASTER'
         await self.middleware.call('ups.dismiss_alerts')
         await self.middleware.call('etc.generate', 'ups')
-        await self._service("nut", "stop", force=True, onetime=True)
+        if master_mode:
```
william-gr: Same here?

PR #3284, opened 2019-06-28
NAS-102323 / 11.2 / Do not disable AD or LDAP on passive controller in unified config

anodos325:
If AD is enabled, but there is a stub smb4.conf on the passive controller, then
if ix-activedirectory start will fail and disable the AD service. This will may
cause the passive controller to cease to renew kerberos tickets and adversely affect
SMB services on failover.

In all AD and LDAP related RC scripts, check if the controller is standby. If it is,
then immediately return. This is to avoid inadvertently disabling the
AD/LDAP service on the passive during boot or normal operations.

### DIFF
--- a/src/freenas/etc/ix.rc.d/ix-activedirectory
+++ b/src/freenas/etc/ix.rc.d/ix-activedirectory
@@ -178,6 +178,11 @@ setup_homedirs()
 
 activedirectory_start()
 {
+	if is_unified_standby; then
+                echo got here
+		return 0
+	fi
+        return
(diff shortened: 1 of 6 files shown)


src/freenas/etc/ix.rc.d/ix-activedirectory
```diff
@@ -178,6 +178,11 @@ setup_homedirs()
 
 activedirectory_start()
 {
+	if is_unified_standby; then
+                echo got here
```
william-gr: left over debug

PR #3286, opened 2019-06-29
NAS-102286 / 11.3 / Multiple Repository Support For Plugins

sonicaj:
This PR introduces following changes making sure we support multiple repositories:

1) New Service to manage plugins i.e creating, listing installed/available etc.
2) New API for listing branches of a github repository.
3) New API for listing available/downloaded releases.
4) New Plugin service now allows multiple repository support.
5) Jail names are now expected when creating a plugin and the api/iocage will not do it automatically from middlewared's end.
6) Backward compatibility with jail.fetch method has been introduced which means plugins can still be created with jail.fetch but if multiple repository support is required, new plugin service should be used.

It does not introduce new api for listing templates as that can already be done by querying jail service with a filter.


src/middlewared/middlewared/plugins/jail.py
```diff
@@ -117,6 +117,168 @@ def query(self, filters=None, options=None):
 
         return filter_list(jails, filters, options)
 
+    @accepts()
```
william-gr: It almost seems like we need a new service called "plugin" to separate the logic?

Could be in the same file.

Thoughts?

sonicaj: @william-gr i couldn't agree more. I was afraid of breaking the API too much this way. `fetch` too is overloaded and having a new service will help us manage that too in a better way. Please advise if you think i should move ahead with this and update the PR.

william-gr: Yes, lets go ahead. We can try to somehow keep backward compatibility if at all possible, but currently there is too much confusion in the API with jails vs plugins

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -36,6 +36,430 @@
...
+    @accepts(
+        Dict(
+            'plugin_create',
+            Str('plugin_name', required=True),
+            Str('jail_name', required=True),
+            List('props', default=[], empty=False),
+            Str('branch', default=None, null=True),
+            Str('plugin_repository', default='https://github.com/freenas/iocage-ix-plugins.git'),
+        )
+    )
```
william-gr: `@job` doesnt work here? It should

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -36,6 +36,430 @@
...
+    @job(
+        lock=lambda args: 'available_plugins_{}_{}'.format(
+            args[-1].get('branch') if args else None,
+            args[-1]['plugin_repository'] if args and args[-1].get(
+                'plugin_repository'
+            ) else 'https://github.com/freenas/iocage-ix-plugins.git'
+        )
+    )
+    def available_plugins(self, job, options):
```
william-gr: Just `available` would be OK I think.

We are already in plugins namespace.

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -117,6 +580,45 @@ def query(self, filters=None, options=None):
...
+    @accepts(
+        Dict(
+            'branches_choices',
+            Str('account', default='freenas'),
+            Str('repository', default='iocage-ix-plugins'),
+        )
+    )
+    def branches_choices(self, options):
+        """
+        List branches for a github `account`'s `repository`.
```
william-gr: Not a fan of relying on github for that. Is there anything we can do about that?

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -36,6 +36,469 @@
...
+            try:
+                for mac in value.split():
+                    MACAddr()(mac)
+
+                if len(value.split()) != 2 or any(value.split().count(v) > 1 for v in value.split()):
+                    raise ValueError('Exception')
+            except ValueError:
+                verrors.add(
+                    f'{schema}.{key}',
+                    'Please Enter two valid and different '
```
Mrt134:
```suggestion
                    'Please enter two valid and different '
```

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -36,6 +36,469 @@
...
+            List('props', default=[], empty=False),
+            Str('branch', default=None, null=True),
+            Str('plugin_repository', default='https://github.com/freenas/iocage-ix-plugins.git'),
+        )
+    )
+    def do_create(self, data):
+        """
+        Create a Plugin.
+
+        `plugin_name` is the name of the plugin as specified by the INDEX file in "plugin_repository" and it's json
```
Mrt134:
```suggestion
        `plugin_name` is the name of the plugin specified by the INDEX file in "plugin_repository" and its JSON
```

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -36,6 +36,469 @@
...
+        file.
+
+        `jail_name` is the name of the plugin jail and is required.
```
Mrt134:
```suggestion
        `jail_name` is the name of the jail that will manage the plugin. Required.
```

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -36,6 +36,469 @@
...
+
+        `props` is a list of iocage props which user can manually set. For plugins, it should be ensured that the
```
Mrt134:
```suggestion
        `props` is a list of jail properties that the user manually sets. Plugins should always set the
```

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -36,6 +36,469 @@
...
+        jails have networking capability set either with dhcp, ip address or nat.
```
Mrt134:
```suggestion
        jail networking capability with DHCP, IP Address, or NAT properties.
```
Might be helpful to have an example (dhcp=on, ip4_addr="em0|192.168.0.1/24", etc) or link to iocage manpage?

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -36,6 +36,469 @@
...
+
+        `plugin_repository` is a git uri which will be used to fetch `plugin_name`'s data.
```
Mrt134:
```suggestion
        `plugin_repository` is a git URI that fetches data for `plugin_name`.
```

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -36,6 +36,469 @@
...
+
+        `branch` is the branch to use for `plugin_repository`. It defaults to system's version. i.e 11.3-RELEASE.
```
Mrt134:
```suggestion
        `branch` is the FreeNAS repository branch to use as the base for the `plugin_repository`. The default is to use the current system version. Examples: 11.3-RELEASE, 11.3-U1, 11.3-U2.
```
Suggested explaining the term and behavior a little more.

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -24,18 +24,483 @@
...
+        """
+        Update plugin `jail`.
+        """
+        await self._get_instance(jail)
+        return await self.middleware.call('jail.update', jail, data)
+
+    @accepts(Str('id'))
+    async def do_delete(self, id):
+        """
+        Delete plugin `jail`.
```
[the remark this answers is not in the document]
sonicaj: Updated :)

PR #3293, opened 2019-07-02
NAS-101010 / 12 / Allow disabling HTTPS protocols

themylogin:


src/middlewared/middlewared/plugins/system.py
```diff
@@ -569,6 +569,10 @@ def __init__(self, *args, **kwargs):
 
         return data
 
+    @accepts()
+    def https_protocols_choices(self):
```
william-gr: `ui_httpsprotocols_choices`

It should also return a dict since thats the default we use for choices every where else (or almost).

Also needs a docstring

src/middlewared/middlewared/plugins/system.py
```diff
@@ -534,6 +534,7 @@ def debug_download(self, job):
 
 
 class SystemGeneralService(ConfigService):
+    HTTPS_PROTOCOLS = ['TLSv1', 'TLSv1.1', 'TLSv1.2']
```
william-gr: According to nginx TLSv1.3 is already a possibility, why are we not allowing it?

gui/system/forms.py
```diff
@@ -1007,6 +1012,10 @@ def __init__(self, *args, **kwargs):
         super(SettingsForm, self).__init__(*args, **kwargs)
         self.original_instance = dict(self.instance.__dict__)
 
+        with client as c:
+            self.fields['stg_guihttpsprotocols'].choices = [
+                (p, p) for p in c.call('system.general.https_protocols_choices')
```
william-gr: Update here to catch up?

PR #3296, opened 2019-07-03
NAS-102387 / 11.3 / Use account plugin for user/group data retrieval

sonicaj:
This commit removes old usages of get_user_object method in notifier plugin and instead now uses account plugin to retrieve relevant details.

### DIFF
--- a/src/middlewared/middlewared/plugins/cron.py
+++ b/src/middlewared/middlewared/plugins/cron.py
@@ -1,3 +1,5 @@
+import contextlib
+
 from middlewared.schema import accepts, Bool, Cron, Dict, Int, Patch, Str
 from middlewared.service import CRUDService, job, private, ValidationErrors
 from middlewared.validators import Range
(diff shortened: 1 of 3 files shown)


william-gr:
@anodos325 do we need to skip the cache entirely here? and use get_uncached?

anodos325:
> @anodos325 do we need to skip the cache entirely here? and use get_uncached?

I think it makes sense to bypass the cache and get_uncached. 

This will return the following:
```
root@freenas[/mnt/Z/src/middlewared]# midclt call dscache.get_uncached_user root
{"pw_name": "root", "pw_uid": 0, "pw_gid": 0, "pw_gecos": "root", "pw_dir": "/root", "pw_shell": "/usr/local/bin/zsh"}
root@freenas[/mnt/Z/src/middlewared]# midclt call dscache.get_uncached_group wheel
{"gr_name": "wheel", "gr_gid": 0, "gr_mem": []}
```

src/middlewared/middlewared/plugins/cron.py
```diff
@@ -47,16 +49,18 @@ def cron_extend(self, data):
...
-                verrors.add(
-                    f'{schema}.user',
-                    'Specified user does not exist'
-                )
+            else:
+                user_data = None
+                with contextlib.suppress(KeyError):
+                    user_data = await self.middleware.call(
+                        'user.query', [['username', '=', user]]
+                    ) or await self.middleware.call('dscache.get_uncached_user', user)
```
william-gr: Why use both methods and not just dscache.get_uncached_user?

PR #3297, opened 2019-07-03
NAS-102386 / 11.3 / Fix key error in smb plugin.

anodos325:
Regression caused by changes related to bindip in smb plugin.


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -561,7 +561,7 @@ def getparm(self, parm, section):
         if new['netbiosname'] and new['netbiosname'].lower() == new['workgroup'].lower():
             verrors.add('smb_update.netbiosname', 'NetBIOS and Workgroup must be unique')
 
-        if data['bindip']:
+        if 'bindip' in data and data['bindip']:
```
william-gr: if data.get('bindip'):

will avoid you one lookup, just for future reference

PR #3301, opened 2019-07-03
QE NAS-102165 Adding jail plugins API testing for 11.3

ericbsd:


tests/api2/jail_plugins.py
```diff
@@ -0,0 +1,319 @@
...
+
+@to_skip
+def test_02_verify_jail_pool():
+    results = GET('/jail/get_activated_pool/')
+    assert results.status_code == 200, results.text
+    assert results.json() == IOCAGE_POOL, results.text
+
+
+@to_skip
+def test_03_get_list_of_instaled_plugin_job_id():
```
sonicaj: typo

tests/api2/jail_plugins.py
```diff
@@ -0,0 +1,291 @@
...
+        else:
+            assert job_state == 'SUCCESS', str(job_results)
+            break
+
+
+@to_skip
+@pytest.mark.parametrize('plugin', plugins_list)
+def test_07_verify_available_plugin_(plugin):
+    for plugin_info in job_results.json()[0]['result']:
+        if plugin in plugin_info:
```
sonicaj: This would need updating as changes are introduced in the plugin list. How about we come up with a more robust solution getting contents ( `requests` module ) of `https://raw.githubusercontent.com/freenas/iocage-ix-plugins/master/INDEX` and then converting to json and verifying list of plugins there and the available returned ?

tests/api2/jail_plugins.py
```diff
@@ -0,0 +1,291 @@
...
+
+@to_skip
+def test_08_add_transmision_plugins():
+    global JOB_ID
+    payload = {
+        "plugin_name": "transmission",
+        "jail_name": "transmission",
+        'props': [
+            'nat=1',
+            'vnet=1',
```
sonicaj: `vnet` and `vnet_default_interface` are automatically set when `nat=1`. So i think let's just keep `nat` in this case as this will also keep on checking that the other two are set as desired.

tests/api2/jail_plugins.py
```diff
@@ -0,0 +1,291 @@
...
+    results = GET("/jail/id/transmission")
+    assert results.status_code == 200, results.text
+    assert isinstance(results.json(), dict), results.text
+    transmission_jail = results.json()
+
+
+@to_skip
+@pytest.mark.parametrize('object', plugins_objects)
+def test_13_verify_transmission_plugin_value_with_jail_value_of_(object):
+    assert transmission_jail[object] == transmission_plugin[object], results.text
```
sonicaj: `object` is a built in variable in python, i would recommend we use some other name here perhaps, like `prop` ? As that accurately reflects the value

tests/api2/jail_plugins.py
```diff
@@ -0,0 +1,291 @@
...
+
+
+@to_skip
+def test_14_get_list_of_available_plugins_with_cache():
```
sonicaj: I am not sure if this needed, `cache` is enabled by default ;)

tests/api2/jail_plugins.py
```diff
@@ -0,0 +1,291 @@
...
+        "force": True
+    }
+    results = POST('/jail/stop/', payload)
+    assert results.status_code == 200, results.text
+
+
+@to_skip
+def test_18_wait_for_transmission_plugin_to_be_down():
+    results = GET('/plugin/id/transmission/')
+    while results.json()['state'] == 'up':
```
sonicaj: What happens if `stop` fails ? ;)
We should probably timeout say after a max of 10 seconds. With `force=true`, the jail should stop quickly.

tests/api2/jail_plugins.py
```diff
@@ -0,0 +1,291 @@
...
+    global results
+    payload = "transmission"
+    results = POST('/jail/start/', payload)
+    assert results.status_code == 200, results.text
+
+
+@to_skip
+def test_20_wait_for_transmission_plugin_to_be_up():
+    results = GET('/plugin/id/transmission/')
+    while results.json()['state'] == 'down':
```
sonicaj: This has the same issue as above, we should have a timeout here

tests/api2/jail_plugins.py
```diff
@@ -0,0 +1,291 @@
...
+        sleep(1)
+        results = GET('/plugin/id/transmission/')
+        assert results.status_code == 200, results.text
+    assert results.json()['state'] == 'up', results.text
+
+
+@to_skip
+def test_21_stop_transmission_jail_before_deleteing():
```
sonicaj: I think we can skip this as stopping has been tested before, we should delete with force  flag enabled which will stop + destroy, so we will test multiple scenarios this way. This would be passed over to jail.

tests/api2/jail_plugins.py
ericbsd: It was to make sure the option works.

tests/api2/jail_plugins.py
ericbsd: We did accomplish that in jail testing.

tests/api2/jail_plugins.py
ericbsd: It was to make sure the option works.

ericbsd: So I could test without cache then.

tests/api2/jail_plugins.py
ericbsd: What about this https://github.com/freenas/freenas/pull/3301/commits/133d0a0cc9f73490b002aebfd433d7047e016c2d

sonicaj:
> Just a few minor notes below. Apart from them we should test `version` returned by `plugin.available` and `plugin.query` too. We can pick one plugin like `rslsync` and ensure that it's `version` is not `N/A`. The same is true for `revision` and `epoch` keys.
> Also this doesn't test multiple repository support for plugins. A new repository should be created or an existing one can be used, i was using a fork i made `https://github.com/sonicaj/iocage-ix-plugins.git`. That should be supplied and a plugin should be created which exists in 11.3-RELEASE branch to see if that's successful or not.

Please let's add tests for these as well

tests/api2/jail_plugins.py
sonicaj: @ericbsd it's much better, thank you ;)

tests/api2/plugin.py
```diff
@@ -0,0 +1,258 @@
...
+    assert job_status['state'] == 'SUCCESS', job_status['results']
+    job_results = job_status['results']
+
+
+@to_skip
+@pytest.mark.parametrize('plugin', plugin_list)
+def test_07_verify_available_plugin_(plugin):
+    for plugin_info in job_results['result']:
+        if plugin in plugin_info:
+            assert isinstance(plugin_info, list), str(job_results)
```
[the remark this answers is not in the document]
ericbsd: like this?
```
def test_07_verify_available_plugin_(plugin):
    for plugin_info in job_results['result']:
        assert isinstance(plugin_info, list), str(job_results)
        if plugin in plugin_info:
            assert plugin in plugin_info, str(job_results)
```

tests/functions.py
```diff
@@ -17,22 +18,25 @@
...
 def GET(testpath, **optional):
-    if "api" in optional:
-        api_v = optional.get('api', None)
-        if api_v == "1":
-            api_url = api1_url
-        elif api_v == "2":
-            api_url = api2_url
-        else:
-            raise ValueError('api parameter should be "1" or "2"')
+    if 'https' in testpath or 'http' in testpath:
```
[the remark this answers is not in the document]
ericbsd: Yeah thanks, I never think to use that, also `testpath.startswith('http')` should be enough since https start with the http.

tests/functions.py
```diff
@@ -305,3 +309,18 @@ def ping_host(host):
...
+    while True:
+        job_results = GET(f'/core/get_jobs/?id={job_id}')
+        job_state = job_results.json()[0]['state']
+        if job_state in ('RUNNING', 'WAITING'):
+            sleep(5)
+        elif job_state == 'SUCCESS' or job_state == 'SUCCESS':
+            return {'state': job_state, 'results': job_results.json()[0]}
+        if timeout == 120:
+            return {'state': 'TIMEOUT', 'results': job_results.json()[0]}
+        timeout += 1
```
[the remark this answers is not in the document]
ericbsd: The logic was to have a 10 minute time out not 2 minutes, the server for testing is really slow it needs lots of times.

tests/functions.py
```diff
@@ -17,22 +18,25 @@
...
-    getit = requests.get(api_url + testpath, headers=header,
-                         auth=auth)
+        if "api" in optional:
+            api_v = optional.get('api', None)
+            if api_v == "1":
+                api_url = api1_url
+            elif api_v == "2":
+                api_url = api2_url
+            else:
+                raise ValueError('api parameter should be "1" or "2"')
```
wblock: "Should" is a recommendation.  If it has to be 1 or 2, use "must".
```suggestion
                raise ValueError('api parameter must be "1" or "2"')
```

tests/functions.py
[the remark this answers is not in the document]
ericbsd: never mind I got it I will change the 120 for 600 and add 5 to the timeout.

tests/functions.py
```diff
@@ -305,3 +309,18 @@ def ping_host(host):
...
+
+def wait_on_job(job_id):
+    global job_results
+    timeout = 0
+    while True:
+        job_results = GET(f'/core/get_jobs/?id={job_id}')
+        job_state = job_results.json()[0]['state']
+        if job_state in ('RUNNING', 'WAITING'):
+            sleep(5)
+        elif job_state == 'SUCCESS' or job_state == 'SUCCESS':
```
wblock: This is redundant.
```suggestion
        elif job_state == 'SUCCESS':
```

tests/functions.py
```diff
@@ -305,3 +309,18 @@ def ping_host(host):
         return False
     else:
         return True
+
+
+def wait_on_job(job_id):
+    global job_results
+    timeout = 0
```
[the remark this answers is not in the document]
ericbsd: Actually, I have changed the code of tests 17, 18, 19, 20, 21, 22 to use wait_on_job

tests/functions.py
ericbsd: Yeah the second is supposed to be 'FAILED'

wblock: Ah.  In that case, I find the "in" easier to read:
```suggestion
        elif job_state in ('SUCCESS', 'FAILED'):
```

ericbsd: Yeah that what I had in mind thanks

tests/api2/plugin.py
```diff
@@ -0,0 +1,379 @@
...
+import sys
+import os
+apifolder = os.getcwd()
+sys.path.append(apifolder)
+from auto_config import pool_name
+from functions import GET, POST, DELETE, wait_on_job
+
+JOB_ID = None
+job_results = None
+not_freenas = GET("/system/is_freenas/").json() is False
```
sonicaj: For `master` this is not needed anymore i believe.

tests/api2/plugin.py
```diff
@@ -0,0 +1,379 @@
...
+    job_results = job_status['results']
+
+
+@to_skip
+@pytest.mark.parametrize('plugin', plugin_list)
+def test_07_verify_available_plugin_(plugin):
+    assert isinstance(job_results['result'], list), str(job_results)
+    for plugin_info in job_results['result']:
+        if plugin in plugin_info['plugin']:
+            assert plugin in plugin_info['plugin'], str(job_results)
```
sonicaj: This and the assert below will always be `true`. We should do something like the following
```
assert plugin in [p['plugin'] for p in job_results['result']]
```

sonicaj: And skip the for loop altogether

tests/api2/plugin.py
```diff
@@ -0,0 +1,379 @@
...
+    job_results = job_status['results']
+
+
+@to_skip
+@pytest.mark.parametrize('plugin', plugin_list)
+def test_19_verify_available_plugin_without_cache_(plugin):
+    assert isinstance(job_results['result'], list), str(job_results)
+    for plugin_info in job_results['result']:
+        if plugin in plugin_info['plugin']:
+            assert plugin in plugin_info['plugin'], str(job_results)
```
sonicaj: Same as above

tests/api2/plugin.py
```diff
@@ -0,0 +1,379 @@
...
+    job_results = job_status['results']
+
+
+@to_skip
+@pytest.mark.parametrize('plugin', plugin_list2)
+def test_31_verify_available_plugin_(plugin):
+    assert isinstance(job_results['result'], list), str(job_results)
+    for plugin_info in job_results['result']:
+        if plugin in plugin_info['plugin']:
+            assert plugin in plugin_info['plugin'], str(job_results)
```
sonicaj: Same as above

tests/api2/plugin.py
```diff
@@ -0,0 +1,379 @@
...
+        "force": True
+    }
+    results = POST('/jail/stop/', payload)
+    assert results.status_code == 200, results.text
+    JOB_ID = results.json()
+
+
+@to_skip
+def test_25_wait_for_transmission_plugin_to_be_down():
+    job_status = wait_on_job(JOB_ID)
```
sonicaj: We don't want to wait for `10` minutes for stopping a plugin. That unnecessarily adds a lot of time to the tests. We should allow a `max_timeout` parameter for `wait_on_job` function which it can use instead and it can default to `600`.

tests/api2/plugin.py
```diff
@@ -0,0 +1,379 @@
...
+    global JOB_ID
+    payload = "transmission"
+    results = POST('/jail/start/', payload)
+    assert results.status_code == 200, results.text
+    JOB_ID = results.json()
+
+
+@to_skip
+def test_23_wait_for_transmission_plugin_to_be_up():
+    job_status = wait_on_job(JOB_ID)
```
sonicaj: The same as below for stopping

tests/functions.py
```diff
@@ -305,3 +309,18 @@ def ping_host(host):
...
+    while True:
+        job_results = GET(f'/core/get_jobs/?id={job_id}')
+        job_state = job_results.json()[0]['state']
+        if job_state in ('RUNNING', 'WAITING'):
+            sleep(5)
+        elif job_state in ('SUCCESS', 'FAILED'):
+            return {'state': job_state, 'results': job_results.json()[0]}
+        if timeout == max_timeout:
```
sonicaj: If `max_timeout` is not a multiple of `5`, the timeout will never occur. We should use `>=` then instead.

ericbsd: True

ericbsd:
I will adjust if we see a failure in the future.

PR #3304, opened 2019-07-04
QE NAS-102411 Update jail API testing with the changes in NAS-102286

ericbsd:


tests/api2/jail.py
```diff
@@ -100,37 +77,14 @@ def test_07_verify_fetch_job_state():
...
-                break
-            time.sleep(5)
-        else:
-            assert '11.2-RELEASE' in get_job.json()[0]['result'], get_job.text
-            break
+def test_07_verify_FreeBSD_release_is_installed():
+    results = POST('/jail/releases_choices', False)
+    assert results.status_code == 200, results.text
+    assert isinstance(results.json(), dict), results.text
+    assert '11.2-RELEASE' in results.json(), results.text
```
sonicaj: Should we be using `RELEASE` variable which was set above and depcits the release we have decided to fetch ?

tests/api2/jail.py
```diff
@@ -326,11 +280,21 @@ def test_20_verify_jail_stopped():
...
 
 
-def test_21_rc_action():
+def test_19_rc_action():
     results = POST('/jail/rc_action/', 'STOP')
     assert results.status_code == 200, results.text
 
 
+def test_20_delete_jail():
+    results = DELETE(f'/jail/id/{JAIL_NAME}/')
```
sonicaj: Let's the `force` flag here too please. If a jail is running and we intend to destroy it without stopping and without providing `force`, it will fail whereas if it's running and `force` is provided it shouldn't fail and destroy the jail.

tests/api2/jail.py
ericbsd: Good idea

tests/api2/jail.py
```diff
@@ -31,44 +31,21 @@ def test_02_verify_iocage_pool():
 
 
 def test_03_verify_list_resources_is_an_integer():
```
sonicaj: The name doesn't reflect the test

ericbsd: Oh yeah, I forgot to change that one.

PR #3306, opened 2019-07-05
NAS-102424 / 11.3 / Improve input validation in filesystem.setacl

anodos325:
Create common private winacl job for filesystem.setperm, filesystem.chown, and filesystem.setacl.
Add explicit check for invalid ACL flag combination.

### DIFF
--- a/src/middlewared/middlewared/plugins/filesystem.py
+++ b/src/middlewared/middlewared/plugins/filesystem.py
@@ -285,9 +285,7 @@ class FilesystemService(Service):
 
         return flagset
 
-    @private
-    @job(lock=lambda args: f'_winacl:{args[0]}')
-    def _winacl(self, job, path, action, uid, gid, options):
+    def _winacl(self, path, action, uid, gid, options):
         winacl = subprocess.run([
             '/usr/local/bin/winacl',
             '-a', action,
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -224,6 +224,7 @@ def statfs(self, path):
             'avail_bytes': statfs.avail_blocks * statfs.blocksize,
         }
 
+    @private
```
themylogin: No need for this decorator to methods, whose names start with `_`

src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -285,6 +289,19 @@ def __convert_to_adv_flagset(self, basic_flag):
 
         return flagset
 
+    @private
+    @job(lock=lambda args: f'_winacl:{args[0]}')
+    def _winacl(self, job, path, action, uid, gid, options):
```
themylogin: Why have you decided to make it a job?

anodos325: This can potentially take several hours to complete if there are a significant number of files involved.

themylogin: So we are calling it asynchronously them, `filesystem.acl` returns success immediately and if any exception happens in `_winacl`, it is never shown anywhere?

william-gr:
Wait until Monday to merge please, I would like to discuss this further.

PR #3307, opened 2019-07-05
NAS-102425 / 11.3 / Expose netdata ips in netdata plugin

sonicaj:
This commit exposes netdata ips which user has configured and via which netdata can be accessed.

### DIFF
--- a/src/middlewared/middlewared/plugins/netdata.py
+++ b/src/middlewared/middlewared/plugins/netdata.py
@@ -37,6 +37,27 @@ class NetDataService(SystemServiceService):
             data['alarms'][alarm].pop('write_path', None)
         return data
 
+    @accepts()
+    async def ips(self):
+        """
+        Returns a list of user configured addresses with which netdata can be accessed.
+        """
+        ips = []
+        for ip in (await self.config())['bind']:
+            if ip == '0.0.0.0':
+                ips.extend([
+                    f'http://{d["address"]}/netdata/'
+                    for d in await self.middleware.call('interface.ip_in_use', {'ipv4': True})
+                ])
+            elif ip == '::':
+                ips.extend([
+                    f'http://{d["address"]}/netdata/'
+                    for d in await self.middleware.call('interface.ip_in_use', {'ipv6': True})
+                ])
+            else:
+                ips.append(f'http://{ip}/netdata/')
+        return ips
+
     @private
     async def list_alarms(self):
         alarms = copy.deepcopy(self._alarms)


src/middlewared/middlewared/plugins/netdata.py
```diff
@@ -37,6 +37,27 @@ def __init__(self, *args, **kwargs):
...
+    @accepts()
+    async def ips(self):
+        """
+        Returns a list of user configured addresses with which netdata can be accessed.
+        """
+        ips = []
+        for ip in (await self.config())['bind']:
+            if ip == '0.0.0.0':
+                ips.extend([
+                    f'http://{d["address"]}/netdata/'
```
themylogin: Why not https?

sonicaj: `http+https` is enabled by default in the system. If user wants to enforce `https`, he/she can just set the relevant settings in system.general and all requests would be proxied to `https`.

src/middlewared/middlewared/plugins/netdata.py
```diff
@@ -37,6 +37,27 @@ def __init__(self, *args, **kwargs):
             data['alarms'][alarm].pop('write_path', None)
         return data
 
+    @accepts()
+    async def ips(self):
```
william-gr: They are urls not IPs.

Also, what is the use case?

src/middlewared/middlewared/plugins/netdata.py
themylogin: I mean we always have https enabled and we are trying to enforce to use it by default, why http is the default value here?

src/middlewared/middlewared/plugins/netdata.py
sonicaj: UI requested to know with configurable binding ip's for netdata, where should they redirect the user to launch netdata.
And yes, i should change that to addresses. I wanted to wait till Monday tod discuss this with you as this has a potential problem with `nginx.conf` as there we direct any request with `/netdata` to localhost which in turn means that whichever addresses nginx is using, netdata will be available on those regardless of the fact if user only wants it on one ip.

src/middlewared/middlewared/plugins/netdata.py
sonicaj: Copying `system.general` behavior where the default is to use `http` unless forced by the user. Please advise if you would like me to change this to `https`.

src/middlewared/middlewared/plugins/netdata.py
william-gr: Sounds like thats not the solution.

They have the IP already from the browse, all they need is a path.

If they dont want to hardcode /netdata/ we can provide a method: `netdata.url_path`

william-gr: Also please request them to create a ticket

sonicaj: Makes sense @william-gr. I think we should expose `netdata.url_path` after making sure that UI is okay with that else we can close this. Also i'll make sure in the future to request them to create a ticket first.

PR #3308, opened 2019-07-05
NAS-102430 / 11.3 / Make service announcement types configurable.

anodos325:
Add multi-select to network->global configuration to enable and disable
mDNS, NetBIOS, and WS-Discovery. More work is still needed to decouple
NetBIOS-NS and WS-Discovery from smbd.


william-gr:
In theory we sbould not be adding new features to the queue of new UI for 11.3 anymore, which this does.
Anybgood reason to have this in 11.3 and not 12?

anodos325:
> In theory we sbould not be adding new features to the queue of new UI for 11.3 anymore, which this does.
> Any good reason to have this in 11.3 and not 12?

The older way of doing mDNS had the option to turn off mDNS advertisements for SMB, the gui checkbox maps to an smb4.conf parameter that doesn't do anything now that we're not compiling with mDNS support. I thought it probably makes more sense to make a global setting than one that's limited to a particular file-shaing protocol. I can always make the mDNS plugin look at the SMB configuration to determine whether to register it, but this seems somewhat ad-hoc. That said, I'm happy to bump it to 12 if you prefer.

william-gr:
For 11.3 I would implement what you said and move this to 12.
Its time we only push for things that are essential to get 11.3 out of the door, this does not seem to be the case.

anodos325:
Re-targeting for 12.

PR #3309, opened 2019-07-08
QE NAS-102444 Fix jail API v2 test 20

ericbsd:


tests/api2/jail.py
```diff
@@ -288,13 +288,14 @@ def test_19_start_jail():
...
+    results = GET(f'/jail/id/{JAIL_NAME}/')
+    assert results.status_code == 200, results.text
     timeout = 0
     while results.json()['state'] == 'down':
         time.sleep(1)
-        results = GET('/plugin/id/transmission/')
+        results = GET(f'/jail/id/{JAIL_NAME}/')
         assert results.status_code == 200, results.text
-        if timeout == 10:
+        if timeout == 60:
```
sonicaj: `60` seconds is a bit too much i think to wait for in this case. `10-15` is more then reasonable. Also please correct `test_19` as that also is starting `transmission` jail which doesn't exist and the return value is a job so the test doesn't show as failed.

PR #3323, opened 2019-07-11
NAS-102494 / 11.3 / Expose compression algorithm selection for import/export of jail

sonicaj:
1) Expose compression algorithm selection for import/export of a jail.
2) Expose path selection for where a compressed jail might live when we import a jail.

### DIFF
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -1244,12 +1244,14 @@ class JailService(CRUDService):
         Dict(
             'options',
             Str('jail', required=True),
-            Str('compression_algorithm', default=None, null=True)
+            Str('compression_algorithm', default='ZIP', enum=['ZIP', 'LZMA'])
         )
     )
     @job(lock=lambda args: f'jail_export:{args[-1].get("jail")}')
     def export(self, job, options):
-        """Exports jail to zip file"""
+        """
+        Export jail to compressed file.
+        """
         jail = options['jail']
         uuid, path, _ = self.check_jail_existence(jail)
         status, jid = IOCList.list_get_jid(uuid)
@@ -1259,7 +1261,7 @@ class JailService(CRUDService):
             self.middleware.call_sync('jail.stop', jail, job=True)
             started = True
 
-        IOCImage().export_jail(uuid, path, compression_algo=options['compression_algorithm'])
+        IOCImage().export_jail(uuid, path, compression_algo=options['compression_algorithm'].lower())
 
         if started:
             self.middleware.call_sync('jail.start', jail, job=True)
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/jail.py
```diff
@@ -1240,10 +1240,19 @@ def progress_callback(content, exception):
...
+        Dict(
+            'options',
+            Str('jail', required=True),
+            Str('compression_algorithm', default='ZIP', enum=['ZIP', 'LZMA'])
+        )
+    )
+    @job(lock=lambda args: f'jail_export:{args[-1].get("jail")}')
+    def export(self, job, options):
+        """
+        Exports jail to compressed file.
```
wblock: I would use present tense for docstrings.  It's shorter and often clearer.
```suggestion
        Export jail to compressed file.
```

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -1252,19 +1261,38 @@ def export(self, job, jail):
...
+            'options',
+            Str('jail', required=True),
+            Str('path', default=None, null=True),
+            Str('compression_algorithm', default=None, null=True, enum=['ZIP', 'LZMA', None])
+        )
+    )
+    @job(lock=lambda args: f'jail_import:{args[-1].get("jail")}')
+    def import_image(self, job, options):
+        """
+        Imports jail from compressed file.
```
wblock: As above.
```suggestion
        Import jail from compressed file.
```

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -1252,19 +1261,38 @@ def export(self, job, jail):
...
+
+        `compression algorithm` if set to None indicates that middlewared will automatically determine
```
wblock: Break into shorter sentences.
```suggestion
        `compression algorithm`: None indicates that middlewared is to automatically determine
```

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -1252,19 +1261,38 @@ def export(self, job, jail):
...
+        which compression algorithm to use based on compressed file extension but if multiple copies are found an
```
wblock:
```suggestion
        which compression algorithm to use based on the compressed file extension. If multiple copies are found, an
```

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -1252,19 +1261,38 @@ def export(self, job, jail):
...
+        exception will be raised.
```
wblock:
```suggestion
        exception is raised.
```

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -1252,19 +1261,38 @@ def export(self, job, jail):
...
 
-        IOCImage().import_jail(jail)
+        `path` is the directory where exported jail lives. It defaults to the iocage's images dataset.
```
wblock: Avoid possessives.
```suggestion
        `path` is the directory where the exported jail lives. It defaults to the iocage images dataset.
```

PR #3327, opened 2019-07-12
NAS-102450 / 11.3 / Set HSTS header if explicitly specified

sonicaj:
This PR introduces changes to only set hsts header explicitly if user wants traffic to be only https based. Also when this is not wanted, we do add the header but set the value to 0 which invalidates the last setting browser saves for the ip/domain.

### DIFF
--- a/gui/system/models.py
+++ b/gui/system/models.py
@@ -84,7 +84,8 @@ class Settings(Model):
         verbose_name=_('WebGUI HTTP -> HTTPS Redirect'),
         default=False,
         help_text=_(
-            'Redirect all incoming HTTP requests to HTTPS'
+            'Redirect all incoming HTTP requests to HTTPS and '
+            'enable the HTTP Strict Transport Security (HSTS) header.'
         ),
     )
     stg_language = models.CharField(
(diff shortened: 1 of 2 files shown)


gui/system/models.py
```diff
@@ -84,7 +84,8 @@ class Settings(Model):
         verbose_name=_('WebGUI HTTP -> HTTPS Redirect'),
         default=False,
         help_text=_(
-            'Redirect all incoming HTTP requests to HTTPS'
+            'Redirect all incoming HTTP requests to HTTPS. This '
```
Mrt134:
```suggestion
            'Redirect all incoming HTTP requests to HTTPS and '
```

gui/system/models.py
```diff
@@ -84,7 +84,8 @@ class Settings(Model):
...
+            'enables HSTS header to be used for each request enhancing security.'
```
Mrt134:
```suggestion
            'enable the HTTP Strict Transport Security (HSTS) header.'
```

Mrt134: Suggested simplifying and combining the two sentences. It might be nice to add further clarification about using this setting:
 "User-agents without Transport Layer Security (TLS) will not be able to connect to this system."

PR #3333, opened 2019-07-17
NAS-102482 / 11.3 / Use new py-libzfs api for datasets retrieval

sonicaj:
This PR updates zfs.dataset namespace to use new dataset serialization api exposed by py-libzfs which is way faster then `zfs.datasets`. Also we expose two structure options for retrieving datasets, first is hierarchical and the second is a flattened approach.

### DIFF
--- a/src/middlewared/middlewared/plugins/zfs.py
+++ b/src/middlewared/middlewared/plugins/zfs.py
@@ -453,17 +453,17 @@ class ZFSDatasetService(CRUDService):
         return filter_list(datasets, filters, options)
 
     def query_for_quota_alert(self):
-        with libzfs.ZFS() as zfs:
-            return [
-                {
-                    k: v.__getstate__()
-                    for k, v in i.properties.items()
-                    if k in ["name", "quota", "used", "refquota", "usedbydataset", "mounted", "mountpoint",
-                             "org.freenas:quota_warning", "org.freenas:quota_critical",
-                             "org.freenas:refquota_warning", "org.freenas:refquota_critical"]
-                }
-                for i in zfs.datasets
-            ]
+        return [
+            {
+                k: v for k, v in dataset['properties'].items()
+                if k in [
+                    "name", "quota", "used", "refquota", "usedbydataset", "mounted", "mountpoint",
+                    "org.freenas:quota_warning", "org.freenas:quota_critical",
+                    "org.freenas:refquota_warning", "org.freenas:refquota_critical"
+                ]
+            }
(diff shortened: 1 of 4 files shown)


sonicaj:
I did, it was almost similar by a degree of `~0.1/2` seconds for `~1400` datasets.

william-gr:
Did you use cdef definition as opposed to def?

sonicaj:
No @william-gr, i'll try that and get back on this asap.

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -423,6 +423,32 @@ def flatten_datasets(self, datasets):
...
+
+        We provide 2 ways how zfs.dataset.query returns dataset's data. First is a flat structure ( default ), which
+        means that all the datasets in the system are returned as separate objects which also contain all the data
+        their is for their children. This retrieval type is slightly slower because of duplicates which exist in
+        each object.
+        Second type is hierarchical where only top level datasets are returned in the list and they contain all the
+        children there are for them in `children` key. This retrieval type is slightly faster.
+        These options are controlled by `query-options.extra.flat` attribute which defaults to true.
+
+        `query-options.extra.custom_properties` controls if user defined properties of datasets should be retrieved
```
william-gr: If uts user properties, why call it custom_properties and not user_properties?

sonicaj: I was reflecting on that when i wrote the docs. I'll update this :)

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -417,50 +418,82 @@ class Config:
...
+        `query-options.extra.user_properties` controls if user defined properties of datasets should be retrieved
+        or not.
+
+        While we provide a way to exclude all properties from data retrieval, we introduce a single attribute
+        `query-options.extra.retrieve_properties` which if set to false will make sure that no property is retrieved
+        whatsoever and overrides any other property retrieval attribute.
+        """
+        options = options or {}
+        extra = options.get('extra', {}).copy()
+        top_level_props = None if extra.get('top_level_properties') is None else extra['top_level_properties'].copy()
```
william-gr: You say mountpoint is top level by default but I am not seeing how? Did you do it in py-libzfs?

sonicaj: Yes, if `top_level_properties` is None and no custom props are specified, `py-libzfs` will add it automatically

PR #3334, opened 2019-07-17
NAS-102523 / 11.3 / Update official repositories choices for plugins

sonicaj:
This commit updates the api we expose for retrieving officially supported plugin repos including an icon uri and a user friendly name for the official repo.

### DIFF
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -146,7 +146,10 @@ class PluginService(CRUDService):
         process_pool = True
 
     OFFICIAL_REPOSITORIES = {
-        'https://github.com/freenas/iocage-ix-plugins.git'
+        'IXSYSTEMS': {
+            'name': 'iXsystems',
+            'git_repository': 'https://github.com/freenas/iocage-ix-plugins.git',
+        }
     }
 
     @accepts()
@@ -154,7 +157,7 @@ class PluginService(CRUDService):
         """
         List officially supported plugin repositories.
         """
-        return {k: k for k in self.OFFICIAL_REPOSITORIES}
+        return self.OFFICIAL_REPOSITORIES
 
     @filterable
     def query(self, filters=None, options=None):


src/middlewared/middlewared/plugins/jail.py
```diff
@@ -146,15 +146,15 @@ class Config:
         process_pool = True
 
     OFFICIAL_REPOSITORIES = {
-        'https://github.com/freenas/iocage-ix-plugins.git'
+        'ixSystems': 'https://github.com/freenas/iocage-ix-plugins.git'
     }
```
william-gr: Sounds like this should be a list, because what if we want to add more metadata to each repository later on?

william-gr: Or a dict with another dict as values

sonicaj: This seems most reasonable to me,
```
{'ix': {'git_repository': 'uri'}}
```
Thoughts ? This would mean that the friendly name would be unique and act as a key, else we can have a list and dicts inside

william-gr: The key can be an enum, e.g.

{"IXSYSTEMS": {..."name": "iXsystems", "git_repository": "..."...}, "COMMUNITY": {...}}

sonicaj: Perfect, i'll update accordingly

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -37,6 +38,10 @@
 SHUTDOWN_LOCK = asyncio.Lock()
 
 
+class OfficialPluginRepositories(enum.Enum):
+    IXSYSTEMS = 'ixSystems'
```
william-gr: iXsystems is what we call it

sonicaj: Updating asap

william-gr: Also not sure how an ENUM helps here?

I mentioned enum in the PR just as an acronym (C lang), not actual enum python module

PR #3335, opened 2019-07-17
NAS-102536 / 11.3 / Ensure that smb_configure only runs after system dataset setup

anodos325:
If samba's private dir does not exist, then smb-related commands will fail
with 'invalid messaging context'. This situation can occur if smb_configure
runs before the system dataset is set up.


themylogin:
backport

PR #3338, opened 2019-07-18
EC2 support

themylogin:


src/freenas-installer/etc/installer12.sh
```diff
@@ -1161,16 +1161,6 @@ menu_install()
     done
 
     chroot /tmp/data /usr/sbin/sysrc -f /boot/loader.conf rc_system=rc
-    chroot /tmp/data /usr/bin/touch /etc/dynamicdiskless
-    chroot /tmp/data /bin/mkdir -p /conf/base/etc
-    chroot /tmp/data /bin/sh -c '/bin/echo 65535 > /conf/base/etc/md_size'
```
william-gr: Why these should be different for EC2?

themylogin: They are the same, we are just not using installer on EC2

william-gr: We can revert these changes now, not used.

We should remove that entire file later

src/freenas/usr/local/libexec/nas/post-install
```diff
@@ -0,0 +1,31 @@
...
+/usr/bin/touch /etc/dynamicdiskless
+
+/bin/mkdir -p /conf/base/etc
+/bin/echo 65535 > /conf/base/etc/md_size
+
+/bin/mkdir -p /conf/base/mnt
+/bin/echo 8192 > /conf/base/mnt/md_size
+
+/bin/mkdir -p /conf/base/usr__local__etc
+/bin/echo 65535 > /conf/base/usr__local__etc/md_size
```
william-gr: I dont think these paths are used in 12, its the same as 11.3.

src/freenas/usr/local/libexec/nas/post-install
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+if [ -f /.ec2 ];
+then
+    /usr/sbin/sysrc -f /boot/loader.conf if_ena_load=YES
+    /usr/sbin/sysrc -f /boot/loader.conf autoboot_delay=-1
+    /usr/sbin/sysrc -f /boot/loader.conf beastie_disable=YES
+    /usr/sbin/sysrc -f /boot/loader.conf boot_multicons=YES
+    /usr/sbin/sysrc -f /boot/loader.conf hint.atkbd.0.disabled=1
+    /usr/sbin/sysrc -f /boot/loader.conf hint.atkbdc.0.disabled=1
```
william-gr: I just noticed that if we changed loader.conf we would get a error in freenas-verify because its part of the base.

Would this work if we changed loader.conf.local instead?

william-gr: @themylogin ping, in case you have forgotten

PR #3340, opened 2019-07-18
NAS-100230 / 11.3 / NAS-100230

sv0:
This pull request adds third option 'multi-line command' to 'Tasks' > 'init/shutdown'


src/middlewared/middlewared/plugins/init_shutdown_script.py
```diff
@@ -16,8 +18,9 @@ class Config:
 
     @accepts(Dict(
         'init_shutdown_script_create',
-        Str('type', enum=['COMMAND', 'SCRIPT'], required=True),
+        Str('type', enum=['COMMAND', 'MULTILINE_COMMAND', 'SCRIPT'], required=True),
```
william-gr: MULTILINE_COMMAND seems like a bit strange name.

Its actually a script, but as a text and not a file path, since it can run any script, e.g.:

#!/usr/bin/env python
print("foo")

What are you thoughts on keeping COMMAND and SCRIPT as type and introduce the new field a `scripttext` or something? Making `script` and `script_text` mutually exclusive using validation?

src/middlewared/middlewared/plugins/init_shutdown_script.py
```diff
@@ -147,10 +166,20 @@ class Config:
...
+                elif task_type == 'SCRIPT':
+                    cmd = task['script']
+                else:
+                    cmd = ''
                 self.middleware.logger.debug(
                     'Execution failed for '
-                    f'{task_type} {task["command"] if task_type == "COMMAND" else task["script"]}: {stderr.decode()}'
+                    f'{task_type} {cmd}: {stderr.decode()}'
                 )
+        if tmp_script:
```
william-gr: If an exception occurs for any reason it will leave a dangling file.

Seems like we need this cleanup method via a try/finally.

src/middlewared/middlewared/plugins/init_shutdown_script.py
sv0: OK. What validation error message should be raised in case if both `script` and `script_text` fields are filled up?

Something like `Only one of  two fields should be provided.` ?

william-gr: Thats seems OK. When we are finished we will ask documentation team their input on the wordings.

In an ideal world this message will only be seem by bad usage of the API and UI will have a selection box and/or hide the fields when appropriate.

PR #3344, opened 2019-07-19
NAS-102570 / 11.3 / Multiple fixes in smb4.conf

anodos325:
Increase default number max winbindd connections. Offline logons limits winbind to single connection per domain. This can limit scalability in large environments.
Turn off user and group enumeration in winbind if the FreeNAS cache is disabled.
Fix smb4.conf generation on TrueNAS HA (unified config had second [global] heading).
Fix ordering of vfs objects for shadow_copy_zfs and ixnas.
vfs_noacl now disables NT ACL support behind the scenes so it no longer should be stacked with zfsacl.
LDAP hostnames is now always a list, so remove this check.
'client use spnego = yes' is now the default, so remove the parameter.
mDNS registration now happens through middleware, so remove this parameter.

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/smb4.conf
+++ b/src/middlewared/middlewared/etc_files/local/smb4.conf
@@ -23,7 +23,7 @@
 
             conf['cifs'] = middleware.call_sync('smb.config')
             conf['ad'] = middleware.call_sync('activedirectory.config')
-            conf['ldap'] = middleware.call_sync('datastore.config', 'directoryservice.ldap')
+            conf['ldap'] = middleware.call_sync('ldap.config')
             conf['nis'] = middleware.call_sync('datastore.config', 'directoryservice.nis')
             conf['gc'] = middleware.call_sync('network.configuration.config')
             conf['shares'] = middleware.call_sync('sharing.smb.query', [['enabled', '=', True]])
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/etc_files/local/smb4.conf
```diff
@@ -149,58 +166,64 @@
...
+                    'ldap group suffix': db['ldap']['ldap_groupsuffix'],
+                    'ldap machine suffix': db['ldap']['ldap_machinesuffix'],
+                    'ldap replication sleep': '1000',
+                    'ldap passwd sync': 'Yes',
+                    'ldap ssl': 'off' if db['ldap']['ldap_ssl'] != 'START_TLS' else 'start tls',
+                    'ldapsam:trusted': 'Yes',
+                    'domain logons': 'Yes',
+                    'workgroup': db['cifs']['workgroup'].upper(),
+                })
+                ldap_pdb_backend = "ldapsam:\"%s\"" & " ".join([
```
william-gr: `&` ? i cant see how that could have ever worked?

src/middlewared/middlewared/etc_files/local/smb4.conf
```diff
@@ -149,58 +166,64 @@
...
+                    "%s://%s" % ("ldaps" if db['ldap']['ldap_ssl'] == 'ON' else "ldap", hostname)
```
william-gr: While you're here might as well convert it to .format() for f-string. `%` is deprecated in our code

PR #3346, opened 2019-07-19
NAS-101734 / 11.2 / Convert size str to int for raw devices

sonicaj:
THis commit adds a migration to convert strings to integers for raw file size in case of vm devices.

### DIFF
--- a/gui/vm/migrations/0008_size_field.py
+++ b/gui/vm/migrations/0008_size_field.py
@@ -1,0 +1,40 @@
+import math
+
+from django.db import migrations
+
+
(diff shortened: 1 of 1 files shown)


sonicaj:
@william-gr changes in vm plugin are not required as on creation we raise a call error in a routine which is executed before this which ensures that existing destination is not truncated

sonicaj:
backport

sonicaj:
As discussed with @william-gr, we don't want this in 11.2, so we will move ahead with 11.3 changes only

PR #3348, opened 2019-07-21
NAS-102566 / 11.3 / Disable legacy UI by default

sonicaj:
1) Disables legacy UI by default.
2) Exposes an unauthenticated api to query legacy UI status.
3) Collects legacy UI status anonymously via our usage plugin.

### DIFF
--- a/gui/system/migrations/0046_legacy_ui.py
+++ b/gui/system/migrations/0046_legacy_ui.py
@@ -1,0 +1,20 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('system', '0045_fromname'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='Advanced',
+            name='adv_legacy_ui',
+            field=models.BooleanField(
+                default=False,
+                help_text='Enable or disable the legacy UI.',
+                verbose_name='Enable legacy UI'
+            ),
+        )
+    ]
--- a/gui/system/models.py
+++ b/gui/system/models.py
@@ -218,6 +218,11 @@ class NTPServer(Model):
 
 
 class Advanced(Model):
+    adv_legacy_ui = models.BooleanField(
+        verbose_name=_('Enable legacy UI'),
+        default=False,
+        help_text=_('Enable or disable the legacy UI.')
+    )
     adv_consolemenu = models.BooleanField(
         verbose_name=_("Show Text Console without Password Prompt"),
         default=False,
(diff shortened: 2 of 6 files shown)


src/middlewared/middlewared/plugins/system.py
```diff
@@ -133,6 +134,8 @@ class Config:
 
         `autotune` when enabled executes autotune script which attempts to optimize the system based on the installed
         hardware.
+
+        `legacy_ui` is disabled by default and when enabled, it will allow end users to be able to use legacy UI.
```
Mrt134:
```suggestion
        `legacy_ui` is disabled by default. Enabling it allows end users to use the legacy UI.
```

src/middlewared/middlewared/plugins/system.py
```diff
@@ -239,6 +242,14 @@ class SystemService(Service):
         """
         return "FreeNAS" if await self.middleware.call("system.is_freenas") else "TrueNAS"
 
+    @no_auth_required
+    @accepts()
+    async def legacy_ui_enabled(self):
+        """
+        Returns a boolean value indicating if legacy UI is enabled/disabled for use by the end user.
```
Mrt134:
```suggestion
        Returns a boolean value indicating if the legacy UI can be used by end users.
```

gui/system/models.py
```diff
@@ -218,6 +218,11 @@ class FreeAdmin:
 
 
 class Advanced(Model):
+    adv_legacy_ui = models.BooleanField(
+        verbose_name=_('Enable legacy UI'),
+        default=False,
+        help_text=_('Enable to use legacy UI.')
```
Mrt134:
```suggestion
        help_text=_('Enable using the legacy UI.')
```

gui/system/migrations/0046_legacy_ui.py
```diff
@@ -0,0 +1,20 @@
...
+        ('system', '0045_fromname'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='Advanced',
+            name='adv_legacy_ui',
+            field=models.BooleanField(
+                default=False,
+                help_text='Enable to use legacy UI.',
```
Mrt134:
```suggestion
                help_text='Enable using the legacy UI.',
```

sonicaj: @Mrt134 do you think this says more like that this field should be enabled using the legacy UI rather then that this field enables legacy UI ?

Mrt134: Hm, it could be interpreted that way. I'll take another stab at it.

PR #3349, opened 2019-07-22
NAS-100230 / 12 / Enhance init/shutdown scripts to allow multiline text box as a script option

sv0:


src/middlewared/middlewared/plugins/init_shutdown_script.py
```diff
@@ -149,9 +169,24 @@ class Config:
...
+                else:
+                    cmd = ''
                 self.middleware.logger.debug(
                     'Execution failed for '
-                    f'{task_type} {task["command"] if task_type == "COMMAND" else task["script"]}: {stderr.decode()}'
+                    f'{task_type} {cmd}: {stderr.decode()}'
+                )
+        if tmp_script:
+            try:
+                os.unlink(tmp_script)
```
william-gr: I think you may have misunderstood me in the earlier comment.

Say an exception occurs anywhere between the creation of the tmpfile and this (e.g. subprocess communicate for w/e reason), what will happen? File will stay there forever?

william-gr:
@themylogin care to review this, please?

src/middlewared/middlewared/plugins/init_shutdown_script.py
```diff
@@ -132,27 +145,50 @@ class Config:
...
+                        cmd = task['command']
+                    elif task_type == 'SCRIPT' and task['script_text']:
+                        cmd = task['script_text']
+                    elif task_type == 'SCRIPT' and task['script']:
+                        cmd = task['script']
+                    else:
+                        cmd = ''
+                    self.middleware.logger.debug(
+                        'Execution failed for '
+                        f'{task_type} {cmd}: {stderr.decode()}'
```
themylogin: `script_text` can be big multi-line string, we don't want that in logs.
Why only stderr? Won't it be better to redirect stderr to stdout and put stdout here?

src/middlewared/middlewared/plugins/init_shutdown_script.py
```diff
@@ -132,27 +145,50 @@ class Config:
...
+                    )
+        except Exception as error:
+            self.middleware.logger.debug(
+                f'{task["type"]} {cmd}: {error.strerror}'
```
themylogin: `Exception` class does not have `strerror` attribute, it would be better to use `{error!r}`

src/middlewared/middlewared/plugins/init_shutdown_script.py
themylogin: It will still be multi-line and confusing. As we don't have any other visible identifier, I would go with `ellipsis(repr(task['script_text'], 64))`

sv0: > It will still be multi-line and confusing. As we don't have any other visible identifier, I would go with `ellipsis(repr(task['script_text'], 64))`

Not sure what you mean.
`script_text` will be always multi-line.
I guess putting the first `script_text` lines to the log will not help a lot to identify the problematic `script_text`.
So I'm not sure what to do.

william-gr: Put some identifier of the task (e.g. id and comment). We dont need the script text.

themylogin: Let's assume that first 64 chars of script (in one line) is the best we can provide to identify it, everything else will be flooding logs.

themylogin: @william-gr ID is useless (not seen in UI) and we don't have comment (maybe we should add one? `name`)

william-gr: Thats too verbose, we just need an identifier, the id and comment are good enough. The script it too verbose either way.

william-gr: Better than a multi line script. middlewared.log is not seen in the UI either so its a moot point.

Which raises the question, why are we even logging it instead of raising a CallError?

william-gr: I will asnwer that myself, to allow other scripts to run.

That needs to be refactored later on, but for the scope of this ticket the ID seems good enough to me. (We can eventualyl show the ID in the UI if thats the case)

themylogin: @william-gr because now it's the only way for these to be seen, especially for those at SHUTDOWN.

As we are in 12.0, maybe we can think if it is worth to store exit code / output of the last run in DB and expose it in UI?

themylogin: We can store them in tmpfs as we do with other job logs, but there still there will be no way to retrieve errors/output for scripts ran at SHUTDOWN.

william-gr: @themylogin we do have a comment in 12.

Everything else seems out of the scope of the ticket to me, which can be improved later.

themylogin:
@sv0 this was left unresolved

"Why only stderr? Won't it be better to redirect stderr to stdout and put stdout here?"

(in "Execution failed for" logging)

sv0:
> @sv0 this was left unresolved
> 
> "Why only stderr? Won't it be better to redirect stderr to stdout and put stdout here?"
> 
> (in "Execution failed for" logging)

Done in commit 635da22

william-gr:
@freenas/docs-team ping

PR #3392, opened 2019-07-26
NAS-102685 / 11.3 / Use winbindd connection manager to monitor domain status

anodos325:
Generate alert when domain goes offline and clear it when domain comes back online. Once a domain goes offline, by default the winbindd connection manager will try to reconnect every 30 seconds. Update state entry accordingly.

### DIFF
--- a/src/middlewared/middlewared/alert/source/active_directory.py
+++ b/src/middlewared/middlewared/alert/source/active_directory.py
@@ -1,6 +1,6 @@
 from datetime import timedelta
 import logging
-from middlewared.alert.base import AlertClass, AlertCategory, Alert, AlertLevel, AlertSource
+from middlewared.alert.base import AlertClass, AlertCategory, Alert, AlertLevel, AlertSource, SimpleOneShotAlertClass
 from middlewared.alert.schedule import CrontabSchedule, IntervalSchedule
 
 log = logging.getLogger("activedirectory_check_alertmod")
@@ -18,6 +18,13 @@ class ActiveDirectoryDomainHealthAlertClass(AlertClass):
     level = AlertLevel.WARNING
     title = "Active Directory Domain Validation Failed"
     text = "Domain validation failed with error: %(verrs)s."
+
+
+class ActiveDirectoryDomainOfflineAlertClass(AlertClass, SimpleOneShotAlertClass):
+    category = AlertCategory.DIRECTORY_SERVICE
+    level = AlertLevel.WARNING
+    title = "Domain Offline"
+    text = "Active Directory Domain \"%(domain)s\" is Offline."
 
 
 class ActiveDirectoryDomainHealthAlertSource(AlertSource):
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -615,7 +615,7 @@ class ActiveDirectoryService(ConfigService):
         Str('idmap_backend', default='RID', enum=['AD', 'AUTORID', 'FRUIT', 'LDAP', 'NSS', 'RFC2307', 'RID', 'SCRIPT']),
         Str('nss_info', null=True, default='', enum=['SFU', 'SFU20', 'RFC2307']),
         Str('ldap_sasl_wrapping', default='SIGN', enum=['PLAIN', 'SIGN', 'SEAL']),
-        Str('createcomputer', null=True),
+        Str('createcomputer'),
         Str('netbiosname'),
         Str('netbiosname_b'),
         List('netbiosalias'),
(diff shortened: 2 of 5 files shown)


src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1331,3 +1336,117 @@ def fill_cache(self, job, force=False):
...
+                self.middleware.call_sync("alert.oneshot_delete",
+                    "ActiveDirectoryDomainOffline", {"domain": m["domain_name_netbios"]}
+                )
+
+        self.state = new_state
+
+    def read_messages(self):
+        while not self.finished.is_set():
+            with open('/var/run/samba4/.wb_fifo') as f:
+                while not self.finished.is_set():
```
william-gr: Why two `while not self.finished.is_set()`

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1331,3 +1336,117 @@ def fill_cache(self, job, force=False):
...
+                    select.select([f], [], [f])
```
william-gr: Dont we need a timeout here? Otherwise it will block forever, even if `self.finished.is_set`

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1331,3 +1336,117 @@ def fill_cache(self, job, force=False):
...
+                        self.parse_msg(data)
+
+        self.logger.debug('exiting winbind messaging thread')
+
+    def run(self):
+        set_thread_name('ad_monitor_thread')
+        try:
+            self.read_messages()
+        except Exception as e:
+            self.logger.debug('Failed to start monitor thread %s', e)
```
william-gr: exc_info=True so we know what the exception is, exactly?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1331,3 +1336,117 @@ def fill_cache(self, job, force=False):
...
+
+    def setup(self):
+        pass
+
+    def cancel(self):
+        self.finished.set()
+
+
+class ADMonitorService(Service):
```
william-gr: Perhaps we should have

```
class Config:
   private = True
```

and remove private decorator from methods?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -601,15 +606,15 @@ class Config:
...
+        Str('site', null=True),
         Int('kerberos_realm', null=True),
         Str('kerberos_principal', null=True),
         Int('timeout'),
         Int('dns_timeout'),
         Str('idmap_backend', default='RID', enum=['AD', 'AUTORID', 'FRUIT', 'LDAP', 'NSS', 'RFC2307', 'RID', 'SCRIPT']),
         Str('nss_info', null=True, default='', enum=['SFU', 'SFU20', 'RFC2307']),
         Str('ldap_sasl_wrapping', default='SIGN', enum=['PLAIN', 'SIGN', 'SEAL']),
-        Str('createcomputer'),
+        Str('createcomputer', null=True),
```
themylogin: As I see it, this model field is NOT NULL in DB

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1331,3 +1336,117 @@ def fill_cache(self, job, force=False):
...
+        self.initialized = False
+        self.lock = threading.Lock()
+
+    @private
+    def start(self):
+        with self.lock:
+            if self.initialized:
+                return
+
+        thread = WBStatusThread(
```
themylogin: Why is this outside of locking block?

src/middlewared/middlewared/plugins/activedirectory.py
themylogin: Maybe it'll be best to say "failed to **run** AD monitor thread"?

src/middlewared/middlewared/alert/source/active_directory.py
```diff
@@ -20,6 +20,13 @@ class ActiveDirectoryDomainHealthAlertClass(AlertClass):
     text = "Domain validation failed with error: %(verrs)s."
 
 
+class ActiveDirectoryDomainOfflineAlertClass(AlertClass, SimpleOneShotAlertClass):
```
themylogin: Who creates this alert?

anodos325: Alert is generated when domain state transitions to offline. Is deleted when it comes back online. see parse_msg().

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1331,3 +1337,142 @@ def fill_cache(self, job, force=False):
...
+                    "alert.oneshot_delete",
+                    "ActiveDirectoryDomainOffline",
+                    {"domain": m["domain_name_netbios"]}
+                )
+
+        self.state = new_state
+
+    def read_messages(self):
+        while not self.finished.is_set():
+            with open('/var/run/samba4/.wb_fifo') as f:
```
william-gr: Do we need to open the file every iteration?
Should this be reversed? while within open?

PR #3398, opened 2019-07-29
We dont need pwenc_check in system.general.config

themylogin:


sonicaj:
@themylogin I believe that is system.advanced in the diff.

themylogin:
@sonicaj thank you!

PR #3402, opened 2019-07-30
NAS-102719 / 11.3 / Streamline smb service starts

anodos325:
Decouple samba daemons and only reload / restart when it's actually needed.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -460,6 +460,21 @@ class SMBService(SystemServiceService):
                 raise CallError(f'Failed to enable {username}: {enableacct.stderr.decode()}')
 
     @private
+    def wsdd_cmd(self, cmd=None):
+        if cmd is None:
+            raise CallError('wsdd command not specified')
+
+        wsdd = subprocess.Popen(['service', 'wsdd', cmd])
+        try:
+            wsdd.communicate(timeout=10)
+        except subprocess.TimeoutExpired:
+            wsdd.kill()
+            wsdd.communicate()
+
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -459,6 +459,21 @@ def check_codec(encoding):
...
+    def wsdd_cmd(self, cmd=None):
+        if cmd is None:
+            raise CallError('wsdd command not specified')
+
+        wsdd = subprocess.Popen(['service', 'wsdd', cmd])
+        try:
+            wsdd.communicate(timeout=10)
+        except subprocess.TimeoutExpired:
+            wsdd.kill()
+            wsdd.communicate()
```
william-gr: Why do you need a communicate after killing?

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -459,6 +459,21 @@ def check_codec(encoding):
             if enableacct.returncode != 0:
                 raise CallError(f'Failed to enable {username}: {enableacct.stderr.decode()}')
 
+    @private
+    def wsdd_cmd(self, cmd=None):
```
william-gr: Do you see this being used outside service plugin?

anodos325: At this point I don't see it being used outside of the SMB plugin. It's fairly tightly coupled with smbd. wsdd won't actually start unless "enable web service discovery" is set in the smb4.conf file.

william-gr: I meant service plugin, as in, why this method is not in service plugin if all this handles in service cmd?

PR #3404, opened 2019-07-31
NAS-102579 / 11.3 / Fix mess in alert service schemas

themylogin:


william-gr:
IIRC we dont have some of these attributes in 11.2. Is it even worth backporting ?

themylogin:
It's worth backporting, it is the only proper way to address this bug. The only thing changed is SNMP.

PR #3411, opened 2019-08-01
NAS-102753 / 11.3 / /usr/local/libexec/nas/scrub to middlewared

themylogin:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3382,6 +3388,64 @@ class Config:
         await self.middleware.call('service.restart', 'cron')
         return response
 
+    @accepts(Str('name'), Int('threshold', default=35))
+    async def scrub(self, name, threshold):
```
william-gr: run might be a better name, pool.scrub.run instead of pool.scrub.scrub : P

src/middlewared/middlewared/etc_files/crontab
```diff
@@ -10,7 +10,7 @@
 %>\
 ${system_crontab}
 
-45 3 * * * root /usr/local/libexec/nas/scrub -t ${system_advanced['boot_scrub']} freenas-boot
+45 3 * * * root /usr/local/bin/midclt call pool.scrub.scrub freenas-boot ${system_advanced['boot_scrub']}
```
william-gr: Should probably redirect stdout/stderr now?

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -916,6 +918,29 @@ def cancel(self):
         self._cancel.set()
 
 
+class ScrubNotStartedAlertClass(AlertClass, OneShotAlertClass):
+    category = AlertCategory.TASKS
+    level = AlertLevel.WARNING
+    title = "Scrub Not Started"
```
william-gr: Scrub Not Started seems strange, perhaps only use it for actual failures and say "Scrub Failed to Start" or something?

Then do not generate it when scrub or resilver is in progress.

themylogin: Do we drop the notification that scrub was not started due to another scrub/resilver?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3382,6 +3388,61 @@ class Config:
...
+            })
+        else:
+            if started:
+                await self.middleware.call('alert.oneshot_create', 'ScrubStarted', name)
+
+    async def __run(self, name, threshold):
+        if name == 'freenas-boot':
+            if not await self.middleware.call('system.is_freenas'):
+                if await self.middleware.call('failover.status') == 'BACKUP':
+                    return
```
william-gr: Now that I come to think about this, I cant think of any reason why we wouldnt run scrub of boot pool in standby node.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3382,6 +3388,61 @@ class Config:
...
+        for match in reversed(list(RE_HISTORY_ZPOOL_SCRUB.finditer(history))):
+            last_scrub = datetime.strptime(match.group(1), '%Y-%m-%d.%H:%M:%S')
+            break
+        else:
+            # creation time of the pool if no scrub was done
+            for match in RE_HISTORY_ZPOOL_CREATE.finditer(history):
+                last_scrub = datetime.strptime(match.group(1), '%Y-%m-%d.%H:%M:%S')
+                break
+            else:
+                raise ScrubError(f'Skipping scrubbing of pool {name}: can\'t get last scrubbing date')
```
william-gr: I am not sure we should raise an error if we cant find the pool create time? Perhaps just log it and let it start the scrub?

src/middlewared/middlewared/plugins/pool.py
themylogin: Good that you've noted it, I've put this in a wrong place

PR #3425, opened 2019-08-06
NAS-102776 / 11.3 / Add health checks for LDAP and NIS

anodos325:
Also add ability to get basic domain info from AD.


yocalebo:
@anodos325 If these alerts are valid for the passive storage controller, can we make sure these don't run?

PR #3428, opened 2019-08-07
NAS-102571 / 12 / Fix position of installer boot logo

freqlabs:
With the old menu now ported to Lua, this manual override is no longer
correct.


william-gr:
sonicj is on a day off, thus i am approving this on his behalf

PR #3438, opened 2019-08-08
NAS-102813 / 11.3 / Rsync task reporting

sonicaj:
This PR introduces alerts for rsync tasks if quiet attribute of a rsync task is disabled.

### DIFF
--- a/src/middlewared/middlewared/plugins/rsync.py
+++ b/src/middlewared/middlewared/plugins/rsync.py
@@ -567,7 +567,15 @@ class RsyncTaskService(CRUDService):
             commandline, rsync['user'], lambda v: job.logs_fd.write(v)
         )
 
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/rsync.py
```diff
@@ -563,11 +563,19 @@ def run(self, job, id):
         rsync = self.middleware.call_sync('rsynctask._get_instance', id)
         commandline = self.middleware.call_sync('rsynctask.commandline', id)
 
+        for klass in ('RsyncSuccess', 'RsyncFailed'):
```
themylogin: I think you should do it after `run_command_with_user_context` so there would be no time gap between alert gone and alert appearance.

PR #3443, opened 2019-08-09
NAS-102825 / 11.3 / Add exception handling for stale and expired winbindd cache entries

anodos325:


src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1318,13 +1318,19 @@ def fill_cache(self, job, force=False):
...
                         try:
                             pwd.getpwuid(int(cached_gid))
                             break
-                        except Exception:
-                            group_data = grp.getgrgid(int(cached_gid))
+                        except KeyError:
+                            try:
+                                group_data = grp.getgrgid(int(cached_gid))
+                            except KeyError:
+                                break
```
william-gr: Shouldnt this be a continue instead of break?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1304,27 +1304,33 @@ def fill_cache(self, job, force=False):
...
 
-            if 'GID2SID' in line:
+            if line.startswith('Key: IDMAP/GID2SID'):
                 cached_gid = ((line.split())[1].split('/'))[2]
                 is_local_group = any(filter(lambda x: x['gid'] == int(cached_gid), local_groups))
                 if is_local_group:
                     continue
 
                 for d in known_domains:
-                    if int(cached_gid) in range(d['low_id'], d['high_id']):
```
freqlabs: Doesn't this change the behavior? Previous range was [ low_id, high_id ), now it's [ low_id, high_id ].

Also of note, checking int membership in a range has been a constant-time operation rather than linear since Python 3.2. You're shaving off some time to init a range object but then giving it back up to parse the int from a string an extra time each iteration. If you're looking for more optimizations, parsing that cached_uid and cached_gid into an ints outside the loops would save you from doing it repeatedly inside a loop.

PR #3445, opened 2019-08-09
NAS-102819 / 12 / Installer improvements for FreeNAS 12

freqlabs:
Remove vestigial installer12.sh script.

Remove the ability to upgrade from legacy versions of FreeNAS (9.2 or
earlier). The suggested upgrade path has long been to upgrade to each
consecutive major release. This upgrade process will have repartitioned
the legacy disk format to create the new ZFS-based system, so it is time
we shed the extra complexity required to support this operation.

When importing the freenas-boot pool to check a disk for a an existing
FreeNAS install, import the pool by its GUID taken from the partition's
label instead of by name in case there are multiple pools with the name
"freenas-boot". Likewise, use the GUID when importing the pool to create
a boot-environment during in-place upgrades.

Fix mistaken use of `local readonly` which only creates a local variable
named readonly instead of the intended readonly local variable.

Garbage collect dead code and trailing whitespace.

Jira ticket: https://jira.ixsystems.com/browse/NAS-102819


freqlabs:
Sorry about the extra housekeeping getting tied up in the changes. I have tested the upgrade and fresh install paths and mostly just want to confirm that we don't need the installer12.sh file still.

william-gr:
Yeah, installer12.sh was used for the freebsd pkg base, which we wont be using anymore.

PR #3450, opened 2019-08-12
NAS-100078 / 12 /  Remote syslog over TLS

sv0:


src/middlewared/middlewared/etc_files/syslogd.py
```diff
@@ -33,8 +33,16 @@ def generate_syslog_conf(middleware):
         else:
             host, port = advanced_config["syslogserver"], "514"
 
-        syslog_conf += f'destination loghost {{ udp("{host}" port({port}) localport(514)); }};\n'
-        syslog_conf += f'log {{ source(src); filter({advanced_config["sysloglevel"].lower()}); destination(loghost); }};\n'
+        if advanced_config.get('syslog_transport') == 'tls':
+            syslog_conf += f'destination loghost {{'
+            syslog_conf += f'syslog("{host}" port({port}) transport("tls") tls(ca-dir("/etc/certificates")));'
+        else:
+            transport = advanced_config.get('syslog_transport', 'tcp')
```
themylogin: These values are always defined, why `.get` with default value? Seems like we are scattering defaults all over the code. And the default is udp, no?

src/middlewared/middlewared/plugins/system.py
```diff
@@ -161,79 +164,85 @@ class Config:
...
         if verrors:
             raise verrors
 
-        if len(set(config_data.items()) ^ set(original_data.items())) > 0:
-            if original_data.get('sed_user'):
-                original_data['sed_user'] = original_data['sed_user'].lower()
-            if config_data.get('sed_user'):
-                config_data['sed_user'] = config_data['sed_user'].lower()
+        # if len(set(config_new.items()) ^ set(config_current.items())) > 0:
+        if config_new != config_current:
```
themylogin: Why?

src/middlewared/middlewared/plugins/system.py
```diff
@@ -161,79 +164,85 @@ class Config:
 
         `legacy_ui` is disabled by default. Enabling it allows end users to use the legacy UI.
         """
-        config_data = await self.config()
-        original_data = config_data.copy()
-        config_data.update(data)
+        config_current = await self.config()
```
themylogin: If you want to refactor this, it seems to me that we are using `old` and `new` in most of our other code

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -1044,7 +1044,14 @@ def __init__(self, *args, **kwargs):
         cert['csr_path'] = os.path.join(
             root_path, f'{cert["name"]}.csr'
         )
-
+        x509 = crypto.load_certificate(
+            crypto.FILETYPE_PEM,
+            cert['certificate']
+        )
+        cert['subject_name_hash'] = '%x' % x509.subject_name_hash()
+        cert['hash_symlink_path'] = os.path.join(
```
sonicaj: subject name hash should be added where we load the cert in cryptokey service as if it fails to load etc, it will be handled appropriately in certificate extend method.

src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -2660,7 +2667,7 @@ def set_defaults(attr):
                 ]
             }
         """
-        ca = await self._get_instance(id)
+        ca = await self._get_instance(id)  # noqa: is assigned to but never used
```
sonicaj: We should not assign to ca but keep the get instance call as it ensures that a ca of that id exists

src/middlewared/middlewared/plugins/system.py
```diff
@@ -75,12 +75,13 @@ class Config:
...
         if data.get('sed_user'):
             data['sed_user'] = data.get('sed_user').upper()
+
         if data.get('sysloglevel'):
             data['sysloglevel'] = data['sysloglevel'].upper()
 
         return data
 
-    async def __validate_fields(self, schema, data):
+    async def validate_advanced_settings(self, schema, data):
```
sonicaj: We should have this method marked as private or keep the old implementation

src/middlewared/middlewared/etc_files/syslogd.py
```diff
@@ -33,8 +33,16 @@ def generate_syslog_conf(middleware):
...
+            syslog_conf += f'destination loghost {{ {transport}("{host}" port({port}) localport(514)); '
+
+        syslog_conf += f'}};\n'
```
sonicaj: f string not required here ;)

src/middlewared/middlewared/etc_files/syslogd.py
```diff
@@ -9,6 +9,53 @@
 logger = logging.getLogger(__name__)
 
 
+def generate_syslog_remote_destination(middleware, advanced_config):
+    result = ""
+    if advanced_config["syslogserver"]:
+        if ":" in advanced_config["syslogserver"]:
+            host, port = advanced_config["syslogserver"].split(":")[:2]
```
sonicaj: Do you think it's good to discard rest of the values silently ( :2 ) ? How about we use `split/rsplit(':', 1)` ?

src/middlewared/middlewared/plugins/system.py
```diff
@@ -116,6 +120,23 @@ class Config:
...
+                await self.middleware.call(
+                    'certificate.query',
+                    [
+                        ('id', '=', data['syslog_tls_certificate']),
+                        ('revoked', '=', False),
+                    ],
+                    {'get': True}
+                )
+            except IndexError:
+                verrors.add(
```
sonicaj: I think it might be better to use `certificate.cert_services_validation` method to check for this. That probably does not yet check for certificate's revocation status, that check can be added in it and then we should use it as it works nicely to impose certain security limits on certs being used for different services.

src/middlewared/middlewared/etc_files/syslogd.py
```diff
@@ -9,6 +9,53 @@
...
+
+        result += 'destination loghost { '
+
+        if advanced_config["syslog_transport"] == "TLS":
+            try:
+                certificate = middleware.call_sync(
+                    "certificate.query",
+                    [
+                        ("id", "=", advanced_config["syslog_tls_certificate"]),
+                        ("revoked", "=", False),
```
sonicaj: An alert should be generated when a certificate is revoked which is in use by a service. We should make amends in `certificates.py` for alert source to reflect this.

src/middlewared/middlewared/etc_files/syslogd.py
sonicaj: Or maybe not allow such values which contain more then one instance of `:` ?

src/middlewared/middlewared/plugins/system.py
```diff
@@ -116,6 +120,11 @@ class Config:
                         f'{schema}.syslogserver',
                         'Port must be in the range of 0 to 65535.'
                     )
+
+        if data['syslog_transport'] == 'TLS':
+            await self.middleware.call('certificate.cert_services_validation', data['syslog_tls_certificate'],
```
sonicaj: Can we please add a check for cert revocation status in this method or is it already there ?

PR #3451, opened 2019-08-12
NAS-102582 / 11.3 / Forbid using netcli when failover is not disabled

themylogin:


src/freenas/etc/netcli
```diff
@@ -76,6 +76,18 @@ from requests.packages.urllib3 import poolmanager
...
 
 
+class FailoverIsEnabledException(Exception):
+    pass
+
+
+def ensure_failover_is_disabled():
+    if not notifier().is_freenas():
+        from freenasUI.failover.models import Failover
+        if Failover.objects.all().exists():
```
william-gr: Not that it matters much but you could probably save a query if you save Failover.objects.all() as `qs`

PR #3457, opened 2019-08-14
NAS-102849 / 11.3 / Use filesystem.chown and filesystem.setperm in account and webdav

anodos325:
Avoid direct chown and chmod in account and webdav plugins. This will improve auditing since they're middleware jobs and ensure proper interaction with ZFS ACLs.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -220,7 +220,13 @@ class UserService(CRUDService):
                 try:
                     os.makedirs(data['home'], mode=int(home_mode, 8))
                     new_homedir = True
-                    os.chown(data['home'], data['uid'], group['gid'])
+                    await self.middleware.call('filesystem.setperm', {
+                        'path': data['home'],
+                        'mode': home_mode,
+                        'uid': data['uid'],
+                        'gid': group['gid'],
+                        'options': {'stripacl': True}
+                    })
                 except FileExistsError:
                     if not os.path.isdir(data['home']):
                         raise CallError(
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/account.py
```diff
@@ -702,11 +726,22 @@ def shell_choices(self):
...
+            'gid': group,
+            'options': {'recursive': True, 'stripacl': True}
+        })
+
         with open(keysfile, 'w') as f:
             f.write(pubkey)
             f.write('\n')
-        os.chmod(keysfile, 0o600)
-        await run('/usr/sbin/chown', '-R', f'{user["username"]}:{group}', sshpath, check=False)
+        await self.middleware.call('filesystem.setperm', {'path': keysfile, 'mode': str(600)})
```
william-gr: str(600) is not right, equals 0o1130, perhaps just `'600'` ?

src/middlewared/middlewared/plugins/account.py
```diff
@@ -229,8 +235,13 @@ class Config:
...
                         )
 
-                    # If it exists, ensure the user is owner
-                    os.chown(data['home'], data['uid'], group['gid'])
+                    # If it exists, ensure the user is owner.
+                    await self.middleware.call('filesystem.chown', {
+                        'path': data['home'],
+                        'uid': data['uid'],
+                        'gid': group['gid'],
+                        'options': {'recursive': True}
```
william-gr: recursive=True is both is a change of behavior, are we sure we want to chown every single file with user homedir?

Legit question, I am trying to think situations where user would not want to have other files changed behind their back.
One I can think of is suid/sgid.

src/middlewared/middlewared/plugins/account.py
william-gr: Err, nvm, brainfart :)

src/middlewared/middlewared/plugins/account.py
anodos325: Yeah, I started thinking more, recursive here is a bad idea.

PR #3463, opened 2019-08-15
 Fix RuntimeError("Middleware is terminating") in the middle of initialization

themylogin:
Event loop is not running between the calls of `loop.run_until_complete`. Some IO thread (spawned e.g. in plugin setup function) that uses `run_coroutine` can get a false feeling of middleware being terminated because event loop is shut down.


themylogin:
@william-gr I only had this on 12.0 (with ~5% probability), but propose to fix 11.3 too because any new thread / timing change from big bugfixes (that still are possible) can introduce this.

PR #3465, opened 2019-08-16
NAS-102845 / 11.3 / Enclose dynamic DNS username in quotes

sonicaj:
This commit makes sure we use quotes to enclose dynamic DNS username as for certain usernames involving email addresses with a '+' symbol, dynamic dns service fails to start if isn't enclosed properly.

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/inadyn.conf
+++ b/src/middlewared/middlewared/etc_files/local/inadyn.conf
@@ -31,7 +31,7 @@ provider ${config['provider']} {
     checkip-path = "${config['checkip_path']}"
 % endif
 % if config['username']:
-    username = ${config['username']}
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/etc_files/local/inadyn.conf
```diff
@@ -31,7 +31,7 @@ provider ${config['provider']} {
     checkip-path = "${config['checkip_path']}"
 % endif
 % if config['username']:
-    username = ${config['username']}
+    username = "${config['username']}"
```
william-gr: Can username have a quote in itself?

sonicaj: I haven't been able to find a reference which indicates that so far.

PR #3466, opened 2019-08-16
NAS-102878 / 11.3 / Fix regression in webdav start

anodos325:
Finish cleaning up permissions-related code in webdav config generation. Share permissions are changed when the share is created.

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/apache24/Includes/webdav.conf
+++ b/src/middlewared/middlewared/etc_files/local/apache24/Includes/webdav.conf
@@ -45,12 +45,9 @@
 		raise ValueError("Invalid auth_type (must be one of 'none', 'basic', 'digest')")
 
 	if path:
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/etc_files/local/apache24/Includes/webdav.conf
```diff
@@ -45,12 +45,9 @@
 		raise ValueError("Invalid auth_type (must be one of 'none', 'basic', 'digest')")
 
 	if path:
+		uid = middleware.call_sync('user.get_user_obj', {'username': 'webdav'})['pw_uid']
+		gid = middleware.call_sync('group.get_group_obj', {'groupname': 'webdav'})['gr_gid']
 		os.chown(path, uid, gid)
```
william-gr: We kept chown? should we remove all that block?

PR #3471, opened 2019-08-16
fix(middlewared): remove many uses of notifier plugin

william-gr:


src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -865,8 +865,7 @@ def extent_naa(self, naa):
                 diskchoices[f'zvol/{snap["name"]}'] = f'{snap["name"]} [ro]'
 
         for disk in await self.middleware.call('disk.get_unused'):
-            size = await self.middleware.call('notifier.humanize_size', disk['size'])
-            diskchoices[disk['name']] = f'{disk["name"]} ({size})'
+            diskchoices[disk['name']] = f'{disk["name"]}|{disk["size"]}'
```
sonicaj: We are changing format here, is this intentional ?

william-gr: Yes, its not job of middleware to provide only formatted data to that extent, idk why it was there before.

PR #3477, opened 2019-08-19
Update trueview_stats.sh

beanpole135:
Quick patch to increase the length of the polling interval for network traffic stats.
Should be backported to 11.2 ASAP as well.


william-gr:
backport

PR #3479, opened 2019-08-19
NAS-102908 / 11.3 / Fix sharing.webdav.update

anodos325:


src/middlewared/middlewared/plugins/webdav.py
```diff
@@ -114,9 +114,9 @@ class Config:
 
             await self._service_change('webdav', 'reload')
 
-        if data['perm']:
+        if new['perm']:
```
william-gr: Isnt this going to apply permission on every update? Even if perm was not changed?

PR #3484, opened 2019-08-20
NAS-102910 / 12 / fix(middlewared/network): make VRRP scan work

william-gr:
Also stop using code from freenasUI.

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -13,17 +13,12 @@ import itertools
 import netif
 import os
 import re
-import sys
 import shlex
 import signal
 import socket
 import subprocess
 import urllib.request
 
-
-# TODO: move scan_for_vrrp to middleware
-if '/usr/local/www' not in sys.path:
-    sys.path.append('/usr/local/www')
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -1491,6 +1486,38 @@ def choices(self, options):
...
 
+    @private
+    def scan_vrrp(self, ifname, count=10, timeout=10):
+        # This runs tcpdump looking for VRRP packets.  If
+        # none are seen we have a gun in the glovebox that times the
+        # tcpdump out after `timeout` seconds. If we do see VRRP packets we stop
+        # after seeing `count` of them. If there were more than `count` VRRP
+        # devices on the broadcast domain we wouldn't get them all.
+        proc = subprocess.Popen(
+            f"tcpdump -l -n {f'-c {count}' if count else ''} -i {ifname} vrrp 2> /dev/null"
```
sonicaj: This will add an extra space between `-n` and `-i` if count is false but i don't think that would hurt ?

william-gr: Its shell, so not a problem, looked better than the alternative of not having any spaces between the f-string and other param ; )

PR #3489, opened 2019-08-21
Bring fenced from TrueNAS

william-gr:


william-gr:
Not looking for an actual review, just moving code from TrueNAS.

PR #3491, opened 2019-08-22
NAS-102894 / 11.2 / Do not start NFS on HA standby node

themylogin:


themylogin:
@william-gr I am not sure how NFS/SMART/Asigra/etc are started on boot then, carp hook is executed on boot?

william-gr:
Yes, carp is disabled until the end of boot process, which then triggers the carp hook script.

william-gr:
Does this need an equivalent for 11.3 rc.conf.py?

themylogin:
Yes, added backport label

PR #3492, opened 2019-08-22
NAS-102919 / 11.2 / change verbiage for "gone alert" emails when alert is cleared

themylogin:


src/middlewared/middlewared/alert/base.py
```diff
@@ -190,10 +190,18 @@ def format_alerts(product_name, hostname, node_map, alerts, gone_alerts, new_ale
         return text + "This is a test alert"
 
     if new_alerts:
-        text += "New alerts:\n" + "".join(["* %s\n" % format_alert(alert, node_map) for alert in new_alerts]) + "\n"
+        if len(gone_alerts) == 1:
+            text += "New alert"
+        else:
+            text += "New alerts"
+        text += ":\n" + "".join(["* %s\n" % format_alert(alert, node_map) for alert in new_alerts]) + "\n"
```
wblock: Why are we still using the old Python 2 string interpolation instead of the easier to read Python3 f() function?
```suggestion
        text += ":\n" + "".join([f"* {format_alert(alert, node_map)}\n" for alert in new_alerts]) + "\n"
```

src/middlewared/middlewared/alert/base.py
```diff
@@ -190,10 +190,18 @@ def format_alerts(product_name, hostname, node_map, alerts, gone_alerts, new_ale
...
 
     if gone_alerts:
-        text += "Gone alerts:\n" + "".join(["* %s\n" % format_alert(alert, node_map) for alert in gone_alerts]) + "\n"
+        if len(gone_alerts) == 1:
+            text += "The following alert has been cleared"
```
wblock: If it is immediately following, the reader can see that.
```suggestion
            text += "This alert has been cleared"
```

src/middlewared/middlewared/alert/base.py
```diff
@@ -190,10 +190,18 @@ def format_alerts(product_name, hostname, node_map, alerts, gone_alerts, new_ale
...
+        else:
+            text += "The following alerts have been cleared"
```
wblock: As above:
```suggestion
            text += "These alerts have been cleared"
```

src/middlewared/middlewared/alert/base.py
```diff
@@ -190,10 +190,18 @@ def format_alerts(product_name, hostname, node_map, alerts, gone_alerts, new_ale
...
+        text += ":\n" + "".join(["* %s\n" % format_alert(alert, node_map) for alert in gone_alerts]) + "\n"
```
wblock: As above with the f() function.

src/middlewared/middlewared/alert/base.py
```diff
@@ -190,10 +190,18 @@ def format_alerts(product_name, hostname, node_map, alerts, gone_alerts, new_ale
...
 
     if alerts:
         text += "Current alerts:\n" + "".join(["* %s\n" % format_alert(alert, node_map) for alert in alerts]) + "\n"
```
wblock: Here too.

src/middlewared/middlewared/alert/base.py
william-gr: Please focus on documentation, no python code : )

PR #3496, opened 2019-08-22
NAS-102930 / 12 / Replace mdnsresponder with avahi

anodos325:
- remove py-bonjour as middleware dependency
- remove most of middleware plugin (leave basic reload functionality in place to keep this a more-or-less drop-in replacement for mDNSResponder)
- move logic for generating dns-sd sdRef to python script in etc_files and have it generate avahi xml service definitions


src/middlewared/middlewared/etc_files/local/avahi/avahi-daemon.conf
```diff
@@ -0,0 +1,52 @@
...
+# License along with avahi; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307
+# USA.
+# See avahi-daemon.conf(5).
+<%
+    import os
+    import sys
+    logger = middleware.logger
+
+    def ipv4_enabled():
```
william-gr: Just an idea, you could perhaps replace all this method with
any(call_sync('interface.ip_in_use', {'ipv4': true, 'ipv6': false}))

src/middlewared/middlewared/plugins/mdns.py
```diff
@@ -1,455 +1,38 @@
...
-        pass
-
-    def cancel(self):
-        self.finished.set()
+from middlewared.service import Service, private, accepts
+from middlewared.service_exception import CallError
+from middlewared.utils import run
 
 
 class mDNSAdvertiseService(Service):
```
william-gr: Do we need this service/plugin?

Could it make more sense to have _start_mdns, _reload_mdns, etc in service.py instead and use those?

PR #3497, opened 2019-08-23
NAS-100417 / 12 / Add Libvirt Support

sonicaj:
This PR aims to integrate libvirt in the VM plugin making sure moving forward we make use of libvirt to manage our VM's.


sonicaj:
@william-gr @themylogin can you please review this ? I am still running some more tests to ensure we don't have regressions but i believe this is ready for a first pass of reviews. Please let me know what you think, thank you

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -7,481 +6,609 @@
...
+            )
+
+        self.domain.undefine()
+        self.domain = None
+
+    def __getattribute__(self, item):
+        retrieved_item = object.__getattribute__(self, item)
+        if callable(retrieved_item) and item in ('start', 'stop', 'restart', 'poweroff', 'undefine_domain', 'status'):
+            if not getattr(self, 'domain', None):
+                raise CallError('Domain attribute not defined, please re-instantiate the VM class')
```
themylogin: I doubt that this internal logic errors should be `CallError`. After a while, we would be forced to i18n all this, and translating programming errors would have no sense. Can it be a `RuntimeError` for example since user is not responsible for this and can't fix anything?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -7,481 +6,609 @@
...
+        shutdown_timeout = 30
+        while shutdown_timeout > 0 and self.status()['state'] == 'RUNNING':
+            shutdown_timeout -= 5
+            time.sleep(5)
+
+        for device in self.devices:
+            device.post_stop_vm()
+
+    def restart(self, vm_data=None):
+        self._before_stopping_checks()
```
themylogin: Already done in `self.stop()`?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -7,481 +6,609 @@
...
+
+        self.domain.shutdown()
+
+        # We wait for 30 seconds before initiating post stop activities for the vm
+        # This is done because the shutdown call above is non-blocking
+        shutdown_timeout = 30
+        while shutdown_timeout > 0 and self.status()['state'] == 'RUNNING':
+            shutdown_timeout -= 5
+            time.sleep(5)
+
```
themylogin: Why is there no check that machine stop is successful (and in restart method there is)?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1210,51 +1422,46 @@ def random_mac(self):
...
+    @private
+    async def wait_for_libvirtd(self, timeout):
+        async def libvirtd_started(middleware):
+            while not await middleware.call('service.started', 'libvirtd'):
+                pass
+
+        try:
+            await asyncio.wait_for(libvirtd_started(self.middleware), timeout=timeout)
+        except asyncio.TimeoutError:
+            # TODO: We can log here but there is no good way to distinguish between a boot and middlewared restart
```
themylogin: `system.ready`?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1649,4 +1873,5 @@ def setup(middleware):
     asyncio.ensure_future(kmod_load())
     asyncio.ensure_future(middleware.call('pool.dataset.register_attachment_delegate',
                                           VMFSAttachmentDelegate(middleware)))
+    asyncio.ensure_future(middleware.call('vm.initialize_vms', 2))  # We use a short timeout here deliberately
```
themylogin: Why is this done twice: here and after system boot?

src/middlewared/middlewared/plugins/vm.py
sonicaj: I had a peek at `system.ready` yesterday and it wasn't working as desired for me i.e on middlewared restarts it still showed the state as `BOOTING`, but i can't reproduce that anymore, so added a todo initially. I'll have this updated.

src/middlewared/middlewared/plugins/vm.py
sonicaj: We want to make sure that we have initialized vms on middlewared restart and on boot. For boot, the first call would fail as libvirtd isn't running yet. Ideally this first call should only be made when `system.ready` returns true. I'll update this to use that. Would you know of any reason where `system.ready` can't be trusted ? It checks for `.bootready` file and it should be there, i wonder why it wasn't their earlier.
For now, i'll update this to correctly reflect that behavior.

src/middlewared/middlewared/plugins/vm.py
sonicaj: My thinking here was that a shutdown signal is being sent to the VM, it's possible that there is some blocking task which is forcing it to not go in shutdown mode. In that case user will see that the status of VM is still running and he/she can issue a poweroff. If we make this a blocking call ( the shutdown one to ensure that shutdown happens ), it would have to be a job or if we want to raise an exception, what's the ideal time to wait for the machine to cleanly shutdown ?

themylogin: My thinking is shutdown timeout should be user-configurable.

Also now you are performing `device.post_stop_vm()` even if vm has not stopped which is not good in any case.

src/middlewared/middlewared/plugins/vm.py
themylogin: > i.e on middlewared restarts it still showed the state as `BOOTING`

This should not happen, if it does, it is a bug

src/middlewared/middlewared/plugins/vm.py
sonicaj: User configurable as in an attribute specific to a VM or all VM's ?

sonicaj: In either cases we would have to come up with a reasonable default as well, do you think 30 seconds is good in that case ?

src/middlewared/middlewared/plugins/vm.py
sonicaj: I'll keep an eye out for it, i wasn't sure if that is known. Thank you for the update

src/middlewared/middlewared/plugins/vm.py
themylogin: > User configurable as in an attribute specific to a VM or all VM's ?

To `stop` operation.

30 seems to be too short, we had something like 90 in FreeBSD rc scripts?

sonicaj: I'll have this updated, also i think we shouldn't run post stop activities if the VM is running in that case. Do you agree ? This can cause issues wrt VM stopping down later and we unable to run post stop for the devices in both poweroff/stop as VM is not active

sonicaj: I think i just reversed it, i should run post stop regardless VM is active or not as the user had a chance to configure the timeout for stopping. Thoughts ?

themylogin: Can we spawn a background coroutine that will perform post stop actions when VM is finally shut down? Regardless of stop job status.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1,487 +1,651 @@
...
+            raise CallError(f'Unable to find boot devices for {self.libvirt_domain_name} domain')
+
+        if len([d for d in self.devices if isinstance(d, VNC)]) > 1:
+            raise CallError('Only one VNC device per VM is supported')
+
+        for device in self.devices:
+            device.pre_start_vm()
+
+        if self.domain.create() < 0:
+            raise CallError(f'Failed to boot {self.vm_data["name"]} domain')
```
themylogin: Changes made by `pre_start_vm` won't be rolled back if error happens

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1,487 +1,651 @@
...
+
+        try:
+            for device in self.devices:
+                device.post_start_vm()
```
themylogin: Exception for one device will prevent subsequent devices from initializing

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1,487 +1,651 @@
...
+
+        self.domain.shutdown()
+
+        # We wait for timeout seconds before initiating post stop activities for the vm
+        # This is done because the shutdown call above is non-blocking
+        while shutdown_timeout > 0 and self.status()['state'] == 'RUNNING':
+            shutdown_timeout -= 5
+            time.sleep(5)
+
+        self.initiate_post_stop_thread()
```
themylogin: Who will do this if the machine has shut down itself?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1,487 +1,651 @@
...
+        for device in self.devices:
+            with contextlib.suppress(Exception):
+                device.pre_stop_vm()
+
+        self.domain.destroy()
+
+        self.initiate_post_stop_thread()
+
+        with contextlib.suppress(Exception):
+            self.stop_devices_thread.join()
```
themylogin: It has two names: `post_stop_thread` and `stop_devices_thread`, we need to choose one

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1,487 +1,651 @@
...
 
-                disk_sector_size = int(device['attributes'].get('sectorsize') or 0)
-                if disk_sector_size > 0:
-                    sectorsize_args = ',sectorsize=' + str(disk_sector_size)
+            with open(device_map_file, 'w') as f:
+                f.write(device_map_data)
+
+            if self.vm_data['grubconfig']:
+                grub_config = self.vm_data['grubconfig'].strip()
+                if grub_config.startswith('/mnt') and os.path.exists(grub_config):
```
themylogin: What does `startswith('/mnt')` check help to do?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1,487 +1,651 @@
...
-                return True
-            else:
-                # XXX: We return true for now to keep the vm.status sane.
-                # It is necessary handle in a better way the bhyve process associated with the vmm.
-                return True
-        elif bhyve_error == 1:
-            return False
+    def post_stop_vm(self, *args, **kwargs):
+        if self.web_process and psutil.pid_exists(self.web_process.pid):
+            os.kill(self.web_process.pid, signal.SIGTERM)
```
themylogin: We have `service.terminate_process` that also sends SIGKILL after specific timeout.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1643,10 +1924,13 @@ class VMFSAttachmentDelegate(FSAttachmentDelegate):
...
-def setup(middleware):
+async def setup(middleware):
     global ZFS_ARC_MAX_INITIAL
     ZFS_ARC_MAX_INITIAL = sysctl.filter('vfs.zfs.arc.max')[0].value
     asyncio.ensure_future(kmod_load())
     asyncio.ensure_future(middleware.call('pool.dataset.register_attachment_delegate',
                                           VMFSAttachmentDelegate(middleware)))
+
+    if await middleware.call('system.ready'):
+        asyncio.ensure_future(middleware.call('vm.initialize_vms', 2))  # We use a short timeout here deliberately
```
themylogin: If middleware dies, what happens to VMs?

sonicaj: VM's live on as they are being handled by libvirt. However VNC won't work anymore in that case as the subprocess would stop as well when middlewared dies

src/middlewared/middlewared/plugins/vm.py
sonicaj: I think it ensures that we don't have this somewhere else in the system. These bits have been copied from legacy logic to support old behavior.

src/middlewared/middlewared/plugins/vm.py
sonicaj: Should they be ?

src/middlewared/middlewared/plugins/vm.py
themylogin: So newly started middleware will see that VMs as already running?

sonicaj: Yes

themylogin: Good, thank you

src/middlewared/middlewared/plugins/vm.py
sonicaj: But we can't access middlewared from here, VMSupervisor can act as a standalone and doesn't need middlewared at all right now. Do you think that aspect needs to be changed ?

themylogin: I doubt it will ever be used as standalone but making no leftover processes is a good reason to inject middleware here.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -158,36 +157,58 @@ def start(self, vm_data=None):
...
 
-    def stop(self):
-        self._before_stopping_checks()
+    def run_post_stop_actions(self):
+        while self.status()['state'] == 'RUNNING':
+            time.sleep(5)
 
         for device in self.devices:
-            device.pre_stop_vm()
+            device.post_stop_vm()
```
themylogin: Now much better, but I think exception handling is also needed here

PR #3500, opened 2019-08-23
NAS-102935 / 11.3 / Remove directory service related sysctls

anodos325:


gui/directoryservice/migrations/0014_remove_sysctls.py
```diff
@@ -0,0 +1,61 @@
...
+        'freenas.directoryservice.activedirectory.timeout.start',
+        'freenas.directoryservice.timeout.reload',
+        'freenas.directoryservice.timeout.restart',
+        'freenas.directoryservice.timeout.started',
+        'freenas.directoryservice.timeout.stop',
+        'freenas.directoryservice.timeout.start'
+    ]
+    sysctls = apps.get_model('system', 'tunable')
+    for s in fn_sysctl:
+        to_delete = sysctls.objects.filter(tun_type='sysctl',
```
william-gr: We should replace all that list with:

sysctls.objects.filter(tun_type='sysctl', tun_var__startswith='freenas.').delete()

PR #3506, opened 2019-08-26
QE NAS-102948 Chage jail/export payload and add test to verify that export job succeed

ericbsd:
Also change all test the verify job to use wait_on_job

### DIFF
--- a/tests/api2/jail.py
+++ b/tests/api2/jail.py
@@ -144,7 +144,7 @@ def test_13_verify_jail_started():
         pytest.skip(freeze_msg)
     freeze = False
     job_status = wait_on_job(JOB_ID, 20)
-    if job_status['state'] in ['TIMEOUT', "FAILED"]:
+    if job_status['state'] in ['TIMEOUT', 'FAILED']:
         freeze = True
         freeze_msg = f"Failed to start jail: {JAIL_NAME}"
     assert job_status['state'] == 'SUCCESS', str(job_status['results'])
@@ -177,7 +177,7 @@ def test_15_verify_exec_job():
         pytest.skip(freeze_msg)
     freeze = False
     job_status = wait_on_job(JOB_ID, 300)
-    if job_status['state'] in ['TIMEOUT', "FAILED"]:
+    if job_status['state'] in ['TIMEOUT', 'FAILED']:
         freeze = True
         freeze_msg = f"Failed to exec jail: {JAIL_NAME}"
     assert job_status['state'] == 'SUCCESS', str(job_status['results'])


tests/api2/jail.py
```diff
@@ -163,40 +139,42 @@ def test_12_start_jail():
...
-        else:
-            results = GET('/jail/')
-            assert results.status_code == 200, results.text
-            assert len(results.json()) > 0, get_job.text
-            assert results.json()[0]['state'].lower() == 'up', results.text
-            break
-
-
-def test_14_export_call():
+    job_status = wait_on_job(JOB_ID, 600)
```
sonicaj: I think that's a lot of time to wait for an ordinary jail to start, how about say 15 seconds ?

tests/api2/jail.py
```diff
@@ -163,40 +139,42 @@ def test_12_start_jail():
...
+    if job_status['state'] in ['TIMEOUT', "FAILED"]:
+        freeze = True
+        freeze_msg = f"Failed to start jail: {JAIL_NAME}"
+    assert job_status['state'] == 'SUCCESS', str(job_status['results'])
+    results = GET(f'/jail/id/{JAIL_NAME}/')
+    assert results.status_code == 200, results.text
+    assert results.json()['state'] == 'up', results.text
+
+
+def test_14_export_jail():
```
sonicaj: A jail must not be running when we export it, so that should be ensured here. I think it's running right now ?

tests/api2/jail.py
```diff
@@ -254,70 +220,50 @@ def test_17_stop_jail():
...
-        else:
-            results = GET('/jail/')
-            assert results.status_code == 200, results.text
-            assert len(results.json()) > 0, get_job.text
-            assert results.json()[0]['state'].lower() == 'down', results.text
-            break
-
-
-def test_19_start_jail():
+    job_status = wait_on_job(JOB_ID, 600)
```
sonicaj: 600 seconds is again too much to wait for stopping. Let's try 15/20 seconds ?

tests/api2/jail.py
```diff
@@ -254,70 +220,50 @@ def test_17_stop_jail():
...
+
+
+def test_20_start_jail():
     results = POST('/jail/start/', JAIL_NAME)
     assert results.status_code == 200, results.text
 
 
-def test_20_wait_for_jail_to_be_up():
+def test_21_wait_for_jail_to_be_up():
+    job_status = wait_on_job(JOB_ID, 600)
```
sonicaj: Same here as above

tests/api2/jail.py
```diff
@@ -213,33 +171,21 @@ def test_15_exec_call():
...
-        else:
-            results = job_status['result']
-            assert get_job.status_code == 200, results.text
-            assert 'exec successful' in results, results.text
-            break
-
-
-def test_17_stop_jail():
+    job_status = wait_on_job(JOB_ID, 300)
+    if job_status['state'] in ['TIMEOUT', "FAILED"]:
```
sonicaj: singe/double quotes clash ; )

tests/api2/jail.py
```diff
@@ -213,33 +171,21 @@ def test_15_exec_call():
...
+        freeze = True
+        freeze_msg = f"Failed to exec jail: {JAIL_NAME}"
```
sonicaj: Same here, i think we should be consistent in our usages

PR #3507, opened 2019-08-26
rm(gui): remove legacy snapshot/replication task scripts

william-gr:


themylogin:
@william-gr I've created corresponding JIRA task https://jira.ixsystems.com/browse/NAS-102953

PR #3508, opened 2019-08-26
NAS-103030 / 12 / enc_helper logic move

william-gr:


src/freenas/usr/local/sbin/enc_helper
```diff
@@ -56,116 +48,53 @@ if __name__ == "__main__":
...
+                    print("ERROR: passphrase not detected.")
+                    log.warn("ERROR: passphrase not detected.")
+            elif cmd == 'syncfrompeer':
                 passphrase = c.call('failover.call_remote', 'failover.encryption_getkey')
-            rv = escrowctl.setkey(passphrase)
-        if not rv:
-            print("Error")
+                rv = c.call('failover.encryption_setkey', passphrase)
+            if not rv:
+                print("Error")
```
yocalebo: Can we print a reasonable error here and state something like "ERROR: failed to set passphrase on this controller" or something like that?

william-gr: Sure, updated

PR #3511, opened 2019-08-28
NAS-102958 / 11.3 / Safely delete object via api v1

sonicaj:
This commit fixes an issue where we did not safely delete an object via api v1 when the model had no model form associated.


sonicaj:
backport

PR #3515, opened 2019-08-29
NAS-102979 / 11.3 / Bug fix for updating VM devices via VM service

sonicaj:
This commit fixes a bug where when we validated a device while updating a VM, we used the vm instance which was saved in the database and not the one which had just been provided. This resulted in checks being performed on stale data and giving wrong results.

### DIFF
--- a/src/middlewared/middlewared/plugins/vm.py
+++ b/src/middlewared/middlewared/plugins/vm.py
@@ -870,7 +870,9 @@ class VMService(CRUDService):
         devices_ids = {d['id']: d for d in await self.middleware.call('vm.device.query')}
         for i, device in enumerate(data.get('devices') or []):
             try:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1440,8 +1442,11 @@ def vnc_bind_choices(self):
                 )
 
     @private
-    async def validate_device(self, device, old=None):
-        vm_instance = await self.middleware.call('vm._get_instance', device['vm']) if device.get('vm') else None
+    async def validate_device(self, device, old=None, vm_instance=None):
+        # We allow vm_instance to be passed for cases where VM devices are being updated via VM and
+        # the device checks should be performed with the modified vm_instance object not the one db holds
+        vm_instance = await self.middleware.call('vm._get_instance', device['vm']) \
```
themylogin: Please rewrite it using regular if/else, this looks very confusing

PR #3516, opened 2019-08-29
sqlalchemy

themylogin:


themylogin:
@william-gr good, thank you for noticing this now. I am just working on HA and I was hoping not to make any business logic changes, but that very change will make our life a lot simpler if I get everything right.

src/middlewared/middlewared/plugins/datastore.py
```diff
@@ -1,74 +1,160 @@
...
+    return reg.search(item) is not None
 
 
 class DatastoreService(Service):
 
     class Config:
         private = True
 
-    def _filters_to_queryset(self, filters, field_prefix=None):
+    thread_pool = ThreadPoolExecutor(1)
```
william-gr: Why not thread_pool within Config, so its done automatically?

src/middlewared/middlewared/plugins/datastore.py
```diff
@@ -1,74 +1,160 @@
...
+        sql = compiled.string
+        binds = []
+        for param in compiled.positiontup:
+            bind = compiled.binds[param]
+            value = bind.value
+            bind_processor = compiled.binds[param].type.bind_processor(self.engine.dialect)
+            if bind_processor:
+                binds.append(bind_processor(value))
+            else:
+                binds.append(value)
```
william-gr: Is all of this safe/fast to run in a coroutine?

src/middlewared/middlewared/plugins/datastore.py
```diff
@@ -145,245 +338,227 @@ def query(self, name, filters=None, options=None):
...
-        except OperationalError as err:
-            raise CallError(err)
-        finally:
-            cursor.close()
-        return rv
-
-    @accepts(List('queries'))
-    def restore(self, queries):
+    @private
+    async def get_backrefs(self, name):
```
william-gr: Seems there is nothing here that grants it to be a coroutine, just possibly a bunch of python computation?

nas_ports/freenas/py-middlewared/Makefile
```diff
@@ -54,6 +54,23 @@ RUN_DEPENDS=	${PYTHON_PKGNAMEPREFIX}ws4py>0:www/py-ws4py@${PY_FLAVOR} \
...
+		${PYTHON_PKGNAMEPREFIX}licenselib>0:freenas/py-licenselib@${PY_FLAVOR} \
+		${PYTHON_PKGNAMEPREFIX}simplejson>0:devel/py-simplejson@${PY_FLAVOR} \
+		${PYTHON_PKGNAMEPREFIX}sqlparse>0:databases/py-sqlparse@${PY_FLAVOR} \
+		${PYTHON_PKGNAMEPREFIX}pycryptodome>0:security/py-pycryptodome@${PY_FLAVOR} \
+		${PYTHON_PKGNAMEPREFIX}openssl>0:security/py-openssl@${PY_FLAVOR} \
+		${PYTHON_PKGNAMEPREFIX}dnspython>0:dns/py-dnspython@${PY_FLAVOR} \
+		${PYTHON_PKGNAMEPREFIX}lxml>0:devel/py-lxml@${PY_FLAVOR} \
+		${PYTHON_PKGNAMEPREFIX}sysctl>0:devel/py-sysctl@${PY_FLAVOR} \
+		${PYTHON_PKGNAMEPREFIX}ipaddr>0:devel/py-ipaddr@${PY_FLAVOR} \
+		${PYTHON_PKGNAMEPREFIX}django110>0:www/py-django110@${PY_FLAVOR} \
```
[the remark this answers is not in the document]
themylogin: To run migrations from e.g. 11.2 to 11.3

src/freenas/etc/netcli
```diff
@@ -36,37 +36,18 @@ import subprocess
...
-
-from django.utils.translation import ugettext as _
-
 _ = lambda s: s
 
 from middlewared.client import Client
-from freenasUI.choices import NICChoices, LAGGType
-from freenasUI.common.system import get_sw_version
-from freenasUI.network.models import GlobalConfiguration, Interfaces, \
-	LAGGInterface, LAGGInterfaceMembers, StaticRoute, VLAN
```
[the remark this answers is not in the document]
themylogin: I've found out about netcli 2 days ago because previously I was only searching *.py files. Guess netcli adaptation will have to be another ticket.

src/middlewared/middlewared/plugins/datastore.py
themylogin: I don't want to run all service methods in this thread pool, it's just variable with same name.

src/middlewared/middlewared/plugins/datastore.py
themylogin: It will be a few (every ore table has maximum "a few" fields of non-trivial type) datetime/json parse/serialize, I think it's fine.

src/middlewared/middlewared/sqlalchemy.py
```diff
@@ -0,0 +1,150 @@
...
+    Boolean, CHAR, DateTime, Integer, SmallInteger, String, Text,
+)  # noqa
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import relationship  # noqa
+from sqlalchemy.types import UserDefinedType
+
+from middlewared.plugins.pwenc import encrypt, decrypt
+
+
+def django_ix(index, table):
```
william-gr: Can you give a quick explanation for dummies why that is needed to be kept in code and not migrated once?

themylogin: If we don't do this, a big useless migration will appear that will re-create indexes. We have to choose one of the evils and I guess I've chosen the wrong one?

william-gr: Well, from my point of view it seems to be more sane to have one big useless migration to keep the original code cleaner without any mentions of "django" : P
Thats just how I see it though, I wont mind too much if you dont think its worth it.

src/middlewared/middlewared/plugins/datastore.py
```diff
@@ -145,245 +338,227 @@ def query(self, name, filters=None, options=None):
...
-            options = {}
-        else:
-            # We do not want to make changes to original options
-            # which might happen with "prefix"
-            options = options.copy()
+        table = self._get_table(name)
+
+        # We do not want to make changes to original options
+        # which might happen with "prefix"
+        options = options.copy()
```
sonicaj: `options` can be `null`. We would want to ensure it's still a dict

src/middlewared/middlewared/plugins/datastore.py
```diff
@@ -145,245 +338,227 @@ def query(self, name, filters=None, options=None):
...
+                    columns.extend(list(alias.c))
+                    from_ = from_.outerjoin(alias, alias.c[foreign_key.column.name] == foreign_key.parent)
 
-        extra = options.get('extra')
-        if extra:
-            qs = qs.extra(**extra)
+            qs = select(columns).select_from(from_)
 
-        prefix = options.get('prefix')
+        prefix = options['prefix']
```
sonicaj: If `options` is `null` and we recreate the dict, it wouldn't have `prefix`. Same is true for below usages

src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -374,6 +409,17 @@ class Config:
         return result
 
 
+class iSCSITargetAuthCredentialModel(sa.Model):
+    __tablename__ = 'services_iscsitargetauthcredential'
+
+    id = sa.Column(sa.Integer(), primary_key=True)
+    iscsi_target_auth_tag = sa.Column(sa.Integer(), default=1)
+    iscsi_target_auth_user = sa.Column(sa.String(120))
+    iscsi_target_auth_secret = sa.Column(sa.String(120), default=None)
```
sonicaj: Do we have to explicitly define `nullable` here or just specifying the `default` as `None` is good ?

src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -374,6 +409,17 @@ class Config:
...
+    iscsi_target_auth_peeruser = sa.Column(sa.String(120))
+    iscsi_target_auth_peersecret = sa.Column(sa.String(120), default=None)
```
sonicaj: Same as above

src/middlewared/middlewared/plugins/keychain.py
```diff
@@ -218,6 +219,15 @@ def process_ssh_keyscan_output(output):
     return [" ".join(line.split()[1:]) for line in output.split("\n") if line and not line.startswith("# ")][-1]
 
 
+class KeychainCredentialModel(sa.Model):
+    __tablename__ = 'system_keychaincredential'
+
+    id = sa.Column(sa.Integer(), primary_key=True)
+    name = sa.Column(sa.String(255), default=6)
```
sonicaj: Default of `6` ?

src/middlewared/middlewared/plugins/keychain.py
```diff
@@ -218,6 +219,15 @@ def process_ssh_keyscan_output(output):
...
+    type = sa.Column(sa.String(255), default='Mail')
+    attributes = sa.Column(sa.JSON(encrypted=True), default=False)
```
sonicaj: `False` as a default, is this required ?

src/middlewared/middlewared/plugins/keyvalue.py
```diff
@@ -1,6 +1,15 @@
...
 from middlewared.schema import Any, Str, accepts
 from middlewared.service import Service
+import middlewared.sqlalchemy as sa
+
+
+class KeyValueModel(sa.Model):
+    __tablename__ = 'system_keyvalue'
+
+    id = sa.Column(sa.Integer(), primary_key=True)
+    key = sa.Column(sa.String(255), default="http")
```
sonicaj: Do we need `http` as a default for `key` ?

src/middlewared/middlewared/plugins/migration.py
```diff
@@ -17,6 +18,13 @@ def load_migrations(middleware):
     return sorted(modules, key=lambda module: module.__name__)
 
 
+class MigrationModel(sa.Model):
+    __tablename__ = 'system_migration'
+
+    id = sa.Column(sa.Integer(), primary_key=True)
+    name = sa.Column(sa.String(255), default=6)
```
sonicaj: `6` as a default for `name` ?

src/middlewared/middlewared/plugins/support.py
```diff
@@ -9,11 +9,27 @@
...
 
 ADDRESS = 'support-proxy.ixsystems.com'
 
 
+class SupportModel(sa.Model):
+    __tablename__ = 'system_support'
+
+    id = sa.Column(sa.Integer(), primary_key=True)
+    enabled = sa.Column(sa.Boolean(), nullable=True, default=True)
+    name = sa.Column(sa.String(200), default=6)
```
sonicaj: `6` as a default for name ?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -666,6 +667,33 @@ def post_stop_vm(self, *args, **kwargs):
         self.web_process = None
 
 
+class VMModel(sa.Model):
+    __tablename__ = 'vm_vm'
+
+    id = sa.Column(sa.Integer(), primary_key=True)
+    name = sa.Column(sa.String(150), default=6)
```
sonicaj: `6` as a default for name ?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -666,6 +667,33 @@ def post_stop_vm(self, *args, **kwargs):
...
+    threads = sa.Column(sa.Integer(), default=1)
+    shutdown_timeout = sa.Column(sa.Integer(), default=90)
+
+
+class VMDeviceModel(sa.Model):
+    __tablename__ = 'vm_device'
+
+    id = sa.Column(sa.Integer(), primary_key=True)
+    dtype = sa.Column(sa.String(50))
+    attributes = sa.Column(sa.JSON(), default=False)
```
sonicaj: `False` as a default, is this required/beneficial ?

src/middlewared/middlewared/plugins/datastore.py
```diff
@@ -1,74 +1,160 @@
...
+    def _get_pk(self, table):
+        return [col for col in table.c if col.primary_key][0]
+
+    def _get_col(self, table, name, prefix=None):
+        col = self._get_col_by_django_name(table, name)
+        if col is not None:
+            return col
+
+        if prefix:
+            col = self._get_col_by_django_name(table, prefix + name)
```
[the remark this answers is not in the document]
themylogin: Unfortunately we have a few models like this

```
class SMBModel(sa.Model):
    __tablename__ = 'services_cifs'

    ...
    cifs_srv_bindip = sa.Column(sa.MultiSelectField(), nullable=True)
    cifs_SID = sa.Column(sa.String(120), nullable=True)
```

src/middlewared/middlewared/plugins/datastore.py
```diff
@@ -77,31 +163,137 @@ def _filters_to_queryset(self, filters, field_prefix=None):
...
+            extend_context_value = await self.middleware.call(extend_context)
         else:
             extend_context_value = None
-        for i in qs:
-            yield django_modelobj_serialize(self.middleware, i, extend=extend, extend_context=extend_context,
-                                            extend_context_value=extend_context_value, field_prefix=field_prefix,
-                                            select=select)
+
+        result = []
+        for i, row in enumerate(qs):
```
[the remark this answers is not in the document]
themylogin: We don't have such tables and most likely will never have, but what problem do you see with `_queryset_serialize` being a coroutine? It calls `await self._serialize` in this loop so it won't process 1000 records without giving other coroutines their CPU share.

src/middlewared/middlewared/plugins/datastore.py
themylogin: Addressed in fae679541e0f3e96d0d6e247eb565ba223deca69

william-gr: Did you inspect all .query usages that we dont use "null" anywhere? Nor UI? We cant change API out of the blue just because something before "didnt make sense" :P

src/middlewared/middlewared/plugins/datastore.py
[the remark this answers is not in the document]
sonicaj: I was curious and wanted to know your opinion about it. We are good, thank you

PR #3518, opened 2019-08-30
NAS-102992 / 11.3 / Fix ldap failover for SMB

anodos325:
- properly configure idmap settings and ldap.conf in presence of multiple LDAP URIs
- add support to get nslcd status from nslcd command socket
- fix DS cache format for LDAP and NIS


src/middlewared/middlewared/etc_files/local/nslcd.conf
```diff
@@ -3,50 +3,50 @@
...
                 else:
                     certpath = cert['certificate_path']
         else:
             ldap = None
 
-        ldap_enabled = ldap['ldap_enable']
+        ldap_enabled = ldap['enable']
 
-        ldap_uri = "%s://%s" % ("ldaps" if ldap['ldap_ssl'] == "on" else "ldap", ldap['ldap_hostname'])
+        ldap_uri = ["%s://%s" % ("ldaps" if ldap['ssl'] == "on" else "ldap", x) for x in ldap['hostname']]
```
william-gr: According to ldap.update, ssl should be ON, OFF or START_TLS

Is .config returning lowercase? if so we should fix that

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -28,6 +34,87 @@ class SSL(enum.Enum):
     USETLS = 'START_TLS'
 
 
+class NlscdConst(enum.Enum):
+    NSLCD_CONF_PATH = '''/usr/local/etc/nslcd.conf'''
+    NSLCD_PIDFILE = '''/var/run/nslcd.pid'''
+    NSLCD_SOCKET = '''/var/run/nslcd/nslcd.ctl'''
```
william-gr: Why triple quote? thats odd : P

anodos325: It's in sample tools in nss-pam-ldapd. It bugs me too so I'll remove it.

src/middlewared/middlewared/etc_files/local/nslcd.conf
anodos325: Ah, that error crept in due to switching from datastore.query to ldap.config. Good catch.

src/middlewared/middlewared/etc_files/local/nss_ldap.conf
```diff
@@ -6,38 +6,33 @@
...
+ssl ${ldap['ssl']}.lower()
         % if certpath:
 tls_cacert ${certpath}
         % endif
 tls_reqcert allow
     % endif
 scope sub
-timelimit 30
+timelimit ${ldap['timeout']}
 bind_timelimit 30
```
william-gr: In nslcd.conf dns_timeout was used. Is this on purpose?

src/middlewared/middlewared/etc_files/local/openldap/ldap.conf
```diff
@@ -12,40 +22,46 @@
...
 % if (ldap_enabled and ldap) or (ad_enabled and ad):
+# This file is used by Samba and py-ldap. If NETWORK_TIMEOUT is too high, then ldap failover
+# in Samba's ldapsam passdb backend may not occur.
 URI ${uri}
 BASE ${base}
+NETWORK_TIMEOUT ${network_timeout}
+TIMEOUT ${timeout}
     % if ssl:
-SSL ${ssl}
+# SSL ${ssl}
```
william-gr: Is this comment also on purpose?

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -493,6 +580,19 @@ def get_dn(self, dn=None, ldap=None):
     async def __set_state(self, state):
         await self.middleware.call('cache.put', 'LDAP_State', state.name)
 
+    @private
+    def get_nslcd_status(self):
+        """
+        Returns internal nslcd state. nslcd will preferentially use the first LDAP server,
+        and only failover if the current LDAP server is unreachable.
+        """
+        ctx = NslcdClient(NlscdConst.NSLCD_ACTION_STATE_GET.value)
```
william-gr: Not important, but it seems a context manager for NslcdClient would be a perfect fit for this use case : P

anodos325: Good idea. I'll do that.

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -28,6 +34,93 @@ class SSL(enum.Enum):
...
+        # write a request header with a request code
+        self.action = action
+        self.write_int32(NlscdConst.NSLCD_VERSION.value)
+        self.write_int32(action)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, typ, value, traceback):
+        self.close()
```
william-gr: Actually we should prob check for an exception and raising it after close

william-gr: Hm, I may be having a brainfart, I feel we talked about this before. Is the exception raised by default?

PR #3529, opened 2019-09-03
fix(autotune): do not use freenasUI

william-gr:


src/autotune/files/autotune.py
```diff
@@ -15,34 +15,21 @@
...
 
-os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'freenasUI.settings')
-
-import django
-django.setup()
-
-from freenasUI.middleware.notifier import notifier
-n = notifier()
-if n.is_freenas():
+c = Client()
```
sonicaj: We should close it ?

william-gr: i thought about it, but its not a long running script, so it should be fine and be taken care of during exit.

PR #3544, opened 2019-09-06
NAS-103051 / 11.3 / Fix ldaps configuration in AD and LDAP enviornments

anodos325:
Affects python ldap config, nslcd.conf, and ldap.conf.
Ensure ca-root-nss.crt is included in ldap CACERTFILE if no custom CAs are on system, otherwise use user-provided ones.
Use cert provided in 'Certificate' dropdown for certificate-based authentication.
Add new parameter "validate_certificates" to Active Directory and LDAP.
Default to using GSSAPI bind in AD environments to meet Samba DC requirement of "strong auth".
Remove unused fields from LDAP model.


gui/directoryservice/migrations/0015_add_ldaps_validation.py
```diff
@@ -0,0 +1,56 @@
...
+    for o in AD.objects.all():
+        if o.ad_dns_timeout == 60:
+            o.ad_dns_timeout = 10
+
+        if o.ad_timeout == 60:
+            o.ad_timeout = 10
+
+        if not o.ad_enable:
+            o.ad_ssl = 'on'
+            o.ad_ldap_sasl_wrapping = 'sign'
```
yocalebo: I'm glad we talked about this! This looks good, however, I'm interested to know why we're defaulting to enabling LDAPS for AD based on the justification that it's needed for Samba AD related services. I thought that functionality was removed from freeNAS?

src/middlewared/middlewared/etc_files/generate_ssl_certs.py
```diff
@@ -19,9 +20,23 @@ def write_certificates(certs):
             with open(cert['csr_path'], 'w') as f:
                 f.write(cert['CSR'])
 
+    """
+    Write unified CA certificate file for use with LDAP.
+    """
+    shutil.copyfile('/usr/local/share/certs/ca-root-nss.crt',
```
freqlabs: What are all these extra certs needed for? Shouldn't the uploaded certificates be enough to trust the user's CA?

gui/directoryservice/migrations/0015_add_ldaps_validation.py
anodos325: FreeNAS can't be an AD DC, but it can still join Samba AD domains.

src/middlewared/middlewared/etc_files/generate_ssl_certs.py
freqlabs: Perhaps only do this when the user has not provided their own CA certificates. I just don't like the idea of being forced to trust someone else's CA when I have intentionally created my own root of trust.

gui/directoryservice/migrations/0015_add_ldaps_validation.py
yocalebo: This recent change resolves my question. Thanks!

src/middlewared/middlewared/etc_files/generate_ssl_certs.py
anodos325: Done.

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -619,57 +678,105 @@ class Config:
...
+        for the NAS. This is a recommended parameter for TrueNAS HA servers.
+
+        `disable_freenas_cache` disable active caching of AD users and groups.
+        When disabled, only users cached in winbind's internal cache will be
+        visible in GUI dropdowns. Disabling active caching is recommended
+        environments with a large amount of users.
+
+        `site` AD site of which the NAS is a member. This parameter is auto-
+        detected during the domain join process. If no AD site is configured
+        for the subnet in which the NAS is configured, then this parameter will
```
a-stjohn:
```suggestion
        for the subnet in which the NAS is configured, then this parameter
```

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -619,57 +678,105 @@ class Config:
...
+        appear as 'Default-First-Site-Name'. Auto-detection is only performed
```
a-stjohn:
```suggestion
        appears as 'Default-First-Site-Name'. Auto-detection is only performed
```

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -619,57 +678,105 @@ class Config:
...
+        during the initial domain join.
+
+        `kerberos_realm` in which the server is located. This parameter will be
```
a-stjohn:
```suggestion
        `kerberos_realm` in which the server is located. This parameter is
```

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -619,57 +678,105 @@ class Config:
...
+        automatically populated during the initial domain join. If the NAS has
+        an AD site configured and that site has multiple kerberos servers, then
+        the kerberos realm will be automatically updated with a site-specific
```
a-stjohn:
```suggestion
        the kerberos realm is automatically updated with a site-specific
```

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -619,57 +678,105 @@ class Config:
...
+        configuration to use those servers. Auto-detection is only performed
+        during initial domain join.
+
+        `kerberos_principal` kerberos principal to use for AD-related
+        operations outside of Samba. After intial domain join, this field will
```
a-stjohn:
```suggestion
        operations outside of Samba. After intial domain join, this field is
```

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -619,57 +678,105 @@ class Config:
...
+        be updated with the kerberos principal associated with the AD machine
```
a-stjohn:
```suggestion
        updated with the kerberos principal associated with the AD machine
```

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -619,57 +678,105 @@ class Config:
...
+
+        `ldap_sasl_wrapping` defines whether ldap traffic will be signed or
+        signed and encrypted (sealed). LDAP traffic that does not originate
+        from Samba defaults to using GSSAPI signing unless it is tunnelled
+        over LDAPs.
+
+        `createcomputer` Active Directory Organizational Unit in which to
+        create the NAS computer object during domain join.
+
+        If blank, then the default OU is used during computer account creation.
```
a-stjohn:
```
        If blank, then the default OU is used during computer account creation.
```

Same as the beginning. I would expand this first mention of the `OU` acronym. I'm not sure what it is.

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -619,57 +678,105 @@ class Config:
...
+        Precreate the computer account in a specific OU. The OU string read from
+        top to bottom without RDNs and delimited by a "/". For example,
+        "createcomputer=Computers/Servers/Unix. Backslashes "\" are used as an
+        escape character, and not as a separator.
+
+        `idmap_backend` provides a plugin interface for Winbind to use varying
+        backends to store SID/uid/gid mapping tables. The correct setting
+        depends on the environment in which the NAS is deployed.
+
+        The Active Directory service will be started after a configuration
```
a-stjohn:
```suggestion
        The Active Directory service is started after a configuration
```

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -619,57 +678,105 @@ class Config:
...
+        update if the service was initially disabled, and the updated
+        configuration sets `enable` to `True`. Likewise, the Active Directory
```
a-stjohn:
```suggestion
        configuration sets `enable` to `True`. The Active Directory
```

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -619,57 +678,105 @@ class Config:
...
+        service will be stopped if `enable` is changed to `False`. If the
```
a-stjohn:
```suggestion
        service is stopped if `enable` is changed to `False`. If the
```

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -619,57 +678,105 @@ class Config:
...
+        configuration is updated, but the initial `enable` state is `True`, and
+        remains unchanged, then the samba server will only be restarted.
```
a-stjohn:
```suggestion
        remains unchanged, then only the samba server is restarted.
```

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -619,57 +678,105 @@ class Config:
...
+
+        During the domain join, a kerberos keytab for the newly-created AD
+        machine account will be generated. It is used for all future
```
a-stjohn:
```suggestion
        machine account is generated. It is used for all future
```

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -414,8 +439,62 @@ class Config:
...
+        format.
+
+        `binddn` specifies the default bind DN to use when performing ldap
+        operations. The bind DN must be specified as a Distinguished Name in
+        LDAP format.
+
+        `anonbind` use anonymous authentication.
+
+        `ssl` establish SSL/TLS-protected connections to the LDAP server(s).
+        GSSAPI signing will be disabled on SSL/TLS-protected connections if
```
a-stjohn:
```suggestion
        GSSAPI signing is disabled on SSL/TLS-protected connections if
```

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -414,8 +439,62 @@ class Config:
...
+        if a bad certificate is provided, the session is immediately terminated.
+        If disabled, TLS_REQCERT allow is set. The server certificate is
+        requested, but all errors are ignored.
+
+        `kerberos_realm` in which the server is located. This parameter is
+        only required for SASL GSSAPI authentication to the remote LDAP server.
+
+        `kerberos_principal` kerberos principal to use for SASL GSSAPI
+        authentication to the remote server. If `kerberos_realm` is specified
+        without a keytab, then the `binddn` and `bindpw` will be used to
```
a-stjohn:
```suggestion
        without a keytab, then the `binddn` and `bindpw` are used to
```

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -414,8 +439,62 @@ class Config:
...
+        perform to obtain the ticket necessary for GSSAPI authentication.
+
+        `timeout` specifies  a  timeout  (in  seconds) after which calls to
+        synchronous LDAP APIs will abort if no response is received.
+
+        `dns_timeout` specifies the timeout (in seconds) after which the
+        poll(2)/select(2) following a connect(2) returns in case of no activity.
```
a-stjohn:
```suggestion
        poll(2)/select(2) following a connect(2) returns in case of no activity
```

I think there's an extra period there on the end?

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -414,8 +439,62 @@ class Config:
...
+        for openldap. For nslcd this specifies the time limit (in seconds) to
+        use when connecting to the directory server. This directly impacts the
+        length of time that the LDAP service will try before failing over to
```
a-stjohn:
```suggestion
        length of time that the LDAP service tries before failing over to
```

william-gr:
@sonicaj can you review this, please?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -905,12 +1010,44 @@ def validate_credentials(self, ad=None):
...
+                capture_output=True
+            )
+            if kinit.returncode != 0:
+                realm = self.middleware.call_sync(
+                    'kerberos.realm.query',
+                    [('realm', '=', ad['domainname'])],
+                    {'get': True}
+                )
+                self.middleware.call_sync('kerberos.realm.delete', realm['id'])
+                raise CallError(f"kinit for domain [{ad['domainname']}] with password failed: {kinit.sderr.decode()}")
```
sonicaj: I believe this should be `stderr` ?

src/middlewared/middlewared/etc_files/generate_ssl_certs.py
```diff
@@ -19,9 +20,24 @@ def write_certificates(certs):
...
                 f.write(cert['CSR'])
 
+    """
+    Write unified CA certificate file for use with LDAP.
+    """
+    if not cacerts:
+        shutil.copyfile('/usr/local/share/certs/ca-root-nss.crt',
+                        '/etc/ssl/truenas_cacerts.pem')
+    else:
+        with open('/etc/ssl/truenas_cacerts.pem', 'a+') as f:
```
freqlabs: Is append still intended here? I suspect it should be truncating the file instead.

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -619,57 +678,105 @@ class Config:
...
-        during an update. If the configuration is updated, but the initial :enable: state was True, then only a samba_server restart
-        command will be issued.
+        `use_default_domain` controls whether domain users and groups will have
+        the pre-windows 2000 domain name prepended to the user account. When
+        enabled, the user will appear as "administrator" rather than
+        "EXAMPLE\administrator"
+
+        `allow_trusted_doms` enable support for trusted domains. If this
+        parameter is enabled, then separate idmap backends _must_ be configured
+        for each trusted domain, and the idmap cache should be cleared.
```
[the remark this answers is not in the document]
anodos325: 'should' implies that the action is recommended. 'must' implies that it is mandatory, which is not the case.

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: Indeed. Thank you.

src/middlewared/middlewared/etc_files/generate_ssl_certs.py
anodos325: No, it is no longer intended.

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: Indeed, it should. Thank you for catching that.

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -230,54 +227,115 @@ def _open(self):
...
-                        self.logger.debug('%s', e)
+                        saved_bind_error = e
+                        self.logger.debug('Failed to initialize start_tls: %s', e)
+                        continue
 
-                if self.ad['kerberos_principal']:
+                if self.ad['certificate'] and SSL(self.ad['ssl']) != SSL.NOSSL:
+                    """
+                    Active Directory permits two means of establishing an
+                    SSL/TLS-protected connection to a DC. The first is by
```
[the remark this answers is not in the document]
anodos325: This one is not user-facing, and is a direct quote of MS documentation so I'll keep it as-is.

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: I think Warren wrote up something that put this in understandable language. So this entry should probably be rewritten using the authorized text.

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: That changes the meaning of the statement. I will remove its 'will', but will maintain the same meaning.

PR #3545, opened 2019-09-09
Clear pool labels from new partitions

freqlabs:
When reusing an old disk, there may be leftover pool labels in strange
places. Clear out any pre-existing labels when formatting disks with new
partitions in order to prevent import/mountroot errors at boot.

Ticket: NAS-103070


freqlabs:
This was discussed after the OS/Services meeting this week.

PR #3549, opened 2019-09-10
NAS-103081 / 12 / feat(middlewared/boot): allow different boot pool name

william-gr:
Ticket:	NAS-103081

### DIFF
--- a/src/middlewared/middlewared/plugins/boot.py
+++ b/src/middlewared/middlewared/plugins/boot.py
@@ -9,21 +9,29 @@ from middlewared.validators import Range
 from bsd import geom
 
 
+BOOT_POOL_NAME = None
+BOOT_POOL_NAME_VALID = ['freenas-boot', 'boot-pool']
+
+
 class BootService(Service):
+
+    @private
+    async def pool_name(self):
+        return BOOT_POOL_NAME
 
     @accepts()
     async def get_state(self):
         """
         Returns the current state of the boot pool, including all vdevs, properties and datasets.
         """
-        return await self.middleware.call('zfs.pool.query', [('name', '=', 'freenas-boot')], {'get': True})
+        return await self.middleware.call('zfs.pool.query', [('name', '=', BOOT_POOL_NAME)], {'get': True})
 
     @accepts()
     async def get_disks(self):
         """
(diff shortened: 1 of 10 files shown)


src/middlewared/middlewared/plugins/boot.py
```diff
@@ -257,3 +265,17 @@ def get_swap_size(self, disk):
...
+
+    pools = (
+        await run('zpool', 'list', '-H', '-o', 'name', encoding='utf8')
+    ).stdout.strip().split()
+    for i in BOOT_POOL_NAME_VALID:
+        if i in pools:
+            BOOT_POOL_NAME = i
+            break
+    else:
+        middleware.logger.error('Failed to detect boot pool name.')
```
themylogin: Maybe a middleware shutdown should be performed in this case?

william-gr: I dont think so. Functionality will be limited and things will fail but it will mostly work.

src/middlewared/middlewared/plugins/boot.py
```diff
@@ -9,21 +9,29 @@
...
 
+BOOT_POOL_NAME = None
+BOOT_POOL_NAME_VALID = ['freenas-boot', 'boot-pool']
+
+
 class BootService(Service):
 
+    @private
+    async def pool_name(self):
+        return BOOT_POOL_NAME
```
themylogin: Can we at least throw an exception here if `BOOT_POOL_NAME` is None then to prevent unexpected things from happening?

william-gr: Also no, I want to be able to run middlewared outside a FreeNAS install

themylogin: This exception won't prevent you from running middleware outside FN, but will prevent you from running boot pool related tasks, no?

william-gr: boot.pool_name is called on other places as well, not only boot pool tasks, e.g. system dataset

william-gr: And lets face it, the change of this being None in a real FreeNAS install is 0.0000000001% : )

PR #3552, opened 2019-09-11
NAS-103080 / 11.3 / Fix too verbose smart log

themylogin:


src/middlewared/middlewared/plugins/disk.py
```diff
@@ -764,7 +764,7 @@ def decrypt(self, job, devices, passphrase=None):
         cp = await run(['smartctl', '-a'] + smartctl_args, check=False, stderr=subprocess.STDOUT,
                        encoding='utf8', errors='ignore')
         if (cp.returncode & 0b11) != 0:
-            self.logger.warning('Failed to run smartctl for %r (%r): %s', disk, smartctl_args, cp.stdout)
+            self.logger.trace('Failed to run smartctl for %r (%r): %s', disk, smartctl_args, cp.stdout)
```
william-gr: What is your opinion about checking for the low-power mode alone? Seems it could be still be useful when a device fails  to open or did not identify. If not perhaps ignore Bit 1 completely

Bit 2 at least seems to have sense to use warning

themylogin: The problem is that device is checked every 5 minutes with collectd and log will be flooded.
In case we get debug from the user, smartctl outputs for failing drives would be collected by debug script, so I think we can afford not to have this in middlewared log.

PR #3553, opened 2019-09-11
Clear pool labels from new partitions (11.2)

freqlabs:
When reusing an old disk, there may be leftover pool labels in strange places. Clear out any preexisting labels when formatting disks with new partitions in order to prevent import/mountroot errors at boot.

Also backported parts of the previous cleanup from 12 and 4k swap alignment from 11.3 as separate commits for diff reduction.

Ticket: NAS-103070

### DIFF
--- a/src/freenas-installer/etc/install.sh
+++ b/src/freenas-installer/etc/install.sh
@@ -399,7 +399,7 @@ create_partitions()
 	fi
 
 	if is_truenas; then
-	    gpart add -t freebsd-swap -s 16g -i 3 ${_disk}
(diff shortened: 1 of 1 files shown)


amotin:
Lets leave this to 11.3, or at very least delay after 11.2-U6 is out.  It does not look a dramatic problem to hurry now.

freqlabs:
11.2-U6 is out, ok to merge?

william-gr:
No, only after Truenas is out

freqlabs:
@william-gr can you merge this? I don't have write access to this branch.

PR #3554, opened 2019-09-11
NAS-103035 / 11.3 / Remove obsolete SMART E-Mail field

themylogin:


gui/services/migrations/0036_remove_smart_email.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def create_smart_alert_service(apps, schema_editor):
+    smart = apps.get_model("services.smart").objects.all()[0]
+    alertservice = apps.get_model('system.alertservice')
+    root_email = apps.get_model('account.bsdusers').objects.filter(bsdusr_username="root")[0].bsdusr_email
+    existing_emails = [s.attributes["email"] or root_email
+                       for s in alertservice.objects.filter(type="Mail")]
+    emails = [x.strip() for x in smart.smart_email.split(",") if x.strip()]
+    for new_email in set(emails) - set(existing_emails):
```
william-gr: Hm, that will email them for every alert source?

themylogin: Yes, these users will start receiving all alerts and we have removed a way to avoid this. I think that's better than someone not receiving smart alert.

PR #3560, opened 2019-09-12
NAS-103028 / 11.3 / Correctly validate disks for VMs

sonicaj:
This commit introduces changes to fix a bug which expected device id to be present when a device was validated in vm's data. However this is not the case always i.e when we create a VM, device id's haven't been initialized.
We used vm's data to ensure we didn't have duplicates for disk devices. Now apart from id's we have paths unique for disks, if we find a path more then once in a VM, that means it has been duplicated. We use this to validate disk devices now.

### DIFF
--- a/src/middlewared/middlewared/plugins/vm.py
+++ b/src/middlewared/middlewared/plugins/vm.py
@@ -1442,10 +1442,40 @@ class VMDeviceService(CRUDService):
                 )
 
     @private
+    async def disk_uniqueness_integrity_check(self, device, vm):
+        # This ensures that the disk is not already present for `vm`
+        def translate_device(dev):
+            # A disk should have a path configured at all times, when that is not the case, that means `dtype` is DISK
+            # and end user wants to create a new zvol in this case.
+            return dev['attributes'].get('path') or f'/dev/zvol/{dev["attributes"]["zvol_name"]}'
+
+        disks = [
+            d for d in vm['devices']
+            if d['dtype'] in ('DISK', 'RAW', 'CDROM') and translate_device(d) == translate_device(device)
+        ]
+        if not disks:
+            # We don't have that disk path in vm devices, we are good to go
+            return True
+        elif len(disks) > 1:
+            # VM is mis-configured
+            return False
+        elif not device.get('id') and disks:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1441,11 +1441,41 @@ def vnc_bind_choices(self):
...
+            return dev['attributes'].get('path') or f'/dev/zvol/{dev["attributes"]["zvol_name"]}'
+
+        disks = [
+            d for d in vm['devices']
+            if d['dtype'] in ('DISK', 'RAW', 'CDROM') and translate_device(d) == translate_device(device)
+        ]
+        if not disks:
+            # We don't have that disk path in vm devices, we are good to go
+            return True
+        elif len(disks) > 1:
```
william-gr: Why a VM cannot have more than 1 disk?

sonicaj: A VM cannot have more then 1 disk having the same path

william-gr: Can you show me where its being checked for same path?

sonicaj: https://github.com/freenas/freenas/pull/3560/files#diff-2624b1c1bd7ddfbcb5b2ddf3655d961eR1454

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1441,11 +1441,41 @@ def vnc_bind_choices(self):
...
+    async def disk_uniqueness_integrity_check(self, device, vm):
+        # This ensures that the disk is not already present for `vm`
+        def translate_device(dev):
+            # A disk should have a path configured at all times, when that is not the case, that means `dtype` is DISK
+            # and end user wants to create a new zvol in this case.
+            return dev['attributes'].get('path') or f'/dev/zvol/{dev["attributes"]["zvol_name"]}'
+
+        disks = [
+            d for d in vm['devices']
+            if d['dtype'] in ('DISK', 'RAW', 'CDROM') and translate_device(d) == translate_device(device)
```
sonicaj: Here

PR #3564, opened 2019-09-13
NAS-102725 / 11.3 / Handle BrokenPipeError in /_upload

themylogin:


src/middlewared/middlewared/main.py
```diff
@@ -506,22 +506,25 @@ def do_copy():
...
+                pass
 
         try:
             job = await self.middleware.call(data['method'], *(data.get('params') or []),
                                              pipes=Pipes(input=self.middleware.pipe()))
-            try:
-                await self.middleware.run_in_thread(copy)
-            finally:
-                await self.middleware.run_in_thread(job.pipes.input.w.close)
+            await self.middleware.run_in_thread(copy)
```
william-gr: Should we even wait for this coroutine to finish? Or just let it run in the "background" with ensure_future?

themylogin: Yes, we need, it's a request handler, we can't allow it to finish until request reading is complete.

Downside is that /_upload won't return job ID until job consumes input but for our current use cases it's OK.

william-gr: Derp, obviously. Stupid question :D

PR #3565, opened 2019-09-13
NAS-102615 / 11.3 / fix(middlwared/network): do not raise error if HA has not been setup yet

william-gr:
Ticket:	NAS-102615

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -577,7 +577,7 @@ class InterfaceService(CRUDService):
     async def __check_failover_disabled(self):
         if await self.middleware.call('system.is_freenas'):
             return
-        if not await self.middleware.call('failover.licensed'):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -577,7 +577,7 @@ def iface_extend(self, iface_state, configs, is_freenas, fake=False):
     async def __check_failover_disabled(self):
         if await self.middleware.call('system.is_freenas'):
             return
-        if not await self.middleware.call('failover.licensed'):
+        if await self.middleware.call('failover.status') == 'SINGLE':
```
sonicaj: I think we don't need `()` here ?

william-gr: correct

sonicaj: Perfect, thank you

PR #3576, opened 2019-09-17
NAS-103131 / 11.3 / Add validation for dyndns service

sonicaj:
This commit adds validation for dyndns service

### DIFF
--- a/src/middlewared/middlewared/plugins/dyndns.py
+++ b/src/middlewared/middlewared/plugins/dyndns.py
@@ -1,5 +1,5 @@
 from middlewared.schema import accepts, Bool, Dict, Int, List, Str
-from middlewared.service import SystemServiceService, private
+from middlewared.service import SystemServiceService, private, ValidationErrors
 
 
 class DynDNSService(SystemServiceService):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/dyndns.py
```diff
@@ -55,6 +55,23 @@ class Config:
...
+    @private
+    async def validate_data(self, data, schema):
+        verrors = ValidationErrors()
+        if data.get('provider') not in await self.provider_choices():
+            verrors.add(
+                f'{schema}.provider',
+                'Please select a valid provider.'
+            )
+
+        if not data.get('domain'):
```
themylogin: Why can't you use empty=False on the List schema?

sonicaj: My bad, i overlooked `empty` option for list schema. Thank you for letting me know

src/middlewared/middlewared/plugins/dyndns.py
```diff
@@ -55,6 +55,17 @@ class Config:
             'ipv6tb@he.net': 'he.net'
         }
 
+    @private
+    async def validate_data(self, data, schema):
+        verrors = ValidationErrors()
+        if data.get('provider') not in await self.provider_choices():
```
themylogin: Thank you, @william-gr, my bad, the provider `custom` also should be accepted; also, why `.get`?

src/middlewared/middlewared/plugins/dyndns.py
```diff
@@ -55,6 +55,25 @@ class Config:
...
+    @private
+    async def validate_data(self, data, schema):
+        verrors = ValidationErrors()
+        provider = data['provider']
+        if provider == 'custom':
+            for k in ('custom_ddns_server', 'custom_ddns_path'):
+                if not data[k]:
+                    verrors.add(
+                        f'{schema}.{k}',
+                        'Required when using custom provider.'
```
Mrt134:
```suggestion
                        'Required when using a custom provider.'
```

PR #3580, opened 2019-09-18
NAS-103108 / 12 / Unify Installer for FN/TN

sonicaj:
This PR aims to unify installer for FN/TN

### DIFF
--- a/src/freenas-installer/etc/install.sh
+++ b/src/freenas-installer/etc/install.sh
@@ -13,7 +13,8 @@ export TERM
 . /etc/avatar.conf
 
 # Boot Pool
-readonly BOOT_POOL="boot-pool"
+BOOT_POOL="boot-pool"
+NEW_BOOT_POOL="boot-pool"
 
 # Constants for base 10 and base 2 units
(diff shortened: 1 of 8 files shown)


src/freenas-installer/etc/install_sata_dom.sh
```diff
@@ -0,0 +1,148 @@
+#!/bin/sh
```
william-gr: I need to confirm that but it has been so long we dropped graid support I am thinking its time we remove this code.

src/freenas-installer/etc/install.sh
```diff
@@ -621,24 +614,24 @@ disk_is_freenas()
     fi
 
     # Make sure this is a FreeNAS boot pool.
-    zdb -l ${part} | grep -qF "name: 'freenas-boot'" || return 1
+    zdb -l ${part} | grep -qF "name: '${BOOT_POOL}'" || return 1
```
william-gr: For upgrades it may still be freenas-boot, we have to account for both.

PR #3588, opened 2019-09-19
NAS-102283 / 12 / Run periodic tasks when system is ready 

sonicaj:
This commit adds the ability to execute periodic tasks when the system is ready. Motivation for this is tasks which require network interaction, as middlewared boots early in the process, networking isn't setup. We still support  attribute but if  is enabled, that takes precedence over.

### DIFF
--- a/src/middlewared/middlewared/main.py
+++ b/src/middlewared/middlewared/main.py
@@ -849,7 +849,7 @@ class Middleware(LoadPluginsMixin):
 
         self.logger.debug('All plugins loaded')
 
-    def __setup_periodic_tasks(self):
+    def _setup_periodic_tasks(self):
         for service_name, service_obj in self.get_services().items():
             for task_name in dir(service_obj):
                 method = getattr(service_obj, task_name)
@@ -1350,7 +1350,8 @@ class Middleware(LoadPluginsMixin):
         await restful_api.register_resources()
         asyncio.ensure_future(self.jobs.run())
 
-        self.__setup_periodic_tasks()
+        if await self.call('system.state') == 'READY':
+            self._setup_periodic_tasks()
 
         # Start up middleware worker process pool
         self.__procpool._start_queue_management_thread()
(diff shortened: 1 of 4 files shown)


sonicaj:
I did consider that but then didn't go through because there might be tasks which don't require the system to be ready like processing alerts ( i think it doesn't require network ? ). If we do that, then these tasks would have to wait until the system has completely booted.

william-gr:
I dont think thats a problem, quite the contrary. I am not sure we want to process alerts before system boots and is in a "stable" state. It can also try to send alerts out which requires network.

sonicaj:
In that case, it makes most sense to only have one parameter here as you mentioned above. I'll update the PR, thank you

william-gr:
Lets wait and see what @themylogin thinks too

src/middlewared/middlewared/main.py
```diff
@@ -849,27 +849,51 @@ def sort_key(plugin__function):
 
         self.logger.debug('All plugins loaded')
 
-    def __setup_periodic_tasks(self):
+    async def __setup_periodic_tasks(self):
+        system_ready_tasks = []
+
+        async def execute_periodic_func(delay, method, service_name, service_obj, method_name, interval):
```
themylogin: There are no awaits inside, you can omit async def

src/middlewared/middlewared/main.py
```diff
@@ -849,27 +849,51 @@ def sort_key(plugin__function):
...
+            self.loop.call_later(
+                delay,
+                functools.partial(
+                    self.__call_periodic_task,
+                    method, service_name, service_obj, method_name, interval
+                )
+            )
+
+        async def system_read_task_event_handler(middleware, event_type, args):
+            if args['id'] == 'ready':
```
themylogin: Becomes a pattern, can we do something to allow subscribing to `system.ready`?

themylogin:
I agree with @william-gr, let's just do all run_on_start tasks when system is ready.

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -1144,7 +1144,7 @@ class Config:
                     args=','.join(share['share_acl'].split())
                 )
 
-    @periodic(3600, run_on_start=False)
+    @periodic(3600)
```
themylogin: Why did you change it here?

william-gr:
@themylogin I agree, should keep it as minimum as possible and merge back to 11.3

PR #3590, opened 2019-09-19
NAS-103157 / 11.3 / fix(middlewared/account): validate home directory on update as well

william-gr:
Ticket:	NAS-103157


src/middlewared/middlewared/plugins/account.py
```diff
@@ -624,6 +619,12 @@ def shell_choices(self):
                     '"Home Directory" must begin with /mnt/ or set to '
                     '/nonexistent.'
                 )
+            elif os.stat(data['home']).st_dev == os.stat('/mnt').st_dev:
```
sonicaj: If it doesn't exist yet ?

william-gr: true, will fix

PR #3593, opened 2019-09-20
NAS-100293 / 12 / Improve iscsi portal choices and validation

sonicaj:
This PR adds changes to introduce ipv6 wildcard address in the iscsi portal choices. Also we update the validation for iscsi portals to ensure that no dynamic address is configured with iscsi portals as it's possible that we might not have it when system boots resulting in ctld failure.


sonicaj:
Yup

sonicaj:
backport

PR #3595, opened 2019-09-20
NAS-103161 / 12 / Create eula pending file on license update

sonicaj:
This commit adds changes which when license is updated, ensure we create a eula pending file


william-gr:
I will think about this for a bit, will review next week

PR #3596, opened 2019-09-20
NAS-103143 / 11.3 / Update plugin index to automatically reflect TN

sonicaj:
This commit introduces changes which ensure that the default plugin index is automatically adjusted to match FN/TN.

### DIFF
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -145,19 +145,22 @@ class PluginService(CRUDService):
     class Config:
         process_pool = True
 
-    OFFICIAL_REPOSITORIES = {
-        'IXSYSTEMS': {
-            'name': 'iXsystems',
-            'git_repository': 'https://github.com/freenas/iocage-ix-plugins.git',
-        }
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/jail.py
```diff
@@ -244,6 +247,7 @@ def do_create(self, data):
         `branch` is the FreeNAS repository branch to use as the base for the `plugin_repository`. The default is to
         use the current system version. Example: 11.3-RELEASE.
         """
+        data['plugin_repository'] = data.get('plugin_repository', self.default_repo())
```
william-gr: data.get('plugin_repository') or self.default_repo()

otherwise default_repo is evaluated even if plugin_repository is passed

sonicaj: Is there a drawback to that ? If it is passed, the passed value will still be respected. Idea is if user passes empty string, we should raise an error

william-gr: The drawback is that we will be calling a function that will not be used

PR #3602, opened 2019-09-23
NAS-102932 / 11.3 / Prevent deadlock for jail plugin services

sonicaj:
This PR introduces the following changes:
1) We run `plugin.available` as a periodic task now.
2) We ensure that if a `plugin.available` job is executing without parameters ( i.e job locking on none ), and we call the same job, this time with default parameters ( i.e 11.3-RELEASE ), we wait for the first job to finish as they are basically the same thing.
3) Changes `jail` and `plugin` service to not execute in a process pool as py-libzfs has been removed for iocage.
4) Updates jail plugin to reflect changes outlined in `3`.

### DIFF
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -142,9 +142,6 @@ def common_validation(middleware, options, update=False, jail=None, schema='opti
 
 class PluginService(CRUDService):
 
-    class Config:
-        process_pool = True
-
     @accepts()
     async def official_repositories(self):
         """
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/job.py
```diff
@@ -254,6 +254,10 @@ def check_pipe(self, pipe):
         if getattr(self.pipes, pipe) is None:
             raise ValueError("Pipe %r is not open" % pipe)
 
+    def check(self, exception=CallError):
```
william-gr: Would you mind we remove this for now and keep the old way?

I had other plans to implement this, combining wait and error.

PR #3603, opened 2019-09-24
NAS-103210 / 11.3 / Force AD domain name to uppercase

anodos325:
Originally this was done in activedirectory.compress(),
but this should actually be done earlier so that it's
impossible to kinit with a lowercase realm name.


anodos325:
> That wont make already existing entries in the db use uppercase. That expected?

In this case it's fine since the kerberos.realm will already be populated in the db, and kerberos._kinit forces uppercase on the AD domainname. I didn't want to write another migration. :)

PR #3613, opened 2019-09-26
NAS-103228 / 12 / Ability to tell if memory is ECC based

sonicaj:
This commit adds ability to tell if system configured memory is ECC based or not.

### DIFF
--- a/src/middlewared/middlewared/plugins/system.py
+++ b/src/middlewared/middlewared/plugins/system.py
@@ -38,6 +38,8 @@ CACHE_POOLS_STATUSES = 'system.system_health_pools'
 FIRST_INSTALL_SENTINEL = '/data/first-boot'
 LICENSE_FILE = '/data/license'
 
+RE_MEMWIDTH = re.compile(r'Total Width:\s*(\d*)')
+
 
 class SystemAdvancedService(ConfigService):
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/system.py
```diff
@@ -453,6 +453,9 @@ def license_update(self, license):
             stdout=subprocess.PIPE,
         )).communicate())[0].decode().strip() or None
 
+        # https://superuser.com/questions/893560/how-do-i-tell-if-my-memory-is-ecc-or-non-ecc/893569#893569
+        ecc_results = re.findall(r'Total Width:\s*(\d*)', (await run(['dmidecode', '-t', '17'])).stdout.decode())
```
william-gr: Could you define/compile the regex at the top, please?

RE_MEMWIDTH = re.compile(r'Total Width:\s*(\d*)')

or something

src/middlewared/middlewared/plugins/system.py
```diff
@@ -38,6 +38,8 @@
 FIRST_INSTALL_SENTINEL = '/data/first-boot'
 LICENSE_FILE = '/data/license'
 
+ECC_REGEX = re.compile(r'Total Width:\s*(\d*)')
```
william-gr: Not to be pedantic. but on every plugin we do this we start it with RE_  :P

sonicaj: Updated :p

PR #3616, opened 2019-09-27
NAS-102449 / 12 / Plugin Updates Alerts

sonicaj:
This PR adds changes which raise an alert for a plugin which needs to be updated.


src/middlewared/middlewared/alert/base.py
```diff
@@ -94,6 +94,7 @@ class AlertCategory(enum.Enum):
     DIRECTORY_SERVICE = "DIRECTORY_SERVICE"
     HA = "HA"
     HARDWARE = "HARDWARE"
+    JAILS = "JAILS"
```
william-gr: Should it be PLUGINS instead?

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -501,6 +502,70 @@ def defaults(self, options):
...
+                self.middleware.logger.error(f'Failed to retrieve plugin versions for {repo}: {job_obj.error}')
+                git_repos[repo] = []
+            else:
+                git_repos[repo] = job_obj.result
+
+        for plugin in installed_plugins:
+            repo = plugin['plugin_repository']
+            plugin_dict = (list(filter(lambda d: d['name'] == plugin['plugin'], git_repos[repo])) or [{}])[0]
+            if not plugin_dict or any(
+                plugin_dict[k] == 'N/A' or plugin[k] == 'N/A' for k in ('version', 'revision', 'epoch')
```
william-gr: didnt we remove version from plugins json?

sonicaj: `plugin_dict` here is data retrieved from `plugin.available` and `plugin` here is data retrieved from `plugin.query`. Both are guaranteed to give pkg version/revision/epoch keys.

src/middlewared/middlewared/alert/source/jails.py
```diff
@@ -0,0 +1,20 @@
+from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, OneShotAlertClass, Alert
+
+
+class PluginUpdateAlertClass(AlertClass, OneShotAlertClass):
+    deleted_automatically = False
+
+    category = AlertCategory.PLUGINS
+    level = AlertLevel.INFO
+    title = 'Plugin Update Available'
+    text = 'Plugin "%(name)s" update is available. Go to ' \
```
wblock:
```suggestion
    text = 'An update is available for the "%(name)s" plugin. Go to ' \
```

src/middlewared/middlewared/alert/source/jails.py
```diff
@@ -0,0 +1,20 @@
...
+           'Plugins->Installed->"%(name)s"->Update and apply the update'
```
wblock: Corrected, but it is a mistake for the middleware to suggest where the user should go in the menus. That is a separate place to track and get out of sync when menus or buttons change. It would be better to be more vague here, leaving out the second sentence entirely.  The user can find the plugin update in the user interface by themselves.  Or if they can't, it's a separate issue we should fix there.
```suggestion
           'Plugins -> Installed -> "%(name)s"->Update to apply the update.'
```

sonicaj: I'll omit the last sentence. Thank you

PR #3617, opened 2019-09-27
Merge master on top of freenas/12-devel

william-gr:


william-gr:
@freenas/middleware  I dont need a detailed review, but just a quick look if the pieces you changed are OK due the conflicts I had to solve.

PR #3620, opened 2019-09-30
fix(cloudsync): Allow bigger `service_account_credentials`

william-gr:
Ticket:	NAS-102938


themylogin:
Strange, I remember encountering and fixing this kind of error but I don't see the fix anywhere... We probably need to do this for all fields named `token` because usually there is a giant JSON string in there which might not be able to fit in 1024 bytes one day or another.

william-gr:
@themylogin Ok, do you want to take over this PR and do this, please? I only did it because I thought it would be a one-liner :P

PR #3633, opened 2019-10-04
NAS-103128 / 11.3 / Remove Netdata Service

sonicaj:
This commit removes netdata support in it's entirety provided by middlewared.


ncmans:
Here's what I found to be the reason why this is removed. Just for others to quickly find it:

https://www.ixsystems.com/community/threads/is-there-anything-like-a-netdata-equivalent-for-freenas.79945/#post-553692

PR #3636, opened 2019-10-07
NAS-102504 / 12 / API for CRUD operations on user props for zfs datsets

sonicaj:
This PR introduces a new API for CRUD operations on user properties for zfs datasets

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2619,6 +2619,26 @@ class PoolService(CRUDService):
             self.dismissed_import_disk_jobs.add(current_import_job["id"])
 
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2619,6 +2619,111 @@ def import_on_boot(self, job):
             self.dismissed_import_disk_jobs.add(current_import_job["id"])
 
 
+class PoolDatasetUserPropService(CRUDService):
+
+    class Config:
+        namespace = 'pool.dataset.userprop'
+
+    @filterable
+    async def query(self, filters=None, options=None):
```
william-gr: Do you think this should be a method, given that filter_list may block, not being a coroutine?

PR #3646, opened 2019-10-10
NAS-103046 / 11.2 / Fix kerberos keytab related string handling

anodos325:
Use 'id' value rather than UI keytab name to generate temporary keytab.
Change freenasldap to use new-style temporary keytab.
Add dedicated parser for ldap and ad config files.

### DIFF
--- a/gui/common/system.py
+++ b/gui/common/system.py
@@ -88,8 +88,32 @@ def get_sw_year():
     return str(datetime.now().year)
 
 
+def get_freenas_var_ds_config(f, var):
+    """
+    False gets picked up in get_freenas_var()
+    and results in default value (if it exists)
+    being set.
+    """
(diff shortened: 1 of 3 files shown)


gui/common/system.py
```diff
@@ -88,8 +88,32 @@ def get_sw_year():
     return str(datetime.now().year)
 
 
+def get_freenas_var_ds_config(f, var):
+    """
+    False gets picked up in get_freenas_var()
+    and results in default value (if it exists)
+    being set.
+    """
+    ds_config = []
```
william-gr: Is the file guaranteed to exist?

anodos325: It should be generated by this time, but there's not a strong guarantee. I'll add a check.

PR #3656, opened 2019-10-15
NAS-103441 / 11.3 / Exclude .system dataset from replication.list_datasets

themylogin:
@william-gr do you think we should do the same for iocage dataset?

### DIFF
--- a/src/middlewared/middlewared/plugins/zettarepl.py
+++ b/src/middlewared/middlewared/plugins/zettarepl.py
@@ -41,7 +41,8 @@ from middlewared.utils.string import make_sentence
 from middlewared.worker import watch_parent
 
 INVALID_DATASETS = (
-    re.compile(r"freenas-boot/?"),
+    re.compile(r"freenas-boot($|/)"),
+    re.compile(r"[^/]+/\.system($|/)")
 )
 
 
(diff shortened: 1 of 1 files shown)


themylogin:
@william-gr @sonicaj problem is that if we replicate entire pool from one NAS to entire pool of another and both are using iocage, replication will fail.

william-gr:
How? Doesnt the datasets get replicated to a sub dataset?

sonicaj:
@themylogin to clarify, are you referring to a case where both pools have iocage activated i.e both pools already contain iocage related datasets ?

themylogin:
> How? Doesnt the datasets get replicated to a sub dataset?

No, you can recursively replicate 192.168.0.1:tank/ -> 192.168.0.2:tank/, why not?

> @themylogin to clarify, are you referring to a case where both pools have iocage activated i.e both pools already contain iocage related datasets ?

Yes. In most common situation we have only one pool, so that's where iocage would be.

william-gr:
That use case doesnt concern me

sonicaj:
@themylogin i am not sure how replication handles clashes this way. If it just adds the children which are not present, functionality wise we are good. Middleware/iocage should work just fine. 

But from user point of view, maybe it would be nice if it was configurable ?

themylogin:
@sonicaj replication won't be able to replicate iocage dataset to already existing dataset and will fail.

@william-gr got it, merging as is.

PR #3657, opened 2019-10-15
NAS-103492 / 11.2 / remove restrictive nfs share logic

yocalebo:
Explanation in https://jira.ixsystems.com/browse/NAS-103492

### DIFF
--- a/src/middlewared/middlewared/plugins/nfs.py
+++ b/src/middlewared/middlewared/plugins/nfs.py
@@ -344,13 +344,6 @@ class SharingNFSService(CRUDService):
 
             used_networks.add(network)
 
-        if not data["hosts"] and not data["networks"]:
-            if used_networks:
-                verrors.add(
-                    f"{schema_name}.networks",
-                    "Another NFS share already exports this dataset for some network"
-                )
-
     @private
     async def extend(self, data):
         data["paths"] = [path["path"]


themylogin:
You've only removed part of the check (that works if new share does not have allowed hosts or networks specified). We should remove entire validate_hosts_and_networks then, but why did it exist in the first place then?

yocalebo:
The `validate_hosts_and_networks` still prevents the user from making an incorrect change.

If user creates 2 shares:
/mnt/tank/dataset -maproot="root" 10.10.10.10
/mnt/tank/dataset/subdir -maproot="root" 10.10.10.10

The `validate_hosts_and_networks` function will error out and say this ip address has already been used which is correct. If you try to write the above to freeBSD exports file, mountd will error out so the function still serves a purpose.

The logic that I removed prevents you from doing this configuration (which is allowed by mountd)
/mnt/tank/dataset 10.10.10.10
/mnt/tank/dataset/subdir

themylogin:
@yocalebo I think I understand you. But now this code will allow
```
/mnt/tank/dataset
/mnt/tank/dataset/subdir
```
Which mountd will also won't like.

I think you just need to replace
```
if used_networks:
```
with
```
if ipaddress.ip_network("0.0.0.0/0") in used_networks:
```
in original code.

yocalebo:
@themylogin Yes you're correct! I was actually typing a response to this to state the same thing. I'll make the changes and investigate to ensure I'm not missing any other edge cases.

william-gr:
@yocalebo ping

yocalebo:
@william-gr @themylogin I'm closing this PR.

This is a very complex problem that will need to be discussed at release council. If you're using NFSv4 then the ENTIRE webUI section that shows  hosts and networks is completely pointless because of how we configure the exports file.

Currently, support isn't having a huge problem in this area it was only 1 customer. I'm not going to shoe-horn a change in here that could potentially cause more problems than what it's worth. I'll revisit this at a later time.

PR #3658, opened 2019-10-15
NAS-103477 / 12 / Add Native ZFS Encryption support

sonicaj:
This PR adds support for native ZFS encryption.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -3238,14 +3238,10 @@ class PoolDatasetService(CRUDService):
             if len(f) == 3:
                 if f[0] in ('id', 'name', 'pool', 'type'):
                     zfsfilters.append(f)
-        datasets = self.middleware.call_sync(
-            'zfs.dataset.query', zfsfilters, {'extra': {
-                'top_level_properties': [
-                    'mountpoint', 'encryption', 'encryptionroot', 'keystatus'
-                ]
-            }}
+
+        return filter_list(
+            self.__transform(self.middleware.call_sync('zfs.dataset.query', zfsfilters)), filters, options
         )
-        return filter_list(self.__transform(datasets), filters, options)
 
     def __transform(self, datasets):
         """
@@ -3253,12 +3249,6 @@ class PoolDatasetService(CRUDService):
         making it match whatever pool.dataset.{create,update} uses as input.
         """
         def transform(dataset):
-            if 'key_loaded' not in dataset:
-                # FIXME: This is a hack, please fix it in the right place
-                dataset['encrypted'] = dataset.pop('encryption') != 'off'
-                dataset['encryption_root'] = dataset.pop('encryptionroot') or None
-                dataset['key_loaded'] = dataset.pop('keystatus') == 'available'
-
             for orig_name, new_name, method in (
                 ('org.freenas:description', 'comments', None),
                 ('org.freenas:quota_warning', 'quota_warning', None),
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2910,8 +2910,8 @@ class PoolDatasetService(CRUDService):
                 key = job.pipes.input.r.read(64)
                 # We would like to ensure key matches specified key format
                 try:
-                    key = key.decode()
-                except UnicodeDecodeError:
+                    key = hex(int(key, 16))[2:]
+                except ValueError:
                     verrors.add(f'{schema}.key_file', 'Please specify a valid key')
                     return {}
 
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -1970,6 +1970,7 @@ class PoolService(CRUDService):
 
         await self.middleware.call('service.reload', 'disk')
         await self.middleware.call_hook('pool.post_import_pool', pool)
+        await self.middleware.call('pool.dataset.sync_db_keys', [['name', '^', pool['name']]])
 
         return True
 
@@ -2243,6 +2244,10 @@ class PoolService(CRUDService):
                         e,
                         exc_info=True,
                     )
+            else:
+                await self.middleware.call(
+                    'datastore.delete', PoolDatasetService.dataset_store, [['name', '^', pool['name']]]
+                )
         else:
             job.set_progress(80, 'Exporting pool')
             await self.middleware.call('zfs.pool.export', pool['name'])
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2977,39 +2977,21 @@ class PoolDatasetService(CRUDService):
         Dict(
             'lock_options',
             Bool('force_umount', default=False),
-            Bool('recursive', default=False),
         )
     )
-    def lock(self, id, options):
-        # TODO:Ask William if we should mount encrypted root children of `id`, they would be unmounted
-        #  as well because of how `umount` works, we would like to mount them again
-        ds = self.middleware.call_sync('pool.dataset._get_instance', id)
+    async def lock(self, id, options):
+        ds = await self._get_instance(id)
 
-        if not options['recursive']:
-            if ds['encrypted'] and ZFSKeyFormat(ds['key_format']['value']) != ZFSKeyFormat.PASSPHRASE:
-                raise CallError('Only datasets which are encrypted with passphrase can be locked')
+        if not ds['encrypted']:
+            raise CallError(f'{id} is not encrypted')
+        elif ds['encrypted'] and ZFSKeyFormat(ds['key_format']['value']) != ZFSKeyFormat.PASSPHRASE:
+            raise CallError('Only datasets which are encrypted with passphrase can be locked')
+        elif not ds['key_loaded']:
+            raise CallError(f'Dataset {id} is already locked')
 
-            if not ds['key_loaded']:
-                raise CallError(f'Dataset {id} is already locked')
-
-        datasets = [
-            ds[0] for ds in self.query_encrypted_datasets(
-                id, {'recursive': options['recursive'], 'key_loaded': True, 'full_output': True}
-            )
-            if ZFSKeyFormat(ds[1]['key_format']['value']) == ZFSKeyFormat.PASSPHRASE
-        ]
-
-        failed = []
-        for dataset, status in zip(
-            datasets, self.execute_zfs_plugin_bulk(
-                'unload_key', lambda d: {'umount': True, 'force_umount': options['force_umount']}, datasets,
-            ),
-        ):
-            if status['error']:
-                failed.append((dataset, status['error']))
-
-        if failed:
-            raise CallError('\n'.join(f'Failed to lock {d}' for d in failed))
+        await self.middleware.call(
+            'zfs.dataset.unload_key', {'umount': True, 'force_umount': options['force_umount'], 'recursive': True}
+        )
 
         return True
 
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -545,8 +545,8 @@ class PoolService(CRUDService):
                     'AES-128-CCM', 'AES-192-CCM', 'AES-256-CCM', 'AES-128-GCM', 'AES-192-GCM', 'AES-256-GCM'
                 ]
             ),
-            Str('passphrase', default=None, null=True, empty=False),
-            Str('key', default=None, null=True, validators=[Range(min=64, max=64)]),
+            Str('passphrase', default=None, null=True, empty=False, private=True),
+            Str('key', default=None, null=True, validators=[Range(min=64, max=64)], private=True),
             register=True
         ),
         Dict(
@@ -3041,8 +3041,8 @@ class PoolDatasetService(CRUDService):
                     Dict(
                         'dataset',
                         Str('name', required=True, empty=False),
-                        Str('key', validators=[Range(min=64, max=64)]),
-                        Str('passphrase', empty=False),
+                        Str('key', validators=[Range(min=64, max=64)], private=True),
+                        Str('passphrase', empty=False, private=True),
                     )
                 ], default=[],
             ),
@@ -3175,8 +3175,8 @@ class PoolDatasetService(CRUDService):
                     Dict(
                         'dataset',
                         Str('name', required=True, empty=False),
-                        Str('key', validators=[Range(min=64, max=64)]),
-                        Str('passphrase', empty=False),
+                        Str('key', validators=[Range(min=64, max=64)], private=True),
+                        Str('passphrase', empty=False, private=True),
                     )
                 ], default=[],
             ),
@@ -3255,8 +3255,8 @@ class PoolDatasetService(CRUDService):
             Bool('generate_key', default=False),
             Bool('key_file', default=False),
             Int('pbkdf2iters', default=350000, validators=[Range(min=100000)]),
-            Str('passphrase', empty=False, default=None, null=True),
-            Str('key', validators=[Range(min=64, max=64)], default=None, null=True),
+            Str('passphrase', empty=False, default=None, null=True, private=True),
+            Str('key', validators=[Range(min=64, max=64)], default=None, null=True, private=True),
         )
     )
     @job(lock=lambda args: f'dataset_change_key_{args[0]}', pipes=['input'], check_pipes=False)
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2952,9 +2952,9 @@ class PoolDatasetService(CRUDService):
     @private
     @job(lock=lambda args: f'sync_encrypted_pool_dataset_keys_{args}')
     def sync_db_keys(self, job, name=None):
-        db_datasets = self.encrypted_roots_query_db(
-            [['OR', [['name', '=', name], ['name', '^', f'{name}/']]]] if name else None, True
-        )
+        filters = [['OR', [['name', '=', name], ['name', '^', f'{name}/']]]] if name else []
+        db_datasets = self.encrypted_roots_query_db(filters, True)
+        encrypted_roots = {d['name']: d for d in self.query(filters) if d['name'] == d['encryption_root']}
         to_remove = []
         check_key_job = self.middleware.call_sync('zfs.dataset.bulk_process', 'check_key', [
             (name, {'key': db_datasets[name]}) for name in db_datasets
@@ -2965,8 +2965,13 @@ class PoolDatasetService(CRUDService):
             return
 
         for dataset, status in zip(db_datasets, check_key_job.result):
-            if not status['result'] or status['error']:
+            if not status['result']:
                 to_remove.append(dataset)
+            elif status['error']:
+                if dataset not in encrypted_roots:
+                    to_remove.append(dataset)
+                else:
+                    self.middleware.logger.error(f'Failed to check encryption status for {dataset}: {status["error"]}')
 
         self.middleware.call_sync('datastore.delete', self.dataset_store, [['name', 'in', to_remove]])
 
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -3083,7 +3083,7 @@ class PoolDatasetService(CRUDService):
             if all(ds.get(k) for k in ('key', 'passphrase')):
                 verrors.add(
                     f'unlock_options.datasets.{i}.dataset.key',
-                    f'Must not be specified when unlock_options.datasets.{i}.dataset.passphrase is supplied.'
+                    f'Must not be specified when passphrase for {ds["name"]} is supplied'
                 )
             keys_supplied[ds['name']] = ds.get('key') or ds.get('passphrase')
 
@@ -3225,7 +3225,7 @@ class PoolDatasetService(CRUDService):
             if all(ds.get(k) for k in ('key', 'passphrase')):
                 verrors.add(
                     f'unlock_options.datasets.{i}.dataset.key',
-                    f'Must not be specified when unlock_options.datasets.{i}.dataset.passphrase is supplied.'
+                    f'Must not be specified when passphrase for {ds["name"]} is supplied'
                 )
             keys_supplied[ds['name']] = ds.get('key') or ds.get('passphrase')
 
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2976,6 +2976,7 @@ class PoolDatasetService(CRUDService):
         check_key_job.wait_sync()
         if check_key_job.error:
             self.middleware.logger.error(f'Failed to sync database keys: {check_key_job.error}')
+            return
 
         for dataset, status in zip(db_datasets, check_key_job.result):
             if not status['result'] or status['error']:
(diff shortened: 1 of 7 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1943,6 +1977,7 @@ def configure_resilver_priority(self):
 
         await self.middleware.call('service.reload', 'disk')
         await self.middleware.call_hook('pool.post_import_pool', pool)
+        await self.middleware.call('pool.dataset.sync_db_keys', [['name', '^', pool['name']]])
```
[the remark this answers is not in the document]
sonicaj: I will certainly improve the filter, however, it won't break as the method only checks if the keys in db are valid and if not so, removes them from db

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2586,6 +2625,17 @@ def import_on_boot(self, job):
                         'Failed to inherit mountpoints for %s', pool['name'], exc_info=True,
                     )
 
+                unlock_job = self.middleware.call_sync(
+                    'pool.dataset.unlock', pool['name'], {'toggle_attachments': False, 'recursive': True}
+                )
+                unlock_job.wait_sync()
+                if unlock_job.error or unlock_job.result['failed']:
+                    fd = ','.join(unlock_job.result['failed']) if not unlock_job.error else ''
```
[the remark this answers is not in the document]
sonicaj: `failed_datasets` -> `fd`, though i agree, it sounds more like file descriptor but the former name was too big. I'll address both concerns : )

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -657,17 +679,25 @@ def pool_extend(self, pool):
...
-        job.set_progress(95, 'Setting pool options')
-        pool_id = None
+        pool_id = z_pool = None
         try:
+            await self.middleware.call(
+                'pool.dataset.insert_or_update', {
+                    'name': data['name'], 'encryption_key': encryption_dict.get('key'),
+                    'key_format': encryption_dict.get('keyformat')
+                }
+            )
```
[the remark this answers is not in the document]
sonicaj: So we do delete it in case an exception occurs below using the dataset name. I saved it in the database before even the pool is created for a hard edge case where for any reason the process is disrupted and we lose keys to the encrypted root dataset. But it doesn't make much sense still to do this first. I'll correct the order

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2749,13 +2799,473 @@ def query(self, filters=None, options=None):
...
+                'Should not be provided when key format is not "PASSPHRASE"'
+            )
+
+        if not verrors:
+            key = encryption_dict['passphrase']
+            if encryption_dict['generate_key']:
+                key = getattr(secrets, f'token_{"hex" if key_format == ZFSKeyFormat.HEX else "bytes"}')(32)
+            elif not key:
+                job.check_pipe('input')
+                key = job.pipes.input.r.read().strip()
```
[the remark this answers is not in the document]
sonicaj: Good catch. 32 bytes is fixed for raw/hex keys, so i should only read that much

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2749,13 +2799,473 @@ def query(self, filters=None, options=None):
...
+        ds = await self._get_instance(id)
+
+        if not ds['encrypted']:
+            raise CallError(f'{id} is not encrypted')
+        elif ds['locked']:
+            raise CallError(f'Dataset {id} is already locked')
+        elif ZFSKeyFormat(ds['key_format']['value']) != ZFSKeyFormat.PASSPHRASE:
+            raise CallError('Only datasets which are encrypted with passphrase can be locked')
+        elif id == (await self.middleware.call('systemdataset.config'))['pool']:
+            raise CallError(f'Please move system dataset to another pool before locking {id}')
```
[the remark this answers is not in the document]
sonicaj: We don't. Hard edge case where user manipulates this on his/her own

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -502,6 +502,142 @@ def query_for_quota_alert(self):
...
+            self.logger.error(f'Failed to check key for {id}', exc_info=True)
+            raise CallError(f'Failed to check key for {id}: {e}')
+
+    @accepts(
+        Str('id'),
+        Dict(
+            'unload_key_options',
+            Bool('recursive', default=False),
+            Bool('force_umount', default=False),
+            Bool('umount', default=False),
```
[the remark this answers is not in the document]
sonicaj: Nope, it is not possible. However i have tried to be consistent here with how ZFS operates exposing option to `umount` it but not doing it implicitly so that the developer is aware this needs to be done. In case this is not done, an exception will be raised

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2749,13 +2799,473 @@ def query(self, filters=None, options=None):
...
+                'datastore.update',
+                self.dataset_store,
+                ds[0]['id'], data
+            )
+        else:
+            await self.middleware.call(
+                'datastore.insert',
+                self.dataset_store,
+                data
+            )
```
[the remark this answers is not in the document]
sonicaj: So far we seem to have no use for the returned values, please advise if we should still do that.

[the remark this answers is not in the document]
william-gr: The idea was to use it in pool create, to delete using primary key and not name which is weirder

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2749,13 +2813,560 @@ def query(self, filters=None, options=None):
         return True
 
 
+class PoolDatasetEncryptionModel(sa.Model):
+    __tablename__ = 'storage_datasetencryption'
```
themylogin: We already have `storage.encrypteddisk`, why can't we name this in same manner? E.g. `storage_encrypteddataset`.

src/middlewared/middlewared/alembic/versions/12.0/2019-10-24-18-45_dataset_model.py
```diff
@@ -0,0 +1,30 @@
...
+# revision identifiers, used by Alembic.
+revision = 'ff2e85cf3c1b'
+down_revision = 'ed69a9a6fab1'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'storage_dataset',
```
themylogin: Your model name is different.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2749,13 +2813,560 @@ def query(self, filters=None, options=None):
...
+        Please refer to websocket documentation for downloading the file.
+        """
+        self.middleware.call_sync('pool.dataset._get_instance', id)
+        sync_job = self.middleware.call_sync('pool.dataset.sync_db_keys', [['name', '^', id]])
+        sync_job.wait_sync()
+
+        datasets = self.encrypted_roots_query_db([['name', '^', id]], decrypt=True)
+        tar_path = None
+        try:
+            tar_path = tempfile.mkstemp()[1]
```
themylogin: Why not tempfile.NamedTemporaryFile?

sonicaj: That will provide me with an open file handle which i'll have to close. That's not the case with `mkstemp`

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2749,13 +2813,564 @@ def query(self, filters=None, options=None):
...
+            j = self.middleware.call_sync('pool.dataset.toggle_attachments', self.__attachments_path(dataset), True)
+            j.wait_sync()
+
+        return {'unlocked': unlocked, 'failed': failed}
+
+    @private
+    @job(lock=lambda args: f'toggle_attachments_{args[0]}')
+    async def toggle_attachments(self, path, enabled=True):
+        for delegate in self.attachment_delegates:
+            delegate.toggle((await delegate.query(path, enabled)), enabled)
```
themylogin: I might been found guilty in not documenting `FSAttachmentDelegate` so I did it now.

Your code disables attachment that are already disabled or enables ones that are already enabled. That has no sense for me. What did you want to do?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2749,13 +2813,564 @@ def query(self, filters=None, options=None):
...
+            'pool.dataset.sync_db_keys', ['OR', [['name', '=', id], ['name', '^', f'{id}/']]]
+        )
+
+    @private
+    def _retrieve_keys_from_compressed_file(self, job, ds_name):
+        job.check_pipe('input')
+        temp_dir = key_file = None
+        data = {}
+        try:
+            with tempfile.NamedTemporaryFile(delete=False, mode='wb') as f:
```
themylogin: Why `delete=False` and everything else outside of context manager block?

sonicaj: The idea is that if for any reason an exception occurs while we are copying from the job pipe to the temporary fie, we handle that separately and reading from the newly written temp file would also mean that i have to close the previous handle, so we accomplish that as well. An exception in the latter section means that we have been provided with a single key file instead of a tar file.

themylogin: Your `key_file` value is overwritten in line 3360 so `f` is never deleted in line 3368, is this intentional and what for?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -2035,8 +2035,13 @@ class VMFSAttachmentDelegate(FSAttachmentDelegate):
...
     async def query(self, path, enabled):
         vms_attached = []
+        vms = {
+            vm['id']: vm for vm in await self.middleware.call(
+                'vm.query', [('status.state', '=' if enabled else '!=', 'RUNNING')]
+            )
+        }
         for device in await self.middleware.call('datastore.query', 'vm.device'):
-            if device['dtype'] not in ('DISK', 'RAW'):
+            if (device['dtype'] not in ('DISK', 'RAW')) or device['vm']['id'] in vms:
```
themylogin: `vms` contains list of VMs that are subject of our interest, right? (started if `enabled is True`, stopped otherwise) Why do you skip devices that are part of these VMs then?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -2037,7 +2037,7 @@ class VMFSAttachmentDelegate(FSAttachmentDelegate):
         vms_attached = []
         vms = {
             vm['id']: vm for vm in await self.middleware.call(
-                'vm.query', [('status.state', '=' if enabled else '!=', 'RUNNING')]
+                'vm.query', [('status.state', '!=' if enabled else '=', 'RUNNING')]
```
themylogin: Classic case of "don’t not avoid double negative boolean logic" :)

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2749,13 +2813,567 @@ def query(self, filters=None, options=None):
...
+        """
+        ds = await self._get_instance(id)
+        if not ds['encrypted']:
+            raise CallError(f'Dataset {id} is not encrypted')
+        elif not ds['encryption_root'] != id:
+            raise CallError(f'Dataset {id} is not an encryption root')
+        elif ds['locked']:
+            raise CallError('Dataset must be unlocked to perform this operation')
+        elif '/' not in id:
+            raise CallError('This operation is not valid for root datasets')
```
wblock: Does "not valid" mean "not allowed" or "cannot be done on root datasets"?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2749,13 +2813,567 @@ def query(self, filters=None, options=None):
...
+                        d['name'] == d['encryption_root']
+                        for d in await self.middleware.run_in_thread(
+                            self.query, [
+                                ['id', '^', f'{id}/'], ['encrypted', '=', True],
+                                ['key_format.value', '!=', ZFSKeyFormat.PASSPHRASE.value]
+                            ]
+                        )
+                    ):
+                        raise CallError(
+                            f'{id} has children which are encrypted with HEX/RAW keys. It is not allowed to '
```
wblock: As above.

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -113,13 +113,28 @@ class Config:
...
+            elif await self.middleware.call(
+                'pool.dataset.query', [
+                    ['name', '=', new['pool']], ['encrypted', '=', True],
+                    ['OR', [['key_format.value', '=', 'PASSPHRASE'], ['locked', '=', True]]]
+                ]
+            ):
+                verrors.add(
+                    'sysdataset_update.pool',
+                    'The system dataset cannot be placed on a pool '
+                    'which has root dataset encrypted with PASSPHRASE or is locked.'
```
wblock: Capitalization is again confusing to the user.  It seems to mean the PASSPHRASE method, not just a passphrase.
```suggestion
                    'which has the root dataset encrypted with PASSPHRASE or locked.'
```

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2749,13 +2813,567 @@ def query(self, filters=None, options=None):
...
+                key = getattr(secrets, f'token_{"hex" if key_format == ZFSKeyFormat.HEX else "bytes"}')(32)[:32]
+            elif not key:
+                job.check_pipe('input')
+                key = job.pipes.input.r.read(32)
+                # We would like to ensure key matches specified key format
+                if key_format == ZFSKeyFormat.HEX:
+                    try:
+                        key.decode()
+                    except UnicodeDecodeError:
+                        verrors.add(f'{schema}.key_file', 'Unable to decode HEX key')
```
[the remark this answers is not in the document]
sonicaj: I am not sure i follow, are you referring to `HEX` ? If yes, can you please elaborate how is it invisible to the user ? ( on the other hand i think i should have quotes around it keeping consistent with other usages of passphrase )

src/middlewared/middlewared/plugins/pool.py
sonicaj: Yes

src/middlewared/middlewared/plugins/sysdataset.py
sonicaj: I am not sure i follow "PASSPHRASE method" ? There are 3 types of keys which can be used to encrypt a dataset and passphrase is one of them. So if a dataset is encrypted with passphrase key type, we don't want to allow system dataset to be placed there of if course the root dataset is locked as well. I don't think `or locked` is okay in this context, please advise. Thank you

wblock: As above, this mixes Python capitalization with spoken (read) words. When speaking to the end user about a key, use "a raw key, a hex key, or a passphrase key".  If you are speaking to the user about Python constants/methods/other code, then use the uppercase names.

src/middlewared/middlewared/plugins/pool.py
darkfiberiru: I would argue that "not valid" sounds like it's an illegal operation so were stopping you while not allowed sounds like some restriction were doing by choice as part of freenas.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -525,6 +536,20 @@ def pool_extend(self, pool):
...
+            'encryption_options',
+            Bool('generate_key', default=False),
+            Bool('key_file', default=False),
+            Int('pbkdf2iters', default=350000, validators=[Range(min=100000)]),
+            Str(
+                'algorithm', default='AES-256-CCM', enum=[
+                    'AES-128-CCM', 'AES-192-CCM', 'AES-256-CCM', 'AES-128-GCM', 'AES-192-GCM', 'AES-256-GCM'
+                ]
+            ),
+            Str('key_format', default='HEX', enum=list(ZFSKeyFormat.__members__)),
```
william-gr: It makes me wonder if we arent just complicating yourselves by giving the user a choice between RAW and HEX? isnt HEX just the RAW but 16bit encoded?

william-gr: To add, uploading a key file sounds a very edge case that we shouldnt probably even expose in the UI. What makes sense is auto generating the key, as we currently do.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2747,13 +2811,580 @@ def query(self, filters=None, options=None):
...
+                    db_key = datasets[ds]
+                    key = io.BytesIO(db_key if isinstance(db_key, bytes) else db_key.encode())
+                    info = tarfile.TarInfo(name=ds)
+                    info.size = len(db_key)
+                    tar.addfile(tarinfo=info, fileobj=key)
+
+            with open(tar_path, 'rb') as f:
+                shutil.copyfileobj(f, job.pipes.output.w)
+        finally:
+            shutil.rmtree(temp_dir or '', ignore_errors=True)
```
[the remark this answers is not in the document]
sonicaj: In case something wrong happens and a temporary directory is not created for any reason ? However an `if` would seem more reasonable here perhaps first. I’ll do that

src/middlewared/middlewared/plugins/pool.py
sonicaj: It’s not possible to do this operation on root datasets. Apologies for the vague comment above

wblock: But that just says the same thing again.  Being specific avoids the problems with "invalid" or "not allowed":
"Root datasets do not have a parent and cannot inherit encryption settings."

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -546,8 +546,9 @@ def pool_extend(self, pool):
                     'AES-128-CCM', 'AES-192-CCM', 'AES-256-CCM', 'AES-128-GCM', 'AES-192-GCM', 'AES-256-GCM'
                 ]
             ),
-            Str('key_format', default='HEX', enum=list(ZFSKeyFormat.__members__)),
+            Str('key_format', default='HEX', enum=[ZFSKeyFormat.PASSPHRASE.value, ZFSKeyFormat.HEX.value]),
```
william-gr: default should be ZFSKeyFormat.HEX.value ?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2883,41 +2884,42 @@ def validate_encryption_data(self, job, verrors, encryption_dict, schema):
...
+                key = job.pipes.input.r.read(64)
                 # We would like to ensure key matches specified key format
-                if key_format == ZFSKeyFormat.HEX:
-                    try:
-                        key.decode()
-                    except UnicodeDecodeError:
-                        verrors.add(f'{schema}.key_file', 'Unable to decode "HEX" key')
-                        return {}
+                try:
+                    key = key.decode()
```
william-gr: only .decode() seems too weak, it will decode utf8 chars.

Should we probably do .decode('ascii') and/or perhaps make sure its hexadecimal?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3075,7 +3082,14 @@ def unlock(self, job, id, options):
 
         if options['key_file']:
             keys_supplied = self._retrieve_keys_from_compressed_file(job, id)
-        keys_supplied.update({d['name']: d['passphrase'] for d in options['datasets']})
+
+        for i, ds in enumerate(options['datasets']):
+            if all(ds.get(k) for k in ('key', 'passphrase')):
+                verrors.add(
+                    f'unlock_options.datasets.{i}.dataset.key',
+                    f'Must not be specified when unlock_options.datasets.{i}.dataset.passphrase is supplied.'
```
william-gr: We could name it as "Must not be specified when passphrase for {dataset} is supplied"

unlock_options.datasets.{i}.dataset.passphrase as part of the text that would eventually show up in the UI is weird.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3194,17 +3221,29 @@ def encryption_summary(self, job, id, options):
...
+        verrors = ValidationErrors()
         if options['key_file']:
             keys_supplied = self._retrieve_keys_from_compressed_file(job, id)
-        keys_supplied.update({d['name']: d['passphrase'] for d in options['datasets']})
 
+        for i, ds in enumerate(options['datasets']):
+            if all(ds.get(k) for k in ('key', 'passphrase')):
+                verrors.add(
+                    f'unlock_options.datasets.{i}.dataset.key',
+                    f'Must not be specified when unlock_options.datasets.{i}.dataset.passphrase is supplied.'
```
william-gr: Same here about unlock_options.datasets.{i}.dataset.passphrase

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3049,7 +3051,12 @@ def export_keys(self, job, id):
             Bool('recursive', default=False),
             List(
                 'datasets', items=[
-                    Dict('dataset', Str('name', required=True), Str('passphrase', required=True, empty=False)),
+                    Dict(
+                        'dataset',
+                        Str('name', required=True, empty=False),
+                        Str('key', validators=[Range(min=64, max=64)]),
```
william-gr: Hm, why would one need to provide a key to unlock a dataset?

sonicaj: Edge case, where user has imported a dataset perhaps which was exported from a different machine. The same can be used to replace the entire usage of `key_file` if desired

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -525,6 +535,21 @@ def pool_extend(self, pool):
...
+            'encryption_options',
+            Bool('generate_key', default=False),
+            Bool('key_file', default=False),
+            Int('pbkdf2iters', default=350000, validators=[Range(min=100000)]),
+            Str(
+                'algorithm', default='AES-256-CCM', enum=[
+                    'AES-128-CCM', 'AES-192-CCM', 'AES-256-CCM', 'AES-128-GCM', 'AES-192-GCM', 'AES-256-GCM'
+                ]
+            ),
+            Str('key_format', default='KEY', enum=[ZFSKeyFormat.PASSPHRASE.value, 'KEY']),
```
william-gr: key_format = KEY? is that intended?

william-gr: Also, we already have `key` and `passphrase` attributes, do we really need `key_format` ?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2749,13 +2799,473 @@ def query(self, filters=None, options=None):
...
+                'key': key, 'load_key': False,
+            }
+        )
+
+        # TODO: Handle renames of datasets appropriately wrt encryption roots and db - this will be done when
+        #  devd changes are in from the OS end
+        await self.insert_or_update({'encryption_key': key, 'key_format': options['key_format'], 'name': id})
+
+    @accepts(Str('id'))
+    async def inherit_parent_encryption_properties(self, id):
```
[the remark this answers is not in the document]
william-gr: This was resolved but this method is still public and has no docstring?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3141,8 +3141,8 @@ def unlock(self, job, id, options):
             else:
                 unlocked.append(name)
 
-        if options['restart_related_attachments'] and not failed:
-            j = self.middleware.call_sync('pool.dataset.restart_related_attachments', self.__attachments_path(dataset))
+        if options['start_related_attachments'] and not failed:
```
william-gr: Not sure it any clearer with "start" ?

We will likely be restarting services because we may have shares that are enable for other pools/datasets?

PR #3659, opened 2019-10-15
NAS-103496 / 11.2 / properly handle rc.conf.local on TrueNAS HA systems

yocalebo:
Jira: https://jira.ixsystems.com/browse/NAS-103496

### DIFF
--- a/src/freenas/etc/rc.conf.local
+++ b/src/freenas/etc/rc.conf.local
@@ -200,7 +200,7 @@ _interface_config()
 			else
 				int_passopt=""
 			fi
-		if [ ${failover_licensed} -eq 1 ]; then
+		if [ ${failover_status} = "MASTER" ]; then
 				carp1_skew="1"
 			fi
 			echo -n "ifconfig_${interface}_alias0=\"inet vhid ${int_vhid} advskew ${carp1_skew}${int_passopt} alias ${int_vip}/32"
@@ -624,7 +624,7 @@ _gen_conf()
 	# 1. FreeNAS systems
 	# 2. TrueNAS single controller systems
 	# 3. ONLY TrueNAS MASTER controllers in HA systems
-	if [ $(failover_licensed) -eq 0 -o "${failover_status}" = "MASTER" ]; then
+	if [ ${failover_licensed} -eq 0 -o "${failover_status}" = "MASTER" ]; then
 		_bool_svcconfig netatalk_enable afp
 		_bool_svcconfig proftpd_enable ftp
 		_bool_svcconfig nfs_server_enable nfs
--- a/src/freenas/etc/rc.conf.local
+++ b/src/freenas/etc/rc.conf.local
@@ -580,8 +580,8 @@ _watchdogd_config()
 _asigra_config()
 {
 	local failover_licensed=$1
-	local failover_status=$2
-        if [ ${failover_licensed} -eq 0 -o "$(failover_status)" = "MASTER" ]; then
+	local failover_status="$2"
+	if [ ${failover_licensed} -eq 0 -o "${failover_status}" = "MASTER" ]; then
 		_bool_svcconfig postgresql_enable asigra
 		_bool_svcconfig dssystem_enable asigra
 	fi
@@ -663,7 +663,7 @@ _gen_conf()
 		_nis_config
 	fi
 
-	_asigra_config ${failover_licensed} $(failover_status)
+	_asigra_config ${failover_licensed} ${failover_status}
 	_snmp_config
 
 	_ldap_config
@@ -680,7 +680,7 @@ _gen_conf()
 		# NFS service is enabled
 		if [ "$(${FREENAS_SQLITE_CMD} ${FREENAS_CONFIG} "SELECT nfs_srv_v4 FROM services_nfs")" = 1 ]; then
 			# NFSv4 is enabled
-			if [ ${failover_licensed} -eq 0 -o "$(failover_status)" = "MASTER" ]; then
+			if [ ${failover_licensed} -eq 0 -o "${failover_status}" = "MASTER" ]; then
 				echo "nfsv4_server_enable=\"YES\""
 			fi
 			if [ "$(${FREENAS_SQLITE_CMD} ${FREENAS_CONFIG} "SELECT nfs_srv_v4_v3owner FROM services_nfs")" = 1 ]; then
@@ -690,7 +690,7 @@ _gen_conf()
 				sysctl vfs.nfsd.enable_stringtouid=1 > /dev/null
 				sysctl vfs.nfs.enable_uidtostring=1 > /dev/null
 			else
-				if [ ${failover_licensed} -eq 0 -o "$(failover_status)" = "MASTER" ]; then
+				if [ ${failover_licensed} -eq 0 -o "${failover_status}" = "MASTER" ]; then
 					echo "nfsuserd_enable=\"YES\""
 				fi
 				sysctl vfs.nfsd.enable_stringtouid=0 > /dev/null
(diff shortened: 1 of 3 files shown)


yocalebo:
@themylogin actually, the root cause is because this commit: 41646cc5676a645ae4364b682c6dcaa2a3e3348f caused the regression. That commit prevents the "mountd_enable="YES"" from being written to /etc/rc.conf.local which prevents mountd from reloading the config correctly and since we're using "quietreload" it hides the warnings. I'll push the proper commit.

src/freenas/etc/rc.conf.local
```diff
@@ -622,7 +622,7 @@ _gen_conf()
 	## Daemons
 	_bool_svcconfig netatalk_enable afp
 	_bool_svcconfig proftpd_enable ftp
-	if [ ${failover_licensed} -ne 1 ]; then
+	if [ "${failover_status}" != "BACKUP" ]; then
```
william-gr: failover_status cannot be relied on, its not guaranteed to be valid by the time the cache is generated.

themylogin:
@yocalebo maybe you should just change it to onereload? (and for all other reloads in that file too)

yocalebo:
@themylogin yeah I agree. If we can't rely on that command, then I'll just do onereload for all services. @william-gr do you disagree with that method? Trying to make sure I don't introduce significant changes in this release.

william-gr:
Need to be careful it wont start/stop services if NFS is not enabled.

themylogin:
@william-gr isn't that what's `one` for?

william-gr:
No, `one` ignores rc var as forces start/stop.

src/freenas/etc/rc.conf.local
yocalebo: @william-gr If we can't trust `failover_status` then there are a couple other places in that script that use that same check. Do we need to worry about those as well?

william-gr: Possibly

yocalebo: So I'm going to be honest here. That config file is potentially breaking many services. For example, if any services rc.conf settings are dictated by the `if [${failover_licensed} -ne 1]` check then we're not setting the rc.conf.local file correctly for TrueNAS HA systems. We're starting the services without issues because carp-state change hook doesn't have any notion of that file, however, if we use the `service` command to manage any services we're not setting rc.conf settings appropriately. I'm not sure where to really go with this one without making logic changes in other portions of the code which would probably be invasive. Any hints on how you'd like this done @william-gr or do we revert the changes in commit: 41646cc5676a645ae4364b682c6dcaa2a3e3348f ?

yocalebo: Further investigation shows that this is also the reason why the `vfs.nfsd.server_max_nfsvers` sysctl isn't being set back to `3` when we restart nfsd daemon via the `service` framework because the rc script is automatically changing that value back because we're not setting `nfsv4_server_enable="YES"` in /tmp/rc.conf.local on TrueNAS HA. The ticket for that issue is here: https://jira.ixsystems.com/browse/NAS-103502 So it seems like either we make invasive changes somewhere else or we revert commit:41646cc5676a645ae4364b682c6dcaa2a3e3348f

yocalebo: I'm inclined to revert the commit at this point because that fix was only to fix some cosmetic console error messages when TrueNAS passive controller booted up.

freqlabs:
```suggestion
	if [ "${failover_status}" = "MASTER" -o "${failover_status}" = "SINGLE" ]; then
```
If it's written this way, are there any other conditions that should be covered? I assume the cache is regenerated and services are started or stopped accordingly when the failover status changes.

freqlabs: Scratch that, $failover_status isn't SINGLE for FreeNAS, so it should be more like
```suggestion
	if [ "${failover_status}" = "MASTER" \
	    -o "${failover_status}" = "SINGLE" \
	    -o ${failover_licensed} -ne 1 ]; then
```

william-gr:
@yocalebo & all. I think reverting the other commit is going to be a step back.

The problem with rc.conf.local is that its cached based on the database hash. Back in the day we improperly used failover_status in rc.conf.local and the CARP was initialized during boot. Now CARP is only initialized after boot is completed and rc.conf.local is cached rendering failover_status useless.

What I think we should do here is analyze every failover_licensed and failover_status usages in rc.conf.local and enabled the blocks when failover is not licensed or failover status is master. Additionally to that we should change the carp script to remove the rc.conf.local cache after pool is imported so the rc vars are re-evaluated.

yocalebo:
Okay sounds good. Do you want me to make those changes to this branch for this PR? Or would you like me to close this one and open a fresh PR?

william-gr:
Whatever is easier

yocalebo:
I will make an associated commit in TrueNAS repo for carp-state-change script.

src/freenas/etc/rc.conf.local
```diff
@@ -201,7 +201,7 @@ _interface_config()
 			else
 				int_passopt=""
 			fi
-			if [ "${failover_status}" = "MASTER" ]; then
+			if [ $(failover_licensed) -eq 0 -o "${failover_status}" = "MASTER" ]; then
```
william-gr: Not necessary in this case, we dont want carp if not failover licensed

src/freenas/etc/rc.conf.local
```diff
@@ -580,7 +580,8 @@ _watchdogd_config()
 _asigra_config()
 {
 	local failover_licensed=$1
-	if [ ${failover_licensed} -eq 0 ]; then
+	local failover_status=$2
+        if [ ${failover_licensed} -eq 0 -o "$(failover_status)" = "MASTER" ]; then
```
william-gr: tab vs spaces

src/freenas/etc/rc.conf.local
```diff
@@ -659,7 +663,7 @@ _gen_conf()
 		_nis_config
 	fi
 
-	_asigra_config ${failover_licensed}
+	_asigra_config ${failover_licensed} $(failover_status)
```
william-gr: ${failover_status}

src/freenas/etc/rc.conf.local
```diff
@@ -676,7 +680,7 @@ _gen_conf()
 		# NFS service is enabled
 		if [ "$(${FREENAS_SQLITE_CMD} ${FREENAS_CONFIG} "SELECT nfs_srv_v4 FROM services_nfs")" = 1 ]; then
 			# NFSv4 is enabled
-			if [ ${failover_licensed} -ne 1 ]; then
+			if [ ${failover_licensed} -eq 0 -o "$(failover_status)" = "MASTER" ]; then
```
william-gr: ${failover_status} and all the others below

src/freenas/etc/rc.conf.local
```diff
@@ -201,7 +200,7 @@ _interface_config()
 			else
 				int_passopt=""
 			fi
-			if [ "${failover_status}" = "MASTER" ]; then
+		if [ ${failover_licensed} -eq 1 ]; then
```
william-gr: I meant to keep using failover_status, we dont want carp1 in both nodes to have skew 1

src/freenas/etc/rc.conf.local
```diff
@@ -614,41 +614,44 @@ _gen_conf()
...
-	## Daemons
-	_bool_svcconfig netatalk_enable afp
-	_bool_svcconfig proftpd_enable ftp
-	if [ ${failover_licensed} -ne 1 ]; then
+	# JIRA NAS-103496
+	# These daemons should be configured to start in the following scenarios:
+	# 1. FreeNAS systems
+	# 2. TrueNAS single controller systems
+	# 3. ONLY TrueNAS MASTER controllers in HA systems
+	if [ $(failover_licensed) -eq 0 -o "${failover_status}" = "MASTER" ]; then
```
william-gr: ${failover_licensed}

src/freenas/etc/rc.conf.local
```diff
@@ -201,7 +200,7 @@ _interface_config()
 			else
 				int_passopt=""
 			fi
-			if [ "${failover_status}" = "MASTER" ]; then
+		if [ "${failover_status}" = "MASTER" ]; then
```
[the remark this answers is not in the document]
yocalebo: I'm battling WSL on Windows 10 and am clearly losing. Should be fixed.

PR #3664, opened 2019-10-15
NAS-103445 / 11.3 / replication.target_unmatched_snapshots

themylogin:


src/middlewared/middlewared/plugins/replication.py
```diff
@@ -686,6 +686,43 @@ def _legacy_extend_context(self):
...
+                "params": [
+                    "PUSH",
+                    ["repl/work", "repl/games"],
+                    "backup",
+                    "SSH",
+                    4
+                ]
+            }
+
+        Returns
```
william-gr: Does it render well? wont this need double spaces and the block below 4 spaces and ::javascript?

themylogin: @william-gr renders as any other method doc (poorly comparing to swagger, but guess we'll work on it someday)

PR #3665, opened 2019-10-16
NAS-103497 / 11.3 / Create generic directoryservice.[get|set]_state functions

anodos325:
Return status of all directory services in one API call. Register event for status changes.

### DIFF
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -846,6 +846,10 @@ class ActiveDirectoryService(ConfigService):
 
     @accepts()
     async def get_state(self):
+        """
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -849,31 +842,11 @@ class Config:
 
     @private
     async def _set_state(self, state):
-        await self.middleware.call('cache.put', 'AD_State', state.name)
+        return await self.middleware.call('directoryservices.set_state', {'activedirectory': state.name})
 
     @accepts()
     async def get_state(self):
```
william-gr: I think we need either to make this private or add a doc stating its a shortcut to directoryservices.get_state for active directory

Ditto for nis/ldap

PR #3668, opened 2019-10-16
NAS-103297 / 11.2 / Refuse to bind netatalk to invalid IP addresses

anodos325:
Only add bind parameter if the specified IP address is valid.


william-gr:
@anodos325 does this need a master equivalent?

PR #3673, opened 2019-10-16
NAS-102893 / 11.2 / Avoid unnecessary Samba starts / stops during sysdataset manipulation

anodos325:
- Only restart system_datasets if the system dataset isn't located on a zpool
- During system_datasets restart, do not pass onerestart to Samba.
  This is to prevent samba from restarting when it's not enabled.


william-gr:
Minor flake8 nit :package: 
src/middlewared/middlewared/plugins/service.py:1072:35: E241 multiple spaces after ','

william-gr:
@anodos325  actually, is it even worth fixing this for 11.2?

PR #3690, opened 2019-10-18
NAS-103496 / 11.3 / properly configure rc.conf.local on TrueNAS HA

yocalebo:


yocalebo:
@william-gr we're already calling etc.generate rc after zpool import in truenas master branch so I believe there doesn't need to be any changes on that side.

src/middlewared/middlewared/etc_files/rc.conf.py
```diff
@@ -102,28 +103,41 @@ def lldp_config(middleware, context):
...
+        }
+
+    """
+    JIRA NAS-103496
+    These daemons should be configured to start in the following scenarios:
+        1. FreeNAS systems
+        2. TrueNAS single controller systems
+        3. On BOTH controllers in TrueNAS HA systems
+    """
+    mapping.update({
```
william-gr: mapping will be undefined licensed or not master

PR #3694, opened 2019-10-21
NAS-103539 / 12 / Fix arc_summary.py l2arc ValueError exception

freqlabs:
Use the Decimal values of the kstat to find the total number of L2ARC
evictions rather than trying to parse the formatted numbers as ints.

Ticket: NAS-103539


freqlabs:
@william-gr I want this backported to 11.2, so should I skip 12 and reopen this against master?

william-gr:
@freqlabs yes

PR #3695, opened 2019-10-21
NAS-103539 / 11.3 / Fix arc_summary.py l2arc ValueError exception

freqlabs:
Use the Decimal values of the kstat to find the total number of L2ARC
evictions rather than trying to parse the formatted numbers as ints.

Ticket: NAS-103539


william-gr:
backport

PR #3706, opened 2019-10-23
Plugins parts

themylogin:


themylogin:
@william-gr interesting part is https://github.com/freenas/freenas/pull/3706/commits/c766a4966f26c932b8bb1447fbbd6dd0824f60b2 the second commit is just splitting big file into small ones

src/middlewared/middlewared/platform.py
```diff
@@ -0,0 +1 @@
+PLATFORM = "freebsd"
```
william-gr: How about using `platform` module instead?

src/middlewared/middlewared/plugins/datastore/read.py
```diff
@@ -0,0 +1,267 @@
...
+from .filter import FilterMixin
+from .schema import SchemaMixin
+
+
+def regexp(expr, item):
+    reg = re.compile(expr, re.I)
+    return reg.search(item) is not None
+
+
+class DatastoreService(Service, FilterMixin, SchemaMixin):
```
william-gr: FilterMixin already inherits SchemaMixin, not sure you need it here as well?

src/middlewared/middlewared/plugins/datastore/write.py
```diff
@@ -0,0 +1,112 @@
+from sqlalchemy import and_, types
+
+from middlewared.schema import accepts, Any, Dict, Str
+from middlewared.service import Service
+
+from .filter import FilterMixin
+from .schema import SchemaMixin
+
+
+class DatastoreService(Service, FilterMixin, SchemaMixin):
```
william-gr: Same here about SchemaMixin

src/middlewared/middlewared/schema.py
```diff
@@ -957,16 +957,12 @@ def nf(*args, **kwargs):
...
-        nf.__name__ = f.__name__
-        nf.__doc__ = f.__doc__
-        # Copy private attrs to new function so decorators can work on top of it
-        # e.g. _pass_app
-        for i in dir(f):
-            if i.startswith('__'):
-                continue
-            if i.startswith('_'):
-                setattr(nf, i, getattr(f, i))
+        from middlewared.utils.type import copy_function_metadata
```
william-gr: Does this really need to be lazy imported?

src/middlewared/middlewared/plugins/datastore/read.py
themylogin: I was thinking that explicit is better than implicit and as I use SchemaMixin methods in this class, I want to state it explicitly.

src/middlewared/middlewared/schema.py
themylogin: Yes, otherwise it would be circular dependency to break which we'll need to move stuff out of utils/__init__.py

PR #3708, opened 2019-10-24
NAS-103591 / 11.3 / Validate directory service config during health checks

anodos325:
When LDAP or AD is enabled, certain assumptions are made about the
service having valid data in the service config. Since the presence
of bad data in the config can adversely affect system behavior
outside of the service in question (and there are some online guides
suggesting using the sqlite command to manually edit the database),
perform an additional configuration validation step the respective
<ds>.started methods. These methods are called in periodic health
checks as part of the directory service alerts, and will convert
the ValidationError into a proper alert. When this check fails,
disable the directory service in question to help recover from
undefined behavior.

### DIFF
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -1124,15 +1124,17 @@ class ActiveDirectoryService(ConfigService):
             return False
 
         await self.common_validate(config, config, verrors)
-        if verrors:
+
+        try:
+            verrors.check()
+        except Exception as e:
             await self.middleware.call(
                 'datastore.update',
                 'directoryservice.activedirectory',
                 config['id'],
                 {'ad_enable': False}
             )
-            cerrors = ' '.join(v.errmsg for v in verrors.errors)
-            raise CallError(cerrors, errno.EINVAL)
+            raise CallError(e) 
 
         netlogon_ping = await run([SMBCmd.WBINFO.value, '-P'], check=False)
         if netlogon_ping.returncode != 0:
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -634,6 +634,27 @@ class ActiveDirectoryService(ConfigService):
             )
         return
 
+    @private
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1100,9 +1118,22 @@ def validate_domain(self, data=None):
...
 
+        await self.common_validate(config, config, verrors)
+        if verrors:
+            await self.middleware.call(
+                'datastore.update',
+                'directoryservice.activedirectory',
+                config['id'],
+                {'ad_enable': False}
+            )
+            cerrors = ' '.join(v.errmsg for v in verrors.errors)
```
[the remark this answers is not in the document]
william-gr: @anodos325 ping

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1100,9 +1118,24 @@ def validate_domain(self, data=None):
...
+        try:
+            verrors.check()
+        except Exception as e:
+            await self.middleware.call(
+                'datastore.update',
+                'directoryservice.activedirectory',
+                config['id'],
+                {'ad_enable': False}
+            )
+            raise CallError(e)
```
william-gr: @anodos325 I think you may have misunderstood me, if its validation error it seems there is no need for CallError.

Perhaps just `raise` here would do the trick?

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -613,10 +644,21 @@ def get_dn(self, dn=None, ldap=None):
...
+        await self.common_validate(ldap, ldap, verrors)
+        if verrors:
+            await self.middleware.call(
+                'datastore.update',
+                'directoryservice.ldap',
+                ldap['id'],
+                {'ldap_enable': False}
+            )
+            cerrors = ' '.join(v.errmsg for v in verrors.errors)
+            raise CallError(cerrors, errno.EINVAL)
```
william-gr: And same here, perhaps just `raise verrors` would do the trick here? Not understanding why this should be a CallError?

PR #3710, opened 2019-10-24
feat(middlewared/system): add `product_type`

william-gr:
Ticket:	NAS-103601


william-gr:
Maybe in the future, I am not sure that buy us much unless we could have a different meaning.

PR #3713, opened 2019-10-24
NAS-103612 / 11.2 / log when vmware vm isn't powered on

yocalebo:
This is in our documentation but we should log it in a log file so TrueNAS support has a way of identifying the perplexing issue for when vmware aware snapshots don't seem to be working when it's by design.


yocalebo:
backport

PR #3721, opened 2019-10-25
NAS-103387 / 12 / Add authorized keys to config

sonicaj:


src/middlewared/middlewared/plugins/config.py
```diff
@@ -16,7 +16,9 @@
 from middlewared.plugins.pwenc import PWENC_FILE_SECRET
 from middlewared.plugins.pool import GELI_KEYPATH
 
-CONFIG_FILES = {'pwenc_secret': PWENC_FILE_SECRET, 'geli': GELI_KEYPATH}
+CONFIG_FILES = {
+    'pwenc_secret': PWENC_FILE_SECRET, 'geli': GELI_KEYPATH, 'authorized_keys': '/root/.ssh/authorized_keys'
```
themylogin: Can this be formatted properly please?

src/middlewared/middlewared/plugins/config.py
```diff
@@ -28,6 +32,7 @@ class ConfigService(Service):
         'configsave',
         Bool('secretseed', default=False),
         Bool('pool_keys', default=False),
+        Bool('authorized_keys', default=False),
```
william-gr: Maybe root_authorized_keys to make it clearer?

Also need the docstring updated.

src/middlewared/middlewared/plugins/config.py
```diff
@@ -37,6 +42,8 @@ class ConfigService(Service):
         `secretseed` will include the password secret seed in the bundle.
 
         `pool_keys` when set will include the geli encryption keys in the bundle.
+
+        `root_authorized_keys` when set will include "root" users "authorized_keys" file in the debug.
```
william-gr: This is not a debug

src/middlewared/middlewared/plugins/config.py
```diff
@@ -37,6 +42,8 @@ class ConfigService(Service):
         `secretseed` will include the password secret seed in the bundle.
 
         `pool_keys` when set will include the geli encryption keys in the bundle.
+
+        `root_authorized_keys` when set will include "root" users "authorized_keys" file in the bundle.
```
william-gr: Just noticed `"root" users` also sounds strange. Perhaps ask docs review :P

sonicaj: :P

wblock: This is really unclear.  Is it supposed to be possessive?  In that case, avoid possessives, they are often unclear.  "When set" is not needed in any of these, and using present tense (include) is clearer than future tense (will include).  Can also factor out "in the bundle", like
```
Create a bundle of security-sensitive information. These options select which information is included in the bundle:

`secretseed`: include password secret seed.
`pool_keys`: include GELI encryption keys.
`root_authorized_keys`: include "authorized_keys" file for the root user.
```

sonicaj:
@wblock so if no of the options is selected, we wouldn't have a bundle. Just the database file. Can you please advise how we should factor that in the documentation ? Thank you

william-gr:
@wblock ping

PR #3729, opened 2019-10-29
NAS-103620 / 11.3 / fix(middlewared/download): do not close pipes for running download

william-gr:
Also increase read timeout for downloads.

Ticket:	NAS-103620

### DIFF
--- a/src/middlewared/middlewared/main.py
+++ b/src/middlewared/middlewared/main.py
@@ -377,7 +377,14 @@ class FileApplication(object):
         self.jobs[job_id] = self.middleware.loop.call_later(
             60, lambda: asyncio.ensure_future(self._cleanup_job(job_id)))
 
+    async def _cleanup_cancel(self, job_id):
+        job = self.jobs.pop(job_id, None)
+        if job:
+            job.cancel()
+
     async def _cleanup_job(self, job_id):
+        if job_id not in self.jobs:
+            return
         self.jobs[job_id].cancel()
         del self.jobs[job_id]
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/main.py
```diff
@@ -377,7 +377,14 @@ def register_job(self, job_id):
         self.jobs[job_id] = self.middleware.loop.call_later(
             60, lambda: asyncio.ensure_future(self._cleanup_job(job_id)))
 
+    async def _cleanup_cancel(self, job_id):
+        job = self.jobs.pop(job_id, None)
```
themylogin: Took me a while to understand what's going on. Can we name it `job_cleanup`? For me it looked like we are canceling the job itself.

src/middlewared/middlewared/main.py
```diff
@@ -439,9 +446,10 @@ def do_copy():
                 asyncio.run_coroutine_threadsafe(resp.write(read), loop=self.loop).result()
 
         try:
+            await self._cleanup_cancel(job_id)
             await self.middleware.run_in_thread(do_copy)
         finally:
-            await self._cleanup_job(job_id)
+            await job.pipes.close()
```
themylogin: Code duplication with `_cleanup_job`

src/middlewared/middlewared/main.py
william-gr: We already have a _cleanup_job method. Why would we call it job_cleanup if its really canceling a future?

themylogin: @william-gr I mean name variable `job_cleanup`

PR #3740, opened 2019-10-30
NAS-103678 / 11.2 / make sure to not run vmware-checkvm on truenas

yocalebo:
Jira: NAS-103678


yocalebo:
@william-gr yep! I'll have to apply it differently since 11.3 uses different mechanism to write rc.conf.

PR #3742, opened 2019-10-30
NAS-103676 / 11.3 / Ignore OS Boot IPMI SEL events

themylogin:
Colud not find anything else to ignore here https://github.com/openbmc/ipmitool/blob/master/include/ipmitool/ipmi_sel.h#L297

### DIFF
--- a/src/middlewared/middlewared/alert/source/ipmi_sel.py
+++ b/src/middlewared/middlewared/alert/source/ipmi_sel.py
@@ -87,6 +87,41 @@ class IPMISELAlertSource(AlertSource):
 
     dismissed_datetime_kv_key = "alert:ipmi_sel:dismissed_datetime"
 
+    # https://github.com/openbmc/ipmitool/blob/master/include/ipmitool/ipmi_sel.h#L297
+
+    IPMI_SENSORS = (
+        "Redundancy State",
+        "Temperature",
+        "Voltage",
+        "Current",
+        "Fan",
+        "Physical Security",
+        "Platform Security",
+        "Processor",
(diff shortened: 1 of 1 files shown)


william-gr:
@yocalebo can we get your input here? What kind of IPMI events should we ignore? Or is it better/easier to say which one we should report?

I can think of some other useless like:
ACPI Device Power State
Availability State
Usage State
System Event

yocalebo:
@william-gr I like the approach of ignoring all and only reporting the events that we know we're interested in but I'm not sure of the ones that we're interested in.

Another one to ignore on TrueNAS HA is `Watchdog2 #0xca Asserted Timer interrupt ()` message which is triggered on a failover and we can't export the zpool within 4 seconds and we trigger the software based watchdog timeout.

william-gr:
Here is the list I think we should alert on:

Battery
Management Subsystem Health
Cable/Interconnect - Config Error
Critical Interrupt
System Event - Undetermined system hardware failure
Event Logging Disabled - Log full
Event Logging Disabled - Log almost full
System Firmware Error
Memory
Power Unit - Soft-power control failure
Power Unit - Failure detected
Power Unit - Predictive failure
Power Supply - All but "Presence detected"
Processor - All but "Presence detected"
Platform Security
Physical Security
Temperature
Voltage
Current
Fan
Redundancy State - All but "Fully Redundant"


Thoughts?

darkfiberiru:
Definitely a good start nothing to add or remove at this moment for me.

yocalebo:
Yeah, looks good. Like that list.

PR #3745, opened 2019-10-31
NAS-103625 / 11.3 / Remove unused alert services attributes

themylogin:


src/middlewared/middlewared/alert/base.py
```diff
@@ -218,6 +218,12 @@ def __init__(self, middleware, attributes):
 
         self.logger = logging.getLogger(self.__class__.__name__)
 
+        # If we remove some attributes, it should not be an error if they are still left in the database
+        schema = self.schema.copy()
```
william-gr: Why dont we have migration to remove these from database?

Wouldnt this allow user to input anything they want and make it part of the database?

themylogin: > Why dont we have migration to remove these from database?

We can have a migration, but we also can avoid having a migration :)

> Wouldn't this allow user to input anything they want and make it part of the database?

This method is not used to validate user's input, it is only used to construct an alert service when sending alerts. The method that validates the input does that according to the proper schema.

PR #3751, opened 2019-11-01
NAS-103607 / 12 / Replication job events

themylogin:


themylogin:
Honestly have not even thought about it, why?

william-gr:
Hooks seems cleaner. Hooks being used for internal middlewared, and events only for clients/websocket

PR #3752, opened 2019-11-01
RunInThreadMixin

themylogin:


src/middlewared/middlewared/utils/run_in_thread.py
```diff
@@ -0,0 +1,13 @@
+import functools
+
+from .io_thread_pool_executor import IoThreadPoolExecutor
+
+
+class RunInThreadMixin:
+    loop = None  # Must be provided by child class
+
+    # Will be shared among all children classes but that's ok for our use case
+    __io_threadpool = IoThreadPoolExecutor('IoThread', 20)
```
william-gr: 20 min thread workers for each process worker seems rather excessive, no?

That pool is mostly used within the main process and I would think its very rare (if even possible) where we use more than 1 within the process worker?

PR #3753, opened 2019-11-04
Travis CI: Update environment, Python and add Arm job

EwoutH:
- Updates build environment from Ubuntu 16.04 Xenial to 18.04 Bionic
 - Updates Python version to 3.8
 - Adds a job on Arm64
 - Adds a job with Python nightly (which is allowed to fail)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,6 @@ jobs:
   - python: nightly
 
 install:
-    - sudo apt-get -qq update
-    - sudo apt-get -y install git
     - pip3 install flake8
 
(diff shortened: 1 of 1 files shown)


EwoutH:
@william-gr Can you review?

EwoutH:
Strange failure on Arm64, it ran fine on my side: https://travis-ci.com/EwoutH/freenas/jobs/252571300

william-gr:
@EwoutH I dotn understand why we need this in the first place. Especially arm64 and nightly.

Are you familiar with freenas at all?

PR #3760, opened 2019-11-05
NAS-103779 / 12 / Expose bindip choices for S3 service

sonicaj:
1) Extends ip choices method to filter static ips
2) Exposes bindip choices method for s3 service
3) Validate bindip attribute for s3 service

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -2050,6 +2050,7 @@ class InterfaceService(CRUDService):
             Bool('ipv6', default=True),
             Bool('loopback', default=False),
             Bool('any', default=False),
+            Bool('static', default=False),
         )
     )
     def ip_in_use(self, choices=None):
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -2082,6 +2085,13 @@ def ip_in_use(self, choices=None):
         if not choices['loopback']:
             ignore_nics.append('lo')
         ignore_nics = tuple(ignore_nics)
+        static_ips = {}
+        if choices['static']:
+            for i in self.middleware.call_sync('interface.query'):
+                for alias in i.get('failover_virtual_aliases') or []:
```
william-gr: For HA we should only show virtual aliases and not regular aliases. We dont want people binding to each node's IP.

src/middlewared/middlewared/plugins/network.py
```diff
@@ -2082,6 +2085,19 @@ def ip_in_use(self, choices=None):
...
+                    static_ips[alias['address']] = alias['address']
+                for alias in (
+                    i['aliases'] if (
+                        not self.middleware.call_sync('failover.config')['disabled'] and self.middleware.call_sync(
```
william-gr: Two things:

1. calling call_sync for every alias for every interface
2. checking only once for `failover.licensed` is good enough

PR #3762, opened 2019-11-05
NAS-103729 / 12 / Expose a new has_private attribute

sonicaj:
This commit introduces a new method on attribute 'has_private' which ensures that if we have a dict or dicts in a list's items, where any dict has an attribute marked as private, that is respected.

### DIFF
--- a/src/middlewared/middlewared/pytest/unit/test_schema.py
+++ b/src/middlewared/middlewared/pytest/unit/test_schema.py
@@ -331,6 +331,14 @@ def test__schema_dict_mixed_args(value, expected, msg):
         assert dictargs(self, value) == expected
 
 
+@pytest.mark.parametrize('items,value', [
+    ([List('b', items=[List('c', private=True)])], [[['a']]]),
+    ([Dict('b', Str('c', private=True))], [{'c': 'secret'}])
+])
+def test__schema_list_private_items(items, value):
+    assert List('a', items=items).dump(value) == '********'
+
+
 def test__schema_list_empty():
 
     @accepts(List('data', empty=False))
--- a/src/middlewared/middlewared/schema.py
+++ b/src/middlewared/middlewared/schema.py
@@ -85,7 +85,7 @@ class Attribute(object):
         return self.private
 
     def dump(self, value):
-        if self.has_private():
+        if self.private:
             return "********"
 
         return value
@@ -472,6 +472,11 @@ class List(EnumMixin, Attribute):
 
     def has_private(self):
         return self.private or any(item.has_private() for item in self.items)
+
+    def dump(self, value):
+        if self.has_private():
+            return '********'
+        return value
 
     def validate(self, value):
         if value is None:
--- a/src/middlewared/middlewared/schema.py
+++ b/src/middlewared/middlewared/schema.py
@@ -81,8 +81,11 @@ class Attribute(object):
                 raise Error(self.name, 'attribute required')
         return value
 
+    def has_private(self):
+        return self.private
+
     def dump(self, value):
-        if self.private:
+        if self.has_private():
             return "********"
 
         return value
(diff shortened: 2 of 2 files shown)


src/middlewared/middlewared/schema.py
```diff
@@ -81,6 +81,9 @@ def clean(self, value):
                 raise Error(self.name, 'attribute required')
         return value
 
+    def has_private(self):
```
themylogin: Can this be named `is_private` and just be used in `dump` like `if self.is_private(): return "***"`?

sonicaj: So `is_private` as a name is quite similar to `private` which means the attribute in question is private. However, `has_private` serves a use case which indicates that the attribute has other children which are private and if the attribute is being used somewhere, the user of the attribute can decide how to handle the children case

themylogin: If anything else is private, children are also private, how else can it be?

sonicaj: How about attribute is not private but it has a few children which are, we do not want to hide the parent then, do we ?

themylogin: When attribute is not private, we dump it, and children will hide themselves when dumping recursively. I've mean use it like this in `List.dump`

src/middlewared/middlewared/schema.py
```diff
@@ -574,6 +577,9 @@ def __init__(self, name, *attrs, **kwargs):
                         raise ValueError(f"Attribute {attr.name} is not required and does not have default value, "
                                          f"this is forbidden in strict mode")
 
+    def has_private(self):
+        return self.private or any(i.private or i.has_private() for i in self.attrs.values())
```
themylogin:
```suggestion
        return self.private or any(i.has_private() for i in self.attrs.values())
```

src/middlewared/middlewared/schema.py
```diff
@@ -468,7 +471,7 @@ def clean(self, value):
         return value
 
     def dump(self, value):
-        if self.private or (self.items and any(item.private for item in self.items)):
+        if self.private or (self.items and any(item.private or item.has_private() for item in self.items)):
```
themylogin:
```suggestion
        if self.has_private():
```

themylogin:
```
def test__1():
    s = List('a', items=[List('b', items=[List('c', private=True)])])
    assert s.dump([[['d']]]) == '********'
```

src/middlewared/middlewared/schema.py
themylogin: Also you may promote this implementation to base class

src/middlewared/middlewared/schema.py
```diff
@@ -467,11 +470,8 @@ def clean(self, value):
                     raise Error(self.name, 'Item#{0} is not valid per list types: {1}'.format(index, found))
         return value
 
-    def dump(self, value):
-        if self.private or (self.items and any(item.private for item in self.items)):
-            return "********"
-
-        return value
+    def has_private(self):
+        return self.private or (self.items and any(item.has_private() for item in self.items))
```
themylogin: `any` will return `False` for empty list
```suggestion
        return self.private or any(item.has_private() for item in self.items)
```

src/middlewared/middlewared/pytest/unit/test_schema.py
```diff
@@ -331,6 +331,14 @@ def dictargs(self, data):
         assert dictargs(self, value) == expected
 
 
+def test__schema_list_private():
+    for items, value in (
```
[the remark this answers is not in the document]
sonicaj: Thank you @themylogin for clarifying. I have addressed your concerns, please do let me know if something's amiss. Thanks again  :)

src/middlewared/middlewared/pytest/unit/test_schema.py
```diff
@@ -331,6 +331,14 @@ def dictargs(self, data):
         assert dictargs(self, value) == expected
 
 
+@pytest.mark.parametrize('items,value', [
+    ([List('b', items=[List('c', private=True)])], [[['a']]]),
+    ([Dict('b', Str('c', private=True))], [{'c': 'secret'}])
+])
+def test__schema_list_private_items(items, value):
+    assert List('a', items=items).dump(value) == '********'
```
william-gr: Can we also add a test with a List with multiple Dict as items with private attributes?

Or not worth it?

william-gr: Err, nevermind me, i am dumb, that what is happening here : )

PR #3768, opened 2019-11-06
Sync master from 19/11/06 to 12-devel

william-gr:


william-gr:
@sonicaj there was a conflict on config and jail plugin, please verify only these

PR #3777, opened 2019-11-08
NAS-103839 / 11.2 / Allow non-GSSAPI binds to LDAP in environments with a kerberos realm

anodos325:
Previous behavior would always perform SASL_GSSAPI if
kerberos realm and keytab were selected in the LDAP form
in the UI. There are situations where kerberos should be
configured, but not used for actual LDAP binds. Decouple
when it makes sense to do so.

While Active Directory permits SASL binds to be performed
on an SSL/TLS-protected connection, it does not permit the
use of SASL-layer encryption/integrity verification
mechanisms on such a connection. While this restriction is
present in Active Directory on Windows 2000 Server
operating system and later, versions prior to Windows
Server 2008 operating system can fail to reject an LDAP
bind that is requesting SASL-layer encryption/integrity
verification mechanisms when that bind request is sent on a
SSL/TLS-protected connection. See MS-ADTS 5.1.1.1.2

Samba AD Domain controllers also require the following
smb.conf parameter in order to permit SASL_GSSAPI on an SSL/
TLS-protected connection:

'ldap server require strong auth = allow_sasl_over_tls'

This means that LDAP server support for SASL_GSSAPI on an SSL/
TLS-protected connection is unpredictable / undefined.

Add workaround if this combination is _required_ for environment
previous behavior can be restored by setting FREENAS_LDAP_FORCE_GSSAPI
to 1 in /etc/directoryservice/rc.DS_Common

### DIFF
--- a/src/freenas/etc/ix.rc.d/ix-ldap
+++ b/src/freenas/etc/ix.rc.d/ix-ldap
@@ -161,7 +161,7 @@ ldap_status()
 			do_gssapi_bind=0
 		fi
 
-		if [ -n "${realm}" -a -n "${principal}" -a -n "${do_gssapi_bind}" ]
+		if [ -n "${realm}" -a -n "${principal}" -a "${do_gssapi_bind}" == 0 ]
 		then
 			local temp=$(mktemp /tmp/tmp.XXXXXX)
 			local cmdfile=$(mktemp /tmp/tmp.XXXXXX)
--- a/src/freenas/etc/ix.rc.d/ix-ldap
+++ b/src/freenas/etc/ix.rc.d/ix-ldap
@@ -14,13 +14,13 @@ LDAP_STATUS_ALERT="/tmp/.ldap_status_alert"
 
 generate_ldapconf()
 {
-	LD_LIBRARY_PATH=/usr/local/lib /usr/local/bin/midclt call etc.generate ldap > /dev/null
+	LD_LIBRARY_PATH=/usr/local/lib /usr/local/bin/midclt call etc.generate ldap 2> /dev/null
 }
 
 generate_nss_ldapconf()
 {
 	rm -f /usr/local/etc/ldap.conf
-	LD_LIBRARY_PATH=/usr/local/lib /usr/local/bin/midclt call etc.generate nss > /dev/null
+	LD_LIBRARY_PATH=/usr/local/lib /usr/local/bin/midclt call etc.generate nss 2> /dev/null
 	ln -sf /usr/local/etc/nss_ldap.conf /usr/local/etc/ldap.conf
 }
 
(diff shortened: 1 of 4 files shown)


src/freenas/etc/ix.rc.d/ix-ldap
```diff
@@ -134,13 +134,31 @@ ldap_status()
...
+		then
+			allow_gssapi_bind=1
+		fi
+
+		if [ "${ldap_anonbind}" = "1" ]
+		then
+			allow_gssapi_bind=1
+		fi
+
+		if [ "force_gssapi" = "0" ]
```
william-gr: ${force_gssapi} ?

anodos325: Indeed

anodos325:
> @anodos325 Maybe I'm failing to understand, but we were actually failing to run GSSAPI ldapsearch'es because they had LDAPS enabled. I thought we discussed that it's not possible to do GSSAPI based requests over LDAPS or TLS connections? This logic looks like you're forcing GSSAPI binds when principle,realm, and ssl is on. I reversed that logic on customer system. So do we still need to look at ix-ldap rc script? After upgrading the customers system and everything failing again, I ended up having to have the logic of
> 
> `if [ -n "${realm}" -a -n "${principle}" -a "${ldap_ssl}" != "on" ]; then run gssapi ldapsearch'es`
ldap_ssl has three values "off", "on", and "start_tls". I'm renaming the "allow_gssapi_bind" to "do_gssapi_bind" in ix-ldap to make it clearer what we're doing. I misremembered the possible "SSL" values for ldap_ssl and so I've fixed that in ix-ldap to reflect reality. In this case since we're in a shell script 0 is True and 1 is False. In the case of freenasldap.py, I apparently inverted logic of want_ssl in simplifying my initial code there, so good catch.
>
> Also, in the `ix-ldap` script the 2 middleware calls aren't redirecting standard error only standard out.
Fixed.

src/freenas/etc/ix.rc.d/ix-ldap
```diff
@@ -134,13 +134,34 @@ ldap_status()
...
+			do_gssapi_bind=1
+		fi
+
+		if [ "${force_gssapi}" = "0" ]
+		then
+			do_gssapi_bind=0
 		fi
 
-		if [ -n "${realm}" -a -n "${principal}" ]
+		if [ -n "${realm}" -a -n "${principal}" -a -n "${do_gssapi_bind}" ]
```
william-gr: Seems this is not how you compare, unless i am missing something:

```
william@stone[~]: do_gssapi_bind=0
william@stone[~]: test -n "${do_gssapi_bind}"
william@stone[~]: echo $?
0
william@stone[~]: do_gssapi_bind=1           
william@stone[~]: test -n "${do_gssapi_bind}"
william@stone[~]: echo $?  
0
```

src/freenas/etc/ix.rc.d/ix-ldap
```diff
@@ -134,13 +134,34 @@ ldap_status()
...
+			do_gssapi_bind=1
+		fi
+
+		if [ "${force_gssapi}" = "0" ]
+		then
+			do_gssapi_bind=0
 		fi
 
-		if [ -n "${realm}" -a -n "${principal}" ]
+		if [ -n "${realm}" -a -n "${principal}" -a "${do_gssapi_bind}" == 0 ]
```
william-gr: Hmm, I don't think `==` works in shell, i think you're looking for `${do_gssapi_bind} -eq 0` ?

PR #3778, opened 2019-11-08
NAS-102358 / 12 / Switch from active/passive to active/standby terms

themylogin:


src/middlewared/middlewared/etc_files/local/smb4.conf
```diff
@@ -362,7 +362,7 @@
             pc = {}
             if db['truenas_conf']['smb_ha_mode'] == 'UNIFIED' and db['truenas_conf']['failover_status'] != 'MASTER':
                 stub_config = {
-                    'netbios name': f"{db['gc']['hostname_virtual']}_PASSIVE",
+                    'netbios name': f"{db['gc']['hostname_virtual']}_STANDBY",
```
themylogin: @anodos325 can we do this safely?

yocalebo: @themylogin yes, you can do this safely without effecting anything.

PR #3780, opened 2019-11-08
NAS-103833 / 11.3 / Add error handling for duplicate gids when doing smb.groupmap_add

anodos325:
Samba's group mapping database maps unix groups to nt groups. It doesn't allow duplicate
generating entries for groups with duplicate gids.  Add some error handling for when
smb.groupmap_add encounters this situation.

Unfortunately, net groupmap add doesn't output to stderr at -d 0 when it encounters
duplicate gids. Wrap smb.groupmap_add in a try / except and only
raise exception if allow_duplicate_gids is False. This indicates
a different error condition.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -830,7 +830,7 @@ class GroupService(CRUDService):
 
         `users` is a list of user ids (`id` attribute from `user.query`).
         """
-
+        allow_duplicate_gid = data['allow_duplicate_gid']
         verrors = ValidationErrors()
         await self.__common_validation(verrors, data, 'group_create')
         verrors.check()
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/account.py
```diff
@@ -851,7 +851,18 @@ class Config:
...
-        await self.middleware.call('smb.groupmap_add', data['name'])
+        try:
+            await self.middleware.call('smb.groupmap_add', data['name'])
+        except Exception as e:
+            """
+            Samba's group mapping database does not allow duplicate gids.
+            Unfortunately, we don't get a useful error message at -d 0.
+            """
+            if not allow_duplicate_gid:
+                raise e
```
william-gr: I think the proper way to re-raise here is just `raise`

PR #3783, opened 2019-11-10
NAS-103881 / 11.3 / Introduce NetMask Validator

sonicaj:


src/middlewared/middlewared/plugins/ipmi.py
```diff
@@ -71,7 +72,7 @@ class IPMIService(CRUDService):
     @accepts(Int('channel'), Dict(
         'ipmi',
         IPAddr('ipaddress'),
-        Str('netmask'),
+        Str('netmask', validators=[Netmask(ipv6=False), IpAddress()]),
```
themylogin: What you needed here was just `IPv4Netmask`, I doubt there would be any need for "either IPv4 netmask or number between 1 and 128" validator. Maybe I am stupid but I had to actually run this code to understand what does `validators=[Netmask(ipv6=False), IpAddress()]` mean.

sonicaj: Umm we do plan on supporting ipv6 in 12, should we still make the validator use ipv4netmask only ?

themylogin: Then make it `Netmask(ipv6=False, prefix_length=False)`

And also `IPAddr('ipaddress', ipv6=False)` (same for gateway) if we don't yet support IPv6

PR #3795, opened 2019-11-12
NAS-103915 / 12 / Disable specifying key file when creating encrypted dataset/pool

sonicaj:


william-gr:
Ah, you removed the job, I am blind. Just the empty file then ; )

sonicaj:
:p

PR #3796, opened 2019-11-12
NAS-103917 / 11.3 / Update debug output for ActiveDirectory

anodos325:
Add more useful output to the AD and SMB debug
- dump passdb.tdb
- get clockskew
- get output of testjoin
- add new fields and remove old ones
- get share configuration


anodos325:
> Is there a ticket yet? If not you/I can add "jira" label in here for the bot to create it.

Not yet, I'll do that. I'm mostly seeking feedback here on what other things we might want to include. I'll probably add another commit for some LDAP-related information as well.

PR #3802, opened 2019-11-13
feat(middlewared/worker): alternative for when kqueue is not available

william-gr:


themylogin:
Are we interested in better Linux implementation (e.g. using `PR_SET_PDEATHSIG`)?

william-gr:
That would be good, do you want to take this over?

themylogin:
Yes, I'll take this over

PR #3805, opened 2019-11-13
NAS-103865 / 11.3 / Use proper ups identifier in reporting

themylogin:


sonicaj:
@themylogin it's possible that this would be different when UPS service is running as a SLAVE. I have opened up another PR addressing an issue on that end. Have you verified if this works as desired when running in SLAVE mode ?

themylogin:
@william-gr we've figured this out in adjacent PR, no changes needed.

PR #3807, opened 2019-11-14
NAS-103944 / 11.3 / Bug fixes for UPS

sonicaj:
This PR introduces the following changes:

1) Correctly set ups details for collectd plugin
2) Ensure we respect the port settings for UPS service

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/nut/upsd.conf
+++ b/src/middlewared/middlewared/etc_files/local/nut/upsd.conf
@@ -2,10 +2,10 @@
     ups_config = middleware.call_sync('ups.config')
 %>\
 % if ups_config['rmonitor']:
-LISTEN 0.0.0.0
-LISTEN ::0
+LISTEN 0.0.0.0 ${ups_config['remoteport']}
+LISTEN ::0 ${ups_config['remoteport']}
 % else:
-LISTEN 127.0.0.1
-LISTEN ::1
+LISTEN 127.0.0.1 ${ups_config['remoteport']}
+LISTEN ::1 ${ups_config['remoteport']}
 % endif
 ${ups_config['optionsupsd']}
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/ups.py
```diff
@@ -31,6 +31,8 @@ class Config:
         data['mode'] = data['mode'].upper()
         data['shutdown'] = data['shutdown'].upper()
         data['toemail'] = [v for v in data['toemail'].split(';') if v]
+        host = 'localhost' if data['mode'] == 'MASTER' else data['remotehost']
+        data['complete_identifier'] = f'{data["identifier"]}@{host}:{data["remoteport"]}'
```
themylogin: IIRC correctly if you don't pop this in do_update, it will raise an exception?

sonicaj: It certainly will and i have made the change locally : ) Testing master/slave configuration right now and will push this change and any other if required : )

sonicaj:
@themylogin I have asked the reporter to share an output of `ls -l /var/db/collectd/rrd/localhost/`. I am not sure what the folder would be named with SLAVE. UPS service worked for me in SLAVE mode but maybe dummy-ups is not handling it okay for the collectd plugin ? I am unsure.

themylogin:
@sonicaj it would be named the same way (after what's before `@`) so there is no special handling needed for slave mode.

sonicaj:
In that case it's ready for review :)

src/middlewared/middlewared/etc_files/local/nut/upsd.conf
```diff
@@ -2,10 +2,10 @@
     ups_config = middleware.call_sync('ups.config')
 %>\
 % if ups_config['rmonitor']:
-LISTEN 0.0.0.0
-LISTEN ::0
+LISTEN 0.0.0.0 ${ups_config['remoteport']}
```
themylogin: You should add some validation on `remoteport` since it's now it's in use for both modes.

PR #3813, opened 2019-11-15
NAS-103887 / 11.3 / Save iocage configs in BE

sonicaj:
This PR introduces changes which save default/jail/template configurations in boot environment before applying an update ensuring when the user rolls back, they have a way of bringing back their iocage jails.


sonicaj:
We have decided to handle this in iocage/iocage instead as it's a much better way to tackle this issue.

PR #3815, opened 2019-11-18
NAS-103906 / 12 / KMIP Support

sonicaj:
This PR aims to add support for TN with an external KMIP server which manages ZFS encrypted keys and SED keys


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2869,16 +2870,31 @@ class Config:
...
+        for ds in datasets:
+            if ds['encryption_key']:
+                keys[ds['name']] = self.middleware.call_sync('pwenc.decrypt', ds['encryption_key'])
+            elif ds['name'] in zfs_keys:
+                keys[ds['name']] = zfs_keys[ds['name']]
+        return keys
+
+    @private
+    def dataset_datastore(self):
+        return self.dataset_store
```
themylogin: Why do you think this is necessary?

src/middlewared/middlewared/alert/source/kmip.py
```diff
@@ -0,0 +1,15 @@
+from middlewared.alert.base import AlertClass, SimpleOneShotAlertClass, AlertCategory, AlertLevel
+
+
+class KMIPConnectionFailedAlertClass(AlertClass, SimpleOneShotAlertClass):
```
themylogin: You should mark this as `deleted_automatically = False` because you don't delete these alerts programmatically and instead they should be deleted when user dismisses them.

src/middlewared/middlewared/plugins/kmip.py
```diff
@@ -0,0 +1,384 @@
...
+            failed = self.sync_zfs_keys_from_db_to_server(ids)
+        else:
+            failed = self.sync_zfs_keys_from_server_to_db(ids)
+        if failed:
+            self.middleware.call_sync(
+                'alert.oneshot_create', ' KMIPZFSDatasetsSyncFailure', {'datasets': ','.join(failed)}
+            )
+
+    @periodic(interval=86400)
+    @job(lock='sync_kmip_keys')
```
themylogin: Lock is not necessary - sync_zfs_keys protects itself. Also it will be better to name it `periodic_sync_keys` so users can understand what is the difference between `sync_zfs_keys` and `sync_keys`

src/middlewared/middlewared/plugins/kmip.py
```diff
@@ -0,0 +1,384 @@
...
+        return self.zfs_keys
+
+
+async def __event_system(middleware, event_type, args):
+    if args['id'] != 'ready':
+        return
+
+    if (await middleware.call('kmip.config'))['enabled']:
+        await middleware.call('kmip.initialize_keys')
+
```
themylogin: Please look how I've split the datastore plugin into few files and try to do the same with this file. Long files are pain to maintain. I can see at least three parts:
* part that interacts with kmip library (connection context manager and operations)
* part that syncs keys
* part that implements ConfigService

src/middlewared/middlewared/plugins/kmip.py
```diff
@@ -0,0 +1,384 @@
...
+            update=True
+        )
+    )
+    async def do_update(self, data):
+        old = await self.config()
+        new = old.copy()
+        new.update(data)
+        verrors = ValidationErrors()
+
+        if not new['server']:
```
themylogin: Can't I clear this field if I disable KMIP?

src/middlewared/middlewared/plugins/kmip.py
```diff
@@ -0,0 +1,384 @@
...
+            result = await self.middleware.run_in_thread(self.test_connection, new)
+            if result['error']:
+                verrors.add('kmip_update.server', f'Unable to connect to KMIP server: {result["exception"]}.')
+
+        force_clear = new.pop('force_clear', False)
+        sync_error = 'KMIP sync is pending, please make sure database and KMIP server ' \
+                     'are in sync before proceeding with this operation.'
+        if old['enabled'] != new['enabled'] and await self.kmip_sync_pending():
+            if force_clear:
+                await self.clear_sync_pending_keys()
```
themylogin: Maybe this should be performed after `verrors.check()`?

src/middlewared/middlewared/plugins/kmip.py
```diff
@@ -0,0 +1,384 @@
...
+            'datastore.query', await self.middleware.call('pool.dataset.dataset_datastore')
+        ):
+            if config['enabled'] and config['manage_zfs_keys'] and ds['encryption_key']:
+                return True
+            elif any(not config[k] for k in ('enabled', 'manage_zfs_keys')) and ds['kmip_uid']:
+                return True
+        return False
+
+    @private
+    def sync_zfs_keys_from_db_to_server(self, ids=None):
```
themylogin: "pull" and "push" are good terms for that

src/middlewared/middlewared/plugins/kmip.py
sonicaj: `sync_zfs_keys` i think is better wrt API because it's not only a periodic job and users can initiate this themselves as well to get remote server and local db in sync. Please advise if you think we should still rename it, I'll take care of it then

src/middlewared/middlewared/plugins/pool.py
sonicaj: I can retrieve dataset's table name via this method without hardcoding it everywhere

src/middlewared/middlewared/etc_files/pykmip/pykmip.conf
```diff
@@ -0,0 +1,14 @@
+<%
+	kmip_config = middleware.call_sync('kmip.config')
+	cert = middleware.call_sync('certificate.query', [['id', '=', kmip_config['certificate']]])
```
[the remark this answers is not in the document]
sonicaj: Automatically joined as in already having the complete certificate in `kmip.config.certificate` ?

src/middlewared/middlewared/plugins/pool.py
sonicaj: And as the table has already been made and likely won't be changed ever, i think it's safe to use the name directly. I'll update this :p

src/middlewared/middlewared/etc_files/pykmip/pykmip.conf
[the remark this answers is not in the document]
themylogin: Sorry, I am not right

src/middlewared/middlewared/etc_files/pykmip/pykmip.conf
```diff
@@ -0,0 +1,14 @@
...
+	ca = middleware.call_sync('certificateauthority.query', [['id', '=', kmip_config['certificate_authority']]])
```
william-gr: Seems we should have `get: True`

src/middlewared/middlewared/plugins/kmip/sync.py
```diff
@@ -0,0 +1,363 @@
...
+                        self.disks_keys[disk['identifier']] = key
+                    continue
+                elif not disk['passwd']:
+                    continue
+
+                self.disks_keys[disk['identifier']] = disk['passwd']
+                destroy_successful = False
+                if disk['kmip_uid']:
+                    # This needs to be revoked and destroyed
+                    destroy_successful = self._revoke_and_destroy_key(disk['kmip_uid'], conn)
```
william-gr: Isnt there anything we can log how/why revoke and destroy failed? I didnt see it in the call chain, maybe I am missing something

src/middlewared/middlewared/plugins/kmip/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+        return
+
+    if (await middleware.call('kmip.config'))['enabled']:
+        await middleware.call('kmip.initialize_keys')
+
+
+async def setup(middleware):
+    middleware.event_subscribe('system', __event_system)
+    if await middleware.call('system.ready'):
+        await middleware.call('kmip.initialize_keys')
```
william-gr: Dont we need to do that before system boots? Before importing pools?

src/middlewared/middlewared/etc_files/pykmip/pykmip.conf
themylogin: We don't as it will raise exception if `kmip_config['certificate']` is `None`

william-gr: Well, we dont check if cert or ca are empty, so it will raise an IndexError wither way

william-gr: Actually, we do, I am blind, again :D

src/middlewared/middlewared/plugins/disk.py
```diff
@@ -112,6 +115,23 @@ def _compress_enclosure(self, disk):
...
+    @filterable
+    async def query_passwords(self, filters=None, options=None):
+        """
+        Query disks SED passwords with `query-filters` and `query-options`.
+        """
+        disks = await self.middleware.call(
+            'datastore.query', self._config.datastore, filters, {**options, 'prefix': self._config.datastore_prefix}
+        )
+        disks_keys = await self.middleware.call('kmip.retrieve_sed_disks_keys')
+        for disk in ([disks] if not isinstance(disks, list) else disks):
```
william-gr: Why would disks not be a list?

src/middlewared/middlewared/plugins/kmip/sed_keys.py
```diff
@@ -0,0 +1,239 @@
...
+            else:
+                return
+        else:
+            failed = self.pull_sed_keys()
+        ret_failed = failed.copy()
+        if failed:
+            if 'Global SED Key' in failed:
+                failed.remove('Global SED Key')
+                self.middleware.call_sync('alert.oneshot_create', 'KMIPSEDGlobalPasswordSyncFailure')
+            if failed:
```
william-gr: Already checked for `if failed` before, guess you want an `else:` ?

src/middlewared/middlewared/plugins/kmip/sed_keys.py
```diff
@@ -0,0 +1,239 @@
...
+    @private
+    def sync_sed_keys(self, job, ids=None, force=False):
+        if not force and not self.middleware.call_sync('kmip.sed_keys_pending_sync'):
+            return
+        config = self.middleware.call_sync('kmip.config')
+        conn_successful = self.middleware.call_sync('kmip.test_connection', None, True)
+        if config['enabled'] and config['manage_sed_disks']:
+            if conn_successful:
+                failed = self.push_sed_keys(ids)
+            else:
```
william-gr: Should we issue an alert if connection failed?

src/middlewared/middlewared/plugins/disk.py
sonicaj: If `get=True` then it would not be a list

src/middlewared/middlewared/plugins/kmip/sed_keys.py
sonicaj: We do raise it above `kmip. test_connection`, `True` there specifies that an alert should be raised

src/middlewared/middlewared/plugins/kmip/sed_keys.py
sonicaj: Umm nope, but i can still improve it however, I'll do that :) To clarify, `Global SED Key` can be in `failed` with other disks which failed to sync as well. So I remove it first and raise an alert for it specially and then handle the other disk case separately

src/middlewared/middlewared/plugins/disk_/sync.py
```diff
@@ -98,6 +98,13 @@ class DiskService(Service, ServiceChangeMixin):
                         'iscsi.extent.query', [['type', '=', 'DISK'], ['path', '=', disk['disk_identifier']]]
                     ):
                         await self.middleware.call('iscsi.extent.delete', extent['id'])
+                    if disk['disk_kmip_uid']:
+                        try:
+                            await self.middleware.call('kmip.delete_kmip_secret_data', disk['disk_kmip_uid'])
```
william-gr: Does this contact KMIP server? Do we really need to wait for it to happen?

src/middlewared/middlewared/plugins/kmip/connection.py
```diff
@@ -0,0 +1,91 @@
+import contextlib
+import socket
+
+from kmip.core import enums
+from kmip.pie.client import ProxyKmipClient
+from kmip.pie.exceptions import ClientConnectionFailure, ClientConnectionNotOpen, KmipOperationFailure
+from kmip.pie.objects import SecretData
```
william-gr: Can you add kmip dependency to the debian package?

src/middlewared/middlewared/plugins/kmip/sed_keys.py
```diff
@@ -0,0 +1,242 @@
+from middlewared.service import job, private, Service
```
william-gr: I think this file could use comments spread all over explaining what and why we are doing the things we are because its a bit convoluted and not direct for people nor familiar.

william-gr:
@themylogin  could really use a second pair of eyes here, I am a bit overwhelmed

src/middlewared/middlewared/plugins/disk_/sync.py
sonicaj: It does and yes we don't. Will make required changes

src/middlewared/middlewared/plugins/system.py
```diff
@@ -213,8 +218,20 @@ class Config:
                 original_data['sed_user'] = original_data['sed_user'].lower()
             if config_data.get('sed_user'):
                 config_data['sed_user'] = config_data['sed_user'].lower()
-
-            # PASSWORD ENCRYPTION FOR SED IS BEING DONE IN THE MODEL ITSELF
+            if config_data['sed_passwd']:
+                config_data['sed_passwd'] = await self.middleware.call('pwenc.encrypt', config_data['sed_passwd'])
```
themylogin: Please use this https://github.com/freenas/freenas/pull/3970/files

src/middlewared/middlewared/plugins/disk.py
```diff
@@ -161,10 +175,7 @@ def _compress_enclosure(self, disk):
         """
 
         old = await self.middleware.call(
-            'datastore.query',
-            self._config.datastore,
-            [('identifier', '=', id)],
-            {'prefix': self._config.datastore_prefix, 'get': True}
+            'datastore.query', 'storage.disk', [['identifier', '=', id]], {'get': True, 'prefix': 'disk_'}
```
william-gr: Why changed self._config.datastore_prefix to text?

src/middlewared/middlewared/plugins/disk.py
```diff
@@ -676,10 +688,12 @@ def check_clean(self, disk):
     @private
     async def sed_unlock_all(self):
         advconfig = await self.middleware.call('system.advanced.config')
-        disks = await self.middleware.call('disk.query')
+        disks = await self.query([], {'extra': {'passwords': True}})
```
william-gr: I personally prefer to keep middleware.call in these situations, because if we change query to being a method this wont silently break

PR #3819, opened 2019-11-19
NAS-103963 / 11.3 / Disable NIC capabilities when using Jail/VM(s) for TN HA

sonicaj:
This PR introduces changes which ensure we disable NIC capabilities for interfaces being used by jail/vm when HA is licensed.

### DIFF
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -841,48 +841,7 @@ class JailService(CRUDService):
         if name:
             iocage.rename(name)
 
-        new = self.middleware.call_sync('jail._get_instance', name or jail['id'])
-        if (
-            any(new[k] != jail[k] for k in ('vnet', 'nat')) or (
-                (new['vnet'] and new['vnet_default_interface'] != jail['vnet_default_interface']) or (
-                    new['nat'] and new['nat_interface'] != jail['nat_interface']
-                )
-            )
-        ) and not self.middleware.call_sync('system.is_freenas') and self.middleware.call_sync('failover.licensed'):
-            # We do this to to re-enable capabilities
-            self.enable_capabilities_for_nic(jail)
         return True
-
-    @private
-    def enable_capabilities_for_nic(self, jail_config):
-        if not self.middleware.call_sync('system.is_freenas') and self.middleware.call_sync('failover.licensed'):
-            nics = self.middleware.call_sync('interface.to_disable_evil_nic_capabilities', False)
-            if jail_config['nat']:
-                nic = self.retrieve_nat_interface(jail_config['nat_interface'])
-            else:
-                nic = self.retrieve_vnet_interface(jail_config['vnet_default_interface'])
-            iface = self.middleware.call_sync(
-                'interface.query', [['name', '=', nic], ['disable_offload_capabilities', '=', False]]
-            )
-            if iface:
-                enable = [
-                    c for c in self.middleware.call_sync('interface.nic_capabilities')
-                    if c not in nics.get(nic, []) and c not in iface[0]['state']['capabilities']
-                ]
-                if not enable:
-                    return
-                if not self.middleware.call_sync('failover.config')['disabled']:
-                    raise CallError(
-                        f'Failed to enable {",".join(enable)} capabilities for {nic} as failover is enabled',
-                        errno.EPERM
-                    )
-                self.middleware.call_sync('interface.enable_capabilities', nic, enable)
-                try:
-                    self.middleware.call_sync('failover.call_remote', 'interface.enable_capabilities', [nic, enable])
-                except Exception as e:
-                    self.middleware.logger.debug(
-                        f'Failed to enable capabilities for {nic} on remote node: {e}'
-                    )
 
     @accepts(
         Str('jail'),
@@ -898,8 +857,6 @@ class JailService(CRUDService):
 
         # TODO: Port children checking, release destroying.
         iocage.destroy_jail(force=options.get('force'))
-        if jail_config['nat'] or jail_config['vnet']:
-            self.enable_capabilities_for_nic(jail_config)
         return True
 
     @private
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -1105,7 +1105,7 @@ class JailService(CRUDService):
             )
             if nic in system_ifaces:
                 conflicts = {'LRO', 'TSO4', 'TSO6', 'VLAN_HWTSO'} if jail['nat'] else {
-                    'TXCSUM', 'TXCSUM_IPV6', 'RXCSUM', 'RXCSUM_IPV6', 'TSO4', 'TSO6'
+                    'TXCSUM', 'TXCSUM_IPV6', 'RXCSUM', 'RXCSUM_IPV6', 'TSO4', 'TSO6', 'LRO',
                 }
                 if not jail['nat']:
                     bridges = await self.middleware.call('jail.retrieve_vnet_bridge', jail['interfaces'])
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -28,7 +28,7 @@ from middlewared.service_exception import CallError, ValidationErrors
 from middlewared.utils import filter_list, run
 from middlewared.validators import IpInUse, MACAddr
 
-from collections import deque, Iterable
+from collections import defaultdict, deque, Iterable
 
 BRANCH_REGEX = re.compile(r'\d+\.\d-RELEASE')
 
@@ -1042,6 +1042,35 @@ class JailService(CRUDService):
         return self.retrieve_default_iface() if iface == 'none' else iface
 
     @private
+    async def nic_capability_checks(self, jails=None):
+        """
+        For vnet/nat based jails, when jail is started, if NIC has certain capabilities set, we experience a hiccup
+        in the network traffic which for failover can result in backup node coming online. This method returns
+        interfaces which will be affected by this based on the jails user has.
+        """
+        jail_nics = defaultdict(set)
+        system_ifaces = {i['name']: i for i in await self.middleware.call('interface.query')}
+        for jail in await self.middleware.call(
+            'jail.query', [['OR', [['vnet', '=', 1], ['nat', '=', 1]]]]
+        ) if not jails else jails:
+            nic = await self.middleware.call(
+                f'jail.retrieve_{"nat" if jail["nat"] else "vnet"}_interface',
+                jail['nat_interface' if jail['nat'] else 'vnet_default_interface']
+            )
+            if nic in system_ifaces:
+                conflicts = {'TSO4', 'TSO6', 'VLAN_HWTSO'} if jail['nat'] else {
+                    'TXCSUM', 'TXCSUM_IPV6', 'RXCSUM', 'RXCSUM_IPV6', 'TSO4', 'TSO6'
+                }
+                needs_update = set(system_ifaces[nic]['state']['capabilities']) & conflicts
+                if not jail['nat']:
+                    bridges = await self.middleware.call('jail.retrieve_vnet_bridge', jail['interfaces'])
+                    if not bridges or not bridges[0]:
+                        continue
+                if needs_update:
+                    jail_nics[nic].update(conflicts)
+        return {k: list(v) for k, v in jail_nics.items()}
+
+    @private
     def start_failover_checks(self, jail_config):
         jail_config = {
             k: ioc_common.check_truthy(v) if k in IOCJson.truthy_props else v for k, v in jail_config.items()
@@ -1054,7 +1083,7 @@ class JailService(CRUDService):
             ) and self.middleware.call_sync('failover.licensed')
         ):
             failover_enabled = not self.middleware.call_sync('failover.config')['disabled']
-            to_disable_nics = self.middleware.call_sync('interface.jail_checks', [jail_config])
+            to_disable_nics = self.middleware.call_sync('jail.nic_capability_checks', [jail_config])
             if to_disable_nics:
                 if failover_enabled:
                     raise CallError(f'Failover must be disabled before creating/updating {jail_config["id"]}')
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -1,5 +1,6 @@
 import asyncio
 import contextlib
+import errno
 import os
 import subprocess as su
 import itertools
@@ -876,7 +877,8 @@ class JailService(CRUDService):
                     return
                 if not self.middleware.call_sync('failover.config')['disabled']:
                     raise CallError(
-                        f'Failed to enable {",".join(enable)} capabilities for {nic} as failover is enabled'
+                        f'Failed to enable {",".join(enable)} capabilities for {nic} as failover is enabled',
+                        errno.EPERM
                     )
                 self.middleware.call_sync('interface.enable_capabilities', nic, enable)
                 try:
@@ -1131,7 +1133,7 @@ class JailService(CRUDService):
                 if failover_enabled:
                     raise CallError(
                         f'Failover must be disabled before creating/updating {jail_config["id"]} '
-                        f'because of the following nics: {",".join(to_disable_nics)}'
+                        f'because of the following nics: {",".join(to_disable_nics)}', errno.EPERM
                     )
                 else:
                     for nic in to_disable_nics:
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -858,10 +858,9 @@ class JailService(CRUDService):
 
     @private
     def enable_capabilities_for_nic(self, jail_config):
-        if not self.middleware.call_sync('system.is_freenas') and self.middleware.call_sync(
-            'failover.licensed'
-        ) and self.middleware.call_sync('failover.config')['disabled']:
+        if not self.middleware.call_sync('system.is_freenas') and self.middleware.call_sync('failover.licensed'):
             nics = self.middleware.call_sync('jail.nic_capability_checks', None, False)
+            nics.update(self.middleware.call_sync('vm.device.nic_capability_checks', None, False))
             if jail_config['nat']:
                 nic = self.retrieve_nat_interface(jail_config['nat_interface'])
             else:
@@ -876,6 +875,10 @@ class JailService(CRUDService):
                 ]
                 if not enable:
                     return
+                if not self.middleware.call_sync('failover.config')['disabled']:
+                    raise CallError(
+                        f'Failed to enable {",".join(enable)} capabilities for {nic} as failover is enabled'
+                    )
                 self.middleware.call_sync('interface.enable_capabilities', nic, enable)
                 try:
                     self.middleware.call_sync('failover.call_remote', 'interface.enable_capabilities', [nic, enable])
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -845,9 +845,13 @@ class JailService(CRUDService):
             iocage.rename(name)
 
         new = self.middleware.call_sync('jail._get_instance', name or jail['id'])
-        if any(new[k] != jail[k] and not new[k] for k in ('vnet', 'nat')) and not self.middleware.call_sync(
-            'system.is_freenas'
-        ) and self.middleware.call_sync('failover.licensed'):
+        if (
+            any(new[k] != jail[k] for k in ('vnet', 'nat')) or (
+                (new['vnet'] and new['vnet_default_interface'] != jail['vnet_default_interface']) or (
+                    new['nat'] and new['nat_interface'] != jail['nat_interface']
+                )
+            )
+        ) and not self.middleware.call_sync('system.is_freenas') and self.middleware.call_sync('failover.licensed'):
             # We do this to to re-enable capabilities
             self.enable_capabilities_for_nic(jail)
         return True
--- a/src/middlewared/middlewared/plugins/jail.py
+++ b/src/middlewared/middlewared/plugins/jail.py
@@ -1415,11 +1415,6 @@ class JailService(CRUDService):
             return
 
         self.check_dataset_existence()
-        if not self.middleware.call_sync('system.is_freenas') and self.middleware.call_sync('failover.licensed'):
-            self.update_defaults({'nat_backed': 'ipfw'})
-            # We start Jail/VM(s) during carp state change hook
-            return
-
         self.logger.debug('Starting jails on boot: PENDING')
         ioc.IOCage(rc=True, reset_cache=True).start(ignore_exception=True)
         self.logger.debug('Starting jails on boot: SUCCESS')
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -1706,8 +1706,76 @@ def choices(self, options):
...
+
         await self.middleware.call_hook('interface.post_sync')
 
+    @private
+    async def nic_capabilities_check(self):
+        nics = await self.jail_checks(
+            None if await self.middleware.call('failover.status') == 'MASTER' else await self.middleware.call(
+                'failover.call_remote', 'jail.query'
+            )
+        )
```
william-gr: call_remote may fail here if standby is not up, we should handle that exception

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1706,8 +1706,76 @@ def choices(self, options):
         except Exception:
             self.logger.info('Failed to sync routes', exc_info=True)
 
+        if not await self.middleware.call('system.is_freenas') and await self.middleware.call('failover.licensed'):
+            await self.nic_capabilities_check()
+
         await self.middleware.call_hook('interface.post_sync')
 
+    @private
+    async def nic_capabilities_check(self):
```
william-gr: A docstring explaining why this is all needed might com in handy if someone else needs to poke at tihs in the future

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1706,8 +1706,76 @@ def choices(self, options):
...
+        system_ifaces = {i['name']: i for i in await self.middleware.run_in_thread(self.query)}
+        conflicts = {'TXCSUM', 'TXCSUM_IPV6', 'RXCSUM', 'RXCSUM_IPV6', 'TSO4', 'TSO6'}
+        for vm_device in await self.middleware.call(
+            'vm.device.query', [
+                ['dtype', '=', 'NIC'], [
+                    'OR', [['attributes.nic_attach', '=', None], ['attributes.nic_attach', '!^', 'bridge']]
+                ]
+            ]
+        ) if not vm_devices else vm_devices:
+            nic = vm_device['attributes'].get('nic_attach') or netif.RoutingTable().default_route_ipv4.interface
```
william-gr: I believe you cant be sure there is a default route set

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1257,6 +1262,21 @@ class Config:
...
+            nics = await self.middleware.call('interface.vm_checks', [vm_device])
+            if nics:
+                if failover_enabled:
+                    raise CallError('Failover must be disabled before attempting to create/update VM NIC device')
+                else:
+                    for nic in nics:
+                        await self.middleware.call('interface.disable_capabilities', nic, nics[nic])
+                        await self.middleware.call(
+                            'failover.call_remote', 'interface.disable_capabilities', [nic, nics[nic]]
+                        )
```
william-gr: Standby node may be down, we need to handle that exception here and other places.

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -989,6 +1006,65 @@ def rc_action(self, action):
 
         return True
 
+    @accepts(Dict('props', additional_attrs=True))
+    def update_defaults(self, props):
```
william-gr: Docstring

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -989,6 +1006,65 @@ def rc_action(self, action):
...
+    def retrieve_nat_interface(self, iface):
+        return self.retrieve_default_iface() if iface == 'none' else iface
+
+    @private
+    def start_failover_checks(self, jail_config):
+        jail_config = {
+            k: ioc_common.check_truthy(v) if k in IOCJson.truthy_props else v for k, v in jail_config.items()
+        }
+        if jail_config.get('dhcp'):
+            jail_config['vnet'] = 1
```
william-gr: Should we do all that inside the if statement?

sonicaj: Well we check if vnet/nat is enabled below, so if we want to do this below, we can use another if clause inside the main one. I'll update this

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -989,6 +1006,65 @@ def rc_action(self, action):
...
+        if (
+            not self.middleware.call_sync('system.is_freenas') and (
+                jail_config['vnet'] or jail_config['nat']
+            ) and self.middleware.call_sync('failover.licensed')
+        ):
+            failover_enabled = not self.middleware.call_sync('failover.config')['disabled']
+            to_disable_nics = self.middleware.call_sync('interface.jail_checks', [jail_config])
+            if to_disable_nics:
+                if failover_enabled:
+                    raise CallError(f'Failover must be disabled before creating/updating {jail_config["id"]}')
```
themylogin: Please explain why? (`because of the following nics: {', ',join(...`)

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -1377,11 +1453,18 @@ def stop_on_shutdown(self):
...
-            pass
+    if args['id'] == 'ready' and await middleware.call('jail.iocage_set_up'):
+        if not await middleware.call_sync('system.is_freenas') and await middleware.call_sync(
+            'failover.licensed'
+        ) and await middleware.call('failover.status') == 'MASTER':
+            await middleware.call('jail.check_dataset_existence')
+            await middleware.call('jail.update_defaults', {'nat_backend': 'ipfw'})
+            # We start Jail/VM(s) during carp state change hook
+        else:
+            try:
```
themylogin: We also get here on BACKUP node, is this desired?

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1706,8 +1706,76 @@ def choices(self, options):
...
+            if nic in system_ifaces and set(system_ifaces[nic]['state']['capabilities']) & conflicts:
+                vm_nics[nic] = list(conflicts)
+        return vm_nics
+
+    @private
+    async def jail_checks(self, jails=None):
+        jail_nics = defaultdict(set)
+        system_ifaces = {i['name']: i for i in await self.middleware.run_in_thread(self.query)}
```
themylogin: Why not `self.middleware.call('interface.query')`?

src/middlewared/middlewared/plugins/network.py
themylogin: Yes, also please name all methods accordingly so they can be reused in the future. `nic_capabilities_check` - not very clear, check for what? `check_nic_capabilities_for_xxx` will be much better. Or even `disable_evil_nic_capabilities` (because you call `disable_capabilities` at the bottom).

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1706,8 +1706,76 @@ def choices(self, options):
...
+                ]
+            ]
+        ) if not vm_devices else vm_devices:
+            nic = vm_device['attributes'].get('nic_attach') or netif.RoutingTable().default_route_ipv4.interface
+            if nic in system_ifaces and set(system_ifaces[nic]['state']['capabilities']) & conflicts:
+                vm_nics[nic] = list(conflicts)
+        return vm_nics
+
+    @private
+    async def jail_checks(self, jails=None):
```
themylogin: I have a feeling this belongs to the jail plugin rather then network

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1584,6 +1604,7 @@ def translate_device(dev):
             elif not await self.disk_uniqueness_integrity_check(device, vm_instance):
                 verrors.add('attributes.path', f'{vm_instance["name"]} has "{path}" already configured')
         elif device.get('dtype') == 'NIC':
+            await self.nic_check(device)
```
themylogin: Shouldn't this be done after you run basic checks that can raise validation errors?

src/middlewared/middlewared/plugins/vm.py
themylogin: Maybe this method `interface.disable_capabilities` should propagate itself to standby node?

src/middlewared/middlewared/plugins/network.py
themylogin: Same for VM method actually.

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1706,8 +1706,76 @@ def choices(self, options):
...
+                if needs_update:
+                    jail_nics[nic].update(conflicts)
+        return {k: list(v) for k, v in jail_nics.items()}
+
+    @private
+    @accepts(Str('iface'), List('capabilities'))
+    async def disable_capabilities(self, iface, capabilities):
+        await self._get_instance(iface)
+        iface = netif.get_interface(iface)
+        iface.capabilities = {c for c in iface.capabilities if c.name not in capabilities}
```
themylogin: We should probably log what capabilities we are disabling.

src/middlewared/middlewared/plugins/vm.py
sonicaj: @themylogin there's still possibility for remote node to get out of sync with master node ( a case is remote node is down when this is called for master ). Please advise if we should still consider that ?

src/middlewared/middlewared/plugins/jail.py
sonicaj: Nope we don't. Backup node would not return true for `jail.iocage_set_up`

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -989,6 +1006,103 @@ def rc_action(self, action):
...
+        system_ifaces = {i['name']: i for i in await self.middleware.call('interface.query')}
+        for jail in await self.middleware.call(
+            'jail.query', [['OR', [['vnet', '=', 1], ['nat', '=', 1]]]]
+        ) if not jails else jails:
+            nic = await self.middleware.call(
+                f'jail.retrieve_{"nat" if jail["nat"] else "vnet"}_interface',
+                jail['nat_interface' if jail['nat'] else 'vnet_default_interface']
+            )
+            if nic in system_ifaces:
+                conflicts = {'TSO4', 'TSO6', 'VLAN_HWTSO'} if jail['nat'] else {
```
themylogin: Do we lose some features / performance when disabling this? Can these capabilities be re-enabled later after jail / VM network configuration?

src/middlewared/middlewared/plugins/jail.py
sonicaj: Because we only want to use `ipfw` firewall for TrueNAS - so we explicitly check that and update the iocage defaults. Also we don't want to start jails here - we do them in carp state change hook

src/middlewared/middlewared/plugins/jail.py
sonicaj: Yes there is a performance regression. If i got you correctly @william-gr, we don't want to re-enable those when these jails/vms have been deleted ?

src/middlewared/middlewared/plugins/jail.py
william-gr: Seems we should always use ipfw for != freenas
And just do the do not start for failover.licensed, no need MASTER check

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1706,8 +1706,49 @@ def choices(self, options):
...
+        This hiccup in case of failover makes backup node come ONLINE as master as there's a hiccup in the
+        master/backup communication. This scenario is common to vnet/nat based jails and VM's, this method checks
+        if the user has such VM's or jails which can bring forward this case and disables certain capabilities for
+        the affected NIC's so that the user is not affected by the interruption which is caused when these NIC's
+        experience a hiccup in the network traffic.
+        """
+        nics = await self.middleware.call(
+            'jail.nic_capability_checks',
+            None if await self.middleware.call('failover.status') == 'MASTER' else await self.middleware.call(
+                'failover.call_remote', 'jail.query', [['OR', [['vnet', '=', 1], ['nat', '=', 1]]]]
```
william-gr: This seems a bit redundant, regarding the call and filters. Should we move all of this inside jail.nic_capability_checks ?

sonicaj: Good idea, I'll update the PR

src/middlewared/middlewared/plugins/jail.py
william-gr: I think If Jails/VMs are deleted we should somehow enable these flags again

themylogin: @william-gr @sonicaj now when I understand what's going on maybe we should add disabling these capabilities to "interface options" and just forbid creating jails / vms using interfaces with capabilities not disabled?

themylogin: Users will have a better understanding on what's going on, why and when, and we won't need to remember to re-enable capabilities again. Explicit is better than implicit.

sonicaj: @william-gr @themylogin's suggestion makes sense to me. Do you think we should correct this with that approach ? We can store capabilities in the database so that we can configure them on system boot before we start jails/vms so they are correctly started

william-gr: Yes, I like that idea, lets do that. But I think we should also add a way to do that "automatically", warning the user what we are about to do and how, when creating a jail/vm.

Its hard to tell the user to go to another screen and do X, Y and Z before he can actually perform what he wanted in the first place.

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -827,8 +844,46 @@ def do_update(self, jail, options):
...
+        ) and not self.middleware.call_sync('system.is_freenas') and self.middleware.call_sync('failover.licensed'):
+            # We do this to to re-enable capabilities
+            self.enable_capabilities_for_nic(jail)
         return True
 
+    @private
+    def enable_capabilities_for_nic(self, jail_config):
+        if not self.middleware.call_sync('system.is_freenas') and self.middleware.call_sync(
+            'failover.licensed'
+        ) and self.middleware.call_sync('failover.config')['disabled']:
```
william-gr: Should we raise in error if HA is not disabled?

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -1377,12 +1531,18 @@ def stop_on_shutdown(self):
     """
     # We need to call a method in Jail service to make sure it runs in the
     # process pool because of py-libzfs thread safety issue with iocage and middlewared
-    if args['id'] == 'ready':
+    if args['id'] == 'ready' and await middleware.call('jail.iocage_set_up'):
+        if not await middleware.call_sync('system.is_freenas'):
+            await middleware.call('jail.check_dataset_existence')
+            await middleware.call('jail.update_defaults', {'nat_backend': 'ipfw'})
+            # We start Jail/VM(s) during carp state change hook
+            if await middleware.call_sync('failover.licensed'):
```
william-gr: call

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -1377,12 +1531,18 @@ def stop_on_shutdown(self):
     """
     # We need to call a method in Jail service to make sure it runs in the
     # process pool because of py-libzfs thread safety issue with iocage and middlewared
-    if args['id'] == 'ready':
+    if args['id'] == 'ready' and await middleware.call('jail.iocage_set_up'):
+        if not await middleware.call_sync('system.is_freenas'):
```
william-gr: call

src/middlewared/middlewared/plugins/network.py
```diff
@@ -818,6 +820,9 @@ def iface_extend(self, iface_state, configs, is_freenas, fake=False):
                         )
                 raise
 
+        if data.get('disable_offload_capabilities'):
+            await self.middleware.call('interface.enable_capabilities', name)
```
william-gr: You mean disable?

src/middlewared/middlewared/plugins/jail.py
sonicaj: I was confused initially, thinking more about it, i think it would make more sense if we did raise an error here.

src/middlewared/middlewared/plugins/jail.py
yocalebo: Can you update the docstring here?

Instead of `we experience a hiccup in the network traffic which for failover can result in backup node coming online.` Just make it `we experience a hiccup in the network traffic which can cause a failover to occur.`

src/middlewared/middlewared/plugins/network.py
```diff
@@ -818,6 +820,9 @@ def iface_extend(self, iface_state, configs, is_freenas, fake=False):
                         )
                 raise
 
+        if data.get('disable_offload_capabilities'):
+            await self.middleware.call('interface.disable_capabilities', name)
```
william-gr: What is this is false, how/where do we enable the capabilities again?

william-gr: Ah, this is the create method, nevermind me

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -989,6 +1048,103 @@ def rc_action(self, action):
...
+                jail_config['vnet'] = 1
+            if not (jail_config['vnet'] or jail_config['nat']):
+                return
+            failover_enabled = not self.middleware.call_sync('failover.config')['disabled']
+            to_disable_nics = self.middleware.call_sync('jail.nic_capability_checks', [jail_config])
+            if to_disable_nics:
+                if failover_enabled:
+                    raise CallError(
+                        f'Failover must be disabled before creating/updating {jail_config["id"]} '
+                        f'because of the following nics: {",".join(to_disable_nics)}'
```
william-gr: I think we need an explicit errno here and for VM so UI can identify it and not present a full traceback

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -827,8 +844,48 @@ def do_update(self, jail, options):
...
+                if not self.middleware.call_sync('failover.config')['disabled']:
+                    raise CallError(
+                        f'Failed to enable {",".join(enable)} capabilities for {nic} as failover is enabled'
+                    )
+                self.middleware.call_sync('interface.enable_capabilities', nic, enable)
+                try:
+                    self.middleware.call_sync('failover.call_remote', 'interface.enable_capabilities', [nic, enable])
+                except Exception as e:
+                    self.middleware.logger.debug(
+                        f'Failed to enable capabilities for {nic} on remote node: {e}'
```
yocalebo: The `remote node` was removed as accepted verbiage for TrueNAS HA. Let's use `for {nic} on standby storage controller`, please.

src/middlewared/middlewared/plugins/jail.py
```diff
@@ -989,6 +1048,103 @@ def rc_action(self, action):
...
+                else:
+                    for nic in to_disable_nics:
+                        self.middleware.call_sync('interface.disable_capabilities', nic, to_disable_nics[nic])
+                        try:
+                            self.middleware.call_sync(
+                                'failover.call_remote', 'interface.disable_capabilities', [nic, to_disable_nics[nic]]
+                            )
+                        except Exception as e:
+                            self.middleware.logger.debug(
+                                f'Failed to disable capabilities for {nic} on remote node: {e}'
```
yocalebo: The `remote node` was removed as accepted verbiage for TrueNAS HA. Let's use `for {nic} on standby storage controller`, please.

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1331,6 +1342,41 @@ def __convert_aliases_to_datastore(self, data):
...
+            if new['disable_offload_capabilities']:
+                await self.middleware.call('interface.disable_capabilities', iface['name'])
+                if failover_licensed:
+                    try:
+                        await self.middleware.call(
+                            'failover.call_remote', 'interface.disable_capabilities', [iface['name']]
+                        )
+                    except Exception as e:
+                        self.middleware.logger.debug(
+                            f'Failed to disable capabilities for {iface["name"]} on remote node: {e}'
```
yocalebo: The `remote node` was removed as accepted verbiage for TrueNAS HA. Let's use `for {nic} on standby storage controller`, please.

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1331,6 +1342,41 @@ def __convert_aliases_to_datastore(self, data):
...
+                if capabilities:
+                    await self.middleware.call('interface.enable_capabilities', iface['name'], capabilities)
+                    if failover_licensed:
+                        try:
+                            await self.middleware.call(
+                                'failover.call_remote', 'interface.enable_capabilities', [iface['name'], capabilities]
+                            )
+                        except Exception as e:
+                            self.middleware.logger.debug(
+                                f'Failed to enable capabilities for {iface["name"]} on remote node: {e}'
```
yocalebo: The `remote node` was removed as accepted verbiage for TrueNAS HA. Let's use `for {nic} on standby storage controller`, please.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1257,6 +1273,56 @@ class Config:
...
+                else:
+                    for nic in nics:
+                        await self.middleware.call('interface.disable_capabilities', nic, nics[nic])
+                        try:
+                            await self.middleware.call(
+                                'failover.call_remote', 'interface.disable_capabilities', [nic, nics[nic]]
+                            )
+                        except Exception as e:
+                            self.middleware.logger.debug(
+                                f'Failed to disable capabilities for {nic} on remote node: {e}'
```
yocalebo: The `remote node` was removed as accepted verbiage for TrueNAS HA. Let's use `for {nic} on standby storage controller`, please.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1407,6 +1474,40 @@ def vnc_bind_choices(self):
...
+                if not (await self.middleware.call('failover.config'))['disabled']:
+                    raise CallError(
+                        f'Failed to enable {",".join(enable)} capabilities for {nic} as failover is enabled'
+                    )
+                await self.middleware.call('interface.enable_capabilities', nic, enable)
+                try:
+                    await self.middleware.call('failover.call_remote', 'interface.enable_capabilities', [nic, enable])
+                except Exception as e:
+                    self.middleware.logger.debug(
+                        f'Failed to enable capabilities for {nic} on remote node: {e}'
```
yocalebo: The `remote node` was removed as accepted verbiage for TrueNAS HA. Let's use `for {nic} on standby storage controller`, please.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1257,6 +1273,56 @@ class Config:
...
+                        except Exception as e:
+                            self.middleware.logger.debug(
+                                f'Failed to disable capabilities for {nic} on remote node: {e}'
+                            )
+
+    @private
+    async def nic_capability_checks(self, vm_devices=None, check_system_iface=True):
+        """
+        For NIC devices, if VM is started and NIC is added to a bridge, if desired nic_attach NIC has certain
+        capabilities set, we experience a hiccup in the network traffic which for failover can result in backup node
```
yocalebo: To be consistent with the docstring in the jail plugin, let's change it to `we experience a hiccup in the network traffic which can cause a failover to occur.`

PR #3826, opened 2019-11-20
NAS-103902 / 12 / Ensure freenas-migrate113 does not have 12 specific migrations

themylogin:


src/middlewared/middlewared/alembic/versions/12.0/2019-11-20_12-22_vm_normalize.py
```diff
@@ -0,0 +1,115 @@
...
+                raise ValueError('Invalid sector size')
+        except (ValueError, TypeError):
+            sector_size = None
+
+        attributes.update({
+            'logical_sectorsize': sector_size,
+            'physical_sectorsize': None,
+        })
+
+        conn.execute("UPDATE vm_device SET attributes = ? WHERE id = ?", (json.dumps(attributes), device.id))
```
sonicaj: Can we access `id` with `device.id` ?

src/middlewared/middlewared/alembic/versions/12.0/2019-11-20_12-22_vm_normalize.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # normalize_mac_address
+
+    for device in conn.execute("SELECT * FROM vm_device WHERE dtype = 'NIC'").fetchall():
+        attributes = json.loads(device['attributes'])
+
+        if not attributes.get('mac') or attributes['mac'] == '00:a0:98:FF:FF:FF':
+            attributes['mac'] = None
+            conn.execute("UPDATE vm_device SET attributes = ? WHERE id = ?", (json.dumps(attributes), device.id))
```
sonicaj: Same here ?

src/middlewared/middlewared/alembic/versions/12.0/2019-11-20_12-22_vm_normalize.py
themylogin: Yes, ResultProxy has both object and dict interface, but I've fixed it to be uniform, thank you.

PR #3828, opened 2019-11-20
NAS-103247 / 12 / core.resize_shell

themylogin:


src/middlewared/middlewared/service.py
```diff
@@ -541,6 +542,17 @@ class ServicePartBase(metaclass=ServicePartBaseMeta):
 
 class CoreService(Service):
 
+    @accepts(Str('id'), Int('cols'), Int('rows'))
+    async def resize_shell(self, id, cols, rows):
```
william-gr: Should this really be a coroutine? Can the input queue put block if its full?

themylogin: It's infinite

PR #3829, opened 2019-11-20
NAS-104035 / 11.3 / Add auto-configuration for kerberized NFS in AD environments

anodos325:
Auto-register kerberos nfs spn. Configure winbindd to use default domain.

### DIFF
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -970,6 +970,18 @@ class ActiveDirectoryService(ConfigService):
             await self._net_ads_join()
             await self._register_virthostname(ad, smb, smb_ha_mode)
             if smb_ha_mode != 'LEGACY':
+                """
+                Manipulating the SPN entries must be done with elevated privileges. Add NFS service
+                principals while we have these on-hand. Once added, force a refresh of the system
+                keytab so that the NFS principal will be available for gssd.
+                """
+                must_update_trust_pw = await self._net_ads_setspn([
+                    f'nfs/{ad["netbiosname"].upper()}.{ad["domainname"]}',
+                    f'nfs/{ad["netbiosname"].upper()}'
+                ])
+                if must_update_trust_pw:
+                    await self.change_trust_account_pw()
+
                 kt_id = await self.middleware.call('kerberos.keytab.store_samba_keytab')
                 if kt_id:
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -108,6 +108,28 @@ def nfs_compress(self, nfs):
...
+            await self.middleware.call(
+                'datastore.update',
+                'directoryservice.activedirectory',
+                ad['id'],
+                {'ad_use_default_domain': True}
+            )
+            await self.middleware.call('etc.generate', 'smb')
+            await self.middleware.call('service.reload', 'cifs') 
+            if await self.middleware.call('smb.get_smb_ha_mode') == 'LEGACY':
+                verrors.add(
```
anodos325: We should probably raise a validation error immediately before reconfiguring the AD service and reloading samba.

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1223,6 +1225,20 @@ def validate_domain(self, data=None):
 
         return neterr.JOINED
 
+    @private
+    async def _net_ads_setspn(self, spn):
+        ad = await self.config()
```
anodos325: This may need to be removed will decide during testing.

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -982,6 +982,8 @@ class Config:
                     ad = await self.config()
 
             ret = neterr.JOINED
+            await self._net_ads_setspn(f'nfs/{ad["netbiosname"]}.{ad[domainname"}')
+            await self._net_ads_setspn(f'nfs/{ad["netbiosname"]}')
```
anodos325: Check return value for these and generate an alert on failure?

src/middlewared/middlewared/plugins/nfs.py
yocalebo: We should wrap the call to `smb.get_smb_ha_mode` in an `if else` statement and raise a validation error when smb is configured in LEGACY mode.

src/middlewared/middlewared/plugins/activedirectory.py
anodos325: I moved this up to the code block where we're performing keytab related tasks on non-legacy AD configurations. Change of trust pw is performed automatically, and resulting keytab uploaded to FN config.

PR #3832, opened 2019-11-21
NAS-103952 / 11.3 / Ensure objects exist before deleting them in iscsi plugin

sonicaj:


william-gr:
Not sure what I feel about this.

This will add a new extra query for every call.

sonicaj:
How about passing the retrieved object to update/delete methods ?

william-gr:
Then you want to change every single plugin to catch up with that change?

sonicaj:
To do it correctly, yes ? An easier solution would be to make sure all update/delete have at least one `_get_instance` call ( however that would lots of duplicating lines ). If you think we should go with the latter, I'll update the PR that way

william-gr:
We dont want to make such a huge change at this point. Just tackle the API call in question

william-gr:
Thanks

sonicaj:
backport

PR #3842, opened 2019-11-26
NAS-104097 / 11.3 / Validate that uploaded file is actually a keytab

anodos325:
Write out uploaded file to temporary location, validate with ktutil,
and then remove. Add new kerberos.keytab.upload method to WS API so
that users can choose to upload a keytab file, or create using a
b64encoded keytab and kerberos.keytab.create. Plumb in a legacy_validate
method into the legacy API so that we can catch keytab errors.

### DIFF
--- a/src/middlewared/middlewared/plugins/kerberos.py
+++ b/src/middlewared/middlewared/plugins/kerberos.py
@@ -2,17 +2,20 @@ import asyncio
 import base64
 import datetime
 import enum
+import io
 import os
+import shutil
 import subprocess
 import time
 from middlewared.schema import accepts, Dict, Int, List, Patch, Str
-from middlewared.service import CallError, ConfigService, CRUDService, periodic, private, ValidationErrors
+from middlewared.service import CallError, ConfigService, CRUDService, job, periodic, private, ValidationErrors
 from middlewared.utils import run, Popen
 
 
 class keytab(enum.Enum):
     SYSTEM = '/etc/krb5.keytab'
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -694,9 +721,19 @@ class Config:
         """
         verrors = ValidationErrors()
         try:
-            base64.b64decode(data['file'])
+            decoded = base64.b64decode(data['file'])
         except Exception as e:
             verrors.add("kerberos.keytab_create", f"Keytab is a not a properly base64-encoded string: [{e}]")
```
william-gr: Should we return earlier?

src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -663,6 +665,31 @@ class Config:
...
+        'keytab_data',
+        Str('name', required=True),
+    ))
+    @job(lock='upload_keytab', pipes=['input'], check_pipes=False)
+    async def upload_keytab(self, job, data):
+        """
+        Upload a keytab file. This method expects the keytab file to be uploaded using
+        the /_upload/ endpoint.
+        """
+        job.check_pipe("input")
```
william-gr: Sorry, just noticed this, I think you can use check_pipes=True in the job decorator and remove this. Right, @themylogin ?

themylogin: Yes

PR #3845, opened 2019-11-27
NAS-104103 / 11.3 / Let @throttle wrapper report throttled function name properly

themylogin:


themylogin:
@william-gr it's his threaded version for wrapping synchronous methods.

william-gr:
@themylogin sorry I dont follow. All websocket calls are gated in call_method with a Semaphore first, doesnt matter if its a coroutine or a method, the dispatcher will take care of that later in the stack call.

What am I missing?

themylogin:
@william-gr this is the other semaphore, that ensures only `max_waiters` method calls are queued up due to throttling.

william-gr:
Thats what I am saying, if we already have a semaphore in the calls do we need another one specifically for throttle? Is it worth all the extra complexity if we are still doing `seconds` wait?

Is there a reason we would want a throttle to have a different/lower semaphore count in that situation?

themylogin:
@william-gr I have not even thought about those semaphores that we already have for websocket calls. As I see it, websocket message processor and specific method decorator are located far away, and I can't think of a way to connect them that does not introduce punching through multiple abstractions (and thus, extra complexity).

william-gr:
You dont have to connect them, do you?

The way it currently is users would not be able to call the throttle function over than 10 times simultaneously anyway, per websocket connection. 

Now, if there is a global `seconds` anyway, what would be the problem in simply sleeping instead of raising the error and doing another semaphore?

themylogin:
> The way it currently is users would not be able to call the throttle function over than 10 times simultaneously anyway, per websocket connection.

I still have that DoS thing in mind, so it's still global now rather than per-connection.

william-gr:
Its global either way because of `seconds`, no? What would be actual the difference?

themylogin:
@william-gr no, it won't be global, `self.semaphores` is the same dictionary as `self.last_calls`

william-gr:
@themylogin Sorry, I am still now following, these semaphores are instantiated per throttle. How are you making them per connection?

themylogin:
@william-gr I am not making them per-connection, but if I understand you correctly your proposal is to use `Application._softhardsemaphore` which are per-connection

william-gr:
@themylogin now I am confused. You said its not global because self.semaphores.

Lets rewind and start from the beginning :)

Why do we need the semaphores per throttle? Only because one may want to have a max_waiters?

themylogin:
@william-gr yes. We don't want to queue pending method calls indefinitely in case of DoS attack.

william-gr:
Thats why I ask, why not? Each connection can queue up to 10 concurrent calls.
And only one call will be actually be made every X seconds, no?

If someone is DoS'ing with multiple connections throttle is not what would solve the problem, wouldnt you agree? But something in middlewared limiting number of max connections per source.

themylogin:
> Each connection can queue up to 10 concurrent calls.

Yes, because of `Application._softhardsemaphore`

> And only one call will be actually be made every X seconds, no?

Yes, because of original @throttle behavior

> If someone is DoS'ing with multiple connections throttle is not what would solve the problem, wouldnt you agree? 

It would not solve the 10000 incoming connections problem but it is also not trying to do it.

> But something in middlewared limiting number of max connections per source.

This is another new issue.

william-gr:
Right, that said, I dont get it, what is the benefit of local semaphores for each throttle if you agreed with everything I said?

themylogin:
@william-gr so we should limit total number of middleware connections and remove these semaphores?

william-gr:
Isnt only the sleep enough?

themylogin:
@william-gr so for now we are leaving ability to open 10000 middleware connections?

william-gr:
For this ticket i imagine so.
We can implement conn limit on 12 and think about backporting

PR #3852, opened 2019-11-27
NAS-100780 / 12 / Add events for vms

sonicaj:
This commit adds events for VM's which emit status for different vm's wrt running/shutdown etc

### DIFF
--- a/src/middlewared/middlewared/plugins/vm.py
+++ b/src/middlewared/middlewared/plugins/vm.py
@@ -755,7 +755,7 @@ class VMService(CRUDService):
     def __init__(self, *args, **kwargs):
         super(VMService, self).__init__(*args, **kwargs)
         self.vms = {}
-        self.libvirt_connection = None
+        self.libvirt_connection = self.libvirt_event_loop_thread = self.libvirt_event_loop_thread_condition = None
 
     @accepts()
     def flags(self):
@@ -1608,6 +1608,10 @@ class VMService(CRUDService):
 
     @private
     def close_libvirt_connection(self):
+        if self.libvirt_event_loop_thread:
+            self.libvirt_event_loop_thread_condition = False
+            self.libvirt_event_loop_thread.join(2.0)
+            self.libvirt_event_loop_thread = None
         if self.libvirt_connection:
             with contextlib.suppress(libvirt.libvirtError):
                 self.libvirt_connection.close()
@@ -1719,12 +1723,13 @@ class VMService(CRUDService):
                 self.middleware.logger.debug(f'Received libvirtd event with unknown domain name {dom.name()}')
 
         def event_loop_execution():
-            while True:
+            self.libvirt_event_loop_thread_condition = True
+            while self.libvirt_event_loop_thread_condition:
                 libvirt.virEventRunDefaultImpl()
 
-        event_thread = threading.Thread(target=event_loop_execution, name='libvirt_event_loop')
-        event_thread.setDaemon(True)
-        event_thread.start()
+        self.libvirt_event_loop_thread = threading.Thread(target=event_loop_execution, name='libvirt_event_loop')
+        self.libvirt_event_loop_thread.setDaemon(True)
+        self.libvirt_event_loop_thread.start()
         self.libvirt_connection.domainEventRegister(callback, None)
         self.libvirt_connection.setKeepAlive(5, 3)
 
--- a/src/middlewared/middlewared/plugins/vm.py
+++ b/src/middlewared/middlewared/plugins/vm.py
@@ -49,7 +49,7 @@ BUFSIZE = 65536
 LIBVIRT_URI = 'bhyve+unix:///system'
 LIBVIRT_AVAILABLE_SLOTS = 29  # 3 slots are being used by libvirt / bhyve
 SHUTDOWN_LOCK = asyncio.Lock()
-VM_BRIDGE = 'virbr0'
+LIBVIRT_LOCK = asyncio.Lock()
 ZFS_ARC_MAX_INITIAL = None
 
 ZVOL_CLONE_SUFFIX = '_clone'
@@ -1031,8 +1031,9 @@ class VMService(CRUDService):
         shutdown, if the VM hasn't exited after a hardware shutdown signal has been sent by the system within
         `shutdown_timeout` seconds, system initiates poweroff for the VM to stop it.
         """
-        if not self.libvirt_connection:
-            await self.wait_for_libvirtd(10)
+        async with LIBVIRT_LOCK:
+            if not self.libvirt_connection:
+                await self.wait_for_libvirtd(10)
         self.ensure_libvirt_connection()
 
         verrors = ValidationErrors()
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1634,6 +1641,68 @@ def initialize_vms(self, timeout=10):
...
+                    'vm.query', emit_type, id=int(dom.name().split('_')[0]), fields={'state': state}
+                )
+            else:
+                self.middleware.logger.debug(f'Received libvirtd event with unknown domain name {dom.name()}')
+
+        def event_loop_execution():
+            while True:
+                libvirt.virEventRunDefaultImpl()
+
+        event_thread = threading.Thread(target=event_loop_execution, name='libvirt_event_loop')
```
william-gr: Is this going to run even if there are no VMs configured?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1618,13 +1626,17 @@ def close_libvirt_connection(self):
         try:
             if not await self.middleware.call('service.started', 'libvirtd'):
                 await asyncio.wait_for(libvirtd_started(self.middleware), timeout=timeout)
+            # We want to do this before initializing libvirt connection
```
william-gr: I just noticed above, unlreated to this change: time.sleep(2), thats blocks the event loop

use await asyncio.sleep

We should also have a limit for waiting

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1634,6 +1654,69 @@ def initialize_vms(self, timeout=10):
...
+                self.middleware.send_event(
+                    'vm.query', emit_type, id=int(dom.name().split('_')[0]), fields={'state': state}
+                )
+            else:
+                self.middleware.logger.debug(f'Received libvirtd event with unknown domain name {dom.name()}')
+
+        def event_loop_execution():
+            self.libvirt_event_loop_thread_condition = True
+            while self.libvirt_event_loop_thread_condition:
+                libvirt.virEventRunDefaultImpl()
```
william-gr: What this method does? Does it block? At what point does it return?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1634,6 +1654,69 @@ def initialize_vms(self, timeout=10):
...
+                    else:
+                        # We undefine/define domain numerous times based on if vm has any new changes
+                        # registered, this is going to reflect that
+                        state = 'UPDATING CONFIGURATION'
+                else:
+                    state = vm_state_mapping.get(dom.state()[0], 'UNKNOWN')
+            except libvirt.libvirtError:
+                state = 'UNKNOWN'
+
+            if dom.name().split('_')[0].isdigit():
```
william-gr: We are doing this name() split twice. Just a nitpick but we should use a var instead

src/middlewared/middlewared/plugins/vm.py
sonicaj: It runs one iteration of the event loop. It will block indefinitely if we don't register any event.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1634,6 +1654,69 @@ def initialize_vms(self, timeout=10):
...
+                self.middleware.send_event(
+                    'vm.query', emit_type, id=int(dom.name().split('_')[0]), fields={'state': state}
+                )
+            else:
+                self.middleware.logger.debug(f'Received libvirtd event with unknown domain name {dom.name()}')
```
william-gr: We should use %s for logging, not f-string

src/middlewared/middlewared/plugins/vm.py
william-gr: You mean every time one event is delivered from libvirtd this will return?

sonicaj: Yes

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1594,6 +1606,10 @@ def status(self, id):
 
     @private
     def close_libvirt_connection(self):
+        if self.libvirt_event_loop_thread:
+            self.libvirt_event_loop_thread_condition = False
+            self.libvirt_event_loop_thread.join(2.0)
```
william-gr: I am not sure why you need to join() then? Wont the thread only end after you close the connection in the next lines, because it will likely be blocked within the loop?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1634,6 +1654,69 @@ def initialize_vms(self, timeout=10):
                 # Whatever happens, we don't want middlewared not booting
                 self.vms[vm_data['name']] = VMSupervisor(vm_data, self.libvirt_connection, self.middleware)
 
+    @private
+    def setup_libvirt_events(self):
```
themylogin: I am continuing to push my idea of making our files smaller (or at least not increasing their size). Please put this to separate file (for existing modules we do it like (`vm_/events.py`)

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1594,6 +1606,10 @@ def status(self, id):
 
     @private
     def close_libvirt_connection(self):
+        if self.libvirt_event_loop_thread:
```
themylogin: Related to my comment  below about `setup_libvirt_events`: if you refactor this out to `teardown_libvirt_events`, you'll be able to hide these three variables into events service part.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1284,13 +1292,16 @@ def do_delete(self, id, data):
...
-        vm = self.middleware.call_sync('vm._get_instance', id)
         if vm['name'] in self.vms:
             self.vms.pop(vm['name']).undefine_domain()
         else:
             VMSupervisor(vm, self.libvirt_connection, self.middleware).undefine_domain()
 
-        return self.middleware.call_sync('datastore.delete', 'vm.vm', id)
+        result = self.middleware.call_sync('datastore.delete', 'vm.vm', id)
+        if not self.middleware.call_sync('vm.query'):
+            self.close_libvirt_connection()
```
themylogin: Do we stop libvirtd when every VM is stopped?

sonicaj: No, that would result in failure to managing them correctly. It's only closed if there are no more VM's left and the last one was deleted.

themylogin: I've meant when all VMs are stopped :)

1. Boot system with no VMs -> libvirtd is not running
2. Create VM - middleware will start libvirtd
3. Stop and delete VM - libvirtd will continue running

Is it like this?

sonicaj: Yes it is. For step 3, we will close the hypervisor connection we have but libvirtd will continue running separately.

themylogin: Why?

sonicaj: Why we still let libvirtd run in the background or why we close the connection ?

themylogin: Why we leave libvirtd running?

sonicaj: We shouldn't, it provides little benefit when the vm is started again for opening a connection to the hypervisor but that's a fine tradeoff. Thank you for pointing it out

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1259,39 +1262,53 @@ def rename_domain(self, old, new):
...
-        elif status.get('state') == 'ERROR':
-            raise CallError('Unable to retrieve VM status. Failed to destroy VM')
-
-        if data['zvols']:
-            devices = self.middleware.call_sync('vm.device.query', [
-                ('vm', '=', id), ('dtype', '=', 'DISK')
-            ])
+        async with LIBVIRT_LOCK:
+            vm = await self._get_instance(id)
+            self.ensure_libvirt_connection()
```
william-gr: Is it safe to run ensure_libvirt_connection synchronously in the event loop?

sonicaj: I think so but maybe i can still separate this to be certain we don't get into any issues.

sonicaj: Actually no, we are checking connection.isalive - definitely updating this

src/middlewared/middlewared/plugins/vm_/events.py
```diff
@@ -0,0 +1,72 @@
+import libvirt
+import threading
+
+from middlewared.service import private, Service
+
+
+class VMService(Service):
+
+    @private
+    def libvirt_connection(self):
```
themylogin: What is the point of this method?

src/middlewared/middlewared/plugins/vm_/events.py
```diff
@@ -0,0 +1,72 @@
...
+
+            vm_id = dom.name().split('_')[0]
+            if vm_id.isdigit():
+                self.middleware.send_event('vm.query', emit_type, id=int(vm_id), fields={'state': state})
+            else:
+                self.middleware.logger.debug('Received libvirtd event with unknown domain name %s', dom.name())
+
+        def event_loop_execution():
+            while self.libvirt_connection():
+                libvirt.virEventRunDefaultImpl()
```
themylogin: Race condition can occur if `close_libvirt_connection` and then `wait_for_libvirtd` are called during `libvirt.virEventRunDefaultImpl` - this loop will never receive null value of `libvirt_connection`, will never terminate and you'll get two `event_thread` running.

SImple solution is not to use `self.libvirt_connection()` in `setup_libvirt_events`. Just pass connection on which you want to react to events on to `setup_libvirt_events` and run this loop while `connection.isAlive()`.

src/middlewared/middlewared/service.py
```diff
@@ -297,7 +297,7 @@ def __init__(self, middleware, parts):
 
         for part in self.parts:
             for name in dir(part):
-                if name.startswith('_'):
+                if name.startswith('__'):
```
themylogin: We won't allow exposed method names to start with `_`, why?

sonicaj: `_get_instance` is a good example here. Compound services can't use it

src/middlewared/middlewared/plugins/vm_/events.py
sonicaj: This looks unlikely, however we would still end up with 2 threads if it does happen as whatever happens `wait_for_libvirtd` is going to create another thread and it should as the event handler should be configured with the new connection and not the old one.

sonicaj: Actually we also stop libvirt after closing the hypervisor connection and then start it again and then open the connection and start a new thread. It looks highly unlikely that this loop will not yield till then, it yields as soon as the connection is closed. Please advise if you think we should still look after this case. We can perhaps synchronise this in a different manner then

src/middlewared/middlewared/service.py
themylogin: Then we need a public alias for _get_instance that will eventually replace it

src/middlewared/middlewared/plugins/vm_/events.py
themylogin: You won't need to synchronize anything if you let the event reading thread die with the connection.

sonicaj: I just learned something new about objects and how python manages their references. Thank you for pointing this out : )

sonicaj: @themylogin when testing this, i see that when we close the connection `isAlive` call fails to correctly determine that the connection is no longer active and results in a traceback which is okay ( would have been better if it correctly told us boolean false ), however we don't want this polluting our logs and we can't ignore it as we would like to log where this happens for a valid/good reason. Suggestions on tackling this ? I have been looking at libvirt source code to see if we could have another flag which might tell us the connection is closed/open but doesn't seem like there is one.

sonicaj: And i found a workaround, no worries :)

src/middlewared/middlewared/service.py
themylogin: We still need to revert this

PR #3856, opened 2019-11-28
NAS-104134 / 11.3 / Don't offer to restart disabled and stopped services on pool unlock

themylogin:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1514,23 +1514,62 @@ def __find_disk_from_topology(self, label, pool):
...
+
+        result = {}
+
+        for k, v in services.items():
+            service = await self.middleware.call('service.query', [['service', '=', k]], {'get': True})
+            if service['enable'] or service['state'] == 'RUNNING':
+                result[k] = v
+
+        try:
+            activated_pool = await self.middleware.call('jail.get_activated_pool')
```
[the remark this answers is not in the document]
sonicaj: If that pool already has jails, it would already be activated and we wouldn't run into this issue, so i guess we can perhaps update this usage to checking if activated pool equals to the current pool name ?

sonicaj:
@themylogin i am not sure actually now - when UI retrieves this list, that pool wouldn't have been unlocked, so we can't determine for jails the activated iocage pool, no ?

themylogin:
@sonicaj please give me the method that returns iocage pool name even if it is locked now

sonicaj:
@themylogin in 12 we have ZOL encryption where this can be done. Here we can't access the pool's content while it's locked. I can think of ways of keeping track of iocage but not sure if we want that for 11.3, plus it wouldn't be useful for this case in 12. @william-gr do we want to do that for 12 ?

themylogin:
@sonicaj I've brought what's been there back, please confirm that there is nothing better we can do in 11.3

sonicaj:
@themylogin i think this is actually the right thing to do. Apologies for all the confusion. If it's not activated, we won't activate it even if we start jails with it

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1514,23 +1514,62 @@ def __find_disk_from_topology(self, label, pool):
...
+        except Exception:
+            activated_pool = None
+        # If iocage is not activated yet, there is a chance that this pool might have it activated there
```
sonicaj: We could perhaps remove this comment though :)

themylogin: For me it looked like quite illogical thing, that's why I've put this comment, why do you want to remove it?

themylogin:
@sonicaj what do you mean, what changes?

src/middlewared/middlewared/plugins/pool.py
sonicaj: Well that call will now never activate iocage nor if we start these jails with `rc_action`, so this comment isn't reflecting reality

src/middlewared/middlewared/service.py
```diff
@@ -69,37 +70,88 @@ class throttle(object):
     on some key (possibly argument of the method).
     """
 
-    def __init__(self, seconds=0, condition=None, exc_class=RuntimeError):
+    def __init__(self, seconds=0, condition=None, exc_class=RuntimeError, max_waiters=10):
```
sonicaj: These changes @themylogin, are they related ?

themylogin: Oh, God, thank you, I have no idea how did it get there

src/middlewared/middlewared/plugins/pool.py
themylogin: Hm, now, if I unlock pool that had iocage activated on it, it won't bring jails back?

sonicaj: So when we import the pool, and if iocage wasn't already active and we execute `rc_action`, there are 2 cases:
1) Pool had jails, in which case the rc_action call should proceed as desired because the `iocage_set_up` function which checks if iocage is activated or not only checks if we have pool and dataset with iocage root dataset on it. So they should succeed.
2) Pool didn't had jails, the `iocage_set_up` call would return false and `rc_action` won't do anything with the pool remaining deactivated

themylogin: So, based on #1, after pool import iocage will become activated?

sonicaj: Yes, that's how it should/will work. I can test this on my end to be certain as well - but if it's not, that's a bug and we have to fix it.

themylogin: So this comment does reflect reality? :)

sonicaj: Also this change makes sense from another perspective, if a user was using iocage on an encrypted pool and then locked it. Now he is using it on a different pool, we shouldn't offer him jail restart option as it will conflict on which pool to use with both showing up as active ( some thing for me to look at to make sure we don't error out if someone manually does this )

sonicaj: You're right 😅

PR #3859, opened 2019-11-28
NAS-104159 / 11.3 / Donot activate iocage while querying if there is an active iocage pool

sonicaj:


src/middlewared/middlewared/plugins/jail.py
```diff
@@ -1501,7 +1503,7 @@ class JailFSAttachmentDelegate(FSAttachmentDelegate):
         except Exception:
             pass
         else:
-            if query_dataset.startswith(os.path.join(activated_pool, 'iocage')):
+            if query_dataset.startswith(os.path.join(activated_pool or '', 'iocage')):
```
themylogin: This won't work correctly:
```
>>> os.path.join('', 'iocage')
'iocage'
```

sonicaj: Yes, it won't. Thank you for catching that

PR #3863, opened 2019-11-28
NAS-103926 / 12 / Introduce pool.attach for extending STRIPE/MIRROR vdevs

sonicaj:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -531,6 +531,95 @@ def pool_extend(self, pool):
...
+                break
+            for v in topology[i]:
+                if v['guid'] == options['target_vdev']:
+                    topology_type = i
+                    vdev = v
+                    break
+        else:
+            verrors.add('pool_attach.target_vdev', 'Unable to locate VDEV')
+            verrors.check()
+        if topology_type in ('cache', 'spares'):
```
themylogin: What about other invalid types? (slog) Can we just rephrase it: `if topology_type == 'data' then process else raise` ?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -531,6 +531,95 @@ def pool_extend(self, pool):
...
+        try:
+            await job.wrap(extend_job)
+        except CallError as e:
+            if pool['encrypt'] > 0:
+                try:
+                    # If replace has failed lets detach geli to not keep disk busy
+                    await self.middleware.call('disk.geli_detach_single', devname)
+                except Exception:
+                    self.logger.warn(f'Failed to geli detach {devname}', exc_info=True)
+            raise e
```
themylogin: Just `raise` so we don't lose original stack trace

src/middlewared/middlewared/plugins/pool.py
sonicaj: There shouldn't be any issue extending log topology ? We already allow it to be mirrored/striped, so users can add redundancy if they want to ?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -531,6 +531,95 @@ def pool_extend(self, pool):
...
+            for v in topology[i]:
+                if v['guid'] == options['target_vdev']:
+                    topology_type = i
+                    vdev = v
+                    break
+        else:
+            verrors.add('pool_attach.target_vdev', 'Unable to locate VDEV')
+            verrors.check()
+        if topology_type in ('cache', 'spares'):
+            verrors.add('pool_attach.target_vdev', f'Attaching disks to {topology_type} not allowed.')
```
william-gr: I dont recall if attaching to cache is really not allowed?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -531,6 +531,95 @@ def pool_extend(self, pool):
...
+        pool = await self._get_instance(oid)
+        verrors = ValidationErrors()
+        if not pool['is_decrypted']:
+            verrors.add('oid', 'Pool must be unlocked for this action.')
+            verrors.check()
+        topology = pool['topology']
+        topology_type = vdev = None
+        for i in topology:
+            if topology_type:
+                break
```
william-gr: I dont understand how topology_type could ever be true here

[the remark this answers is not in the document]
sonicaj: It will be when we set it in the nested loop. However this will get missed in the last iteration. I should move this after the nested loop. ( last key would be spare, we don't support it but still the error would be wrong )

src/middlewared/middlewared/plugins/pool.py
sonicaj: Not allowed by ZFS and us as well in create

william-gr: Are you completely sure you cannot turn a single cache device into a mirror?

sonicaj:
```
root@freenas[~]# zpool attach check 15975562645825302336 /dev/da5
cannot attach /dev/da5 to 15975562645825302336: device is in use as a cache
```

sonicaj:
```
	if (avail_spare)
		return (zfs_error(hdl, EZFS_ISSPARE, msg));

	if (l2cache)
		return (zfs_error(hdl, EZFS_ISL2CACHE, msg));
```
In `zpool_vdev_attach`

PR #3866, opened 2019-11-29
NAS-104108 / 12 / Platform-specific etc files scaffolding

themylogin:


src/middlewared/middlewared/plugins/etc.py
```diff
@@ -94,7 +101,7 @@ class EtcService(Service):
             {'type': 'mako', 'path': 'dhclient.conf'},
         ],
         'nfsd': [
-            {'type': 'py', 'path': 'nfsd'},
+            {'type': 'py', 'path': 'nfsd', 'platform': 'Linux'},
```
william-gr: Did you mean FreeBSD?

PR #3867, opened 2019-11-29
Merge master from 191129

william-gr:


william-gr:
@sonicaj i am most interested in a review of the vm plugin, it diverged a bit

PR #3876, opened 2019-11-30
NAS-104165 / 12 / Platform specific system.info implementation

sonicaj:


src/middlewared/middlewared/utils/__init__.py
```diff
@@ -331,6 +331,10 @@ def sw_version_is_stable():
     return version['stable']
 
 
+def cpu_model():
```
themylogin: What for?

src/middlewared/middlewared/utils/osc/linux/system.py
```diff
@@ -0,0 +1,19 @@
...
+logger = logging.getLogger(__name__)
+
+__all__ = ['cpu_model']
+
+
+RE_CPU_MODEL = re.compile(r'Model name:\s*(.*)')
+
+
+def cpu_model():
+    cp = subprocess.Popen('lscpu', stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
themylogin: You can just read from `/proc/cpuinfo`

src/middlewared/middlewared/utils/__init__.py
```diff
@@ -331,6 +331,10 @@ def sw_version_is_stable():
     return version['stable']
 
 
+def get_cpu_model():
```
themylogin: I am still not convinced this is necessary

sonicaj: I wasn't sure about it myself. Should we use `osc` directly in system plugin ?

themylogin: Yes, we import from osc directly

src/middlewared/middlewared/plugins/system.py
```diff
@@ -7,6 +7,7 @@
 from middlewared.service import CallError, ConfigService, no_auth_required, job, private, Service, ValidationErrors
 import middlewared.sqlalchemy as sa
 from middlewared.utils import Popen, run, start_daemon_thread, sw_buildtime, sw_version
+from middlewared.utils import osc
```
william-gr: Combine with line above please

sonicaj: oops, sorry

PR #3878, opened 2019-12-02
NAS-102806 / 12 / Add support for userspace quotas to middleware

anodos325:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4184,149 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+        Ref('query-options'),
+    )
+    @item_method
+    async def get_groupquota(self, id, filters, options):
+        """
+        Returns a list of group quotas on the ZFS dataset `id`. Supports
+        `query-filters` and `query-options`.
+        """
+        quota_list = []
+        quota_list = []
```
yocalebo: `quota_list` defined twice?

anodos325: Yup. Good catch.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4184,148 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+        """
+        units = {
+            'TB': 1099511627776,
+            'GB': 1073741824,
+            'MB': 1048576,
+            'KB': 1024,
+            'B': 1
+        }
+        cmd = ['zfs', 'set']
+        for q in data:
```
anodos325: Probably need to add validation for length of list. May want to set upper-bound on how many quotas to set in a single command.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4184,148 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+            if m is None:
+                self.logger.debug('Invalid userspace quota: %s', quota)
+                continue
+
+            entry.update(m.groupdict())
+
+            try:
+                entry['username'] = (
+                    await self.middleware.call('user.get_user_obj', {'uid': entry['uid']})
+                )['pw_name']
```
william-gr: Seems we could cache this value in a dict, so we dont do it multiple times for the same uid?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4184,148 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+
+    @accepts(
+        Str('id', required=True),
+        List('quotas', items=[
+            Dict(
+                'userspace_quota',
+                Str('type', enum=['USER', 'GROUP'], required=True),
+                Str('id', required=True),
+                Int('quota', required=True),
+                Str('unit', enum=['B', 'KB', 'MB', 'GB', 'TB'], default='B')
```
william-gr: For the API I think it usually makes sense to always ask/use bytes, and let the caller (UI) handle about unit?

anodos325: Yeah, that makes sense. Keep it simpler.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4184,137 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+        `query-filters` and `query-options`.
+        """
+        quota_list = []
+        userquota = await run(
+            ['zfs', 'userspace', '-H', '-n', '-p', '-o', 'name,used,quota', id],
+            check=False
+        )
+        if userquota.returncode != 0:
+            raise CallError('Failed to get userquota for %s: [%s]',
+                            id, userquota.stderr.decode())
```
william-gr: CallError does not accept args like that, need to use f-string

src/middlewared/middlewared/plugins/pool.py
sonicaj: I agree

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4184,137 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+        return filter_list(quota_list, filters, options)
+
+    @accepts(
+        Str('id', required=True),
+        List('quotas', items=[
+            Dict(
+                'userspace_quota',
+                Str('type', enum=['USER', 'GROUP'], required=True),
+                Str('id', required=True),
+                Str('quota', required=True),
```
sonicaj: We should use Int schema here which will validate that `quota` indeed is an integer

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4184,137 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+
+        return filter_list(quota_list, filters, options)
+
+    @accepts(
+        Str('id', required=True),
+        List('quotas', items=[
+            Dict(
+                'userspace_quota',
+                Str('type', enum=['USER', 'GROUP'], required=True),
+                Str('id', required=True),
```
sonicaj: Do we want to validate this ?

william-gr: We should probably add a _get_instance call to make sure it exists

anodos325: I will add some validation for the (uid/gid/name), and take William's suggestion to also validate the dataset.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4182,131 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+    @accepts(
+        Str('id', required=True),
+        Str('type', enum=['USER', 'GROUP']),
+        Ref('query-filters'),
+        Ref('query-options'),
+    )
+    @item_method
+    async def get_userquota(self, id, type, filters, options):
+        """
+        Returns a list of the specified `type` of  quotas on the ZFS dataset `id`.
```
wblock: Use present tense (s/Returns/Return/).  It's a small thing in this one, but makes this sort of thing easier to read.  More instances below.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4182,131 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+        Supports `query-filters` and `query-options`. used_bytes and used_percentage
+        may not instantly update as space is used.
+
+        Each quota entry has the following fields:
```
wblock: Avoid "the following" when items are immediately following.
s/the following/these/

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4182,131 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+
+        `id` - the uid or gid to which the quota applies.
+
+        `name` - the user or group name to which the quota applies. Value will be
```
wblock: Present tense:
s/will be/is/
s/could not be/cannot be/

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4182,131 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+        """
+        Set a user or group quota on the specified ZFS dataset. The quota limits
+        the amount of disk space consumed by files that are owned by the specified
+        users or groups.
+
+        `id` the name of the target ZFS dataset.
+
+        `quotas` specifies a list of `quota_entry` entries to apply to dataset.
+
+        `quota_entry` entries have the following required parameters:
```
wblock: s/the following/these/

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4182,131 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+
+        `type`: specifies whether the quota applies to a user or a group.
+
+        `id`: the uid, gid, or name to which the quota applies.
+
+        `quota`: the quota size in bytes. Setting a value of `0` will remove
```
wblock: Present tense:
s/will remove/removes/

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4182,131 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+        the user or group quota.
+        """
+        MAX_QUOTAS = 100
+        if len(data) > MAX_QUOTAS:
+            raise ValidationError(
+                'pool.dataset.set_userspace_quota.',
+                f'The number of userspace quotas that may be set in single API call is limited to {MAX_QUOTAS}.'
```
wblock: s/may/can/

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4182,147 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+        `quota` - the quota size in bytes.
+
+        `used_bytes` - the amount of bytes the user has written to the dataset.
+
+        `used_percentage` - the percentage of the user or group quota consumed.
+        """
+        dataset = (await self._get_instance(id))['name']
+        quota_list = []
+        quota_get = await run(
+            ['zfs', f'{type.lower()}space', '-H', '-n', '-p', '-o', 'name,used,quota', dataset],
```
sonicaj: We should be making `zfs` specific calls in zfs plugin please. I forgot about it earlier.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4182,147 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
         return data
 
+    @accepts(
+        Str('id', required=True),
+        Str('type', enum=['USER', 'GROUP']),
+        Ref('query-filters'),
+        Ref('query-options'),
+    )
+    @item_method
+    async def get_userquota(self, id, type, filters, options):
```
sonicaj: Should we perhaps rename this to `get_quota` as it can retrieve quotas for groups as well

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4182,147 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
             raise CallError(pjob.error)
         return data
 
+    @accepts(
+        Str('id', required=True),
+        Str('type', enum=['USER', 'GROUP']),
```
sonicaj: Also perhaps we could rename it to something diffferent then just `type` as `type` variable is predefined in python

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4182,147 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+                Int('quota', required=True),
+            )
+        ], default=[{
+            'type': 'USER',
+            'id': '0',
+            'quota': 0
+        }])
+    )
+    @item_method
+    async def set_userquota(self, id, data):
```
sonicaj: We should probably rename this as well ?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4182,147 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+        if len(data) > MAX_QUOTAS:
+            verrors.add(
+                'pool.dataset.set_userquota.',
+                f'The number of userspace quotas that can be set in single API call is limited to {MAX_QUOTAS}.'
+            )
+
+        cmd = ['zfs', 'set']
+        for q in data:
+            type = q["type"].lower()
+            if not q["id"].isdigit():
```
sonicaj: Should we perhaps verify digits as well if they exist or not ?

src/middlewared/middlewared/plugins/pool.py
william-gr: I dont see a problem naming it type, is there an issue with it?

src/middlewared/middlewared/plugins/pool.py
anodos325: I'm someone unsure of what to do here. `get_quota` might be confusing because we have dataset quotas as well as user and group quotas that apply to datasets. In one iteration I had `get_user_or_group_quota` if you think that's less confusing.

yocalebo: I like `get_zfs_user_or_group_quota`.

src/middlewared/middlewared/plugins/pool.py
sonicaj: Technically no, just a suggestion :p

william-gr: I think we should just rename the python variable to type_ and keep "type" for the API name

sonicaj: Agreed :)

src/middlewared/middlewared/plugins/pool.py
anodos325: Went with maximal length

src/middlewared/middlewared/plugins/pool.py
anodos325: Changed to quota_type

src/middlewared/middlewared/plugins/pool.py
anodos325: Done

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4182,112 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+
+        `quota`: the quota size in bytes. Setting a value of `0` removes
+        the user or group quota.
+        """
+        MAX_QUOTAS = 100
+        dataset = (await self._get_instance(id))['name']
+        verrors = ValidationErrors()
+        if len(data) > MAX_QUOTAS:
+            verrors.add(
+                'pool.dataset.set_userquota.',
```
william-gr: First argument of verrors is the schema name, in this case just `quotas`

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4182,112 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+                f'The number of userspace quotas that can be set in single API call is limited to {MAX_QUOTAS}.'
+            )
+
+        quota_list = []
+        for q in data:
```
william-gr: For proper verror index do something like: for i, q in enumerate(data):

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4182,6 +4182,112 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+                        f'{quota_type} {q["id"]} is not valid.'
+                    )
+            else:
+                id_type = 'uid' if quota_type == 'user' else 'gid'
+                try:
+                    await self.middleware.call(f'{quota_type}.get_{quota_type}_obj',
+                                               {id_type: q["id"]})
+                except Exception:
+                    verrors.add(
+                        f'pool.dataset.set_userquota',
```
william-gr: quotas.{i}.id

PR #3880, opened 2019-12-03
NAS-104188 / 12 / Correct path for ups driver choices

sonicaj:


src/middlewared/middlewared/plugins/ups.py
```diff
@@ -81,7 +81,7 @@ def driver_choices(self):
         Returns choices of UPS drivers supported by the system.
         """
         ups_choices = {}
-        driver_list = '/conf/base/usr__local__etc/nut/driver.list'
+        driver_list = '/conf/base/etc/local/nut/driver.list'
```
william-gr: Well, how about we fix it for Linux while we are at it? It is somewhat trivial?

sonicaj: I agree

william-gr:
@themylogin i will let you review this one ; )

PR #3888, opened 2019-12-04
NAS-104204 / 12 / Add support more more quota types

anodos325:
Rename get_user_or_group_quota and set_user_or_group_quota to simply get_quota and set_quota. Add support for USEROBJQUOTA, GROUPOBJQUOTA, and DATASET quotas. Lay foundation for project quotas as well, but don't expose yet. DATASET quotas are a wrapper around existing code related to dataset quota and refquota.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -4223,10 +4223,6 @@ class PoolDatasetService(CRUDService):
         )
         return filter_list(quota_list, filters, options)
 
-    @accepts()
-    async def quota_type_choices(self):
-        return ['DATASET', 'USER', 'USEROBJ', 'GROUP', 'GROUPOBJ']
-
     @accepts(
         Str('ds', required=True),
         List('quotas', items=[
@@ -4300,7 +4296,7 @@ class PoolDatasetService(CRUDService):
                     )
                     continue
 
-                if dataset_quotas.get(q['id'].lower(), None) is not None:
+                if dataset_quotas.get(q['id'].lower()) is not None:
                     verrors.add(
                         f'quotas.{i}.id',
                         f'Setting multiple values for {q["id"]} for quota_type "DATASET" is not permitted'
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -4183,19 +4183,19 @@ class PoolDatasetService(CRUDService):
         return data
 
     @accepts(
-        Str('id', required=True),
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4206,52 +4206,80 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
-        dataset = (await self._get_instance(id))['name']
+        dataset = (await self._get_instance(ds))['name']
         quota_list = await self.middleware.call(
-            'zfs.dataset.get_user_or_group_quota', dataset, quota_type.lower()
+            'zfs.dataset.get_quota', dataset, quota_type.lower()
         )
         return filter_list(quota_list, filters, options)
 
+    @accepts()
+    async def quota_type_choices(self):
```
william-gr: We usually only have choices for things that are not in the enum. If you think we should keep that we will need a docstring

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4260,33 +4288,69 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+            if q["quota_type"] == 'DATASET':
+                if q['id'] not in ['QUOTA', 'REFQUOTA']:
                     verrors.add(
                         f'quotas.{i}.id',
-                        f'{quota_type} {q["id"]} is not valid.'
+                        f'id for quota_type DATASET must be either "QUOTA" or "REFQUOTA"'
                     )
+                    continue
+
+                if dataset_quotas.get(q['id'].lower(), None) is not None:
```
william-gr: Second argument of .get (None) is a no-op. None is the default, JFYI

william-gr:
@sonicaj please review as well

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4260,33 +4288,69 @@ def __handle_zfs_set_property_error(self, e, properties_definitions):
...
+                    )
+                    continue
+
+                dataset_quotas.update({
+                    q['id'].lower(): q['quota_value']
+                })
+
+                continue
+
+            if q["quota_type"] not in ['PROJECT', 'PROJECTOBJ']:
```
sonicaj: Can `quota_type` ever be `PROJECT/PROJECTOBJ` ? I think not ?

anodos325: At the time I was writing this, I was considering future expansion of available quota types to include these as well. Project quotas will probably require a separate design doc and ticket, but I thought it's not a bad idea to at least keep some of the code here for future use. That said, I can remove if it's a problem.

PR #3890, opened 2019-12-04
NAS-104216 / 11.2 / ensure 'progressfile' isn't empty

yocalebo:
NAS-104216

`* Active Controller - Unable to run alert source 'Replication'
File "/usr/local/lib/python3.6/site-packages/middlewared/plugins/alert.py", line 469, in __run_source
alerts = (await alert_source.check()) or []
.....`
`
File "/usr/local/lib/python3.6/site-packages/middlewared/plugins/replication.py", line 55, in replication_extend
pid = int(f.read())
ValueError: invalid literal for int() with base 10: ''`

### DIFF
--- a/src/middlewared/middlewared/plugins/replication.py
+++ b/src/middlewared/middlewared/plugins/replication.py
@@ -52,7 +52,6 @@ class ReplicationService(CRUDService):
         progressfile = f'/tmp/.repl_progress_{data["id"]}'
         if os.path.exists(progressfile):
             with open(progressfile, 'r') as f:
-                f.seek(0)
                 pid = f.read().strip()
 
             if len(pid) > 0:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/replication.py
```diff
@@ -52,14 +52,18 @@ class Config:
...
             with open(progressfile, 'r') as f:
-                pid = int(f.read())
-            title = await self.middleware.call('notifier.get_proc_title', pid)
-            if title:
-                reg = re.search(r'sending (\S+) \((\d+)%', title)
-                if reg:
-                    data['status'] = f'Sending {reg.groups()[0]}s {reg.groups()[1]}s'
-                else:
-                    data['status'] = 'Sending'
+                f.seek(0)
```
themylogin: What do you intend to do?

yocalebo: @themylogin sorry, do you mean with the `f.seek(0)` call? I put that in there to ensure we start at the beginning of the file before reading it's contents. Customer received a traceback in the alert system because the replication progress file was empty.

themylogin: But files are always opened at the beginning?

yocalebo: @themylogin just a habit I have I guess. If f.read() is called twice in the same context, it will return EOF. The traceback showed an empty string and I couldn't definitively say if it was because it was empty. I'll go ahead and remove it.

PR #3891, opened 2019-12-05
NAS-104132 / 12 / Make disk.sync* work on linux

sonicaj:


src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -0,0 +1,105 @@
...
+
+    async def get_serials(self):
+        raise NotImplementedError()
+
+    def get_disks(self):
+        disks = {}
+        lshw_disks = self.retrieve_lshw_disks_data()
+
+        for block_device in filter(
+            lambda b: b.name not in ('sr0',),
```
themylogin: Why not `startswith('sr')` like in FreeBSD?

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -0,0 +1,105 @@
...
+                for c in child.getchildren():
+                    if not len(c.getchildren()):
+                        data[c.tag] = c.text
+                    elif c.tag == 'capabilities':
+                        for capability in filter(lambda d: d.text.endswith('rotations per minute'), c.getchildren()):
+                            data['rotationrate'] = capability.get('id')[:-3]
+                lshw_disks[data['logicalname']] = data
+        return lshw_disks
+
+    def get_disk(self, name):
```
themylogin: `@private`

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -0,0 +1,105 @@
...
+
+        return self.get_disk_details(block_device, disk, self.retrieve_lshw_disks_data())
+
+    @private
+    def get_disk_details(self, block_device, disk, lshw_disks):
+        dev_data = block_device.__getstate__()
+        disk.update({
+            'name': dev_data['name'],
+            'sectorsize': dev_data['io_limits']['logical_sector_size'],
+            'number': ord(dev_data['name'][-1].lower()) - 96,
```
themylogin: We can have more than 26 disks. They are named like that:
```
#1	sda
#2	sdb
...
#26	sdz
#27	sdaa
#28	sdab
...
#52	sdaz
#53	sdba
#54	sdbb
...
```
Also please do not hardcode magic numbers, use `ord('a')`

themylogin: https://github.com/torvalds/linux/blob/master/drivers/scsi/sd.c#L3249 proof

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -0,0 +1,105 @@
...
+                disk['lunid'] = cp_stdout.strip().split()[0].decode()
+                if disk['lunid'].startswith('0x'):
+                    disk['lunid'] = disk['lunid'][2:]
+
+        if disk['serial'] and disk['lunid']:
+            disk['serial_lunid'] = f'{disk["serial"]}_{disk["lunid"]}'
+
+        return disk
+
+    async def get_valid_zfs_partition_type_uuids(self):
```
themylogin: Why don't we make this shared between OSes? Are there any obstacles to import Linux-created pool in FreeBSD?

src/middlewared/middlewared/plugins/device_/device_info_linux.py
sonicaj: Base class has `@private`

src/middlewared/middlewared/plugins/device_/device_info_linux.py
sonicaj: Good point. Ideally there shouldn't be

src/middlewared/middlewared/plugins/device.py
```diff
@@ -22,44 +17,20 @@ class DeviceService(Service):
 
         Currently only SERIAL is supported.
```
william-gr: While we are here, lets update this text please

src/middlewared/middlewared/plugins/device_/device_info_base.py
```diff
@@ -0,0 +1,39 @@
+import re
+
+from middlewared.service import private, ServicePartBase
+
+RE_SERIAL_NUMBER = re.compile(r'Serial Number:\s+(?P<serial>.+)', re.I)
+
+
+class DeviceInfoBase(ServicePartBase):
+
+    RE_SERIAL_NUMBER = RE_SERIAL_NUMBER
```
william-gr: I dont quite understand why declare again in the class instead of just define it here to begin with?

src/middlewared/middlewared/plugins/device_/device_info_freebsd.py
```diff
@@ -0,0 +1,103 @@
...
+            return disks
+        for g in klass.geoms:
+            # Skip cd*
+            if g.name.startswith('cd'):
+                continue
+            disks[g.name] = await self.get_disk_details(self.disk_default.copy(), g)
+        return disks
+
+    @private
+    def retrieve_disk_geom_klass(self):
```
william-gr: I dont quite understand why we have to call it klass instead of class, its the full method name and not just the "class" keyword. But just pointing that out, no need to change ;)

src/middlewared/middlewared/plugins/device_/device_info_freebsd.py
```diff
@@ -0,0 +1,103 @@
...
+            'name': disk_geom.name,
+            'mediasize': disk_geom.provider.mediasize,
+            'sectorsize': disk_geom.provider.sectorsize,
+            'stripesize': disk_geom.provider.stripesize,
+        })
+        if disk_geom.provider.config:
+            disk.update({k: v for k, v in disk_geom.provider.config.items() if k not in ('fwheads', 'fwsectors')})
+            if disk['rotationrate'] is not None:
+                disk['rotationrate'] = int(disk['rotationrate']) if disk['rotationrate'].isdigit() else None
+            if disk['descr'] is not None and not disk['descr']:
```
william-gr: Seems redundant, isnt just `if not disk['descr']` the same thing?

PR #3894, opened 2019-12-05
NAS-104164 / 12 / interface for Linux

themylogin:


src/middlewared/middlewared/plugins/interface/lagg.py
```diff
@@ -0,0 +1,63 @@
+from middlewared.service import Service
```
william-gr: Prety minor but can we rename it to `lag` or linkaggregation?

FreeBSD is the only operating system I found to call that kind of interface `lagg`

src/middlewared/middlewared/plugins/interface/netif_linux/interface.py
```diff
@@ -0,0 +1,131 @@
...
+from .lagg import LaggMixin
+from .utils import bitmask_to_set, run
+from .vlan import VlanMixin
+
+logger = logging.getLogger(__name__)
+
+__all__ = ["Interface"]
+
+CLONED_PREFIXES = [
+    'lo', 'tun', 'tap', 'br', 'epair', 'carp', 'vlan', 'bond', 'pflog', 'pfsync',
```
william-gr: No pflog/pfsync/epair in Linux, I imagine?

src/middlewared/middlewared/plugins/interface/lagg.py
themylogin: It's named lagg in database and API, is it really worth it to rename?

william-gr: Where do we have lagg in the API? I made sure we used LINK_AGGREGATION instead

themylogin: Ok, I am wrong, but it's still lagg in the database, rename that too?

william-gr: I dont think we have to go as far as renaming the database

src/middlewared/middlewared/plugins/interface/capabilities_freebsd.py
```diff
@@ -0,0 +1,43 @@
...
+                iface.capabilities = current | {capability}
+            except OSError:
+                pass
+            else:
+                enabled.append(capability.name)
+        if enabled:
+            self.middleware.logger.debug(f'Enabled {",".join(enabled)} capabilities for {iface}')
+
+    def disable_capabilities(self, iface, capabilities):
+        self.middleware.call_sync('interface._get_instance', iface)
```
sonicaj: Let's please use `.get_instance` in favour of `._get_instance` as private methods can cause issues  with service parts

PR #3895, opened 2019-12-05
Correctly retrieve disk identifier with gpt partition uuids

sonicaj:


src/middlewared/middlewared/plugins/disk.py
```diff
@@ -804,7 +804,7 @@ def decrypt(self, job, devices, passphrase=None):
         if klass:
             for g in klass.geoms:
                 for p in g.providers:
-                    if p.name == name:
+                    if p.name.rsplit('p', 1)[0] == name:
```
william-gr: Wouldnt it be g.name in the outside loop?

PR #3902, opened 2019-12-09
NAS-104251 / 12 / Perform cleanup when the installer loops

freqlabs:
Make sure we can use the install menu option more than once.

JIRA: NAS-104251


freqlabs:
It looked odd to have Cancel for no password when a few lines later there is a check that  prevents one from leaving the password fields blank, saying "Empty password is not secure."
But I think I had the logic wrong anyway, so I've reverted that change. Now it's just the cleanup fix.

PR #3912, opened 2019-12-10
Clean up idmap plugin

anodos325:
Add comments. Add validation for domain. Fix cascading on delete on idmap.domain entry deletion.


src/middlewared/middlewared/alembic/versions/2019-12-10_19-44_fix_idamp_foreign_keys.py
```diff
@@ -0,0 +1,40 @@
...
+revision = '17fe2353a0de'
+down_revision = 'dcf5c178714b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('directoryservice_idmap_ad', schema=None) as batch_op:
+        batch_op.create_foreign_key(batch_op.f('fk_directoryservice_idmap_ad_idmap_ad_domain_id_directoryservice_idmap_domain'), 'directoryservice_idmap_domain', ['idmap_ad_domain_id'], ['idmap_domain_name'], ondelete='CASCADE')
```
themylogin: This will crash update if for some reason invalid values are present in altered column. Please see `2019-09-27_08-20_fix_lacking_foreign_keys` how we prevent that.

anodos325: Great. This is why I flagged you for review :)

PR #3915, opened 2019-12-11
Freenas/12 devel merge 20191211

william-gr:


william-gr:
@themylogin @anodos325 not looking for detailed review, but just a quick glance if everything looks how it should because I had a few conflicts.

PR #3917, opened 2019-12-11
NAS-104269 / 11.2 / raise custom exception and log it

yocalebo:
We have a customer where this ValueError is being raised in `get_reserved_portfd` so all `hardware = True` related alerts are being sent to proactive support because the customer has a gold contract. This single system has opened up > 250 support tickets with us.

All 250+ tickets are being generated because of that custom ValueError. I change it so that we raise a custom exception and simply log it.

### DIFF
--- a/src/middlewared/middlewared/client/client.py
+++ b/src/middlewared/middlewared/client/client.py
@@ -18,9 +18,7 @@ import sys
 import threading
 import time
 import uuid
-import logging
 
-log = logging.getLogger('middlewared.client')
 
 class Event(TEvent):
 
@@ -165,7 +163,6 @@ class WSClient(WebSocketClient):
             else:
                 break
         if fd < 0:
-            log.debug('Failed to reserve a privileged port')
             raise ReserveFDException()
         return fd
 
--- a/src/middlewared/middlewared/client/client.py
+++ b/src/middlewared/middlewared/client/client.py
@@ -18,7 +18,9 @@ import sys
 import threading
 import time
 import uuid
+import logging
 
+log = logging.getLogger('middlewared.client')
 
 class Event(TEvent):
 
@@ -46,6 +48,10 @@ class Event(TEvent):
 
 
 CALL_TIMEOUT = int(os.environ.get('CALL_TIMEOUT', 60))
+
+
+class ReserveFDException(Exception):
+    pass
 
 
 class WSClient(WebSocketClient):
@@ -159,7 +165,8 @@ class WSClient(WebSocketClient):
             else:
                 break
         if fd < 0:
-            raise ValueError('Failed to reserv a privileged port')
+            log.debug('Failed to reserve a privileged port')
+            raise ReserveFDException()
         return fd
 
     def connect(self):
--- a/src/middlewared/middlewared/plugins/alert.py
+++ b/src/middlewared/middlewared/plugins/alert.py
@@ -401,6 +401,7 @@ class AlertService(Service):
                         else:
                             raise
                 except ReserveFDException:
+                    self.logger.debug('Failed to reserve a privileged port')
                     pass
 
                 except Exception:
--- a/src/middlewared/middlewared/plugins/alert.py
+++ b/src/middlewared/middlewared/plugins/alert.py
@@ -402,7 +402,6 @@ class AlertService(Service):
                             raise
                 except ReserveFDException:
                     self.logger.debug('Failed to reserve a privileged port')
-                    pass
 
                 except Exception:
                     alerts_b = [
--- a/src/middlewared/middlewared/plugins/alert.py
+++ b/src/middlewared/middlewared/plugins/alert.py
@@ -20,6 +20,7 @@ from middlewared.alert.base import (
     ProThreadedAlertService
 )
 from middlewared.alert.base import UnavailableException, AlertService as _AlertService
+from middlewared.client.client import ReserveFDException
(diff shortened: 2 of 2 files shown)


src/middlewared/middlewared/client/client.py
```diff
@@ -47,6 +49,20 @@ def wait(self, timeout=None):
 
 CALL_TIMEOUT = int(os.environ.get('CALL_TIMEOUT', 60))
 
+class ReservFDError(Exception):
```
themylogin: Reserv**e**FDError?

src/middlewared/middlewared/client/client.py
```diff
@@ -47,6 +49,20 @@ def wait(self, timeout=None):
...
+    """
+    This class is used as a custom exception raised in `get_reserved_portfd`.
```
themylogin: Too verbose comment that involves talking about unrelated subsystems (alerts). The best place to explain alert-related part is in the alert plugin code, where you catch this exception and skip it.

src/middlewared/middlewared/client/client.py
```diff
@@ -110,6 +126,8 @@ def __init__(self, url, *args, **kwargs):
                             sock.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 0)
                         except (AttributeError, socket.error):
                             pass
+                except ReservFDError:
+                    log.debug('Failed to reserv a privileged port')
```
themylogin: I think we should not handle this exception here, client should not be initialized if we failed to reserve port. This code will just crash client a few lines below in `WebSocket.__init__(self, sock, ...)` with `NameError: sock`

src/middlewared/middlewared/client/client.py
```diff
@@ -159,7 +177,7 @@ def get_reserved_portfd(self, blacklist=None):
             else:
                 break
         if fd < 0:
-            raise ValueError('Failed to reserv a privileged port')
+            raise ReservFDError
```
themylogin: `raise ReservFDError()`

src/middlewared/middlewared/client/client.py
themylogin: Alert plugin already handles some exceptions in a special way in `__run_source` method, you can just add this new exception there.

src/middlewared/middlewared/client/client.py
yocalebo: Kept the name inline with the message that was raised (i.e `Failed to reserv privileged port`). I'll correct the spelling.

src/middlewared/middlewared/client/client.py
```diff
@@ -159,7 +165,8 @@ def get_reserved_portfd(self, blacklist=None):
             else:
                 break
         if fd < 0:
-            raise ValueError('Failed to reserv a privileged port')
+            log.debug('Failed to reserve a privileged port')
```
themylogin: We don't need logging here anymore

yocalebo: Does the alert system log this somewhere? I think we would still like to know when this is failing somehow, right?

themylogin: We raise exception, it's enough. Caller will log if necessary..

src/middlewared/middlewared/plugins/alert.py
```diff
@@ -399,6 +400,9 @@ def restore(self, id):
                             pass
                         else:
                             raise
+                except ReserveFDException:
```
william-gr: How are we going to know the alert failed to run on remote side if we just pass here?

yocalebo: We won't, that's why I had a debug log message. I'd like to add that back at the caller side if you agree.

william-gr: Definitely is a strong word. It could be a one-off.

Why you dont want to raise on hardware alerts? So a support ticket is not created? I dont see a harm in only logging it. Whats the benefit of raising?

themylogin: > Whats the benefit of raising?
We'll get "running alert on passive node failed due to ReserveFDException" for all non-hardware alert sources if we raise. Otherwise it would only be a line in log file.

yocalebo: Support is fine with just a log entry.

william-gr: Yeah, not sure its any useful to raise an alert with that error?

yocalebo: @william-gr I don't believe it is. So do you want me to push PR to log message in alert plugin?

william-gr: Please do, unless @themylogin feels strongly about the alert instead? ; )

themylogin: If support is fine with log, I am also fine :)

src/middlewared/middlewared/plugins/alert.py
```diff
@@ -399,6 +400,10 @@ def restore(self, id):
...
+                    self.logger.debug('Failed to reserve a privileged port')
+                    pass
```
william-gr: I will be picky here and say `pass` has become a no-op/useless

yocalebo: :)

PR #3930, opened 2019-12-13
Ability to retrieve datasets in flat/hierarchical structure

sonicaj:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3516,6 +3516,14 @@ def _retrieve_keys_from_file(self, job):
     def query(self, filters=None, options=None):
         """
         Query Pool Datasets with `query-filters` and `query-options`.
+
+        We provide 2 ways to how dataset's data is retrieved. First is a flat structure ( default ), which
```
wblock:
```suggestion
        We provide two ways to retrieve a dataset. The first is a flat structure (default), where
```

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3516,6 +3516,14 @@ def _retrieve_keys_from_file(self, job):
...
+        means that all the datasets in the system are returned as separate objects which also contain all the data
```
wblock:
```suggestion
        all datasets in the system are returned as separate objects which contain all data
```

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3516,6 +3516,14 @@ def _retrieve_keys_from_file(self, job):
...
+        their is for their children. This retrieval type is slightly slower because of duplicates which exist in
```
wblock:
```suggestion
        there is for their children. This retrieval type is slightly slower because of duplicates in
```

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3516,6 +3516,14 @@ def _retrieve_keys_from_file(self, job):
...
+        each object.
+        Second type is hierarchical where only top level datasets are returned in the list and they contain all the
```
wblock:
```suggestion
        The second type is hierarchical, where only top level datasets are returned in the list. They contain all the
```

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3516,6 +3516,14 @@ def _retrieve_keys_from_file(self, job):
...
+        children there are for them in `children` key. This retrieval type is slightly faster.
```
wblock:
```suggestion
        children in the `children` key. This retrieval type is slightly faster.
```

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3516,6 +3516,14 @@ def _retrieve_keys_from_file(self, job):
...
+        These options are controlled by `query-options.extra.flat` attribute which defaults to true.
```
wblock:
```suggestion
        These options are controlled by the `query-options.extra.flat` attribute (default true).
```

PR #3931, opened 2019-12-13
NAS-104328 / 12 / Clean up idmap plugin

anodos325:
Add comments. Add validation for domain. Fix cascading on delete
of idmap.domain entry.


src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -603,7 +636,18 @@ class Config:
     )
     async def do_create(self, data):
         """
-        Create an entry in the idmap backend table.
+        `domain` specifies the domain for which the idmap backend is being created. Numeric id, short-form
+        domain name, or long-form DNS domain name of the domain may be specified. Entry must be entered as
+        it appears in `idmap.domain`.
+
+        `range_low` and `rang_high` specify the UID and GID range for which this backend is authoritative.
```
william-gr: rang_high typo

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -128,6 +128,32 @@ class Config:
...
 
+    @private
+    async def _validate_domain_info(self, domain, verrors):
+        """
+        Only domains that have been configured as idmap domains
+        are permitted.
+        """
+        configured_domains = await self.middleware.call('idmap.domain.query')
+
+        id = domain.get('id', None)
```
william-gr: JFYI, .get('foo') == .get('foo', None)

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -603,7 +636,18 @@ class Config:
...
+        it appears in `idmap.domain`.
+
+        `range_low` and `rang_high` specify the UID and GID range for which this backend is authoritative.
+
+        `rangesize` defines the number of uids/gids available per domain range.
+
+        `readonly` turn the module into read-only mode. No new ranges will be allocated nor will new mappings
+        be created in the idamp pool.
+
+        `ignore builtin` ignore any mapping requests for the BUILTIN domain.
```
william-gr: ignore_builtin

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -603,7 +636,18 @@ class Config:
...
-        Create an entry in the idmap backend table.
+        `domain` specifies the domain for which the idmap backend is being created. Numeric id, short-form
+        domain name, or long-form DNS domain name of the domain may be specified. Entry must be entered as
+        it appears in `idmap.domain`.
+
+        `range_low` and `rang_high` specify the UID and GID range for which this backend is authoritative.
+
+        `rangesize` defines the number of uids/gids available per domain range.
+
+        `readonly` turn the module into read-only mode. No new ranges will be allocated nor will new mappings
```
wblock:
```suggestion
        `readonly` sets the module to read-only mode. No new ranges will be allocated. New mappings
```

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -603,7 +636,18 @@ class Config:
...
+        be created in the idamp pool.
```
wblock:
```suggestion
        will not be created in the idmap pool.
```

src/middlewared/middlewared/plugins/idmap.py
wblock:
```suggestion
        `ignore builtin` ignores mapping requests for the BUILTIN domain.
```

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -787,7 +847,13 @@ class Config:
     )
     async def do_create(self, data):
         """
-        Create an entry in the idmap backend table.
+        `domain` specifies the domain for which the idmap backend is being created. Numeric id, short-form
```
wblock: This repeated identical comment should be factored out into a single comment at the start of the section, then referred back to from these multiple entries.

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -1190,7 +1274,15 @@ class Config:
     )
     async def do_create(self, data):
         """
-        Create an entry in the idmap backend table.
+        The idmap_tdb plugin is the default backend used by winbindd for storing SID/uid/gid mapping tables.
+        In contrast to read only backends like idmap_rid, it is an allocating backend. This means that it
```
wblock:
```suggestion
        In contrast to read-only backends like idmap_rid, it is an allocating backend. This means that it
```

PR #3932, opened 2019-12-15
NAS-104200 / 12 / Platform dependent boot plugin

sonicaj:


src/middlewared/middlewared/plugins/boot_/boot_loader.py
```diff
@@ -0,0 +1,48 @@
...
+import platform
+import shutil
+import tempfile
+
+from middlewared.service import private, Service
+from middlewared.utils import run
+
+IS_LINUX = platform.system().lower() == 'linux'
+
+
```
william-gr: In this case seems to make more sense to have one file for linux and for freebsd? There is no shared code at all, just big `if` blocks.

sonicaj: I am not sure i follow ? Are you saying we should have 2 files because there is no shared code ?

william-gr: yes

src/middlewared/middlewared/plugins/boot_/boot_loader_freebsd.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class BootService(Service, BootLoaderBase):
+
+    async def install_loader(self, dev):
+        if (await self.middleware.call('boot.get_boot_type')) == 'EFI':
+            with tempfile.TemporaryDirectory() as tmpdirname:
+                await run('mount', '-t', 'msdosfs', f'/dev/{dev}p1', tmpdirname, check=False)
+                try:
+                    os.makedirs(f'{tmpdirname}/efi/boot')
```
william-gr: I think there is a exist_ok=True arg to that method

src/middlewared/middlewared/plugins/boot_/format.py
```diff
@@ -0,0 +1,122 @@
...
+    async def format(self, dev, options):
+        """
+        Format a given disk `dev` using the appropriate partition layout
+        """
+        job = await self.middleware.call('disk.wipe', dev, 'QUICK')
+        await job.wait()
+        if job.error:
+            raise CallError(job.error)
+
+        disk_details = await self.middleware.call('device.get_disk', dev)
```
william-gr: I just noticed that device.get_disk returns a dict if the disk is not found?

Isn't that wrong/weird? Shouldnt it return None?

src/middlewared/middlewared/plugins/boot_/format.py
```diff
@@ -0,0 +1,122 @@
...
+            raise CallError(f'Details for {disk_details["name"]} not found.')
+
+        swap_size = options.get('swap_size')
+        commands = []
+        partitions = []
+        efi_boot = (await self.middleware.call('boot.get_boot_type')) == 'EFI'
+        if not IS_LINUX:
+            commands.append(('gpart', 'create', '-s', 'gpt', '-f', 'active', f'/dev/{dev}'))
+            partitions.append(
+                ('efi' if efi_boot else 'freebsd-boot', 272629760 if efi_boot else 524288),
```
william-gr: Can you put in a comment explaining what these magic numbers are, please?

src/middlewared/middlewared/plugins/boot_/format.py
```diff
@@ -0,0 +1,122 @@
...
+            )
+            if options.get('swap_size'):
+                partitions.append(('freebsd-swap', options['swap_size']))
+            if options.get('size'):
+                partitions.append(('freebsd-zfs', options['size']))
+        else:
+            partitions.extend([
+                ('BIOS boot partition', 1048576),
```
william-gr: Same here

src/middlewared/middlewared/plugins/boot_/format.py
```diff
@@ -0,0 +1,122 @@
...
+                ['gpart', 'add', '-t', 'freebsd-zfs', '-i', '2', '-a', '4k'] + (
+                    ['-s', str(options['size']) + 'B'] if options.get('size') else []
+                ) + [dev]
+            )
+
+        for command in commands:
+            p = await run(*command, check=False)
+            if p.returncode != 0:
+                raise CallError(
+                    '%r failed:\n%s%s' % (' '.join(command), p.stdout.decode('utf-8'), p.stderr.decode('utf-8'))
```
william-gr: Lets not use old formatting style, please. Lets prefer .format instead.

src/middlewared/middlewared/plugins/device_/device_info_base.py
```diff
@@ -43,3 +44,11 @@ class DeviceInfoBase(ServicePartBase):
     @private
     async def get_disk(self, name):
         raise NotImplementedError()
+
+    @private
+    async def get_zfs_part_type(self):
```
william-gr: Arent these two more suited for the disk plugin?

src/middlewared/middlewared/plugins/boot_/boot_loader_freebsd.py
sonicaj: Yeah there is and i actually had it in my mind i'll be covering it. Forgot :( Thank you for catching it :)

PR #3934, opened 2019-12-16
NAS-104316 / 11.3 / Fix newfs_msdos: /dev/da0p1: Operation not permitted

themylogin:


Tsaukpaetra:
I've recently experienced this myself; actually it happened specifically when installation failed due to an IO timeout, I rebooted and tried to install again and the terminal would forevermore say Operation not permitted. It wasn't apparently mounted anywhere, and nothing would resolve issue except smashing the partition table with another OS....

william-gr:
@themylogin  should we close this?

themylogin:
@william-gr we were not able to understand who is keeping partition open. sleep helps, but IIRC sleep is not acceptable solution. Also IIRC, it only happens on 11.2, not 11.3

PR #3945, opened 2019-12-19
NAS-104348 / 11.3 / ZfsDeadmanAlertClass

themylogin:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -38,6 +39,17 @@
 ZPOOL_KILLCACHE = '/data/zfs/killcache'
 
 
+class ZfsDeadmanAlertClass(AlertClass, SimpleOneShotAlertClass):
+    category = AlertCategory.SYSTEM
+    level = AlertLevel.WARNING
+    title = "Device Is Causing Slow I/O on Pool"
+    text = "Device %(vdev)s is causing slow I/O on pool %(pool)s."
+
+    expires_after = timedelta(hours=4)
```
william-gr: For the sake of keeping changes minimal for RELEASE, do we really need this new feature to expire?
For this specifically if deadman happened, wouldnt we want to keep the alert until the user acknowledge/dismiss it?

themylogin: Alexander said it might be too pushy. However, we can probably backport this to 12 and for 11.3 just have `delete_automatically = False`

william-gr: I think thats the safest thing to do, especially because we wont have a RC2 now.

themylogin:
backport

src/middlewared/middlewared/plugins/pool.py
themylogin: @william-gr done

PR #3953, opened 2019-12-20
fix(middlewared/network): carp pass with sqlalchemy

william-gr:


william-gr:
In django ORM we automatically generated int_pass on model save.

This is the password used for the CARP IP.

william-gr:
I can put in a comment if you think its gonna be useful?

themylogin:
@william-gr thank you for explanation, now I get it. I did not knew that CARP has pass and I was sure I've ported all field defaults so to me that's came from nowhere.

PR #3964, opened 2019-12-26
NAS-104109 / 12 / middlewared systemd service

themylogin:


src/middlewared/debian/middlewared.service
```diff
@@ -0,0 +1,18 @@
...
+
+Before=network-pre.target
+Wants=network-pre.target
+
+Wants=systemd-modules-load.service local-fs.target
+After=systemd-modules-load.service local-fs.target
+
+[Service]
+Type=notify
+ExecStart=/usr/bin/middlewared --log-handler=file
```
william-gr: Does it work if we add the truenas overlay directory?

themylogin: I think if I fix invalid imports, we'll be fine

PR #3969, opened 2019-12-26
fix(middlewared/filesystem): increase file_receive size limit

william-gr:
This is used in failover plugin to transfer database.

### DIFF
--- a/src/middlewared/middlewared/plugins/filesystem.py
+++ b/src/middlewared/middlewared/plugins/filesystem.py
@@ -206,7 +206,7 @@ class FilesystemService(Service):
     @private
     @accepts(
         Str('path'),
-        Str('content', max_length=512000),
+        Str('content', max_length=2048000),
         Dict(
             'options',
             Bool('append', default=False),


src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -206,7 +206,7 @@ def stat(self, path):
     @private
     @accepts(
         Str('path'),
-        Str('content', max_length=512000),
+        Str('content', max_length=2048000),
```
themylogin: We do transfers in chunks, any sane limit will work, what is the reason for this?

william-gr: Because thats the limit on the method size. We dont want a client sending a payload of 500MB in a single method call message.

themylogin: I mean what is the reason of increasing this limit?

william-gr: I want to make the transfer in a single call. Splitting it in 3-4 chunks will make it slower

PR #3973, opened 2019-12-26
NAS-104197 / 12 / account.sync_builtin

themylogin:


src/middlewared/middlewared/plugins/account.py
```diff
@@ -1082,3 +1082,7 @@ class Config:
                     f'{schema}.users',
                     f'Following users do not exist: {", ".join(map(str, notfound))}',
                 )
+
+
+async def setup(middleware):
+    await middleware.call('user.sync_builtin')
```
william-gr: We are running this everytime middlewared starts.

Shouldnt we do something to run it only once? After install/upgrade?

PR #3979, opened 2019-12-27
Move CARP script to middlewared

william-gr:


william-gr:
@themylogin this is a raw conversion from the devd carp script to within middlewared. No actual functional changes. There are tons of things that could be and will be improved but for this first pass I am simply making it work, without changing anything, not even stupid things like `run("touch {FILE}")`.

That said, please review only functionality and things that may be utterly wrong.

william-gr:
@themylogin ping

PR #3983, opened 2019-12-30
NAS-102517 / 12 / Introduce "managedby" field to track ZFS dataset users

sonicaj:


src/middlewared/middlewared/restful.py
```diff
@@ -532,7 +541,9 @@ def get_post_args(self, req, resp, **kwargs):
         if method.get('item_method') is True:
             method_args.insert(0, kwargs['id'])
 
-        method_kwargs = {}
+        method_kwargs = {
+            'app': Application(req.headers.get('X-Real-Remote-Addr'), req.headers.get('X-Real-Remote-Port'))
+        } if method.get('pass_application') else {}
```
themylogin: Can we please use regular syntax for things like this? Condition check happens first and then conditioned code is executed, and most people are used to read it this way.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3775,6 +3777,17 @@ def transform(dataset):
...
+        if not data.get('managedby') and app:
+            uri = None
+            if app.rest and app.host:
+                uri = app.host
+            elif app.websocket and app.request.headers.get('X-Real-Remote-Addr'):
+                uri = app.request.headers.get('X-Real-Remote-Addr')
+            if uri and uri not in [
+                '::1', '127.0.0.1', *[d['address'] for d in await self.middleware.call('interface.ip_in_use')]
+            ]:
+                data['managedby'] = uri
```
themylogin: Reading ticket description, I see that this field will be used to determine if this dataset has been created by TrueCommand. I can't understand how will we distinguish TrueCommand from Web UI in this case, will we match `managedby` with IPv4|IPv6 regex? Seems awkward.

themylogin:
@william-gr I don't see why we need to do the same for update (the original ticket says about creating datasets for external's system purposes, I wonder which ones) and it's completely illogical to me to set some property on deleted dataset :)
Also the original ticket talks about setting the same property on shares, I don't see this implemented here, or have we decided not to do it.

william-gr:
@themylogin I am forward thinking, doing it for update and delete will make pass_app simply work when we have a use case for these. I just dont see why we would have to only do it for create, making it an unicorn.

Yes, we are only handling datasets here.

src/middlewared/middlewared/plugins/pool.py
william-gr: Not just TrueCommand, any API users, so they can identify themselves via name. IP should be a fallback in case no string is specified.

themylogin:
@william-gr I don't understand where unicorn is. This code looks smooth to me, no far-fetched special cases, or what did you mean?.

william-gr:
I mean by only do_create behaving as expected with regards to pass_app, not do_update or do_delete

themylogin:
It's arguable whether `managed_by` should be changeable with `update`. IMO for current purpose the better name is `created_by`

How do you expect `do_delete` to behave with regards to this new field? o_O

william-gr:
Well, but I just told you I dont expect it to be used in this case. I just want the middleware to be consistent, and behave the same way for do_update and do_delete :). I am not asking to make manage_by work with these.

themylogin:
@william-gr another issue: what will UI print? `"This dataset was created by {managed_by}"`? If we put a name/intent there, we'll need to distinguish name and intent because they require different prepositions to form a phrase. I also don't see any need for putting creator IP there. If we really want it, maybe we can put it to different field to solve preposition problem?

william-gr:
`This dataset was created by 'TrueCommand'`

`This dataset was created by '10.20.20.4'`

They kind of read the same to me? It will always be a name, intent can be put into "comments". Not sure we need to complicate ourselves further?
Although I can see some value of having both.

themylogin:
Ok, I understand you about intent. But still having `TrueCommand@10.20.20.4` can be more informative IMO.

PR #3984, opened 2019-12-30
NAS-104537 / 11.2 / fix race condition with sync_encrypted and volume_import methods

yocalebo:
@william-gr 
before this commit, I could reproduce UNIQUE constraint violation errors ~50% of the time. After this change, I couldn't get it to happen again and the database was updated accordingly via the devd event calling the `sync_encrypted` method.

### DIFF
--- a/gui/middleware/notifier.py
+++ b/gui/middleware/notifier.py
@@ -2038,7 +2038,7 @@ class notifier(metaclass=HookMetaclass):
         self.__rmdir_mountpoint(vol_mountpath)
 
     def volume_import(self, volume_name, volume_id, key=None, passphrase=None, enc_disks=None):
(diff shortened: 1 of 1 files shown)


william-gr:
@yocalebo so if an exception occurs, whatever it is, we will have dangling entries in the database, despite the import failing?

yocalebo:
@william-gr I don't believe so because the devd event shouldn't be triggered if we fail to import the zpool which means the entries should never be written.

william-gr:
@yocalebo You removed code to delete all objects. Which means we will have a att least dangling Volume entry if the pool dont import. It will be seen from the UI pools list.

yocalebo:
@william-gr Thanks, fixed it.

PR #3989, opened 2019-12-31
NAS-104552 / 11.2 / only run jail.query if system is licensed for jails

yocalebo:
Jira: https://jira.ixsystems.com/browse/NAS-104552

### DIFF
--- a/gui/storage/models.py
+++ b/gui/storage/models.py
@@ -250,8 +250,7 @@ class Volume(Model):
         jails_licensed = False
         with client as c:
             jails_licensed = c.call('system.feature_enabled', 'JAILS')
-        if jails_licensed:
-            with client as c:
+            if jails_licensed:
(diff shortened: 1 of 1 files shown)


gui/storage/models.py
```diff
@@ -247,9 +247,13 @@ def has_attachments(self):
             for vm_attached in vms_attached:
                 attachments['vm'].append(vm_attached.get('device_id'))
 
+        jails_licensed = False
         with client as c:
-            attachments['jails_ng'] = [
-                ij['host_hostuuid'] for ij in c.call('jail.query')]
+            jails_licensed = c.call('system.feature_enabled', 'JAILS')
+        if jails_licensed:
```
william-gr: Can we put this inside the first with statement so we dont have to open two connections in case jails is enabled?

yocalebo: Sure!

PR #3991, opened 2020-01-02
NAS-104331 / 12 / Platform dependent Pool plugin

sonicaj:


src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -121,6 +134,27 @@ def get_disk_details(self, block_device, disk, lshw_disks):
...
                 disk['blocks'] = int(disk['size'] / disk['sectorsize'])
 
+        if not disk['size'] and os.path.exists(os.path.join(disk_sys_path, 'size')):
+            with open(os.path.join(disk_sys_path, 'size'), 'r') as f:
+                disk['blocks'] = int(f.read().strip())
+            disk['size'] = disk['mediasize'] = disk['blocks'] * disk['sectorsize']
+
+        if not disk['serial']:
+            serial_cp = subprocess.Popen(
+                ['sg_vpd', '--quiet', '--page=0x80', block_device.path],
```
william-gr: Have you verified if this work with SATA? First glance makes me think its only for SCSI/SAS

src/middlewared/middlewared/plugins/pool_/attach_disk.py
```diff
@@ -0,0 +1,97 @@
...
+            'pool_attach',
+            Str('target_vdev', reuired=True),
+            Str('new_disk', reuired=True),
+            Str('passphrase'),
+        )
+    )
+    @job(lock=lambda args: f'pool_attach_{args[0]}')
+    async def attach(self, job, oid, options):
+        """
+        For FreeBSD platform, if the `oid` pool is passphrase GELI encrypted, `passphrase`
```
william-gr: Lets just say "TrueNAS Core/Enterprise" instead, please

src/middlewared/middlewared/plugins/pool_/import_disk_freebsd.py
```diff
@@ -0,0 +1,270 @@
+import asyncio
```
william-gr: How about the linux version for this? You will tackle this later?

src/middlewared/middlewared/plugins/pool_/replace_disk.py
```diff
@@ -0,0 +1,132 @@
...
+        Str('passphrase', private=True),
+    ))
+    @job(lock='pool_replace')
+    async def replace(self, job, oid, options):
+        """
+        Replace a disk on a pool.
+
+        `label` is the ZFS guid or a device name
+        `disk` is the identifier of a disk
+        `passphrase` is only valid for FreeBSD platform where pool is GELI encrypted
```
william-gr: TrueNAS Core/Enterprise

src/middlewared/middlewared/plugins/device_/device_info_linux.py
themylogin:
```
[server] /home/themylogin> sg_vpd --quiet --page=0x80 /dev/sde
  Unit serial number:      WD-WCAVY0211173
```

src/middlewared/middlewared/plugins/pool_/import_disk_freebsd.py
themylogin: @william-gr you can assign this part to me since I wrote it :)

src/middlewared/middlewared/pytest/unit/common/smart/test_smartctl.py
```diff
@@ -10,19 +10,28 @@
 
 @pytest.mark.asyncio
 async def test__get_smartctl_args__disk_nonexistent():
-    assert await get_smartctl_args(None, {}, "ada0") is None
+    with patch("middlewared.common.smart.smartctl.IS_LINUX", Mock(return_value=False)):
```
themylogin: You are mocking it with callable, while it is a scalar value. `bool(Mock(return_value=False))` is `True` so your tests are behaving like we are on Linux. Have you ran them, are they passing? (they should not, if they are passing, it is a ringing bell)

src/middlewared/middlewared/plugins/device_/device_info_linux.py
sonicaj: @william-gr the above example shows it's working nicely with SATA. Thank you @themylogin  :) ( also we should already have SATA drive serial from `lshw`, this is specifically for nvme drives which don't show up in `lshw` disk class )

PR #3993, opened 2020-01-02
NAS-104573 / 11.3 / Fix replication manual snapshot timezone

themylogin:


src/middlewared/middlewared/plugins/system.py
```diff
@@ -1067,6 +1068,16 @@ def _get_index(country_columns, column):
 
         raise CallError('Unable to connect to any of the specified UI addresses:\n' + '\n'.join(errors))
 
+    @private
+    async def set_timezone(self):
```
william-gr: Arent't we already doing this in _reload_tinmeservices?

themylogin: Have not thought about that. Maybe we should not do it there? It's a code duplication.

william-gr: If you're going to inspect all use cases, I dont mind either way

themylogin: @william-gr there are no, in TN too (actually that's an issue, who handles timezone change on idle node?)

william-gr:
_reload_timeservices does

service.* calls are replicated to standby

src/middlewared/middlewared/plugins/system.py
william-gr: _reload_timeservices does

service.* calls are replicated to standby

themylogin: @william-gr then guess I'll stop playing with fire and trying to look smart :)

PR #3996, opened 2020-01-03
NAS-104565 / 12 / Platform dependent swap configuration

sonicaj:


src/middlewared/middlewared/plugins/disk_/configure_swap.py
```diff
@@ -0,0 +1,206 @@
...
+                if not IS_LINUX and not dumpdev:
+                    dumpdev = await self.middleware.call('disk.dumpdev_configure', part_a)
+                name = await self.middleware.call('disk.new_swap_name')
+                if not name:
+                    # Which means maximum has been reached and we can stop
+                    break
+                part_a_path, part_b_path = all_partitions[part_a]['path'], all_partitions[part_b]['path']
+                if IS_LINUX:
+                    cp = await Popen(
+                        f'echo "y" | mdadm --create {os.path.join("/dev/md", name)} '
```
william-gr: In FreeBSD we dont write blocks in the disk so its not seen after reboot and make things simpler. Is there a way to do that in linux?

sonicaj: Yes it should be with `--build`, I'll update.

src/middlewared/middlewared/plugins/disk_/configure_swap.py
```diff
@@ -0,0 +1,223 @@
...
+                existing_swap_devices['partitions'].append(device)
+
+        # disk.get_mirrors is going to get us complete path for linux and mirror/swapname for freebsd
+        # point to note is that we don't get .eli suffix from above - so we have to be careful here and
+        # use encrypted_provider instead
+        for mirror in mirrors:
+            # If the mirror is degraded or disk is not in a pool lets remove it
+            mirror_name = (mirror['encrypted_provider'] or mirror['real_path']) if IS_LINUX else (
+                mirror['encrypted_provider'] or mirror['path']
+            ).strip('/dev/')
```
themylogin: `strip` does not work this way, also, can you please use regular if / else block?

src/middlewared/middlewared/plugins/disk_/configure_swap.py
```diff
@@ -0,0 +1,223 @@
...
+                'path': swap_device,
+                'encrypted_provider': all_partitions[unused_partitions[0]]['encrypted_provider'],
+            }
+
+        created_swap_devices = []
+        for name, data in create_swap_devices.items():
+            if IS_LINUX:
+                if not data['encrypted_provider']:
+                    cp = await Popen(
+                        f'echo "y" | cryptsetup -d /dev/urandom open --type plain {name} {name.split("/")[-1]}',
```
themylogin: `asyncio.create_subprocess_exec` has `stdin` argument, let's use it instead of the shell.

src/middlewared/middlewared/plugins/disk_/configure_swap.py
```diff
@@ -0,0 +1,223 @@
...
+            )
+            if swap_mirror and (len(mirror['providers']) == 1 or any(
+                p['disk'] not in disks for p in mirror['providers']
+            )):
+                await self.middleware.call('disk.swaps_remove_disks', [p['disk'] for p in mirror['providers']])
+                if mirror_name in existing_swap_devices['mirrors']:
+                    existing_swap_devices['mirrors'].remove(mirror_name)
+            else:
+                if swap_mirror and mirror_name not in existing_swap_devices['mirrors']:
+                    create_swap_devices[mirror_name if IS_LINUX else mirror_name.rsplit('.eli', 1)[0]] = {
```
themylogin: Please group all platform-dependent assignments in one place, it's very hard to understand what's going on now

src/middlewared/middlewared/plugins/disk_/configure_swap.py
```diff
@@ -0,0 +1,223 @@
...
+        disks = [i async for i in await self.middleware.call('pool.get_disks')]
+        disks.extend(await self.middleware.call('boot.get_disks'))
+        existing_swap_devices = {'mirrors': [], 'partitions': []}
+        mirrors = await self.middleware.call('disk.get_mirrors')
+        encrypted_mirrors = {m['encrypted_provider']: m for m in mirrors if m['encrypted_provider']}
+        all_partitions = {p['name']: p for p in await self.middleware.call('disk.list_all_partitions')}
+
+        for device in await self.middleware.call('disk.get_swap_devices'):
+            if device in encrypted_mirrors or device.startswith(('/dev/md', 'mirror/')):
+                # This is going to be complete path for linux and mirror/swapname.eli for freebsd
```
william-gr: Why for linux is absolute path and for freebsd it returns relative? Thats a bit confusing, no?

src/middlewared/middlewared/plugins/disk_/configure_swap.py
```diff
@@ -0,0 +1,223 @@
...
+
+                # If mirror has been configured automatically (not by middlewared)
+                # and there is no geli attached yet we should look for core in it.
+                if not IS_LINUX and mirror['config_type'] == 'AUTOMATIC' and not mirror['encrypted_provider']:
+                    await run(
+                        'savecore', '-z', '-m', '5', '/data/crash/', f'/dev/{mirror_name}',
+                        check=False
+                    )
+
+        if (len(existing_swap_devices['mirrors']) + len(existing_swap_devices['partitions'])) > MIRROR_MAX:
```
william-gr: This feels an odd check.

How could we ever have mixed mirrors and single partition swap?

If thats is happening we have a bug. We have one or another. And we shouldnt have more than one partition. If there is more than one partition it would be a mirror instead.

src/middlewared/middlewared/plugins/disk_/configure_swap.py
```diff
@@ -0,0 +1,223 @@
...
+                part_a_path, part_b_path = all_partitions[part_a]['path'], all_partitions[part_b]['path']
+                try:
+                    await self.middleware.call(
+                        'disk.create_mirror', name, {
+                            'paths': [part_a_path, part_b_path],
+                            'extra': {'level': 1} if IS_LINUX else {},
+                        }
+                    )
+                except CallError:
+                    self.logger.warning('Failed to create swap mirror %s: %s', name)
```
william-gr: Two %s, one arg

src/middlewared/middlewared/plugins/disk_/configure_swap.py
```diff
@@ -0,0 +1,223 @@
...
+                    self.logger.warning(f'Failed to make swap for %s: %s', name, e.stderr.decode())
+                    continue
+            elif not IS_LINUX and not data['encrypted_provider']:
+                try:
+                    await run('geli', 'onetime', name)
+                except subprocess.CalledProcessError as e:
+                    self.logger.warning('Failed to encrypt swap partition %s: %s', name, e.stderr.decode())
+                    continue
+
+            name = name if IS_LINUX else f'{name}.eli'
```
william-gr: if not IS_LINUX:
   name = f'{name}.eli'

reads better, than assigning name to itself (name = name)

src/middlewared/middlewared/plugins/disk_/mirror_base.py
```diff
@@ -0,0 +1,35 @@
+from middlewared.schema import accepts, Dict, Str, List
+from middlewared.service import filterable, private, ServicePartBase
+
+
+class DiskMirrorBase(ServicePartBase):
```
william-gr: I think it could be useful if we rename all mentions of disk mirror to disk swap mirror, otherwise feels like we are handling any regular mirrors for the system.

src/middlewared/middlewared/plugins/disk_/configure_swap.py
```diff
@@ -0,0 +1,223 @@
+import os
```
themylogin: Name it `swap_configure.py` so it matches `swap_remove.py`

src/middlewared/middlewared/plugins/disk_/swap_remove.py
```diff
@@ -0,0 +1,52 @@
...
+IS_LINUX = platform.system().lower() == 'linux'
+
+
+class DiskService(Service):
+
+    @private
+    async def swaps_remove_disks(self, disks):
+        """
+        Remove a given disk (e.g. ["da0", "da1"]) from swap.
+        it will offline if from swap, remove it from the gmirror (if exists)
```
william-gr: Might not want to mention gmirror/geli here in the common interface

src/middlewared/middlewared/plugins/disk_/swap_configure.py
```diff
@@ -0,0 +1,221 @@
...
+        create_swap_devices = {}
+        disks = [i async for i in await self.middleware.call('pool.get_disks')]
+        disks.extend(await self.middleware.call('boot.get_disks'))
+        existing_swap_devices = {'mirrors': [], 'partitions': []}
+        mirrors = await self.middleware.call('disk.get_swap_mirrors')
+        encrypted_mirrors = {m['encrypted_provider']: m for m in mirrors if m['encrypted_provider']}
+        all_partitions = {p['name']: p for p in await self.middleware.call('disk.list_all_partitions')}
+
+        for device in await self.middleware.call('disk.get_swap_devices'):
+            if device in encrypted_mirrors or device.startswith(('/dev/md', 'mirror/')):
```
william-gr: If its complete path now, why `mirror/` and not `/dev/mirror/` ?

src/middlewared/middlewared/plugins/disk_/swap_mirror_freebsd.py
```diff
@@ -0,0 +1,51 @@
...
+from middlewared.service import CallError, Service
+from middlewared.utils import filter_list, run
+
+from .swap_mirror_base import DiskSwapMirrorBase
+
+
+class DiskService(Service, DiskSwapMirrorBase):
+
+    async def create_swap_mirror(self, name, options):
+        cp = await run('gmirror', 'create', name, *(options['paths']), check=False, encoding='utf8')
```
william-gr: Not important, but why *(options['paths']) and not *options['paths']

src/middlewared/middlewared/plugins/disk_/swap_configure.py
sonicaj: True, fixed

src/middlewared/middlewared/plugins/disk_/swap_mirror_freebsd.py
sonicaj: No reason at all, thank you for pointing it out

PR #3997, opened 2020-01-03
NAS-103610 / 12.0 / feat(middlewared/failover): persistent HA connection and status event

william-gr:


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/remote.py
```diff
@@ -0,0 +1,252 @@
...
+            if "'NoneType' object has no attribute 'text_message'" in str(e):
+                raise CallError('Remote connection closed.', errno.EREMOTE)
+            else:
+                raise
+        except ClientException as e:
+            raise CallError(str(e), e.errno or errno.EFAULT)
+
+    def subscribe(self, name, callback):
+        # Only subscribe if we are already connected, otherwise simply register it
+        if name not in self._subscriptions and self.is_connected():
```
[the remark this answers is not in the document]
william-gr: The way subscriptions are being used probably not, but I will address it

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/remote.py
```diff
@@ -0,0 +1,255 @@
...
+            try:
+                cb(self.middleware)
+            except Exception:
+                logger.error('Failed to run on_disconnect for remote client', exc_info=True)
+
+    def call(self, *args, **kwargs):
+        try:
+            if not self.connected.wait(timeout=20):
+                if self.remote_ip is None:
+                    raise CallError('Unable to determine remote node IP')
```
themylogin: How about we store that exception in Client class and raise it? So we can see why we are unable to determine remote node IP.

william-gr: There is no reason why. It just means it couldnt detect A or B

PR #3998, opened 2020-01-03
NAS-104587 / 11.2 / fix calling _upload multiple times on GELI import

yocalebo:
@william-gr This fixes the `_upload` endpoint from being called multiple times with the `disk.decrypt` method.

### DIFF
--- a/gui/storage/forms.py
+++ b/gui/storage/forms.py
@@ -981,7 +981,7 @@ class AutoImportDecryptForm(Form):
             ]
 
     def done(self, *arg, **kwargs):
-        super(AutoImportDecryptForm, self).done()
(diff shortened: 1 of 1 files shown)


gui/storage/forms.py
```diff
@@ -980,15 +980,10 @@ def _populate_disk_choices(self):
                 for i in c.call('disk.get_encrypted', {'unused': True})
             ]
 
-    def clean(self):
+    def done(self, *arg, **kwargs):
+        super(AutoImportDecryptForm, self).done()
```
william-gr: I think we may need to pass *args and **kwargs here?

yocalebo: Whoops, just saw this. Fixed it.

PR #4000, opened 2020-01-05
NAS-104595 / 11.3 / Fix importing pool with new specified name

sonicaj:


sonicaj:
This isn't directly used by the UI, however any API user would file a bug as the end point's docs suggest the corrected behaviour where user is able to change the pool's name while importing it.

william-gr:
Lets fix this for 12 then

sonicaj:
backport

PR #4002, opened 2020-01-06
NAS-104585 / 11.3 / Generate log message on autotune failures

anodos325:


src/middlewared/middlewared/plugins/system.py
```diff
@@ -222,6 +222,8 @@ def autotune(self, conf='loader'):
                 f'--userland-reserved={userland_reserved}', '--conf', conf
             ], capture_output=True
         )
+        if cp.returncode != 0:
+            self.logger.debug('autotune for [%s] failed.', conf, exc_info=True)
```
william-gr: We probably want a warn instead of debug here, since debug is not the default.

exc_info=True would also not be useful here since there isnt a traceback

anodos325: Good points. Don't know what I was thinking :)

PR #4017, opened 2020-01-07
NAS-104627 / 11.3 / Validate iscsi target/extent before deletion

sonicaj:


src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -606,14 +606,25 @@ class Config:
     @accepts(
         Int('id'),
         Bool('remove', default=False),
+        Bool('validate', default=True),
```
william-gr: Small nit, `validate` feels strange.

Can we use `force` and default to False instead?

sonicaj: Sure

PR #4020, opened 2020-01-08
NAS-102573 / 12.0 / Add FSRVP protocol support to middleware

anodos325:
Have middleware configure correct settings for FSRVP protocol support
in Samba. This relies on the vfs module zfs_fsrvp, and allows clients
to manage snapshots on the TrueNAS server through the FSRVP protocol.
Linux and BSD clients can use rpcclient to administer snapshots and
Windows clients can use DiskShadow.exe. The protocol requires clients
be able to generate shares to snapshots on the fly, and so a writable
libsmbconf backend was required for shares. Utilize existing registry
share capabilities in samba to provide this feature.


src/middlewared/middlewared/etc_files/local/smb4_share.conf
```diff
@@ -9,21 +9,15 @@
 
         def get_db_config():
             db = {}
+            db['globals'] = middleware.call_sync('sharing.smb._get_global_params', None)
+            logger.debug('globals: %s', db['globals'])
```
william-gr: Leftover debug or on purpose?

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -952,6 +992,285 @@ class Config:
 
         return vuid
 
+    @private
```
william-gr: I think we dont need `@private` for methods starting with underscore

src/middlewared/middlewared/etc_files/local/smb4_share.conf
```diff
@@ -9,21 +9,15 @@
 
         def get_db_config():
             db = {}
+            db['globals'] = middleware.call_sync('sharing.smb._get_global_params', None)
```
themylogin: When we compose service from multiple files (which I think we'll need to do now) we don't import methods whose names begin with `_`.  Our new agreement is to use normal method names and just make them private.

src/middlewared/middlewared/etc_files/local/smb4_share_load.py
```diff
@@ -0,0 +1,20 @@
+import logging
+
+from middlewared.utils import run
+
+logger = logging.getLogger(__name__)
+
+
+async def reload_registry_shares(middleware):
+    drop = await run(['/usr/local/bin/net', 'conf', 'drop'], check=False)
```
themylogin: We're also working on Linux product now. On FreeBSD we have `/usr/local/bin` in `PATH` so just `net` will work and most likely it will also work in Linux from `/usr/bin`?

themylogin:
@anodos325 great work, but I agree with William: we have the ability to split plugins into multiple file now (please note directories in `plugins/`) and it would be very convenient to at least stop growing our files (which are already giant and hard to navigate) so It would be great if you move all new methods into separate file (or files).

src/middlewared/middlewared/etc_files/local/smb4_share.conf
anodos325: Okay. I will switch methods to not begin with an underscore.

anodos325:
> @anodos325 great work, but I agree with William: we have the ability to split plugins into multiple file now (please note directories in `plugins/`) and it would be very convenient to at least stop growing our files (which are already giant and hard to navigate) so It would be great if you move all new methods into separate file (or files).

I've moved smb.sharesec into a different file (since it was a different service). I was thinking of keeping this particular commit generic for 11.3 / 12.0 so that I can backport if absolutely needed (hopefully not). I can make a future PR to break up LDAP / AD / SMB into separate files inside directories.

src/middlewared/middlewared/plugins/smb_sharesec.py
```diff
@@ -0,0 +1,409 @@
...
+    LABEL = 10
+
+
+class ShareSec(CRUDService):
+
+    class Config:
+        namespace = 'smb.sharesec'
+
+    @private
+    async def _parse_share_sd(self, sd, options=None):
```
themylogin: You don't need to mark them with @private if they start with underscore, they won't be visible outside this class

src/middlewared/middlewared/plugins/smb_sharesec.py
```diff
@@ -0,0 +1,409 @@
...
+    @private
+    async def _view_all(self, options=None):
+        """
+        Return Security Descriptor for all shares.
+        """
+        share_sd_list = []
+        idx = 1
+        share_entries = (await self._sharesec(action='--view-all')).split('\n\n')
+        for share in share_entries:
+            parsed_sd = await self.middleware.call('smb.sharesec._parse_share_sd', share, options)
```
themylogin: This (`smb.sharesec._parse_share_sd`) won't work, middleware does not see these methods

themylogin:
@anodos325 makes sense

src/middlewared/middlewared/etc_files/local/smb4_share.conf
anodos325: Leftover. Removed.

PR #4027, opened 2020-01-09
NAS-104646 / 12.0 / Add support for platform-specific SMB-related commands and paths

anodos325:
Linux-side of enums are placeholder values. I will properly set them once I have made ix-samba package for Linux.

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/smb4_share_load.py
+++ b/src/middlewared/middlewared/etc_files/local/smb4_share_load.py
@@ -3,7 +3,11 @@ import logging
 from middlewared.utils import run
 from middlewared.plugins.smb import SMBCmd, SMBPath
 
+import platform
+
 logger = logging.getLogger(__name__)
+
+IS_LINUX = platform.system().lower() == 'linux'
 
 
 async def reload_registry_shares(middleware):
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/etc_files/local/smb4_share_load.py
```diff
@@ -3,15 +3,20 @@
 from middlewared.utils import run
 from middlewared.plugins.smb import SMBCmd, SMBPath
 
+import platform
+
 logger = logging.getLogger(__name__)
 
+IS_LINUX = platform.system().lower() == 'linux'
```
william-gr: Is this used?

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -45,23 +48,29 @@ class SMBHAMODE(enum.IntEnum):
 
 
 class SMBCmd(enum.Enum):
-    NET = '/usr/local/bin/net'
-    PDBEDIT = '/usr/local/bin/pdbedit'
-    SHARESEC = '/usr/local/bin/sharesec'
-    SMBCONTROL = '/usr/local/bin/smbcontrol'
-    SMBPASSWD = '/usr/local/bin/smbpasswd'
-    WBINFO = '/usr/local/bin/wbinfo'
+    NET = ('/usr/local/bin/net', 'net')
```
william-gr: Do we really need both? Why not just the binary name? /usr/local/bin is in PATH

PR #4032, opened 2020-01-09
NAS-104634 / 11.3 / Apply autotune after middleware starts

anodos325:


src/freenas/etc/ix.rc.d/ix-autotune
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+#
+# $FreeBSD$
+#
+#
+
+# PROVIDE: ix-autotune
+# REQUIRE: NETWORKING
```
william-gr: If we need to reboot we probably want to do it earlier than networking, possibly right after middlewared?

src/freenas/etc/ix.rc.d/ix-autotune
```diff
@@ -0,0 +1,34 @@
...
+# PROVIDE: ix-autotune
+# REQUIRE: middlewared
+# BEFORE: NETWORKING
+
+. /etc/rc.freenas
+
+do_autotune()
+{
+	/usr/local/bin/midclt call system.advanced.autotune sysctl
+	/usr/local/bin/midclt call system.advanced.autotune loader
```
william-gr: Just noticed the return code is the result of the call as text, not the actual exit code of the process

PR #4041, opened 2020-01-10
Fix smb.getparm for registry config

anodos325:
param.LoadParm().get() in samba doesn't return values for registry shares.
Have smb.getparm call "net conf getparm" for non-global sections, and give
same return types as param.LoadParm().get().

### DIFF
--- a/src/middlewared/middlewared/plugins/smb_/registry.py
+++ b/src/middlewared/middlewared/plugins/smb_/registry.py
@@ -21,7 +21,15 @@ class SharingSMBService(CRUDService):
         samba's registry.tdb file.
         """
         action = kwargs.get('action')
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/smb_/registry.py
```diff
@@ -88,6 +96,20 @@ class Config:
...
+    async def reg_getparm(self, share, parm):
+        to_list = ['vfs objects', 'hosts allow', 'hosts deny']
+        try:
+            ret = await self.netconf(action='getparm', share=share, args=[parm])
+        except Exception as e:
+            if f"Error: given parameter '{parm}' is not set." in e.errmsg:
+                # Copy behavior of samba python binding
+                return None
+            else:
+                raise e
```
william-gr: I think this should be just `raise` or it will reraise the exception in na improper way

PR #4057, opened 2020-01-14
NAS-102430 / 12.0 / Make service announcements configurable

anodos325:


src/middlewared/middlewared/alembic/versions/12.0/2020-01-14_13-26_add_service_announcments.py
```diff
@@ -0,0 +1,40 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('network_globalconfiguration', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('gc_service_announcement', sa.JSON(type=dict), nullable=True))
+
+    op.execute("UPDATE network_globalconfiguration SET gc_service_announcement = '{"mdns": true,  "wsd": true, "netbios": false)'
```
william-gr: Json is not valid, missing }

src/middlewared/middlewared/plugins/network.py
```diff
@@ -155,6 +161,13 @@ def network_config_extend(self, data):
             Str('hostname_virtual', validators=[Match(r'^[a-zA-Z\.\-\0-9]+$')]),
             Str('domain', validators=[Match(r'^[a-zA-Z\.\-\0-9]+$')]),
             List('domains', items=[Str('domains')]),
+            Dict(
+                'service_announcement',
+                Bool('netbios', default=False),
+                Bool('mdns', default=True),
+                Bool('wsd', default=True),
+                register=True,
```
william-gr: Do we need register=True? I dont see that schema being referenced in other places, I may be msising something

Maybe what we want  here is update=True, so not all keys are populated if not provided?

src/middlewared/middlewared/alembic/versions/12.0/2020-01-14_13-26_add_service_announcments.py
```diff
@@ -0,0 +1,40 @@
...
+revision = '39a133a04496'
+down_revision = '133f2d9049d2'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('network_globalconfiguration', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('gc_service_announcement', sa.JSON(type=dict), nullable=True))
```
themylogin: It won't work IIRC, it should be just `sa.TEXT()`

src/middlewared/middlewared/plugins/network.py
themylogin: @william-gr will `update` work for inline schemas?

william-gr: Good question, I dont know, we should verify it

src/middlewared/middlewared/alembic/versions/12.0/2020-01-14_13-26_add_service_announcments.py
anodos325: Okay. Thanks.

src/middlewared/middlewared/plugins/network.py
themylogin: It will not, it will be hard to implement, it will break PATCH restful method semantics and cause a lot of other confusion.

anodos325: Removed this.

src/middlewared/middlewared/alembic/versions/12.0/2020-01-14_13-26_add_service_announcments.py
anodos325: Indeed. Laptop was on ~5% battery left when I force-pushed that. Whoops.

src/middlewared/middlewared/alembic/versions/12.0/2020-01-14_13-26_add_service_announcments.py
```diff
@@ -0,0 +1,40 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('network_globalconfiguration', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('gc_service_announcement', sa.Text(), nullable=True))
+
+    defaults = '{"mdns": true, "wsd": true, "netbios": false}'
+    op.execute(f"UPDATE network_globalconfiguration SET gc_service_announcement = \'{defaults}\'")
```
themylogin: Sorry I forgot to mention this the first time, we also need to set this column back to nullable

src/middlewared/middlewared/alembic/versions/12.0/2020-01-14_13-26_add_service_announcments.py
```diff
@@ -0,0 +1,43 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('network_globalconfiguration', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('gc_service_announcement', sa.Text(), nullable=True))
+
+    defaults = '{"mdns": true, "wsd": true, "netbios": false}'
```
william-gr: Shouldnt we be setting mdns value based on the current value of zeroconf?

PR #4063, opened 2020-01-15
NAS-104579 / 12.0 / Shell choices fixes

sonicaj:

1) Exposes a knob to ensure that consumers of middlewared are able to retrieve shell choices by specifying a user id. This ensures that netcli is not returned for non-root users.
2) Fixes a bug where` netcli.sh` path in `/etc/shells` would have double `//` in the start.
3) Migration for existing users having invalid `netcli.sh` value for shell.
4) Add validation for shell for users create/update

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -496,16 +496,20 @@ class UserService(CRUDService):
 
         return pk
 
-    @accepts()
-    def shell_choices(self):
+    @accepts(Int('oid', default=None))
+    def shell_choices(self, oid=None):
         """
         Return the available shell choices to be used in `user.create` and `user.update`.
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/account.py
```diff
@@ -496,16 +496,20 @@ def do_home_copy():
 
         return pk
 
-    @accepts()
-    def shell_choices(self):
+    @accepts(Int('oid', default=None))
```
william-gr: Minor nit, `oid` name feels odd, is it used anywhere else? As its shown in the docs I thought we could say simply `id` or `user_id` ?

sonicaj: Updated :)

PR #4065, opened 2020-01-16
NAS-104674 / 12.0 / fix(middlewared/pool): use a lock for syncing encrypted disk table

william-gr:
There is a race condition with ZFS events.

Ticket:	NAS-104674


william-gr:
backport

PR #4067, opened 2020-01-16
NAS-104859 / 11.3 / Fix crash on trying to save deleted replication task state

themylogin:


william-gr:
Do we need 12 backport?

PR #4068, opened 2020-01-16
Update migrate113 port

themylogin:


nas_ports/freenas/freenas-migrate113/Makefile
```diff
@@ -1,7 +1,8 @@
 # $FreeBSD$
 
 PORTNAME=	freenas-migrate113
-PORTVERSION=	2019112901
+PORTVERSION=	0.1
```
william-gr: Why the version change?

0.1 is smaller than 2019112901 so we need PORTEPOCH set if we really need that

themylogin: My copypasted script from zettarepl port did it :)

nas_ports/freenas/freenas-migrate113/Makefile
```diff
@@ -1,7 +1,8 @@
 # $FreeBSD$
 
 PORTNAME=	freenas-migrate113
-PORTVERSION=	2019112901
+PORTVERSION=	2020000000
+PORTREVISION=	1
```
william-gr: Dont need the revision, fwiw

themylogin: It will be the one that is incremented every time

PR #4070, opened 2020-01-16
NAS-104708 / 12.0 / Generate nginx configuration after interfaces have synced

sonicaj:
This PR introduces changes to ensure that we generate nginx configuration after interfaces have synced.


sonicaj:
backport

PR #4076, opened 2020-01-17
NAS-104735 / 12.0 / Simplify SMB share parameters

anodos325:
Remove unused DB columns from services_cifs
Remove VFS objects, showhiddenfiles, guestonly and guest only fields
During migration, move any parameters that can't be moved into new
fields into SMB auxiliary parameters.

Add 'acl', 'streams', 'aapl_name_mangling', 'durablehandle' fields.
This will expose the ability to enable SMB2 durable handles to end-users
as well as apply a common set of non-backward-compatible name mangling
features that also may be of benefit to MacOS users. "vfs_fruit" is
not enabled by setting the global (services_cifs) parameter "aapl_extensions"
which is more descriptive of what it actually does (enables AAPL protocol
extensions).

Add ability to get preset configurations for shares for specific tasks. These
are potentially useful parameter combinations that are non-obvious.


src/middlewared/middlewared/alembic/versions/12.0/2020-01-16_19-16_simplify_smb_shares.py
```diff
@@ -0,0 +1,102 @@
...
+        batch_op.add_column(sa.Column('cifs_acl', sa.Boolean(), nullable=True))
+        batch_op.add_column(sa.Column('cifs_aapl_name_mangling', sa.Boolean(), nullable=True))
+        batch_op.add_column(sa.Column('cifs_path_suffix', sa.String(length=255), nullable=True))
+        batch_op.add_column(sa.Column('cifs_streams', sa.Boolean(), nullable=True))
+        batch_op.add_column(sa.Column('cifs_durablehandle', sa.Boolean(), nullable=True))
+
+    conn = op.get_bind()
+    smb_shares = [dict(row) for row in conn.execute("SELECT * FROM sharing_cifs_share").fetchall()]
+    for share in smb_shares:
+        aux_params = share.get('cifs_auxsmbconf', '').split()
```
william-gr: Did you mean split('\n') ?

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -265,14 +321,15 @@ def getparm(self, parm, section):
...
+        Bool('aapl_extensions'),
         Bool('localmaster'),
         Str('guest'),
         Str('admin_group', required=False, default=None, null=True),
         Str('filemask'),
         Str('dirmask'),
         Bool('ntlmv1_auth'),
         List('bindip', items=[IPAddr('ip')], default=[]),
-        Str('smb_options', max_length=None),
+        List('smb_options', default=[]),
```
william-gr: Curious what you all (@freenas/middleware) think about this. How should we handle backward compatibility of the API? Should we care?

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -675,32 +773,8 @@ class Config:
...
-            'shell_snap',
-            'streams_xattr',
-            'shadow_copy2',
-            'winmsa',
-            'zfs_space',
-            'zfsacl'
-        ]
-
-        return vfs_modules
+    async def presets(self):
```
william-gr: Need docstring so its not empty in the docs

src/middlewared/middlewared/plugins/smb_/registry.py
```diff
@@ -247,10 +238,20 @@ class Config:
                 "level2 oplocks": "no",
                 "oplocks": "no"
             })
+            if data['durablehandle']:
+                self.logger.warn("Disabling durable handle support on SMB share [%s] "
+                                 "due to NFS export of same path.")
+                self.middleware.call('datastore.update', 'sharing.cifs_share',
```
william-gr: missing await

src/middlewared/middlewared/plugins/smb.py
anodos325: Whoops forgot to undo that one. I was considering switching to a list because I hated all the \n characters.

src/middlewared/middlewared/plugins/smb_/registry.py
anodos325: Thank you.

src/middlewared/middlewared/alembic/versions/12.0/2020-01-16_19-16_simplify_smb_shares.py
anodos325: yes

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -265,14 +321,15 @@ def getparm(self, parm, section):
...
+        Bool('aapl_extensions'),
         Bool('localmaster'),
         Str('guest'),
         Str('admin_group', required=False, default=None, null=True),
         Str('filemask'),
         Str('dirmask'),
         Bool('ntlmv1_auth'),
         List('bindip', items=[IPAddr('ip')], default=[]),
-        Str('smb_options', max_length=None),
+        Str('smb_options', default=[]),
```
william-gr: Forgot to revert to max_length=None as well

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -68,6 +68,67 @@ def platform(self):
...
+        'acl': False,
+        'streams': False,
+        'durablehandle': False,
+        'auxsmbconf': '\n'.join([
+            'oplocks = no',
+            'level 2 oplocks = no',
+            'strict locking = yes',
+        ])
+    }}
+    PRIVATE_DATASETS = {"verbose_name": "Private SMB Datasets", "params": {
```
yocalebo: Being nit-picky, but `PRIVATE_DATASETS` is confusing to me because you're turning on feature of `ixnas` related to home directories. Maybe name it something related to home directories?

anodos325: Home directories would make it more confusing. What I need to actually do is deprecate that parameter name (keep it around for backwards-compatibility), and instead call it something like "ixnas:dataset_autocreate". With these changes,  users could potentially have automatic share path creation for things other than username. (For instance, primary group, netbios name of computer, IP address, etc, etc).

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -589,18 +681,21 @@ class Config:
...
-        if 'noacl' in data['vfsobjects']:
-            if not await self.middleware.call('filesystem.acl_is_trivial', data['path']):
-                verrors.add(
-                    f'{schema_name}.vfsobjects',
-                    f'The "noacl" VFS module is incompatible with the extended ACL on {data["path"]}.'
-                )
+        if not data['acl'] and not await self.middleware.call('filesystem.acl_is_trivial', data['path']):
+            verrors.add(
+                f'{schema_name}.acl',
+                f'ACL detected on {data["path"]}. ACLs must be stripped prior to creation'
```
themylogin: It'll be `creationof SMB share`, you've forgot to add whitespace

src/middlewared/middlewared/alembic/versions/12.0/2020-01-16_19-16_simplify_smb_shares.py
```diff
@@ -0,0 +1,102 @@
...
+        if share['cifs_guestonly']:
+            aux_params.append('guest only = yes')
+
+        if has_nondefault_vfs_objects:
+            aux_params.append(f"vfs objects = {vfs_objects.replace(',', ' ')}")
+
+        new_aux = '\n'.join(aux_params)
+        op.execute(f"UPDATE sharing_cifs_share SET cifs_acl = {has_acl}, cifs_streams = {has_streams}, "
+                   f"cifs_aapl_name_mangling = {has_catia}, cifs_durablehandle= {set_durable}, "
+                   f"cifs_path_suffix = '', cifs_auxsmbconf = \'{new_aux}\' WHERE id={share['id']}")
```
themylogin: `new_aux` can contain single quotes, please use parameters binding
```
        conn.execute("UPDATE vm_device SET attributes = ? WHERE id = ?", (
            json.dumps(vnc_device['attributes']), vnc_device['id']
        ))
```

src/middlewared/middlewared/plugins/smb_/registry.py
```diff
@@ -247,10 +238,20 @@ class Config:
                 "level2 oplocks": "no",
                 "oplocks": "no"
             })
+            if data['durablehandle']:
+                self.logger.warn("Disabling durable handle support on SMB share [%s] "
+                                 "due to NFS export of same path.")
```
william-gr: Missing argument to %s

PR #4077, opened 2020-01-17
NAS-104733 / 11.3 / Allow users to specify custom port for LDAP

anodos325:


src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -155,9 +155,15 @@ def _open(self):
             saved_simple_error = None
             saved_gssapi_error = None
             for server in self.hosts:
+                parts = server.split(':', 1)
```
william-gr: @anodos325 

```
>>> 'foo:'.split(':')
['foo', '']
```

which means port will be empty string, not sure we care enough and/or want to validate it?

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -442,13 +469,27 @@ class Config:
...
                 "The LDAP hostname parameter is required."
             )
+        for idx, uri in enumerate(new["uri_list"]):
+            parsed = urlparse(uri)
+            try:
+                port = parsed.port
+
+            except ValueError:
+                verrors.add(f"ldap_update.hostname.{idx}",
+                            "Port number [{port}] is not an integer.")
```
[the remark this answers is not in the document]
themylogin: Better to say "Invalid port number" because
```
>>> urllib.parse.urlparse('url://localhost:65536').port
  File "/usr/lib/python3.6/urllib/parse.py", line 171, in port
    raise ValueError("Port out of range 0-65535")
ValueError: Port out of range 0-65535
```

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -3,6 +3,7 @@
 import errno
 import fcntl
 import grp
+import ipaddr
```
william-gr: Should use ipaddress module instead, ipaddr is deprecated

PR #4078, opened 2020-01-18
Migrate users to minimum of signed LDAP SASL wrapping

anodos325:


william-gr:
@anodos325 can we also update it in migrate113 repo, please? Just so we dont end up forgetting

PR #4082, opened 2020-01-20
NAS-104542 / 11.3 / Fix VMware snapshot error reporting

themylogin:


themylogin:
@william-gr I've tested this, there should be no more fixes

PR #4095, opened 2020-01-21
NAS-104554 / 11.3 / Use UTC datetimes for job start / end time

themylogin:


themylogin:
@william-gr no, it just returns epoch * 1000 there
We can't convert anything because `datetime.now()` and `datetime.utcnow()` are both naive (as all other datetimes we use).
I think we should stick to UTC datetimes (as we already do) because we are the one product where timezone can be changed during runtime.

PR #4096, opened 2020-01-21
NAS-104576 / 11.3 / Make pool.dataset.kill_processes only restart services that have shares. Other services we have to stop completely

themylogin:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3265,9 +3265,14 @@ def __attachments_path(self, dataset):
...
-                    self.logger.info('Restarting service %r that holds dataset %r', process['service'], oid)
-                    await self.middleware.call('service.restart', process['service'])
+                service = process.get('service')
+                if service is not None:
+                    if any(attachment_delegate.service == service for attachment_delegate in self.attachment_delegates):
+                        self.logger.info('Restarting service %r that holds dataset %r', service, oid)
+                        await self.middleware.call('service.restart', service)
+                    else:
+                        self.logger.info('Stopping service %r that holds dataset %r', service, oid)
+                        await self.middleware.call('service.stop', service)
```
william-gr: I am probably misunderstanding what is going on here.
Cant be there other pools using that service?

themylogin: What do you mean by "pools are using the service"?

We were doing `service.restart` so that, if our pool attachment delegates (that close samba shares without interrupting other connections) fail, we still had the ability to detach the pool by restarting them completely.
What I have missed is that there are other services like FTP or S3 that, if they are configured to store data in that pool, no matter how many times you restart them, they'll still be "using" that pool. The only option is to stop them.

william-gr: Yes, but you're stopping a service that is holding that dataset.

Are we ever trying to start it again after detachment? Or, how can you be sure that this service in question will not start without that dataset?

themylogin: > Are we ever trying to start it again after detachment?

No

> how can you be sure that this service in question will not start without that dataset?

Do I need to be? If we've exported minio pool, what is the point of having minio started?

themylogin: It's explicitly not AFP, CIFS, iSCSI or NFS that can have different shares on different datasets/pools. Everything else we'll need to terminate.

william-gr: We could have WebDAV? Which we dont currently have attachments?

My point is that piece of code does not make it clear. If destroying a pool which were in use by FTP and S3, should the user be properly notified about it via attachment as well? Simply stopping the service seems odd

william-gr: I mean, its fine for 11.3, I imagine, but not as a final solution. Unless again, I am misunderstanding how this all is working

themylogin: > If destroying a pool which were in use by FTP and S3, should the user be properly notified about it via attachment as well? Simply stopping the service seems odd

The problem is how will this attachment detach itself? Set FTP root to /dev/null? :) Stopping the service seems fine to me. I agree that we need to extend API a little bit to say "these services will be restarted, but these we'll need to stop" (now it says "we'll restart FTP" but it will really stop it)

william-gr: Wont that also stop WebDAV?

But yes, agreed on everything else

PR #4102, opened 2020-01-22
NAS-104764 / 12.0 / Drop iscsi initiator tag column

sonicaj:
This PR drops iSCSI initiator tag column as it is not being used anywhere. We use each initiator's database row id to keep track of mappings in the UI as well which was the only place it was being used before.


amotin:
Just in case, I think it could be added since the tag is 15-bit value, so it can not be just a plain sequential auto-increment, otherwise at some (likely distant) future it will overflow.  But if it is not a problem for how the database work, then probably OK.

amotin:
Or may be I confused it with portal tag, which I was talking about, in which case disregard. :)

PR #4103, opened 2020-01-22
NAS-104880 / 11.2 / Allow SSSD to be id provider for winbindd if idmap backend is nss

anodos325:


william-gr:
@anodos325 ping. Also would we need this for U8?

anodos325:
> @anodos325 ping. Also would we need this for U8?

Yes. I'll make one more minor UI change and then remove WIP tag.

PR #4106, opened 2020-01-23
Set default for SMB share purpose to DEFAULT_SHARE

anodos325:


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -469,7 +469,7 @@ class Config:
 
     @accepts(Dict(
         'sharingsmb_create',
-        Str('purpose', enum=[x.name for x in SMBSharePreset], default="DEFAULT"),
+        Str('purpose', enum=[x.name for x in SMBSharePreset], default="NO_PRESET"),
```
william-gr: Using SMBSharePreset.NO_PRESET might be more future proof, but not required.

Question, why not DEFAULT_SHARE instead of NO_PRESET, any especial reason?

PR #4115, opened 2020-01-24
Temporarily fix sshd_config for Linux

william-gr:


src/middlewared/middlewared/etc_files/local/ssh/sshd_config
```diff
@@ -41,7 +41,8 @@ ClientAliveCountMax 3
 % if 'ClientAliveInterval' not in ssh_config['options']:
 ClientAliveInterval 15
 % endif
-% if 'NoneEnabled' not in ssh_config['options']:
+## Scale doesnt have HPN patches yet
+% if middleware.call_sync('system.platform') != 'LINUX' and 'NoneEnabled' not in ssh_config['options']:
```
sonicaj: We already have `platform` in etc plugins. How about we use that instead ?

PR #4118, opened 2020-01-26
NAS-104213 / 12.0 / Devd alternative for TN Scale

sonicaj:


src/middlewared/middlewared/plugins/zfs_/events_linux.py
```diff
@@ -0,0 +1,57 @@
...
+            if middleware.call_sync('system.state') != 'SHUTTING_DOWN':
+                middleware.logger.error('Failed to retrieve ZFS events: %s', str(e))
+            else:
+                break
+
+        time.sleep(1)
+
+
+async def setup(middleware):
+    start_daemon_thread(target=setup_zfs_events_process, args=(middleware,), name='retrieve_zfs_events_thread')
```
william-gr: You need to use set_thread_name from utils.osc

src/middlewared/middlewared/plugins/zfs_/events_linux.py
```diff
@@ -0,0 +1,59 @@
...
+                child_conn.send(event)
+    finally:
+        if zevent_fd is not None:
+            os.close(zevent_fd)
+
+
+def setup_zfs_events_process(middleware):
+    set_thread_name('retrieve_zfs_events_thread')
+    while True:
+        start_time = time.time_ns() / (10 ** 9)
```
themylogin: This is non-monotonic time which can cause issues when system clock is adjusted. We need other way to solve this issue. What "old" events are coming in?

sonicaj: So basically when we start reading events with opening a fd to /dev/zfs, we start from the first event zfs has logged and then complete the trail of events coming to a blocking call where we wait for new events to occur. This means that if a scrub was run before, we would have old events of scrub if the thread starts after that scrub and we need to skip those old events and only consider new events given to us.

PR #4122, opened 2020-01-27
NAS-104773 / 11.3 / Provide more accurate controller statuses for alert e-mails

themylogin:


src/middlewared/middlewared/plugins/alert.py
```diff
@@ -279,20 +279,25 @@ def __init__(self, middleware):
...
+                        'A': 'Active Controller (A)',
+                        'B': 'Standby Controller (B)',
+                    }
+                else:
+                    nodes = {
+                        'A': 'Standby Controller (A)',
+                        'B': 'Active Controller (B)',
+                    }
+            else:
+                nodes[node] = f'{status.title()} Controller ({node})'
```
william-gr: That would give us "Master Controller" and "Backup Controller"?

Is that really what we want?

william-gr: @themylogin ping

william-gr: In that case seems we should just say "Controller A/B" ?

themylogin: I agree it will sound less confusing for user but I don't want to hide from us what was the failover state when this happened and make debugging harder if not impossible. This will be (should be) a rare case when user sees this and when he sees this I want to have at least some clue on what's going on.

william-gr: Fair enough

PR #4133, opened 2020-01-28
NAS-104281 / 12.0 / Add explicit mappings for BUILTIN\Administrators, Users, and Guests

anodos325:
Add flag to indicate whether a user or group is an SMB one.
Create explicit mappings for Windows BUILTIN groups Administrators, Users, and Guests.


src/middlewared/middlewared/alembic/versions/12.0/2020-01-28_00-31_add-smb-flag-to-accounts.py
```diff
@@ -0,0 +1,49 @@
...
+    op.execute("UPDATE account_bsdgroups SET bsdgrp_smb = 0 WHERE bsdgrp_builtin = 1")    
+    op.execute("UPDATE account_bsdusers SET bsdusr_smb = 0 WHERE bsdusr_builtin = 1")    
+
+    with op.batch_alter_table('account_bsdgroups', schema=None) as batch_op:
+        batch_op.alter_column('bsdgrp_smb', nullable=False)
+
+    with op.batch_alter_table('account_bsdusers', schema=None) as batch_op:
+        batch_op.alter_column('bsdusr_smb', nullable=False)
+
+    op.execute("INSERT INTO account_bsdgroups "
```
themylogin: As 12.0 is cross-platform, we'll do it a different way there. Please see `src/middlewared/middlewared/assets/account/builtin`

PR #4142, opened 2020-01-29
NAS-104863 / 12.0 / fix(middlewared/usage): only count STRIPE if we are sure

william-gr:


william-gr:
We dont support different vdev types, if it is different somehow, then we dont care

PR #4143, opened 2020-01-29
NAS-104846 / 12.0 / Gracefully handle case where plugin repository might be misconfigured

sonicaj:
This PR introduces changes where we gracefully handle the case when plugin repository is misconfigured and we don't either have the INDEX or plugin's manifest file.

### DIFF
--- a/src/middlewared/middlewared/plugins/jail_freebsd.py
+++ b/src/middlewared/middlewared/plugins/jail_freebsd.py
@@ -415,7 +415,9 @@ class PluginService(CRUDService):
                 )
             except Exception as e:
                 resource_list = []
-                self.middleware.logger.debug(f'Failed to retrieve plugins for {options["plugin_repository"]}: {e}')
+                self.middleware.logger.debug(
+                    'Failed to retrieve plugins for %s: %s', options['plugin_repository'], str(e)
+                )
 
         for plugin in resource_list:
             plugin.update({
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/jail_freebsd.py
```diff
@@ -409,9 +409,13 @@ def available(self, job, options):
...
-            resource_list = ioc.IOCage(skip_jails=True).fetch(
-                list=True, plugins=True, header=False, branch=branch, git_repository=options['plugin_repository']
-            )
+            try:
+                resource_list = ioc.IOCage(skip_jails=True).fetch(
+                    list=True, plugins=True, header=False, branch=branch, git_repository=options['plugin_repository']
+                )
+            except Exception as e:
+                resource_list = []
+                self.middleware.logger.debug(f'Failed to retrieve plugins for {options["plugin_repository"]}: {e}')
```
william-gr: for %s: %s

src/middlewared/middlewared/plugins/jail_freebsd.py
```diff
@@ -455,7 +459,10 @@ def defaults(self, options):
             index = plugins_obj.retrieve_plugin_index_data(plugins_obj.git_destination)
 
         if options['plugin'] not in index:
-            raise CallError(f'{options["plugin"]} not found')
+            raise CallError(
+                f'{options["plugin"]} not found, possible reasons are plugin repository\'s '
+                f'INDEX file is missing or {options["plugin"]} plugin\'s manifest file might be missing.'
```
william-gr: Too technical error message for end user. Could we say something along the lines of plugin repository is corrupted? What do you think?

PR #4148, opened 2020-01-30
NAS-104879 / 11.3 / Exclude unknown encytypes from ktutil choices

anodos325:
Unknown enctype may break processing of ktutil output.


src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -752,6 +752,10 @@ class Config:
             for line in kt_list_output.splitlines():
                 fields = line.split()
                 if len(fields) >= 4 and fields[0] != 'Vno':
+                    if fields['1'] == 'unknown':
```
william-gr: Err, after I merged I saw this: fields['1'], with 1 as string

william-gr: @anodos325

PR #4156, opened 2020-01-30
NAS-104802 / 11.3 / Provide more useful text output for winbind errors

anodos325:


src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1181,7 +1181,16 @@ def validate_domain(self, data=None):
...
+            wberr = netlogon_ping.stderr.decode().strip('\n')
+            err = errno.EFAULT
+            for wb in WBCErr:
+                if wb.err() in wberr:
+                    humanized_text = wb.value[0]
+                    wberr.replace(wb.err(), humanized_text)
+                    err = wb.value[1]
+                    break
+
+            raise CallError(wberr.replace(wberr, humanized_text), err)
```
william-gr: humanized_test could somehow be undefined if the condition is never true. Seems we should better be safe than sorry

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1181,7 +1181,16 @@ def validate_domain(self, data=None):
 
         netlogon_ping = await run([SMBCmd.WBINFO.value, '-P'], check=False)
         if netlogon_ping.returncode != 0:
-            raise CallError(netlogon_ping.stderr.decode().strip('\n'))
+            wberr = netlogon_ping.stderr.decode().strip('\n')
+            err = errno.EFAULT
+            for wb in WBCErr:
+                if wb.err() in wberr:
+                    humanized_text = wb.value[0]
+                    wberr.replace(wb.err(), humanized_text)
```
william-gr: Did you mean `wberr = wberr.replace(wb.err(), humanized_text)` ?

anodos325: You caught me before I finished refactoring :)

PR #4157, opened 2020-01-30
NAS-104853 / 12.0 / Fix looking up iscsi authentication group not being found

sonicaj:
This commit fixes an issue where we failed validation by trying to look up authentication groups via id whereas the consumer expects it to use via tag


william-gr:
@sonicaj this is the PR i meant if we should backport

sonicaj:
Yup we do want to backport this one ;)

PR #4181, opened 2020-02-04
NAS-104437 / 11.2 / fix dangling sockets that are not being closed

yocalebo:
There will be an associated change in the truenas repo since I changed the `Client` keyword argument name. The good news is that there is only 1 consumer that uses that parameter.

### DIFF
--- a/src/middlewared/middlewared/client/client.py
+++ b/src/middlewared/middlewared/client/client.py
@@ -4,18 +4,14 @@ from .utils import ProgressBar
 from collections import defaultdict, namedtuple, Callable
 from threading import Event as TEvent, Lock, Thread
 from ws4py.client.threadedclient import WebSocketClient
-from ws4py.websocket import WebSocket
 
 import argparse
-from base64 import b64decode, b64encode
-import ctypes
+from base64 import b64decode
 import errno
 import os
 import pickle
 import socket
-import ssl
 import sys
-import threading
 import time
 import uuid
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/client/client.py
```diff
@@ -48,126 +44,55 @@ def wait(self, timeout=None):
...
-                fd = libc.rresvport(pport)
-                os.close(oldfd)
-                retries -= 1
+        host = ''
+        try:
+            with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
+                s.connect(('169.254.10.1', 80))
+                host = s.getsockname()[0]
+        except Exception:
+            pass
```
freqlabs: Seems dubious to `pass` here.

yocalebo: I pass there under the premise that we don't have to have the local IP address for which we `.bind()` to as it will be built based on the URI parameter that get's called in the failover plugin. What would you suggest we do in that scenario? Log it, maybe?

freqlabs: Oh I actually missed that you have `host=''` above this.
```python
import socket
s = socket.socket()
s.bind(('', 123))
s.connect(('localhost', 22))
while True:
    pass
```
```
# bind.py &
# sockstat | grep 123   
root     python3.6  15498 3  tcp4   127.0.0.1:123         127.0.0.1:22
?        ?          ?     ?  tcp4   127.0.0.1:22          127.0.0.1:123
```
I just tested the above code which shows we can bind to a wildcard address and only set a specific port. The kernel will pick the correct source address when we connect based on the address we connect to. So we don't benefit from this extra socket work here.

yocalebo: Even better!

src/middlewared/middlewared/client/client.py
```diff
@@ -48,126 +44,44 @@ def wait(self, timeout=None):
...
-                os.close(oldfd)
-                retries -= 1
+        if self.privileged_ports_blacklist is None:
+            self.privileged_ports_blacklist = [860]
+        else:
+            self.privileged_ports_blacklist = list(set(self.privileged_ports_blacklist + [860]))
+
+        for port in (x for x in range(600, 1024) if x not in self.privileged_ports_blacklist):
+            try:
+                self.sock.bind((host, port))
```
freqlabs:
```suggestion
                self.sock.bind(('', port))
```

yocalebo: missed it in my previous commit. thanks for catching it

yocalebo:
Unfortunately, this does not fix the dangling sockets and the memory leak still exists. I'm going to close this PR but open a fresh one with a correct title.

PR #4182, opened 2020-02-04
NAS-104363 / 12.0 / Platform dependent debug

sonicaj:


src/freenas/usr/local/libexec/freenas-debug/network/network.sh
```diff
@@ -51,21 +60,43 @@ network_func()
...
 		echo
 
-		if $(ifconfig ${i}|grep -q '\bUP\b')
+		if [ "$iface_up" -eq 0 ];
 		then
-			ips=$(ifconfig ${i}|grep '\binet\b'|awk '{ print $2 }'|xargs)
-			ips6=$(ifconfig ${i}|grep '\binet6\b'|awk '{ print $2 }'|xargs)
+			if [ "$is_linux" -eq 0 ]; then
+				ips=$(ip address show dev ${i} | grep '\binet\b' | awk '{ print $2 }'| xargs)
+				ips6=$(ip address show dev ${i} | grep '\binet6\b' | awk '{ print $2 }'| xarg)
```
yocalebo:
```suggestion
				ips6=$(ip address show dev ${i} | grep '\binet6\b' | awk '{ print $2 }'| xargs)
```

src/freenas/usr/local/bin/ixdiagnose
```diff
@@ -55,7 +59,8 @@ dumpfile_to_submit_exists ()
 truncate_files ()
 {
 	local limit=$1
-	local IFS=$'\n'
+	local IFS='
```
themylogin: Can you please elaborate?

src/freenas/usr/local/bin/ixdiagnose
```diff
@@ -80,14 +85,17 @@ truncate_files ()
 	done
 }
 
-test -e /etc/rc.conf && . /etc/rc.conf
-[ "$dumpdir" = "" ] && dumpdir="$dumpdir"
+if [ "$is_linux" -eq 1 ]; then
```
themylogin: We'll be misunderstanding this as "if we are on linux"

src/freenas/usr/local/bin/ixdiagnose
sonicaj: We have different quotes behaviour in linux where value of `ixdiagnose` changes to `ixdiag ose` with n being stripped out. This keeps the IFS set to newline and is consistent with both platforms.

src/freenas/usr/local/bin/ixdiagnose
sonicaj: In the next line i have a comment clarifying that this section is freebsd specific, please advise if you think i should clarify this further somehow. Thank you

freqlabs: We could create a small library to source in scripts, with something along the lines of:
```sh
case $(uname -s) in
FreeBSD)
	is_freebsd=true
	is_linux=false
	;;
Linux)
	is_freebsd=false
	is_linux=true
	;;
esac

is_freebsd() {
	$is_freebsd
}

is_linux() {
	$is_linux
}
```

That will give us easy to use functions:
```sh
source /etc/system # or whatever the lib gets called

if is_freebsd; then
	do_freebsd_things
elif is_linux; then
	do_linux_things
fi
```

This is what has been done in the ZFS test suite to support multiple platforms.

sonicaj: Thank you @freqlabs, i have updated the code base to use such a script.

src/freenas/etc/system_info
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
```
william-gr: Not sure how useful it would be to have a system wide shell library like this, but if we choose this path, it should be somewhere inside /usr/local/share, not /etc, IMO.

src/freenas/usr/local/bin/ixdiagnose
```diff
@@ -80,14 +84,17 @@ truncate_files ()
 	done
 }
 
-test -e /etc/rc.conf && . /etc/rc.conf
-[ "$dumpdir" = "" ] && dumpdir="$dumpdir"
+if is_linux; then
```
freqlabs:
```suggestion
if is_freebsd; then
```

src/freenas/usr/local/bin/ixdiagnose
```diff
@@ -138,29 +145,41 @@ else
...
+tar -c${h_args}f - -C /var log | tar -C $dir -xf -
+tar -c${h_args}f - -C /var/tmp fndebug | tar -C $dir -xf -
 if [ "${limit}" != "-1" ] ; then
 	truncate_files "${limit}"
 fi
 
-if ! dumpfile_to_submit_exists $dumpdir ; then
-	if [ $? = 2 ] ; then
-		exit 2
+if is_linux; then
```
freqlabs:
```suggestion
if is_freebsd; then
```

src/freenas/usr/local/bin/ixdiagnose
```diff
@@ -179,20 +198,23 @@ if [ -f /etc/version ] ; then
...
-set +e
-ls /dev/ses* 2> /dev/null | while read -r ses; do
-	name=$(echo ${ses}|cut -d "/" -f 3)
-	/sbin/camcontrol smpphylist ${ses} > "$dir/smpphylist.${name}"
-	/usr/local/bin/smp_rep_phy_err_log ${ses} > "$dir/smp_rep_phy_err_log.${name}"
-	cat $dir/smpphylist.${name} | tail -n +3 | while read -r num val desc; do
-		if [ -n "${desc}" ]; then
-			/usr/local/bin/smp_rep_phy_err_log ${ses} -p ${num} > "$dir/smp_rep_phy_err_log.${name}.${num}"
-		fi
+if is_linux; then
```
freqlabs:
```suggestion
if is_freebsd; then
```

src/freenas/usr/local/bin/ixdiagnose
```diff
@@ -138,29 +145,41 @@ else
 	$mydir/freenas-debug -A
 fi
 
+if is_linux; then
+	h_args="h"
+else
+	h_args="H"
+fi
```
freqlabs: Maybe we should be using `h` on both if we're going to start doing it on Linux.

src/freenas/usr/local/share/system_info
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+
+case $(uname -s) in
+FreeBSD)
+	is_freebsd=true
+	is_linux=false
+	;;
+Linux)
+	is_linux=true
+	is_freebsd=false
```
freqlabs: nit
```suggestion
	is_freebsd=false
	;;
```

src/freenas/usr/local/bin/ixdiagnose
sonicaj: @freqlabs there shouldn't be any issues ideally. I'm just trying to avoid touching freebsd's original behaviour as much as I can. Do you see any specific reason for us using that ?

freqlabs: Copying the targets of symlinked files instead of just symlinks may be more useful for debugging, and more shared code is a bonus.

PR #4191, opened 2020-02-05
NAS-104968 / 11.2 / simply privileged port logic for local sockets on TrueNAS HA

yocalebo:
Jira: https://jira.ixsystems.com/browse/NAS-104968

### DIFF
--- a/src/middlewared/middlewared/client/client.py
+++ b/src/middlewared/middlewared/client/client.py
@@ -52,8 +52,8 @@ class WSClient(WebSocketClient):
 
     def __init__(self, url, *args, **kwargs):
         self.client = kwargs.pop('client')
-        self.use_privileged_port = kwargs.pop('use_privileged_port, False')
-        self.privileged_ports_blacklist = kwargs.pop('reserved_ports_blacklist')
+        self.use_privileged_port = kwargs.pop('use_privileged_port', False)
+        self.privileged_ports_blacklist = kwargs.pop('reserved_ports_blacklist', None)
         self.reserved_fd = None
         self.protocol = DDPProtocol(self)
         super(WSClient, self).__init__(url, *args, **kwargs)
--- a/src/middlewared/middlewared/client/client.py
+++ b/src/middlewared/middlewared/client/client.py
@@ -4,18 +4,14 @@ from .utils import ProgressBar
 from collections import defaultdict, namedtuple, Callable
 from threading import Event as TEvent, Lock, Thread
 from ws4py.client.threadedclient import WebSocketClient
-from ws4py.websocket import WebSocket
 
 import argparse
-from base64 import b64decode, b64encode
-import ctypes
+from base64 import b64decode
 import errno
 import os
 import pickle
 import socket
-import ssl
 import sys
-import threading
 import time
 import uuid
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/client/client.py
```diff
@@ -48,126 +44,45 @@ def wait(self, timeout=None):
...
     pass
 
 
 class WSClient(WebSocketClient):
+
     def __init__(self, url, *args, **kwargs):
         self.client = kwargs.pop('client')
-        reserved_ports = kwargs.pop('reserved_ports')
-        reserved_ports_blacklist = kwargs.pop('reserved_ports_blacklist')
+        self.use_privileged_port = kwargs.pop('use_privileged_port, False')
```
themylogin: `kwargs.pop('use_privileged_port', False)`?

src/middlewared/middlewared/client/client.py
```diff
@@ -48,126 +44,45 @@ def wait(self, timeout=None):
...
+        self.privileged_ports_blacklist = kwargs.pop('reserved_ports_blacklist')
```
themylogin: Why does above have the default and this does not?

src/middlewared/middlewared/client/client.py
yocalebo: Yeah, typo. Thanks!

src/middlewared/middlewared/client/client.py
yocalebo: Yep, just missed it. Thanks!

src/middlewared/middlewared/client/client.py
```diff
@@ -48,126 +44,45 @@ def wait(self, timeout=None):
...
-            raise ReserveFDException()
-        return fd
+            return
+
+        raise BindToPrivilegedPortFailed()
 
     def connect(self):
         self.sock.settimeout(10)
+
+        if self.used_privileged_port:
```
freqlabs:
```suggestion
        if self.use_privileged_port:
```

freqlabs:
Can you get rid of https://github.com/freenas/freenas/pull/4191/files#diff-55e798ad4bfe395d1e2d66159add11c8R220-R223 too since we're not needing the special `__del__` method anymore.

yocalebo:
@freqlabs yeah, will do. However, I'm going to raise the Exception still. Just remove the `del` call.

freqlabs:
The exception will still be raised, you don't have to catch it and reraise it.

freqlabs:
Thanks

src/middlewared/middlewared/client/client.py
```diff
@@ -48,126 +44,45 @@ def wait(self, timeout=None):
...
-                retries -= 1
+        if self.privileged_ports_blacklist is None:
+            self.privileged_ports_blacklist = [860]
+        else:
+            self.privileged_ports_blacklist = list(set(self.privileged_ports_blacklist + [860]))
+
+        for port in (x for x in range(600, 1024) if x not in self.privileged_ports_blacklist):
+            try:
+                self.sock.bind(('', port))
+            except Exception:
```
william-gr: Exception seems too generic here, what kind of expected exceptions will it raise?

william-gr: Also, does this by default try to reuse ports? (e.g. tcp connections that are closed but still on timeout state)

yocalebo: > Also, does this by default try to reuse ports? (e.g. tcp connections that are closed but still on timeout state)

Great question. In summary, I have not changed what `ws4py` is doing. However, I do believe there is a bug in the `Client\__init__` code in that library. I'll try and explain.

There are 2 scenarios.

1. The standby controller is offline.
--In this scenario, all new connections will try to bind using a local port however the TCP connections never get past the `SYN_SENT` portion of the handshake. So the `TIME_WAIT` state is never reached.

2. The standby controller is online and connections are established and then suddenly goes offline.
--In this scenario, the existing connections will go into `TIMED_WAIT` because the server did not gracefully close the existing connections.
--All new local client connections can and will use the SAME local port, however, it will not work to the same remote connection.
(i.e. bound to 169.254.10.1:600 -> 169.254.10.2:6000 -> TIMED_WAIT
       bound to 169.254.10.1:600 -> 169.254.10.2:6000 -> WILL NOT WORK because it's to the same destination)

Essentially, I've kept the same logic as what was there. However, I do believe that specifying `SO_REUSEADDR` on client based connections is incorrect and is a bug with ws4py. (which can be fixed as well.)

yocalebo: There is a 3rd situation that is a little more subtle, that might even explain some other issues.

3. The standby controllers middlewared process restarts or crashes and then restarts BEFORE the `TIMED_WAIT` sockets on the client side close.
--In this scenario, any buffered data in the `TIMED_WAIT` state will push that data to the server and if the client has more than 1 local socket that has buffered data, how does the server know which data to act on in order?

yocalebo:
This is too much change for 11.2. I'll close this and re-open a PR for 11.2 only with minimal changes.

PR #4193, opened 2020-02-06
NAS-104983 / 12.0 / Bug fix for select option

sonicaj:
This commit fixes an issue where we selected fields before going through extending phase while retrieving objects which resulted in extend phase not working for most services as it is dependent on keys which might not be present in the data as it had already been pruned.

### DIFF
--- a/src/middlewared/middlewared/plugins/datastore/read.py
+++ b/src/middlewared/middlewared/plugins/datastore/read.py
@@ -219,7 +219,7 @@ class DatastoreService(Service, FilterMixin, SchemaMixin):
 
     async def _serialize(self, obj, table, aliases, relationships, extend, extend_context, extend_context_value,
                          field_prefix, select):
-        data = self._serialize_row(obj, table, aliases, select)
+        data = self._serialize_row(obj, table, aliases)
         data.update(relationships)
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/datastore/read.py
```diff
@@ -230,18 +230,15 @@ def _get_queryset_joins(self, table):
...
 
     def _strip_prefix(self, k, field_prefix):
         return k[len(field_prefix):] if field_prefix and k.startswith(field_prefix) else k
 
-    def _serialize_row(self, obj, table, aliases, select):
+    def _serialize_row(self, obj, table, aliases):
         data = {}
 
         for column in table.c:
-            if select and column.name not in select:
```
themylogin: We should not remove this as this will skip handling relationships if we are not selecting them

src/middlewared/middlewared/plugins/datastore/read.py
```diff
@@ -230,18 +230,15 @@ def _get_queryset_joins(self, table):
             else:
                 data = await self.middleware.call(extend, data)
 
-        return data
+        return data if not select else {k: v for k, v in data.items() if k in select}
```
themylogin: Let's use "normal" if condition please

sonicaj:
backport

PR #4196, opened 2020-02-07
NAS-104437 / 11.2 / fix asyncio mem leak in _run_in_conn_threadpool method

yocalebo:
@themylogin This is the original asyncio memory leak that was fixed in commit: 969eb2c0db3505008a9d1aa167761f43ab7ff3a7

However, there is a call path where `_run_in_conn_threadpool` gets called which still has the problem.


yocalebo:
@themylogin this commit is wrong. I just realized I'm calling shutdown on all threads in `__threadpool`. I'll close this and reopen with correct commit.

PR #4197, opened 2020-02-07
NAS-104437 / 11.2 / fix asyncio mem leak

yocalebo:


yocalebo:
There are multiple places where we're leaking memory with respect to asyncio and tracebacks. This commit aims to fix all instances by writing a wrapper function and calling it as necessary. This is with @freqlabs advice and input. Thanks!

PR #4201, opened 2020-02-09
NAS-105016 / 12.0 / OpenStack Swift support for cloud sync

JGoutin:
OpenStack Swift support for cloud sync.

### DIFF
--- a/src/middlewared/middlewared/rclone/remote/swift.py
+++ b/src/middlewared/middlewared/rclone/remote/swift.py
@@ -1,0 +1,57 @@
+from middlewared.rclone.base import BaseRcloneRemote
+from middlewared.schema import Str
+
+
+class OpenStackSwiftRcloneRemote(BaseRcloneRemote):
+    name = "OPENSTACK_SWIFT"
+    title = "OpenStack Swift"
+
+    buckets = True
+    bucket_title = "Container"
+
+    fast_list = True
+
+    rclone_type = "swift"
+
+    credentials_schema = [
+        Str("user", required=True,
+            title="User name (OS_USERNAME)"),
+        Str("key", required=True,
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/rclone/remote/swift.py
```diff
@@ -0,0 +1,57 @@
...
+        Str("application_credential_id", default="",
+            title="Application Credential ID (OS_APPLICATION_CREDENTIAL_ID)"),
+        Str("application_credential_name", default="",
+            title="Application Credential Name "
+                  "(OS_APPLICATION_CREDENTIAL_NAME)"),
+        Str("application_credential_secret", default="",
+            title="Application Credential Secret "
+                  "(OS_APPLICATION_CREDENTIAL_SECRET)"),
+        Str("auth_version", default="",
+            title="AuthVersion - set to (1,2,3) if your auth URL has no "
```
themylogin: Have you tested what does `set to (1,2,3)` mean? Set to that exact string (but it is `Type: int`)? Or set to any of these values?

JGoutin: This is the API version used, the value can be `1`, `2` or ` (or `0` by default). This is used only on some clouds.
And yes, [it is a Int](https://rclone.org/swift/#swift-auth-version), I'll fix this.

themylogin:
@JGoutin would you have some time to ensure that I've designed corresponding UI form properly?  https://jira.ixsystems.com/browse/NAS-105021

JGoutin:
> @JGoutin would you have some time to ensure that I've designed corresponding UI form properly? https://jira.ixsystems.com/browse/NAS-105021

This looks good for me. :+1:

PR #4205, opened 2020-02-10
NAS-104993 / 12.0 / Send events on disk change

themylogin:


themylogin:
@william-gr that's not what I had in mind :) fixed

PR #4206, opened 2020-02-10
NAS-104500 / 12.0 / ssh weak_ciphers

themylogin:


themylogin:
@freenas/middleware ping :)

themylogin:
@william-gr multiselect "weak ciphers" widget, I've already created an issue https://jira.ixsystems.com/browse/NAS-105023

src/middlewared/middlewared/plugins/ssh.py
```diff
@@ -66,6 +67,7 @@ def bindiface_choices(self):
         Str('sftp_log_level', enum=["", "QUIET", "FATAL", "ERROR", "INFO", "VERBOSE", "DEBUG", "DEBUG2", "DEBUG3"]),
         Str('sftp_log_facility', enum=["", "DAEMON", "USER", "AUTH", "LOCAL0", "LOCAL1", "LOCAL2", "LOCAL3", "LOCAL4",
                                        "LOCAL5", "LOCAL6", "LOCAL7"]),
+        List('weak_ciphers', items=[Str('cipher', enum=['aes128-cbc', 'NONE'])]),
```
william-gr: AES181-CBC, uppercase?

src/middlewared/middlewared/alembic/versions/12.0/2020-02-10_09-47_ssh_weak_ciphers.py
```diff
@@ -0,0 +1,61 @@
...
+revision = '06bfbd354deb'
+down_revision = '4abbf75347b2'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('services_ssh', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('ssh_weak_ciphers', sa.TEXT(), nullable=True))
```
william-gr: Curious why create with nullable True to only make it False later? Why not create it with `[]` as default? Not possible?

themylogin: Possible, but it will have the default value stored in the schema which is not always desirable. I guess this is not the case for NAS and it's just the matter of habit, thank you for reminding.

PR #4208, opened 2020-02-10
NAS-105032 / 11.3 / Provide mechanism for querying all disk temps simultaneously.

anodos325:
Return temperatures for all disks if disk.temperatures is passed an empty list.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk.py
+++ b/src/middlewared/middlewared/plugins/disk.py
@@ -719,6 +719,18 @@ class DiskService(CRUDService):
 
         return None
 
+    @private
+    async def disks_for_temperature_monitoring(self):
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/disk.py
```diff
@@ -728,6 +728,8 @@ def decrypt(self, job, devices, passphrase=None):
         """
         Returns temperatures for a list of device `names` using specified S.M.A.R.T. `powermode`.
         """
+        if len(names) == 0:
+            names = [x['name'] for x in await self.query()]
```
themylogin: Shouldn't this be `disk.disks_for_temperature_monitoring` instead?

PR #4217, opened 2020-02-11
NAS-105068 / 12.0 / Correctly show partition number in freebsd for disks

sonicaj:
Some drives might have partition names like da0s1, this commit fixes an issue where we didn't anticipate this format.


src/middlewared/middlewared/plugins/disk_/disk_info_freebsd.py
```diff
@@ -8,7 +8,7 @@
 
 from .disk_info_base import DiskInfoBase
 
-RE_DISKPART = re.compile(r'^([a-z]+\d+)(p\d+)?')
+RE_DISKPART = re.compile(r'^([a-z]+\d+)([ps]\d+)?')
```
william-gr: Why s? s is supposed to be slice, not GPT partitioned disk

PR #4224, opened 2020-02-13
NAS-104437 / 11.2 / remove ctypes and use setsockopt() in WSClient

yocalebo:
Running this since last night and graphing CLOSED sockets as well as memory usage for middlewared and it has stayed < 200MB of resident memory.

This keeps same functionality as the ctype method.

### DIFF
--- a/src/middlewared/middlewared/client/client.py
+++ b/src/middlewared/middlewared/client/client.py
@@ -67,7 +67,6 @@ class WSClient(WebSocketClient):
         for retry in range(5):
             self.sock.setsockopt(socket.IPPROTO_IP, IP_PORTRANGE, IP_PORTRANGE_LOW)
             self.sock.bind(('', 0))
-            _host, port = self.sock.getsockname()
 
             # The old socket can't be closed before we bind the new socket or
             # we have the possibility of binding to the same port.
--- a/src/middlewared/middlewared/client/client.py
+++ b/src/middlewared/middlewared/client/client.py
@@ -66,9 +66,15 @@ class WSClient(WebSocketClient):
 
         oldsock = None
 
-        for retry in range(5):
+        n_retries = 5
+        for retry in range(n_retries):
             self.sock.setsockopt(socket.IPPROTO_IP, IP_PORTRANGE, IP_PORTRANGE_LOW)
-            self.sock.bind(('', 0))
+
+            try:
+                self.sock.bind(('', 0))
+            except OSError:
+                time.sleep(0.1)
+                continue
 
             # The old socket can't be closed before we bind the new socket or
             # we have the possibility of binding to the same port.
@@ -82,7 +88,7 @@ class WSClient(WebSocketClient):
             # If we're at last pass in loop and get here, break out
             # so we don't set up a socket just to close it essentially
             # making it a NO-OP.
-            if retry == 4:
+            if retry == n_retries - 1:
                 break
 
             oldsock = self.sock
--- a/src/middlewared/middlewared/client/client.py
+++ b/src/middlewared/middlewared/client/client.py
@@ -77,14 +77,17 @@ class WSClient(WebSocketClient):
             if port not in self.reserved_ports_blacklist:
                 return
 
+            # If we're at last pass in loop and get here, break out
+            # so we don't set up a socket just to close it essentially
+            # making it a NO-OP.
+            if retry == 4:
+                break
+
             oldsock = self.sock
 
             self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/client/client.py
```diff
@@ -55,110 +51,42 @@ class ReserveFDException(Exception):
...
-                retries -= 1
-                continue
-            else:
-                break
-        if fd < 0:
-            raise ReserveFDException()
-        return fd
+        while retries > 0:
+            self.sock.setsockopt(socket.IPPROTO_IP, IP_PORTRANGE, IP_PORTRANGE_LOW)
+            _host, port = self.sock.getsockname()
```
freqlabs: We need to actually bind the socket before this.

src/middlewared/middlewared/client/client.py
```diff
@@ -55,110 +51,42 @@ class ReserveFDException(Exception):
...
+
+            if port not in self.reserved_ports_blacklist:
+                return
+
+            newsock = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
+            newsock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+            newsock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            self.sock.close()
```
freqlabs: And this close has to happen after we bind the new socket.

src/middlewared/middlewared/client/client.py
yocalebo: Good catch, thanks!

src/middlewared/middlewared/client/client.py
```diff
@@ -55,110 +51,49 @@ class ReserveFDException(Exception):
...
+            _host, port = self.sock.getsockname()
+
+            # close here because we previously bind() to a port in blacklist
+            if retry > 0:
+                oldsock.close()
+
+            _host, port = self.sock.getsockname()
+            if port not in self.reserved_ports_blacklist:
+                return
+
```
william-gr: If we are in the last iteration of the loop, running the lines below would be a no-op, no?

yocalebo: You're right. I'll fix.

src/middlewared/middlewared/client/client.py
```diff
@@ -55,110 +51,54 @@ class ReserveFDException(Exception):
...
+        if self.reserved_ports_blacklist is None:
+            self.reserved_ports_blacklist = []
+
+        # defined in net/in.h
+        IP_PORTRANGE = 19
+        IP_PORTRANGE_LOW = 2
+
+        oldsock = None
+
+        for retry in range(5):
```
william-gr: Minor nit but just to be more future-proof, can we use n_retries = 5?

Then we can use `if retry == n_reties - 1:` down below

That is if we ever change retry number we wont end up forgetting the stop condition and its also clearer what it means

src/middlewared/middlewared/client/client.py
```diff
@@ -55,110 +51,54 @@ class ReserveFDException(Exception):
...
+            self.sock.setsockopt(socket.IPPROTO_IP, IP_PORTRANGE, IP_PORTRANGE_LOW)
+            self.sock.bind(('', 0))
```
william-gr: How will this fail if we cant allocate more ports, e.g. all taken? Will it raise an exception?

src/middlewared/middlewared/client/client.py
yocalebo: I agree. I'll fix it!

src/middlewared/middlewared/client/client.py
```diff
@@ -66,9 +66,15 @@ def get_reserved_portfd(self):
...
-        for retry in range(5):
+        n_retries = 5
+        for retry in range(n_retries):
             self.sock.setsockopt(socket.IPPROTO_IP, IP_PORTRANGE, IP_PORTRANGE_LOW)
-            self.sock.bind(('', 0))
+
+            try:
+                self.sock.bind(('', 0))
+            except OSError:
+                time.sleep(0.1)
```
freqlabs: Good catch

src/middlewared/middlewared/client/client.py
```diff
@@ -82,7 +88,7 @@ def get_reserved_portfd(self):
             # If we're at last pass in loop and get here, break out
             # so we don't set up a socket just to close it essentially
             # making it a NO-OP.
-            if retry == 4:
+            if retry == n_retries - 1:
```
freqlabs: Good

PR #4230, opened 2020-02-15
Fix problem related with add/update periodic snapshots for multiple days (API)

bastiao:
Fix problem related with add/update periodic snapshots for multiple days only in API v1.


themylogin:
Actually, that's an incorrect approach to this problem. @bastiao you were just using API the wrong way, we'll update the docs

PR #4233, opened 2020-02-17
NAS-105100 / 11.2 / fix WSClient when ALUA is enabled

yocalebo:
1. Port 860 isn't the ALUA port and hasn't been since the addition of binding to reserved ports in 2017
2. If ALUA is enabled, I lessen the available reserved ports by 26. I believe this to be an acceptable trade-off because of the lessening of the technical debt in `WSClient`


yocalebo:
I've pushed a commit in TrueNAS repo to keep code churn on the 11.2 branch as minimal as possible. I'll push this fix to 11.3 and 12.0 respectively.

PR #4237, opened 2020-02-17
Verbose logging for TN Scale

sonicaj:
This PR introduces changes where we make TN Scale logs more verbose wrt retrieving disks and creating pools.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -624,11 +624,6 @@ class PoolService(CRUDService):
         try:
             job.set_progress(90, 'Creating ZFS Pool')
 
-            if IS_LINUX:
-                self.middleware.logger.debug(
-                    'Pool %s is being created with topology %s', data['name'], data['topology'],
-                )
-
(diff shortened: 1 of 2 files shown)


sonicaj:
For disks, we are seeing a case where potentially the disk name is a malformed string which shouldn't be possible ideally. Because of this, when we try to do a low level probe, it fails. For pool creation we are getting a random error `No such pool or dataset`. I am unable to reproduce them so far but they do happen once in a while, so this is to get more insight in the context of the issue.

william-gr:
Perhaps we should only log this in case create fails?

sonicaj:
That’s a very good idea. I’ll make amends

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -670,6 +671,10 @@ def pool_extend(self, pool):
             )
         except Exception as e:
             # Something wrong happened, we need to rollback and destroy pool.
+            if IS_LINUX:
+                self.middleware.logger.debug(
+                    'Pool %s is being created with topology %s', data['name'], data['topology'],
```
william-gr: You mean "pool %s failed to create..." ?

PR #4239, opened 2020-02-19
NAS-104841 / 12.0 / Better information on which services will be restarted and which will…

themylogin:
… be stopped during pool export

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -3403,15 +3403,22 @@ class PoolDatasetService(CRUDService):
         return result
 
     @private
-    async def kill_processes(self, oid, restart_services, max_tries=5):
-        manually_restart_services = []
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3403,15 +3403,21 @@ def __attachments_path(self, dataset):
...
+            if service is not None:
+                if any(attachment_delegate.service == service for attachment_delegate in self.attachment_delegates):
+                    need_restart_services.append(service)
+                else:
+                    need_stop_services.append(service)
+        if (need_restart_services or need_stop_services) and not control_services:
+            raise CallError('Some services have open files and need to be restarted or stopped', errno.EBUSY, {
+                'code': 'control_services',
+                'restart_services': need_restart_services,
+                'stop_services': need_stop_services,
```
william-gr: Can we keep `services` key combining both for backward compatibility?

We have frozen 12 target and dont want to keep adding more to the UI pile.

PR #4241, opened 2020-02-19
NAS-105072 / 12.0 / Correctly redirect host port in nginx

sonicaj:
This commit fixes a bug where we redirected to port 80 by default and thus not respecting the port. To be specific, this caused issues when we have port forwarding to FN UI, the forwarded port on the host got emitted by nginx rewrite rule.


william-gr:
@sonicaj i think we forgot to merge this?

PR #4244, opened 2020-02-21
NAS-105010 / 11.3 / fix(cloud sync): Take Snapshot option can not be used for datasets th…

themylogin:
…at have nested datasets

### DIFF
--- a/src/middlewared/middlewared/plugins/cloud_sync.py
+++ b/src/middlewared/middlewared/plugins/cloud_sync.py
@@ -748,6 +748,11 @@ class CloudSyncService(CRUDService):
                 verrors.add(f"{name}.snapshot", "This option can only be enabled for PUSH tasks")
             if data["transfer_mode"] == "MOVE":
                 verrors.add(f"{name}.snapshot", "This option can not be used for MOVE transfer mode")
+            if await self.middleware.call("pool.dataset.query",
+                                          [["name", "^", os.path.relpath(data["path"], "/mnt") + "/"],
+                                           ["type", "=", "FILESYSTEM"]]):
+                verrors.add(f"{name}.snapshot", "This option is only available for datasets that have no further "
+                                                "nesting")
 
     @private
     async def _validate_folder(self, verrors, name, data):


src/middlewared/middlewared/plugins/cloud_sync.py
```diff
@@ -748,6 +748,10 @@ class Config:
                 verrors.add(f"{name}.snapshot", "This option can only be enabled for PUSH tasks")
             if data["transfer_mode"] == "MOVE":
                 verrors.add(f"{name}.snapshot", "This option can not be used for MOVE transfer mode")
+            if await self.middleware.call("pool.dataset.query",
+                                          [["name", "^", os.path.relpath(data["path"], "/mnt") + "/"],
+                                           ["type", "=", "FILESYSTEM"]]):
+                verrors.add(f"{name}.snapshot", "This option can not be used for datasets that have nested datasets")
```
william-gr: "datasets that have nested datasets" looks odd at first glance...

lets get opinion from @freenas/docs-team

Mrt134: Couple of refinement ideas on either side of positive/negative logic:
"This option is only available for non-parent datasets."
"This option is not available for parent datasets."

themylogin: Can "parent dataset" really be used to say "a dataset that has children"?

william-gr: I think so, but I will leave it up to @freenas/docs-team / @Mrt134

william-gr: @freenas/docs-team  ping

Mrt134: "parent" dataset was the term that appeared the most when referring to similar directory structuring, so I think it'll be largely understood.

Mrt134: I bounced it back and forth with Jaron a bit, and it seems "parent/child" terminology might not be immediately understood. How about `This option is only available for datasets that have no further nesting.` ?

william-gr: @themylogin

PR #4248, opened 2020-02-24
NAS-105194 / 11.3 / Fix ix-update

themylogin:


william-gr:
@themylogin is there a ticket for this?

themylogin:
@william-gr no, I've just seen this error when booting

PR #4251, opened 2020-02-24
NAS-104840 / 12.0 / WebDAVFSAttachmentDelegate

themylogin:


src/middlewared/middlewared/plugins/webdav.py
```diff
@@ -271,5 +275,36 @@ class Config:
...
+        return results
+
+    async def get_attachment_name(self, attachment):
+        return attachment['name']
+
+    async def delete(self, attachments):
+        for attachment in attachments:
+            await self.middleware.call('datastore.delete', 'sharing.webdav_share', attachment['id'])
+
+        await self._service_change('webdav', 'reload')
```
william-gr: Wouldnt we have to stop in this case?

src/middlewared/middlewared/plugins/webdav.py
```diff
@@ -271,5 +275,36 @@ class Config:
...
+
+    async def toggle(self, attachments, enabled):
+        for attachment in attachments:
+            await self.middleware.call('datastore.update', 'sharing.webdav_share', attachment['id'],
+                                       {'webdav_enabled': enabled})
+
+        await self._service_change('webdav', 'reload')
```
william-gr: Shouldnt we start or stop depending on enabled?

src/middlewared/middlewared/plugins/webdav.py
themylogin: We don't stop services when exporting pools because they could be serving other pools.

william-gr: Ah, of course, thanks for clarifying

PR #4254, opened 2020-02-25
NAS-104326 / 12.0 / Platform dependent iSCSI plugin

sonicaj:
This PR introduces changes to use SCST for making iSCSI plugin platform dependent for TN Scale


src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -509,11 +380,38 @@ async def do_delete(self, id):
         """
         Delete iSCSI Authorized Access of `id`.
         """
-        await self._get_instance(id)
+        config = await self._get_instance(id)
+        if not await self.query([['tag', '=', config['tag'], ['id', '!=', id]]]):
+            usages = await self.is_in_use_by_portals_targets(id)
+            if usages['in_use']:
+                raise CallError(f'Authorized access "{id}" is in use by following object(s): {usages["usages"]}')
```
themylogin, line 387: Won't `usages["usages"]` be self-sufficient string here?

src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -571,8 +469,8 @@ class iSCSITargetExtentModel(sa.Model):
     iscsi_target_extent_xen = sa.Column(sa.Boolean(), default=False)
     iscsi_target_extent_rpm = sa.Column(sa.String(20), default='SSD')
     iscsi_target_extent_ro = sa.Column(sa.Boolean(), default=False)
-    iscsi_target_extent_legacy = sa.Column(sa.Boolean(), default=False)
     iscsi_target_extent_enabled = sa.Column(sa.Boolean(), default=True)
+    iscsi_target_extent_vendor = sa.Column(sa.Text(), default='TrueNAS')
```
themylogin, line 473: I've migrated those `default=` from Django because I did not have time back then to figure out what defaults were not set by middleware plugins causing NULL errors. For new fields we should always have defaults in plugin and not rely on model in setting them.

src/middlewared/middlewared/etc_files/scst.conf.mako
```diff
@@ -0,0 +1,120 @@
+<%
+	import os
```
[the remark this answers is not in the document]
sonicaj: Yup, I thought that we are good with using tabs/spaces but just need to be consistent ? I can change to spaces though if desired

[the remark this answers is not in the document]
themylogin: My impression is that we only have tabs in legacy bash scripts, Makefiles and stuff like that. It's the first python file in which I see tabs, and 8 characters wide indents in GitHub reviews looks weird, not speaking about formatting issues we'll get if one day we decide to move code to/from this file..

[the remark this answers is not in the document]
william-gr: I agree, we should unify to spaces for everything inside middlewared/

No need to retroactively change it now though

src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -1660,7 +1580,8 @@ async def delete(self, attachments):
 
         await self._service_change('iscsitarget', 'reload')
 
-        for lun_id in lun_ids:
+        # For SCALE, reload action will remove existing LUN(s)
+        for lun_id in (lun_ids if osc.IS_FREEBSD else ()):
             await run(['ctladm', 'remove', '-b', 'block', '-l', str(lun_id)], check=False)
```
william-gr, line 1585: I has been discussed with mav that this should not be necessary, right @themylogin ?

Shall we remove this block

themylogin: @william-gr we still need this block in order to wait for LUN(s) to be removed

william-gr: @themylogin that already happens on reload, so this block is actually prone to race condition so I dotn see the point in keeping it here.

We should add a sleep after reload at least until there are events from OS.

src/middlewared/middlewared/plugins/service_/services/iscsitarget.py
```diff
@@ -23,3 +23,11 @@ async def before_stop(self):
         if osc.IS_FREEBSD:
             with contextlib.suppress(IndexError):
                 sysctl.filter("kern.cam.ctl.ha_peer")[0].value = ""
```
william-gr, line 25: isnt this broken in sysctl module and should use subprocess?

themylogin: Why? I don't remember this

sonicaj: I think it should be, we changed that usage in ctld.py, but interestingly I think stopping iscsi service works, confirming.

sonicaj: It doesn't, that sysctl I think is set/exists on HA systems so that raised an index error and I thought it worked. I'll correct this usage

src/middlewared/middlewared/plugins/iscsi.py
themylogin: @sonicaj please replace this with `sleep(10)` everywhere except for `terminate_luns_for_pool`

src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -1581,8 +1582,8 @@ async def delete(self, attachments):
         await self._service_change('iscsitarget', 'reload')
 
         # For SCALE, reload action will remove existing LUN(s)
```
themylogin, line 1584: Create dataset on SCALE, create zvol there, mount it via iSCSI, then delete dataset. Won't you get EBUSY because reload does not happen immediately?

sonicaj: No ?

sonicaj: When I delete the dataset, the reload action removes the LUN and the zvol is deleted without any issue.

sonicaj: It would be probably because the `scstadmin` command actually waits for the action to complete ?

themylogin: Then what does "For SCALE, reload action will remove existing LUN(s)" mean if the one that is really doing the job is `scstadmin`?

sonicaj: So `iscsitarget` service reload action has been overloaded to use `scstadmin`, which means that reload of `iscsitarget` will result in LUN being removed via `scstadmin`. Please let me know if you think we should remove that comment ?

PR #4255, opened 2020-02-25
Skip sending update event if row does not exist

sonicaj:
This commit introduces changes where we skip sending update event if the row for which the event is desired has already been removed from the database.


Allison-terminate:
I want out of this project NOW. O keep asking the dumb system how to cancel
my account.

PR #4259, opened 2020-02-26
NAS-105186 / 11.3 / Fix idmap generation for RFC2307 backend

anodos325:


william-gr:
@anodos325 will this need a backport to 12?

PR #4269, opened 2020-02-26
Serial port fixes for TN Scale

sonicaj:


src/middlewared/middlewared/etc_files/grub.py
```diff
@@ -36,7 +36,7 @@ def render(service, middleware):
         default_grub = RE_CMDLINE_LINUX.sub(
             r'\1"root={}{}"'.format(
                 root,
-                f' console=ttyS0,{advanced["serialspeed"]} console=tty0'
+                f' console={advanced["serialport"]},{advanced["serialspeed"]} console=tty1'
```
william-gr: Serial port is supposed to be a hexdecimal address, does this really work?

sonicaj: Yup. I was curious about it as well and if we should be saving port instead of name. But it works as desired. I’ll confirm once more before merging

PR #4272, opened 2020-02-27
NAS-104980 / 12.0 / Use warden jail mac for epair inside jail

sonicaj:
This commit introduces changes where we use warden jail mac for the epair inside the jail instead of the one outside in the host.


sonicaj:
backport

PR #4277, opened 2020-02-28
NAS-105003 / 11.3 / Really fix Content-type E-Mail issue

themylogin:


src/middlewared/middlewared/plugins/mail.py
```diff
@@ -350,7 +350,7 @@ def read_json():
         for key, val in list(extra_headers.items()):
             # We already have "Content-Type: multipart/mixed" and setting "Content-Type: text/plain" like some scripts
             # do will break python e-mail module.
-            if key.lower() == "сontent-type":
+            if key.lower() == "content-type":
```
william-gr: I must be blind, what changed here? I all looks the same in this diff

themylogin: @william-gr it was cyrillic c, don't know why this happened

PR #4305, opened 2020-03-03
Uniform platform detection

themylogin:


src/middlewared/middlewared/plugins/disk_/swap_remove.py
```diff
@@ -1,9 +1,5 @@
-import platform
-
 from middlewared.service import private, Service
-from middlewared.utils import run
-
-IS_LINUX = platform.system().lower() == 'linux'
+from middlewared.utils import osc, run
```
william-gr: osc is not used anywhere here?

PR #4325, opened 2020-03-04
fix(ssh): use openssh service (from ports), not builtin sshd

william-gr:


src/middlewared/middlewared/plugins/service_/services/ssh.py
```diff
@@ -7,7 +7,7 @@ class SSHService(SimpleService):
 
     etc = ["ssh"]
 
-    freebsd_rc = "sshd"
+    freebsd_rc = "openssh"
     freebsd_pidfile = "/var/run/sshd.pid"
```
themylogin: There should also be `freebsd_procname = "sshd"` then

william-gr: Why? Shouldnt the pidfile take precedence?

PR #4328, opened 2020-03-04
NAS-105295 / 11.3 / Bug fixes for swap configuration

sonicaj:
1) Fix race condition where `swaps_configure` can be called when another call to the same endpoint is already executing resulting in undesired behaviour.
2) Do not try to `savecore` for partitions which are already being used in swap.
3) Configure dump device again if current dump device is removed from swap.
4) Fix race condition where sometimes creating a pool would not result in swap being configured as we rely on zfs event to configure swap which is raised as soon as the pool is created. However, during configuration we use database to find out current pools and retrieve valid disks - but if the pool hasn't been saved yet in the database, we don't find those disks resulting in swap not being configured.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk.py
+++ b/src/middlewared/middlewared/plugins/disk.py
@@ -1952,7 +1952,7 @@ async def devd_zfs_hook(middleware, data):
         'misc.fs.zfs.pool_destroy',
         'misc.fs.zfs.pool_import',
     ):
-        asyncio.ensure_future(middleware.call('disk.swaps_configure'))
+        middleware.call('disk.swaps_configure')
 
 
 async def _event_system_ready(middleware, event_type, args):
--- a/src/middlewared/middlewared/plugins/disk.py
+++ b/src/middlewared/middlewared/plugins/disk.py
@@ -1965,7 +1965,6 @@ async def devd_zfs_hook(middleware, data):
     # it in sync every time there is a change.
     if data.get('type') in (
         'misc.fs.zfs.config_sync',
-        'misc.fs.zfs.pool_create',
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/disk.py
```diff
@@ -1469,7 +1471,11 @@ def check_clean(self, disk):
                         # Try to save a core dump from that.
                         # Only try savecore if the partition is not already in use
                         # to avoid errors in the console (#27516)
-                        await run('savecore', '-z', '-m', '5', '/data/crash/', f'/dev/{p.name}', check=False)
+                        cp = await run('savecore', '-z', '-m', '5', '/data/crash/', f'/dev/{p.name}', check=False)
+                        if cp.returncode:
+                            self.middleware.logger.debug(
```
themylogin: Why `debug`?

src/middlewared/middlewared/plugins/disk.py
```diff
@@ -1566,6 +1572,24 @@ def check_clean(self, disk):
...
+        if not os.path.exists('/dev/dumpdev'):
+            try:
+                os.unlink('/dev/dumpdev')
+            except OSError:
+                pass
+            os.symlink(f'/dev/{name}', '/dev/dumpdev')
+            cp = await run('dumpon', f'/dev/{name}')
+            if cp.returncode:
+                self.middleware.logger.debug(
+                    'Failed to specify "%s" device for crash dumps: %s', f'/dev/{name}', cp.stderr.decode()
```
themylogin: Why `debug`?

sonicaj:
backport

PR #4333, opened 2020-03-05
NAS-105207 / 11.3 / Propagate environment update to worker processes

themylogin:


src/middlewared/middlewared/plugins/system.py
```diff
@@ -599,6 +600,24 @@ def debug(self, job):
                 shutil.copyfileobj(f, job.pipes.output.w)
         job.pipes.output.w.close()
 
+    _environ = {}
```
william-gr: Would you agree these are more suited to the core service?

src/middlewared/middlewared/worker.py
```diff
@@ -131,6 +133,16 @@ def watch_parent():
...
 
 
+def receive_environ():
+    def callback(*args, **kwargs):
+        environ_update(kwargs['fields'])
+
+    c = Client('ws+unix:///var/run/middlewared-internal.sock', py_exceptions=True)
+    c.subscribe('system.environ', callback)
+
+    environ_update(c.call('system.environ'))
```
william-gr: We dont have any problems of garbage collection getting rid of client as disconnecting do we? Since it should spawn a new thread waiting for responses...

Just making sure

src/middlewared/middlewared/environ.py
```diff
@@ -0,0 +1,19 @@
+import os
```
william-gr: How about we put it inside utils? or common?

Dont feel strongly about it if you dont think its necessary, jsut not sure if I would fit this into the middlewared level of things.

src/middlewared/middlewared/worker.py
themylogin: It should be a permanent connection that lives during worker entire lifetime

PR #4339, opened 2020-03-05
feat(middlewared/etc): use /etc/cron.d instead appending to system crontab

william-gr:


william-gr:
Nevermind the review, there is more work to do here

themylogin:
We **do** want to keep all this /etc/cron.daily/man-db, right?

william-gr:
I imagine so, at least for now

src/middlewared/middlewared/etc_files/cron.py
```diff
@@ -0,0 +1,5 @@
+import shutil
+
+
+def render(service, middleware):
+    shutil.copyfile('/conf/base/etc/crontab.template', '/etc/crontab')
```
themylogin: We can just move this file to template folder and avoid doing this

william-gr: Ah, of course

william-gr:
@themylogin ah, wait, i think my version was the right one. cron(8) says it will reload the files if the mtime of the file has changed. What do you think?

themylogin:
@william-gr when will we need to reload this file, isn't it static?

william-gr:
@themylogin according to the manpage thats the only file that is watched for. So if middlewared cron.d changes alone, it wont reload.

themylogin:
@william-gr have not thought of that. Then the best solution I can see is touching `/etc/crontab` from within `src/middlewared/middlewared/etc_files/crontab` (that way we can keep it being a template, just in case)

william-gr:
I have looked at the source code, and it seems to reload regardless of that, looking for mtime of any cron.d

themylogin:
@william-gr but how does it know when to check if we don't touch /etc/crontab at all?

william-gr:
it only needs to check the first time, which we do. I will check for the mtime of cron.d/middlewared whenever its generated

PR #4341, opened 2020-03-05
Fix integration tests

themylogin:


tests/api2/rsync.py
```diff
@@ -64,7 +64,7 @@ def test_07_Testing_rsync_access():
 
 def test_08_Starting_rsyncd_service():
     results = POST("/service/start/",
-                   {'service': 'rsyncd'}
+                   {'service': 'rsync'}
```
william-gr: Wasnt it named rsyncd before? Why this test started failing now?

themylogin: It was always named `rsync` (surrounding tests refer to it like that). Old API was returning 200 for invalid service name, that's why it was not failing.

william-gr: Got it, thanks

PR #4348, opened 2020-03-06
NAS-105246 / 11.3 / Verify that certificates exist in ldap and ad common_validate()

anodos325:


anodos325:
> cert_services_validation

I like that idea.

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -653,6 +653,10 @@ class Config:
         if not new["enable"]:
             return
 
+        if new["certificate"]:
+            await self.middleware.call("certificate.cert_services_validation",
```
sonicaj: This would be doing 2 things, 
1) It will raise verrors which I think we do not want and would only like after we have validated the bits down as well ? ( We can use verrors.extend here with the verrors being returned from the call )
2) Schema name being passed should be `activatedirectory_update.certificate`

PR #4353, opened 2020-03-06
Add Apache to Scale

william-gr:


william-gr:
Etc plugin is supposed to created dirs

sonicaj:
Would etc plugin be able to find `{APACHE_DIR}/webdav_config` to render ?

william-gr:
Err, I am dumb. Thanks for the sanity check

PR #4359, opened 2020-03-06
Generate smb4.conf if needed during smb_configure

anodos325:
If for some reason /usr/local/etc/smb4.conf is deleted, then
etc.generate smb_configure will fail.

Fix traceback in smb.update when smb.conf has invalid aux parameter.

### DIFF
--- a/src/middlewared/middlewared/etc_files/smb_configure.py
+++ b/src/middlewared/middlewared/etc_files/smb_configure.py
@@ -27,6 +27,9 @@ def get_config(middleware):
     """
     Set basic configuration
     """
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/etc_files/smb_configure.py
```diff
@@ -27,6 +27,9 @@ def get_config(middleware):
     """
     Set basic configuration
     """
+    if not os.path.exists(SMBPath.GLOBALCONF.platform()):
+        middleware.call_sync("etc.generate", "smb")
```
william-gr: @anodos325  Why not put the smb4.conf in the "smb_configure" etc group?

PR #4362, opened 2020-03-09
NAS-105221 / 12.0 / Add Support for Special Allocation Classes

sonicaj:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -466,6 +466,7 @@ def pool_extend(self, pool):
         Str('name', required=True),
         Bool('encryption', default=False),
         Str('deduplication', enum=[None, 'ON', 'VERIFY', 'OFF'], default=None, null=True),
+        Int('special_small_block_size', validators=[Range(min=512)]),
```
william-gr: Is this a create only option?

william-gr: @sonicaj if it isnt, I dont see why we should put this option in pool create?

Why is this any more special that all others properties?

sonicaj: I thought it might be beneficial for the users to specify this for the root dataset in one step of pool creation like we do for `deduplication`, I'll have it removed.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -709,6 +723,7 @@ def pool_extend(self, pool):
         'pool_create', 'pool_update',
         ('rm', {'name': 'name'}),
         ('rm', {'name': 'encryption'}),
+        ('rm', {'name': 'deduplication'}),
```
william-gr: Why this?

sonicaj: It is a dataset attribute and `pool.update` does not do anything with it

PR #4366, opened 2020-03-09
NAS-105339 / 11.2 / fix index out of range

yocalebo:
Jira: https://jira.ixsystems.com/browse/NAS-105339


yocalebo:
@william-gr hmm, not sure what's making the request. Just spotted in the logs. I'll keep investigating.

PR #4379, opened 2020-03-10
NAS-105324 / 11.3 / Increase timeout for grub call to 5 seconds

sonicaj:
It is possible in some cases grub takes a bit more then 2 seconds to start up especially when the system has just booted and is setting up other services.


sonicaj:
Libvirt is doing that for us, I've made a note for myself to test this bit as well when making sure libvirt is a go for 12 for all major platforms

sonicaj:
In short, no, not for now at least ; )

PR #4398, opened 2020-03-12
NAS-105240 / 12.0 / Conform to updated standards of trusted certificates

sonicaj:
This PR introduces changes where we update the default value of `not_valid_after` attribute of a digital certificate conforming to https://support.apple.com/en-us/HT210176.

It should be noted that we allow specifying EKU extension, so certificates should set them themselves based on their usage, i.e `SERVER_AUTH`/`CLIENT_AUTH` as it's required for server certificates now.


sonicaj:
backport

PR #4405, opened 2020-03-12
NAS-105030 / 11.3 / Fix exporting pool with iSCSI shares

themylogin:
@william-gr do we want to put OS team to work instead to figure out what is the proper solution?

### DIFF
--- a/src/middlewared/middlewared/plugins/iscsi.py
+++ b/src/middlewared/middlewared/plugins/iscsi.py
@@ -11,6 +11,7 @@ from middlewared.utils import filter_list, run
 from middlewared.utils.path import is_child
 from middlewared.validators import IpAddress, Range
 
+import asyncio
 import bidict
 import errno
 import hashlib
@@ -1538,8 +1539,8 @@ class ISCSIFSAttachmentDelegate(FSAttachmentDelegate):
 
         await self._service_change('iscsitarget', 'reload')
 
-        for lun_id in lun_ids:
-            await run(['ctladm', 'remove', '-b', 'block', '-l', str(lun_id)], check=False)
+        if lun_ids:
+            await asyncio.sleep(10)  # iscsitarget does not remove luns immediately
 
     async def toggle(self, attachments, enabled):
         lun_ids = []


amotin:
I'd definitely prefer proper fix rather than this bad workaround. I haven't heard about this problem till now. To me it looks like a valid CTL (OS team) bug.  I'd just like to better understand why after doing iscsitarget reload, which should remove the LUNs by itself we are doing that once more via ctladm?  Is there some race or insufficient wait this supposed to close?

themylogin:
@amotin If I have iscsi volume mounted somewhere, remove it from ctl config and do iscsitarget reload, I won't be able to export pool (EBUSY) until I run ctladm remove.

amotin:
I think the problem was misdiagnosed from the beginning.  `service ctld reload` sends SIGHUP to ctld process, making it reload the configuration.  But I suppose it does not wait for completion.  Calling `ctladm remove` explicitly makes you duplicate ctld behavior, and that I guess causes race.  What I think we should do is somehow reliably wait for reload completion.  It would also be good to fix the race causing the panic, but that I believe is a consequence.

william-gr:
@themylogin @amotin so how should we go about this?
How can we reliably wait for reload to complete? I imagine this is not trivial, the easiest way I could think of is start sending devd events when that happens but that seems a stretch of devd functionality.

Should we just try to sleep for a couple seconds for now?

amotin:
Doing sleep before ctladm remove is a workaround for workaround, since the ctladm remove should not be needed there by itself, since ctld reload after updating config should do all by itself.  The only problem I see is that it does not report completion, since signals simply have no return channel.  If ctld reload does not work in some situations, please show me one. I've just tried to simulate it manually on 11.3-U2 and it removed the LUN for me.  If the problem is that deletion takes some time, then just some sleep() without call to ctladm remove should do the trick, though probably unreliable, since I don't remember how long it may take.  Better way would probably be just to wait while LUN disappear from ctladm devlist after reloading ctld.
Alternatively, if there is a wish to fix the present supposed deadlock between ctld reload and ctladm remove, please let me look on some system in that state.

themylogin:
@amotin it happens on HA system, Bonnie was able to reproduce this for me when I worked on this https://jira.ixsystems.com/browse/NAS-105030

william-gr:
@amotin watching ctladm devlist is also not great because its not event driven.

How about using a devd event when ctld finishes reloading? Is that such a bad idea? :P AFAIK its the only way we have a way to report things to userspace?

I suppose we could also just also sleep a couple seconds after service ctld reload? (Never intended to keep it like this PR after what you said)

amotin:
@william-gr ctld runs in user-space. I don't remember I've sent devd messages from user-space to user-space.  Alternatively I think kernel ctl could be made to send devd events on all changes like this.

william-gr:
@amotin so what should we do here?

For 11.3 remove the ctladm remove and add a sleep X after ctld reload?

For 12 or later create an OS ticket to somehow get an event when that happens?

amotin:
So far I've made a patch for CTL among other things fixing the panic.  Going to merge it down after some time.  How to make ctld report reload completion I haven't decided yet.

PR #4408, opened 2020-03-13
NAS-105433 / 12.0 / Restart SMB if required after running smb_configure

anodos325:
If for some reason samba is running when this is called then samba should be restarted.


anodos325:
> Wouldnt that make more sense in the service plugin?
> 
> Feels odd manipulating services within etc plugin

That's a good point. I will move this into the service plugin and maybe get rid of the etc plugin entirely.

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -324,6 +331,60 @@ def getparm(self, parm, section):
...
+            await self.middleware.call("smb.synchronize_group_mappings")
+            await self.middleware.call("admonitor.start")
+            job.set_progress(60, 'generating SMB share configuration.')
+            await self.middleware.call("etc.generate", "smb_share")
+
+        job.set_progress(70, 'Checking SMB server status.')
+        if await self.middleware.call("service.started", "cifs"):
+            job.set_progress(80, 'Restarting SMB service.')
+            await self.middleware.call("service.restart", "cifs")
+        job.set_progress(100, 'Finished configuring SMB.')
```
anodos325: This is probably an excessive amount of progress updates

src/middlewared/middlewared/plugins/smb_/groupmap.py
```diff
@@ -94,3 +95,44 @@ class Config:
...
+        must_remove_cache = False
+        if groupmap:
+            sids_fixed = await self.middleware.call('smb.fixsid', groupmap)
+            if not sids_fixed:
+                groupmap = []
+
+        for b in SMBBuiltin:
+            entry = list(filter(lambda x: b.value[0] == x['unixgroup'], groupmap))
+            if b.name == 'ADMINISTRATORS':
+                if len(await self.middleware.call('group.query', [('gid', '=', 544)])) > 1:
```
william-gr: Not need to change, but you could use {"count": True}

PR #4411, opened 2020-03-16
NAS-100045 / 12.0 / NAS-100045 PCI-E passthru support (libvirt based)

stattin42:
PCI-E passthru support in vm plugin
ok to test

### DIFF
--- a/src/middlewared/middlewared/plugins/vm.py
+++ b/src/middlewared/middlewared/plugins/vm.py
@@ -55,8 +55,8 @@ SHUTDOWN_LOCK = asyncio.Lock()
 LIBVIRT_LOCK = asyncio.Lock()
 LIBVIRT_HOSTDEV = False  # Whether libvirt supports hostdev for bhyve or not
 PPT_BASE_NAME = 'ppt'
-RE_PCICONF_PPTDEVS = re.compile(r'^(' + re.escape(PPT_BASE_NAME) \
-                                + '[0-9]+@pci.*:)(([0-9]+:){2}[0-9]+).*$', flags=re.I)
+RE_PCICONF_PPTDEVS = re.compile(
+    r'^(' + re.escape(PPT_BASE_NAME) + '[0-9]+@pci.*:)(([0-9]+:){2}[0-9]+).*$', flags=re.I)
 ZFS_ARC_MAX_INITIAL = None
 
 ZVOL_CLONE_SUFFIX = '_clone'
--- a/src/middlewared/middlewared/plugins/vm.py
+++ b/src/middlewared/middlewared/plugins/vm.py
@@ -55,7 +55,7 @@ SHUTDOWN_LOCK = asyncio.Lock()
 LIBVIRT_LOCK = asyncio.Lock()
 LIBVIRT_HOSTDEV = False  # Whether libvirt supports hostdev for bhyve or not
 PPT_BASE_NAME = 'ppt'
-RE_PCICONF_PPTDEVS = re.compile(r'^(' + re.escape(PPT_BASE_NAME)
+RE_PCICONF_PPTDEVS = re.compile(r'^(' + re.escape(PPT_BASE_NAME) \
                                 + '[0-9]+@pci.*:)(([0-9]+:){2}[0-9]+).*$', flags=re.I)
 ZFS_ARC_MAX_INITIAL = None
 
--- a/src/middlewared/middlewared/plugins/vm.py
+++ b/src/middlewared/middlewared/plugins/vm.py
@@ -8,7 +8,7 @@ import middlewared.sqlalchemy as sa
 from middlewared.utils import Nid, osc, Popen, run
 from middlewared.utils.asyncio_ import asyncio_map
 from middlewared.utils.path import is_child
-from middlewared.validators import Range
+from middlewared.validators import Range, Match
 
 import middlewared.logger
 import asyncio
@@ -54,8 +54,9 @@ LIBVIRT_BHYVE_NSMAP = {'bhyve': LIBVIRT_BHYVE_NAMESPACE}
 SHUTDOWN_LOCK = asyncio.Lock()
 LIBVIRT_LOCK = asyncio.Lock()
 LIBVIRT_HOSTDEV = False  # Whether libvirt supports hostdev for bhyve or not
-LIBVIRT_BHYVE_NAMESPACE = 'http://libvirt.org/schemas/domain/bhyve/1.0'
-LIBVIRT_BHYVE_NSMAP = {'bhyve': LIBVIRT_BHYVE_NAMESPACE}
+PPT_BASE_NAME = 'ppt'
+RE_PCICONF_PPTDEVS = re.compile(r'^(' + re.escape(PPT_BASE_NAME)
+                                + '[0-9]+@pci.*:)(([0-9]+:){2}[0-9]+).*$', flags=re.I)
 ZFS_ARC_MAX_INITIAL = None
 
 ZVOL_CLONE_SUFFIX = '_clone'
@@ -288,24 +289,16 @@ class VMSupervisor:
         else:
             # Source bus/slot seen before. Reuse the old guest slot.
             guest_slot = item['guest_bsf'][1]
-            # No need to map the same bus/slot/function more than once.
             # Check if the function has already been mapped.
+            # No need to map the same bus/slot/function more than once.
             while (item is not None and item['host_bsf'] != host_bsf):
                 # Not same function. Check next item.
                 item = next(mylist, None)
-            if item is None:
-                # Host bus/slot seen before, but function is new.
-                # Map it on the same guest slot as other functions of same host
-                # bus/slot.
-                pass  # placeholder for logging
         if item is None:
             # This host bus/slot/function is not mapped yet.
             # Add passthru device
             guest_bsf = [0, guest_slot, host_bsf[2]]
             return guest_bsf
-        else:
-            # This host bus/slot/function is already mapped. Do not add it again.
-            pass  # placeholder for logging
         return None
 
     def construct_xml(self):
@@ -372,7 +365,7 @@ class VMSupervisor:
         )
 
     def commandline_xml(self):
-        commandline_args = filter(bool, self.commandline_args())
+        commandline_args = self.commandline_args()
         return [create_element(
             etree.QName(LIBVIRT_BHYVE_NAMESPACE, 'commandline'), attribute_dict={
                 'children': [
@@ -387,7 +380,7 @@ class VMSupervisor:
     def commandline_args(self):
         args = []
         for device in filter(lambda d: isinstance(d, (PCI, VNC)), self.devices):
-            args.append(device.bhyve_args())
+            args += filter(None, [device.bhyve_args()])
         return args
 
     def os_xml(self):
@@ -542,12 +535,16 @@ class VMSupervisor:
                 device_xml = device.xml(slot=pci_slot())
             elif isinstance(device, PCI):
                 # PCI passthru section begins here
-                pptdev = device.data['attributes'].get('pptdev')
-                host_bsf = list(map(int, pptdev.split('/')))
-                guest_bsf = self.guest_pptdev(ppt_maps, pci_slot, host_bsf)
+                # Check if ppt device is available for passthru. Map, only if available.
+                host_bsf = device.ppt_map['host_bsf']
+                if host_bsf is not None:
+                    guest_bsf = self.guest_pptdev(ppt_maps, pci_slot, host_bsf)
+                else:
+                    guest_bsf = None
+                device.ppt_map['guest_bsf'] = guest_bsf
                 if guest_bsf is not None:
-                    ppt_maps.append({'host_bsf': host_bsf, 'guest_bsf': guest_bsf})
-                device_xml = device.xml(host_bsf=host_bsf, guest_bsf=guest_bsf)
+                    ppt_maps.append(device.ppt_map)
+                device_xml = device.xml()
                 # PCI passthru section ends here
             else:
                 device_xml = device.xml()
@@ -678,38 +675,57 @@ class PCI(Device):
 
     schema = Dict(
         'attributes',
-        Str('pptdev', default=None, required=True),
+        Str('pptdev', required=True, validators=[Match(r'([0-9]+/){2}[0-9]+')]),
     )
 
-    def __init__(self, data, middleware=None):
-        Device.__init__(self, data, middleware)
-        self.ppt_map = {'host_bsf': None, 'guest_bsf': None}
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.init_ppt_map()
+
+    def init_ppt_map(self):
+        iommu_type = self.middleware.call_sync('vm.device.get_iommu_type')
+        pptdevs = self.middleware.call_sync('vm.device.pptdev_choices')
+        pptdev = self.data['attributes'].get('pptdev')
+        self.ppt_map = {
+            'host_bsf': list(
+                map(int, pptdev.split('/'))) if pptdev in pptdevs and iommu_type else None,
+            'guest_bsf': None
+        }
 
     def xml(self, *args, **kwargs):
-        host_bsf = kwargs.pop('host_bsf')
-        guest_bsf = kwargs.pop('guest_bsf')
+        if LIBVIRT_HOSTDEV:
+            # Generate xml if libvirt supports hostdev for bhyve.
+            # If passthru is performed by means of additional command-line arguments
+            # to the bhyve process using the <bhyve:commanline> element under domain,
+            # the xml is TYPICALLY not needed. An EXCEPTION is when there are devices
+            # for which the pci address is not under the control of and set by
+            # middleware and generation of the xml can reduce the risk for conflicts.
+            # It appears that when assigning addresses to other devices libvirt avoids
+            # the pci address provided in the xml also when libvirt does not (fully)
+            # support hostdev for bhyve.
+            host_bsf = self.ppt_map['host_bsf']
+            guest_bsf = self.ppt_map['guest_bsf']
 
-        self.ppt_map = {'host_bsf': host_bsf, 'guest_bsf': guest_bsf}
-        return create_element(
-            'hostdev', mode='subsystem', type='pci', managed='no', attribute_dict={
-                'children': [
-                    create_element(
-                        'source', attribute_dict={
-                            'children': [
-                                create_element('address', domain='0x0000',
-                                               bus='0x{:04x}'.format(host_bsf[0]),
-                                               slot='0x{:04x}'.format(host_bsf[1]),
-                                               function='0x{:04x}'.format(host_bsf[2])),
-                            ]
-                        }
-                    ),
-                    create_element('address', type='pci', domain='0x0000',
-                                   bus='0x{:04x}'.format(guest_bsf[0]),
-                                   slot='0x{:04x}'.format(guest_bsf[1]),
-                                   function='0x{:04x}'.format(guest_bsf[2])),
-                ]
-            }
-        ) if guest_bsf is not None else None
+            return create_element(
+                'hostdev', mode='subsystem', type='pci', managed='no', attribute_dict={
+                    'children': [
+                        create_element(
+                            'source', attribute_dict={
+                                'children': [
+                                    create_element('address', domain='0x0000',
+                                                   bus='0x{:04x}'.format(host_bsf[0]),
+                                                   slot='0x{:04x}'.format(host_bsf[1]),
+                                                   function='0x{:04x}'.format(host_bsf[2])),
+                                ]
+                            }
+                        ),
+                        create_element('address', type='pci', domain='0x0000',
+                                       bus='0x{:04x}'.format(guest_bsf[0]),
+                                       slot='0x{:04x}'.format(guest_bsf[1]),
+                                       function='0x{:04x}'.format(guest_bsf[2])),
+                    ]
+                }
+            ) if guest_bsf is not None else None
 
     def bhyve_args(self, *args, **kwargs):
         if not LIBVIRT_HOSTDEV:
@@ -1919,8 +1935,9 @@ class VMDeviceService(CRUDService):
 
     @accepts()
     def pptdev_choices(self):
-        PPT_BASE_NAME = 'ppt'
-
+        """
+        Available choices for PCI passthru device.
+        """
         proc = subprocess.Popen(['/usr/sbin/pciconf', '-l'],
                                 stdout=subprocess.PIPE,
                                 stderr=subprocess.DEVNULL, text=True)
@@ -1933,10 +1950,8 @@ class VMDeviceService(CRUDService):
 
         lines = outs.split('\n')
         pptdevs = {}
-        regexp = r'^(' + re.escape(PPT_BASE_NAME)\
-                 + '[0-9]+@pci.*:)(([0-9]+:){2}[0-9]+).*$'
         for line in lines:
-            object = re.match(regexp, line, flags=re.I)
+            object = RE_PCICONF_PPTDEVS.match(line)
             if object:
                 pptdev = object.group(2).replace(':', '/')
                 pptdevs[pptdev] = pptdev
@@ -2133,7 +2148,7 @@ class VMDeviceService(CRUDService):
             return False
 
     @private
-    async def get_iommu_type(self):
+    def get_iommu_type(self):
         def check_iommu(type):
             IOMMU_TEST = {'VT-d': {'arglist': ['/usr/sbin/acpidump', '-t'],
                                    'string': 'DMAR'},
@@ -2281,17 +2296,10 @@ class VMDeviceService(CRUDService):
                     verrors.add('attributes.nic_attach', 'Not a valid choice.')
             await self.failover_nic_check(device, verrors, 'attributes')
         elif device.get('dtype') == 'PCI':
-            pptdev = device['attributes'].get('pptdev')
-            if not pptdev:
-                verrors.add('attributes.pptdev', 'PPT device is required.')
-            elif re.match(r'([0-9]+/){2}[0-9]+', pptdev) is None:
-                verrors.add('attribute.pptdev', 'PPT device format is invalid. Should be on form bus#/slot#/fcn#.')
-            else:
-                pptdevs = await self.middleware.call('vm.device.pptdev_choices')
-                if pptdev not in pptdevs:
-                    verrors.add('attribute.pptdev', 'Not a valid choice.')
-            IOMMU_TYPE = await self.get_iommu_type()
-            if IOMMU_TYPE is None:
+            if device['attributes'].get('pptdev') not in (
+                    await self.middleware.call('vm.device.pptdev_choices')):
+                verrors.add('attribute.pptdev', 'Not a valid choice. The PCI device is not available for passthru.')
+            if (await self.middleware.call('vm.device.get_iommu_type')) is None:
                 verrors.add('attribute.pptdev', 'IOMMU support is required.')
         elif device.get('dtype') == 'VNC':
             if vm_instance:
--- a/src/middlewared/middlewared/plugins/vm.py
+++ b/src/middlewared/middlewared/plugins/vm.py
@@ -1951,13 +1951,17 @@ class VMDeviceService(CRUDService):
                 self.logger.warn(f'An error occured when checking the PCI configuration '
                                  f'for devices available for PCI passthru. {e}.')
             else:
-                lines = outs.split('\n')
-                for line in lines:
-                    object = RE_PCICONF_PPTDEVS.match(line)
-                    if object:
-                        pptdev = object.group(2).replace(':', '/')
-                        self.pptdevs[pptdev] = pptdev
-
+                if proc.returncode < 1:
+                    lines = outs.split('\n')
+                    for line in lines:
+                        object = RE_PCICONF_PPTDEVS.match(line)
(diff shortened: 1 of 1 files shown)


freqlabs:
As a slight heads-up, I have patches for libvirt to support hostdev pci passthrough for bhyve, but I need to revisit them to address the feedback I received in the process of upstreaming. I'm not sure if that will make it into a version of libvirt that we use for 12.

stattin42:
Thanks @freqlabs. I found some version of your patches when I was looking for information about libvirt, but was not sure about their official status. Also found a pointer, maybe also from you, to use <bhyve:commandline> to pass arguments through to bhyve. Thanks for the hint.

This PR generates XML for hostdev pci passthrough but, as this is not yet fully supported by current version of libvirt, the PR also provides an overlay with <bhyve:commandline> arguments to provide bhyve with the corresponding arguments. I say hostdev pci passthrough is "not yet fully supported" because, while libvirt does not generate the corresponding bhyve arguments, libvirt seems to respect the pci address specification provided by a hostdev element. Thus, this PR relies on hostdev elements for (at least) pci address reservation and then (if needed) uses <bhyve:commandline> for passing the arguments to bhyve.

I have defined LIBVIRT_HOSTDEV to control the behaviour depending on whether libvirt supports hostdev pci passthrough or not. Once hostdev pci passthrough is fully supported by libvirt, the <bhyve:commandline> overlay should (I hope) not be needed anymore.

@freqlabs, where should I look for the latest version of your libvirt patches; to track the progress?

freqlabs:
> @freqlabs, where should I look for the latest version of your libvirt patches; to track the progress?

This branch has the remainder of my patches which have not been upstreamed yet: https://gitlab.com/freqlabs/libvirt/-/tree/freebsd-master
It is the three most recent commits as of now. I will eventually be splitting the commits up further for upstreaming.

william-gr:
Thanks @stattin42 ! @sonicaj will review your code once he has some spare time, as he is the one responsible and wrote most of the vm plugin.

sonicaj:
Thank you for your contribution @stattin42, it's much appreciated. Can you please rebase and update your branch and we can give it a final review before merging ?

stattin42:
Thanks @sonicaj! The PR has been rebased and updated now.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -605,6 +674,55 @@ class RAW(StorageDevice):
...
+    def __init__(self, data, middleware=None):
+        Device.__init__(self, data, middleware)
+        self.ppt_map = {'host_bsf': None, 'guest_bsf': None}
+
+    def xml(self, *args, **kwargs):
+        host_bsf = kwargs.pop('host_bsf')
+        guest_bsf = kwargs.pop('guest_bsf')
+
+        self.ppt_map = {'host_bsf': host_bsf, 'guest_bsf': guest_bsf}
+        return create_element(
```
sonicaj, line 693: If I understand correctly, this is not required right now as PCI currently works with bhyve args with the libvirt version we have. How about we use the `LIBVIRT_HOSTDEV` variable here as well to make sure we don't generate this and add it to the xml even though it won't work and we have to make it work via command line args ?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -269,6 +272,42 @@ def poweroff(self):
...
+            guest_slot = nid()
+        else:
+            # Source bus/slot seen before. Reuse the old guest slot.
+            guest_slot = item['guest_bsf'][1]
+            # No need to map the same bus/slot/function more than once.
+            # Check if the function has already been mapped.
+            while (item is not None and item['host_bsf'] != host_bsf):
+                # Not same function. Check next item.
+                item = next(mylist, None)
+            if item is None:
```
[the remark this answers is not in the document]
stattin42: Comment probably also not needed.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -269,6 +272,42 @@ def poweroff(self):
...
+                # bus/slot.
+                pass  # placeholder for logging
+        if item is None:
+            # This host bus/slot/function is not mapped yet.
+            # Add passthru device
+            guest_bsf = [0, guest_slot, host_bsf[2]]
+            return guest_bsf
+        else:
+            # This host bus/slot/function is already mapped. Do not add it again.
+            pass  # placeholder for logging
```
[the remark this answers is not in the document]
stattin42: Exisitng comments before the else clause probably sufficient.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -308,13 +347,32 @@ def construct_xml(self):
...
+            )
+
         return create_element(
             'domain', type='bhyve', id=str(self.vm_data['id']),
             attribute_dict={'children': domain_children}, nsmap=LIBVIRT_BHYVE_NSMAP,
         )
 
     def commandline_xml(self):
-        commandline_args = self.commandline_args()
+        commandline_args = filter(bool, self.commandline_args())
```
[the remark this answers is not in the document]
stattin42: Thanks for hightlighting the if condition. I had overlooked that.

src/middlewared/middlewared/plugins/vm.py
stattin42: Maybe ok now since the fbuf/vncserver is passed with command-line arg. Before, ordering of devices was not consistent due to fbuf/vncserver not getting a fixed PCI address. I think previously there was also a USB controller 'floating around'; not anymore. Those devices caused PCI address conflicts. Inclusion of the xml also when using bhyve args avoided the conflicts as libvirt seemed to avoid the PCI address provided in the xml for the passthru device when libvirt assigned addresses/slots for other devices.

I have made xml generation conditioned on LIBVIRT_HOSTDEV, but added a comment about potential address conflicts, in case the issue reappears.

stattin42:
You are probably already aware, but just in case, there is an associated PR for the webui [here](https://github.com/freenas/webui/pull/3792).

stattin42:
Thank you @sonicaj for the detailed review and helpful comments. I have tried to address your comments and commited the updates. I have a couple of questions though:

1. Can we assume that slot numbers can and will be managed for all relevant devices such that the xml generation can be safely skipped for PCI passthru devices while the solution with bhyve args is used? See comments above and in the code regarding slot conflicts which I observed when I was working with the first version of the 12.0 PR.

2. Ideally discovery of iommu and pptdevs should only need to be done once at startup as changes to those typically require restart of the host. I am not sure, however, where in the vm plugin would be the best place to do this and then which would be the best way to make this information available to other methods. Do you have any advise or preference?

src/middlewared/middlewared/plugins/vm.py
sonicaj: Thank you for elaborating @stattin42, I see what the issue is. So what happens is that libvirt chooses slots for the controller itself and earlier it was doing that for VNC too. Now moving on, libvirt is still handling that for `tablet`, however I have created a PR to address that ( https://github.com/freenas/freenas/pull/4549 ). So we shouldn't be facing that issue anymore. 

I think it's safe to say that we are good for this. Can you please rebase and test to be certain ?

sonicaj:
@stattin42 for VM's if we want to run some action on boot, we do that here
https://github.com/freenas/freenas/blob/8c25faa4cac2d1aeff926e05e5a1eea1d8756f2d/src/middlewared/middlewared/plugins/vm.py#L2166.

Please let me know if there are any issues. Thank you

src/middlewared/middlewared/plugins/vm.py
sonicaj: @stattin42 I have been going through libvirt source and it is going to keep on adding an extra controller each time we cross the AHCI devices allowed in one slot which means there's potential for conflict. Unfortunately there doesn't seem to be a good way to resolve this except for patching bhyve driver to do this properly as it's just consuming an entire slot. So I think the best case for us is to keep on using xml arguments as well for passthrough.

Also can we remove `LIBVIRT_HOSTDEV` variable references ? When libvirt bhyve driver has support for it, we will not be keeping backward compatibility like this, we will just remove the bhyve args code ; ) Older version releases use a different stable branch so I don't anticipate a scenario where this variable and it's usages will help us. Please use tomorrow nightlies for testing as we are updating ports tree and we will have a newer version of libvirt in to work with.

stattin42: Ok. Thanks for checking @sonicaj. I just tested with PR #4549. Worked with my test configurations (with only two ahci devices). Will update though, based on your findings for more complex configurations, to always include xml for passthrough. I will remove the LIBVIRT_HOSTDEV variable too, but keep the comments with reasons for using both xml and bhyve args.

stattin42:
> 
> 
> @stattin42 for VM's if we want to run some action on boot, we do that here
> 
> https://github.com/freenas/freenas/blob/8c25faa4cac2d1aeff926e05e5a1eea1d8756f2d/src/middlewared/middlewared/plugins/vm.py#L2166
> .
> 
> Please let me know if there are any issues. Thank you

Thanks @sonicaj! I chose a slightly different approach. Since most(?) people might not use pci passthrough or only use it occasionally, I thought it would be good to check for iommu support and available pptdevs only when first needed and then cache the results to avoid the overhead/delay due to repeated spawning of shell commands. Let me know if you see a problem with this approach. Thanks!

sonicaj:
Thank you for addressing the reviews @stattin42, is the PR ready for review ?

stattin42:
@sonicaj, PR is rebased, tested and should be ready for review now.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1993,6 +2155,38 @@ def translate_device(dev):
...
+                    outs, errs = proc2.communicate(timeout=5)
+                except subprocess.TimeoutExpired:
+                    proc2.kill()
+                    outs, errs = proc2.communicate()
+                if outs:
+                    return type
+            return None
+
+        if self.iommu_type is self.NotSet:
+            for type in ['VT-d', 'amdvi']:
```
sonicaj, line 2184: We shouldn't use words which are already in use by python i.e `type`

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1737,6 +1866,14 @@ class Config:
         datastore = 'vm.device'
         datastore_extend = 'vm.device.extend_device'
 
+    class NotSet:
```
sonicaj, line 1869: We can remove this based on below

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1804,6 +1941,30 @@ def nic_attach_choices(self):
         """
         return self.middleware.call_sync('interface.choices', {'exclude': ['epair', 'tap', 'vnet']})
 
+    @accepts()
+    def pptdev_choices(self):
+        """
+        Available choices for PCI passthru device.
+        """
+        if self.pptdevs is self.NotSet:
```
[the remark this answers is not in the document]
stattin42: Ok. This was a leftover from the original plan to probe at boot; i.e.,  before actually parsing any vm configuration. I wanted to avoid useless retries when PPT was not used and iommu or pptdevs were not available. Now that the probe is anyway only done when PPT is configured it is probably reasonable to assume that iommu support and pptdev should typically be available and a repeat of the probe may be desirable or at least acceptable.

sonicaj:
@stattin42 is this ready for review ?

stattin42:
@sonicaj, yes, ready for review. Sorry for (again) not making that clear.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1804,6 +1938,38 @@ def nic_attach_choices(self):
...
+            try:
+                proc = subprocess.Popen(['/usr/sbin/pciconf', '-l'],
+                                        stdout=subprocess.PIPE,
+                                        stderr=subprocess.DEVNULL, text=True)
+                outs, errs = proc.communicate(timeout=5)
+            except (subprocess.TimeoutExpired, OSError, ValueError) as e:
+                if proc is not None:
+                    proc.kill()
+                    outs, errs = proc.communicate()
+                self.logger.warn(f'An error occured when checking the PCI configuration '
```
sonicaj, line 1957: Please see the logging message below. Also while we are here can we please make sure to only use `f-strings` where required ?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1804,6 +1938,38 @@ def nic_attach_choices(self):
...
+            else:
+                if proc.returncode < 1:
+                    lines = outs.split('\n')
+                    for line in lines:
+                        object = RE_PCICONF_PPTDEVS.match(line)
+                        if object:
+                            pptdev = object.group(2).replace(':', '/')
+                            self.pptdevs[pptdev] = pptdev
+                else:
+                    self.logger.warn(f'An error occured when checking the PCI configuration '
```
sonicaj, line 1968: Same here for the `f-strings`

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1804,6 +1938,38 @@ def nic_attach_choices(self):
...
+                                        stderr=subprocess.DEVNULL, text=True)
+                outs, errs = proc.communicate(timeout=5)
+            except (subprocess.TimeoutExpired, OSError, ValueError) as e:
+                if proc is not None:
+                    proc.kill()
+                    outs, errs = proc.communicate()
+                self.logger.warn(f'An error occured when checking the PCI configuration '
+                                 f'for devices available for PCI passthru. {e}.')
+            else:
+                if proc.returncode < 1:
```
sonicaj, line 1960: Is `-1` returncode okay ? I suspect okay would be `0`, can you please confirm ?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1804,6 +1938,38 @@ def nic_attach_choices(self):
...
+                proc = subprocess.Popen(['/usr/sbin/pciconf', '-l'],
+                                        stdout=subprocess.PIPE,
+                                        stderr=subprocess.DEVNULL, text=True)
+                outs, errs = proc.communicate(timeout=5)
+            except (subprocess.TimeoutExpired, OSError, ValueError) as e:
+                if proc is not None:
+                    proc.kill()
+                    outs, errs = proc.communicate()
+                self.logger.warn(f'An error occured when checking the PCI configuration '
+                                 f'for devices available for PCI passthru. {e}.')
```
sonicaj, line 1958: Please use `:` instead of `.`

src/middlewared/middlewared/plugins/vm.py
sonicaj: About the logging as well please

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1993,6 +2160,52 @@ def translate_device(dev):
...
+                    if proc2 is not None:
+                        proc2.kill()
+                        outs, errs = proc2.communicate()
+                    elif proc1 is not None:
+                        proc1.kill()
+                    self.logger.warn(f'An error occured when checking for iommu ({iommu_type}). {e}.')
+                else:
+                    if proc1.poll() is None:  # make sure proc1 has terminated
+                        proc1.kill()
+                    if proc1.returncode < 1 and proc2.returncode < 2:
```
sonicaj, line 2193: Sam here about returncode, can you please confirm if all return codes below these are valid ?

src/middlewared/middlewared/plugins/vm.py
stattin42: Not sure I understand your reference to "the logging message below". Do you refer to using "self.middleware.logger.error" or are suggesting some rewording of the message (in addition to changing the . to :)?

src/middlewared/middlewared/plugins/vm.py
stattin42: Ok. Done.

src/middlewared/middlewared/plugins/vm.py
stattin42: Don't know of a situation where return code would be `<0` (except if terminated due to a signal). Unlikely, but not impossible this could happen, so agree it would be prudent to be conservative and only accept `0`. Better throw an extra error which can be analysed than risk intermittent malfunction with no clues why. Done.

src/middlewared/middlewared/plugins/vm.py
stattin42: See above. Done(?)

src/middlewared/middlewared/plugins/vm.py
stattin42: See above. For grep (proc2) return codes 0 and 1 are good. Done.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1993,6 +2160,52 @@ def translate_device(dev):
...
+                        if outs:
+                            return iommu_type
+                    else:
+                        self.logger.warn(f'An error occured when checking for iommu ({iommu_type}). '
```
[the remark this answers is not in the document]
stattin42: Done.

stattin42:
@sonicaj, thank you for your helpful review. I have updated and tested the PR. It is ready for rereview.

stattin42:
The PR is rebased on current master and ready for review.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1982,6 +2149,53 @@ def translate_device(dev):
         else:
             return False
 
+    @private
+    def get_iommu_type(self):
+        def check_iommu(iommu_type):
+            IOMMU_TEST = {'VT-d': {'arglist': ['/usr/sbin/acpidump', '-t'],
```
sonicaj, line 2155: @stattin42 this looks good to me, however I was thinking do we really need to run `grep` ? How about just parsing the output in python instead of managing 2 processes and ensuring everything is good wrt their management ?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1982,6 +2149,53 @@ def translate_device(dev):
...
+                                   'string': 'DMAR'},
+                          'amdvi': {'arglist': ['/sbin/sysctl', 'hw'],
```
sonicaj, line 2157: We could simply use `sysctl hw.vmm.amdvi.enable` ?
Btw we have a python module `sysctl` which can be used like this
```
sysctl.filter('hw.vmm.amdvi.enable')[0].value
```

src/middlewared/middlewared/plugins/vm.py
stattin42: @sonicaj, I have revised to skip grep in favour of parsing the output in python. To further shield middlewared from poential misbehaviour of the external utility and to avoid the need to buffer the complete output in python I also switched to asyncio.

src/middlewared/middlewared/plugins/vm.py
stattin42: Yes, `sysctl hw.vmm.amdvi.enable` would certainly reduce the amount of output to parse. The `sysctl` python module looks nice too. Only drawback would seem to be that the detection mechanisms become further differentiated. I have made an attempt though to abstract the difference by including a detection mechanism/function in the IOMMU_TESTS dictionary. This way the different tests can still be looped over without conditions or exceptions. Maybe there is a better way?

sonicaj: I think I can only advise once I see it ; )

stattin42:
@sonicaj, the PR has been updated based on your review and rebased. Ready for review.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1987,6 +2154,57 @@ def translate_device(dev):
...
+                if proc.returncode == 0:
+                    return found
+                else:
+                    self.middleware.logger.error('An error occured when checking for iommu '
+                                                 f'({iommu_type}): Subprocess ({cmd_args}) '
+                                                 f'exited with status {proc.returncode}.')
+            return None
+
+        async def int_sysctl(iommu_type, iou):
+            res = sysctl.filter(iou)
```
sonicaj, line 2193: `sysctl` module should be used in synchronous functions, I am sorry I should have clarified in my last comment

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1987,6 +2154,57 @@ def translate_device(dev):
...
+                        found = True
+                else:
+                    break  # Reached EOF
+            return found
+
+        async def ext_util(iommu_type, cmd_args, pattern):
+            found = False
+            proc = None
+            try:
+                proc = await asyncio.create_subprocess_exec(*cmd_args,
```
sonicaj, line 2174: So we already have this in place and we can just reuse the existing implementation reducing duplication and making it more neat ; )
Please see this example - https://github.com/freenas/freenas/blob/66c51a77288731d3eaeca3a2dc756c4d0e12669b/src/middlewared/middlewared/plugins/disk_/swap_configure.py#L78

You can use this for subprocess calls

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1987,6 +2154,57 @@ def translate_device(dev):
         else:
             return False
 
+    @private
+    def get_iommu_type(self):
```
sonicaj, line 2158: There is no reason to keep this synchronous anymore ? I think we can just use a subprocess call for the sysctl and keep it simple ?

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1987,6 +2154,57 @@ def translate_device(dev):
...
+            return True if res and res[0].value else False
+
+        IOMMU_TESTS = {'VT-d': {'mech': ext_util,
+                                'args': {'cmd_args': ['/usr/sbin/acpiXdump', '-t'],
+                                         'pattern': r'DMAR'}},
+                       'amdvi': {'mech': int_sysctl,
+                                 'args': {'iou': 'hw.vmm.amdvi.enable'}}}
+        if self.iommu_type is None:
+            for key, value in IOMMU_TESTS.items():
+                if asyncio.run(value['mech'](key, **value['args'])):
```
sonicaj, line 2203: We have a framework in place for asynchronous/synchronous functions. I think if you still want to use this approach ( separate functions ), it's better to have separate asynchronous/synchronous functions with the `private` tag and then call them here.

stattin42:
@sonicaj, in the event that you try the PR, I had introduced an error today in the spelling of the utility `acpidump` so VT-d support was not properly detected :-( Fixed and tested now.

src/middlewared/middlewared/plugins/vm.py
stattin42: Fair enough. Then I also don't need to worry about the sysctl call being used in synchronous function.

sonicaj: Yup, `run` would take care of that nicely ; )

src/middlewared/middlewared/plugins/vm.py
stattin42: I'll revert to using subprocess call for sysctl; see above.

src/middlewared/middlewared/plugins/vm.py
sonicaj: btw just to be clear, python `sysctl` module should only be used in synchronous function and it should not be used in an `asynchronous` one as it has the potential to block the event loop

src/middlewared/middlewared/plugins/vm.py
stattin42: If I understand correctly `middlewared.utils.run` will not return until the subprocess has finished and the complete output will be buffered in memory before it becomes available to the caller. Probably ok if the subprocess is well bahaved and does not produce too much output, but in case it misbehaves could it not potentially make middlewared unstable or run out of memory?

My intention was to avoid the need to buffer much in python and to avoid getting stuck by doing asynchronous readlines and by terminating the subprocess after a timeout if it does not complete within reasonable time. A bit safer I think although, admittedly, some more code. What do you think?

stattin42: I just saw there is similar handling in this example:
https://github.com/freenas/freenas/blob/66c51a77288731d3eaeca3a2dc756c4d0e12669b/src/middlewared/middlewared/plugins/disk_/hddstandby_force.py#L58-L88
albeit the scope of the timeout timer is slightly different (restarted for each line read instead of only before the first line).

stattin42: @sonicaj, unless you have a strong preference I think it would be prudent to keep the extra protection. The PR has been updated (less this aspect) and is ready for review. 

Thank you for constructive feedback and suggestions!

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1798,6 +1932,38 @@ def nic_attach_choices(self):
...
 
+    @accepts()
+    def pptdev_choices(self):
+        """
+        Available choices for PCI passthru device.
+        """
+        if not self.pptdevs:
+            proc = None
+            try:
+                proc = subprocess.Popen(['/usr/sbin/pciconf', '-l'],
```
sonicaj, line 1943: Same here please, we should be fine with just `run`

stattin42: @sonicaj, I am not _aware_ of a _current_ issue with `acpidump` or `sysctl`, apart from potentially generating a lot of output depending on how they are invoked. From a maintenance, robustness and security point of view, however, it may be beneficial to consider not only the current situation, but also how to detect and mitigate it in case the situation changes or there is an error. E.g., due to a bug triggered only under certain circumstances or due to an upgrade of the external tools otherwise unrelated to middlewared. I think it would be a sound principle to apply more generally and personally I would favour an evolution of the FN/TN async framework to provide this protection. 

As this is not my project and I understand you have more things to consider, e.g., code maintenance, I am fine with changing to `run`. Hoping to see a future evolution of the framework though to offer isolation/protection from unexpected behaviour/misbehaviour of external utilities.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1987,6 +2154,53 @@ def translate_device(dev):
...
+                                                            stdout=asyncio.subprocess.PIPE,
+                                                            stderr=asyncio.subprocess.DEVNULL)
+                found = await asyncio.wait_for(check_util_output(proc, pattern), timeout=5)
+            except (asyncio.TimeoutError, OSError, ValueError) as e:
+                if proc is not None:
+                    proc.kill()
+                self.middleware.logger.error('An error occured when checking for iommu '
+                                             f'({iommu_type}): {e}.')
+            else:
+                if proc.returncode == 0:
```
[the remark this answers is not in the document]
stattin42: Right, it seems I forgot that or assumed that the process would be more wellbehaved after reaching EOF than what I assumed for other aspects :-P Thanks for spotting this. Will fix.

src/middlewared/middlewared/plugins/vm.py
stattin42: See reply to same question in other comment.

src/middlewared/middlewared/plugins/vm.py
stattin42: @sonicaj, the PR is updated according to our discussion so far and ready for review.

stattin42:
@sonicaj, the PR is updated according to our discussion so far and ready for review.

src/middlewared/middlewared/plugins/vm.py
```diff
@@ -1987,6 +2140,24 @@ def translate_device(dev):
...
+                       'amdvi': {'cmd_args': ['/sbin/sysctl', '-i', 'hw.vmm.amdvi.enable'],
+                                 'pattern': r'vmm\.amdvi\.enable: 1'}}
+
+        if self.iommu_type is None:
+            for key, value in IOMMU_TESTS.items():
+                sp = await run(*value['cmd_args'], check=False)
+                if sp.returncode:
+                    raise CallError(f'Failed to check support for iommu ({key}): {sp.stderr.decode()}')
+                else:
+                    if re.search(value['pattern'], sp.stdout.decode()):
```
sonicaj, line 2156: I am getting a decode error in my system for this
```
freenas# midclt call vm.device.get_iommu_type
'utf-8' codec can't decode byte 0xf1 in position 7552: invalid continuation byte
  File "/usr/local/lib/python3.8/site-packages/middlewared/main.py", line 135, in call_method
    result = await self.middleware._call(message['method'], serviceobj, methodobj, params, app=self,
  File "/usr/local/lib/python3.8/site-packages/middlewared/main.py", line 1158, in _call
    return await methodobj(*args)
  File "/usr/local/lib/python3.8/site-packages/middlewared/plugins/vm.py", line 2156, in get_iommu_type
    if re.search(value['pattern'], sp.stdout.decode()):
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf1 in position 7552: invalid continuation byte
```

I think you should probably use `br'DMAR'` and not try to decode it. It should be safe I assume in this case

src/middlewared/middlewared/plugins/vm.py
sonicaj: Thank you for your concerns @stattin42, if we notice such a case ( or as time permits ), we will implement adequate measures to ensure we are good.

src/middlewared/middlewared/plugins/vm.py
stattin42: @sonicaj, thank you for testing. Interesting and, although the output of acpidump is heavily dependant on system hw, somewhat unexpected. On my system, I don't get that error :-) I agree that using bytes instead of string is probably ok in this case. Will update accordingly. 
Trying to understand what caused the decode error on your system, would you mind sharing the output line with the offending bytes?

PR #4412, opened 2020-03-16
NAS-105426 / 12.0 / Reflect unlock scenario in encryption summary for encrypted datasets

sonicaj:
This PR introduces changes so that we reflect which datasets will be unlocked in `pool.dataset.encryption_summary` based on the provided keys and current pool state.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2222,21 +2222,54 @@ class PoolDatasetService(CRUDService):
 
         Keys/passphrase can be supplied to check if the keys are valid.
 
+        It should be noted that there are 2 keys which show if a recursive unlock operation is
+        done for `id`, which dataset will be unlocked and if not why it won't be unlocked. The keys
+        namely are "unlock_successful" and "unlock_error". The former is a boolean value showing if unlock
+        would succeed/fail. The latter is description why it failed if it failed.
+
+        If a dataset is already unlocked, it will show up as true for "unlock_successful" regardless of what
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2276,9 +2309,38 @@ def encryption_summary(self, job, id, options):
             results.append({
                 'name': ds_name, 'key_format': ZFSKeyFormat(data['key_format']['value']).value,
```
themylogin: Can we format this according to PEP8 please?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2276,9 +2309,38 @@ def encryption_summary(self, job, id, options):
...
+        for ds in sorted(results, key=lambda d: d['name'].count('/')):
+            for i in range(1, ds['name'].count('/') + 1):
+                check = ds['name'].rsplit('/', i)[0]
+                if check in failed:
+                    failed.add(ds['name'])
+                    ds['unlock_error'] = f'Child cannot be unlocked when parent "{check}" is locked'
+
+            if ds['valid_key']:
+                if not ds['unlock_error']:
+                    ds['unlock_successful'] = not bool(ds['unlock_error'])
```
themylogin: Can't it be anything but `True` here?

sonicaj: It will be `False` if `unlock_error` is set above in the loop

themylogin: I mean, `not bool(ds['unlock_error'])` will always be `True` in this `if not ds['unlock_error']:` branch?

PR #4421, opened 2020-03-17
NAS-105452 / 12.0 / Load openzfs-debug when booting kernel-debug

freqlabs:
Also removed redundant modules path variable from loader.conf.

### DIFF
--- a/src/freenas/boot/loader.conf
+++ b/src/freenas/boot/loader.conf
@@ -15,6 +15,9 @@ debug.ddb.textdump.pending=1
 
 hw.hptrr.attach_generic=0
 
+# Possible kernel module locations (in addition to /boot/${kernel})
+module_path="/boot/modules;/usr/local/modules"
+
 # Load firmware for isp FC cards.
 ispfw_load="YES"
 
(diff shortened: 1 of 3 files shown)


freqlabs:
@william-gr Do we generate loader.conf.local during install or will I need to keep loader.conf bootable when there is not a loader.conf.local on first boot?

amotin:
I have got the same question. We definitely generate some tunables from the installer, for example to pass serial console settings. But I am not sure it is the same scripts, so ability of the first boot worries me too.

amotin:
Also I don't remember whether we allow choice of debug/normal kernel from boot menu. Definitely on the first page, but any deeper? I guess debug module may be unhappy to boot with normal kernel. Is it?

freqlabs:
Ah we can pick a debug kernel at the boot menu. That will require some customization too.

william-gr:
loader.conf.local is not generated the same way for install, as far as I can remember loader is manipulated directly from the installer for the first time so loader.conf needs to remain bootable

freqlabs:
I'll merge this now, but the hook isn't going to be invoked until we merge it from stable/12. I have tested by manually editing config.lua to add the hook and made sure it works as intended.

PR #4423, opened 2020-03-17
NAS-105443 / 12.0 / Get list of SMB IP addresses from interface.ip_in_use

anodos325:
- Get full list of addresses
- Validate user-provided addresses at smb.conf generation time
  to avoid having stale IP address prevent SMB access.

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/smb4.conf
+++ b/src/middlewared/middlewared/etc_files/local/smb4.conf
@@ -79,7 +79,17 @@
             smbpasswd by default connects to 127.0.0.1 as an SMB client. For this reason, localhost is added
             to the list of bind ip addresses here.
             """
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/etc_files/local/smb4.conf
```diff
@@ -79,7 +79,17 @@
             smbpasswd by default connects to 127.0.0.1 as an SMB client. For this reason, localhost is added
             to the list of bind ip addresses here.
             """
-            if db['cifs']['bindip']:
+            allowed_interfaces = middleware.call_sync('smb.bindip_choices')
+            validated_bind_ips = []
+            for interface in db['cifs']['bindip']:
```
william-gr: Isnt interface an odd name if its really an IP? or am I misreading what it means?

PR #4437, opened 2020-03-19
fix(middlewared/client): run callback in another thread

william-gr:
This will avoid the client blocking waiting on the callback


william-gr:
@themylogin  TrueNAS HA, we can have deadlock on events.

Say an event with a callback tries to use the same client to make another call.

I am not sure its ever safe to run callbacks synchronously since it will always halt the client?

PR #4439, opened 2020-03-20
Failover support for Native ZFS Encryption

sonicaj:
1) Removes `escrowd` and use cache plugin instead.
2) Introduces support for automatic unlocking of encrypted datasets on failover event.
3) We remove ability to set passphrase from netcli as new UI allows the user to log in and unlock the pool from there.


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -802,75 +700,144 @@ def _disabled_reasons(self, app):
...
+                keys['zfs'] = {
+                    k: v for k, v in keys['zfs'].items() if k != dataset and not k.startswith(f'{dataset}/')
+                }
+            await self.middleware.call('cache.put', 'failover_encryption_keys', keys)
+            if options['sync_keys']:
+                await self.sync_keys_with_remote_node(lock=False)
 
     @private
     @job()
     def encryption_attachall(self, job):
```
yocalebo: Since you're here and making all these wonderful changes, it would be best to name this function to something not so ambiguous. :)

Maybe something like `def attach_all_geli_providers`

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -903,36 +870,29 @@ def encryption_attachall(self, job):
...
+                    })
+                except Exception as e:
                     failed_volume += 1
+                    self.logger.error('Failed to import %s pool: %s', pool['name'], str(e))
 
-            if failed_drive > 0:
-                job.set_progress(None, f'{failed_drive} can not be attached.')
-                self.logger.error('%d can not be attached.', failed_drive)
+        if failed_drive > 0:
+            job.set_progress(None, f'{failed_drive} can not be attached.')
```
yocalebo: `{failed_drive}` is an integer which doesn't represent the actual geli provider. The `encrypted_disk['encrypted_provider']` should be what you want.

sonicaj: I think original author meant something like `2 drives cannot be attached` as not sure if in progress it would be of much benefit telling the paths and they can be a lot too if we have more disks in the pool.
Nice catch ; ) I'll make amends, it will look really weird right now : p

src/freenas/usr/local/lib/middlewared_truenas/alert/source/failover.py
```diff
@@ -193,14 +193,17 @@ def check_sync(self):
                 pass
             try:
                 if len(fobj['phrasedvolumes']) > 0:
-                    if not self.middleware.call_sync('failover.encryption_status'):
-                        alerts.append(Alert(NoFailoverEscrowedPassphraseAlertClass))
+                    keys = self.middleware.call_sync('failover.encryption_keys')['geli']
```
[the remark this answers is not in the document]
sonicaj: I am not sure I follow, isn't this supposed to run on MASTER controller ? And if yes, we are going to get the alert on MASTER controller which is good ? If the user does a failover, this will run on the new node which will have become MASTER and we should be good there too ?

src/freenas/usr/local/lib/middlewared_truenas/alert/source/failover.py
```diff
@@ -100,11 +100,11 @@ class CTLHALinkAlertClass(AlertClass):
...
 
 
-class NoFailoverEscrowedPassphraseAlertClass(AlertClass):
+class NoFailoverPassphraseKeysAlertClass(AlertClass):
     category = AlertCategory.HA
     level = AlertLevel.CRITICAL
-    title = "No Escrowed Passphrase for Failover"
-    text = "No escrowed passphrase for failover. Automatic failover disabled."
+    title = "No Pool Passphrase Key for Failover"
+    text = "Passphrase key for %(pool)r not available for failover. Automatic failover disabled."
```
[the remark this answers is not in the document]
william-gr: People do not know what GELI is, its an implementation detail. We should probably say Legacy encryption or something along these lines.

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -783,8 +682,9 @@ def _disabled_reasons(self, app):
          2. the quantity of disks are the same between
             controllers but serials do not match
         """
-        local_boot_disks = await self.middleware.call('zfs.pool.get_disks', 'freenas-boot')
-        remote_boot_disks = await self.middleware.call('failover.call_remote', 'zfs.pool.get_disks', ['freenas-boot'])
+        boot_pool = await self.middleware.call('boot.pool_name')
+        local_boot_disks = await self.middleware.call('zfs.pool.get_disks', boot_pool)
```
william-gr: How about boot.get_disks instead?

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1722,13 +1753,21 @@ def sync_internal_ips(middleware, iface, carp1_skew, carp2_skew, internal_ip):
             middleware.logger.warn(f'Failed to run {verb}({service})', exc_info=True)
 
 
+async def ready_system_sync_keys(middleware):
+    await middleware.call('failover.call_remote', 'failover.sync_keys_with_remote_node')
```
william-gr: I think it may make more sense to do this in the event plugin, like it was done before?

src/middlewared/middlewared/plugins/disk_/encryption_freebsd.py
```diff
@@ -75,17 +75,9 @@ def geli_attach_single(self, dev, key, passphrase=None, skip_existing=False):
...
-            self.__geli_notify_passphrase(passphrase)
         else:
             self.logger.debug(f'{dev} already attached')
 
-    def __geli_notify_passphrase(self, passphrase):
-        if passphrase:
-            with open(passphrase) as f:
-                self.middleware.call_hook_sync('disk.post_geli_passphrase', f.read())
-        else:
-            self.middleware.call_hook_sync('disk.post_geli_passphrase', None)
```
william-gr: Was this not used?

sonicaj: We were using this hook before but with this hook we are not able to map a disk with a pool. So I changed this and we are using import/export/unlock/lock/passphrase hooks to keep the cache updated.

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1268,6 +1230,21 @@ def upgrade_finish(self, job):
         self.middleware.call_sync('keyvalue.set', 'HA_UPGRADE', False)
         return True
 
+    @private
+    async def sync_keys_with_remote_node(self, lock=True):
```
william-gr: sync_keys_to_remote_node seems to be a more direct name

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
```diff
@@ -363,7 +363,8 @@ def carp_master(self, fobj, ifname, vhid, event, user_override, forcetakeover):
                 # For reboots, /tmp is cleared by virtue of being a memory device.
                 # If someone does a kill -9 on the script while it's running the lockfile
                 # will get left dangling.
-                self.logger.warn('Aquired failover master lock')
+                self.middleware.call_sync('failover.call_remote', ['failover.sync_keys_with_remote_node'])
```
william-gr: We dont need this here, I misunderstood when you asked.

We will already sync when entering in backup and thats enough

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
```diff
@@ -751,14 +762,14 @@ def carp_backup(self, fobj, ifname, vhid, event, user_override):
...
-
-                if fobj['phrasedvolumes']:
-                    self.logger.warn('Setting passphrase from master')
-                    run('LD_LIBRARY_PATH=/usr/local/lib /usr/local/sbin/enc_helper syncfrompeer')
+                detach_all_job = self.middleware.call_sync('failover.encryption_detachall')
+                detach_all_job.wait_sync()
+                if detach_all_job.error:
+                    self.logger.error('Failed to detach geli providers: %s', detach_all_job.error)
 
+                self.middleware.call_sync('failover.call_remote', ['failover.sync_keys_with_remote_node'])
```
william-gr: This seems wrong?

self.middleware.call_sync('failover.call_remote', 'failover.sync_keys_with_remote_node')

sonicaj: It is wrong, thank you for catching that. I would've picked it up after i rolled a new build : p

src/freenas/usr/local/lib/middlewared_truenas/alert/source/failover_sync.py
```diff
@@ -18,3 +18,18 @@ class FailoverSyncFailedAlertClass(AlertClass, SimpleOneShotAlertClass):
...
+
+
+class FailoverKeysSyncFailedAlertClass(AlertClass, SimpleOneShotAlertClass):
+    deleted_automatically = False
+
+    category = AlertCategory.HARDWARE
+    level = AlertLevel.CRITICAL
+    title = "Automatic Keys Sync to Peer Failed"
+    text = (
+        "Failed to sync keys/passphrases configuration information to standby storage "
```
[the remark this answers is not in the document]
Mrt134: Here's some ideas for the verbiage:
`title = Syncing Encryption Keys to Peer Failed`, `Failed to Sync Encryption Details to Peer`, `Error syncing encryption keys to peer`

`text = Failed to sync encrypted pool configuration details to the standby controller. To manually sync this information with the standby controller, please go to the System > Failover page and select the Sync to Peer option.`

`The automatic synchronization of legacy encryption passphrases or current encryption keys with the standby controller has failed. Please go to System > Failover and manually sync to peer.`

src/freenas/etc/netcli
```diff
@@ -1539,14 +1539,27 @@ def main_menu():
...
-                    time.sleep(5)
+            with open('/tmp/.failover_needop', 'r') as f:
+                failed_pools = [p.strip() for p in f.read().strip().split('\n') if p.strip()]
+
+            if passphrase and failed_pools:
+                yes = input('Enter passphrase for pools which failed to decrypt ? (y/n)')
+                if yes.lower().startswith('y'):
+                    cache = []
+                    for pool in failed_pools:
+                        key = input(f'Please enter passphrase for "{pool}" pool: ')
```
william-gr: What if input is empty? User just pressed enter or ctrl+d?

src/freenas/usr/local/lib/middlewared_truenas/alert/source/failover.py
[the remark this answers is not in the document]
yocalebo: I mis-read this originally. Looks good.

PR #4457, opened 2020-03-24
Change default cipher to AES-256-GCM

sonicaj:


freqlabs:
Do we only support encrypting the whole pool or is there somewhere another default for encrypted single datasets, too?

sonicaj:
@freqlabs the support is present at a dataset level and the default is only listed here and this schema is referenced where we create single datasets - so they get the default we set here too.

PR #4459, opened 2020-03-24
NAS-105499 / 12.0 / API Keys

themylogin:


src/middlewared/middlewared/plugins/api_key.py
```diff
@@ -0,0 +1,145 @@
...
+        except ValueError:
+            return None
+
+        try:
+            db_key = await self.middleware.call("datastore.query", "account.api_key", [("id", "=", key_id)],
+                                                {"get": True})
+        except MatchNotFound:
+            return None
+
+        if not pbkdf2_sha256.verify(key, db_key["key"]):
```
william-gr: Is this lightweight enough to not block the event loop?

src/middlewared/middlewared/restful.py
```diff
@@ -18,19 +18,28 @@
...
+            raise web.HTTPUnauthorized()
+
+        try:
+            if not await middleware.call('auth.check_user', username, password):
+                raise web.HTTPUnauthorized()
+        except web.HTTPUnauthorized:
+            raise
+        except Exception:
+            raise web.HTTPUnauthorized()
+    elif auth.startswith('API-Key '):
```
william-gr: This is not a valid authentication schema recognized by IANA, is it?

src/middlewared/middlewared/plugins/api_key.py
```diff
@@ -0,0 +1,145 @@
...
+
+        return db_key
+
+    async def _validate(self, schema_name, data, id=None):
+        verrors = ValidationErrors()
+
+        await self._ensure_unique(verrors, schema_name, "name", data["name"], id)
+
+        if verrors:
+            raise verrors
```
william-gr: We jave verrors.check() for shorter, JFYI

src/middlewared/middlewared/plugins/api_key.py
themylogin:
```
[desktop] ~> python3 -m timeit -s 'from passlib.hash import pbkdf2_sha256; s = "*" * 4096; h = pbkdf2_sha256.encrypt("*" * 64)' 'pbkdf2_sha256.verify(s, h)'   
100 loops, best of 3: 15.7 msec per loop
```

src/middlewared/middlewared/restful.py
themylogin: Is `Bearer` ok?

PR #4460, opened 2020-03-25
Fix collectd service deadlock

themylogin:


themylogin:
@william-gr there is no more "start rrdcached when starting collectd and start collectd when starting rrdcached" infinite recursion.

PR #4469, opened 2020-03-26
feat(middleawred/system): add SCALE to system.product_type

william-gr:


Ornias1993:
It's still funny how almost no FreeNAS user actually reads the source code.
Two lines of code, tell more than a thousant press releases :)

Nice work on Scale btw ;)

PR #4470, opened 2020-03-26
fix(middlewared/reporting): realtime event working (minus interface stats)

william-gr:


src/middlewared/middlewared/plugins/reporting.py
```diff
@@ -1060,30 +1065,75 @@ def run(self):
...
-                    break
-                data['cpu']['temperature'][i] = v[0].value
+            if osc.IS_FREEBSD:
+                for i in itertools.count():
+                    v = sysctl.filter(f'dev.cpu.{i}.temperature')
+                    if not v:
+                        break
+                    data['cpu']['temperature'][i] = v[0].value
+            elif osc.IS_LINUX:
+                cp = subprocess.run(['sensors', '-j'], capture_output=True, text=True)
```
themylogin: Can this be worth to use https://pypi.org/project/PySensors/ to avoid launching new process every time in a loop?

william-gr: Sure, I can create you a ticket to create a package for that and change usage

PR #4477, opened 2020-03-27
NAS-105227 / 12.0 / FREENAS-MINI-3.0-E enclosures

themylogin:


src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class EnclosureService(Service):
+    @private
+    async def map_enclosures(self, enclosures):
+        info = await self.middleware.call("system.info")
+        for product_mapping in MAPPINGS:
+            if product_mapping.product_re.match(info["system_product"]):
+                for version_mapping in product_mapping.mappings:
+                    if version_mapping.version_re.match(info["system_product_version"]):
```
william-gr: Is there any hardware you're seeing that will need the version?

src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
```diff
@@ -0,0 +1,108 @@
...
+
+from middlewared.service import Service, private
+
+ProductMapping = namedtuple("ProductMapping", ["product_re", "mappings"])
+VersionMapping = namedtuple("VersionMapping", ["version_re", "slots"])
+MappingSlot = namedtuple("MappingSlot", ["num", "slot", "identify"])
+
+
+MAPPINGS = [
+    ProductMapping(re.compile(r"To Be Filled By O.E.M."), [
```
william-gr: Why this?

themylogin: That's what was needed to test on test06, do you want me to remove that?

src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
themylogin: NEP says that mapping can vary from version to version

src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
william-gr: Yes, I was thinking just was just be there while testing in the box, not actually commit it to the repo. Or wait until we have a final say of what the system-productname will be

src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
william-gr: Can you remind me which section?

william-gr: Ah, nevermind. its there just for future use and not actually used yet.

PR #4480, opened 2020-03-30
NAS-105557 / 12.0 / fix resolve_hostname method

yocalebo:
If you were to send `192.168` to this method, it would fail the ip check and raise a ValueError but then it would be expanded to `192.0.0.168` by the `gethostbyname()` call. This fixes that problem and furthermore, this method did not support IPv6 which I've fixed too.


yocalebo:
pushed wrong branch. will reopen

PR #4481, opened 2020-03-30
NAS-105557 / 12.0 / fix resolve_hostname async validator

yocalebo:
If you send a short-hand IP address (i.e. `192.168`) to this method, it will raise a ValueError as expected, however the `gethostbyname()` method auto-expands the IP address to `192.0.0.168`.

Furthermore, `gethostbyname()` doesn't support IPv6 which I've fixed.


yocalebo:
I've added a synchronous method in `validators` module that essentially does the same as this method. Since this method is put into a thread and they do the same thing, I'll add the synchronous method and then import and use that method here to remove code duplication.

PR #4482, opened 2020-03-30
NAS-105558 / 12.0 / add ResolveHostOrIp validator

yocalebo:
I need a synchronous method for validating Hosts or IPs for the gluster peer API. This class does that and is directly based off the `resolve_hostname` method in the `async_validators` module (which I've also pushed a PR to fix a problem).

### DIFF
--- a/src/middlewared/middlewared/alert/source/update.py
+++ b/src/middlewared/middlewared/alert/source/update.py
@@ -68,7 +68,6 @@ class HasUpdateAlertSource(ThreadedAlertSource):
         if not path:
             return
 
-
         updates = None
         try:
             if PendingUpdates:
--- a/src/middlewared/middlewared/service.py
+++ b/src/middlewared/middlewared/service.py
@@ -578,7 +578,7 @@ class ServicePartBaseMeta(ServiceBase):
             else:
                 original_argspec = inspect.getfullargspec(original_method)
             if original_argspec != inspect.getfullargspec(new_method):
-                raise RuntimeError(f"Signature for method {name!r} does not match between {klass!r} and it's base " 
+                raise RuntimeError(f"Signature for method {name!r} does not match between {klass!r} and it's base "
                                    f"{base!r}")
 
             copy_function_metadata(original_method, new_method)
--- a/src/middlewared/middlewared/validators.py
+++ b/src/middlewared/middlewared/validators.py
@@ -3,10 +3,28 @@ import ipaddress
 import re
 from urllib.parse import urlparse
 import uuid
+import socket
 
 from zettarepl.snapshot.task.naming_schema import validate_snapshot_naming_schema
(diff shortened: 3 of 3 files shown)


yocalebo:
@william-gr I need a synchronous method for gluster peer API so I've created this one. This is essentially the same as `resolve_hostname` method in the `async_validators` module. However, instead of code duplication, I was going to use this class in the `async_vaildators` module since it gets put into a thread anyways.

william-gr:
@yocalebo yes, I think this wont work out. Doesnt this hits the network trying to resolve the hostname?

It may block the event loop while validating this value. It should probably be done within the method to run this in a thread.

Let me know if I didnt get this right @themylogin @sonicaj

yocalebo:
> 
> 
> @yocalebo yes, I think this wont work out. Doesnt this hits the network trying to resolve the hostname?
> 
> It may block the event loop while validating this value. It should probably be done within the method to run this in a thread.

Hmm, I didn't think about it when this class is called by itself. The gluster API I'm writing is wrapped in a job so it's not blocking the main event loop. I'm wondering if I should just write a validator that is local to the gluster API and use it there?

yocalebo:
> 
> 
> Yes, it will run network in the main loop. Currently we don't have infrastructure for async validators in schema, and I can explain why I think we don't need it, so I would propose to use async validator explicitly.

Is it okay to call an async validator in synchronous function?

`
@accepts(
        blah
)
@job(lock='blah', validators=[resolve_hostanme()])
`

themylogin:
> The gluster API I'm writing is wrapped in a job so it's not blocking the main event loop.

Event loop is common for everyone. Also, schema-level validations are performed before launching the job (so we can access sanitized arguments in `@job(lock=lambda args: ...)`), which is one of the reasons I object to have validators that perform I/O.

themylogin:
@yocalebo `@job` does not have `validators` argument.

Guess for you very case you can just do `asyncio.run_coroutine_threadsafe(resolve_hostname(...), loop=asyncio.get_event_loop()).result()` and we'll think of more elegant way to do this if this issue rises again.

yocalebo:
> 
> 
> @yocalebo `@job` does not have `validators` argument.
> 
> Guess for you very case you can just do `asyncio.run_coroutine_threadsafe(resolve_hostname(...), loop=asyncio.get_event_loop()).result()` and we'll think of more elegant way to do this if this issue rises again.

The `@jobs` was a typo, sorry. But thanks for the clarification. I'll close this PR.

One last thing, however, the `resolve_method` has incorrect behavior when provided a short-hand IP (i.e. 192.168) because the `gethostbyname` auto-expands that to `192.0.0.168`. I closed a PR earlier that fixes that issue. I'll push that PR again and have you review for a fix.

PR #4483, opened 2020-03-30
NAS-105557 / 12.0 / fix resolve_hostname when provided short-hand IPs 12.0

yocalebo:


src/middlewared/middlewared/async_validators.py
```diff
@@ -25,7 +25,13 @@ def resolve_host_name_thread(hostname):
                 ip(hostname)
                 return hostname
             except ValueError:
-                return socket.gethostbyname(hostname)
+                # gethostbyadd() is called here so that if a short-hand IP
+                # was provided (i.e. 192.168), gethostaddr() will raise the
+                # exception as intended. The getaddrinfo() will auto-expand
+                # short-hand IP addresses which isn't desired.
+                socket.gethostbyaddr(hostname)
+                result = socket.getaddrinfo(hostname, None flags=socket.AI_CANONNAME)
```
themylogin: Syntax error

src/middlewared/middlewared/async_validators.py
```diff
@@ -25,7 +25,13 @@ def resolve_host_name_thread(hostname):
                 ip(hostname)
                 return hostname
             except ValueError:
-                return socket.gethostbyname(hostname)
+                # gethostbyadd() is called here so that if a short-hand IP
+                # was provided (i.e. 192.168), gethostaddr() will raise the
+                # exception as intended. The getaddrinfo() will auto-expand
+                # short-hand IP addresses which isn't desired.
+                socket.gethostbyaddr(hostname)
```
themylogin: Why two queries, won't `socket.getaddrinfo` raise same exception?

yocalebo: No, `socket.getaddrinfo` will auto-expand the IP address. `gethostbyaddr` does not auto-expand and will raise an Exception as expected.

src/middlewared/middlewared/async_validators.py
yocalebo: Thanks

src/middlewared/middlewared/async_validators.py
yocalebo: @themylogin ping

themylogin: @yocalebo sorry, comment was pending for some reason. Do you really need to perform DNS query to detect that short-hand IP was provided?

yocalebo: It should only do a DNS lookup if the IP address is valid. So if you provide `192.168`, it won't auto-expand the IP address and fail immediately because it will say `Unknown Host`

yocalebo: Nope, I'm wrong. Disregard previous comment.

yocalebo: @themylogin There's no easy way around this that I can think of. I'm trying to resolve an edge-case with this fix anyways, so I'll close this for now. I've worked around this in another way wrt to the gluster peer API

PR #4490, opened 2020-03-30
NAS-105573 / 12.0 / Add NTPD user

sonicaj:
NPTD now runs with `ntpd` user, so we create that user to ensure `ntpd` runs as desired.


src/middlewared/middlewared/alembic/versions/12.0/2020-03-30_13-05_ntpd_user.py
```diff
@@ -0,0 +1,40 @@
+"""
```
themylogin: Manipulating built-in users/groups through DB migrations won't work good for cross-platform app.

We use assets/account/builtin/<platform>/(group|passwd) file instead. `account.sync_builtin` method does all the job.

PR #4494, opened 2020-03-31
NAS-105536 / 11.3 / Start AD on standby controller once active successfully starts

anodos325:
"Legacy" AD configuration requires that the AD service be started and running on both controllers. Make call to standby controller to start AD once we've confirmed that it's working on the active controller.


src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1015,6 +1015,13 @@ class Config:
             await self.middleware.call('activedirectory.get_cache')
             if ad['verbose_logging']:
                 self.logger.debug('Successfully started AD service for [%s].', ad['domainname'])
+
+            if smb_ha_mode == "LEGACY":
+                job.set_progress(95, 'starting active directory on standby controller')
+                try:
+                    await self.middleware.call('failover.call_remote', 'activedirectory.start')
```
william-gr: Wont that cause an infinite loop of calls? a node call the remote node, and that remote node call the other remote node again?

PR #4499, opened 2020-04-01
NAS-105398 / 12.0 / initial API for adding gluster peers to the cluster

yocalebo:


src/middlewared/middlewared/plugins/gluster_/peer.py
```diff
@@ -0,0 +1,122 @@
...
+        return await resolve_hostname(*args)
+
+    @private
+    def common_validation(self, hostname):
+
+        verrors = ValidationErrors()
+
+        loop = self.middleware.loop
+        try:
+            asyncio.run_coroutine_threadsafe(
```
themylogin, line 57: You can do `self.middleware.call_sync('gluster.peer.resolve_host_or_ip')`, that's what it is there for, to be able to switch between threads/asyncio easily

yocalebo: Thank you!

themylogin:
LGTM, thank you!

PR #4501, opened 2020-04-01
fix(middlewared/disk): import geom meodule on FreeBSD only

william-gr:


themylogin:
Can we rename it to `_freebsd` and put a condition in `pool.create`?

william-gr:
Sure

william-gr:
@themylogin haven't tested yet, will do as soon as you approve

PR #4505, opened 2020-04-02
NAS-105595 / 12.0 / Retrieve correct primary key on insertion of record in database

sonicaj:
This PR adds changes to ensure that we retrieve correct primary key on insertion of a record in database.

### DIFF
--- a/src/middlewared/middlewared/plugins/datastore/write.py
+++ b/src/middlewared/middlewared/plugins/datastore/write.py
@@ -1,4 +1,5 @@
 from sqlalchemy import and_, types
+from sqlalchemy.sql import sqltypes
 
 from middlewared.schema import accepts, Any, Dict, Str
 from middlewared.service import Service
@@ -29,10 +30,10 @@ class DatastoreService(Service, FilterMixin, SchemaMixin):
 
         await self.middleware.call('datastore.execute_write', table.insert().values(**insert))
         pk_column = self._get_pk(table)
-        if isinstance(pk_column.type, (types.String, types.TEXT)):
-            pk = insert[pk_column.name]
-        else:
+        if type(pk_column.type) == sqltypes.Integer:
             pk = (await self.middleware.call('datastore.fetchall', 'SELECT last_insert_rowid()'))[0][0]
+        else:
+            pk = insert[pk_column.name]
 
         await self._handle_relationships(pk, relationships)
 
--- a/src/middlewared/middlewared/plugins/datastore/write.py
+++ b/src/middlewared/middlewared/plugins/datastore/write.py
@@ -18,7 +18,6 @@ class DatastoreService(Service, FilterMixin, SchemaMixin):
         Insert a new entry to `name`.
         """
         table = self._get_table(name)
-
         insert, relationships = self._extract_relationships(table, options['prefix'], data)
 
         for column in table.c:
@@ -29,7 +28,11 @@ class DatastoreService(Service, FilterMixin, SchemaMixin):
                     insert.setdefault(column.name, '')
 
         await self.middleware.call('datastore.execute_write', table.insert().values(**insert))
-        pk = (await self.middleware.call('datastore.fetchall', 'SELECT last_insert_rowid()'))[0][0]
+        pk_column = self._get_pk(table)
+        if isinstance(pk_column.type, (types.String, types.TEXT)):
+            pk = insert[pk_column.name]
+        else:
+            pk = (await self.middleware.call('datastore.fetchall', 'SELECT last_insert_rowid()'))[0][0]
 
         await self._handle_relationships(pk, relationships)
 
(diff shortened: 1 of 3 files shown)


sonicaj:
So that it returns the complete row it inserted instead of just the id ?
We use `datastore.insert` everywhere currently to only retrieve the id on `do_create`. We could get the `id` from the row returned. Do you think we should proceed with changing this behaviour ?

sonicaj:
Also for this specific case, `disk_identifier` was already set. Generally it works well to only get the `id` from `datastore.insert` - alternate would be something like `data['id'] = middleware_call('datastore.insert', data)['id']`

themylogin:
It should return primary key, we are not interested in `last_insert_rowid()` if we are not using rowid as primary key. `_handle_relationships` will also fail with this approach.

src/middlewared/middlewared/plugins/datastore/write.py
```diff
@@ -29,7 +28,11 @@ class Config:
                     insert.setdefault(column.name, '')
 
         await self.middleware.call('datastore.execute_write', table.insert().values(**insert))
-        pk = (await self.middleware.call('datastore.fetchall', 'SELECT last_insert_rowid()'))[0][0]
+        row_id = (await self.middleware.call('datastore.fetchall', 'SELECT last_insert_rowid()'))[0][0]
+        pk = (await self.middleware.call(
```
themylogin, line 32: We can probably avoid extra query. If primary key is integer, we return rowid. If primary key is not integer, we already have it in `insert`. Also, please add tests.

sonicaj: That would be generally true ( at least right now it is I think ) but if say we were to have a table which has an integer as a primary key which is not `id` ?

themylogin: In SQLite integer primary key will automatically be an alias for rowid

src/middlewared/middlewared/plugins/datastore/write.py
```diff
@@ -29,7 +28,11 @@ class Config:
                     insert.setdefault(column.name, '')
 
         await self.middleware.call('datastore.execute_write', table.insert().values(**insert))
-        pk = (await self.middleware.call('datastore.fetchall', 'SELECT last_insert_rowid()'))[0][0]
+        pk_column = self._get_pk(table)
+        if isinstance(pk_column.type, (types.String, types.TEXT)):
```
themylogin, line 32: The only type for primary key being rowid is integer, for all other types (including e.g. bigint sqlite requires us to specify pk explicitly)

src/middlewared/middlewared/pytest/unit/plugins/test_datastore.py
```diff
@@ -636,6 +639,44 @@ class DefaultModel(Model):
...
+
+        assert len(await ds.query("test.stringprimarykey", [["string_id", "=", pk]])) == 1
+
+
+@pytest.mark.asyncio
+async def test__insert_default_integer_pk_value():
+    async with datastore_test() as ds:
+        pk = await ds.insert("test.default", {"string": "VALUE"})
+
+        assert len(await ds.query("test.default", [["id", "=", pk]])) == 1
```
themylogin, line 669: What does this test check for?

src/middlewared/middlewared/pytest/unit/plugins/test_datastore.py
```diff
@@ -636,6 +639,44 @@ class DefaultModel(Model):
...
+    integer_id = sa.Column(sa.String(100), primary_key=True)
+    value = sa.Column(sa.Integer(), nullable=True)
+
+
+@pytest.mark.asyncio
+async def test__insert_string_pk_record():
+    async with datastore_test() as ds:
+        pk = await ds.insert("test.stringprimarykey", {"string_id": "unique_key", "value": 1})
+
+        assert len(await ds.query("test.stringprimarykey", [["string_id", "=", pk]])) == 1
```
themylogin, line 661: Won't it be better to assert returned value from ds.insert?

sonicaj: I think this is better as we ensure that the value returned is actually the primary key as reflected by the database and not somehow just being returned from `ds.insert`

src/middlewared/middlewared/pytest/unit/plugins/test_datastore.py
sonicaj: So it checks to ensure the case where we have integer as the primary key

src/middlewared/middlewared/pytest/unit/plugins/test_datastore.py
themylogin: We can do both asserts

themylogin: Tests like this tend to hide situations where two wrongs make one right (but are still wrongs)

PR #4510, opened 2020-04-03
NAS-104333 / 12.0 / Proper service autostart in Linux

themylogin:


src/middlewared/middlewared/etc_files/hostname.py
```diff
@@ -0,0 +1,10 @@
+import logging
+
+from middlewared.utils import run
+
+logger = logging.getLogger(__name__)
```
sonicaj, line 5: Are we logging ?

sonicaj: Or maybe we should log if `systemctl` command fails for some reason ?

themylogin: It's just mine (and many other's) standard .py file template

PR #4511, opened 2020-04-03
NAS-105227 / 12.0 / FreeNAS enclosures

themylogin:


src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
```diff
@@ -19,6 +19,17 @@
             MappingSlot(0, 4, False),
         ]),
     ]),
+    ProductMapping(re.compile(r"FREENAS-MINI-3.0-X\+?"), [
```
william-gr, line 22: the X+ will probably have different mapping because its a different motherboard. Will need to wait to confirm though

PR #4523, opened 2020-04-07
NAS-105316 / 12.0 / SCALE migration

themylogin:


src/freenas/usr/local/sbin/ix-update
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash -e
+
+readonly UPDATE_FAILED_LOG="/data/update.failed"
+
+if [ ! -f /data/need-update ]
+then
+    exit 0
+fi
+
+has_alembic=$(echo "SELECT name FROM sqlite_master WHERE name = 'alembic_version'" | sqlite3 /data/freenas-v1.db)
```
sonicaj, line 10: Are we supporting upgrades from 11.3 -> TN Scale ? If no, I think it's safe to assume that we are going to have alembic ?

themylogin: Yes, we are supporting these, that's why we have freenas-migrate113 debian package :)

themylogin: @william-gr why?

william-gr: Big jumps like that are the recipe for failure. We simply cant properly support that.
We need to eventually add some coded to 12 for hardware dependent bits to allow a smooth upgrade (e.g. interface card names) that cannot be made only through db migration.
That also includes HA/CARP

themylogin: We already have that types of migrations, e.g. https://github.com/freenas/freenas/pull/4524

william-gr: No, thats not what I meant.

For example, interfaces are identified by name. They are not the same in SCALE. We need someway to identify when being OS agnostic, e.g. mac address.
We dont have that in 11.2 or 11.3. And we cannot do that only via db migration, we first need to run middlewared on 12 to do that kind of thing before loading SCALE.
We cannot load a 11.2 or 11.3 database into SCALE and expect it to work.

src/freenas/usr/local/sbin/ix-update
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -e
+
+readonly UPDATE_FAILED_LOG="/data/update.failed"
+
+if [ ! -f /data/need-update ]
+then
+    exit 0
+fi
+
+migrate > $UPDATE_FAILED_LOG 2>&1
```
sonicaj, line 10: I doubt that this is going to fail but still for any reason if it might, I think we are not going to exit the script which means that we remove the error log below anyways. We could perhaps check return code ?

themylogin: This script is `#!/bin/bash -e`, it will exit on first command failure.

PR #4524, opened 2020-04-07
Migrate bsd shells to linux

themylogin:


themylogin:
@sonicaj these modules are loaded by `load_modules` so same rules apply as with plugins: if module name ends with `_linux` it is not loaded on FreeBSD.

PR #4529, opened 2020-04-08
NAS-105644 / 11.3 / Fix traceback in cert_services_validation

anodos325:


src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -852,7 +852,7 @@ def cert_issuer(cert):
                     f'{cert["name"]} certificate is malformed'
                 )
 
-            if not cert['key_length']:
+            if not cert.get('key_length'):
```
sonicaj, line 855: We should make changes to ensure that `key_length` and `key_type` is there when there is no private key as that is the only case when this key is not going to be there. Secondly we should add another check here to ensure the cert in question has a private key and add a verror accordingly

sonicaj:
backport

PR #4531, opened 2020-04-08
NAS-104383 / 12.0 / QE-NAS-104383 Added API tests for native ZFS encryption

ericbsd:


tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+    results = POST('/pool/dataset/', payload)
+    assert results.status_code == 200, results.text
+
+
+def test_07_delete_not_encrypted_dataset():
+    results = DELETE(f'/pool/dataset/id/{dataset_url}/')
+    assert results.status_code == 200, results.text
+
+
+def test_08_create_a_dataset_with_inherit_encryption_true_on_a_normal_pool():
```
sonicaj, line 108: So the parent here is not encrypted, let's please verify that the dataset created states the same

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+            'generate_key': False,
+            'pbkdf2iters': 100000,
+            'algorithm': 'AES-128-CCM',
+            'passphrase': 'my_passphrase',
+        },
+        'encryption': True,
+        'inherit_encryption': False
+    }
+    results = POST('/pool/dataset/', payload)
+    assert results.status_code == 200, results.text
```
sonicaj, line 64: Let's please assert that the dataset is indeed encrypted by `PASSPHRASE`

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+        'id': dataset,
+        'change_key_options': {
+            'key': dataset_token_hex,
+        }
+    }
+    results = POST(f'/pool/dataset/change_key/', payload)
+    assert results.status_code == 200, results.text
+    job_id = results.json()
+    job_status = wait_on_job(job_id, 120)
+    assert job_status['state'] == 'SUCCESS', str(job_status['results'])
```
sonicaj, line 86: Let's please assert that the dataset in question is now encrypted with `HEX`. ( Please have a peek at `key_format` attribute in the data returned )

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+    assert results.status_code == 200, results.text
+
+
+def test_06_create_not_encrypted_dataset_on_a_normal_pool():
+    payload = {
+        'name': dataset,
+        'encryption': False,
+    }
+    results = POST('/pool/dataset/', payload)
+    assert results.status_code == 200, results.text
```
sonicaj, line 100: Let's assert that the dataset is indeed not encrypted

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+    payload = {
+        'name': dataset,
+        'encryption_options': {
+            'generate_key': True,
+        },
+        'encryption': True,
+        'inherit_encryption': False
+    }
+    results = POST('/pool/dataset/', payload)
+    assert results.status_code == 200, results.text
```
sonicaj, line 175: Let's please assert that the dataset is indeed encrypted with `HEX`

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+
+
+@skip_for_ha
+def test_14_delete_generate_key_encrypted_root():
+    results = DELETE(f'/pool/dataset/id/{dataset_url}/')
+    assert results.status_code == 200, results.text
+
+
+@skip_for_ha
+def test_15_create_an_encrypted_root_with_generate_key():
```
sonicaj, line 185: Please correct the function name to reflect the reality of the operation being performed

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+    payload = {
+        'name': dataset,
+        'encryption_options': {
+            'key': dataset_token_hex,
+        },
+        'encryption': True,
+        'inherit_encryption': False
+    }
+    results = POST('/pool/dataset/', payload)
+    assert results.status_code == 200, results.text
```
sonicaj, line 195: Let's please assert to ensure dataset is indeed encrypted with `HEX` key here

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+        'id': dataset,
+        'change_key_options': {
+            'passphrase': 'my_passphrase'
+        }
+    }
+    results = POST('/pool/dataset/change_key/', payload)
+    assert results.status_code == 200, results.text
+    job_id = results.json()
+    job_status = wait_on_job(job_id, 120)
+    assert job_status['state'] == 'SUCCESS', str(job_status['results'])
```
sonicaj, line 223: Let's please assert that it is indeed `PASSPHRASE` encrypted now

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+    }
+    results = POST('/pool/dataset/lock', payload)
+    assert results.status_code == 200, results.text
+
+
+@skip_for_ha
+def test_19_verify_passphrase_encrypted_root_is_locked():
+    payload = {
+        'id': dataset
+    }
```
sonicaj, line 242: Let's please pass the `passphrase` for the dataset as well and then ensure that `unlock_successful` attribute says `True` which means that if we unlock the dataset in question, it will indeed get unlocked. It might be nice to test with a wrong passphrase as well to ensure the data reported by middleware is accurate.

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+
+
+@skip_for_ha
+def test_24_create_a_passphrase_encrypted_pool():
+    global pool_id
+    payload = {
+        'name': pool_name,
+        'encryption': True,
+        'encryption_options': {
+            'algorithm': 'AES-128-CCM',
```
sonicaj, line 328: Let's please assert the `algorithm` to ensure that the specified one is indeed being used

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+        'id': dataset,
+        'change_key_options': {
+            'key': dataset_token_hex,
+        }
+    }
+    results = POST(f'/pool/dataset/change_key/', payload)
+    assert results.status_code == 200, results.text
+    job_id = results.json()
+    job_status = wait_on_job(job_id, 120)
+    assert job_status['state'] == 'SUCCESS', str(job_status['results'])
```
sonicaj, line 374: This is going to fail ( please use latest nightlies to test this ), and we should mark it as such. It is not allowed for a passphrase encrypted root to have children encrypted roots which are encrypted with a `KEY`.

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+
+
+@skip_for_ha
+def test_27_create_a_not_encrypted_dataset_on_a_passphrase_encrypted_pool():
+    payload = {
+        'name': dataset,
+        'encryption': False,
+    }
+    results = POST('/pool/dataset/', payload)
+    assert results.status_code == 200, results.text
```
sonicaj, line 390: Please assert to verify that it indeed is not encrypted

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+
+
+@skip_for_ha
+def test_29_create_a_dataset_to_inherit_encryption_from_the_passphrase_encrypted_pool():
+    payload = {
+        'name': dataset,
+        'inherit_encryption': True
+    }
+    results = POST('/pool/dataset/', payload)
+    assert results.status_code == 200, results.text
```
sonicaj, line 406: Let's assert to ensure that the `encrypted_root` property reflects the name of the root dataset here

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+    payload = {
+        'name': dataset,
+        'encryption_options': {
+            'key': dataset_token_hex,
+        },
+        'encryption': True,
+        'inherit_encryption': False
+    }
+    results = POST('/pool/dataset/', payload)
+    assert results.status_code == 200, results.text
```
sonicaj, line 492: Please assert that the dataset is indeed encrypted with a `HEX` key and that it is indeed an encryption root itself

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+        'id': dataset,
+        'change_key_options': {
+            'passphrase': 'my_passphrase',
+        }
+    }
+    results = POST(f'/pool/dataset/change_key/', payload)
+    assert results.status_code == 200, results.text
+    job_id = results.json()
+    job_status = wait_on_job(job_id, 120)
+    assert job_status['state'] == 'SUCCESS', str(job_status['results'])
```
sonicaj, line 507: Please assert that it changed successfully

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+        'lock_options': {
+            'force_umount': True
+        }
+    }
+    results = POST('/pool/dataset/lock', payload)
+    assert results.status_code == 200, results.text
+
+
+@skip_for_ha
+def test_38_verify_passphrase_encrypted_root_is_locked():
```
sonicaj, line 523: We don't need to use `encryption_summary` to verify if it's locked, simply retrieving the data for the dataset in question will retrieve this information via `locked` attribute

sonicaj: However, as above we could probably use this call to ensure that `encryption_summary` does say that the dataset can be unlocked with the right passphrase

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+
+
+@skip_for_ha
+def test_42_create_an_not_encrypted_dataset_on_a_key_encrypted_pool():
+    payload = {
+        'name': dataset,
+        'encryption': False,
+    }
+    results = POST('/pool/dataset/', payload)
+    assert results.status_code == 200, results.text
```
sonicaj, line 594: Please verify that it is indeed not encrypted

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+
+
+@skip_for_ha
+def test_44_create_an_dataset_with_inherit_encryption_from_the_key_encrypted_pool():
+    payload = {
+        'name': dataset,
+        'inherit_encryption': True
+    }
+    results = POST('/pool/dataset/', payload)
+    assert results.status_code == 200, results.text
```
sonicaj, line 610: Please verify that it is indeed encrypted and it's encryption root is the parent ( which is root dataset in this case )

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,870 @@
...
+    job_status_result = job_status['results']['result']
+    for dictionary in job_status_result:
+        if dictionary['name'] == dataset:
+            assert dictionary['locked'] is True, str(job_status_result)
+            break
+    else:
+        assert False, str(job_status_result)
+
+
+def test_52_verify_the_chid_of_the_encrypted_root_parent_is_locked():
```
sonicaj, line 691: We can verify that the child is locked in the previous test as well using the same data returned

tests/api2/pool_dataset_encryption.py
ericbsd: > Let's please pass the `passphrase` for the dataset as well and then ensure that `unlock_successful` attribute says `True` which means that if we unlock the dataset in question, it will indeed get unlocked.

I am not sure what you mean there is a test already for that.

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,1006 @@
...
+                    'passphrase': 'my_passphrase'
+                }
+            ]
+        }
+    }
+    results = POST('/pool/dataset/unlock/', payload)
+    assert results.status_code == 200, results.text
+    job_id = results.json()
+    job_status = wait_on_job(job_id, 120)
+    assert job_status['state'] == 'SUCCESS', str(job_status['results'])
```
sonicaj, line 341: We should check that the `unlock` job actually says it unlocked - the data is present in the result retrieved for the job

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,1006 @@
...
```
sonicaj, line 658: We should verify the call returns the datasets unlocked correctly ( like above )

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,1006 @@
...
+                    'passphrase': 'my_passphrase2'
+                }
+            ]
+        }
+    }
+    results = POST('/pool/dataset/unlock/', payload)
+    assert results.status_code == 200, results.text
+    job_id = results.json()
+    job_status = wait_on_job(job_id, 120)
+    assert job_status['state'] == 'FAILED', str(job_status['results'])
```
sonicaj, line 845: Same here as well please

tests/api2/pool_dataset_encryption.py
```diff
@@ -0,0 +1,1006 @@
...
+                    'passphrase': 'my_passphrase2'
+                }
+            ]
+        }
+    }
+    results = POST('/pool/dataset/unlock/', payload)
+    assert results.status_code == 200, results.text
+    job_id = results.json()
+    job_status = wait_on_job(job_id, 120)
+    assert job_status['state'] == 'SUCCESS', str(job_status['results'])
```
sonicaj, line 889: Here too, please

PR #4533, opened 2020-04-08
NAS-103670 / 12.0 / Truecommand Cloud Secure Connection

sonicaj:
This PR introduces changes where we add a `truecommand` service to configure TN so it is able to access TC containers created by iX Portal via wireguard connection.


sonicaj:
@themylogin @william-gr can you please have a pass and let me know what we should change ? Just a heads up that this hasn't been tested yet with the real iX Portal API, so there might be minor tweaks required to make this work once we finally test it.

src/middlewared/middlewared/plugins/truecommand/__init__.py
```diff
@@ -0,0 +1,23 @@
+import asyncio
+
+
+async def _event_system(middleware, event_type, args):
+    if args['id'] == 'ready':
+        await middleware.call('truecommand.start_truecommand_service')
+    elif args['id'] == 'shutdown':
+        # Stop wireguard here please if we have it enabled
+        await middleware.call('truecommand.stop_truecommand_service')
```
william-gr, line 9: Do we have to care about that? Also wont the service be stopped anyway because of rc.d normal shutdown process?

src/middlewared/middlewared/plugins/truecommand/portal.py
```diff
@@ -0,0 +1,165 @@
...
+            ):
+                status_dict.update({
+                    'state': PortalResponseState.FAILED,
+                    'error': f'Malformed ACTIVE response received by iX Portal with {", ".join(response)} keys'
+                })
+            elif response['nas_pubkey'] != config['wg_public_key']:
+                status_dict.update({
+                    'state': PortalResponseState.FAILED,
+                    'error': f'Public key "{response["nas_pubkey"]}" of TN from iX Portal does not '
+                             f'match TN Config public key "{config["wg_public_key"]}".'
```
william-gr, line 127: Can we be a bit more verbose and say TrueNAS?

src/middlewared/middlewared/plugins/truecommand/portal.py
```diff
@@ -0,0 +1,165 @@
...
+    @job(lock='poll_ix_portal_api_truecommand')
+    async def poll_api_for_status(self, job):
+        await self.middleware.call('truecommand.set_status', Status.CONNECTING.value)
+        config = await self.middleware.call('datastore.config', 'system.truecommand')
+
+        while config['enabled']:
+            try:
+                status = await self.middleware.call('truecommand.poll_once', config)
+            except Exception as e:
+                status = {
```
themylogin, line 24: Too broad exception clause for method that can be aborted: you'll ignore CancelledError this way

src/middlewared/middlewared/plugins/truecommand/update.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+class TrueCommandModel(sa.Model):
+    __tablename__ = 'system_truecommand'
+
+    id = sa.Column(sa.Integer(), primary_key=True)
+    api_key = sa.Column(sa.String(128), default=None, nullable=True)
+    api_key_state = sa.Column(sa.String(128), default='DISABLED', nullable=True)
+    wg_public_key = sa.Column(sa.String(255), default=None, nullable=True)
+    wg_private_key = sa.Column(sa.String(255), default=None, nullable=True)
```
themylogin, line 21: Shouldn't it and all other secrets be `EncryptedText`?

src/middlewared/middlewared/plugins/truecommand/connection.py
```diff
@@ -0,0 +1,33 @@
...
+        try:
+            with async_timeout.timeout(timeout):
+                async with aiohttp.ClientSession(
+                    raise_for_status=True
+                ) as session:
+                    req = await session.post(
+                        self.PORTAL_URI,
+                        data=json.dumps(payload or {}),
+                        headers={'Content-type': 'application/json'},
+                        proxy=os.environ.get('http_proxy'),
```
william-gr, line 25: I think there is possible to have a https_proxy as well?

src/middlewared/middlewared/plugins/truecommand/connection.py
```diff
@@ -0,0 +1,33 @@
...
+class TruecommandAPIMixin:
+
+    PORTAL_URI = 'https://portal.ixsystems.com/api'
+
+    async def _post_call(self, options=None, payload=None):
+        options = options or {}
+        timeout = options.get('timeout', 15)
+        response = {'error': None, 'response': {}}
+        try:
+            with async_timeout.timeout(timeout):
```
william-gr, line 17: Just making sure this is to be `with` and not `async with` ?

src/middlewared/middlewared/plugins/truecommand/wireguard.py
```diff
@@ -0,0 +1,90 @@
...
+        public_key, stderr = await cp.communicate(input=private_key)
+        if cp.returncode:
+            raise CallError(
+                f'Failed to generate public key for wireguard with exit code ({cp.returncode}): {stderr.decode()}'
+            )
+
+        return {'wg_public_key': public_key.decode().strip(), 'wg_private_key': private_key.decode().strip()}
+
+    @private
+    @periodic(1800, run_on_start=False)
```
william-gr, line 38: HEALTH_CHECK_SECONDS instead of 1800?

src/middlewared/middlewared/plugins/truecommand/connection.py
sonicaj:
```
aiohttp supports plain HTTP proxies and HTTP proxies that can be upgraded to HTTPS via the HTTP CONNECT method. aiohttp does not support proxies that must be connected to via https://.
```
I think we should be safe, or do you think that is not the case ?

src/middlewared/middlewared/plugins/truecommand/wireguard.py
william-gr: @sonicaj this was resolved but its still 1800 and no explanation why ;)

sonicaj: Because I have fixed it locally and hadn't pushed, waiting for your reply on the first comment ;) ( i usually mark comments as resolved when I have fixed them locally so as to know which comments are left to take care of, so no worries, pushing asap )

src/middlewared/middlewared/plugins/truecommand/update.py
```diff
@@ -0,0 +1,176 @@
...
+from .enums import Status, StatusReason
+
+TRUECOMMAND_UPDATE_LOCK = asyncio.Lock()
+
+
+class TrueCommandModel(sa.Model):
+    __tablename__ = 'system_truecommand'
+
+    id = sa.Column(sa.Integer(), primary_key=True)
+    api_key = sa.Column(sa.String(128), default=None, nullable=True)
```
themylogin, line 18: Shouldn't this be encrypted?

src/middlewared/middlewared/plugins/truecommand/update.py
```diff
@@ -0,0 +1,176 @@
...
+    api_key_state = sa.Column(sa.String(128), default='DISABLED', nullable=True)
+    wg_public_key = sa.Column(sa.String(255), default=None, nullable=True)
+    wg_private_key = sa.Column(sa.EncryptedText(), default=None, nullable=True)
+    wg_address = sa.Column(sa.String(255), default=None, nullable=True)
+    tc_public_key = sa.Column(sa.String(255), default=None, nullable=True)
```
themylogin, line 23: And this?

src/middlewared/middlewared/alembic/versions/12.0/2020-04-16_07-59_truecommand.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def upgrade():
+    op.create_table(
+        'system_truecommand',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('api_key', sa.String(length=128), nullable=True),
+        sa.Column('api_key_state', sa.String(length=128), nullable=True),
+        sa.Column('wg_public_key', sa.String(length=255), nullable=True),
+        sa.Column('wg_private_key', sa.String(length=255), nullable=True),
```
themylogin, line 26: IIRC `sa.EncryptedText()` produces `sa.Text()`?

src/middlewared/middlewared/alert/source/truecommand.py
```diff
@@ -0,0 +1,47 @@
+from middlewared.alert.base import Alert, AlertClass, AlertCategory, AlertLevel, OneShotAlertClass
+
+
+class TruecommandConnectionDisabledAlertClass(AlertClass, OneShotAlertClass):
+    deleted_automatically = False
+
+    category = AlertCategory.SYSTEM
+    level = AlertLevel.CRITICAL
+    title = 'Truecommand API Key Disabled by iX Portal'
```
themylogin, line 9: Isn't it's brand name True**C**ommand?

src/middlewared/middlewared/plugins/truecommand/connection.py
themylogin: Use `aiohttp.ClientSession(trust_env=True)` instead of this

src/middlewared/middlewared/plugins/truecommand/update.py
sonicaj: Should public key be encrypted ?

src/middlewared/middlewared/alembic/versions/12.0/2020-04-16_07-59_truecommand.py
```diff
@@ -0,0 +1,38 @@
...
+TrueCommand Service Model
+
+Revision ID: 58339783792c
+Revises: 171f5b91c36e
+Create Date: 2020-04-06 07:59:08.663553+00:00
+
+"""
+from alembic import op
+import sqlalchemy as sa
+import middlewared.sqlalchemy as m_sa
```
themylogin, line 11: You should never import anything that can alter it's behavior in migrations, they are meant to be written once and then never changed.

src/middlewared/middlewared/alembic/versions/12.0/2020-04-16_07-59_truecommand.py
```diff
@@ -0,0 +1,38 @@
...
+down_revision = '171f5b91c36e'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'system_truecommand',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('api_key', m_sa.EncryptedText(), nullable=True),
```
themylogin, line 24: If you regenerate this migration, it will yield `sa.Text()` here, why have you decided to change it like this?

PR #4537, opened 2020-04-09
NAS-105664 / 12.0 / Raise an alert for already expired certificates

sonicaj:


sonicaj:
@themylogin I think that would be too noisy for the user. Thoughts ?

sonicaj:
The user who reported this had around 4-5 expired certs and had no intention of removing them

sonicaj:
Actually we should, user can just choose to ignore those alerts if he finds them noisy. I'll make amends, thank you

PR #4539, opened 2020-04-09
NAS-105615 / 12.0 / Add logs to determine if we are able to retrieve rotation rate for disks

sonicaj:


src/middlewared/middlewared/plugins/device_/device_info_freebsd.py
```diff
@@ -67,6 +67,11 @@ def retrieve_geom_class(self, class_name):
                     disk['type'] = 'SSD'
                 else:
                     disk['type'] = 'HDD'
+            else:
+                self.middleware.logger.debug(
+                    'Unable to retrieve rotation rate for %s. Rotation rate reported by DISK geom is "%s"',
+                    disk['name'], disk_geom.provider.config.get('rotationrate')
+                )
```
william-gr, line 74: Why did we get a ValueError exception in the 11.3 code?

sonicaj: So for USB i have seen `rotationrate` being `unknown`, that's possibly why and that's why I log the value of `rotationrate` if it's there

PR #4547, opened 2020-04-12
Fix share_to_smbconf - path_suffix with domain

Lokarde:
Domain variable is %D not %d
%d is for server process id
Src: https://www.samba.org/samba/docs/using_samba/ch06.html#samba2-CHP-6-TABLE-2


anodos325:
Good catch on the typo.

PR #4551, opened 2020-04-13
NAS-102818 / 12.0 / autotune: Prune stale ZFS tuning

freqlabs:
These values are now the defaults in OpenZFS.


william-gr:
Are these sysctls gone or just unused? We may need a migration to remove these sysctls from database

freqlabs:
The sysctls are still there, we just aren't changing them from the default values anymore.

PR #4561, opened 2020-04-14
Add flag to ignore model specific registers in libvirt

sonicaj:
( keeping in line with 11.3 )


src/middlewared/middlewared/plugins/vm.py
```diff
@@ -297,7 +297,7 @@ def construct_xml(self):
                 'features', attribute_dict={
                     'children': [
                         create_element('acpi'),
-                        create_element('apic'),
```
william-gr, line 300: Curious why removing apic?

sonicaj: Because it translates to `-I` flag which is non-existent in `bhyve`

PR #4563, opened 2020-04-14
NAS-105726 / 12.0 / loader.conf: Increase size of kernel stacks

freqlabs:
We can have some pretty deep stacks in ZFS.  Increase the stack size to
avoid double faults.

Ticket: https://jira.ixsystems.com/browse/NAS-105726


freqlabs:
Addressing this in the debug kernel config instead.

PR #4571, opened 2020-04-15
Add utility functions to convert local filesystem ACL into a Windows SD

anodos325:


src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -693,10 +714,33 @@ async def sid_to_name(self, sid_str):
...
+    async def unixid_to_sid(self, data):
+        """
+        Samba generates SIDs for local accounts that lack explicit mapping in
+        passdb.tdb or group_mapping.tdb with a prefix of S-1-22-1 (users) and
+        S-1-22-2 (groups). This is not returned by wbinfo, but for consistency
+        with what appears when viewed over SMB protocol we'll do the same here.
+        """
+        unixid = data.get("id")
+        id = IDType[data.get("id_type", "GROUP")]
+        wb = None
```
yocalebo, line 726: Is there a possibility that the `wb` variable returns errno != 0 but has no data in it? If that happens, the `wb.decode().strip()` will fail because it will be of None type, right?

Maybe before returning wb.decode() check to make sure it's `isinstance(wb, bytes)`?

william-gr: If thats the case it will be empty string `b''`, never None as far as i can tell

we could just remove wb = None since its superfluous

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -693,10 +714,33 @@ async def sid_to_name(self, sid_str):
...
+
         if wb.returncode != 0:
-            self.logger.debug("wbinfo failed with error: %s",
-                              wb.stderr.decode().strip())
+            self.logger.warning("Could not convert [%d] to SID: %s",
+                                unixid, wb.stderr.decode().strip())
+            if WBCErr.DOMAIN_NOT_FOUND.err() in wb.stderr.decode():
+                is_local = await self.middleware.call(
+                    f'{"user" if id == IDType.USER else "group"}.query',
+                    [(f'{"uid" if id == IDType.USER else "gid"}', '=', unixid)],
```
william-gr, line 738: f-string seems feels odd

just `"uid" if id == IDType.USER else "gid"` would do it

anodos325: I agree.

PR #4573, opened 2020-04-16
fix(middlewared): asyncio popen in python 3.8

william-gr:


william-gr:
@themylogin it was setting utf8 by default but i dont think it ever worked. All usages I could see would still call decode() after communicate

william-gr:
It was good, there are some places we should probably remove encoding from Popen, I will do that now

PR #4575, opened 2020-04-16
NAS-105767 / 12.0 / Fix usages of sysctl module to use subprocess instead

sonicaj:
We have an issue with the underlying sysctl module not setting the values correctly and resulting in a SIGSEGV, so we instead use subprocess here to set the sysctl values.


sonicaj:
backport

PR #4577, opened 2020-04-16
NAS-105768 / 12.0 / Fix serial console settings

freqlabs:
The installer tries to detect when we are booting from the serial
console menu entry so it can persist that choice in loader.conf.

With TN12 the efi console driver in loader now has built in support
for the EFI serial protocol.  Enabling comconsole and efi both at the
same time leads to doubled output on the serial port when firmware
presents a serial port via EFI.  In some cases the efi serial console
can be disabled by turning off console redirection in system firmware.
In other cases, such as in bhyve, efi only has a serial console and
there is no video console.  There is also hardware (X-series) which only
has access to BIOS via serial console redirection, so it cannot be
disabled.

Don't enable efi console when we are doing a UEFI installation and
console was set to comconsole.  The efi console automatically uses the serial
port which may duplicate comconsole output.  We need to use comconsole
because efi does not necessarily have access to the selected serial port
and the port and speed cannot be configured.

When generating loader.conf.local, check if efi ConOut has a serial port
and if so do not enable the efi console in loader.

Also enable loader_color in loader.conf, which enables ANSI code output
in loader so that terminal settings can be reset to a clean state when
loader is initializing.  This fixes some common issues with the menu
not being visible after earlier boot stages leave the terminal in a
weird state.

### DIFF
--- a/src/freenas-installer/etc/install.sh
+++ b/src/freenas-installer/etc/install.sh
@@ -374,37 +374,71 @@ install_loader()
     return 0
 }
 
+nasdb()
+{
+    local mnt=$1
+    local query=$2
+
+    chroot "${mnt}" /usr/local/bin/sqlite3 /data/freenas-v1.db "${query}"
+}
+
+videoconsole()
+{
+    if [ "$BOOTMODE" = "UEFI" ]; then
+	echo efi
+    else
+	echo vidconsole
+    fi
+}
+
 save_serial_settings()
 {
-    _mnt="$1"
+    local mnt="$1"
 
-    # If the installer was booted with serial mode enabled, we should
-    # save these values to the installed system
-    USESERIAL=$((`sysctl -n debug.boothowto` & 0x1000))
-    if [ "$USESERIAL" -eq 0 ] ; then return 0; fi
+    local RB_SERIAL=0x1000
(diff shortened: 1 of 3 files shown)


william-gr:
@amotin notices this PR open for a few days. Are we waiting on something before merging this?

freqlabs:
The first commit was tested in bhyve and on X-series.
In bhyve with UEFI firmware the console output looks good in the installer and installed system now via the default boot entry.
On X-series we tested BIOS mode via the serial boot entry to confirm no regression in the existing production workflow. In UEFI mode the default boot entry works correctly.

When the serial console is configured by middleware however, we do still see the output on the serial console doubled. Disabling serial console redirection is not a viable workaround for X-series because there is no video console to access BIOS from, so disabling serial console redirection would lock you out. Furthermore, it does not address the issue for bhyve.

Middleware should not configure efi and boot_multicons in loader.conf.local on UEFI when only a serial console is present. Those should also not be set in the loader environment by the serial boot menu entry in this case. Then the installer can detect a serial-only boot configuration and enable the serial console in the database without issue.

The second commit adds logic to middleware so that we enable only comconsole when efi has a serial port.

PR #4583, opened 2020-04-17
Allow making plugin directories be platform specific

sonicaj:


src/middlewared/middlewared/utils/__init__.py
```diff
@@ -350,6 +350,9 @@ def load_modules(directory, base=None, depth=0):
 
     for f in dirs:
         if depth > 0:
+            if any(f.endswith(f'_{suffix}') for suffix in ('freebsd', 'linux')):
```
william-gr, line 353: f.endswith(('_freebsd', '_linux')) is easier :P

sonicaj: True that

PR #4584, opened 2020-04-17
NAS-103598 / 12.0 / initial API for adding gluster volumes

yocalebo:


src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -0,0 +1,569 @@
...
+
+    class Config:
+        namespace = 'gluster.volume'
+
+    def __volume_wrapper(self, method, args=None, kwargs=None):
+
+        result = b''
+
+        try:
+            if args and not kwargs:
```
themylogin, line 22:
```
args = args or ()
kwargs = kwargs or {}
```
And you won't need these checks

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -0,0 +1,569 @@
...
+                result = method(**kwargs)
+            if args and kwargs:
+                result = method(*args, **kwargs)
+            if not args and not kwargs:
+                result = method()
+        except GlusterCmdException as e:
+            rc, out, err = e.args[0]
+            err = err if err else out
+            raise CallError(f'{err.decode().strip()}')
+        except Exception:
```
themylogin, line 34: This is not necessary

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -0,0 +1,569 @@
...
+
+    @private
+    def create_volume(self, data):
+
+        volname = data.pop('volname')
+        volbricks = data.pop('volbricks')
+
+        bricks = []
+        for i in volbricks:
+            peer = i.get('peer_name')
```
themylogin, line 57: `get` implies that key can be absent and we'll be happy with `None`, but we won't, `brick = peer + ':' + path` will crash if `peer` or `path` are `None`, and the problem is that exception you'll get will be less helpful than `KeyError`.

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -0,0 +1,569 @@
...
+        return result
+
+    @private
+    def restart_volume(self, data):
+
+        volname = data.pop('volname')
+
+        args = (volname,)
+
+        result = self.__volume_wrapper(volume.restart, args, kwargs=data)
```
themylogin, line 108: Well... why don't you use args and kwargs the usual way?

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -0,0 +1,569 @@
...
+        self.common_validation(job)
+
+        result = self.list_volumes()
+
+        return result
+
+    @accepts(Dict(
+        'glustervolume_optreset',
+        Str('volname', required=True),
+        Str('opt', default=None),
```
themylogin, line 413: You'll also need `null=True` so `null` value could be passed explicitly

src/middlewared/middlewared/plugins/gluster_linux/volume.py
yocalebo: Thank you!

src/middlewared/middlewared/plugins/gluster_linux/volume.py
yocalebo: Thanks, should be fixed in latest PR.

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -14,25 +14,19 @@ class GlusterVolumeService(CRUDService):
     class Config:
         namespace = 'gluster.volume'
 
-    def __volume_wrapper(self, method, args=None, kwargs=None):
+    def __volume_wrapper(self, method, *args, **kwargs):
 
         result = b''
 
+        args = args or ()
```
themylogin, line 21: Not needed anymore

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -81,9 +72,7 @@ def start_volume(self, data):
 
         volname = data.pop('volname')
 
-        args = (volname,)
-
-        result = self.__volume_wrapper(volume.start, args, kwargs=data)
+        result = self.__volume_wrapper(volume.start, volname, data)
```
themylogin, line 75: `**data`?

src/middlewared/middlewared/plugins/gluster_linux/volume.py
yocalebo: @themylogin `peer_name` and `peer_path` will be enforced at schema level and require that those are populated with data.

themylogin: SO it should be `i['peer_name']` then

yocalebo: thanks!

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -0,0 +1,544 @@
...
+        'glustervolume_create',
+        Str('volname', required=True),
+        List('volbricks', items=[
+            Dict(
+                'brick',
+                Str('peer_name', required=True),
+                Str('peer_path', required=True),
+            ),
+        ], required=True),
+        Int('replica', required=False),
```
themylogin, line 222: `required=False` is the default, no need for visual noise

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -0,0 +1,542 @@
...
+            volname,
+            src_brick,
+            new_brick,
+            **data)
+
+        return result
+
+    @accepts(Dict(
+        'glustervolume_create',
+        Str('volname', required=True),
```
william-gr, line 212: Since the namespace already is called gluster.volume, perhaps it could be just "name" and "bricks", removing the suffix? Thoughts?

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -0,0 +1,542 @@
...
+        self.common_validation(job)
+
+        return self.create_volume(data)
+
+    @accepts(Dict(
+        'glustervolume_start',
+        Str('volname', required=True),
+        Bool('force'),
+    ))
+    @job(lock='glustervolume_start')
```
william-gr, line 256: Is the volume name always the unique identifier for all operations related to volume?

If thats the case I think we should change all these methods to comply with other CRUD services we have, and so the REST API is based on that.

e.g.

@item_method
def start(self, job, id, options):
  ...

For anything other than create/update/delete that uses the volume name as unique identifier.

That way for REST it would also work as:
/gluster/volume/id/{name}/start/

I dont know if it works with a job (never tried that), needs verification.

src/middlewared/middlewared/plugins/gluster_linux/utils.py
```diff
@@ -0,0 +1,36 @@
+def validate_gluster_jobs(self, verrors, job):
+    """
+    Gluster CLI operations need to run synchronously.
+    """
```
william-gr, line 4: Curious about why we choose to raise an error when another operation is already running instead of simply make it wait?

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -0,0 +1,542 @@
...
+    def do_delete(self, job, data):
+        """
+        Delete a gluster volume.
+
+        `volname` Name of the volume to be deleted
+        """
+
+        self.common_validation(job)
+
+        result = self.delete_volume(data)
```
william-gr, line 325: Also not sure how much it buys us to have methods like that, which all it does is call the wrapper.

Why not do

```
result = self.__volume_wrapper(volume.delete, data['volname'])
```
here

src/middlewared/middlewared/plugins/gluster_linux/utils.py
```diff
@@ -0,0 +1,36 @@
...
+
+    peer_job = self.middleware.call_sync(
+        'core.get_jobs', [
+            ['method', '^', 'gluster.peer'],
+            ['state', 'in', ['RUNNING']],
```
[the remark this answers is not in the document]
yocalebo: Thanks!

src/middlewared/middlewared/plugins/gluster_linux/peer.py
```diff
@@ -47,17 +50,20 @@ async def resolve_host_or_ip(self, hostname, verrors):
...
-                'gluster.peer.resolve_host_or_ip', hostname, verrors)
-        except Exception:
-            if verrors:
-                raise verrors
-            raise
+        if hostname:
+            try:
+                self.middleware.call_sync(
+                    'gluster.peer.resolve_host_or_ip', hostname, verrors)
+            except Exception:
```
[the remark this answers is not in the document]
yocalebo: Fixed it, thanks

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -0,0 +1,542 @@
...
+        result = b''
+
+        try:
+            result = method(*args, **kwargs)
+        except GlusterCmdException as e:
+            rc, out, err = e.args[0]
+            err = err if err else out
+            raise CallError(f'{err.decode().strip()}')
+
+        if isinstance(result, bytes):
```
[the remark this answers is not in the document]
yocalebo: This all ends up being run using `subprocess`. I've found, for example, if you try to `start` a gluster volume that has already been started it will not raise an exception, it simply returns `stdout` so I need to decode it at that point.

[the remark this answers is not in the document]
william-gr: Right, but why the isinstance call? isnt it always supposed to return bytes?

src/middlewared/middlewared/plugins/gluster_linux/utils.py
```diff
@@ -0,0 +1,36 @@
+def validate_gluster_jobs(self, verrors, job):
+    """
+    Gluster CLI operations need to run synchronously.
+    """
+
+    peer_job = self.middleware.call_sync(
+        'core.get_jobs', [
+            ['method', '^', 'gluster.peer'],
+            ['state', '=', ['RUNNING']],
```
william-gr, line 9: just 'RUNNING', no array

src/middlewared/middlewared/plugins/gluster_linux/volume.py
yocalebo: Agree 100%, fixed.

src/middlewared/middlewared/plugins/gluster_linux/utils.py
yocalebo: I actually don't know how to make the current job wait for another job using middlewared facilities. I saw that there is a `core.wait_job` method that I can then run `wait_sync` method on but I found 2 issues.

1. the current job doesn't actually wait for the other job to finish
2. `core.get_jobs` has a global timeout of 60 seconds and I didn't see an easy way of changing that

Again, however, I might be missing something obvious that's already been baked into middlewared, I just couldn't find it.

src/middlewared/middlewared/plugins/gluster_linux/volume.py
yocalebo: I think I see what you're saying. So instead of `glustervolume_create` change it to `name_create` inside the `@accepts(Dict(`? And then do the same for the brick operations too?

src/middlewared/middlewared/plugins/gluster_linux/utils.py
william-gr: Thats the "lock" parameter of a job.

If all the gluster jobs use the same lock name, only one of these jobs can run at the same given time. They do not to be unique across them.

src/middlewared/middlewared/plugins/gluster_linux/volume.py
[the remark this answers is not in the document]
william-gr: Fair enough, thanks!

src/middlewared/middlewared/plugins/gluster_linux/utils.py
yocalebo: Oh! Didn't even think about doing it that way. Great idea!

src/middlewared/middlewared/plugins/gluster_linux/volume.py
william-gr: No, I was saying `name` instead of `volname`, and `bricks` instead of `volbricks` as attribute/param name

yocalebo: Understood, I mostly stuck those names because the called methods use those names. However, most of the time those names didn't matter since the called method uses those parameters as positional args anyways.

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -0,0 +1,456 @@
...
+
+        return result
+
+    @accepts(Dict(
+        'glustervolume_replacebrick',
+        Str('name', required=True),
+        Dict(
+            'src_brick',
+            Str('peer_name', required=True),
+            Str('peer_path', required=True),
```
william-gr, line 428: I think you have to set the Dict as required=True too

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -0,0 +1,456 @@
...
+            `peer_name` IP or DNS name of the peer
+            `peer_path` The full path of the brick
+
+        `operation` The operation to be performed
+            `START` Start the removal of the brick(s)
+            `STOP` Stop the removal of the brick(s)
+            `COMMIT` Commit the removal of the brick(s)
+            `STATUS` Display status of the removal of the brick(s)
+
+        `repica` Replica count
```
william-gr, line 414: typo

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -0,0 +1,456 @@
...
+        """
+        Add bricks to a gluster volume.
+
+        `name` Gluster volume name
+        `bricks` List of brick paths.
+            `peer_name` IP or DNS name of the peer
+            `peer_path` The full path of the brick to be added
+
+        `stripe` Stripe count
+        `repica` Replica count
```
william-gr, line 376: typo

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -0,0 +1,456 @@
...
+        """
+
+        result = self.__volume_wrapper(volume.vollist)
+
+        return result
+
+    @accepts(Dict(
+        'glustervolume_optreset',
+        Str('name', required=True),
+        Str('opt'),
```
william-gr, line 307: Isnt opt required too? Or it will reset all options if not set?

yocalebo: It will reset all options if not set.

william-gr: Its even written in the docstring, duh :P sorry

src/middlewared/middlewared/plugins/gluster_linux/volume.py
yocalebo: Yes, the volume name is always the unique identifier. I'll make this comply like other REST APIs.

yocalebo: `@item_method` can be used with `@job` decorator. I confirmed testing it. I've pushed PR fixing this.

PR #4587, opened 2020-04-20
Reduce spam from rotation rate logs

sonicaj:


src/middlewared/middlewared/plugins/device_/device_info_freebsd.py
```diff
@@ -67,7 +67,7 @@ async def get_disk_details(self, disk, disk_geom):
                     disk['type'] = 'SSD'
                 else:
                     disk['type'] = 'HDD'
-            else:
+            elif disk_geom.provider.config.get('rotationrate') != 'unknown':
```
william-gr, line 70: What disk `type` is going to be in this case?

sonicaj: It's going to be `UNKNOWN` ( we default to that )

PR #4606, opened 2020-04-21
NAS-105680 / 11.3 / Fix rclone not liking paths like /mnt/data/deeply/nested/.zfs/snapshot/cloud_sync-1-20200421170217/.

themylogin:


src/middlewared/middlewared/plugins/cloud_sync.py
```diff
@@ -175,7 +175,7 @@ async def rclone(middleware, job, cloud_sync):
                 await middleware.call("zfs.snapshot.create", dict(snapshot, recursive=recursive))
 
                 relpath = os.path.relpath(path, dataset["mountpoint"])
```
william-gr, line 177: According to flake8 this isnt used anymore? did you mean to put relpath after snapshot_name below?

PR #4620, opened 2020-04-21
NAS-105839 / 12.0 / Fix disks on FreeBSD 12 and improve events logic

themylogin:


william-gr:
What do you want to backpor to 11.3, exactly?

themylogin:
@william-gr `include_expired` part, enclosure sync queries expired disks, it will crash `disk.sync_all` if disk is removed

PR #4630, opened 2020-04-23
Validate home directory path is unlocked during user creation

anodos325:
If dataset is locked and a path component of the user home directory
specified during dataset creation, then we may end up creating
a home dir that gets mounted over when the dataset is unlocked.

Add a function to check whether a given path contains a component that
is locked. Depending on how creative users are with symbolic links,
in the future we may need to add checks for both paths.

### DIFF
--- a/src/middlewared/middlewared/plugins/filesystem.py
+++ b/src/middlewared/middlewared/plugins/filesystem.py
@@ -12,7 +12,6 @@ import pwd
 import select
 import shutil
 import subprocess
-from pathlib import Path
 
 from middlewared.main import EventSource
 from middlewared.schema import Bool, Dict, Int, Ref, List, Str, UnixPerm, accepts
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -878,6 +879,21 @@ def setacl(self, job, data):
...
         job.set_progress(100, 'Finished setting ACL.')
 
+    @private
+    async def path_is_encrypted(self, path):
+        p = Path(path)
+        if not p.is_absolute():
+            raise CallError(f"[{path}] is not an absolute path.", errno.EINVAL)
+
+        rp = p.resolve()
+        our_pool = p.parts[2]
```
william-gr, line 889: Could raise an IndexError if anyone wrongfully pass /mnt, /foo

src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -878,6 +879,21 @@ def setacl(self, job, data):
...
+
+        for ds in await self.middleware.call('pool.dataset.query', [('pool', '=', our_pool)]):
```
william-gr, line 891: This will loop through all datasets, is there some way we can restrict to the path we are checking?

PR #4631, opened 2020-04-23
NAS-105851 / 12.0 / add zfs arc size to realtime reporting plugin

yocalebo:


src/middlewared/middlewared/plugins/reporting.py
```diff
@@ -1060,9 +1061,21 @@ def run(self):
...
 
+            # ZFS ARC Size (raw value is in Bytes)
+            data['zfs'] = {}
+            if osc.IS_FREEBSD:
+                data['zfs']['arc_size'] = sysctl.filter('kstat.zfs.misc.arcstats.size')[0].value
+            elif osc.IS_LINUX:
+                with open('/proc/spl/kstat/zfs/arcstats') as f:
+                    rv = f.read()
+                    for l in rv.split('\n'):
+                        if l.startswith('size'):
```
sonicaj, line 1076: How about using regex to do this ?

william-gr: regex is probably slower, but also likely cleaner
i am fine with either

yocalebo: Because I was curious, I tested the speed of doing it both ways to see how much slower regex was. I had a function do the `startswith` method and then I had a function compile a regex. Ran each function 1000 times and calculated total time it took to run. This is what I found:

0.0851650238037 <--- `startswith` method
0.0983519554138 <--- `regex` method

I'll stick to startswith for now :)

sonicaj: This is good as well ; ) Normally we do compilation earlier - so on each run, it is not compiled again and again reducing the time spent

PR #4641, opened 2020-04-24
NAS-105836 / 12.0 / Gracefully handle unexpected LDAP schemas

anodos325:
Windows LDAP servers (AD environment) have slightly different return to query for SambaDomain than we see on OpenLDAP server. Log a debug message with the contents that we failed to parse and then move on. In the future, it may make sense to add a validation error to redirect users to the AD form if a query to rootDSE shows it's an AD DC.

### DIFF
--- a/src/middlewared/middlewared/plugins/ldap.py
+++ b/src/middlewared/middlewared/plugins/ldap.py
@@ -320,17 +320,20 @@ class LDAPQuery(object):
         res = []
         for r in results:
             parsed_data = {}
-            for k, v in r[1].items():
-                try:
-                    v = list(i.decode() for i in v)
-                except Exception:
-                    v = list(str(i) for i in v)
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -320,17 +320,21 @@ def parse_results(self, results):
...
+                        v = list(i.decode() for i in v)
+                    except Exception:
+                        v = list(str(i) for i in v)
+                    parsed_data.update({k: v})
+
+                res.append({
+                    'dn': r[0],
+                    'data': parsed_data
+                })
+            except AttributeError:
```
sonicaj, line 335: Do you think we should perhaps check if `r[1]` exists and is a dict instead ? We can log the other case where that doesn't happen. It just seems clearer to see as to where the issue lies

PR #4642, opened 2020-04-24
NAS-105764 / 12.0 / LAGG interface uses MAC of first port; Reconfigure it if first port h…

themylogin:
…as changed and stop losing DB interface order by using set


themylogin:
backport

PR #4644, opened 2020-04-25
NAS-105906 / 12.0 / Only try to retrieve attachments for pools which are in available state

sonicaj:
If a pool is offline/faulted, we can't retrieve the data on those pools and in this case we should just return an empty list when trying to get attachments on the pool.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -1549,7 +1549,7 @@ class PoolService(CRUDService):
         share, asking for confirmation.
         """
         pool = await self.get_instance(oid)
-        return await self.middleware.call('pool.dataset.attachments', pool['name'])
+        return await self.middleware.call('pool.dataset.attachments_with_path', pool['path'])
 
     @item_method
     @accepts(Int('id'))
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1549,7 +1549,10 @@ async def attachments(self, oid):
         share, asking for confirmation.
         """
         pool = await self.get_instance(oid)
-        return await self.middleware.call('pool.dataset.attachments', pool['name'])
+        if pool['status'] in ('OFFLINE', 'FAULTED'):
```
william-gr, line 1552: I think we can still have attachments in these states? e.g. shares configured for the pool/dataset that is now gone

themylogin: What is the problem you are trying to solve?

sonicaj: @themylogin if the pool is not available or we are not able to get the metadata like list of datasets or if that list is malformed, `pool.dataset.attachments` fails as we try to find the dataset in question and then using that dataset's path we find related attachments etc. As @william-gr pointed out, we can still/should find attachments and I'll be making the necessary amends.

PR #4649, opened 2020-04-27
NAS-105472 / 12.0 / remove pf from truenas HA

yocalebo:


yocalebo:
I've done some research and found 1 scenario where failover behavior would change. There is a short-window when the standby controller is becoming active, and the MAC address of the VIP migrates to the switchport of the standby controller, the PF firewall sill silently drop connection requests from clients. If I remove this, then there is a possibility for which clients could take longer to reconnect to the shares. This is because the error that they would receive at the TCP level would change. (i.e. they could get RST from a closed port which could cause longer failover times). The firewall and how it operates, in theory, prevents this scenario from happening so I'm closing this PR.

PR #4657, opened 2020-04-28
NAS-105891 / 12.0 / Make local plugin version retrieval failure non fatal

sonicaj:
This commit introduces changes where we make sure that if for any reason we are not able to retrieve version of installed plugins, we make the failure non fatal and still allow plugin.query to succeed.

### DIFF
--- a/src/middlewared/middlewared/plugins/jail_freebsd.py
+++ b/src/middlewared/middlewared/plugins/jail_freebsd.py
@@ -673,7 +673,8 @@ class PluginService(CRUDService):
         try:
             conn = sqlite3.connect(db)
         except sqlite3.Error as e:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/jail_freebsd.py
```diff
@@ -673,7 +673,8 @@ def read_plugin_pkg_db(self, db, pkg):
         try:
             conn = sqlite3.connect(db)
         except sqlite3.Error as e:
-            raise CallError(e)
+            self.middleware.logger.error('Failed to connect to "%s" database : %s', db, str(e))
```
william-gr, line 676: for the future, we can use %r instead of "%s"

PR #4665, opened 2020-04-28
NAS-105930 / 12.0 / allow configuration of VIP when none exist

yocalebo:


src/middlewared/middlewared/plugins/network.py
```diff
@@ -644,8 +644,12 @@ async def __check_failover_disabled(self):
             return
         if await self.middleware.call('failover.status') == 'SINGLE':
             return
-        if not (await self.middleware.call('failover.config'))['disabled']:
-            raise CallError('Disable failover before performing interfaces changes.')
```
william-gr, line 648: Shouldnt failover.status be returning SINGLE if there is no VIP and HA is not configured yet? I wonder what it is returning and if thats not the real regression

yocalebo: No, not in this scenario. The system is being configured with an HA license and also has a zpool configured meaning it won't return `SINGLE`.

william-gr: So why does this work in 11.3?

william-gr:
```
        interfaces = await self.middleware.call('interface.query')
        if not any(filter(lambda x: x['state']['carp_config'], interfaces)):
            return 'SINGLE'
```

william-gr: It should be SINGLE if there is no carp configured

yocalebo: Not sure if this doesn't work on 11.3. I'll investigate there.

yocalebo:
This is an API bug so I will be opening a PR in the releng repo to have it fixed over there.

PR #4666, opened 2020-04-28
NAS-105795 / 11.3 / Delete SMB group mappings by SID value rather than NT group

anodos325:
If for some reason Samba's group_mapping.tdb ends up with a stale
entry, removal by NT group name can be impossible. Convert return
of smb.groupmap_list to dictionary keyed by unix group name. Use
this return to convert unix group name to SID to then perform the
groupmap deletion.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -323,14 +323,15 @@ class SMBService(SystemServiceService):
 
     @private
     async def groupmap_list(self):
-        groupmap = []
+        groupmap = {}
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -365,6 +366,30 @@ async def groupmap_add(self, group):
                 f'Failed to generate groupmap for [{group}]: ({gm_add.stderr.decode()})'
             )
 
+    @private
+    async def groupmap_delete(self, data):
+        ntgroup = data.get("ntgroup")
+        sid = data.get("sid")
+        if not ntgroup and not sid:
```
william-gr, line 373: should be or, not and?

PR #4673, opened 2020-04-29
Add to debug

anodos325:


themylogin:
@sonicaj If the middleware is crashed, we don't care for these missing bits. I like the code being simpler and exposing more necessary information.

william-gr:
backport

PR #4674, opened 2020-04-29
NAS-105747 / 12.0 / Encrypt SSH and certificate private keys

themylogin:
Let's use `EncryptedTextField` furthermore, avoiding `pwenc.encrypt/pwenc.decrypt` (from now on, no plugins use that, only DBAL)


themylogin:
@anodos325 can you take a look at this please to ensure that DS functionality is still alright?

anodos325:
> @anodos325 can you take a look at this please to ensure that DS functionality is still alright?

Looks good to me.

PR #4678, opened 2020-04-29
NAS-105951 / 11.3 / Add slight optimizations for adding user accounts

anodos325:
Only alter passdb entry on user update if updated information will
impact the entry. This means changing the username, changing the "locked"
status, or changing the password.

Remove extraneous SMB service reload on user update. This behavior appears
to stem from a past age when FreeNAS user the smbpasswd backend for
samba's passdb.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -349,6 +349,20 @@ class UserService(CRUDService):
 
         verrors.check()
 
+        must_change_pdb_entry = False
+        for k in ('username', 'password', 'locked'):
+            new_val = data.get(k)
+            old_val = user.get(k)
+            if new_val is not None and old_val != new_val:
+                if k == 'username':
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/account.py
```diff
@@ -334,6 +334,7 @@ async def do_update(self, pk, data):
             user['group'] = group['id']
 
         await self.__common_validation(verrors, data, 'user_update', pk=pk)
+        must_change_pdb_entry = True if data.get("password") or data.get("username") or data.get("locked") else False
```
themylogin, line 337: Should comparison with original state be used instead? IIRC, `username` will also be present in UI requests. Also, `locked` transition from `True` to `False` won't be captured like this.

anodos325: Indeed. I apparently had a brain-fart while writing that.

PR #4679, opened 2020-04-29
Ping TC IP address if handshake check is successful

sonicaj:
This commit introduces changes where we ping TC ip address to determine if we have an active connection with TC. Why we do this is because it's possible TC container changed it's IP address and the keys are still valid, which means we would get a handshake from the wireguard connection but we are not connected to the TC container anymore. In this case we should get in an error state and poll the portal to see if we should be using updated address for the TC container.

### DIFF
--- a/src/middlewared/middlewared/plugins/truecommand/update.py
+++ b/src/middlewared/middlewared/plugins/truecommand/update.py
@@ -49,8 +49,10 @@ class TruecommandService(ConfigService):
             else:
                 status_reason = 'Waiting for connection from Truecommand.'
 
-        if config['remote_address']:
-            config['remote_address'] = config['remote_address'].split('/', 1)[0]
+        config['remote_ip_address'] = config['remote_url'] = config.pop('remote_address')
+        if config['remote_ip_address']:
+            config['remote_ip_address'] = config.pop('remote_ip_address').split('/', 1)[0]
+            config['remote_url'] = f'http://{config["remote_ip_address"]}/'
 
         config.update({
             'status': self.STATUS.value,
--- a/src/middlewared/middlewared/plugins/truecommand/wireguard.py
+++ b/src/middlewared/middlewared/plugins/truecommand/wireguard.py
@@ -3,7 +3,7 @@ import subprocess
 import time
 
 from middlewared.service import accepts, CallError, no_auth_required, periodic, pass_app, private, Service, throttle
-from middlewared.utils import Popen, run
+from middlewared.utils import osc, Popen, run
 
 from .enums import Status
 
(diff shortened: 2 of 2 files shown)


src/middlewared/middlewared/plugins/truecommand/wireguard.py
```diff
@@ -84,6 +84,15 @@ async def wireguard_connection_health(self):
                     # We never established handshake with TC if timestamp is 0, otherwise it's been more
                     # then 30 minutes, error out please
                     health_error = True
+                else:
+                    # It's possible that IP of TC changed and we just need to get up to speed with the
+                    # new IP. So if we have a correct handshake, we should ping the TC IP to see if it's
+                    # still reachable
+                    config = await self.middleware.call('datastore.config', 'system.truecommand')
+                    cp = await run(['ping', '-c', '3', '-t', '5', '-q', str(config['remote_address'])], check=False)
```
william-gr, line 92: why -c 3?

yocalebo: if you remove `-c 3` and keep `-t 5` it means the TC can't be more than 1 hop away from us and that single packet has to be successful. Why not just do a `-c 4 or 5` ?

sonicaj: I am not sure I follow ? Idea is to keep pinging for a maximum of 5 seconds. If we do `-c 4/5`, that can potentially take more time then 5 seconds if the IP is not reachable

yocalebo: Hah, so -t on Debian is TTL setting and on freebsd is timeout setting......a subtle difference that changes the behavior. Disregard 😬

sonicaj: good catch ; ) I updated the PR to make sure it works on debian as well. Thank you for pointing it out

src/middlewared/middlewared/plugins/truecommand/update.py
```diff
@@ -50,7 +50,7 @@ async def tc_extend(self, config):
                 status_reason = 'Waiting for connection from Truecommand.'
 
         if config['remote_address']:
-            config['remote_address'] = config['remote_address'].split('/', 1)[0]
+            config['remote_address'] = f'http://{config["remote_address"].split("/", 1)[0]}/'
```
william-gr, line 53: I was thinking we could do both... IP and url, so UI doesnt have to parse either

src/middlewared/middlewared/plugins/truecommand/wireguard.py
```diff
@@ -99,7 +113,8 @@ async def connected(self, app):
         connected = Status(tc_config['status']) == Status.CONNECTED
         return {
             'connected': connected,
-            'truecommand_ip': tc_config['remote_address'] if connected else None,
+            'truecommand_ip': tc_config['remote_ip_address'] if connected else None,
+            'truecommand_address': tc_config['remote_address'] if connected else None,
```
william-gr, line 117: On second thought, dont you think address is confusing?

Address can refer to many things. IP Address, URL Address.
How about we use IP and URL to lessen the confusion?

sonicaj: Right again ; )

PR #4693, opened 2020-04-30
NAS-105928 / 12.0 / Make sure all devices are properly expanded in a pool

sonicaj:
1) Where we make sure we don't try to expand spare/cache devices as that is not supported by ZFS.
2) When we try to resize a partition on expanding pool, we should use the same alignment value ( `4k` ) which was used when we formatted the disk, if we don't do that, it tries to use the disk's native size which can result in the resizing failing.
3) We resize spare/cache disks anyways because for cache, whenever the pool is going to be imported again, zfs will register the new size. For spare, whenever the spare is going to be used, the new capacity will be registered.
4) We make sure that we properly expand MIRROR/RAIDZ vdevs and bring their relevant devices online accordingly.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/expand.py
+++ b/src/middlewared/middlewared/plugins/pool_/expand.py
@@ -59,8 +59,10 @@ class PoolService(Service):
             if osc.IS_FREEBSD:
                 sysctl.filter('kern.geom.debugflags')[0].value = 16
             geli_resize = []
+            # spare/cache devices cannot be expanded
+            topology = {k: v for k, v in pool['topology'].items() if k not in ('spare', 'cache')}
             try:
-                for vdev in sum(pool['topology'].values(), []):
+                for vdev in sum(topology.values(), []):
                     if vdev['type'] != 'DISK':
                         logger.debug('Not expanding vdev of type %r', vdev['type'])
                         continue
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/pool_/expand.py
```diff
@@ -59,8 +59,10 @@ async def expand(self, job, id, options):
             if osc.IS_FREEBSD:
                 sysctl.filter('kern.geom.debugflags')[0].value = 16
             geli_resize = []
+            # spare/cache devices cannot be expanded
+            topology = {k: v for k, v in pool['topology'].items() if k not in ('spare', 'cache')}
```
william-gr, line 63: Why cant they? dont we also have to resize their partition?

src/middlewared/middlewared/plugins/pool_/expand.py
```diff
@@ -59,64 +59,90 @@ async def expand(self, job, id, options):
...
-                        logger.warning('Or to resize provider back: %r',
-                                       join_commandline(rollback_cmd))
-                        geli_resize.append((geli_resize_cmd, rollback_cmd))
+                    partitions = []
+                    disks = vdev['children'] if vdev['type'] != 'DISK' else [vdev]
+                    skip_vdev = None
+                    for child in disks:
+                        if child['status'] != 'ONLINE':
+                            skip_vdev = f'Device "{child["device"]}" status is not ONLINE ' \
+                                        f'( Reported status is {child["status"]})'
```
william-gr, line 75: Missing extra space after ( ?

PR #4696, opened 2020-05-01
Migrate default SMB bind ip to not null

anodos325:


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -179,7 +179,7 @@ class SMBModel(sa.Model):
     cifs_srv_filemask = sa.Column(sa.String(120))
     cifs_srv_dirmask = sa.Column(sa.String(120))
     cifs_srv_smb_options = sa.Column(sa.Text())
-    cifs_srv_bindip = sa.Column(sa.MultiSelectField(), nullable=True)
+    cifs_srv_bindip = sa.Column(sa.MultiSelectField(), nullable=False)
```
themylogin, line 182: `nullable=False` is the default

PR #4704, opened 2020-05-04
NAS-105965 / 12.0 / fix rollback operation on TN HA

yocalebo:
This is the path of least resistance and ensures the databases are sync'ed after rollback changes. I've tried to write code that actually compares values of db and the changes that need to be rolled back but that becomes infinitely complex and is bound for errors. I'm okay with this fix because this only occurs on HA systems and only when a rollback takes place.

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
@@ -1618,7 +1618,7 @@ async def hook_license_update(middleware, *args, **kwargs):
     await middleware.call('service.restart', 'failover')
     await middleware.call('failover.status_refresh')
 
-async def hook_pre_rollback_setup_ha(middleware, *args, **kwargs):
+async def hook_post_rollback_setup_ha(middleware, *args, **kwargs):
     """
     This hook needs to be run after a NIC rollback operation and before
     an `interfaces.sync` operation on a TrueNAS HA system
@@ -1860,7 +1860,7 @@ async def setup(middleware):
     middleware.register_hook('pool.post_change_passphrase', hook_pool_change_passphrase, sync=False)
     middleware.register_hook('interface.pre_sync', interface_pre_sync_hook, sync=True)
     middleware.register_hook('interface.post_sync', hook_setup_ha, sync=True)
-    middleware.register_hook('interface.pre_rollback', hook_pre_rollback_setup_ha, sync=True)
+    middleware.register_hook('interface.post_rollback', hook_post_rollback_setup_ha, sync=True)
     middleware.register_hook('pool.post_create_or_update', hook_setup_ha, sync=True)
     middleware.register_hook('pool.post_create_or_update', hook_sync_geli, sync=True)
     middleware.register_hook('pool.post_export', hook_pool_export, sync=True)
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -641,8 +641,7 @@ async def __restore_datastores(self):
         # breaks `failover.status` on TrueNAS HA systems. This means the
         # following `datastore.insert` operations fail because we can't
         # determine the Active/Standby controller.
-        if await self.middleware.call('failover.licensed'):
-            await self.middleware.call('failover.sync_to_peer')
+        await self.middleware.call_hook('rollback.post_sync')
```
william-gr, line 644: Here is an idea, how about we put it within the rollback method, after __restore_datastores call and name it `interface.pre_rollback` so the hook name is consistent with the others? We try to name it similar to the plugin/service

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1618,6 +1618,21 @@ async def hook_license_update(middleware, *args, **kwargs):
...
 
+async def hook_post_rollback_setup_ha(middleware, *args, **kwargs):
+    """
+    This hook needs to be run after a NIC rollback operation is run
+    on a TrueNAS HA system
+    """
+    if not await middleware.call('failover.licensed'):
+        return
+
+    if await middleware.call('failover.status') == 'UNKNOWN':
```
william-gr, line 1629: I am not sure we can trust failover.status to tell us that, how about check NO_PING in disabled_reasons or actually try to ping it before? Or even catch an specific exception?

I imagine either would these would be more bullet proof

src/middlewared/middlewared/plugins/network.py
yocalebo: I agree, but you mean `interface.post_rollback` right since we're running after the rollback operation occurs?

yocalebo: Oh I see, you want it to run after the `__restore_datastores` method but before the `sync` method?

william-gr: Its a grey line, maybe.. we do rollback the internal state, but dont actually rollback until after .sync runs
I am fine with either

src/middlewared/middlewared/plugins/network.py
```diff
@@ -683,6 +677,13 @@ async def rollback(self):
         # We do not check for failover disabled in here because we may be reverting
         # the first time HA is being set up and this was already checked during commit.
         await self.__restore_datastores()
+
+        # All entries are deleted from the network tables on a rollback operation.
+        # This breaks `failover.status` on TrueNAS HA systems.
+        # Because of this, we need to manually sync the database to the standby
+        # controller.
+        await self.middleware.call_hook('interfaces.pre_rollback')
```
william-gr, line 685: Typo, interfaces vs interface

yocalebo: Great catch :scream:

yocalebo:
Before you approve, I'm going to change to `interface.post_rollback` as it makes more sense grammatically. :smile:

PR #4710, opened 2020-05-04
NAS-105940 / 12.0 / NAS-105940: Fix problem related with write tunables in rc.conf

bastiao:
Fix problem related with write tunables in rc.conf


william-gr:
@bastiao I will merge because of time constraints, but next time you can merge your PRs ; )

PR #4723, opened 2020-05-05
Fix replication test

themylogin:


william-gr:
@themylogin  wrong backport label?

themylogin:
@william-gr yes, no need for backport

PR #4728, opened 2020-05-05
NAS-106023 / 12.0 / Use CLDAP for determining domain and site information

anodos325:
This removes dependency openldap dependency in AD plugin and
simplifies our code significatnly. It will also make future support
for LDAP channel binding simpler.

### DIFF
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -2,10 +2,7 @@ import datetime
 import enum
 import errno
 import grp
-import ipaddress
 import json
-import ldap
-import ldap.sasl
 import ntplib
 import os
 import pwd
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -24,10 +20,24 @@
...
 except ImportError:
     MSG_WINBIND_ONLINE = 9
+    MUST_USE_KERBEROS = 2
+    param = None
+    Net = None
+    Credentials = None
+    nbt = None
+    gensec = None
+
+LP_CTX = param.get_context()
```
william-gr, line 40: if the ImportError happens on SCALE we need this to be inside the try statement...

If not we should get rid of try/except at this point

themylogin, line 40: If we got into `ImportError` then it will be `AttributeError: 'NoneType' object has no attribute 'get_context'` as `params` is `None`

PR #4729, opened 2020-05-05
NAS-105954 / 12.0 / Generate SSH config after syncing interfaces

sonicaj:
This commit introduces changes where we generate SSH configuration after syncing interfaces as binded SSH interfaces at boot time have no ips and they are not reflected in the ssh configuration when it's generated at boot.


sonicaj:
backport

PR #4750, opened 2020-05-07
NAS-106049 / 11.3 / Fix errors tracebacks during system dataset migration

anodos325:
Export of pool and related system dataset migration triggered
two tracebacks. One regression in smb_configure
caused by groupmap_list output format changes and the other
in reporting setup due to failure to check whether symlink
already exists.

### DIFF
--- a/src/middlewared/middlewared/etc_files/smb_configure.py
+++ b/src/middlewared/middlewared/etc_files/smb_configure.py
@@ -199,30 +199,6 @@ def set_SID(middleware, config):
 
 
 """
-    Validate contents of group_mapping.tdb, which maps local Unix groups to Windows group
-    SIDs. This file should:
-    1) contain no duplicate or inconsistent entries
-    2) contain no group names that are identical to usernames
-
-    To do: add validation of SID values in the tdb file.
-"""
-
-
-def get_groups(middleware):
-    _groups = {}
-    groups = middleware.call_sync('group.query', [('builtin', '=', False)])
-    for g in groups:
-        key = str(g['group'])
-        _groups[key] = []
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/etc_files/smb_configure.py
```diff
@@ -290,7 +290,7 @@ def validate_group_mappings(middleware, conf):
             groupmap = []
```
themylogin, line 290: Then you'll need `groupmap = {}` here

src/middlewared/middlewared/plugins/reporting.py
```diff
@@ -890,7 +890,7 @@ def setup(self):
             os.makedirs(os.path.join(pwd, 'localhost'))
 
         # Create "${hostname}" -> "localhost" symlink if necessary
-        if hostname != 'localhost':
+        if hostname != 'localhost' and not os.path.exists(os.path.join(pwd, hostname)):
```
themylogin, line 893: This should always not exist, there is code above that removes both symlinks and directories.

themylogin: Seconds thought: if someone creates file named `localhost` or `f"{hostname}"` there, our code might crash, we won't create corresponding directories, maybe that was the case?

anodos325: Okay. I will back out this particular change until I can reliably reproduce / have a better idea on what caused the issue.

src/middlewared/middlewared/etc_files/smb_configure.py
anodos325: Indeed.

PR #4752, opened 2020-05-07
NAS-105931 / 12.0 / NAS-105931: Add birthday dates for software installation

bastiao:


src/middlewared/middlewared/plugins/system.py
```diff
@@ -1520,6 +1525,10 @@ async def setup(middleware):
         if os.path.exists('/usr/local/sbin/beadm'):
             await firstboot(middleware)
 
+            # Sync clock
```
themylogin, line 1528: Why do you only do it if `beadm` exists?

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1520,6 +1525,10 @@ async def setup(middleware):
...
+            middleware.logger.debug('Synchornization the clock for system birthday')
+            sync_clock()
```
themylogin, line 1530: You only get current time here, not storing it anywhere

src/middlewared/middlewared/utils/__init__.py
```diff
@@ -442,3 +442,21 @@ def get_service(self, name):
 
     def get_services(self):
         return self._services
+
```
themylogin, line 445: This file is too big already, please put all new utils to separate modules

src/middlewared/middlewared/plugins/system.py
```diff
@@ -768,6 +771,7 @@ class SystemGeneralModel(sa.Model):
     stg_guihttpsredirect = sa.Column(sa.Boolean(), default=False)
     stg_language = sa.Column(sa.String(120), default="en")
     stg_kbdmap = sa.Column(sa.String(120))
+    stg_birthday = sa.Column(sa.DateTime(), nullable=True)
```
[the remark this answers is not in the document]
themylogin: You're also lacking database migration (`cd /usr/local/lib/python3.8/site-packages/middlewared && alembic revision --autogenerate`, it will appear in `alembic` subdir)

src/middlewared/middlewared/utils/__init__.py
```diff
@@ -442,3 +442,21 @@ def get_service(self, name):
...
+# Sync the clock
+def sync_clock():
+    DEFAULT_NTP_SERVER="freebsd.pool.ntp.org"
```
yocalebo, line 448: @themylogin We should probably use the ntp servers that the end-user has set first and then only use the `DEFAULT_NTP_SERVER` if all else fails. TrueNAS customers block lots of things out of their network.

src/middlewared/middlewared/utils/__init__.py
```diff
@@ -442,3 +442,21 @@ def get_service(self, name):
...
+    clock = None
+    try:
+        response = client.request(DEFAULT_NTP_SERVER)
+        if response.version:
+            server_alive = True
+            clock = response.tx_time
+            
+    except Exception:
+        # Internet not available, get datetime from host
+        clock = datetime.datetime.now()
```
william-gr, line 460: Relying on system clock should be avoided, loses the entire point of doing this unless we are sure ntpd did its things.

Also odds are on first install people do not even have network configured, so this should be retried on network changes if the birth time is null.

src/middlewared/middlewared/utils/__init__.py
william-gr: Definitely. Even better if there is some way to query local ntpd for it instead of doing ourselves.

src/middlewared/middlewared/plugins/system.py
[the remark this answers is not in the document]
themylogin: I recall my `nullable` comment, seems we'll need `null` state until we are able to receive correct system time.

src/middlewared/middlewared/utils/__init__.py
themylogin: Still seems worth to try default NTP servers (we don't have any other on install anyway)

themylogin: @yocalebo @william-gr reminds me about one thing... Can we take advantage of the fact that we can get NTP server via DHCP?

william-gr: We have to retry until it works, the odds are none of this will work right after install

yocalebo: > @yocalebo @william-gr reminds me about one thing... Can we take advantage of the fact that we can get NTP server via DHCP?

Yes, option 42 provides NTP but DHCP isn't supported on TN HA systems and we really push paying customers to use static IP addresses on everything.

src/middlewared/middlewared/utils/__init__.py
bastiao: Make sense. Now it is isolated.

src/middlewared/middlewared/plugins/system.py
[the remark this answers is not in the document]
bastiao: Make sense, some.

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1134,6 +1138,7 @@ async def ui_certificate_choices(self):
             Str('sysloglevel', enum=['F_EMERG', 'F_ALERT', 'F_CRIT', 'F_ERR', 'F_WARNING', 'F_NOTICE',
                                      'F_INFO', 'F_DEBUG', 'F_IS_DEBUG']),
             Str('syslogserver'),
+            Str('birthday', empty=False),
```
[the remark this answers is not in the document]
bastiao: Right, no need to do in that way.

src/middlewared/middlewared/utils/__init__.py
bastiao: The logic was reimplemented according with feedback.

src/middlewared/middlewared/utils/__init__.py
bastiao: Make sense. Changed that.

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1358,12 +1365,35 @@ def set_language(self):
...
 
+# Update Birthday Date
+async def _update_birthday(middleware):
+    # Sync clock
+    middleware.logger.debug('Synchornization the clock for system birthday')
+    birthday = None
+    timeout=60
+
+    while birthday == None:
+        birthday = sync_clock()
```
themylogin, line 1376: `sync_clock` performs I/O, you can't do this in asynchronous methods. I propose to move it to system service as private method so you can take advantage of I/O thread pool we have.

src/middlewared/middlewared/plugins/system.py
```diff
@@ -568,11 +573,47 @@ async def info(self):
...
             'system_manufacturer': manufacturer,
             'ecc_memory': ecc_memory,
         }
 
+    async def _convert_datetime(self, response, timezone_setting=timezone.utc):
+        return datetime.fromtimestamp(response.tx_time, timezone_setting)
+
+    # Sync the clock
+    @private
+    async def sync_clock(self):
```
william-gr, line 587: Dont think you fully understood Vladimir.

ntplib is synchronous which can block an event loop, which means this has to be a method, not a coroutine.

You're also mixing other synchronous calls and async calls.

src/middlewared/middlewared/plugins/system.py
```diff
@@ -585,22 +585,18 @@ async def _convert_datetime(self, response, timezone_setting=timezone.utc):
     # Sync the clock
     @private
     async def sync_clock(self):
```
themylogin, line 587: This should not be async, it performs blocking I/O in `NTPClient`

src/middlewared/middlewared/plugins/system.py
```diff
@@ -568,11 +573,47 @@ async def info(self):
...
+            if response.version:
+                clock = await self._convert_datetime(response)
+        except Exception:
+            # Cannot connect to NTP server
+            pass
+
+        if clock is not None:
+            return clock
+
+        # If it fails, tries the list of default ntp servers configurable
```
william-gr, line 602: We are already querying local ntp which has the ntp servers configured, I dont really see the point of doing these things below?

src/middlewared/middlewared/plugins/system.py
```diff
@@ -568,11 +573,47 @@ async def info(self):
...
             'boottime': datetime.fromtimestamp(psutil.boot_time()),
             'datetime': datetime.utcnow(),
-            'timezone': (await self.middleware.call('datastore.config', 'system.settings'))['stg_timezone'],
+            'birthday': birthday_date,
+            'timezone': timezone_setting,
             'system_manufacturer': manufacturer,
             'ecc_memory': ecc_memory,
         }
 
+    async def _convert_datetime(self, response, timezone_setting=timezone.utc):
```
themylogin, line 582: This should not be `async`, it does not `await` anything

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1359,11 +1401,35 @@ def set_crash_reporting(self):
...
+    # Sync clock
+    middleware.logger.debug('Synchornization the clock for system birthday')
+    birthday = None
+    timeout = 60
+
+    while birthday is None:
+        birthday = await middleware.call('system.sync_clock')
+        # Wait until be able to sync the clock
+        if birthday is None:
+            await asyncio.sleep(timeout)
```
william-gr, line 1415: Dont think we should retry every 60 seconds, this seems overly aggressive. Perhaps only once a day, and at most 5 retries or so?

We could probably also put it in a hook to retry after network changes are made (e.g. interface.post_sync)

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1359,11 +1401,35 @@ def set_crash_reporting(self):
...
+    })
+
+
 async def _event_system(middleware, event_type, args):
     global SYSTEM_READY
     global SYSTEM_SHUTTING_DOWN
     if args['id'] == 'ready':
         SYSTEM_READY = True
+        # Check if birthday is already setted
+        birthday = middleware.sync_call('system.info')['birthday']
```
themylogin, line 1429: Sorry, but I am under impression that this code has never been executed before pushing

bastiao: Simulated now with: `midclt call core.event_send system ADDED '{"id": "ready"}'`

src/middlewared/middlewared/plugins/system.py
bastiao: Added both as suggested.

src/middlewared/middlewared/plugins/system.py
bastiao: Make sense.

src/middlewared/middlewared/plugins/system.py
bastiao: Done.

src/middlewared/middlewared/plugins/system.py
bastiao: Done

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1359,11 +1384,47 @@ def set_crash_reporting(self):
         CrashReporting.enabled_in_settings = self.middleware.call_sync('system.general.config')['crash_reporting']
 
 
+async def _update_birthday_data(middleware, birthday=None):
+    if birthday is None:
```
themylogin, line 1388: This method should not set the birthday if it is already set (loop from `_update_birthday` won't be aborted if `'interface.post_sync'`, and next iteration will overwrite birthday value that has already been set)

src/middlewared/middlewared/plugins/system.py
```diff
@@ -568,11 +573,30 @@ async def info(self):
...
             'datetime': datetime.utcnow(),
-            'timezone': (await self.middleware.call('datastore.config', 'system.settings'))['stg_timezone'],
+            'birthday': birthday_date,
+            'timezone': timezone_setting,
             'system_manufacturer': manufacturer,
             'ecc_memory': ecc_memory,
         }
 
+    def _convert_datetime(self, response, timezone_setting=timezone.utc):
+        return datetime.fromtimestamp(response.tx_time, timezone_setting)
```
william-gr, line 583: This method is useless as its not used nor meant to me used anywhere, just add another thing to the stack call.

src/middlewared/middlewared/plugins/system.py
```diff
@@ -568,11 +573,30 @@ async def info(self):
...
+    # Sync the clock
+    @private
+    def sync_clock(self):
+        client = ntplib.NTPClient()
+        clock = None
+        # Tries to get default ntpd server
+        try:
+            response = client.request("localhost")
+            if response.version:
+                clock = self._convert_datetime(response)
```
william-gr, line 594: Lets just use datetime.fromtimestamp(response.tx_time, timezone_setting)

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1359,11 +1384,47 @@ def set_crash_reporting(self):
...
+        settings = await middleware.call('datastore.config', 'system.settings')
+        await middleware.call('datastore.update', 'system.settings', settings['id'], {
+            'stg_birthday': birthday,
+        })
+
+
+# Update Birthday Date
+async def _update_birthday(middleware):
+    # Sync clock
+    middleware.logger.debug('Synchornization the clock for system birthday')
```
william-gr, line 1400: typo

src/middlewared/middlewared/utils/__init__.py
```diff
@@ -11,7 +11,6 @@
 from datetime import datetime, timedelta
 from functools import wraps
 from threading import Lock
-
```
william-gr, line 14: Superfluous change, please revert it back

src/middlewared/middlewared/plugins/system.py
```diff
@@ -568,11 +573,27 @@ async def info(self):
...
+    def sync_clock(self):
+        client = ntplib.NTPClient()
+        clock = None
+        # Tries to get default ntpd server
+        try:
+            response = client.request("localhost")
+            if response.version:
+                clock = datetime.fromtimestamp(response.tx_time, timezone.utc)
+        except Exception:
+            # Cannot connect to NTP server
```
william-gr, line 593: Might be worth a debug level log here since this is not expected to happen

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1359,11 +1381,59 @@ def set_crash_reporting(self):
         CrashReporting.enabled_in_settings = self.middleware.call_sync('system.general.config')['crash_reporting']
 
 
+async def _update_birthday_data(middleware, birthday=None):
+    middleware.logger.debug('Synchornization _update_birthday_data interface')
```
william-gr, line 1385: typo, Synchornization

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1359,11 +1381,59 @@ def set_crash_reporting(self):
...
+    # Check if it is exists already
+    system_obj = (await middleware.call('system.info'))
```
william-gr, line 1387: Please dont use superfluous parenthesis, we dont this anywhere else in our code
just  `await middleware.call('system.info')` is OK

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1359,11 +1381,59 @@ def set_crash_reporting(self):
...
+# Update Birthday Date
+async def _update_birthday(middleware):
+    # Sync clock
+    middleware.logger.debug('Synchronization the clock for system birthday')
+    birthday = None
+    timeout = 3600 * 24
+
+    middleware.register_hook('interface.post_sync', _update_birthday_data)
+
+    while birthday is None:
```
william-gr, line 1414: I just thought of something we have to handle.

We dont want to set birthday for upgraded systems, only new installs.

Thoughts? @themylogin

themylogin: We can also set it to minimal mtime of files in `/data`; however, chances are that this time would be incorrect.

bastiao: for upgrade systems, are they not driven by the same database without erase?
what would be required to do in those files in `/data`?

themylogin: @bastiao they are, but some files are never touched since the installation, that's why I think minimal mtime will be precise enough. Not sure if it is correct to implement this, as original ticket does not explain what for do we need the birthday date. Thoughts? @william-gr

william-gr: Lets forget about upgrades and only do this for new installs.

For upgrades we will set a value in the database to let us know to not try to set a birthday time.
We will try to set birthday only to new installs.

src/middlewared/middlewared/alembic/versions/12.0/2020-05-08_15-18_birthday.py
```diff
@@ -0,0 +1,34 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('system_settings', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('stg_birthday', sa.DateTime(), nullable=True))
+
+    op.execute("UPDATE system_settings SET stg_birthday = '" + str(datetime(1970, 1, 1, 0, 0, 0)) + "'")
```
william-gr, line 24: This will set 1970 even for new installs, wont it @themylogin 

Can you explain to him how to make it behave properly, please?

src/middlewared/middlewared/plugins/system.py
```diff
@@ -568,11 +574,28 @@ async def info(self):
...
+        clock = None
+        # Tries to get default ntpd server
+        try:
+            response = client.request("localhost")
+            if response.version:
+                clock = datetime.fromtimestamp(response.tx_time, timezone.utc)
+        except Exception:
+            # Cannot connect to NTP server
+            self.middleware.logger.debug('Error while connecting to NTP server')
+            pass
```
william-gr, line 596: Remove useless `pass` please

src/middlewared/middlewared/alembic/versions/12.0/2020-05-08_15-18_birthday.py
themylogin: @bastiao do not set any default value in migration. Put some flag (e.g. `set_birthday: true`) into `system_keyvalue` table in installer (`src/freenas-installer/etc/installer12.sh`). Pop this flag in middleware and set birthday accordingly.

william-gr: @themylogin  dont we set any environment var when doing install vs upgrade to make this simpler/easier?

themylogin: @william-gr forgot about that, updated my comment

william-gr: @themylogin why put anything under system_keyvalue?

Why not just update the stg_birthday value based on that env var, directly?

themylogin: @william-gr update how? Set to `0000-00-00`?

william-gr: His intent was to set datetime(1970, 1, 1, 0, 0, 0) if its an upgrade

so only run that UPDATE if `os.environ["FREENAS_INSTALL"] != "yes"`

themylogin: Do we want to have 1970 for upgrades?

themylogin: @william-gr updated my comment once more

@bastiao please only do it if `os.environ.get("FREENAS_INSTALL") != "yes"`

src/middlewared/middlewared/plugins/system.py
william-gr: @bastiao forgot this

bastiao:
> Please do merge using Squash button and make the commit message nice ;)

still need approval from @themylogin and @yocalebo,  I guess.

PR #4759, opened 2020-05-10
NAS-106059 / 12.0 / NAS-106059: Fix pool.processes raising validation error

bastiao:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3577,7 +3577,21 @@ async def processes(self, oid):
         ]
         """
         result = []
-        dataset = await self.get_instance(oid)
+        dataset = None
+        path=None
```
themylogin, line 3581: PEP8 checked did not catch this for some reason

sonicaj:
I think we should check if root dataset can be found in `pool.processes` and not `pool.dataset.processes` and if it can't be found, we return an empty list from there ( `pool.processes` ). `pool.dataset.processes` should raise an error if the dataset in question does not exist. Thoughts @themylogin ?

themylogin:
I agree with you @sonicaj

william-gr:
@bastiao please go ahead and merge using Squash button like Vladimir said.

Thank you

PR #4760, opened 2020-05-10
NAS-105924 / 12.0 / NAS-105924: Added API for network probe

bastiao:


src/middlewared/middlewared/service.py
```diff
@@ -870,6 +871,29 @@ def ping(self):
...
+    def _ping_host(self, host, count, timeout):
+        process = run(['ping', '-c', f'{count}', '-W',  f'{timeout}', host])
+        if process.returncode != 0:
+            return False
+        else:
+            return True
+
+    @accepts(
+        Str('hostname'),
+        Int('timeout', default=10000),
```
themylogin, line 885: Why such a big timeout?

src/middlewared/middlewared/service.py
```diff
@@ -870,6 +871,29 @@ def ping(self):
         "ping" protocol message.
         """
         return 'pong'
+    
+
+    def _ping_host(self, host, count, timeout):
+        process = run(['ping', '-c', f'{count}', '-W',  f'{timeout}', host])
```
sonicaj, line 877: Can we please adjust the flags to reflect timeout in freebsd/debian ? I think you want `-t` for freebsd and `-w` for debian as uppercase `-W` is not a deadline.

src/middlewared/middlewared/service.py
```diff
@@ -870,6 +871,33 @@ def ping(self):
         "ping" protocol message.
         """
         return 'pong'
+    
+
+    def _ping_host(self, host, count, timeout):
+        if osc.IS_LINUX:
+            process = run(['ping', '-c', f'{count}', '-w',  f'{timeout}', host])
```
sonicaj, line 878: I don't think we should be using count or at least not defaulting to 1 only ? Just timeout should be good. Would you agree @themylogin ?

themylogin: Why? `ping -c1` is how we always check availability of the host, aren't we?

william-gr: First ping request can sometimes be lost, depending on network topology/firewall.

We are also waiting some clarification o Ken about other checks, e.g. ssh/middleware

sonicaj: We are exposing a timeout option and the consumer would expect that if no connection is made we would be waiting at least timeout X seconds. If we only send 1 request and that fails, we don't wait the remaining timeout to ensure another request worked ?

themylogin: Ok, agreed

william-gr:
@freenas/middleware what is your opinion on the place this service resides? is core.ping_remote reasonable enough or would anywhere else make sense?

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +872,32 @@ def ping(self):
...
+    @accepts(
+        Str('hostname'),
+        Int('timeout', default=10),  # seconds
+    )
+    def ping_remote(self, hostname, timeout):
+        """
+        Utility method call a ping in the other node.
+        The other node will just return "pong".
+        """
+        ping_host = self._ping_host(hostname, 1, timeout)
```
yocalebo, line 895: If we're pinging a remote system that we've never talked to before, then it's almost guaranteed to drop that packet because we don't have an ARP address for that IP or the switch has to update it's arp table.. I'd change the count to at least 2 probably 3.

bastiao: @yocalebo you have a good point! If there is no MAC address in ARP table, the first probably fails. Changed.

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +872,32 @@ def ping(self):
...
+            return False
+        else:
+            return True
+
+    @accepts(
+        Str('hostname'),
+        Int('timeout', default=10),  # seconds
+    )
+    def ping_remote(self, hostname, timeout):
+        """
```
yocalebo, line 891: Let's fix up the docstring.

`Method that will send a ICMPv4 echo request to "hostname" and will wait up to "timeout".`

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +872,32 @@ def ping(self):
...
+        Utility method call a ping in the other node.
+        The other node will just return "pong".
+        """
+        ping_host = self._ping_host(hostname, 3, timeout)
```
yocalebo, line 895: I would remove the `count` argument because we will be waiting no matter what depending on what the timeout is set to.

Also, we should probably cap the `timeout` value to something reasonable?

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +872,32 @@ def ping(self):
...
+    @accepts(
+        Str('hostname'),
+        Int('timeout', validators=[Range(min=1, max=60)], default=10),  # seconds
+    )
+    def ping_remote(self, hostname, timeout):
+        """
+        Method that will send an ICMPv4 echo request to "hostname"
+        and will wait up to "timeout" for a reply.
+        """
+        ping_host = self._ping_host(hostname, timeout)
```
william-gr, line 895: Just return ping_host please (boolean), I dont think a string brings any value.

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +872,32 @@ def ping(self):
...
+
+        if process.returncode != 0:
+            return False
+        else:
+            return True
+
+    @accepts(
+        Str('hostname'),
+        Int('timeout', validators=[Range(min=1, max=60)], default=10),  # seconds
+    )
```
william-gr, line 889: Lets make extend this API interface to be more future-proof, please.

Something like:

```
Dict(
    'options',
    Str('type', enum=['ICMP', 'ICMPV4', 'ICMPV6'], default='ICMP'),
    Str('hostname', required=True),
    Int('timeout', validators=[Range(min=1, max=60)], default=10),
)
```

We will probably have to resolve the hostname first and give preference to IPv4.

This way we could extend it in the future to include more protocols without breaking the API (backward compatible)

src/middlewared/middlewared/service.py
```diff
@@ -872,31 +872,50 @@ def ping(self):
...
     )
-    def ping_remote(self, hostname, timeout):
+    def ping_remote(self, options):
         """
-        Method that will send an ICMPv4 echo request to "hostname"
+        Method that will send an ICMP echo request to "hostname"
         and will wait up to "timeout" for a reply.
         """
-        ping_host = self._ping_host(hostname, timeout)
+        if options['type'] == 'ICMP' or options['type'] == 'ICMPV4':
```
william-gr, line 910: Hmm, I dont think ping can do ipv6?

ICMP should do either v4 or v6, depending on what its resolving to if its a hostname.

src/middlewared/middlewared/service.py
```diff
@@ -907,10 +908,19 @@ def ping_remote(self, options):
...
-            ping_host = self._ping_host(options['hostname'], options['timeout'])
-        elif options['type'] == 'ICMPV6':
-            ping_host = self._ping6_host(options['hostname'], options['timeout'])
+        ip = None
+        try:
+            ip = IpAddress()
+            ip(options['hostname'])
+            ip = options['hostname']
+        except ValueError:
+            ip = socket.gethostbyname(options['hostname'])
```
william-gr, line 917: I have the feeling gethostbyname only resolves to IPv4 ;)

src/middlewared/middlewared/service.py
```diff
@@ -907,10 +908,19 @@ def ping_remote(self, options):
         Method that will send an ICMP echo request to "hostname"
         and will wait up to "timeout" for a reply.
         """
-        if options['type'] == 'ICMP' or options['type'] == 'ICMPV4':
-            ping_host = self._ping_host(options['hostname'], options['timeout'])
-        elif options['type'] == 'ICMPV6':
-            ping_host = self._ping6_host(options['hostname'], options['timeout'])
```
william-gr: So If I specify ICMPV6 your code may resolve to ipv4 and will just erroneously work

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +873,67 @@ def ping(self):
         """
         return 'pong'
 
+    def _ping_host(self, host, timeout, counter=3):
+        if osc.IS_LINUX:
+            process = run(['ping', '-w', f'{timeout}', '-c', f'{counter}', host])
```
william-gr, line 878: Linux ping can do ipv6, you need -4

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +873,67 @@ def ping(self):
...
+        ip_found = True
+        try:
+            ip = IpAddress()
+            ip(options['hostname'])
+            ip = options['hostname']
+        except ValueError:
+            ip_found = False
+        if not ip_found:
+            try:
+                ip = socket.getaddrinfo(options['hostname'], None)[0][4][0]
```
william-gr, line 921: I dont think you're doing the right thing still.

If you specify ICMPV6 it should force resolving to IPV6, right now it doesnt. It will just fail if it resolves to a IPV4 by default.

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +873,70 @@ def ping(self):
...
+            ip(options['hostname'])
+            ip = options['hostname']
+        except ValueError:
+            ip_found = False
+        if not ip_found:
+            try:
+                if options['type'] == 'ICMPV6':
+                    ip = socket.getaddrinfo(options['hostname'], None, socket.AF_INET6)[0][4][0]
+                else:
+                    ip = socket.getaddrinfo(options['hostname'], None, socket.AF_INET)[0][4][0]
```
william-gr, line 924: We dont want to force AF_INET then type is ICMP

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +873,70 @@ def ping(self):
         """
         return 'pong'
 
+    def _ping_host(self, host, timeout, counter=3):
+        if osc.IS_LINUX:
+            process = run(['ping', '-4 -w', f'{timeout}', '-c', f'{counter}', host])
```
william-gr, line 878: this wont work... '-4', '-w'

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +873,74 @@ def ping(self):
...
+        if not ip_found:
+            try:
+                if options['type'] == 'ICMP':
+                    ip = socket.getaddrinfo(options['hostname'], None)[0][4][0]
+                elif options['type'] == 'ICMPV4':
+                    ip = socket.getaddrinfo(options['hostname'], None, socket.AF_INET)[0][4][0]
+                elif options['type'] == 'ICMPV6':
+                    ip = socket.getaddrinfo(options['hostname'], None, socket.AF_INET6)[0][4][0]
+                else:
+                    return False
```
william-gr, line 928: the else is impossible, unreachable code, please remove.

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +873,74 @@ def ping(self):
...
+            except socket.gaierror:
+                return False
+
+        addr = ipaddress.ip_address(ip)
+        ping_host = False
+        if addr.version == 4 and (options['type'] == 'ICMPV4' or options['type'] == 'ICMP'):
```
william-gr, line 934: Sorry, something more came up in last review after your changes.

If you specify ICMPV4 and give a IPV6 it should error out, give a validation error. Not simply return False as it does right now.

Once you add the validation the second part of these both if statements wont be necessary.

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +873,90 @@ def ping(self):
...
+                    'core.ping_remote',
+                    'The provided hostname cannot be resolved'
+                )
+
+        verrors.check()
+
+        addr = ipaddress.ip_address(ip)
+        if not addr.version == 4 and (options['type'] == 'ICMP' or options['type'] == 'ICMPV4'):
+            verrors.add(
+                'core.ping_remote',
```
william-gr, line 939: The first argument is the schema path, not method name

e.g. options.type or options.hostname

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +873,90 @@ def ping(self):
...
+                    'The provided hostname cannot be resolved'
+                )
+
+        verrors.check()
+
+        addr = ipaddress.ip_address(ip)
+        if not addr.version == 4 and (options['type'] == 'ICMP' or options['type'] == 'ICMPV4'):
+            verrors.add(
+                'options.type',
+                'Requested to reach via ICMP or ICMPV4, but an IPv4 was not found'
```
yocalebo, line 940: Let's change verbiage as this doesn't make sense. We'll use an f-string and print the `addr` variable as well.

f'`Requested ICMPv4 protocol, but the address provided "{addr}" is not a valid IPv4 address.`'

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +873,90 @@ def ping(self):
...
+            )
+        if not addr.version == 6 and options['type'] == 'ICMPV6':
+            verrors.add(
+                'options.type',
+                'Requested to reach via ICMPV6, but an IPv6 was not found'
```
yocalebo, line 945: Same here as above.

f'`Requested ICMPv6 protocol, but the address provided "{addr}" is not a valid IPv6 address.`'

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +873,90 @@ def ping(self):
         """
         return 'pong'
 
+    def _ping_host(self, host, timeout, counter=3):
+        if osc.IS_LINUX:
+            process = run(['ping', '-4', '-w', f'{timeout}', '-c', f'{counter}', host])
```
yocalebo, line 878: Specifying `count` on Linux is ignored when you specify `-w` so let's remove the `counter` keyword please.

To make it even more confusing, freeBSD has differing behavior when specifying a timeout and a count.

On freeBSD, it will send `count` packets and wait `timeout` which further validates the reason why we should remove the `counter` keyword.

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +873,90 @@ def ping(self):
...
+            process = run(['ping', '-4', '-w', f'{timeout}', host])
+        else:
+            process = run(['ping', '-t', f'{timeout}', host])
+
+        if process.returncode != 0:
+            return False
+        else:
+            return True
+
+    def _ping6_host(self, host, timeout, counter=3):
```
william-gr, line 887: counter is still here and you're not using it.

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +873,90 @@ def ping(self):
...
+                f'Requested ICMPv6 protocol, but the address provided "{addr}" is not a valid IPv6 address.'
+            )
+        verrors.check()
+
+        ping_host = False
+        if addr.version == 4:
+            ping_host = self._ping_host(ip, options['timeout'])
+        elif addr.version == 6:
+            ping_host = self._ping6_host(ip, options['timeout'])
+
```
william-gr, line 954: `return ping_host`, instead.

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +873,90 @@ def ping(self):
...
+            return False
+        else:
+            return True
+
+    def _ping6_host(self, host, timeout):
+        if osc.IS_LINUX:
+            process = run(['ping6', '-w', f'{timeout}', host])
+        else:
+            process = run(['ping6', '-X', f'{timeout}', host])
+
```
william-gr, line 892: `return process.returncode == 0`

src/middlewared/middlewared/service.py
```diff
@@ -871,6 +873,90 @@ def ping(self):
         """
         return 'pong'
 
+    def _ping_host(self, host, timeout):
+        if osc.IS_LINUX:
+            process = run(['ping', '-4', '-w', f'{timeout}', host])
+        else:
+            process = run(['ping', '-t', f'{timeout}', host])
+
```
william-gr, line 881: `return process.returncode == 0`

PR #4763, opened 2020-05-11
NAS-104899 / 12.0 / SCALE update

themylogin:


themylogin:
@william-gr this can't be merged right away, as resulting middleware package:
1) Won't be able to do `system.version` as it reads `/data/manifest.json` which is added to rootfs by new build script
2) Will generate invalid grub configuration as root dataset needs to be named differently and have kernel version property
Seems to me that it also depends on new installer?

src/freenas-installer/scale/build_archive.py
```diff
@@ -0,0 +1,124 @@
...
+    parser.add_argument("--skip-rootfs", action="store_true")
+    parser.add_argument("version")
+    args = parser.parse_args()
+
+    os.chdir(os.path.normpath(os.path.dirname(__file__)))
+
+    if not args.skip_rootfs:
+        with contextlib.suppress(FileNotFoundError):
+            os.unlink("/usr/share/debootstrap/scripts/truenas")
+        subprocess.run("ln -s /usr/share/debootstrap/scripts/bullseye /usr/share/debootstrap/scripts/truenas",
```
kmoore134, line 29: @themylogin 

Looks like you still have lots of code for doing debootstrap and similar steps in here. Did you intend to use the 'make update' functionality of truenas-build where we already bootstrap the rootfs image, install packages, etc?

src/middlewared/middlewared/etc_files/grub.py
```diff
@@ -1,52 +1,27 @@
...
+        cmdline.append(f"console={advanced['serialport']},{advanced['serialspeed']} console=tty1")
+
+    config.append(f'GRUB_TERMINAL="{" ".join(terminal)}"')
+    config.append(f'GRUB_CMDLINE_LINUX="{" ".join(cmdline)}"')
+    config.append("")
+
+    with open("/etc/default/grub.d/truenas.cfg", 'w') as f:
+        f.write("\n".join(config))
+
+    await run(["update-grub"], encoding="utf-8", errors="ignore")
```
william-gr, line 27: We lost the log in case this fails

src/middlewared/middlewared/plugins/etc.py
```diff
@@ -296,6 +297,8 @@ async def generate(self, name):
             if osc.IS_LINUX:
                 if entry_path.startswith('local/'):
                     entry_path = entry_path[len('local/'):]
+            if 'output_path' in entry:
+                entry_path = entry['output_path']
```
william-gr, line 301: Isnt that what local_path already does?

src/middlewared/middlewared/plugins/update_/utils.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding=utf-8 -*-
+import itertools
+import re
+
+SCALE_MANIFEST_FILE = "/data/manifest.json"
+SCALE_UPDATE_SERVER = "http://pivnoy.thelogin.ru/scale"
```
william-gr, line 6: @kmoore134 what should this be? http://download.truenas.com/scale

kmoore134: For updates? Lets use: https://update.freenas.org/scale/ for now. 

Eventually we will do a DNS re-direct to update.truenas.org though, and we can update our location.

PR #4772, opened 2020-05-12
ensure log dir exists on TN HA systems

yocalebo:


src/fenced/fenced/main.py
```diff
@@ -65,6 +65,17 @@ def panic(reason):
...
 
+def ensure_log_dir_exists():
+    """
+    On HA systems, we need to ensure that `LOG_DIR` is
+    created so that logging to `LOG_DIR`/fenced.log works.
+    """
+    if not os.path.exists(LOG_DIR):
+        os.makedirs(LOG_DIR, exist_ok=True)
+
+    return
```
william-gr, line 76: makes more sense to do all that within the logging.py of fenced?
And reuse the log file var?

Also I am not sure why the fenced in this repo is diverging so much with the 11.3 in truenas repo

yocalebo: I'll check repos and see what the divergence is.

src/fenced/fenced/logging.py
```diff
@@ -26,6 +28,19 @@ def handleError(self, record):
...
 
 
+def ensure_logdir_exists():
+    """
+    We need to ensure that the directory in `LOG_FILE` exists
+    so logging works
+    """
+    dirname = os.path.dirname(LOG_FILE)
+
+    if not os.path.exists(dirname):
```
william-gr, line 38: this check is probably a no-op, we could just drop it and just try os.makedirs since exist_ok=True anyway? Thoughts?

src/fenced/fenced/logging.py
```diff
@@ -26,6 +28,19 @@ def handleError(self, record):
...
+        os.makedirs(dirname, exist_ok=True)
+
+    return
+
+
 def setup_logging(foreground):
```
william-gr, line 44: And maybe just call ensure_logdir_exists in here? Then we dont have to change main.py

src/fenced/fenced/logging.py
yocalebo: If I don't get `dirname` then `os.makedirs` actually makes a `/root/syslog/fenced.log` directory.

william-gr: No, I mean `if not os.path.exists(dirname):`, we can drop that condition and just do `os.makedirs(dirname, exist_ok=True)`

src/fenced/fenced/logging.py
```diff
@@ -26,6 +28,19 @@ def handleError(self, record):
...
+    """
+    We need to ensure that the directory in `LOG_FILE` exists
+    so logging works
+    """
+    dirname = os.path.dirname(LOG_FILE)
+
+    if not os.path.exists(dirname):
+        os.makedirs(dirname, exist_ok=True)
+
+    return
```
william-gr, line 41: Also just return is superfluous

src/fenced/fenced/logging.py
yocalebo: oh derp, will fix.

yocalebo:
backport

PR #4778, opened 2020-05-14
NAS-106044 / 12.0 / fix fenced on TN HA systems

yocalebo:
The issue is that fenced was not detecting when remote keys changed because the `get_keys` method is returning the `failed_keys` in the place of `remote_keys` in the tuple.

### DIFF
--- a/src/fenced/fenced/disks.py
+++ b/src/fenced/fenced/disks.py
@@ -88,12 +88,12 @@ class Disks(dict):
 
         def callback(i, fs, failed):
             try:
-                host_key, remote_keys = i.result()
+                host_key, remote_key = i.result()
+                remote_keys.update(remote_key)
                 if host_key is None:
                     failed.add(fs[i])
                     return
                 keys.add(host_key)
-                remote_keys.union(remote_keys)
             except Exception:
                 failed.add(fs[i])
 


src/fenced/fenced/disks.py
```diff
@@ -67,7 +67,7 @@ def _run_batch(self, method, args=None, disks=None, done_callback=None):
             done_notdone = fut_wait(fs.keys(), timeout=30)
         finally:
             executor.shutdown(wait=False)
-        failed = {fs[i] for i in done_notdone.not_done}
+        failed = [fs[i] for i in done_notdone.not_done]
```
william-gr, line 70: Why are we changing from a set to a list?

yocalebo: Mistake when pushing my local changes when I was debugging.

yocalebo: I fixed it.

src/fenced/fenced/disks.py
```diff
@@ -97,7 +98,7 @@ def callback(i, fs, failed):
             except Exception:
                 failed.add(fs[i])
 
-        failed = self._run_batch('get_keys', done_callback=callback)
+        remote_keys = self._run_batch('get_keys', done_callback=callback)
```
william-gr, line 101: That doesnt seem right, _run_batch is supposed to return failures, not get_keys values

yocalebo: The reason why is because the `callback` is making a check to see if the `host_key` is None and if it is it will add the disk to the `failed` set and then return.

In this scenario, host_key will always be `None` because the other controller has the reservation keys set.

william-gr: and why host_key is None? I dont see how/why it would be.

Even if it can be this does not seem like the right way to handle this as its not a failure and we will be adding it there.
It will mix actual failures with remote_keys results.

yocalebo: Spotted the actual problem. 👍 PR updated.

PR #4792, opened 2020-05-17
NAS-106141 / 12.0 / Platform dependent NTP plugin

sonicaj:
1) Makes NTP plugin platform dependent
2) Introduces changes to allow receiving responses from specified ntp servers. ( http://support.ntp.org/bin/view/Support/AccessRestrictions\#Section_6.5.1.2 )


src/middlewared/middlewared/migration/004_ntp_default_servers.py
```diff
@@ -0,0 +1,20 @@
+async def migrate(middleware):
+    if await middleware.call('system.product_type') != 'SCALE':
+        return
+
+    # There is a drawback with this approach, it will virtually not allow any user to use freebsd ntp server
```
themylogin, line 5: What do you mean, why?

src/middlewared/middlewared/migration/004_ntp_default_servers.py
```diff
@@ -0,0 +1,20 @@
+async def migrate(middleware):
```
themylogin, line 1: This file should be named `0003_ntp_default_servers_linux.py` because
- We only have 0001 and 0002, don't we?
- All other files are named with 4 digits, why be different here?
- If upgrading 11.3 -> 12.0 -> SCALE, this migration will be marked as executed in 12.0 and won't be ran again after upgrading to SCALE. If we name file `_linux` it will only be loaded on Linux which is what we desire.

src/middlewared/middlewared/etc_files/ntp.conf
```diff
@@ -26,4 +26,7 @@ restrict -6 default ignore
 restrict 127.0.0.1
 restrict -6 ::1
 restrict 127.127.1.0
+% for ntp in ntp_query:
```
themylogin, line 29: Shouldn't this be Linux-only?

sonicaj: Nope, this is true for freebsd as well

src/middlewared/middlewared/migration/004_ntp_default_servers.py
sonicaj: Ah, I thought that these migrations are applied on each upgrade but your previous comment explains that. This comment is no longer true, i'll make amends. Thank you.

src/middlewared/middlewared/migration/0004_ntp_default_servers_linux.py
```diff
@@ -0,0 +1,18 @@
+async def migrate(middleware):
+    if await middleware.call('system.product_type') != 'SCALE':
```
themylogin, line 2: Not necessary anymore

src/middlewared/middlewared/etc_files/localtime_config.py
```diff
@@ -1,19 +1,36 @@
...
+        with contextlib.suppress(OSError):
+            os.unlink('/etc/localtime')
+        os.symlink(os.path.join('/usr/share/zoneinfo', system_config['timezone']), '/etc/localtime')
+        cp = subprocess.Popen(
+            ['systemctl', 'daemon-reload'], stdout=subprocess.DEVNULL, stderr=subprocess.PIPE
+        )
+        _, stderr = cp.communicate()
+        if cp.returncode:
+            middleware.logger.error(
+                'Failed to reload systemctl daemon after timezone configuration: %s', stderr.decode()
```
yocalebo, line 24: Verbiage seems strange here. Maybe you mean `Failed to reload ntp daemon`?

sonicaj: Nope, we are not reloading ntp daemon. The intention is to reload systemctl daemon so it reflects the timezone change as well.

src/middlewared/middlewared/etc_files/localtime_config.py
```diff
@@ -1,19 +1,36 @@
...
-        os.path.join('/usr/share/zoneinfo/', system_config['timezone']),
-        '/etc/localtime'
-    )
+    if osc.IS_LINUX:
+        with contextlib.suppress(OSError):
+            os.unlink('/etc/localtime')
+        os.symlink(os.path.join('/usr/share/zoneinfo', system_config['timezone']), '/etc/localtime')
+        cp = subprocess.Popen(
+            ['systemctl', 'daemon-reload'], stdout=subprocess.DEVNULL, stderr=subprocess.PIPE
+        )
```
william-gr, line 20: Do we need to reload if localtime has not actually changed?

src/middlewared/middlewared/etc_files/localtime_config.py
```diff
@@ -1,19 +1,36 @@
...
+        _, stderr = cp.communicate()
```
william-gr, line 21: stderr = communicate()[1] ;)

PR #4794, opened 2020-05-17
NAS-106143 / 12.0 / Generate debug for SCST in SCALE

sonicaj:


debian/debian/control
```diff
@@ -9,6 +9,7 @@ Homepage: http://www.truenas.com
 Package: truenas
 Architecture: all
 Depends: collectd,
+         cpuid,
```
william-gr, line 12: For what reason? commit doesnt say

sonicaj: It's used in debug, I forgot to add it when we ported ixdiagnose to linux.

PR #4799, opened 2020-05-18
Add validation for smb usernames and groupnames

anodos325:
Improve validation for usernames and groupnames when "smb" is selected. Convert groupmap and passdb synchronization to jobs as these can potentially take a significant amount of time to complete in environments with large numbers of SMB users.


src/middlewared/middlewared/plugins/account.py
```diff
@@ -677,6 +678,19 @@ async def __common_validation(self, verrors, data, schema, pk=None):
                     f'The username "{data["username"]}" already exists.',
                     errno.EEXIST
                 )
+            if 'smb' in data:
```
william-gr, line 681: That doesnt mean smb is True?

PR #4802, opened 2020-05-19
NAS-106116 / 12.0 / remove autotune generated nmbclusters sysctl

yocalebo:
Historically, it seems the `kern.ipc.nmbclusters` was added to autotune script back in 2016 but is prone to not working and isn't working on the x-series or m-series platforms for TrueNAS.

The reason why it's not working is because this value is a loader type and freeBSD will autogenerate a value based on RAM that is detected at boot. We are miscalculating the value which means we are setting a `sysctl` value that is < then value that freeBSD is autogenerating. You can not alter this value to make it smaller after the system has booted. You must set a loader type at your desired value and then reboot.

I spoke with @amotin about this value and the autogenerated value by freeBSD should be sufficient.


src/middlewared/middlewared/alembic/versions/12.0/2020-05-18_14-29_nmbclusters-fix.py
```diff
@@ -0,0 +1,98 @@
...
+revision = '730c995cbd37'
+down_revision = '434ea5397cd3'
+branch_labels = None
+depends_on = None
+
+
+def get_db_sysctl_value():
+    TABLE = 'system_tunable'
+    SELECT_FROM = f'SELECT * FROM {TABLE}'
+    WHERE = ' WHERE tun_var = "kern.ipc.nmbclusters"'
```
william-gr, line 23: Should we also filter for comment = Generated by autotune?

src/middlewared/middlewared/alembic/versions/12.0/2020-05-18_14-29_nmbclusters-fix.py
```diff
@@ -0,0 +1,98 @@
...
+        except Exception:
+            pass
+
+    return curr_val
+
+
+def remove_db_entry(db_value, os_value):
+    TABLE = 'system_tunable'
+    DELETE_FROM = f'DELETE FROM {TABLE}'
+    WHERE = f' WHERE id = {db_value["id"]}'
```
william-gr, line 69: We shouldnt really miss uppercase which usually means global vars within a method scope. These vars also seems superfluous and do not bring any value

Perhaps TABLE could be global and the rest just part of execute string?

src/middlewared/middlewared/alembic/versions/12.0/2020-05-18_14-29_nmbclusters-fix.py
yocalebo: Yeah, I thought about that too. I'll filter for that as well.

src/middlewared/middlewared/alembic/versions/12.0/2020-05-18_14-29_nmbclusters-fix.py
yocalebo: yep, agreed. I fixed.

PR #4805, opened 2020-05-19
NAS-106164 / 12.0 / check HA status before running update.check_available

yocalebo:
`UNAVAILABLE` status doesn't really make sense if HA is non-functional.


src/middlewared/middlewared/plugins/update.py
```diff
@@ -435,6 +436,10 @@ def check_available(self, attrs=None):
             not self.middleware.call_sync('system.is_freenas') and
```
william-gr, line 436: While we are here lets remove is_freenas call please, no longer necessary.

PR #4819, opened 2020-05-21
NAS-106183 / 12.0 / Make netatalk logging configurable

anodos325:
Use same options as SMB service so that the options for AFP and SMB
are more consisitent.


src/middlewared/middlewared/plugins/afp.py
```diff
@@ -59,6 +69,7 @@ async def compress(self, afp):
         Str('global_aux', max_length=None),
         Str('map_acls', enum=['RIGHTS', 'MODE', 'NONE']),
         Str('chmod_request', enum=['PRESERVE', 'SIMPLE', 'IGNORE']),
+        Str('loglevel', enum=[x.name for x in AFPLogLevel], default=AFPLogLevel.MINIMUM.name),
```
william-gr, line 72: I think default is basically a no-op for update

PR #4825, opened 2020-05-22
replace sleep by loop in host_key test 02 with timeout

ericbsd:


william-gr:
How about @pytest.mark.timeout(60) instead?

ericbsd:
ha did not think to look for that in the pytest module for something like that. Thanks

PR #4829, opened 2020-05-24
NAS-106114 / 12.0 / NAS-106114: Fix issue related with disable or remove tunable, do not reset to default values

bastiao:
Fix issue related with toggling the "enable" checkbox in the tunable section does nothing and delete do not also recover the default value.


bastiao:
> You should set the default value before you set it the first time, otherwise you cant know what it was.

Yes, it is what is setted. Setted the default value first, then reload the service (it will set the sysctl). I also considered that the value could be added/removed/added again and the value will be setted only the first time.

Can you give a bit more details of what are the expected behaviour?

src/middlewared/middlewared/etc_files/sysctl_config.py
```diff
@@ -1,20 +1,49 @@
...
-                'Failed to set sysctl %s -> %s: %s',
-                tuneable['var'], tuneable['value'], ret.stderr.decode(),
+        value_default = None
+        if tuneable['type'].lower() == 'sysctl':
+            try:
+                value_default = middleware.call_sync('tunable.get_default_value', tuneable['var'])
+            except KeyError:
+                pass
+        if tuneable['enabled'] is True:
+            if value_default is None and tuneable['type'].lower() == 'sysctl':
```
william-gr, line 17: Why are you checking for type if you are already filtering for type in the query?

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -99,6 +112,24 @@ async def do_delete(self, id):
         Delete Tunable of `id`.
         """
         tunable = await self._get_instance(id)
+        await self.lower(tunable)
+        if tunable['type'].lower() == 'sysctl':
```
william-gr, line 116: Dont think none of this is required as everything is handled in sysctl etc file?

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -24,6 +25,18 @@ class Config:
...
+    def __init__(self, *args, **kwargs):
+        super(TunableService, self).__init__(*args, **kwargs)
+        self.__default_sysctl = {}
+
+    @private
+    async def get_default_value(self, oid):
+        return self.__default_sysctl[oid]
+
+    @private
+    async def set_default_value(self, oid, value):
```
william-gr, line 37: Dont think you should overwrite default values, second run would overwrite default values with custom values?

bastiao: It should not occur with current implementation, but still, agree, that we can add protection in the method, itself.

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -99,6 +113,24 @@ async def do_delete(self, id):
         Delete Tunable of `id`.
         """
         tunable = await self._get_instance(id)
+        await self.lower(tunable)
+        if tunable['type'].lower() == 'sysctl':
+            # Restore the default value, if it is possible.
+            value_default = None
+            try:
+                value_default = self.__default_sysctl[tunable["var"]]
```
william-gr, line 121: Lets use get_default_value so we dont mess with __default_sysctl outside the getter/setter

src/middlewared/middlewared/etc_files/sysctl_config.py
```diff
@@ -1,20 +1,49 @@
...
+            pass
+        if tuneable['enabled'] is True:
+            if value_default is None:
+                try:
+                    value_default = sysctl.filter(tuneable['var'])[0].value
+                    middleware.call_sync('tunable.set_default_value', tuneable['var'], value_default)
+                except IndexError:
+                    # Not able to set the default value
+                    middleware.logger.error(
+                        'Failed to get sysctl default value of %s', tuneable['var']
```
william-gr, line 24: , exc_info=True

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -99,6 +113,24 @@ async def do_delete(self, id):
         Delete Tunable of `id`.
         """
         tunable = await self._get_instance(id)
+        await self.lower(tunable)
+        if tunable['type'].lower() == 'sysctl':
```
william-gr, line 117: We already do .lower(), not necessary to repeat it

src/middlewared/middlewared/etc_files/sysctl_config.py
```diff
@@ -1,20 +1,49 @@
 import subprocess
+import sysctl
+
+default_sysctl = {}
```
william-gr, line 4: Where is this used?

src/middlewared/test_email.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/local/bin/python
```
william-gr, line 1: Wrong commit ;)

PR #4839, opened 2020-05-26
NAS-106237 / 11.3 / Avoid race between zpool import completion and share config generation

anodos325:
11.3-U2 added a sanity check when generating share configuration to
avoid adding shares to non-existent paths. In some cases configuration
file generation may happen before zpool import completes. This causes
shares on the pool with slow import to not be visible.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2490,11 +2490,6 @@ class PoolService(CRUDService):
             self.logger.warn('Failed to configure collectd', exc_info=True)
 
         try:
-            self.middleware.call_sync('etc.generate', 'ftp')
-        except Exception:
-            self.logger.warn('Failed to configure ftp', exc_info=True)
-
-        try:
             self.middleware.call_sync('etc.generate', 'syslogd')
         except Exception:
             self.logger.warn('Failed to configure syslogd', exc_info=True)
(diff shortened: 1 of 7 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2526,6 +2527,26 @@ async def dismiss_current_import_disk_job(self):
         if current_import_job:
             self.dismissed_import_disk_jobs.add(current_import_job["id"])
 
+    @private
+    def notify_available(self, pool, max_iter=12):
+        path = f'/mnt/{pool["name"]}'
+        available = False
+        tries = 0
+        while not available:
+            available = os.path.exists(path)
```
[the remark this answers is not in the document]
anodos325: Yes, that makes much more sense. Thanks. I have a bit of conference-brain today and will look at this again tomorrow. Right now the reduced commit I just made makes more sense, which may not be a good sign :)

src/middlewared/middlewared/plugins/webdav.py
```diff
@@ -237,6 +237,10 @@ async def pool_post_import(middleware, pool):
     """
     Makes sure to reload WebDAV if a pool is imported and there are shares configured for it.
     """
+    if pool is None:
+        asyncio.ensure_future(middleware.call('etc.generate', 'webdav'))
+        return
+
     path = f'/mnt/{pool["name"]}'
     if await middleware.call('sharing.webdav.query', [
```
william-gr, line 245: what if we combine both statements?

if pool is None or ...:
    asyncio.ensure_future(middleware.call('service.reload', 'webdav'))

anodos325: My thought was to avoid a restart of apache in this case because the service's rc script will take care of starting it later in boot.

william-gr: Ah, makes sense!

PR #4845, opened 2020-05-27
NAS-105943 / 11.3 / disk.sync_all improvements

themylogin:
* Wait for devd in boot-time disk.sync_all
* Better logging in disk.sync_all
* No unnecessary updates in disk.sync_all

### DIFF
--- a/src/middlewared/middlewared/plugins/device.py
+++ b/src/middlewared/middlewared/plugins/device.py
@@ -3,10 +3,11 @@ import os
 import socket
 
 from middlewared.schema import accepts, Str
-from middlewared.service import Service
+from middlewared.service import private, Service
 
 from bsd import devinfo, geom
 
+DEVD_CONNECTED = False
 DEVD_SOCKETFILE = '/var/run/devd.pipe'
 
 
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/device.py
```diff
@@ -58,11 +59,21 @@ async def _get_disk(self):
             disks[g.name] = disk
         return disks
 
+    @private
+    async def devd_connected(self):
+        global DEVD_CONNECTED
```
william-gr, line 64: Is global necessary? Thought onlt when we need to modify the var

src/middlewared/middlewared/plugins/disk.py
```diff
@@ -1023,7 +1023,17 @@ async def sync_all(self, job):
...
+            if i > 0:
+                await asyncio.sleep(1)
+
+            if await self.middleware.call('device.devd_connected'):
+                break
+        else:
+            self.logger.warning('Starting disk.sync_all when devd is not connected yet')
+
         sys_disks = list((await self.middleware.call('device.get_info', 'DISK')).keys())
+        self.logger.info('Found disks: %r', sys_disks)
```
william-gr, line 1036: Hmm how about job.logs_fd.write so we dont "pollute" logs? Too wierd?

themylogin: Yes, it would be hard to find it there and match with other events. It'll only be one line on system boot / pool create/destroy / disk unplug, I don't see an issue here.

I've imagined `job.logs_fd` to be like separate log file for big external program output (e.g. `rclone`)

PR #4848, opened 2020-05-27
NAS-106197 / 11.3 / Alert collisions are inevitable on TN HA; ignore them

themylogin:


gui/system/migrations/0047_auto_20200527_1038.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
yocalebo, line 1: Should we change the name of this migration to what this is doing like all the other migration files?

gui/system/migrations/0047_auto_20200527_1038.py
```diff
@@ -0,0 +1,19 @@
...
+# Generated by Django 1.10.8 on 2020-05-27 10:38
+from __future__ import unicode_literals
```
yocalebo, line 3: Not using `unicode_literals` ? or maybe I am blind

Also, not using `models`?

william-gr:
@themylogin ping about migrate113

PR #4853, opened 2020-05-28
NAS-106246 / 12.0 / Remove legacy AD sysctls from the config file

anodos325:


src/middlewared/middlewared/alembic/versions/12.0/2020-05-28_12-16_remove-ds-sysctls.py
```diff
@@ -0,0 +1,41 @@
...
+
+TABLE = 'system_tunable'
+
+
+def remove_ad_sysctls():
+    conn = op.get_bind()
+    conn.execute(
+        f'DELETE FROM {TABLE}'
+        ' WHERE tun_var'
+        ' LIKE "freenas.directoryservice%"'
```
william-gr, line 27: Are there any other sysctls that we had under freenas namespace that we should/could remove?

PR #4865, opened 2020-05-29
NAS-106159 / 12.0 / debug: handle malformed progress reporting lines

william-gr:
Other tools can output similar output.

Ticket:	NAS-106251


william-gr:
backport

PR #4871, opened 2020-06-01
NAS-106171 / 12.0 / SMB: store next rid in configuration database

anodos325:
Samba stores the next rid to allocate in passdb.tdb. The group_mapping.tdb
will check this value before auto-generating a SID for the Unix <-> NT
group mapping. If users for some reason delete or move passdb.tdb after
creating entries in group_mapping.tdb, overlap between assigned RIDs
can occur.

To avoid these types of issues, we will persistently store the next
rid in the freenas configuration file. New field will be initialized to 0
during the migration, but the next time that a user or group is added
the next_rid counter will be migrated from the passdb.tdb file. If for
some reason the migration fails, we'll set the counter to 5000, which
should be typically above samba's algorithmic base for normal user behavior
(no potential for collision if UID / GID below 2000), and significantly
higher than what we'd have with default behavior. (Requires 4,000 combined
local users and groups).


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -339,6 +341,23 @@ def getparm(self, parm, section):
         except Exception as e:
             raise CallError(f'Attempt to query smb4.conf parameter [{parm}] failed with error: {e}')
 
+    @private
+    async def get_next_rid(self):
```
william-gr, line 345: Any chances for a race condition here? Do we need a lock?
I didnt follow the callers so I really dont know

PR #4873, opened 2020-06-01
NAS-106171 / 11.3 / SMB: Add RID check after initial SMB configuration

anodos325:
Add check to see whether any duplicate RIDs appear in the group_mapping
and passdb databases. If duplicates exist, remove the group_mapping
database and re-create it.

### DIFF
--- a/src/middlewared/middlewared/etc_files/smb_configure.py
+++ b/src/middlewared/middlewared/etc_files/smb_configure.py
@@ -297,4 +297,5 @@ def render(service, middleware):
     if conf['passdb_backend'] == "tdbsam":
         middleware.call_sync('smb.synchronize_passdb')
         validate_group_mappings(middleware, conf)
+        middleware.call_sync('smb.check_rid_conflict')
         middleware.call_sync('admonitor.start')
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -550,6 +550,31 @@ class SMBService(SystemServiceService):
             raise CallError(f'Attempt to query smb4.conf parameter [{parm}] failed with error: {e}')
 
     @private
+    @job(lock="rid_check")
+    async def check_rid_conflict(self, job):
+        pdb_by_sid = {}
+        passdb_jobs = await self.middleware.call(
+            'core.get_jobs',
+            [('method', '=', 'smb.synchronize_passdb'), ('status', 'in', ['WAITING', 'RUNNING'])]
+        )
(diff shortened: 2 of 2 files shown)


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -549,6 +549,27 @@ def getparm(self, parm, section):
...
 
+    @private
+    @job(lock="rid_check")
+    async def check_rid_conflict(self, job):
+        pdb_by_sid = {}
+        passdb_jobs = await self.middleware.call(
+            'core.get_jobs',
+            [('method', '=', 'smb.synchronize_passdb'), ('status', 'in', ['WAITING', 'RUNNING'])]
+        )
+        for pdb_job in passdb_jobs:
```
william-gr, line 560: get_jobs is not supposed to return Job objects so you cant wait on them like that

william-gr: @anodos325 ping

anodos325: You're correct. I'll fix it in a little bit :)

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -549,6 +549,31 @@ def getparm(self, parm, section):
...
+    @private
+    @job(lock="rid_check")
+    async def check_rid_conflict(self, job):
+        pdb_by_sid = {}
+        passdb_jobs = await self.middleware.call(
+            'core.get_jobs',
+            [('method', '=', 'smb.synchronize_passdb'), ('status', 'in', ['WAITING', 'RUNNING'])]
+        )
+        while passdb_jobs:
+            asyncio.sleep(5)
```
william-gr, line 561: aawit asyncio.sleep

src/middlewared/middlewared/etc_files/smb_configure.py
```diff
@@ -297,4 +297,5 @@ def render(service, middleware):
     if conf['passdb_backend'] == "tdbsam":
         middleware.call_sync('smb.synchronize_passdb')
         validate_group_mappings(middleware, conf)
+        middleware.call_sync('smb.check_rid_conflict')
```
william-gr, line 300: Do we want to wait for this to complete?

anodos325: No. It may take a long time to complete if there are _huge_ numbers of groups and it won't interfere with other things we are doing. So intention is to background it.

PR #4876, opened 2020-06-02
NAS-104133 / 12.0 / Boot Environment support in SCALE

sonicaj:


src/freenas/etc/grub.d/10_truenas_linux
```diff
@@ -292,7 +292,10 @@ if ! zfs list $roots_dataset > /dev/null 2>&1 ; then
   roots_dataset="freenas-boot/ROOT"
 fi
 OS_BASE="$OS"
-for dataset in `zfs list -H -o name -r $roots_dataset | sort -r` ; do
+boot_pool=`echo "$roots_dataset" | cut -d '/' -f1`
+activated_be=`zpool get -H -o value bootfs "$boot_pool"`
+datasets="$activated_be\n`zfs list -H -o name -r $roots_dataset | grep -v $activated_be | sort -r`"
```
themylogin, line 297: If `activated_be` is e.g. `boot-pool/ROOT/TrueNAS-13.0`, it will exclude `boot-pool/ROOT/TrueNAS-13.0U1`, can we make it exclude only exact match just in case?

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1433,6 +1433,12 @@ async def _event_system(middleware, event_type, args):
     global SYSTEM_SHUTTING_DOWN
     if args['id'] == 'ready':
         SYSTEM_READY = True
+        if os.path.exists(FIRST_INSTALL_SENTINEL):
```
william-gr, line 1436: Cant BSD reach this code?

sonicaj: It most certainly will, thank you for pointing it out :)

PR #4883, opened 2020-06-03
NAS-106150 / 12.0 / Bug fixes for swap devices configuration

sonicaj:
1) Fixes a race condition where if a device is configured with `dumpon` and the pool in which it is in is exported, starts `swap_configure` again which potentially means that when we try to wipe the disks, the disks have already been again configured to be used in swap.
2) Do not unnecessarily reconfigure swap on each disk wipe.
3) Make sure we have a device configured with `dumpon` after we remove the existing device from swap. It will not be configured if the device was in a mirror and a subsequent call of `swap_configure` is not made which in most cases is.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/swap_remove.py
+++ b/src/middlewared/middlewared/plugins/disk_/swap_remove.py
@@ -39,18 +39,14 @@ class DiskService(Service):
 
         swap_devices = await self.middleware.call('disk.get_swap_devices')
         for mirror in await self.middleware.call('disk.get_swap_mirrors'):
-            destroyed_mirror = False
-            for provider in mirror['providers']:
-                if providers.pop(provider['id'], None) and not destroyed_mirror:
-                    devname = mirror['encrypted_provider'] or mirror['real_path']
-                    if devname in swap_devices:
-                        await run('swapoff', devname)
-                    if mirror['encrypted_provider']:
-                        await self.middleware.call(
-                            'disk.remove_encryption', mirror['encrypted_provider']
-                        )
-                    await self.middleware.call('disk.destroy_swap_mirror', mirror['name'])
-                    destroyed_mirror = True
+            devname = mirror['encrypted_provider'] or mirror['real_path']
+            if devname in swap_devices:
+                await run('swapoff', devname)
+            if mirror['encrypted_provider']:
+                await self.middleware.call(
+                    'disk.remove_encryption', mirror['encrypted_provider']
+                )
+            await self.middleware.call('disk.destroy_swap_mirror', mirror['name'])
 
         configure_swap = False
         for p in providers.values():
--- a/src/middlewared/middlewared/plugins/disk_/swap_remove.py
+++ b/src/middlewared/middlewared/plugins/disk_/swap_remove.py
@@ -1,17 +1,27 @@
 import os
 
 from middlewared.utils.osc import IS_LINUX
-from middlewared.service import private, Service
+from middlewared.service import job, private, Service
 from middlewared.utils import run
 
 
 class DiskService(Service):
 
     @private
+    @job(lock='swaps_configure')
     async def swaps_remove_disks(self, disks):
         """
         Remove a given disk (e.g. ["da0", "da1"]) from swap.
         It will offline if from swap, removing encryption and destroying the mirror ( if part of one ).
+        """
+        return await self.swaps_remove_disks_internal(disks)
+
+    @private
+    async def swaps_remove_disks_internal(self, disks):
+        """
+        We have a separate endpoint for this to ensure that no other swap related operations not do swap devices
+        removal while swap configuration is in progress - however we still need to allow swap configuration process
+        to remove swap devices and it can use this endpoint directly for that purpose.
         """
         providers = {}
         for disk in disks:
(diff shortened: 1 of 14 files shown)


src/middlewared/middlewared/plugins/disk_/wipe.py
```diff
@@ -45,20 +45,29 @@ async def wipe_quick(self, dev, size=None):
...
+        Dict(
+            'options',
+            Str('mode', enum=['QUICK', 'FULL', 'FULL_RANDOM'], required=True),
+            Bool('synccache', default=True),
+            Dict('swap_removal_options', default=None, additional_attrs=True),
+        ),
     )
     @job(lock=lambda args: args[0])
-    async def wipe(self, job, dev, mode, sync):
+    async def wipe(self, job, dev, options):
```
william-gr, line 56: Its a public method, I would like to avoid changing API, making it backward compatible whenever possible.

src/middlewared/middlewared/plugins/disk_/swap_remove.py
```diff
@@ -55,5 +62,6 @@ async def swaps_remove_disks(self, disks):
                 except OSError:
                     pass
 
-        if configure_swap:
+        # Let consumer explicitly deny swap configuration if desired
+        if configure_swap and not options.get('no_configure_swap'):
```
themylogin, line 66: Let's use `configure_swap=True` instead

src/middlewared/middlewared/service.py
```diff
@@ -28,6 +28,21 @@
...
 PeriodicTaskDescriptor = namedtuple("PeriodicTaskDescriptor", ["interval", "run_on_start"])
 get_or_insert_lock = asyncio.Lock()
+LOCKS = {}
+
+
+def lock(lock_str):
+    def lock_fn(fn):
+        f_lock = LOCKS.get(lock_str)
+        if not f_lock:
+            f_lock = asyncio.Lock()
```
william-gr, line 38: Future suggestion

f_lock = LOCKS[lock_str] = asyncio.Lock()

src/middlewared/middlewared/service.py
```diff
@@ -28,6 +28,18 @@
...
 PeriodicTaskDescriptor = namedtuple("PeriodicTaskDescriptor", ["interval", "run_on_start"])
 get_or_insert_lock = asyncio.Lock()
+LOCKS = {}
+
+
+def lock(lock_str):
+    def lock_fn(fn):
+        f_lock = LOCKS[lock_str] = LOCKS.get(lock_str) or asyncio.Lock()
+
+        async def l_fn(*args, **kwargs):
```
themylogin, line 38: For wrappers we use `functools.wraps`

src/middlewared/middlewared/service.py
```diff
@@ -28,6 +28,18 @@
 
 PeriodicTaskDescriptor = namedtuple("PeriodicTaskDescriptor", ["interval", "run_on_start"])
 get_or_insert_lock = asyncio.Lock()
+LOCKS = {}
+
+
+def lock(lock_str):
+    def lock_fn(fn):
+        f_lock = LOCKS[lock_str] = LOCKS.get(lock_str) or asyncio.Lock()
```
themylogin, line 36: `LOCKS = defaultdict(asyncio.Lock)`

william-gr:
@themylogin why not? Why have yet another facility for locking of job abstract all that away?

Can we simply reduce the boilerplate necessary to call and wait for the job instead?

themylogin:
@william-gr I think that using a job just to use locking is abusing the instrument. Primary job feature is unlimited execution time if called externally, here we don't need this. Streaming output or logs come after long execution time, I don't see this here too. Jobs require special handling, caller needs to remember whether something is a job or not, that's inconvenient. If you want to add locking to something that was not a job before, you'll have to find all callers and change them. I see this `@lock` decorator as a good way to get advantage of just locking. We can unite it's namespace with job locks when we need it in the future.

william-gr:
I dont quite agree with that. You talk like job is an overkill, the overhead is close to zero?
Jobs can handle multiple calls with a limit barrier as well so if 10 calls are made it gets into a queue.

I am not seeing a valid technical reason to add another locking facility.

Also these are long running tasks, (re)configuring is not a limited time task. it can take quite some time if the swap is in use and needs to be readjusted.

themylogin:
Yes, the overhead is close to zero, but developer's time of changing a non-job to job is not zero.

Alternative to this decorator is using common `asyncio.Lock()` instance explicitly, I've just thought that decorator can be useful.

This is long-running task, I agree, but it is not spawned directly by user, it is called by other middleware method, which is a job.

william-gr:
Are there other use cases this could be used? If we are changing methods to being a job twice a year a little extra work is expected to happen. Introducing another mechanism+code that would not have justified use-cases in exchange of some minimal extra work is a bit odd.

Another thing that I can think of is that this only handles coroutines, not the case with jobs.

themylogin:
> Are there other use cases this could be used?

Yes, we have a lot of jobs in AD, iocage or KMIP, which are only jobs because they need locking

> Another thing that I can think of is that this only handles coroutines, not the case with jobs.

Proper implementation would handle both (by re-using job locks), this is just something to do fast before 12.0-BETA1

william-gr:
I don't agree with you because I see a different purpose for jobs, not just locking in these cases. But you both win, I wont argue anymore :)

PR #4885, opened 2020-06-04
NAS-106306 / 12.0 / Make sure MTU is properly configured at boot for interfaces

sonicaj:
This commit introduces changes where if we have a network config with a chain `physical iface -> vlan -> bridge` ( all having say MTU of 9000 ), we won't be able to set a custom MTU for the ifaces if specified because interfaces are synced after bridges have been setup.
Why this is an issue is because when we add the members to the bridge we configure their MTU and if up in the chain we have lower MTU values, we can't change the MTU.

### DIFF
--- a/src/middlewared/middlewared/plugins/interface/bridge.py
+++ b/src/middlewared/middlewared/plugins/interface/bridge.py
@@ -1,4 +1,4 @@
-from middlewared.service import Service
+from middlewared.service import private, Service
 
 from .netif import netif
 
@@ -8,7 +8,27 @@ class InterfaceService(Service):
     class Config:
         namespace_alias = 'interfaces'
 
-    def bridge_setup(self, bridge, sync_interface_opts):
+    @private
+    def pre_bridge_setup(self, bridge, sync_interface_opts):
+        name = bridge['interface']['int_interface']
+        members_database = set(bridge['members'])
+        for member in members_database:
+            sync_interface_opts[member]['skip_mtu'] = True
+
+        try:
+            iface = netif.get_interface(name)
+        except KeyError:
+            return
+
+        members = set(iface.members)
+        for member in members - members_database:
+            # These interfaces may be added dynamically for Jails/VMs
+            if member.startswith(('vnet', 'epair', 'tap')):
+                continue
+            iface.delete_member(member)
+
+    @private
+    def bridge_setup(self, bridge):
         name = bridge['interface']['int_interface']
         self.logger.info(f'Setting up {name}')
         try:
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -1732,23 +1732,41 @@ async def sync(self, wait_dhcp=False):
...
+        # 4) Sync bridge interfaces
 
-            cloned_interfaces.append(name)
-            await self.middleware.call('interface.bridge_setup', bridge, sync_interface_opts)
+        for bridge in bridges:
+            await self.middleware.call('interface.pre_bridge_setup', bridge, sync_interface_opts)
 
-        self.logger.info('Interfaces in database: {}'.format(', '.join(interfaces) or 'NONE'))
-        # Configure VLAN before BRIDGE so MTU is configured in correct order
+        # Configure LAGG before VLAN so MTU is configured in correct order
```
themylogin, line 1748: This is not only related to LAGG. Let's say `Configure VLAN interfaces MTU last as they are restricted by underlying interfaces MTU`

src/middlewared/middlewared/plugins/interface/bridge.py
```diff
@@ -8,7 +8,28 @@ class InterfaceService(Service):
...
+    def pre_bridge_setup(self, bridge, sync_interface_opts):
+        name = bridge['interface']['int_interface']
+        members_database = set(bridge['members'])
+
+        try:
+            iface = netif.get_interface(name)
+        except KeyError:
+            return
+        finally:
+            for member in members_database:
```
themylogin, line 21: Why is it in `finally` block? It'll have more sense to be before `iface = netif.get_interface(name)`. Right now we have the code that is executed in non-obvious sequence.

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1732,23 +1732,41 @@ async def sync(self, wait_dhcp=False):
...
+        # 4) Sync bridge interfaces
 
-            cloned_interfaces.append(name)
-            await self.middleware.call('interface.bridge_setup', bridge, sync_interface_opts)
+        for bridge in bridges:
+            await self.middleware.call('interface.pre_bridge_setup', bridge, sync_interface_opts)
 
-        self.logger.info('Interfaces in database: {}'.format(', '.join(interfaces) or 'NONE'))
-        # Configure VLAN before BRIDGE so MTU is configured in correct order
+        # VLAN interfaces MTU last as they are restricted by underlying interfaces MTU
```
themylogin, line 1748: `Set VLAN interfaces MTU last`?

PR #4888, opened 2020-06-04
NAS-106336 / 12.0 / HA hardware detection for SCALE

yocalebo:


yocalebo:
@william-gr This is the initial work to get z-series hardware detection working on SCALE but I'm marking Pending QA and WIP until I can get my hands on x and m serires hardware. Could you review the initial though to see if you like direction?

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -170,165 +179,44 @@ def licensed(self):
 
     @private
     def ha_mode(self):
```
william-gr, line 181: This might as well be a coroutine now since there is nothing synchronous

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/detect_enclosure_freebsd.py
```diff
@@ -0,0 +1,128 @@
+import subprocess as sb
```
william-gr, line 1: We dont use this anywhere else in the codebase, I would prefer we keep the same without abbreviations to not confuse other people.

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/detect_enclosure_freebsd.py
```diff
@@ -0,0 +1,128 @@
...
+
+    @private
+    def detect(self):
+
+        # First detect if this is a BHYVE instance
+        manufacturer = sb.run(
+            ['dmidecode -s system-product-name'],
+            stdout=sb.PIPE,
+            stderr=sb.PIPE,
+            shell=True
```
william-gr, line 25: Lets avoid use shell whenever we can please.

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/detect_enclosure_freebsd.py
```diff
@@ -0,0 +1,128 @@
...
+
+            devlist = sb.run(
+                ['/sbin/camcontrol devlist'],
+                stdout=sb.PIPE,
+                stderr=sb.PIPE,
+                shell=True
+            )
+
+            if devlist.stdout:
+                devlist = devlist.stdout.decode()
```
william-gr, line 42: If we always decode, better use encoding='utf8', and errors='ignore' in subprocess run

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -169,167 +190,41 @@ def licensed(self):
...
-
-            if mode is None:
-                mode = hardware, node
-            return mode
-        else:
-            return 'MANUAL', None
+        return self.HA_MODE
 
     @accepts()
     def hardware(self):
```
william-gr, line 203: We should probably turn this into a coroutine as well

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -169,167 +190,41 @@ def licensed(self):
...
           ULTIMATE
+          BHYVE
           MANUAL
         """
-        return self.ha_mode()[0]
+
+        return self.middleware.call_sync('failover.ha_mode')[0]
 
     @accepts()
     def node(self):
```
william-gr, line 218: And this

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
yocalebo: I didn't change those to not change the API? If I change these to coroutines, I'll need to change all callers too, right?

william-gr: User API is unchanged. You need to change if they are using self and not self.middleware.call

PR #4889, opened 2020-06-04
Make sure autogenerated extent serial is unique

sonicaj:


src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -808,14 +808,17 @@ async def extent_serial(self, serial):
...
 
-                ltg = await self.query()
+                ltg = await self.query([], {'order_by': ['id']})
                 if len(ltg) > 0:
-                    lid = ltg[0]['id']
+                    lid = ltg[-1]['id']
                 else:
                     lid = 0
-                return f'{mac.strip()}{lid:02}'
+                if osc.IS_LINUX:
```
william-gr, line 818: hmm why mac differs for linux? maybe it shouldnt?

sonicaj: To clarify, are you referring to the behaviour of us not setting 15 letter limit which we are doing for freebsd ?

william-gr: I guess.

Why is nic['state']['link_address'] the same between bsd and linux?

sonicaj: Yes it is. We don't need to enforce the 15 letter limit in SCALE because this limit is specific to ctld.

william-gr: Perhaps the limit should be enforced in the ctl config generation then, and not here?

sonicaj: I am not sure I follow, we should not allow the user to set a serial exceeding 15 chars ? So if we implicitly trim it in ctl config generation, do you think that's okay and we should proceed with it ?

PR #4894, opened 2020-06-08
NAS-105029 / 12.0 / NAS-105029: Platform dependent mail plugin

bastiao:


sonicaj:
This will be done in another PR

PR #4895, opened 2020-06-08
NAS-104332 / 12.0 / NAS-104332: Platform dependent Reporting plugin

bastiao:


sonicaj:
This will be done in another PR

PR #4896, opened 2020-06-08
NAS-106388 / 11.3 / AD: query time from regular DC if NTP broken on PDC emulator

anodos325:
PDC emulator should be authoritative for the domain for domain members,
but if for some reason the NTP server is not working on it, or if it
is unreachable, query time from a DC on the network. If this also
fails, then allow exception to be raised because it's a notable
error condition. Otherwise, only generate warnings in the middleware
log.

### DIFF
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -1119,9 +1119,22 @@ class ActiveDirectoryService(ConfigService):
             self.logger.warning("Unable to find PDC emulator via DNS.")
             return {'pdc': None, 'timestamp': '0', 'clockskew': 0}
 
-        c = ntplib.NTPClient()
-        response = c.request(pdc[0]['host'])
-        ntp_time = datetime.datetime.fromtimestamp(response.tx_time)
+        try:
+            c = ntplib.NTPClient()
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1119,9 +1119,23 @@ def check_clockskew(self, ad=None):
...
+        except ntplib.NTPException:
+            self.logger.warning("NTP request to PDC Emulator failed. Retrying with regular DC",
+                                exc_info=True)
+
+            dc = ActiveDirectory_DNS(conf=ad, logger=self.logger).get_n_working_servers(SRV['PDC'], 1)
+            if not dc:
+                self.logger.warning("Unable to find Domain Controller via DNS.")
+                return {'pdc': None, 'timestamp': '0', 'clockskew': 0}
+
+            c = ntplib.NTPClient()
```
themylogin, line 1135: Why do it the second time?

anodos325: Because I was copying and pasting while working on something else. Will fix.

themylogin: I mean why do you do NTP request the second time?

PR #4902, opened 2020-06-08
NAS-106397 / 12.0 / SCALE system.version

themylogin:


src/middlewared/middlewared/utils/osc/linux/app.py
```diff
@@ -16,7 +16,7 @@ def get_app_version():
 
     return {
         'stable': not RE_UNSTABLE.match(manifest['version']),
-        'version': manifest['version'][len('TrueNAS-'):],
-        'fullname': manifest['version'],
+        'version': manifest['version'],
+        'fullname': f"TrueNAS-{manifest['version']}",
```
william-gr, line 20: What system.version looks like after this?

themylogin: `MASTER-20200603-042400`, do you want `13.0-MASTER-20200603-042400` instead?

william-gr: We need some mention of SCALE.

Currently it is: midclt call system.version
TrueNAS-12.0-MASTER-202006030424

We probably want something along the lines of:
TrueNAS-SCALE-XX.XX-MASTER-202006030424

PR #4903, opened 2020-06-08
Fix await for failover.node

sonicaj:


sonicaj:
Awesome, i'll close this then

PR #4904, opened 2020-06-08
NAS-106284 / 12.0 / fix virtual and node-b hostname config on HA

yocalebo:


william-gr:
How does that fix things? The intent was to not return these fields unless HA licensed

PR #4906, opened 2020-06-08
NAS-106405 / 12.0 / fix scale enclosure detection for non-truenas hardware

yocalebo:


yocalebo:
I'll fix the specific traceback 👍

PR #4927, opened 2020-06-11
NAS-104338 / 12.0 / Platform dependent tunable plugin

sonicaj:


src/middlewared/middlewared/plugins/system.py
```diff
@@ -1626,12 +1626,13 @@ async def setup(middleware):
...
 
-    asyncio.ensure_future(middleware.call('system.advanced.autotune', 'sysctl'))
+    if osc.IS_FREEBSD:
+        asyncio.ensure_future(middleware.call('system.advanced.autotune', 'sysctl'))
 
     if sysctl:
         await update_timeout_value(middleware)
 
-    for srv in ['initshutdownscript', 'tunable', 'vm']:
+    for srv in (['initshutdownscript', 'tunable', 'vm'] if osc.IS_FREEBSD else []):
```
themylogin, line 1635: Why?

sonicaj: FreeBSD imposed a limit on shutdown defaulting to 120 secs, so even if we had tasks running, that was enforced and middleware was forcibly killed. So this hook is specific to FreeBSD and that's why this change has been made

themylogin: Put entire loop under this condition then so it is clear

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1626,12 +1626,13 @@ async def setup(middleware):
     await middleware.call('system.general.set_language')
     await middleware.call('system.general.set_crash_reporting')
 
-    asyncio.ensure_future(middleware.call('system.advanced.autotune', 'sysctl'))
+    if osc.IS_FREEBSD:
+        asyncio.ensure_future(middleware.call('system.advanced.autotune', 'sysctl'))
 
     if sysctl:
```
themylogin, line 1632: `if osc.IS_FREEBSD:` probably?

src/middlewared/middlewared/migration/0005_tunables_linux.py
```diff
@@ -0,0 +1,7 @@
+async def migrate(middleware):
```
william-gr, line 1: Why not a regular migration for this?

sonicaj: That will be executed for users in CORE as well - however when they upgrade to SCALE, it won't be run again whereas this will be

PR #4929, opened 2020-06-11
NAS-106430 / 12.0 / platform independent force failover method

yocalebo:


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -328,21 +328,24 @@ async def get_ips(self, app):
...
-        for i in self.middleware.call_sync('interface.query', [('failover_critical', '!=', None)]):
+
+        for i in await self.middleware.call('interface.query', [('failover_critical', '!=', None)]):
             if i['failover_vhid']:
-                self.middleware.call_sync('failover.event', i['name'], i['failover_vhid'], 'forcetakeover')
+                await self.middleware.call('failover.event', i['name'], i['failover_vhid'], 'forcetakeover')
                 break
-        return False
+
+        return True
```
themylogin, line 348: Is this intentional change?

yocalebo: Yeah, not sure why we would return False when we're successful at forcing the failover?

yocalebo: In the interest of keeping logic the same, I changed it back for now.

PR #4931, opened 2020-06-11
NAS-106448 / 12.0 / Reload systemd daemon after enabling/disabling services

sonicaj:


src/middlewared/middlewared/etc_files/systemd.py
```diff
@@ -36,3 +36,7 @@ async def render(service, middleware):
         is_enabled = {"enabled": True, "disabled": False}[is_enabled]
         if enable != is_enabled:
             await run(["systemctl", "enable" if enable else "disable", service])
+
+    # Reload systemd daemon as on first boot, services by default are enabled and we do disable them above
+    # but the change is not reflected in systemd until we reload the daemon
+    await run(["systemctl", "daemon-reload"])
```
themylogin, line 42: Can we do it only if `enable != is_enabled` was true at least once?

PR #4941, opened 2020-06-12
NAS-106467 / 12.0 / Introduce preset file for TN package in SCALE

sonicaj:
This PR adds changes where we ship a systemd preset file indicating which services are enabled/disabled by default by us. Why this is necessary is because systemd defaults to enabling all services and on first boot, even though we disable them during `ix-etc` execution, the boot order remains unchanged and the services are running the first time. ( Point to note is that by default systemd enables all services, so it's not required to specify which services we mark as enabled in the preset file, it's just been done for consistency and transparency ).

### DIFF
--- a/src/middlewared/middlewared/plugins/service_/services/ssh.py
+++ b/src/middlewared/middlewared/plugins/service_/services/ssh.py
@@ -12,6 +12,7 @@ class SSHService(SimpleService):
     freebsd_pidfile = "/var/run/sshd.pid"
 
     systemd_unit = "ssh"
+    systemd_extra_units = ["ix-ssh-keys"]
 
     async def before_start(self):
         await self.middleware.call("service.reload", "mdns")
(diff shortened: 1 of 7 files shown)


src/middlewared/middlewared/plugins/service_/services/ssh.py
```diff
@@ -12,6 +12,7 @@ class SSHService(SimpleService):
     freebsd_pidfile = "/var/run/sshd.pid"
 
     systemd_unit = "ssh"
+    systemd_extra_units = ["ix-ssh-keys"]
```
themylogin, line 15: Why don't you remove it completely? `after_start` / `after_reload` do the job

src/middlewared/middlewared/plugins/service_/services/ssh.py
```diff
@@ -25,3 +24,9 @@ async def before_reload(self):
 
     async def after_reload(self):
         await self.middleware.call("ssh.save_keys")
+
+    async def before_restart(self):
```
themylogin, line 28: This is only executed if `restartable = True`

PR #4943, opened 2020-06-15
NAS-106474 / 12.0 / Make sure sysctl keys/values are in sync with reality for tunable plugin

sonicaj:

1) If a sysctl var is changed, we don't set that value back to it's default.
2) On a sysctl tunable deletion we should remove it from memory as well.

### DIFF
--- a/src/middlewared/middlewared/plugins/tunables.py
+++ b/src/middlewared/middlewared/plugins/tunables.py
@@ -102,7 +102,7 @@ class TunableService(CRUDService):
         """
         Update Tunable of `id`.
         """
-        old = await self._get_instance(id)
+        old = await self.get_instance(id)
 
         new = old.copy()
         new.update(data)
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -120,25 +120,35 @@ async def do_update(self, id, data):
...
 
+        if old['type'] == 'SYSCTL' and old['var'] in self.__default_sysctl and (
+            old['var'] != new['var'] or old['type'] != new['type']
+        ):
+            default_value = self.__default_sysctl.pop(old['var'])
+            cp = await run(['sysctl', f'{old["var"]}={default_value}'], check=False, encoding='utf8')
+            if cp.returncode:
+                self.middleware.logger.error(
+                    'Failed to set sysctl %r -> %r : %s', old['var'], default_value, cp.stderr
+                )
```
william-gr, line 131: Isnt the etc plugin supposed to do that?

sonicaj: We won't have this information in etc plugin - this is an edge case

william-gr: Ah, different type, thanks

PR #4950, opened 2020-06-15
NAS-105029 / 12.0 / Platform dependent mail plugin

sonicaj:


src/middlewared/middlewared/plugins/mail.py
```diff
@@ -424,7 +425,10 @@ def read_json():
                 raise CallError(str(e))
             syslog.syslog(f'Failed to send email to {", ".join(to)}: {str(e)}')
             if isinstance(e, smtplib.SMTPAuthenticationError):
-                raise CallError(f'Authentication error ({e.smtp_code}): {e.smtp_error}', errno.EAUTH)
+                raise CallError(
+                    f'Authentication error ({e.smtp_code}): {e.smtp_error}',
+                    errno.EAUTH if osc.IS_FREEBSD else errno.EFAULT
```
william-gr, line 430: Why different errcode?

sonicaj: We don't have `EAUTH` in linux

william-gr: EACCES then, or EPERM?

PR #4951, opened 2020-06-16
QE-355 Change tunable test 07 to user type1 variable for scale and core

ericbsd:


ericbsd:
> Is the choice of `variable` random ?

I don't know I did not create tunable API test I just did lower case it.

ericbsd:
@sonicaj I agree on that, I will PM you soon.

PR #4952, opened 2020-06-16
NAS-104332 / 12.0 / Platform dependent Reporting plugin

sonicaj:


src/middlewared/middlewared/etc_files/default/rrdcached.mako
```diff
@@ -0,0 +1,18 @@
+<%
+    if not middleware.call_sync('reporting.setup'):
+		# Let's exit this if setup related disk operations fail
```
william-gr, line 3: mix of space vs tabs

PR #4962, opened 2020-06-17
NAS-106381 / 12.0 / get rid of /tmp/failover.json

yocalebo:


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
```diff
@@ -116,18 +113,55 @@ def event(self, ifname, vhid, event):
...
+        return data
+
     @private
     def _event(self, ifname, vhid, event):
 
+        fobj = self.generate_failover_data()
+
+        if not fobj:
+            self.logger.warning('Unable to acquire failover information! Exiting.')
+            return
```
william-gr, line 158: fobj could never be false, so this is a no-op

yocalebo: I've removed the check.

PR #4965, opened 2020-06-17
NAS-106366 / 12.0 / loosen restrictions on network config for HA

yocalebo:
@william-gr made a fresh PR for this issue using the correct branch name 😃

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -1058,11 +1058,7 @@ class InterfaceService(CRUDService):
                             'VLAN MTU cannot be bigger than parent interface.',
                         )
 
-        failover_licensed = False
-        is_freenas = await self.middleware.call('system.is_freenas')
-        if not is_freenas:
-            failover_licensed = await self.middleware.call('failover.licensed')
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -1076,24 +1072,29 @@ async def _common_validation(self, verrors, schema_name, data, itype, update=Non
...
-                        f'{schema_name}.{i}',
-                        'This attribute is required when configuring HA.',
-                    )
-                    found = False
-            if found:
-                if len(data['aliases']) != len(data['failover_aliases']):
-                    verrors.add(
-                        f'{schema_name}.failover_aliases',
-                        'Number of IPs must be the same between controllers.',
-                    )
```
william-gr, line 1094: Just saw this now.. did we lose this check?

william-gr: Doesnt have to compare controllers, just the number of ips for node A vs node B

Lets add it back.

william-gr: As well in 11.3 if we ended up removing it

PR #4966, opened 2020-06-17
NAS-106366 / 11.3 / add back quantity of ip validation

yocalebo:


src/middlewared/middlewared/plugins/network.py
```diff
@@ -1015,6 +1015,12 @@ async def _common_validation(self, verrors, schema_name, data, itype, update=Non
                     'Failover needs to be disabled to perform network configuration changes.'
                 )
 
+            if len(data['aliases']) != len(data['failover_aliases']):
```
william-gr, line 1018: Are you sure we can guarantee there will be a failover_aliases and a aliases keys?

PR #4969, opened 2020-06-18
NAS-106511 / 12.0 / asyncio/thread pool/process pool juggling improvements

themylogin:
* Do not run async functions in process pool
* Do not use new IO thread if we are calling synchronous method from call_sync
* Do not use new process when calling process pool method from process pool

### DIFF
--- a/src/middlewared/middlewared/worker.py
+++ b/src/middlewared/middlewared/worker.py
@@ -9,14 +9,13 @@ import setproctitle
 from . import logger
 from .common.environ import environ_update
 from .utils import LoadPluginsMixin
-from .utils.io_thread_pool_executor import IoThreadPoolExecutor
 import middlewared.utils.osc as osc
-from .utils.run_in_thread import RunInThreadMixin
+from .utils.service.call import ServiceCallMixin
 
 MIDDLEWARE = None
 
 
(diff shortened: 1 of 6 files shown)


src/middlewared/middlewared/worker.py
```diff
@@ -94,17 +87,7 @@ def set_progress(self, percent, description=None, extra=None):
 
 def main_worker(*call_args):
     global MIDDLEWARE
-    loop = asyncio.get_event_loop()
-    coro = MIDDLEWARE._run(*call_args)
-    try:
-        res = loop.run_until_complete(coro)
-    except SystemExit:
```
william-gr, line 101: I think we should still catch SystemExit? Arbitrary methods could still raise it?

src/middlewared/middlewared/worker.py
```diff
@@ -94,17 +87,7 @@ def set_progress(self, percent, description=None, extra=None):
...
-        raise RuntimeError('Worker call raised SystemExit exception')
-    # TODO: python cant pickle generator for obvious reasons, we should implement
-    # it using Pipe.
-    if inspect.isgenerator(res):
```
william-gr, line 105: Why is this no longer necessary?

PR #4974, opened 2020-06-18
NAS-106512 / 12.0 / ACLs: fix acl_is_trivial on FreeBSD platform

anodos325:
Due to ACL parsing error in POSIX1e code on FreeBSD, paths
were being incorrectly labelled as having an ACL. Fix POSIX1e
parsing on FreeBSD, but bypass this for acl_is_trivial() since
we don't currently support POSIX1e ACLs on FreeBSD through middleware.

### DIFF
--- a/src/middlewared/middlewared/plugins/filesystem.py
+++ b/src/middlewared/middlewared/plugins/filesystem.py
@@ -451,9 +451,12 @@ class FilesystemService(Service):
         if not os.path.exists(path):
             raise CallError(f'Path not found [{path}].', errno.ENOENT)
 
-        if osc.IS_LINUX or not os.pathconf(path, 64):
+        if osc.IS_LINUX:
             posix1e_acl = self.getacl_posix1e(path, True)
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -707,7 +710,11 @@ def getacl_posix1e(self, path, simplified):
             raise CallError(f"Failed to get POSIX1e ACL on path [{path}]: {gfacl.stderr.decode()}")
 
         # linux output adds extra line to output if it's an absolute path and extra newline at end.
-        for entry in (gfacl.stdout.decode().splitlines())[:-(int(osc.IS_LINUX))]:
+        entries = gfacl.stdout.decode().splitlines()
+        if osc.IS_LINUX:
+            entries[:-1]
```
william-gr, line 715: entries = entries[:-1] ?

anodos325: Indeed

PR #4977, opened 2020-06-18
NAS-106482 / 12.1 / nvme reservation support for fenced

yocalebo:
This is 100% dependent on the PR in the py-cam library here: [https://github.com/freenas/py-cam/pull/3](PR)

### DIFF
--- a/src/fenced/fenced/disks.py
+++ b/src/fenced/fenced/disks.py
@@ -1,14 +1,13 @@
-# Copyright (c) 2019 iXsystems, Inc.
+# Copyright (c) 2020 iXsystems, Inc.
 # All rights reserved.
 # This file is a part of TrueNAS
 # and may not be copied and/or distributed
 # without the express permission of iXsystems.
 
-from collections import defaultdict
 from concurrent.futures import ThreadPoolExecutor, wait as fut_wait
-from functools import partial
 
 import cam
+import nvme
 import logging
 
 CAM_RETRIES = 5
@@ -113,8 +112,14 @@ class Disk(object):
         self.fence = fence
         self.name = name
         self.pool = pool
-        self.cam = cam.CamDevice(f'/dev/{name}')
         self.curkey = None
+        self.nvme = None
+        self.cam = None
+
+        if self.name.startswith('nvd'):
+            self.nvme = nvme.NvmeDevice(f'/dev/{name}')
+        else:
+            self.cam = cam.CamDevice(f'/dev/{name}')
 
     def __repr__(self):
         return f'<Disk: {self.name}>'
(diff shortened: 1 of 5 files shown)


yocalebo:
Tested this on an M-series all weekend.

1. fenced foreground
2. fenced foreground force
3. fenced exiting if already running
4. fenced exits if keys change
5. fenced starts if no reservation keys are set on any disks
6. fenced starts if some reservations are set on disks that aren't ours and some are

It's run without any issues. I've tried to keep the changes to fenced as minimal as possible.

src/fenced/fenced/disks.py
```diff
@@ -109,12 +108,17 @@ def reset_keys(self, newkey):
...
-    def __init__(self, fence, name, pool=None):
+    def __init__(self, fence, name, is_nvme=False, pool=None):
         self.fence = fence
         self.name = name
         self.pool = pool
-        self.cam = cam.CamDevice(f'/dev/{name}')
         self.curkey = None
+        self.is_nvme = is_nvme
+
+        if self.is_nvme:
```
william-gr, line 118: Isnt the name enough for figuring out if its nvme? Why the is_nvme

src/fenced/fenced/fence.py
```diff
@@ -48,27 +48,42 @@ def get_hostid(self):
...
 
+        return unsupported, remote_keys
+
+    def load_disks(self):
+
+        self._disks.clear()
+
+        # TODO: blacklist disks used by dumpdev
+        devices = sysctl.filter('kern.disks')[0].value.split()
+        drives = [i for i in devices if i.startswith('da')]
```
william-gr, line 80: i.startswith(('da', 'nvd'))

src/fenced/fenced/disks.py
yocalebo: Just thought it looked a little better than doing `if name.startswith('nvd')` calls. I can change it to that easily enough.

william-gr: You're already sort of doing that when instantiating the Disk object, so if you dont change the API and keep it inside as an implementation detail will probably keep it saner

yocalebo: Understood 👍

src/fenced/fenced/disks.py
```diff
@@ -113,9 +112,13 @@ def __init__(self, fence, name, pool=None):
         self.fence = fence
         self.name = name
         self.pool = pool
-        self.cam = cam.CamDevice(f'/dev/{name}')
         self.curkey = None
 
+        if self.name.startswith('nvd'):
```
william-gr, line 117: we could still have is_nvme here
or always have self.cam and self.nvme set to None by default

Comparing startswith() of everytime is repetitive, sorry if you thought thats what i wanted us to do

PR #4980, opened 2020-06-18
NAS-106519 / 12.0 / Fix rrdtool path in stats plugin

sonicaj:


src/middlewared/middlewared/plugins/stats.py
```diff
@@ -102,7 +103,7 @@ async def get_data(self, data_list, stats):
             ])
         proc = await Popen(
             [
-                '/usr/local/bin/rrdtool', 'xport',
+                RRD_TOOL_PATH, 'xport',
```
william-gr, line 106: Just use rrdtool, no absolute path

PR #4959, opened 2020-06-16
NAS-106336 / 11.3 / loosen restrictions on network config for HA

yocalebo:

PR #4999, opened 2020-06-23
Fix bugs introduced in #4959

themylogin:
@yocalebo what did you intended to do with https://github.com/freenas/freenas/pull/4959/files#diff-c44b6d54b740097b6b338bb38f20ffd3L1004 ? It broke on FreeNAS where we don't have failover plugin

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -1001,7 +1001,7 @@ class InterfaceService(CRUDService):
                             'VLAN MTU cannot be bigger than parent interface.',
                         )
 
-        if not await self.middleware.call('failover.licensed'):
+        if await self.middleware.call('system.is_freenas') or not await self.middleware.call('failover.licensed'):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -1001,7 +1001,7 @@ async def _common_validation(self, verrors, schema_name, data, itype, update=Non
                             'VLAN MTU cannot be bigger than parent interface.',
                         )
 
-        if not await self.middleware.call('failover.licensed'):
+        if await self.middleware.call('system.is_freenas') or not await self.middleware.call('failover.licensed'):
```
william-gr, line 1004: failover plugin should always be available in 12, no?

william-gr: Ah, this is 11.3, ooops

themylogin:
backport

themylogin:
backport

PR #5006, opened 2020-06-24
NAS-106575 / 12.1 / Fix 12 config upload

themylogin:


william-gr:
What if someone uploads a config file from 11.2 or 11.3?

themylogin:
@william-gr migrate113 will handle that

PR #5007, opened 2020-06-24
NAS-106381 / 12.1 / remove /tmp/failover.json

yocalebo:


william-gr:
@yocalebo so what exactly are we actually solving here removing this file other than improving the code?

Reason I ask is because we dont want major changes for stable branches and this would cause a code divergence with master that may not be worth so early in the 12.0 cycle.

yocalebo:
backport

PR #5011, opened 2020-06-24
NAS-106579 / 12.1 / Remove deprecated AD parameters

anodos325:
After redesign of how we perform domain auto-discovery and config
in 12.0 several LDAP-related fields are now no longer used. Add
migration to remove them, and configure the internal LDAP client
to always seal.


src/middlewared/middlewared/alembic/versions/12.0/2020-06-24_17-14_cleanup-ad-parameters.py
```diff
@@ -0,0 +1,29 @@
+"""cleanup AD parameters
+
+Revision ID: 25962b409a1e
+Revises: 8ac8158773c4
```
themylogin, line 4: New revisions have been merged since then

anodos325: Thanks for pointing that out.

themylogin:
@william-gr I've pushed one commit to separate migrations into different directories; please, tell me what do you think

PR #5013, opened 2020-06-24
NAS-106578 / 12.0 / regenerate /tmp/failover.json on pool create or export

yocalebo:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1548,6 +1548,11 @@ async def unlabel(disk):
         # scrub needs to be regenerated in crontab
         await self.middleware.call('service.restart', 'cron')
 
+        # If this is an HA system, then we need to restart the failover service
+        # so that /tmp/failover.json is regenerated
+        if await self.middleware.call('failover.licensed'):
+            await self.middleware.call('service.restart', 'failover')
```
william-gr, line 1554: Hmm why not register a post.pool_export hook in failover.py?

yocalebo: @william-gr to keep change as minimal as possible since this is now stable.

william-gr: Ok, makes sense, thanks

PR #5016, opened 2020-06-24
Do not permit users to set root as a guest account

anodos325:
This one should not need explanation.


anodos325:
> Is this validation applicable on create as well ?

Not a CRUD method.

PR #5017, opened 2020-06-25
NAS-106582 / 12.1 / allow updating via webUI from master to internal

yocalebo:


yocalebo:
Yeah, was the reason why I just return True. We shouldn't be pushing INTERNAL builds publicly 😉

PR #5022, opened 2020-06-25
NAS-106586 / 12.1 / Refresh failover status before starting services after becoming master

anodos325:
If we don't refresh failover status prior to restarting services, then we
may end up passing a stale cached status to etc scripts and cause
services to come up misconfigured.


yocalebo:
We need to refresh status on backup event too in the rare event we have a "clean" failover.

PR #5025, opened 2020-06-25
Remove legacy SMB-related items from failover events

anodos325:
We no longer maintain sentinel files to indicate users / groups. ix-ldap script no longer exists. We should also refresh failover
status on switching to carp backup.


william-gr:
backport

PR #5030, opened 2020-06-25
QE-356 / 12.0 / Changing TN Core and Scale API test to support pytest-dependency 

ericbsd:
QE-357 Moving HA test from releng provision to the TN API test


ericbsd:
> Why is this being opened up as a separate PR in 12.0 branch when the 12.1 branch seems to be the same? Why not just backport so that we don't have to review twice?

AH I was not aware of that option

yocalebo:
> > Why is this being opened up as a separate PR in 12.0 branch when the 12.1 branch seems to be the same? Why not just backport so that we don't have to review twice?
> 
> AH I was not aware of that option

Yep, all you need to do is add the `backport 12.0` label on the PR in the master repo. Then, whenever your PR gets merged in the master repo the bugclerk will automatically cherry-pick everything back to the 12.0 repo. 👍

ericbsd:
Ok cool I will close that then

PR #5032, opened 2020-06-25
NAS-106587 / 12.1 / fix keyerror traceback when disabling wsd

yocalebo:


william-gr:
Just curious where the KeyError happens? service.start with a unknown name should raise it, or didnt in the past.

yocalebo:
@william-gr Seems like this is fallout from service redesign in the past. The same keyerror happened with netbios and nbmd. I'm talking to @anodos325 right now to see what is best/correct way of fixing this particular issue.

PR #5045, opened 2020-06-26
NAS-106601: Don't alert on non-fatal rsync return codes

blkeller:
This is a fix for JIRA issue NAS-106601:
https://jira.ixsystems.com/browse/NAS-106601

Would it be OK if I also backported this to the 11.3 stable branch?


themylogin:
backport

PR #5075, opened 2020-06-30
Fix regression in check_updated_keytab()

anodos325:


anodos325:
Directly cherry-picked fix into PR for 12.0 since it wasn't merged yet.

PR #5077, opened 2020-06-30
NAS-106639 / 12.1 / Fix traceback in ldap.conf generation script when AD enabled

anodos325:
Ensure that our CACERT file is always populated, but remove any
parameters that are None. This will allow libads to use our GUI
cacerts if for some reason the idmap backend requires it, which
may be a requirement if environment has a trusted domain with an
external LDAP server providing IDs via idmap_rfc2307.

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/openldap/ldap.conf
+++ b/src/middlewared/middlewared/etc_files/local/openldap/ldap.conf
@@ -22,11 +22,13 @@
 
         ad = middleware.call_sync('activedirectory.config')
         ad_enabled = ad['enable']
+        idmap = middleware.call_sync('idmap.query',
+                                     [('name', '=', 'DS_TYPE_ACTIVEDIRECTORY')],
+                                     {'get': True})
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/etc_files/local/openldap/ldap.conf
```diff
@@ -22,11 +22,13 @@
 
         ad = middleware.call_sync('activedirectory.config')
         ad_enabled = ad['enable']
+        idmap = middleware.call_sync('idmap.query',
+                                     [('name', '=', 'DS_TYPE_ACTIVEDIRECTORY')],
+                                     {'get': True})
 
-        if ad['enable'] and ad['idmap_backend'] in ["rfc2307", "ldap"]:
+        if ad['enable'] and idmap in ["rfc2307", "ldap"]:
```
sonicaj, line 29: Is `idmap` going to be a string at this point ?

anodos325: Heh. This means I've been sitting at the computer too long.

PR #5078, opened 2020-06-30
NAS-106640 / 12.1 / Do not update grub on first boot when system is ready

sonicaj:
We used to update grub on first boot after system was in ready state to account for the initial-install BE, however with latest changes introduced, grub is updated when system is in ready state as opposed to during boot which means that we don't have to update grub anymore on firstboot exclusively

### DIFF
--- a/src/middlewared/middlewared/plugins/system.py
+++ b/src/middlewared/middlewared/plugins/system.py
@@ -40,8 +40,6 @@ SYSTEM_READY = False
 # Flag telling whether the system is shutting down
 SYSTEM_SHUTTING_DOWN = False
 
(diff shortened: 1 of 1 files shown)


themylogin:
Latest changes being what?

sonicaj:
Checkpoints in etc plugin

sonicaj:
To clarify, we had a race condition with ix-etc and local-fs.target, we tried to mount fstab for SCALE and that failed due to the race condition. So once the system is in ready state, we mount fstab and then generate grub config instead of doing that in initial etc checkpoint. 

With this change, the first boot grub update is not required

PR #5084, opened 2020-07-01
NAS-106648 / 12.1 / Make registry configuration aware of locked datasets

anodos325:
Samba reads its running configuration from registry.tdb. Sharing.smb.update
will apply any share config updates to the registry. The exact behavior required
in the update method depends on that status of the datasets underlying the
old path and new path for the share. ASCII diagram added to relevant section.


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -732,29 +732,78 @@ async def do_update(self, id, data):
         if old['purpose'] != new['purpose']:
             await self.apply_presets(new)
 
+        old_is_locked = (await self.query([("id", "=", id)], {'get': True}))['locked']
```
sonicaj, line 735: We should use get_instance for this

anodos325: Yeah, forgot to fix that. Good catch.

PR #5085, opened 2020-07-01
QE-358 / 12.1 / Update plugins test to supports community plugins

ericbsd:


tests/api2/test_310_plugin.py
```diff
@@ -284,8 +281,8 @@ def test_29_get_list_of_available_plugins_job_id_on_custom_repos(request):
     depends(request, ["pool_04"], scope="session")
     global JOB_ID
     payload = {
-        "plugin_repository": repos_url,
-        "branch": "11.3-RELEASE"
+        "plugin_repository": repos_url2,
```
sonicaj, line 284: It should work with `repos_url2` nicely but I don't believe `repos_url2` is going to be updated as frequently as `repos_url`. We should I think be using that.

ericbsd: it is only to test custom repos not to test the official repos, repos_url was wrong from the start. If it is to use repos_url I should remove it since using repos_url is a duplicate of test 5.

sonicaj: Ah, thanks for clarifying. My motivation was that we would catch any regressions to `transmission` plugin as well but you got that covered ; )

PR #5102, opened 2020-07-05
NAS-104341 / 12.1 / Platform dependent VM plugin

sonicaj:


src/middlewared/middlewared/plugins/sysctl/sysctl_info_linux.py
```diff
@@ -0,0 +1,66 @@
...
+from .sysctl_info_base import SysctlInfoBase
+
+
+ZFS_MODULE_PARAMS_PATH = '/sys/module/zfs/parameters'
+
+
+class SysctlService(Service, SysctlInfoBase):
+
+    def get_value(self, sysctl_name):
+        raise NotImplementedError
```
themylogin, line 16: Why `NotImplementedError`?

src/middlewared/middlewared/plugins/sysctl/sysctl_info_linux.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def not_found_error(self, name):
```
themylogin, line 18: Why a method for something used only once?

sonicaj: I think it will be useful moving on and I was using it in multiple places before as well but then changed my code to use something different. However, if you think we should not have this, I can make the required change : )

src/middlewared/middlewared/plugins/vm/pci_freebsd.py
```diff
@@ -0,0 +1,60 @@
+import re
+
+from middlewared.service import CallError, Service
+from middlewared.utils import run
+
+from .pci_base import PCIInfoBase
+
+
+RE_AMDVI = re.compile(br'vmm\.amdvi\.enable: 1')
```
william-gr, line 9: Why the `b` modifier?

sonicaj: Iirc We are unable to decode the output of subprocess calls on some machines so the regex match is done without decoding.

src/middlewared/middlewared/plugins/sysctl/sysctl_info_linux.py
```diff
@@ -0,0 +1,69 @@
...
+            raise CallError(f'Unable to retrieve value of "{sysctl_name}" sysctl : {cp.stderr.decode()}')
+        return cp.stdout.decode().split('=')[-1].strip()
+
+    def get_arc_max(self):
+        return self.get_arcstats()['c_max']
+
+    def get_arc_min(self):
+        return self.get_arcstats()['c_min']
+
+    def read_value_from_file(self, path, name):
```
themylogin, line 30: Where is this method used?

PR #5107, opened 2020-07-06
Change the order of the API test for 12.1 HA

ericbsd:
rename test_009_interface.py to test_005_interface.py
rename test_005_pool_dataset_encryption.py to test_007_pool_dataset_encryption.py
rename test_007_pool.py to test_009_pool.py
remove controller_a=ha from test_009_pool.py to run it using vip
remove fenced --force from test_003_network.py


ericbsd:
Test on 12.1 nightly and it is working

PR #5119, opened 2020-07-07
NAS-106697 / 12.1 / We'll have to replicate system dataset if we want full replication

themylogin:


william-gr:
@themylogin why is this necessary? The ticket has no explanation at all

themylogin:
@william-gr `zfs -R` ("full replication") can't exclude datasets. zettarepl raises validation error when we try to exclude `.system` from full replication.

PR #5124, opened 2020-07-07
NAS-106703 / 12.1 / Ensure that permissions for tmp are correct during smb.configure

anodos325:
Samba uses /tmp for its IPC share path and incorrect permissions will break RPC preventing Windows clients from connecting.


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -78,6 +78,7 @@ class SMBPath(enum.Enum):
     RUNDIR = ('/var/run/samba4', '/var/run/samba', 0o755, True)
     LOCKDIR = ('/var/lock', '/var/lock', 0o755, True)
     LOGDIR = ('/var/log/samba4', '/var/log/samba', 0o755, True)
+    IPCSHARE = ('/var/tmp', '/tmp', 0o777, True)
```
william-gr, line 81: Should this be 0x1777?

anodos325: Yes

PR #5136, opened 2020-07-08
NAS-106717 / 12.0 / Disable auto-rollback for vfs_tmprotect

anodos325:
This particular feature should probably only be exposed in master until it can be properly tested with multiple
clients and usage patterns.


william-gr:
@anodos325 should we backport to 12.1?

anodos325:
> @anodos325 should we backport to 12.1?

No. I'm currently pretending that I will have plenty of time to have this ready for 12.1.

PR #5138, opened 2020-07-08
NAS-106651 / 12.1 / configure advskew for CARP on HA systems correctly

yocalebo:


src/middlewared/middlewared/plugins/interface/configure.py
```diff
@@ -137,14 +137,18 @@ def configure(self, data, aliases, wait_dhcp=False, options=None):
...
-        # carp must be configured after removing addresses
-        # in case removing the address removes the carp
         if carp_vhid:
-            if not self.middleware.call_sync('system.is_freenas') and not advskew:
-                if self.middleware.call_sync('failover.node') == 'A':
+            if self.middleware.call_sync('failover.licensed') and not advskew:
+                if 'NO_FAILOVER' in self.middleware.call_sync('failover.disabled_reasons'):
+                    if self.middleware.call_sync('failover.vip.get_states')[0]:
+                        advskew = 20
+                    elif self.middleware.call_sync('failover.status') != 'MASTER':
```
william-gr, line 145: Should this just be an else? otherwise advskew may be undefined

yocalebo: Yep, I'll fix.

src/middlewared/middlewared/plugins/interface/configure.py
```diff
@@ -137,14 +137,18 @@ def configure(self, data, aliases, wait_dhcp=False, options=None):
                     self.logger.debug('{}: removing possible valid_lft and preferred_lft on {}'.format(name, addr))
                     iface.replace_address(addr)
 
-        # carp must be configured after removing addresses
-        # in case removing the address removes the carp
```
william-gr, line 141: This comment is still valid, no?

yocalebo: I don't understand the comment honestly. This is configuring CARP on newly created interfaces. It's not creating CARP after removing addresses. Do we allow the ability to remove IP addresses without removing the virtual IP?

william-gr: This code does not only run on new interfaces, it used to sync interface which means addresses can be removed and CARP needs to be reconfigured

yocalebo: I don't see how that's possible because when you remove an interface that means it already has an `advskew` set. So on line 110, we break out of this method because `advskew` gets set to something.

william-gr: Not removing the interface, removing an alias

PR #5143, opened 2020-07-09
NAS-106685 / 11.3 / Double-check the return value of disk.identifier_to_device in disk.sync_all

themylogin:


william-gr:
@themylogin not necessary in 12.1?

themylogin:
@william-gr I'll backport from 12.0 to 12.1, that'll be easier as I know there would be conflicts and I'll only need to resolve them once

PR #5149, opened 2020-07-09
NAS-106735 / 12.1 / Add support for nested VM's in SCALE

sonicaj:


sonicaj:
@william-gr I am not sure. Looks like `host-passthrough` has better performance and `host-model` just gives the edge on migrating domain which we are not doing. Thoughts https://wiki.openstack.org/wiki/LibvirtXMLCPUModel#:~:text=%22host%2Dpassthrough%22%20%2D%20this,the%20host%20CPU%20is%20matched.&text=The%20guest%20can%20only%20be%20migrated%20to%20an%20exactly%20matching%20host%20CPU. ?

william-gr:
```
CPU is specified by providing a base CPU model name (which is a shorthand for a set of feature flags), a set of additional feature flags, and the topology (sockets/cores/threads). The libvirt KVM driver provides a number of standard CPU model names (defined in /usr/share/libvirt/cpu_map.xml):

"486", "pentium", "pentium2", "pentiumpro", "coreduo", "n270", "pentiumpro", "qemu32", "kvm32", "cpu64-rhel5", "cpu64-rhel5", "kvm64", "pentiumpro", "Conroe" "Penryn", "Nehalem", "Westmere", "pentiumpro", "cpu64-rhel5", "cpu64-rhel5", "Opteron_G1", "Opteron_G2", "Opteron_G3, "Opteron_G4"
```

There is this whole list of possible values so I am not sure we want restrict ourselves to only passthrough?

sonicaj:
@william-gr I will update the migration once the other PR gets in master, please let me know if everything looks in place apart from the migration. Thank you

src/middlewared/middlewared/plugins/vm/supervisor/supervisor_linux.py
```diff
@@ -57,3 +57,29 @@ def devices_xml(self):
 
         devices.extend([create_element('serial', type='pty'), create_element('video')])
         return create_element('devices', attribute_dict={'children': devices})
+
+    def cpu_xml(self):
+        cpu_elem = super().cpu_xml()
+        if self.vm_data['cpu_mode']:
+            cpu_elem.set('mode', self.vm_data['cpu_mode'].lower())
+        elif self.vm_data['cpu_model']:
```
william-gr, line 65: Do we really need two fields and use features+vendor xml?

On libvirt if I choose Opteron_G1 all it does in XML is:

<cpu mode="custom" match="exact" check="none">
    <model fallback="forbid">Opteron_G1</model>
</cpu>

sonicaj: The 2 fields serve different purposes.
`cpu_mode` defines how the CPU will be configured, would it be a custom model or host-passthrough/host-model.
`cpu_model` shows which CPU will be used if custom cpu mode is desired. 
So in my opinion we should not be mixing them by using a single field because they represent different things.

About the features + vendor xml, I am doing that for completeness, we could in the future allow control at a feature level allowing user to force/require somethings. I am not 100% sure if there would be any drawbacks if we don't specify this, but I think there shouldn't be. If you think we should remove that, please let me know and I can confirm and remove.

william-gr: Yes, I think we dont need features + vendor as that only complicates things for no real benefit.

If we can avoid reading ~65 files and running 65 subprocesses to get the models, even better.

src/middlewared/middlewared/plugins/vm/vms.py
```diff
@@ -52,10 +54,15 @@ async def bootloader_options(self):
...
         vm['devices'] = await self.middleware.call('vm.device.query', [('vm', '=', vm['id'])])
         vm['status'] = await self.middleware.call('vm.status', vm['id'])
+        if osc.IS_FREEBSD:
+            vm.pop('cpu_mode', None)
+            vm.pop('cpu_model', None)
         return vm
 
     @accepts(Dict(
         'vm_create',
+        Str('cpu_mode', default=None, null=True, enum=[None, 'HOST-MODEL', 'HOST-PASSTHROUGH']),
```
william-gr, line 64: CUSTOM instead of None?

PR #5153, opened 2020-07-09
NAS-106732 / 12.1 / fix adding/deleting aliases on HA systems

yocalebo:


yocalebo:
@william-gr This actually fixes the inverse of adding aliases (deleting) as well.

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1501,15 +1501,28 @@ async def hook_setup_ha(middleware, *args, **kwargs):
...
+
+        # In the event #2 happens listed above, there is a race condition that
+        # must be accounted for. When an alias is added to an already configured interface,
+        # a CARP event will be triggered and the interface will go from MASTER to INIT->BACKUP->MASTER which
+        # generates a devd event that is processed by the failover.event plugin.
+        # It takes a few seconds for the kernel to transition the CARP interface from BACKUP->MASTER.
+        # However, we refresh the failover.status while this interface is transitioning.
+        # This means that failover.status returns 'ERROR'. Since we have already checked that there
+        # is a zpool, we simply account for the 'ERROR' status and continue.
+        if 'MASTER' or 'ERROR' in await middleware.call('failover.status'):
```
william-gr, line 1525: That will be always True because bool('MASTER') is True

william-gr: Also should we run this only if the master flag is true for the given node, to ensure there will be no recursion of calls?

PR #5155, opened 2020-07-10
NAS-106713 / 11.3 / Prevent queued and then disabled cron tasks from executing.

themylogin:
Also prevent rsync jobs from queueing up.

I'll backport from 12.0 to 12.1


william-gr:
backport

PR #5170, opened 2020-07-13
NAS-106778 / 12.1 / Allow retrieving cpu mode choices

sonicaj:


sonicaj:
It was for the UI, if you think we should hard code in the UI, we can do that ?

william-gr:
Yes, we can hardcode these values.

sonicaj:
Great. I'll just close this.

PR #5181, opened 2020-07-14
NAS-106719 / 12.1 / Terminate orphaned multiprocessing bits

themylogin:


themylogin:
@william-gr because we don't do clean multiprocessing shutdown https://github.com/freenas/freenas/blob/master/src/middlewared/middlewared/main.py#L1431

william-gr:
Should we be trying to kill the process group?

e.g.

pgrp = os.getpgid(pid)
os.killpg(pgrp, signal.SIGKILL)

themylogin:
@william-gr before doing `os._exit`?

william-gr:
Yes, would that be enough to make sure all children are killed?

src/middlewared/middlewared/main.py
```diff
@@ -1122,6 +1124,13 @@ async def _run_in_conn_threadpool(self, method, *args, **kwargs):
         """
         return await self.run_in_executor(self.__ws_threadpool, method, *args, **kwargs)
 
+    def __terminate_orphaned_multiprocessing(self):
```
william-gr, line 1127: Given that we are now killing pg, do we still need this?

I feel somewhat unsure about killing generic python processes like that, which we are not 100% being leftovers from middleware.

themylogin: These processes will be left alive if middleware was killed by OOM-killer on Linux (or other SIGKILL).

I think it's safe to kill them considering we only do so when process parent PID is 1 which won't be true if a multiprocessing python program that created them is still alive.

william-gr: Just dont think its our job to handle that if this is the case, sounds more like a python multiprocessing issue?

PR #5186, opened 2020-07-14
NAS-106785 / 12.1 / Revert "Rename files in /root that should not be overwritten if changed"

yocalebo:
This reverts commit 8733acbad33436bc042241c36957767a853f23e6.


yocalebo:
backport

PR #5200, opened 2020-07-15
NAS-106806 / 12.1 / treat CARP NONE state as INIT

yocalebo:


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/detect_virtual_ips_freebsd.py
```diff
@@ -35,6 +35,9 @@ async def get_states(self, interfaces=None):
                 backups.append(iface['name'])
             elif iface['state']['carp_config'][0]['state'] == 'INIT':
```
william-gr, line 36: Not a requirement, but just a thought:

`elif iface['state']['carp_config'][0]['state'] in (None, 'INIT'):`

PR #5209, opened 2020-07-15
NAS-106812 / 12.1 / Make retrieving cert extensions non-fatal

sonicaj:
Some certificates can have extensions with binary data which we can't parse without explicit mapping for each extension. The current case covers the most of extensions nicely and if it's required to map certain extensions which can't be handled by above we can do as users request

### DIFF
--- a/src/middlewared/middlewared/plugins/crypto.py
+++ b/src/middlewared/middlewared/plugins/crypto.py
@@ -433,15 +433,15 @@ class CryptoKeyService(Service):
                 cert_info['san'] = [s.strip() for s in ext.__str__().split(',') if s]
 
             try:
-                cert_info['extensions'][
-                    re.sub(r"^(\S)", lambda m: m.group(1).upper(), ext.get_short_name().decode())
-                ] = ext.__str__()
-            except crypto.Error:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -432,9 +432,16 @@ def get_x509_subject(self, obj):
...
                 cert_info['san'] = [s.strip() for s in ext.__str__().split(',') if s]
 
-            cert_info['extensions'][
-                re.sub(r"^(\S)", lambda m: m.group(1).upper(), ext.get_short_name().decode())
-            ] = ext.__str__()
+            try:
+                cert_info['extensions'][
+                    re.sub(r"^(\S)", lambda m: m.group(1).upper(), ext.get_short_name().decode())
+                ] = ext.__str__()
+            except crypto.Error:
```
william-gr, line 439: Should we at lost log it to know something failed?

sonicaj: Yup, also I try to at least retrieve the name if possible of the extension

PR #5213, opened 2020-07-16
NAS-100025 / 12.1 / Allow disabling outbound network activities

themylogin:
@william-gr this does not affect:
* ACME
* Cloud Sync
* KMIP
* NTP
* Replication
* Rsync
* VMWare

(and God knows what else)


william-gr:
Have not forgotten this, just thinking a bit more on a few ideas and discussing with Support first.

themylogin:
Good idea, we can do it.

Do we want to register pieces corresponding to services I've listed above? (services that do not have much sense without external networking)

william-gr:
Yes, I think we should register anything that can use outbound network. We can try to give a higher level description to what it means/does.

william-gr:
@themylogin ping?

themylogin:
@william-gr I am not sure how can we disable NTP. Stop the service completely?

Also what do we do with AD/LDAP? NFS Kerberos?

Also I have not covered some very specific (and initiated on user's explicit desire) cases like ssh-keyscan or remote TrueNAS API call when creating SSH connection, what do you think should we do with these?

william-gr:
@themylogin For NTP, sicne we cant control that service separately, I think so.

I think what we have right now is a good start, we cam improve later on user feedback.

However they way this is going seems like it would make more sense to be in the network section?
Curious what @freenas/middleware opinions.

src/middlewared/middlewared/plugins/acme_protocol.py
```diff
@@ -414,6 +418,8 @@ def update_txt_record(self, data):
 
     @private
     def update_txt_record_route53(self, domain, challenge, key, access_key_id, secret_access_key):
+        self.middleware.call_sync('system.advanced.will_perform_network_activity', 'acme')
```
sonicaj, line 421: We should do this above at line 399 as this will be specific to route53, however middleware consumes the above endpoint and these are called internally by the service

sonicaj:
@william-gr I do think that this makes more sense in the network section based on the options we expose.

anodos325:
I don't think that disabling NTP should be allowed if AD / or LDAP + Kerberos are enabled. Correct time is _required_ for these protocols to work.

src/middlewared/middlewared/plugins/rsync.py
```diff
@@ -734,3 +736,4 @@ async def restart_reload_services(self, attachments):
 async def setup(middleware):
     await middleware.call('pool.dataset.register_attachment_delegate', RsyncModuleFSAttachmentDelegate(middleware))
     await middleware.call('pool.dataset.register_attachment_delegate', RsyncFSAttachmentDelegate(middleware))
+    await middleware.call('network.general.register_activity', 'rsync', 'rsync')
```
william-gr, line 739: Rsync, uppercase as its a name?

src/middlewared/middlewared/plugins/vmware.py
```diff
@@ -634,3 +641,7 @@ def _alert_vmware_login_failed(self, vmsnapobj, e):
 
     def _delete_vmware_login_failed_alert(self, vmsnapobj):
         self.middleware.call_sync("alert.oneshot_delete", "VMWareLoginFailed", vmsnapobj["hostname"])
+
+
+async def setup(middleware):
+    await middleware.call('network.general.register_activity', 'vmware', 'VMware')
```
william-gr, line 647: VMWare is probably too broad of a term, should we say VMWare Snapshots or something?

PR #5216, opened 2020-07-16
NAS-106827 / 12.0 / Remove extra debug statements from directory service refresh

anodos325:


src/middlewared/middlewared/etc_files/pwd_db.py
```diff
@@ -2,4 +2,5 @@
 
 
 async def render(service, middleware):
+    middleware.logger.debug("XXXX::: GENERATING MASTERPW")
```
william-gr, line 5: Why this

anodos325: Whoops. Didn't realize the branch wasn't clean.

PR #5220, opened 2020-07-17
NAS-106843 / 12.1 / FC ports in target mode by default

william-gr:
Ticket:	NAS-106843

### DIFF
--- a/src/freenas/boot/loader.conf
+++ b/src/freenas/boot/loader.conf
@@ -34,12 +34,6 @@ if_qlxgbe_load="YES"
 ocs_fc_load="YES"
 smartpqi_load="YES"
 
-# Put FC ports in initiator mode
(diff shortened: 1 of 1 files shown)


amotin:
Role 0 is not target, it is nothing.

william-gr:
@amotin does this need to be set to = 0? or is that the default? Should I remove these lines?

william-gr:
@amotin ping

amotin:
When the kernel is built with ISP_TARGET_MODE (as we do) the default is 0, so setting to 0 should not be needed.  When ISP_TARGET_MODE is not used (default FreeBSD kernel) the default is 1, that is why on FreeNAS we've had setting to 2, which is initiator, since some people just used FC disk JBODs without tuning.

PR #5221, opened 2020-07-17
NAS-106840 / 12.1 / prevent invalid VHIDs

yocalebo:


src/middlewared/middlewared/plugins/network.py
```diff
@@ -1075,6 +1075,12 @@ async def _common_validation(self, verrors, schema_name, data, itype, update=Non
                     'Failover needs to be disabled to perform network configuration changes.'
                 )
 
+                if len(data.get('failover_vhid', -1) <= 0 or > 255:
```
william-gr, line 1078: How about the validators in the schema, using Min and Max?
Or is this better for some reason? I dont really know

yocalebo: I've simply forgotten about those validators actually. Much better approach, I'll fix.

PR #5222, opened 2020-07-17
NAS-104615 / 12.1 / Configure kernel dumps in SCALE

sonicaj:


src/freenas/etc/systemd/system/kdump-tools-dump.service.d/override.conf
```diff
@@ -0,0 +1,3 @@
+[Service]
+ExecStart=
```
william-gr, line 2: Why two ExecStart is necessary?

src/freenas/usr/local/bin/execute_dump.sh
```diff
@@ -0,0 +1,53 @@
...
+    LIMIT 1
+  ")
+
+if [ -z "$sys_pool"]; then
+  echo "No system dataset pool found. Aborting dump."
+  reboot -f
+fi
+
+zpool list | grep -q "$sys_pool"
+if [ $? -eq 1 ]; then
```
william-gr, line 33: You mean -eq 0?

src/freenas/etc/systemd/system/kdump-tools-dump.service.d/override.conf
sonicaj: First one is to clear the default `ExecStart` action, if it's not done it will execute it's core dump saving script

src/freenas/usr/local/bin/execute_dump.sh
william-gr: No, nevermind :)

sonicaj: Nope, if we are not able to find that pool, then we try to import it

src/middlewared/middlewared/etc_files/grub.py
```diff
@@ -1,17 +1,22 @@
...
-        await run(["truenas-grub.py"], encoding="utf-8", errors="ignore")
-    except subprocess.CalledProcessError as e:
-        middleware.logger.error("truenas-grub.py error:\n%s", e.stderr)
-        raise
+def render(service, middleware):
+    tn_grub = module_from_file('truenas-grub', '/usr/local/bin/truenas-grub.py')
+    config = tn_grub.retrieve_grub_config()
+    kdump_enabled = middleware.call_sync('system.advanced.config')['kdump_enabled']
+    if kdump_enabled:
+        config['GRUB_CMDLINE_LINUX'].append(
```
themylogin, line 12: Why can't we do this in `truenas-grub` script?

src/middlewared/middlewared/plugins/etc.py
```diff
@@ -144,6 +144,10 @@ class EtcService(Service):
              'local_path': 'local/proftpd.conf'},
             {'type': 'py', 'path': 'local/proftpd'},
         ],
+        'kdump': [
+            {'type': 'mako', 'path': 'default/kdump-tools',
+             'local_path': 'default/kdump-tools.mako', 'platform': 'Linux'},
```
themylogin, line 149: Can't we just strip `.mako` from template file name and thus avoid repeating ourselves?

src/middlewared/middlewared/alert/source/kdump.py
```diff
@@ -0,0 +1,9 @@
+from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, SimpleOneShotAlertClass
+
+
+class KdumpNotReadyAlertClass(AlertClass, SimpleOneShotAlertClass):
+    deleted_automatically = False
+    level = AlertLevel.CRITICAL
```
themylogin, line 6: Is it really CRITICAL?

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1444,6 +1456,16 @@ async def _event_system(middleware, event_type, args):
...
+
+        if osc.IS_LINUX and not FIRST_BOOT and (await middleware.call('system.advanced.config'))['kdump_enabled']:
+            cp = await run(['kdump-config', 'status'], check=False)
+            if cp.returncode:
+                middleware.logger.error('Failed to retrieve kdump-config status: %s', cp.stderr.decode())
+            else:
+                if not RE_KDUMP_CONFIGURED.findall(cp.stdout.decode()):
+                    await middleware.call('alert.oneshot_create', 'KdumpNotReady', None)
+                else:
+                    await middleware.call('alert.oneshot_delete', 'KdumpNotReady', None)
```
themylogin, line 1468: What if we had the alert, disabled kdump and rebooted? It will stay forever in this case.

src/middlewared/middlewared/etc_files/grub.py
sonicaj: Because `truenas-grub` is executed during installation of scale as well and we don't have middleware running at that point to calculate crash kernel memory.

themylogin: What's so complicated in the `required_crash_kernel_memory` that it has to be a middleware method?

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1444,6 +1455,19 @@ async def _event_system(middleware, event_type, args):
         # If it is not defined yet, it will try to define
         if birthday is None:
             asyncio.ensure_future(_update_birthday(middleware))
+
+        if osc.IS_LINUX:
+            if (await middleware.call('system.advanced.config'))['kdump_enabled']:
+                cp = await run(['kdump-config', 'status'], check=False)
+                if cp.returncode:
+                    middleware.logger.error('Failed to retrieve kdump-config status: %s', cp.stderr.decode())
```
themylogin, line 1463: Again, alert won't be deleted in this code path

src/middlewared/middlewared/plugins/system_/kdump_linux.py
```diff
@@ -0,0 +1,22 @@
+import math
+
+from middlewared.service import accepts, Service
+
+
+class SystemService(Service):
+
+    @accepts()
+    async def required_crash_kernel_memory(self):
```
themylogin, line 9: Is this used anywhere?

src/middlewared/middlewared/alert/source/kdump.py
```diff
@@ -0,0 +1,9 @@
+from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, SimpleOneShotAlertClass
+
+
+class KdumpNotReadyAlertClass(AlertClass, SimpleOneShotAlertClass):
+    deleted_automatically = False
+    level = AlertLevel.WARNING
+    category = AlertCategory.SYSTEM
+    title = 'System Not Ready For Kdump'
+    text = 'System is not ready for Kdump, please refer to logs.'
```
themylogin, line 9: Which logs?

src/middlewared/middlewared/plugins/system.py
sonicaj: There is a possibility it is configured in this case. We were just unable to retrieve the status of kdump-config which in various scenarios i have tested should not happen and is just a fail safe

PR #5229, opened 2020-07-20
NAS-106858 / 12.1 / Clarify bootloader options to be more verbose

sonicaj:


src/middlewared/middlewared/plugins/vm/vms.py
```diff
@@ -12,7 +12,12 @@
 from .vm_supervisor import VMSupervisorMixin
 
 
-BOOT_LOADER_OPTIONS = ['UEFI', 'UEFI_CSM'] + (['GRUB'] if osc.IS_FREEBSD else [])
+BOOT_LOADER_OPTIONS = {
+    'UEFI': 'UEFI',
+    'UEFI_CSM': 'Legacy BIOS',
+}
+if osc.IS_FREEBSD:
+    BOOT_LOADER_OPTIONS['GRUB'] = '11.2 Rancher VM\'s'
```
william-gr, line 20: As far as I know thats not just 11.2 rancher VMS and in theory could be used for anything? It was made more generic

PR #5234, opened 2020-07-20
NAS-104359 / 20.10 / VRRP implementation for SCALE HA

yocalebo:


src/middlewared/middlewared/etc_files/keepalived.py
```diff
@@ -0,0 +1,52 @@
...
+        middleware.logger.warning(
+            'Chassis position for this node could not be determined.'
+        )
+        return
+
+    info = middleware.call_sync('interface.query')
+
+    with open(KEEPALIVED_FILE, 'w') as f:
+        for i in info:
+            f.write(f'vrrp_instance {i["int_interface"]} {{\n)'
```
william-gr, line 31: Wouldnt a mako etc make more sense for this?

src/middlewared/middlewared/plugins/interface/vrrp_linux.py
```diff
@@ -0,0 +1,68 @@
...
+            'network.interfaces',
+            [('int_interface', '=', ifname)],
+        )
+        configured_vips = [
+            i['int_vip'] for i in info if i['int_vip']
+        ]
+
+        # if there are no VIPs then there is no reason to continue
+        if not configured_vips:
+            self.logger.warning(
```
william-gr, line 28: For all logging operations we should use %s/%r/%s instead of f-string, its useful to for logging to not receive the already formatted string because it can aggregate on same errors messages in the future.

src/middlewared/middlewared/etc_files/keepalived.conf.mako
```diff
@@ -0,0 +1,51 @@
...
+
+    if licensed and node == 'MANUAL':
+        middleware.logger.error(
+            'Chassis position could not be determined.'
+            ' Keepalived config not generated.'
+        )
+
+    config = middleware.call_sync('failover.config')
+
+    advert_int = .68 # default 2 second timeout
```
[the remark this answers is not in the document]
yocalebo: We wait 2 seconds by default with CARP on freeBSD before we start to transition from one state to another. We have to do that part in middleware with CARP. However, VRRP does that for us. So .68 is the advertisement interval that makes VRRP wait 2 seconds before it sends a state change message to us.

src/middlewared/middlewared/etc_files/keepalived.conf.mako
```diff
@@ -0,0 +1,51 @@
...
+    if config['timeout'] > 2:
+        # https://tools.ietf.org/html/rfc5798#section-6.1
+        # Formula is there so I solved for (y) to get advert_int
+        # I round to the 100th decimal place since VRRP version 3
+        # uses centisecond intervals by default
+        advert_int = round(((128 * config['timeout']) / 385), 2)
+
+    info = middleware.call_sync('datastore.query', 'network_interfaces')
```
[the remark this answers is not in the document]
yocalebo: The problem I have with interface.query is that the top-level `aliases` key is a list of IP addresses on the interface, however, I don't know which one is the primary IP. So when I generate this file I need to know which one is the primary IP address. Or maybe I always takes the 0th entry in that list?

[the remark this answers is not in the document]
william-gr: Is there really a notion of primary IP? In practice everything is an alias as far as I know. At least in FreeBSD.

Problem with direct datastore manipulation is that things can change behind our back and break things, while interface.query is supposed to follow an API and data can be changed there.
I think its fair to say first entry is the "first/primary"

src/middlewared/middlewared/etc_files/keepalived.conf.mako
```diff
@@ -0,0 +1,51 @@
+<%
+    licensed = middleware.call_sync('failover.licensed'):
+    node = middleware.call_sync('failover.node')
+
+    if licensed and node == 'MANUAL':
+        middleware.logger.error(
+            'Chassis position could not be determined.'
+            ' Keepalived config not generated.'
+        )
```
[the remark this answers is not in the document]
yocalebo: Sorry, not sure I understand. Do you mean a file that mako generates? Or are you talking about the `/tmp/.ha_mode` file that is written to disk on HA systems?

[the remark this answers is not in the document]
william-gr: If this condition is met, we should abort here and not try to generate keepalived.conf, right?

And this file should ideally be deleted. There is an exception that will make it happen.

src/middlewared/middlewared/etc_files/keepalived.conf.mako
```diff
@@ -0,0 +1,58 @@
...
+    advert_int ${j}
+    % endfor
+    nopreempt
+    virtual_router_id 20
+    priority 254
+    version 3
+    unicast_peer {
+    % for j in i['failover_aliases']:
+        ${j['address']}
+    % endfor
```
william-gr, line 50: Shouldn't it be dependent on node A or B?
If A use i['failover_aliases'] and if B i['aliases'] ?

src/middlewared/middlewared/plugins/device_/vrrp_events_linux.py
```diff
@@ -0,0 +1,25 @@
...
+        os.mkfifo(VRRP_FIFO_FILE)
+    except FileExistsError:
+        pass
+
+    while True:
+        with open(VRRP_FIFO_FILE) as f:
+            middleware.logger.info('vrrp fifo connection established')
+            # all vrrp messages are terminated with a newline
+            for line in f:
+                middleware.call_hook_sync('vrrp.fifo', data=line)
```
william-gr, line 21: Wonder if we should be extra safe here and wrap it all around a try/except so it keeps retrying after X seconds if for any reason we get an exception

src/middlewared/middlewared/plugins/interface/configure.py
```diff
@@ -33,19 +33,17 @@ def configure(self, data, aliases, wait_dhcp=False, options=None):
...
-            ipv4_field = 'int_ipv4address_b'
-            ipv6_field = 'int_ipv6address'
-            alias_ipv4_field = 'alias_v4address_b'
-            alias_ipv6_field = 'alias_v6address_b'
-        else:
-            ipv4_field = 'int_ipv4address'
-            ipv6_field = 'int_ipv6address'
-            alias_ipv4_field = 'alias_v4address'
-            alias_ipv6_field = 'alias_v6address'
+        if self.middleware.call_sync('failover.licensed'):
```
william-gr, line 36: This is a change of behavior, we dont need license to know if node is A and B, I am skeptical about regressions if we change this because we need to do this based on hardware alone, and not if its licensed.

src/middlewared/middlewared/plugins/network.py
```diff
@@ -390,19 +387,23 @@ def query(self, filters, options):
             i['int_interface']: i
             for i in self.middleware.call_sync('datastore.query', 'network.interfaces')
         }
-        is_freenas = self.middleware.call_sync('system.is_freenas')
-        if not is_freenas:
+        licensed = self.middleware.call_sync('failover.licensed')
```
william-gr, line 390: We need to configure internal interfaces before a license is applied so we can contact the other node via middleware, otherwise the license cant be uploaded to the other node.

Think we should check for ENTERPRISE product type

william-gr (changes requested):

src/middlewared/middlewared/etc_files/keepalived.conf.mako
```diff
@@ -0,0 +1,58 @@
...
+    if licensed and node == 'MANUAL':
+        middleware.logger.error(
+            'Chassis position could not be determined.'
+            ' Keepalived config not generated.'
+        )
+	raise FileShouldNotExist()
+
+    config = middleware.call_sync('failover.config')
+
+    advert_int = [.68] # default 2 second timeout
```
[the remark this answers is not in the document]
yocalebo: Because I don't know how to make mako write a single variable by itself. This was the easiest solution at the time. If you know how, please enlighten me 😄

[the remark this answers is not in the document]
yocalebo: nvm, figured it out 😉

src/middlewared/middlewared/etc_files/keepalived.conf.mako
yocalebo: No, this is inside the `unicast_peer` block so it's actually the other sides IP address. However, i'm clearly not accounting for the A and B controller. I'll fix.

src/middlewared/middlewared/plugins/device_/vrrp_events_linux.py
yocalebo: I've tried hard to ensure that the FIFO is created in this method before the keepalived service is started. If the keepalived service starts before middlewared, and it has created the FIFO then the FIFO file gets removed which would cause an exception. If middlewared creates the FIFO, then the file persists even if middlewared is restarted which is what we want. However, I can add an exception for safe measure.

src/middlewared/middlewared/plugins/interface/configure.py
yocalebo: apologies, glanced over that when I was replacing all the `is_freenas` calls and didn't realize it.

src/middlewared/middlewared/plugins/device_/vrrp_events_linux.py
```diff
@@ -0,0 +1,32 @@
...
+                # all vrrp messages are terminated with a newline
+                for line in f:
+                    middleware.call_hook_sync('vrrp.fifo', data=line)
+        except Exception:
+            middleware.logger.error('vrrp fifo connection not established, retrying')
+            # sleep for `TIMEOUT` before trying to open FIFO and send event again
+            time.sleep(TIMEOUT)
+
+
+def setup(middleware):
```
william-gr, line 31: Hum, so this will start the thread regardless of being a HA licensed system?

yocalebo: whoops

src/middlewared/middlewared/plugins/device_/vrrp_events_linux.py
```diff
@@ -0,0 +1,35 @@
...
+
+    # only run on licensed systems
+    if middleware.call_sync('failover.licensed'):
```
william-gr, line 34: Do we need to also do that when the license is updated/installed?

src/middlewared/middlewared/plugins/device_/vrrp_events_linux.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def vrrp_hook_license_update(middleware, prev_product_type, *args, **kwargs):
+
+    # get new product_type
+    new_product_type = middleware.call_sync('system.product_type')
+
+    if prev_product_type != 'SCALE_ENTERPRISE':
+        if new_product_type == 'SCALE_ENTERPRISE':
+            vrrp_fifo_listen(middleware)
```
william-gr, line 18: Hmm, not exactly what I had in mind as we still do not keep track of a running thread and its not clear how this is handled within the event loop as its a blocking call to a thread...


how about:

VRRP_THREAD = None

...

global VRRP_THREAD
if licensed:
    VRRP_THREAD = start_daemon_thread(...)


...


global VRRP_THREAD
if prod_type != new_prod_type and VRRP_THREAD is None:
    VRRP_THREAD = start_daemon_thread(...)

src/middlewared/middlewared/plugins/device_/vrrp_events_linux.py
```diff
@@ -0,0 +1,57 @@
...
+                middleware.logger.error('vrrp fifo connection not established, retrying')
+                # sleep for `TIMEOUT` before trying to open FIFO and send event again
+                time.sleep(TIMEOUT)
+
+
+def setup(middleware):
+
+    # register hook to be called if/when a license has been uploaded
+    # to system changing the product type to 'SCALE_ENTERPRISE'
+    middleware.register_hook('vrrp.post_license_update', vrrp_hook_license_update, sync=False)
```
william-gr, line 53: Sorry, I was meaning used the same existing hook, just implemented the call here, not a new hook.

src/middlewared/middlewared/plugins/device_/vrrp_events_linux.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def vrrp_hook_license_update(middleware, prev_product_type, *args, **kwargs):
+
+    global VRRP_THREAD
+
+    # get new product_type
+    new_product_type = middleware.call_sync('system.product_type')
+
+    if prev_product_type != 'SCALE_ENTERPRISE' and new_product_type = 'SCALE_ENTERPRISE':
```
william-gr, line 18: ==

william-gr (changes requested):

william-gr (approved):

PR #5238, opened 2020-07-21
NAS-106871 / 12.0 / Fix migrations state

themylogin:


william-gr:
@themylogin I just need to understand why the migration is for rsync but the change to cifs model

themylogin:
@william-gr there were two separate issues:
* rsync model had enabled field NULL in the database (while it was NOT NULL in model like it should be)
* cifs model had secrets field NOT NULL in the model (while it was NULL in the database like it should be; there are `is None` checks against it)

PR #5241, opened 2020-07-21
NAS-106878 / 12.1 / exclude nvme boot drives from fenced on freeBSD

yocalebo:


src/fenced/fenced/fence.py
```diff
@@ -55,10 +56,27 @@ def load_disks(self):
         unsupported = []
         remote_keys = set()
 
+        disks = sysctl.filter('kern.disks')[0].value.split()
+
+        # Running fenced exluding all disks is not allowed
+        try:
+            if not len(set(self._exclude_disks) - set(disks)):
+                raise PanicExit('Excluding all disks is not allowed')
+        except PanicExit:
```
william-gr, line 65: Why have it in a try/except if we are catching PanicExit to just raise it again?

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
```diff
@@ -390,10 +391,10 @@ def carp_master(self, fobj, ifname, vhid, event, user_override, forcetakeover):
                 self.logger.warn('Acquired failover master lock')
                 self.logger.warn('Starting fenced')
                 if not user_override and not fasttrack and not forcetakeover:
-                    error, output = run('LD_LIBRARY_PATH=/usr/local/lib /usr/local/bin/fenced')
+                    error, output = run(f'LD_LIBRARY_PATH=/usr/local/lib /usr/local/bin/fenced -ed {fobj["boot_disks"]}')
                 else:
                     error, output = run(
-                        'LD_LIBRARY_PATH=/usr/local/lib /usr/local/bin/fenced --force'
+                        f'LD_LIBRARY_PATH=/usr/local/lib /usr/local/bin/fenced --force --ed {fobj["boot_disks"]}'
```
william-gr, line 397: -ed or --ed? does it matter?

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
```diff
@@ -390,10 +391,10 @@ def carp_master(self, fobj, ifname, vhid, event, user_override, forcetakeover):
                 self.logger.warn('Acquired failover master lock')
                 self.logger.warn('Starting fenced')
                 if not user_override and not fasttrack and not forcetakeover:
-                    error, output = run('LD_LIBRARY_PATH=/usr/local/lib /usr/local/bin/fenced')
+                    error, output = run(f'LD_LIBRARY_PATH=/usr/local/lib /usr/local/bin/fenced -ed {fobj["boot_disks"]}')
```
william-gr, line 394: I have doubts this would work given this:

>>> f'{a}'
"['asas', 'asdasd']"
>>> f'-ed {a}'
"-ed ['asas', 'asdasd']"


Should we use {" ".join(fobj["boot_disks"])} ?

yocalebo: argsparse does handle a list object but I'll change to join method.

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
yocalebo: it definitely matters, so a typo on my part. fixed.

src/fenced/fenced/fence.py
yocalebo: don't like panic'ing in this scenario so I've added another exception.

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
william-gr: What if boot_disks is empty, could that be a possibility in case some parse fails? And if so, would that cause a call error with an empty -ed arg?

yocalebo: Yes, it would. argparse exits early if -ed is given with an empty value.

william-gr: Shouldnt we do something about it, then?

PR #5247, opened 2020-07-22
NAS-102423 / 12.1 / NFS share aliases

themylogin:


src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -370,6 +375,12 @@ async def do_delete(self, id):
 
     @private
     async def validate(self, data, schema_name, verrors, old=None):
+        if len(data["aliases"]) and len(data["aliases"]) != len(data["paths"]):
```
william-gr, line 378: Should we make sure its set only for SCALE?

PR #5256, opened 2020-07-22
NAS-106892 / 12.1 / fix FCHBANotPresent alert on SCALE

yocalebo:


src/freenas/usr/local/lib/middlewared_truenas/alert/source/fc_hba_not_present.py
```diff
@@ -24,6 +25,10 @@ class FCHBANotPresentAlertSource(ThreadedAlertSource):
     products = ("ENTERPRISE",)
 
     def check_sync(self):
+
+        if osc.IS_LINUX:
+            return []
```
william-gr, line 30: Wouldnt the products = ("ENTERPRISE",) above already make it only run on BSD?

yocalebo:
Hmm, actually yes it should looking at alert plugin source. My local install of SCALE doesn't have the product_type changes that I merged in this morning. I need to go back and update the LicenseStatus alert now. I'll fix.

PR #5265, opened 2020-07-23
Properly log if we are unable to create swap mirror

sonicaj:


src/middlewared/middlewared/plugins/disk_/swap_configure.py
```diff
@@ -139,8 +139,8 @@ async def swaps_configure(self):
                             'extra': {'level': 1} if osc.IS_LINUX else {},
                         }
                     )
-                except CallError:
-                    self.logger.warning('Failed to create swap mirror %s', swap_path)
+                except CallError as e:
+                    self.logger.warning('Failed to create swap mirror %s: %s', swap_path, str(e))
```
william-gr, line 143: %s already does str(e), JFYI, no need to change

PR #5267, opened 2020-07-23
NAS-106887 / 12.1 / sysctl must run before ix-netif on HA

yocalebo:


yocalebo:
@william-gr I know we've already talked about this in DM but I'm pushing my changes here before further discussion tomorrow.

src/freenas/etc/ix.rc.d/ix-netif
```diff
@@ -219,6 +219,20 @@ _interface_config()
...
+
+	# enable CARP preemption
+	/sbin/sysctl net.inet.carp.preempt=1 > /dev/null 2>&1
+
+	# setting ifdown and senderr sysctl values to 0 prevents CARP
+	# from automatically changing the advskew and therefore the
+	# priority of the interface. TrueNAS's HA code handles this
+	# logic.
+	/sbin/sysctl net.inet.carp.ifdown_demotion_factor=0 > /dev/null 2>&1
+	/sbin/sysctl net.inet.carp.senderr_demotion_factor=0 > /dev/null 2>&1
```
william-gr, line 234: So the problem I have with this, is that ix-netif is the last place I would look for default sysctl values.

I would be happier if we kept sysctl.conf

yocalebo: fair enough

PR #5271, opened 2020-07-24
NAS-106874 / 12.1 / WebDAV on SCALE

themylogin:


src/middlewared/middlewared/etc_files/local/apache24/Includes/webdav.conf
```diff
@@ -25,7 +25,7 @@
 		uid = 0
 		gid = 0
 
-	subprocess.run(['chown', '-Rx', f'{uid}:{gid}', oscmd], check=False)
+	subprocess.run(['chown', '-R', f'{uid}:{gid}', oscmd], check=False)
```
william-gr, line 28: Why are we removing x?

themylogin: It meant `stay on one filesystem`, there is no corresponding option in Linux (but there also is no sense in it, we don't mount any filesystems in /etc/apache2)

PR #5272, opened 2020-07-24
NAS-106851 / 12.1 / Provide CPU temperature in FreeBSD-compatible format

themylogin:


themylogin:
@william-gr FreeBSD uses Kelvins and multiplies by 10 to store float in integer (37,1 -> 371)

PR #5277, opened 2020-07-24
NAS-106923 / 12.1 / fix sync'ing of encryption keys (if any) on 12 HA

yocalebo:


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1623,8 +1623,31 @@ async def service_remote(middleware, service, verb, options):
...
+                    'failover.update_encryption_keys', {
+                        'pools': [
+                            {'name': p['name'], 'passphrase': passphrase or ''}
+                            for p in encrypted_pools
+                        ],
+                        'sync_keys': False,
+                    }
+                )
+        else:
+            await middleware.call('failover.call_remote', 'failover.sync_keys_to_remote_node')
```
william-gr, line 1650: If I am seeing correctly there is quite a big block of code identical here and in the event that could be "deduplicated"?

yocalebo: yes, completely so. I'll fix.

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1108,32 +1108,47 @@ def upgrade_finish(self, job):
...
-                        'alert.oneshot_create', 'FailoverKMIPKeysSyncFailed', {'error': str(e)}
-                    )
-                    self.middleware.logger.error(
-                        'Failed to sync KMIP keys with remote node: %s', str(e), exc_info=True
-                    )
-                else:
-                    await self.middleware.call('alert.oneshot_delete', 'FailoverKMIPKeysSyncFailed', None)
+                    assert (await self.middleware.call('failover.call_remote', 'core.ping')) == 'pong'
+                except Exception:
+                    self.middleware.logger.error('Failed to contact standby node.', exc_info=True)
```
sonicaj, line 1124: We should log that this happened during syncing keys just for reference if we want to know why keys did not sync

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1623,8 +1638,32 @@ async def service_remote(middleware, service, verb, options):
...
+
+        # when middlewared is setup, we only run this if we're the standby controller
+        if not await middleware.call('failover.is_backup_node'):
+            return
+
+        try:
+            await middleware.call('failover.call_remote', 'failover.sync_keys_to_remote_node')
+        except Exception as e:
+            if e.errno == CallError.ENOMETHOD:
+                # We're talking to an <= 11.3 system, so use the old API
```
sonicaj, line 1652: I am not sure if we need this in 2 places ? When backup node comes back online it should perform this sync. Do you think it's safe to remove the code in event.py or from here ?

yocalebo: Spoke with William about that and to be safe, we need to keep it in both places. The event plugin is really the only way of knowing when we're the BACKUP controller.

We also need to sync keys on "setup" of middlewared so we'll leave it there too.

sonicaj: In that case let's please have a common method for this which we can call in both places.

yocalebo: A prime example of when the event plugin isn't used but the setup is used is if someone restarts middlewared on the backup controller.

sonicaj: It might be nice to have a wrapper around `sync_keys_to_remote_node` which we could use in these two places and then remove down the road

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1107,33 +1107,100 @@ def upgrade_finish(self, job):
...
+        TODO:
+            Once TruenAS >= 11.3 < 12 has been EOL'd, we should remove this method
+            (should call `failover.call_remote` `failover.sync_keys_to_remote_node`)
+        """
+
+        if not await self.middleware.call('failover.licensed'):
+            return
+
+        # only sync keys if we're the BACKUP node
+        if (await self.middleware.call('failover.is_backup_node')) != 'BACKUP':
```
william-gr, line 1123: Does this return a boolean?

yocalebo: uhh yeah, typo on my part.

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1107,33 +1107,100 @@ def upgrade_finish(self, job):
...
+        if not await self.middleware.call('failover.licensed'):
+            return
+
+        # only sync keys if we're the BACKUP node
+        if (await self.middleware.call('failover.status')) != 'BACKUP':
+            return
+
+        # make sure we can contact the MASTER node
+        try:
+            assert (await self.middleware.call('failover.call_remote', 'core.ping')) == 'pong'
```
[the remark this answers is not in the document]
yocalebo: Yeah, I like that idea too. I'll open another improvement ticket and put that in later release.

PR #5279, opened 2020-07-25
smb4.conf lines beginning #; are not errors, they're ignorable comments and should be silently ignored

stilez:
Don't treat smb4.conf comments as errors. Just silently ignore them


anodos325:
Actually, I'm considering increasing how strict we are about what gets put into SMB auxiliary parameters and simply raising validation errors on things that aren't of the form "foo = bar".

stilez:
Never underestimate the value of being able to.comment ones config in situ.  Config comments may be needed for many reasons, not least to mark workarounds, future fix needs, and unobvious causes for config lines. 

As comments are harmless and obvious, may I request that there's no reason to treat them as unacceptable.

william-gr:
@stilez I imagine this can be closed in favor of the related PR?

stilez:
> @stilez I imagine this can be closed in favor of the related PR?

@william-gr Yes

PR #5282, opened 2020-07-27
NAS-106953 / 12.1 / Improve validation for SMB service and shares

anodos325:
Starting in TrueNAS 12.0 smb configuration relies more heavily on libsmbconf
and registry.tdb. Auxiliary parameters are no longer merely placed directly
into a text file, but must be parsed and applied through libsmbconf.
Additionally, new shares may be generated through vfs_zfs_fsrvp, and
included into the global SMB configuration. This hybrid configuration
necessitates more robust validation and normalization of user-provided
auxiliary parameters. As such, auxiliary parameters are not treated as simple
key-value stores of the form "key = value".

Changes in this regard are as follows:
1) Unnecessary new lines are stripped
2) Validation errors are raised on comments
3) Validation errors are raised on entries that are not of the form "key = value"

Validation for SMB service configuration has been moved into a new separate
function. Checks added for collisions between NetBIOS aliases and workgroups.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -442,8 +442,11 @@ class SMBService(SystemServiceService):
         into the registry.
         """
         job.set_progress(60, 'generating SMB share configuration.')
+        await self.middleware.call('cache.put', 'SMB_REG_INITIALIZED', False)
         await self.middleware.call("etc.generate", "smb_share")
         await self.middleware.call("smb.import_conf_to_registry")
+        await self.middleware.call('cache.put', 'SMB_REG_INITIALIZED', True)
+        os.unlink(SMBPath.SHARECONF.platform())
 
         """
         It is possible that system dataset was migrated or an upgrade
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -902,6 +946,37 @@ async def close_share(self, share_name):
...
+                continue
+            kv = entry.split('=', 1)
+            if len(kv) != 2:
+                verrors.add(
+                    f'{schema_name}.auxsmbconf',
+                    f'Auxiliary parameters must be in the format of "key = value": {entry}'
+                )
+                continue
+
+            parsed_auxsmbconf += f'\n{entry}' if len(parsed_auxsmbconf) != 0 else entry
```
william-gr, line 975: Isnt this piece of code pretty much the same as the one in validate_smb ?

Perhaps we should "deduplicate" it?

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -902,6 +925,40 @@ async def close_share(self, share_name):
...
+        verrors = ValidationErrors()
+        parsed_auxsmbconf = ""
+        for entry in data.splitlines():
+            if entry == "":
+                # silently remove empty lines
+                continue
+            if entry.startswith('#'):
+                verrors.add(
+                    f'{schema_name}',
+                    f'Comments are not permitted in SMB auxiliary parameters: {entry}'
```
themylogin, line 945: I am sorry if I am being ignorant, but Is there any reason for that?

themylogin: I see we have a related PR https://github.com/freenas/freenas/pull/5279

Even if you decide to forbid comments for some reason, then use

`if entry.startswith(('#', ';')):`

william-gr: I am also starting to lean towards allowing comments...

If its a single line aux parameter I may want to note why its there?

anodos325: Configuration is written into registry.tdb (a key-value store), which doesn't allow insertion of comments (as opposed to legacy file-based configuration. If we prefer to allow this I can let it through the validation and strip them before sending to the registry.

william-gr: To be fair we ever allowed this, and that was our ultimate mistake.

I think we should either allow comments (doesnt seem like such a terrible idea TBH) or add a migration removing them.

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -715,6 +750,7 @@ async def do_create(self, data):
             'datastore.insert', self._config.datastore, data,
             {'prefix': self._config.datastore_prefix})
 
+        await self.strip_comments(data)
```
william-gr, line 753: We would now be returning data with comments stripped?

Should we change return to use get_instance and remove .extend call?

anodos325: Yeah, that's a good idea.

src/middlewared/middlewared/plugins/smb.py
stilez: Config workarounds, future fixes, and unobvious params, are often enough needed. Removing the ability to make config self-documenting where possible seems a poor option.  @anodos325 seems to neatly solve the problem, by allowing comments in the GUI config text, but stripping them out before actual use in the registry. That would work and solves the issue that PR #5279 attempts to fix.

PR #5288, opened 2020-07-28
Fix regression in aux param merge in SMB share creation

anodos325:


anodos325:
Validation method should catch any issues before we merge presets with user-provided config options. Remove try-except so that we fail more vocally if validation fails to catch an issue.

PR #5295, opened 2020-07-28
NAS-106960 / 12.0 / Do not restart ctld on BACKUP -> MASTER

themylogin:


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
```diff
@@ -531,7 +531,9 @@ def carp_master(self, fobj, ifname, vhid, event, user_override, forcetakeover):
                 c.execute('SELECT srv_enable FROM services_services WHERE srv_service = "iscsitarget"')
                 ret = c.fetchone()
                 if ret and ret[0] == 1:
-                    self.run_call('service.start', 'iscsitarget')
+                    # Do not restart iscsitarget as it is not necessary (but breaks FC connections)
+                    if not self.middleware.call_sync('service.started', 'iscsitarget'):
+                        self.run_call('service.start', 'iscsitarget')
```
william-gr, line 536: Why starting iscsi service makes it restart?

amotin: That is what I asked too.  But it does.

themylogin: @william-gr we used `restart` RC command to start the vast majority of our services in 11.3, that's why `_freebsd_start` uses `restart` instead of `start` so as not to break anything.

PR #5308, opened 2020-07-29
NAS-106999 / 12.1 / Human-readable error for deleting used cloud credential

themylogin:


src/middlewared/middlewared/plugins/cloud_sync.py
```diff
@@ -611,6 +611,9 @@ async def do_delete(self, id):
         """
         Delete Cloud Sync Credentials of `id`.
         """
+        if await self.middleware.call("cloudsync.query", [["credential", "=", id]]):
+            raise CallError("This credential is used")
```
william-gr, line 615: Should we use errno.EBUSY?

@Mrt134 any thoughts in the error message?

themylogin: @Mrt134 ping

yocalebo: Is there any way to specify which cloud task the credential is being used by?

yocalebo: That way you could say something like `This credential is in use by: blah`. If not, then the verbiage you have is fine.

PR #5316, opened 2020-07-30
NAS-106946 / 12.1 / Multiple idmap generation and validation fixes

anodos325:
During course of writing regression tests for idmap plugin, I discovered multiple issues.

- Make smb.conf generation more forgiving about default / missing keys in backend['options'].
- Wait for idmap cache clear to complete before returning idmap updates
- Add new methods to write idmap secrets to secrets.tdb. "net idmap set secret" performs an smb.conf check, which we want 
   to avoid during idmap setup.
- Force a directory services secrets backup after idmap change if it resulted in new secrets being written to secrets.tdb.
- Fix / improve validation for some edge-cases.

See NAS-107011 for regression tests.

### DIFF
--- a/src/middlewared/middlewared/plugins/idmap.py
+++ b/src/middlewared/middlewared/plugins/idmap.py
@@ -42,6 +42,7 @@ class IdmapBackend(enum.Enum):
             'unix_primary_group': {"required": False, "default": False},
             'unix_nss_info': {"required": False, "default": False},
         },
+        'has_secrets': False,
         'services': ['AD'],
     }
     AUTORID = {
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -646,9 +687,23 @@ async def do_update(self, id, data):
...
         new['options'] = final_options
+
+        if new['options'].get('ldap_user_dn_password'):
+            try:
+                DSType[new["name"]]
+                domain = (await self.middleware.call("smb.config"))['workgroup']
+            except KeyError:
+                domain = new["name"]
+
+            secret = data['options'].pop('ldap_user_dn_password')
```
william-gr, line 701: Do we really want this to pop from `data`
And we are redefining secret on next line which I imagine is the only real thing we need

PR #5326, opened 2020-07-31
NAS-107019 / 11.3 / Fix issue with AD machine account keytab storage on HA platform

anodos325:
netbiosname key in activedirectory config is HA configuration
aware and what is used for domain join. Use that rather than
the SMB netbiosname to ensure that we remove the correct keys
when creating the AD machine account keytab in our config file.

Regression test coverage provided in NAS-106978


anodos325:
Issue does not affect 12.0 +

PR #5327, opened 2020-07-31
Make failover faster by not doing failover.status_refresh when it's not necessary

themylogin:


themylogin:
@yocalebo 35-40 seconds handling these two subsequent events
```
[2020/07/31 06:13:36] (WARNING) FailoverService._event():155 - Ignoring event:INIT on internal interface ntb0
[2020/07/31 06:13:56] (WARNING) FailoverService._event():155 - Ignoring event:INIT on internal interface ntb0
```
(I was rebooting the other node to initiate failover)

@amotin reported even worse result of two minutes

yocalebo:
@themylogin interesting. Thanks. I like this change and makes sense.

william-gr:
backport

PR #5337, opened 2020-08-01
NAS-106994 / 12.0 / Allow user to explicitly set ipv4 for openvpn server/client protocol

sonicaj:


sonicaj:
backport

src/middlewared/middlewared/plugins/vpn.py
```diff
@@ -10,6 +10,9 @@
 from middlewared.validators import Port, Range
 
 
+PROTOCOLS = ['UDP4', 'UDP6', 'TCP4', 'TCP6']
```
william-gr, line 13: Why not still have UDP/TCP?

sonicaj: I am not sure, if user selects `UDP6`, he/she should have dual stack behaviour with ipv4 / ipv6 functioning ( https://wiki.archlinux.org/index.php/OpenVPN#Connect_to_the_server_via_IPv6 ).
However I also found another link where it states that if user specifies just `udp`, it will default to using `getaddrinfo` to see what protocol / network server supports ( https://serverfault.com/questions/651832/openvpn-with-mixed-ipv4-and-ipv6-clients ).

I think we can add it for completeness. Updating the PR.

PR #5340, opened 2020-08-03
NAS-107022 / 12.1 / save failover logs in /root/syslog like 11.3

yocalebo:


src/middlewared/middlewared/logger.py
```diff
@@ -278,6 +294,12 @@ def _set_output_file(self):
             os.chmod(ZETTAREPL_LOGFILE, 0o640)
         except OSError:
             pass
+        try:
+            dirname = os.dirname(FAILOVER_LOGFILE)
+            os.makedirs(dirname, exist_ok=True)
+            os.chmod(FAILOVER_LOGFILE, 0o640)
```
themylogin, line 300: Why, isn't everything under `/root` already protected?

yocalebo: It should be, I've removed the chmod call.

PR #5348, opened 2020-08-03
NAS-106974 / 12.0 / Fix avahi configuration bugs on HA platform

anodos325:
Do not run mdns on passive storage controller. Remove internal interfaces from avahi configuration.

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
@@ -577,6 +577,7 @@ class FailoverService(Service):
 
                 self.run_call('service.restart', 'collectd', {'ha_propagate': False})
                 self.run_call('service.restart', 'syslogd', {'ha_propagate': False})
+                self.run_call('service.restart', 'mdns', {'ha_propagate': False})
 
                 for i in (
                     'smartd', 'ftp', 'lldp', 'rsync', 's3', 'snmp', 'ssh', 'tftp', 'webdav',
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/service_/services/mdns.py
```diff
@@ -13,15 +13,17 @@ class MDNSService(SimpleService):
     systemd_unit = "avahi-daemon"
 
     async def start(self):
+        failover_status = await self.middleware.call("failover.status")
         announce = (await self.middleware.call("network.configuration.config"))["service_announcement"]
-        if not announce["mdns"]:
+        if not announce["mdns"] or failover_status not in ['SINGLE', 'MASTER']:
```
yocalebo, line 18: I'd rather stop the avahi service on a BACKUP event since that's the only sure way of knowing what state we're in.

anodos325: Okay. Sounds fair.

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
```diff
@@ -525,6 +525,7 @@ def carp_master(self, fobj, ifname, vhid, event, user_override, forcetakeover):
                 self.run_call('etc.generate', 'ssl')
                 # Now we restart the appropriate services to ensure it's using correct certs.
                 self.run_call('service.restart', 'http')
+                self.run_call('service.start', 'mdns', {'ha_propagate': False})
```
yocalebo, line 528: I know it's impossible to see by looking at the code, but can we move this call below the "critical" services. Right above the `for i in (smartd, ...)` line would make sense.

PR #5355, opened 2020-08-04
NAS-107069 / 11.3 / Symlink /usr/share/skel to /etc/skel in FreeBSD

anodos325:
pam_mkhomedir appears to default to trying to get files from "/etc/skel".
This possibly indicates a regression in the port, but since it can
be hard to predict behavior of applications that use pam libraries
and hardcode skel path to Linux default location, make this path a
symlink to the FreeBSD path.


william-gr:
backport

PR #5362, opened 2020-08-04
NAS-107074 / 12.1 / Permissions are incorrect on home directory move

anodos325:
In case of moving location of home directory from path a to path b, if path b
does not exist, we need to explicitly set permissions as well strip any ACL
on the path in order to ensure that premissions from original path are preserved.

Preference for new mode is as follows:
1. user-supplied mode in 'user.update' call
2. existing mode on old path
3. default mode. Since this is most likely an error condition, default is 0700.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -18,6 +18,7 @@ import random
 import shlex
 import shutil
 import string
+import stat
 import subprocess
 import time
 
@@ -326,7 +327,6 @@ class UserService(CRUDService):
                         'path': dest_file,
                         'uid': data['uid'],
                         'gid': group['gid'],
-                        'options': {'recursive': True}
                     })
 
             data['sshpubkey'] = sshpubkey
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/account.py
```diff
@@ -415,11 +419,18 @@ async def do_update(self, pk, data):
         if home_copy and not os.path.isdir(user['home']):
             try:
                 os.makedirs(user['home'])
-                await self.middleware.call('filesystem.chown', {
+                mode_to_set = user.get('home_mode')
+                if not mode_to_set:
+                    mode_to_set = oct(old_mode)[-3:] if old_mode else '700'
```
themylogin, line 424: The correct way is
```
>>> '%03o' % 493
'755'
```

william-gr: Not really, % is old formatting style we are avoiding

william-gr:
```
>>> f'{493:03o}'
'755'
```

PR #5373, opened 2020-08-04
NAS-107035 / 11.3 / Properly calculate swap partition size on 4k sector disks

themylogin:
@william-gr I'll backport from 12.0 to 12.1

### DIFF
--- a/src/middlewared/middlewared/plugins/disk.py
+++ b/src/middlewared/middlewared/plugins/disk.py
@@ -1798,8 +1798,10 @@ class DiskService(CRUDService):
             # The GPT header takes about 34KB + alignment, round it to 100
             if size - 100 <= swapgb * 1024 * 1024:
                 raise CallError(f'Your disk size must be higher than {swapgb}GB')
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/disk.py
```diff
@@ -1798,16 +1798,18 @@ def format(self, disk, swapgb, sync=True):
             # The GPT header takes about 34KB + alignment, round it to 100
             if size - 100 <= swapgb * 1024 * 1024:
                 raise CallError(f'Your disk size must be higher than {swapgb}GB')
+            sectorsize = g.provider.sectorsize
```
william-gr, line 1801: Is there any chance this could be null or empty? Would it be safer to put `or 512`?

Also is this not needed for 12.1?

william-gr: @themylogin ping

themylogin: @william-gr I'll backport from 12.0 to 12.1

PR #5383, opened 2020-08-05
NAS-107087 / 12.1 / add TCP keepalive settings for remote client

yocalebo:


src/middlewared/middlewared/client/client.py
```diff
@@ -27,6 +28,25 @@
     LIBZFS = True
 
 
+class TCPKeepAlive(enum.Enum):
+
+    KEEPIDLE = (256, 4)
+    KEEPINTVL = (512, 5)
+    KEEPCNT = (1024, 6)
```
william-gr, line 35: why not use socket.TCP_KEEP*?

yocalebo: It's not exported in the socket module from what I've gathered.

william-gr: Maybe not in FreeBSD, but its there for linux at least

yocalebo: yep, it is in freeBSD too....i must have typo'ed earlier. I'll fix.

william-gr:
@amotin how does that look to you?

amotin:
> @amotin how does that look to you?

It is what I am using in CTL for HA interconnect. Except I am also setting TCP_KEEPINIT to 3.  Works great in my case.

PR #5387, opened 2020-08-06
NAS-107099 / 12.1 / Do not display previous replication task status after deleting it and…

themylogin:
… creating new replication task.

This happens because SQLite IDs can duplicate deleted rows.

### DIFF
--- a/src/middlewared/middlewared/plugins/zettarepl_/state.py
+++ b/src/middlewared/middlewared/plugins/zettarepl_/state.py
@@ -83,11 +83,26 @@ class ZettareplService(Service):
         for task_id in set(old_definition_errors.keys()) | set(self.definition_errors.keys()):
             self._notify_state_change(task_id)
 
-    def set_hold_tasks(self, hold_tasks):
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/zettarepl_/state.py
```diff
@@ -83,12 +83,28 @@ def set_definition_errors(self, definition_errors):
...
         self.hold_tasks = hold_tasks
         for task_id in set(old_hold_tasks.keys()) | set(self.hold_tasks.keys()):
             self._notify_state_change(task_id)
 
+        task_ids = (
+            {f"periodic_snapshot_{k}" for k in definition["periodic-snapshot-tasks"]} |
+            {f"replication_{k}" for k in definition["replication-tasks"]} |
+            set(hold_tasks.keys())
+        )
+        self.middleware.logger.warning("TASK_IDS %r", task_ids)
```
william-gr, line 97: Is this a leftover debug or do we really need this?

PR #5396, opened 2020-08-06
NAS-107107 / 12.0 / Clear any potential stale state after leaving AD domain

anodos325:
re-initialize secrets.tdb. Rename old one to a backup file. Odds of
repeatedly joining / leaving AD are extremely low and so we don't have
to worry about a couple of these accumulating.

Clear out the backup of the domain secrets in our config file.

Flush samba's cache after successfully leaving the domain.


anodos325:
This is primarily a concern during our automated tests where we join / leave AD domain repeatedly.

PR #5398, opened 2020-08-07
NAS-107115 / 12.0 / Newly created builtin users should not default to 'smb'

anodos325:
Also add migration to turn off SMB for builtin accounts.


anodos325:
ntpd ended up with smb flag set, which generated some log spam. This should fix the spam issue and prevent it from happening again.

PR #5402, opened 2020-08-07
NAS-107123 / 12.1 / Add catia mappings for special Apple characters

anodos325:


anodos325:
I was originally planning to add these as a parameter for vfs_catia "catia:mappings = fruit" or something like that, but providing correct behavior is more important as we approach beta2 / rc1.

anodos325:
Copy-paste of code reuse of 0xf00d. Code is now more hygienic.

PR #5409, opened 2020-08-10
NAS-107141 / 12.1 / remove excess logging info when syncing disks

yocalebo:
On large systems, the `Found disks` logging line can cause 100's of lines of information to be printed to `middlewared.log`.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/sync.py
+++ b/src/middlewared/middlewared/plugins/disk_/sync.py
@@ -91,7 +91,7 @@ class DiskService(Service, ServiceChangeMixin):
                 ik: iv for ik, iv in ov.items() if ik in ('name', 'ident', 'lunid', 'serial')
             } for ok, ov in sys_disks.items()
         }
-        self.logger.info(f'Found disks: {log_info}')
+        self.logger.info('Found disks: %r', log_info)
 
         seen_disks = {}
         serials = []
--- a/src/middlewared/middlewared/plugins/disk_/sync.py
+++ b/src/middlewared/middlewared/plugins/disk_/sync.py
@@ -68,12 +68,9 @@ class DiskService(Service, ServiceChangeMixin):
         Synchronize all disks with the cache in database.
         """
         # Skip sync disks on standby node
-        if (
-            not await self.middleware.call('system.is_freenas') and
-            await self.middleware.call('failover.licensed') and
-            await self.middleware.call('failover.status') == 'BACKUP'
-        ):
-            return
+        if await self.middleware.call('failover.licensed'):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/disk_/sync.py
```diff
@@ -86,7 +83,6 @@ async def sync_all(self, job):
                 self.logger.warning('Starting disk.sync_all when devd is not connected yet')
 
         sys_disks = await self.middleware.call('device.get_disks')
-        self.logger.info('Found disks: %r', sys_disks)
```
william-gr, line 89: @themylogin  do you remember why we put it there? Should we log at least the names?

themylogin: Yes, we have issues when syncing disks not resolved yet, this actually helped me once. @yocalebo can we please leave this line and log only `name`, `ident`, `serial` and `lunid`?

yocalebo: Yeah I can do that. But `ident` and `serial` are the same right?

src/middlewared/middlewared/plugins/disk_/sync.py
```diff
@@ -86,7 +83,15 @@ async def sync_all(self, job):
...
+
+        # output logging information to middlewared.log in case we sync disks
+        # when not all the disks have been resolved
+        log_info = {
+            ok: {
+                ik: iv for ik, iv in ov.items() if ik in ('name', 'ident', 'lunid', 'serial')
+            } for ok, ov in sys_disks.items()
+        }
+        self.logger.info(f'Found disks: {log_info}')
```
themylogin, line 94: Let's stick to old-style formatting for logging please

src/middlewared/middlewared/plugins/disk_/sync.py
```diff
@@ -86,6 +83,14 @@ async def sync_all(self, job):
...
         sys_disks = await self.middleware.call('device.get_disks')
+
+        # output logging information to middlewared.log in case we sync disks
+        # when not all the disks have been resolved
+        log_info = {
+            ok: {
+                ik: iv for ik, iv in ov.items() if ik in ('name', 'ident', 'lunid', 'serial')
+            } for ok, ov in sys_disks.items()
+        }
         self.logger.info('Found disks: %r', sys_disks)
```
themylogin, line 94: Now yo're logging old variable

yocalebo: Typo, fixed.

PR #5418, opened 2020-08-10
NAS-107135 / 12.0 / Ensure passdb insertions and removals correspond with 'smb' checkbox

anodos325:
"smb" True -> False should result in passdb removal
"smb" False -> True without password should raise validation error
"smb" False -> True with password should result in passdb insertion


anodos325:
Forgot to backport to 12.0.

PR #5431, opened 2020-08-11
NAS-106875 / 12.1 / Add directory services stats to usage plugin

sonicaj:
We have the following stats gathered for directory services:
```
{
  "directory_services": {
    "state": {
      "activedirectory": "DISABLED",
      "ldap": "DISABLED",
      "nis": "DISABLED"
    },
    "kerberos_realm_populated": false,
    "has_samba_schema": false
  }
}
```

### DIFF
--- a/src/middlewared/middlewared/plugins/usage.py
+++ b/src/middlewared/middlewared/plugins/usage.py
@@ -137,6 +137,18 @@ class UsageService(Service):
             'ha_licensed': await self.middleware.call('failover.licensed'),
         }
 
+    async def gather_directory_service_stats(self, context):
+        config = await self.middleware.call('ldap.config')
+        return {
+            'directory_services': {
+                'state': await self.middleware.call('directoryservices.get_state'),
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/usage.py
```diff
@@ -137,6 +137,16 @@ async def gather_ha_stats(self, context):
             'ha_licensed': await self.middleware.call('failover.licensed'),
         }
 
+    async def gather_directory_service_stats(self, context):
+        config = await self.middleware.call('ldap.config')
+        return {
+            'directory_services': {
+                'state': await self.middleware.call('directoryservices.get_state'),
+                'kerberos_realm_populated': bool(config['kerberos_realm']),
+                'has_samba_schema': config['has_samba_schema'],
```
william-gr, line 146: these are specific to ldap so perhaps it should be in its own ldap dict? thoughts @anodos325

PR #5432, opened 2020-08-11
NAS-107152 / 12.1 / Retry gathering usage starts if we experience a failure

sonicaj:


src/middlewared/middlewared/plugins/usage.py
```diff
@@ -13,27 +13,35 @@
...
+                            headers={'Content-type': 'application/json'},
+                            proxy=os.environ.get('http_proxy'),
+                        )
+                except Exception as e:
+                    # We still want to schedule the next call
+                    self.middleware.logger.error(e)
+                    retries -= 1
+                    if retries:
+                        self.middleware.logger.debug('Retrying gathering stats after 5 minutes')
+                        await asyncio.sleep(300)
```
william-gr, line 42: 5 minutes is too low, we talked about an hour. Make it at least 30 minutes please.

src/middlewared/middlewared/plugins/usage.py
```diff
@@ -13,27 +13,35 @@
...
+
+    FAILED_RETRIES = 3
+
     class Config:
         private = True
 
     async def start(self):
         if (
             await self.middleware.call('system.general.config')
         )['usage_collection']:
```
william-gr, line 25: This should be within the loop too.

src/middlewared/middlewared/plugins/usage.py
sonicaj: No problem, we can have it at 30 minutes. It was not mentioned in the ticket at all so defaulted to 5 minutes but yes 30-60 min is indeed more reasonable.

src/middlewared/middlewared/plugins/usage.py
```diff
@@ -13,27 +13,35 @@
...
-                        headers={"Content-type": "application/json"},
-                        proxy=os.environ.get("http_proxy"),
+                        data=await self.middleware.call('usage.gather'),
+                        headers={'Content-type': 'application/json'},
+                        proxy=os.environ.get('http_proxy'),
                     )
             except Exception as e:
                 # We still want to schedule the next call
-                self.logger.error(e)
+                self.middleware.logger.error(e)
```
william-gr, line 38: Why the change of logger?

sonicaj: Consistency ? We normally use `middleware.logger.level`, i can update to the old one if you think that's best

william-gr: Where do we use middleware.logger? Thats just wrong and we shouldnt be doing that for services.

william-gr: @sonicaj ping

PR #5449, opened 2020-08-12
NAS-107160 / 11.3 / Log sharesec delete errors

anodos325:
If for some reason the path is ripped out before the SMB share is deleted
this can cause failure to remove entry from share_info.tdb.
In this situation, log what happened but allow share deletion to continue.


anodos325:
Only affects 11.3-stable.

PR #5453, opened 2020-08-12
NAS-107166 / 11.3 / Move NFS kerberos SPN setup into separate function

anodos325:
This is to provide support a mechanism to add these entries after
the fact, and is a prep step for providing APIs for the webui team
to prompt users to add kerberos SPNs to AD in the NFS form (if AD
is enabled).

Regression test added in NAS-107169

### DIFF
--- a/src/middlewared/middlewared/etc_files/nfsd.py
+++ b/src/middlewared/middlewared/etc_files/nfsd.py
@@ -5,13 +5,13 @@ from middlewared.utils import run
 logger = logging.getLogger(__name__)
 
 
-async def get_exports(config, shares, kerberos_keytabs):
+async def get_exports(config, shares, has_nfs_principal):
     result = []
 
     if config["v4"]:
         if config["v4_krb"]:
             result.append("V4: / -sec=krb5:krb5i:krb5p")
-        elif kerberos_keytabs:
+        elif has_nfs_principal:
             result.append("V4: / -sec=sys:krb5:krb5i:krb5p")
         else:
             result.append("V4: / -sec=sys")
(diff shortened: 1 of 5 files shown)


william-gr:
@anodos325 there is a small PEP8 nit forbidding the merge

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -25,9 +25,8 @@ class Config:
 
     @private
     async def nfs_extend(self, nfs):
-        nfs["v4_krb_enabled"] = (
-            nfs["v4_krb"] or bool(await self.middleware.call("datastore.query", "directoryservice.kerberoskeytab"))
-        )
+        keytab_has_nfs = await self.middleware.call("kerberos.keytab.has_nfs_principal")
+        nfs["v4_krb_enabled"] = ( nfs["v4_krb"] or keytab_has_nfs )
```
william-gr, line 29: flake8 here, `(nfs["v4_krb"] or keytab_has_nfs)`

src/middlewared/middlewared/etc_files/nfsd.py
```diff
@@ -5,13 +5,13 @@
...
 
-async def get_exports(config, shares, kerberos_keytabs):
+async def get_exports(config, shares, has_nfs_principal):
     result = []
 
     if config["v4"]:
         if config["v4_krb"]:
             result.append("V4: / -sec=krb5:krb5i:krb5p")
-        elif kerberos_keytabs:
+        elif has_nfs_principal:
```
yocalebo, line 14: So, in theory, if the user doesn't have kerberos authentication enabled but they have NFS SPN entries, then we generate /etc/exports so. We're still starting gssd with either situation right? Because if the user
1. doesn't enable kerberos auth
2. but has an NFS SPN entry
Do we start gssd?

anodos325: That's a good question. Underlying logic isn't substantively changed except to be aware that not all keytabs are for NFS.

Prior to this PR though an AD user with v4 enabled _always_ had "V4: / -sec=sys:krb5:krb5i:krb5p" even when no krb5 was intended for shares.

PR #5454, opened 2020-08-12
NAS-107108 / 11.3 / Google Drive alternate_export option

themylogin:


src/middlewared/middlewared/rclone/remote/google_drive.py
```diff
@@ -19,6 +19,10 @@ class GoogleDriveRcloneRemote(BaseRcloneRemote):
     credentials_oauth = True
     refresh_credentials = ["token"]
 
+    task_schema = [
+        Bool("alternate_export", title="Use an alternate set of export URLs for drive documents", default=False),
```
william-gr, line 23: I would like to understand what that does in practice, why there is no workaround and why it shouldnt be always enabled

themylogin: https://rclone.org/drive/#drive-alternate-export
```
--drive-alternate-export
Use alternate export URLs for google documents export.,

If this option is set this instructs rclone to use an alternate set of export URLs for drive documents. Users have reported that the official export URLs can't export large documents, whereas these unofficial ones can.
```

I don't want it always enabled because unofficial API can change at any moment.

william-gr: @themylogin  can you please do a bit more research other than only reading the rclone manpage?

We dont want obscure options exposed in the UI for no reason. That option was added in May 2018.

There is some useful comments in https://issuetracker.google.com/issues/36761333 from 2019.

Perhaps that means alternate can always be used or we can properly implement it in rclone now?

themylogin: @william-gr you are right https://github.com/rclone/rclone/pull/4503

william-gr:
@themylogin Great, thanks for investigating further. I suppose this can be closed then.

PR #5478, opened 2020-08-14
NAS-100008 / 12.1 / Better unhealthy pool notification

themylogin:


src/middlewared/middlewared/plugins/disk_/zfs_guid.py
```diff
@@ -0,0 +1,55 @@
...
+                    await self.middleware.call(
+                        "datastore.update", "storage.disk", disk["identifier"], {"zfs_guid": None}, {"prefix": "disk_"},
+                    )
+
+
+async def hook(middleware, pool):
+    await middleware.call("disk.sync_zfs_guid", pool)
+
+
+async def setup(middleware):
```
william-gr, line 53: We should also hook to zfs events, i think at least config_sync ?

william-gr:
@themylogin do you think its safe to merge this to 12.0?

themylogin:
@william-gr hooks are async so if they fail we'll only suffer a traceback in the logs and nothing more. If alert fails (the code is small, I doubt it can fail), we'll just get a traceback instead of one alert. Seems safe to me.

william-gr:
@themylogin Ok, then I think we should merge this for 12.0, that way we can more easily work on mapping pool devs to disks and have Pool Status page show the disk.

Would you agree?

PR #5508, opened 2020-08-18
NAS-107256 / 12.1 / Be prepared to receive tons of zfs deadman messages

themylogin:


yocalebo:
Do we know if a scrub causes a ZfsDeadman alert for every vdev in the zpool? Lots of our customers have 100+ vdevs so we could still have potentially `n` number of alerts in 60 seconds `n` being the count of vdevs.

themylogin:
That'll be 100 alerts which is bad anyway, I am not sure what should we do, throttle them by pool instead? (i.e. drop random events if they occur too often)

yocalebo:
@themylogin tough problem to solve. The reason why I ask if scrubs can cause ZfsDeadman events then I'm wondering if we simply ignore them if one is triggered on a zpool while a scrub is running.

As well as keep your timed approach.

william-gr:
@themylogin @yocalebo how about we limit by number of events per pool in a given time?

Up to say 5 events in the last 5 minutes?

In theory these deadman events should not be happening at this pace, so getting that number of alerts (5 in 5 min) is enough to tell the user some odd is happening.
We are only seeing that many events because ZFS is misbehaving?

yocalebo:
@william-gr do we just choose the first 5 vdevs that are generated?

william-gr:
Thats the idea, within 5 minutes...

Even on a large system if we get that many events from deadman there is probably something else off, right? And not specific to a disk.

yocalebo:
@william-gr I'd have to agree.

PR #5526, opened 2020-08-21
NAS-107351 / 12.1 / datasets *or zvols* in several places, and a little comment on overlapping tasks

stilez:


src/middlewared/middlewared/plugins/snapshot.py
```diff
@@ -95,11 +95,13 @@ async def do_create(self, data):
         """
         Create a Periodic Snapshot Task
 
-        Create a Periodic Snapshot Task that will take snapshots of specified `dataset` at specified `schedule`.
+        Create a Periodic Snapshot Task that will take snapshots of specified `dataset` or `zvol` at specified `schedule`.
```
william-gr, line 98: This is a parameter name, should remain unchanged.

For ZFS perspective everything is a dataset, what we call dataset is a dataset of FILESYSTEM type, zvol is a dataset of VOLUME type.

stilez: The target audience isn't people who think technically about ZFS and its developer terminology. It's people who use the web UI, and in the web UI we universally do distinguish datasets and volumes that way. So stating only datasets here strongly implies "not volumes" to an end user.

william-gr: This is the API, text not available to end users. There is no `zvol` attribute in the API.

stilez: Damn. I wish I understood the codebase better :) Id like to offer direct PRs not just ask others to do the work. would that be okay?

william-gr: Any contribution in actual code will always be very welcome ;)
As long as you're comfortable receiving harsh feedback of people with OCD :D

stilez: And as long as you're okay reviewing rubbish PRs from people who don't really have a clue! :)

william-gr: I can approve as soon as you revert that line ; )

src/middlewared/middlewared/plugins/snapshot.py
```diff
@@ -95,11 +95,13 @@ async def do_create(self, data):
...
         Recursive snapshots can be created if `recursive` flag is enabled. You can `exclude` specific child datasets
-        from snapshot.
+        or zvols from the snapshot.
         Snapshots will be automatically destroyed after a certain amount of time, specified by
         `lifetime_value` and `lifetime_unit`.
+        If multiple periodic tasks create snapshots at the same time (for example hourly and daily at 00:00) the snapshot
+        will be kept until the last of these tasks reaches its expiry time.
```
william-gr, line 104: There is a whitespace at the end of this line too.

src/middlewared/middlewared/plugins/snapshot.py
stilez: Reverted. Don't know github well enough to see it, but it' not letting me re-revert it so hopefully it agrees its done. If not please do fix.

src/middlewared/middlewared/plugins/snapshot.py
stilez: Space fixed. As above, can't see but believe its done.

william-gr:
Thanks!

PR #5530, opened 2020-08-25
NAS-107213 / 12.1 / Fix validation for netbios names that are NULL

anodos325:
Also expand validation of netbios names for legacy HA configuration.
Predictable and consistent local user authentication with legacy HA
setup requires user to configure at least one netbios alias.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -215,9 +215,19 @@ class SMBService(SystemServiceService):
     @private
     async def smb_extend(self, smb):
         """Extend smb for netbios."""
-        smb['netbiosname_local'] = smb['netbiosname']
-        if not await self.middleware.call('system.is_freenas') and await self.middleware.call('failover.node') == 'B':
-            smb['netbiosname_local'] = smb['netbiosname_b']
+
+        ha_mode = SMBHAMODE[(await self.get_smb_ha_mode())]
+
+        if ha_mode == SMBHAMODE.STANDALONE:
+            smb['netbiosname_local'] = smb['netbiosname']
+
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -510,7 +510,7 @@ async def validate_smb(self, new, verrors):
             1. They contain invalid characters for NetBIOS protocol
             2. The name is identical to the NetBIOS workgroup.
             """
-            if not i:
+            if not new.get(i):
```
william-gr, line 513: What if for some reason netbiosname_b is null in the database and its not a truenas enterprise? Wont it still be an issue?

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -215,9 +215,19 @@ class Config:
...
+        if ha_mode == SMBHAMODE.STANDALONE:
+            smb['netbiosname_local'] = smb['netbiosname']
+
+        elif ha_mode == SMBHAMODE.LEGACY:
+            failover_node = await self.middleware.call('failover.node')
+            smb['netbiosname_local'] = smb['netbiosname'] if failover_node == 'A' else smb['netbiosname_b']
+
+        elif ha_mode == SMBHAMODE.UNIFIED:
+            ngc = await self.middleware.call('network.configuration.config')
+            smb['netbiosname_local'] = ngc['hostname_virtual']
```
yocalebo, line 230: Do we validate that `hostname_virtual` is provided at this point? I can't remember.

anodos325: This is the extend method so validation doesn't happen here.

yocalebo: Derp

PR #5536, opened 2020-08-25
NAS-107370 / 20.10 / Make sure hook is being called for a specific pool

william-gr:


william-gr:
retest

william-gr:
retest this please

themylogin:
@william-gr shouldn't we query all pools and sync zfs guid for each then?

william-gr:
@themylogin No, I am just fixing a traceback, I am not even convinced that hook is required to begin with, sysevent.fs.zfs.config_sync is already called when a pool is imported.

themylogin:
@william-gr then why don't you remove the hook completely?

william-gr:
@themylogin because I am not 100% sure thats the case and I didnt create the plugin.

themylogin:
@william-gr let me re-check and remove it then?

william-gr:
As you wish

themylogin:
@william-gr another important fix we want in 12.0. I have not found other renamed ZFS events like this.

src/middlewared/middlewared/plugins/disk_/zfs_guid.py
```diff
@@ -49,7 +49,7 @@ async def sync_zfs_guid(self, pool_id_or_pool):
 
 async def devd_zfs_hook(middleware, data):
     if data.get("type") in (
-        "misc.fs.zfs.config_sync",
+        "sysevent.fs.zfs.config_sync",
```
william-gr, line 52: Did this event name was actually renamed in devd? Did you verify yourself?

themylogin: `sysevent.fs.zfs.config_sync` is the event that is now appearing in devd pipe if that's what you're talking about, I have not been examining actual ZFS code.

PR #5541, opened 2020-08-26
NAS-107352 / 20.10 / filesystem.statfs for SCALE

themylogin:


src/middlewared/middlewared/plugins/filesystem_/stat_linux.py
```diff
@@ -0,0 +1,42 @@
...
+from middlewared.utils.path import is_child
+
+from .stat_base import FilesystemServiceBase
+
+
+class FilesystemService(Service, FilesystemServiceBase):
+
+    def statfs(self, path):
+        try:
+            st = os.statvfs(path)
```
william-gr, line 16: @anodos325 thought maybe we could use the same for FreeBSD.

Is os.statvfs available in FreeBSD just the same?

For a quick search we are only using it in UI to get free bytes in VM section.

themylogin: @william-gr yes, it is the same, we will only lose flags and fsid

william-gr:
@anodos325 would like your input here too

themylogin:
@anodos325 ping

PR #5561, opened 2020-08-27
NAS-106173 / 12.0 / adding code to wait on middleware to be online in api2/host_key.py

ericbsd:
added remove host_key from the skip test


ericbsd:
backport

PR #5564, opened 2020-08-27
NAS-107382 / 20.10 / Add support for kubernetes

sonicaj:


src/middlewared/middlewared/etc_files/systemd.py
```diff
@@ -10,7 +10,9 @@
 async def render(service, middleware):
     services = []
     services_enabled = {}
-    for service in await middleware.call("datastore.query", "services.services", [], {"prefix": "srv_"}):
+    for service in await middleware.call(
+        "datastore.query", "services.services", [["service", "nin", ["kubernetes"]]], {"prefix": "srv_"}
```
william-gr, line 14: We we are treating kubernetes as a special service, why does it have to be in services table?

src/middlewared/middlewared/plugins/docker_linux/images.py
```diff
@@ -0,0 +1,29 @@
...
+    class Config:
+        namespace = 'docker.images'
+
+    @private
+    async def load_images_from_file(self, path):
+        if not os.path.exists(path):
+            raise CallError(f'"{path}" path does not exist.')
+
+        # FIXME: Please do this in a better way
+        cp = await Popen(f'docker load < {path}', shell=True, stdout=subprocess.DEVNULL, stderr=subprocess.PIPE)
```
william-gr, line 22: Doesnt the API support stdin so we dont have to use shell?

src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @private
+    @lock('kubernetes_status_change')
+    async def status_change(self, config, old_config):
+        k3s_running = await self.middleware.call('service.started', 'kubernetes')
+        if config['pool'] and config['pool'] != old_config['pool']:
+            await self.setup_pool()
+            if not k3s_running:
+                await self.middleware.call('service.start', 'docker')
+                await asyncio.sleep(5)
```
william-gr, line 49: Why sleep?

src/middlewared/middlewared/plugins/kubernetes_linux/nodes.py
```diff
@@ -0,0 +1,42 @@
+from middlewared.schema import Dict, List, Str
+from middlewared.service import accepts, ConfigService
+
+from .k8s import api_client, nodes
+
+
+class KubernetesNodeService(ConfigService):
```
william-gr, line 7: Is this going to be a public service?

sonicaj: Nope, this is an internal service

src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
sonicaj: We try to setup the images too quickly and docker daemon has not initialised properly resulting in it failing our command.

src/middlewared/middlewared/plugins/docker_linux/images.py
sonicaj: It does for importing a single image, however I'll improve this - this was right now just done to have it working.

src/middlewared/middlewared/etc_files/systemd.py
sonicaj: We could either show that as Applications in the services table or let the user know it's k8s, so if you think we should maybe rename this to Applications, we can do that. 
Why we need to have it in services table is that not every user will want it and for ones who do, can start / stop / enable it then as desired.

william-gr: Well, dont think it should be a configurable service? Like right now we have Plugins and we cant just turn off plugins. That would be odd. Either you install them or you dont

src/middlewared/middlewared/etc_files/cni/net.d/multus.d/multus.kubeconfig.mako
```diff
@@ -0,0 +1,24 @@
...
+users:
+- name: multus
+  user:
+    token: "${multus['token']}"
+contexts:
+- name: multus-context
+  context:
+    cluster: local
+    user: multus
+current-context: multus-context
```
Ornias1993, line 24: Might want to add a newline here

src/middlewared/middlewared/etc_files/cni/net.d/kube-router.d/kubeconfig.mako
```diff
@@ -0,0 +1,25 @@
...
+users:
+- name: kube-router
+  user:
+    token: ${kube_router['token']}
+contexts:
+- name: kube-router-context
+  context:
+    cluster: cluster
+    user: kube-router
+current-context: kube-router-context
```
Ornias1993, line 25: might want to add a new line here too.

Ornias1993 (approved):
i'm not enough into the codebase (yet) to comment on the code itself.
But designwise it looks pretty neat, nicely structured an you covered for a lot of failure scenario's 👍

src/middlewared/middlewared/plugins/service_/services/all.py
```diff
@@ -22,6 +23,8 @@
 
 from .pseudo.ad import ActiveDirectoryService, LdapService, NisService
 from .pseudo.collectd import CollectDService, RRDCacheDService
+from .pseudo.docker_linux import DockerService
```
themylogin, line 26: I've imagined pseudo services as ones that were really one-shot FreeBSD rc scripts and we'll eliminate them all someday. These two seem to be real services.

src/middlewared/middlewared/plugins/service_/services/kubernetes_linux.py
```diff
@@ -0,0 +1,30 @@
...
+class KubernetesService(SimpleService):
+    name = 'kubernetes'
+    etc = ['k3s']
+    systemd_unit = 'k3s'
+
+    async def before_start(self):
+        await self.middleware.call('kubernetes.validate_k8s_fs_setup')
+
+    async def _start_linux(self):
+        await self.middleware.call('service.start', 'docker')
```
themylogin, line 15: Doesn't this code belong to `before_start`?

src/middlewared/middlewared/plugins/service_/services/kubernetes_linux.py
```diff
@@ -0,0 +1,30 @@
...
+        await self._unit_action('Start')
+
+    async def after_start(self):
+        asyncio.ensure_future(self.middleware.call('kubernetes.post_start'))
+
+    async def before_stop(self):
+        await self.middleware.call('k8s.node.add_taints', [{'key': 'ix-svc-stop', 'effect': 'NoExecute'}])
+        await asyncio.sleep(10)
+
+    async def _stop_linux(self):
```
themylogin, line 26: Same for `after_stop`

src/middlewared/middlewared/plugins/service_/services/pseudo/docker_linux.py
```diff
@@ -0,0 +1,13 @@
+from middlewared.plugins.service_.services.base import SimpleService
+
+
+class DockerService(SimpleService):
+    name = 'docker'
+    etc = ['docker']
+    systemd_unit = 'docker'
+
+    async def _start_linux(self):
```
themylogin, line 9: Why override with same method body?

themylogin (changes requested):

src/middlewared/middlewared/etc_files/docker.py
```diff
@@ -3,13 +3,18 @@
 
 
 def render(service, middleware):
-    sys_config = middleware.call_sync('systemdataset.config')
-    if not sys_config['path']:
+    config = middleware.call_sync('kubernetes.config')
+    if not config['pool']:
```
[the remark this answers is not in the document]
sonicaj: I'm not sure I follow, how will it keep the old config file ? We will be overwriting that below when we write out the config file. However if you are referring to keeping the old dataset, it won't be an issue and I think the user can destroy it himself if he/she wants ?

src/middlewared/middlewared/etc_files/rancher/k3s/flags.py
```diff
@@ -0,0 +1,17 @@
...
+def render(service, middleware):
+    config = middleware.call_sync('kubernetes.config')
+    kube_controller_args = f'node-cidr-mask-size={ipaddress.ip_network(config["cluster_cidr"]).prefixlen}'
+    os.makedirs('/etc/rancher/k3s', exist_ok=True)
+    with open('/etc/rancher/k3s/flags.yaml', 'w') as f:
+        f.write(yaml.dump({
+            'cluster-cidr': config['cluster_cidr'],
+            'service-cidr': config['service_cidr'],
+            'cluster-dns': config['cluster_dns_ip'],
+            'data-dir': os.path.join('/mnt', config['dataset'], 'k3s'),
```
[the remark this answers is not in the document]
sonicaj: It is but this file is only generated after k3s has been configured so it won't be null in that case. I'll still add a check for good measure

src/middlewared/middlewared/etc_files/docker.py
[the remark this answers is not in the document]
themylogin: If you `return` here, `/etc/docker/daemon.json` will keep old `data-root` path, won't it?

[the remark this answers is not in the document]
sonicaj: But we wouldn't be starting docker daemon in that case ? For good measure I can still remove the file for just in case cases.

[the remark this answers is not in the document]
themylogin: You have a point, thank you for explanation?

src/middlewared/middlewared/plugins/kubernetes_linux/update.py
```diff
@@ -106,8 +106,16 @@ async def do_update(self, job, data):
...
         0.0.0.0 and the cluster in this case will automatically manage which IP address to use for managing traffic
-        for default NAT network. If it is desired that traffic uses a certain interface / ip address, that IP address
-        can be specified and the NAT network will use related IP address and it's routes to manage the traffic.
+        for default NAT network.
+
+        By default kubernetes pods will be using default gateway of the system for outward traffic. This might
+        not be desirable for certain users who want to separate NAT traffic over a specific interface / route. System
+        will create a L3 network which will be routing the traffic towards default gateway for NAT.
+
+        If users want to restrict traffic over a certain gateway / interface, they can specify a default route
```
Ornias1993, line 115: I needed to read it two times before realising this would also be the way to deal with pushing k8s traffic through a vlan. Maybe clear this up a bit?
```suggestion
        If users want to restrict traffic over a certain gateway / interface / vlan, they can specify a default route
```

sonicaj: Thank you @Ornias1993 for pointing it out however I am not sure about it because vlan is an interface ? Let's ask @freenas/docs-team for review so they can recommend what would be clearer and we can adjust accordingly.

Ornias1993: > vlan is an interface

I agree, so I got what you meant in the end... But if this is user facing, some users might not fully grasp the "virtual interface = still an interface" thingy...

Nice work this week though 👍

src/middlewared/middlewared/etc_files/cni/net.d/multus.d/multus.kubeconfig.mako
```diff
@@ -0,0 +1,24 @@
+<%
+    config = middleware.call_sync('datastore.query', 'services.kubernetes', [], {'get': True})
```
william-gr, line 2: Why using the datastore? Isnt there a service interface for this datastore?

src/middlewared/middlewared/etc_files/rancher/k3s/flags.py
```diff
@@ -0,0 +1,29 @@
...
+
+FLAGS_PATH = '/etc/rancher/k3s/config.yaml'
+
+
+def render(service, middleware):
+    shutil.rmtree('/etc/cni/net.d', ignore_errors=True)
+    config = middleware.call_sync('kubernetes.config')
+    if not config['pool']:
+        with contextlib.suppress(OSError):
+            os.unlink(FLAGS_PATH)
```
william-gr, line 16: If there isnt a pool shouldnt we stop here?

src/middlewared/middlewared/plugins/docker_linux/images.py
```diff
@@ -0,0 +1,32 @@
...
+from middlewared.service import CallError, private, CRUDService
+
+
+DEFAULT_DOCKER_IMAGES_PATH = '/usr/local/share/docker_images/docker-images.tar'
+
+
+class DockerImagesService(CRUDService):
+
+    class Config:
+        namespace = 'docker.images'
```
william-gr, line 13: Is this really supposed to be a public service?

src/middlewared/middlewared/plugins/docker_linux/images.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @private
+    async def load_images_from_file(self, path):
+        if not os.path.exists(path):
+            raise CallError(f'"{path}" path does not exist.')
```
william-gr, line 18: errno.ENOEXIST

src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
```diff
@@ -0,0 +1,135 @@
...
+from middlewared.service import CallError, private, Service
+
+
+class KubernetesService(Service):
+
+    @private
+    async def post_start(self):
+        # TODO: Add support for migrations
+        await asyncio.sleep(10)
+        # 10 secs should be enough for the node to come up online and kube-api servers to start accepting calls.
```
william-gr, line 15: Thius should be a FIXME, waiting arbitrary amount of time is no good

src/middlewared/middlewared/etc_files/cni/net.d/multus.d/multus.kubeconfig.mako
sonicaj: We use it because with the service interface, we pop out the cni config as it's of no use to the user and is just used for internal handling of the CNI plugins.

src/middlewared/middlewared/plugins/docker_linux/images.py
sonicaj: Yes, we build on this service for the complete docker images part which we want to expose in the UI

src/middlewared/middlewared/plugins/docker_linux/images.py
sonicaj: I believe you refer to `ENOENT` ?

src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class KubernetesService(Service):
+
+    @private
+    async def post_start(self):
+        # TODO: Add support for migrations
+        async def node_online(middleware):
+            while not (await middleware.call('k8s.node.config'))['node_configured']:
+                await asyncio.sleep(2)
```
william-gr, line 16: This seems like an overengineering use of coroutines since its just a sleep, could very well be a while with a counter

src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
```diff
@@ -0,0 +1,141 @@
...
+        await self.middleware.call('k8s.cni.setup_cni')
+        await self.middleware.call(
+            'k8s.node.remove_taints', [
+                k['key'] for k in (node_config['spec']['taints'] or []) if k['key'] in ('ix-svc-start', 'ix-svc-stop')
+            ]
+        )
+        # We should wait for around 20 seconds so that pods are scheduled and start executing on the node as it is
+        # then that kube-router configures routes in the main table which we would like to add to kube-router table
+        # because it's internal traffic will also be otherwise advertised to the default route specified
+        await asyncio.sleep(20)
```
william-gr, line 45: Lets have a FIXME/TODO here too, sleeps make me twitch :P

src/middlewared/middlewared/plugins/docker_linux/images.py
william-gr: Yes

william-gr (approved):

src/middlewared/middlewared/plugins/kubernetes_linux/cni.py
```diff
@@ -0,0 +1,114 @@
...
+                    cni_config[cni] = await service_accounts.get_service_account_details(
+                        context['core_api'], config[cni]['service_account']
+                    )
+
+        await self.middleware.call(
+            'datastore.update', 'services.kubernetes', kube_config['id'], {'cni_config': cni_config}
+        )
+        await self.middleware.call('etc.generate', 'cni')
+        await self.middleware.call('service.start', 'kuberouter')
+        await asyncio.sleep(5)
```
themylogin, line 39: Is this the best way to wait for routing tables to appear?

src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
```diff
@@ -0,0 +1,148 @@
...
+
+            self.middleware.call_sync('alert.oneshot_delete', 'ApplicationsConfigurationFailed', None)
+
+    @private
+    async def status_change_internal(self):
+        await self.validate_k8s_fs_setup()
+        await self.middleware.call('service.start', 'docker')
+        # This is necessary because docker daemon requires a couple of seconds after starting to initialise itself
+        # properly, if we try to load images without the delay that will fail and will only correct after a restart
+        await asyncio.sleep(5)
```
themylogin, line 119: We have TrueNAS installed on machines slow as hell, with rootfs on USB sticks, shouldn't we use another method to wait for docker to appear?

src/middlewared/middlewared/plugins/service_/services/kubernetes.py
```diff
@@ -0,0 +1,35 @@
...
+    etc = ['k3s']
+    systemd_unit = 'k3s'
+
+    async def before_start(self):
+        await self.middleware.call('kubernetes.validate_k8s_fs_setup')
+        await self.middleware.call('service.start', 'docker')
+        await self._systemd_unit('cni-dhcp', 'start')
+
+    async def _start_linux(self):
+        await self._unit_action('Start')
```
themylogin, line 17:
```suggestion
        await super()._start_linux()
```

themylogin (changes requested):
Outstanding job! But I am in doubt wrt to large amount of `sleep`s

sonicaj:
@themylogin i agree that we should eliminate all of these sleeps. I was thinking of handling this in another PR but if you think we should perhaps do this here, I can do that as well ?

themylogin:
@sonicaj I also agree that we are short in time for ALPHA, so I've commented on the two I think are the most likely to fail and for the rest FIXME will be enough for me for now. Docker being initialized is a matter of socket existing, right? And kuberouter is the matter of having a table existing?

themylogin (approved):

src/middlewared/middlewared/plugins/kubernetes_linux/cni.py
Ornias1993: You could just do a call to the docker API and wait while that call fails.

src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
Ornias1993: You could just do a call to the docker API and wait while that call fails.

PR #5568, opened 2020-08-28
NAS-107385 / 20.10 / Use AES-256-CCM as default zfs cipher

sonicaj:
This commit adds changes to use AES-256-CCM as default cipher as it's faster then AES-256-GCM and is the default used by zfs(8)


sonicaj:
Just discussed with @freqlabs, zfs man page is out of date, `gcm` should be used which we are already using. Closing the PR.

PR #5577, opened 2020-08-31
NAS-107409 / 20.10 / ensure nfs shares are within zpool mountpoint

yocalebo:


src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -389,6 +390,10 @@ async def validate(self, data, schema_name, verrors, old=None):
         if data["alldirs"] and len(data["paths"]) > 1:
             verrors.add(f"{schema_name}.alldirs", "This option can only be used for shares that contain single path")
 
+        # need to make sure that the nfs share is within the zpool mountpoint
+        for i in data["paths"]:
+            await check_path_resides_within_volume(verrors, self.middleware, schema_name, i)
```
william-gr, line 395: I wonder if `schema_name` should be `f'{schema_name}.paths.{idx}` instead

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -389,6 +390,12 @@ async def validate(self, data, schema_name, verrors, old=None):
         if data["alldirs"] and len(data["paths"]) > 1:
             verrors.add(f"{schema_name}.alldirs", "This option can only be used for shares that contain single path")
 
+        # need to make sure that the nfs share is within the zpool mountpoint
+        for idx, i in enumerate(data["paths"]):
+            await check_path_resides_within_volume(
+                verrors, self.middleware, f'{schema_name}.path.{idx}', i
```
william-gr, line 396: I think `paths` is the attribute name in the schema, not `path`

PR #5582, opened 2020-08-31
NAS-107388 / 20.10 / add gluster path validation to all sharing services

yocalebo:


src/middlewared/middlewared/async_validators.py
```diff
@@ -7,13 +7,24 @@
...
-        verrors.add(name, "The path must reside within a volume mount point")
+        verrors.add(name, "The path must reside within a pool mount point")
+
+    # we must also make sure that any sharing service does not point to
+    # anywhere within the ".glusterfs" dataset since the clients need
+    # to go through the appropriate gluster client to write to the cluster.
+    # If data is modified directly on the gluster resources, then it will
+    # cause a split-brain scenario which means the data that was modified
+    # would not be sync'ed with other nodes in the cluster.
+    if "/.glusterfs/" in rp:
```
sonicaj, line 26: I believe this should be in the gluster service itself as it is specific to gluster.

william-gr: How? We dont want any other services to be shared with a gluster path

yocalebo: The way I could do this is add a method in gluster plugin, and then change all service plugins to call it. Essentially doing the same thing that's already done with the `check_path_resides_within_volume` method and all the service plugins. Putting it here is me mostly being lazy. I don't mind either way.

william-gr: Putting this check in every service also make them dependent of the gluster service, so its essentially the same thing but duplicated in every plugin.

yocalebo: I would agree, I also don't want to run "os.path.realpath" again in the gluster method if I add it that way. I think leaving it here is the best place for now.

sonicaj: @william-gr @yocalebo you are right. I think we should just ensure that this is actually what we want to filter then like user creating a path `/mnt/pool/somedir/somedir/.glusterfs/asd` will also result in an error.
Not sure if this is error is desirable in that situation.

anodos325: What if we switch to using pathlib?
```
from pathlib import Path
rp = Path(path).resolve()
```
Then walk through the parents of rp and check for `.glusterfs` that is also a mountpoint
```
for p in rp.parents:
    if p.is_mount() and p == ".glusterfs":
        verrors.add(name, "A path being used by Gluster is not allowed")
```

yocalebo: @anodos325 I agree. TrueCommand is hardcoded to put the top-level `.glusterfs` dataset at the root of the zpool but it'd be nice let user put it wherever they liked with proper validation.

william-gr (approved):

anodos325 (approved):

PR #5591, opened 2020-09-01
NAS-107438 / 11.3 / Reintroduce special behavior for HOMES path

anodos325:
The SMB plugin had undocumented behavior prior to 11.3 whereby in non-AD
enviorments, administrators could leave the SMB path empty for HOMES
shares and thereby allow samba to auto-generate the path based on the
user's home directory path configured in the GUI.


anodos325:
I will need to take another look at this and decide perhaps to limit scope to when LDAP directory service is enabled. This could result in automatic creation on /nonexistent for local users.

PR #5592, opened 2020-09-02
NAS-107437 / 11.3 / Expand validation for user home directories

anodos325:
Re-implement validation from legacy_ui preventing users from setting zpool mountpoints as a home directory.

Require owner read and execute on a home directory.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -18,6 +18,7 @@ import string
 import stat
 import subprocess
 import time
+from pathlib import Path
 
 SKEL_PATH = '/usr/share/skel/'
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/account.py
```diff
@@ -663,11 +663,26 @@ async def __common_validation(self, verrors, data, schema, pk=None):
                         f'The path for the home directory "({data["home"]})" '
                         'must include a volume or dataset.'
                     )
+                elif os.path.realpath(data['home']) in [x['path'] for x in await self.middleware.call('pool.query')]:
+                    verrors.add(
+                        f'{schema}.home',
+                        f'The specified path is a ZFS pool mountpoint "({data["home"]})" '
+                    )
```
william-gr, line 670: We could probably get around without that extra expensive pool.query call?
Disallowing two level deep path?
/mnt/$anything?

PR #5598, opened 2020-09-02
NAS-107418 / 12.0 / proper validation on excluded disks in fenced

yocalebo:
argsparse is a string separated by commas for each disk specified in the exclusion parameter. Convert that to a list and be explicit with checking if that disk is in the exclusion list.

### DIFF
--- a/src/fenced/fenced/fence.py
+++ b/src/fenced/fenced/fence.py
@@ -32,7 +32,7 @@ class Fence(object):
 
     def __init__(self, interval, exclude_disks):
         self._interval = interval
-        self._exclude_disks = exclude_disks
+        self._exclude_disks = exclude_disks.split(',') if exclude_disks else exclude_disks
         self._disks = Disks(self)
         self._reload = False
         self.hostid = None
@@ -68,6 +68,7 @@ class Fence(object):
             # Only care about da and nvd devices
             if not i.startswith(('da', 'nvd')):
                 continue
+
             # You can pass an "--exclude-disks" argument to fenced
             # to exclude disks from getting SCSI reservations.
             # fenced is called in 12+ with the exclusion flag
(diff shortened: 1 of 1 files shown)


src/fenced/fenced/fence.py
```diff
@@ -68,13 +68,16 @@ def load_disks(self):
...
                 continue
+
             # You can pass an "--exclude-disks" argument to fenced
             # to exclude disks from getting SCSI reservations.
             # fenced is called in 12+ with the exclusion flag
             # because the newer generation M-series are going to
             # have NVMe boot drives
-            if i in self._exclude_disks:
-                continue
+            for j in self._exclude_disks:
```
william-gr, line 77: I dont understand what is changing here? Seems like its essentially the same thing but less performant?

yocalebo: argsparse is passing a string like this `ada0,ada1` on systems with SATA DOMs.

if the disk is `da0` and the DOM is `ada0` the `if i in "ada0,ada1"` was met and is actually excluding `da0`.

william-gr: This doesnt explain this change, it will be exactly the same thing just written in different python.

yocalebo: It does though. `self._excluded_disks` was a single line string of `ada0,ada1`.

>>> ed = 'ada0,ada1'
>>> i = 'da0'
>>> i in ed
True

william-gr: You already handled this earlier doing the split.
This change in `for` loop is a no-op.

yocalebo: Sorry, I thought you were talking about the `self._exclude_disks` attribute.

PR #5601, opened 2020-09-02
NAS-107445 / 12.0 / Remove unprotected setfacl / chmod call

anodos325:
Permissions operations during home directory move bypass our middleware API for
permissions. Move the copy and permissions operation into a middleware job and
use API.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -1,6 +1,6 @@
 from middlewared.schema import accepts, Any, Bool, Dict, Int, List, Patch, Str
 from middlewared.service import (
-    CallError, CRUDService, ValidationErrors, item_method, no_auth_required, pass_app, private, filterable
+    CallError, CRUDService, ValidationErrors, item_method, no_auth_required, pass_app, private, filterable, job
 )
 import middlewared.sqlalchemy as sa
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/account.py
```diff
@@ -692,6 +685,22 @@ async def set_root_password(self, app, password, options):
         root = await self.middleware.call('user.query', [('username', '=', 'root')], {'get': True})
         await self.middleware.call('user.update', root['id'], {'password': password})
 
+    @job(lock="home_copy")
```
themylogin, line 688: We can use lock that includes `home_new` in its name so multiple independent calls could be performed simultaneously

themylogin (changes requested):

src/middlewared/middlewared/plugins/account.py
anodos325: Good idea

william-gr: Should probably mark it as private as well

Also we should probably keep the change as minimal as possible in the 12.0 backport

themylogin (approved):

william-gr (changes requested):
Need to fix PEP8 before I can merge

william-gr (approved):

PR #5613, opened 2020-09-03
NAS-107459 / 11.3 / open file and then loop to reduce cpu cycles

yocalebo:
profiling middlewared has shown that activedirectory plugin uses quite a few cpu cycles. Instead of looping and opening the file repeatedly, open the file once and then loop.

### DIFF
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -1873,8 +1873,8 @@ class WBStatusThread(threading.Thread):
         self.state = new_state
 
     def read_messages(self):
-        while not self.finished.is_set():
-            with open('/var/run/samba4/.wb_fifo') as f:
+        with open('/var/run/samba4/.wb_fifo') as f:
+            while not self.finished.is_set():
                 data = f.read()
                 for msg in data.splitlines():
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1873,8 +1873,8 @@ def parse_msg(self, data):
         self.state = new_state
 
     def read_messages(self):
-        while not self.finished.is_set():
-            with open('/var/run/samba4/.wb_fifo') as f:
+        with open('/var/run/samba4/.wb_fifo') as f:
```
william-gr, line 1876: Can the fifo be closed if samba is restarted and we need to keep retrying still? @anodos325

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1873,8 +1873,8 @@ def parse_msg(self, data):
...
+            while not self.finished.is_set():
                 data = f.read()
```
william-gr, line 1878: Should we break if data == '' ?

yocalebo:
Working with @anodos325 directly. Will open a new PR if necessary.

PR #5635, opened 2020-09-08
Add GPU support for k8s cluster

sonicaj:
This PR adds support for setting up k8s cluster to make use of GPU available on the system. Right now this is limited to NVIDIA based on the hardware available but we can easily extend this for other vendors.


themylogin (approved):

src/middlewared/middlewared/plugins/hardware_linux/gpu.py
```diff
@@ -0,0 +1,36 @@
+import re
+
+from middlewared.service import Service
+from middlewared.utils import run
+
+
+RE_VENDOR = re.compile(r'description:\s*VGA compatible controller[\s\S]*vendor:\s*(.*)')
+
+
+class HardwareService(Service):
```
william-gr, line 10: We already have a device service, perhaps we can extend it for GPU?

src/middlewared/middlewared/etc_files/docker.py
```diff
@@ -1,15 +1,45 @@
 import json
 import os
+import re
 import subprocess
 
 
+RE_LDCONFIG = re.compile(r'@/sbin/ldconfig')
```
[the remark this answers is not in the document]
william-gr: Same question

src/middlewared/middlewared/plugins/device_/gpu_linux.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+RE_VENDOR = re.compile(r'description:\s*VGA compatible controller[\s\S]*vendor:\s*(.*)')
+
+
+class DeviceService(Service):
+
+    GPU = None
+
+    async def available_gpu(self):
```
william-gr, line 14: This is not using the same API...
I meant adding a `GPU` entry to device.get_info

src/middlewared/middlewared/etc_files/docker.py
```diff
@@ -27,7 +23,7 @@ def nvidia_configuration(middleware):
 
 
 def gpu_configuration(middleware):
-    available_gpu = middleware.call_sync('hardware.available_gpu')
+    available_gpu = middleware.call_sync('device.get_gpus')
```
william-gr, line 26: We are not supposed to call them directly, but device.get_info GPU, because the others are private methods, unless strictly necessary

PR #5636, opened 2020-09-09
NAS-107501 / 20.10 / Fix pdbedit parsing when SMB service aux param invalid

anodos325:
When user adds invalid auxilary parameter(s) to the smb.conf, pdbedit
invocations (or any utilty that ultimately calls lpcfg_map_parameter()),
will result in additional warning messages being printed even at DEBUG 0
on stdout. This may break parsing of output from the CLI utility, and in
the case of pdbedit, will prevent updating of user password.

This commit will change two things. First is a minimal workaround for the
passdb update to splitlines() and use only last element of resulting list.
Warnings in this specific situation are prepended to the command's output
and so will be safely discarded. Additionally, validation for smb_options
(auxiliary parameters for the SMB service) will be expanded to add a
call to LoadParm.dump_a_parameter(<aux param>). This function will raise
a RuntimeException if the parameter being queried does not exist in Samba's
internal param table.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -1028,6 +1028,21 @@ class SharingSMBService(SharingService):
                     'are not permitted.'
                 )
 
+            if schema_name == 'smb_update.smb_options' and ':' not in kv[0]:
+                """
+                lib/param doesn't validate params containing a colon.
+                this dump_a_parameter() wraps around the respective lp_ctx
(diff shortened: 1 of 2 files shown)


william-gr (approved):
That can wait 12.0-U1 since its an edge case regarding aux params?

anodos325:
> That can wait 12.0-U1 since its an edge case regarding aux params?

Yes. It's just part of that continuing effort to catch / handle edge cases. No harm in waiting for U1.

PR #5637, opened 2020-09-09
NAS-107503 / 20.10 / Allow keytab setup through NFS plugin

anodos325:
Kerberized NFS requires that the system keytab contain an NFS
Service Principal Name (SPN) entry. For various reasons a user may
need to configure kerberos authentication for NFS at a later date
after initial directory services setup. This PR introduces a method
to do this via the NFS plugin. Currently, only the Active Directory
(AD) directory service is supported.

The API call uses the user-provided credentials to kinit, and then
uses the resulting kerberos ticket for a privileged account to modify
the AD computer account and generate NFS SPN entries for the TrueNAS
server in the AD environment. Once this is complete, the system keytab
/etc/krb5.keytab is updated and the privileged kerberos ticket is destroyed.

### DIFF
--- a/src/middlewared/middlewared/plugins/nfs_/krb5.py
+++ b/src/middlewared/middlewared/plugins/nfs_/krb5.py
@@ -1,0 +1,102 @@
+import errno
+from middlewared.service import private, Service
+from middlewared.service_exception import CallError
+from middlewared.plugins.idmap import DSType
+from middlewared.plugins.directoryservices import DSStatus
+from middlewared.schema import accepts, Dict, Str
+
+
+class NFSService(Service):
+
+    class Config:
+        service = "nfs"
+        service_verb = "restart"
+        datastore_prefix = "nfs_srv_"
+        datastore_extend = 'nfs.nfs_extend'
+
+    @private
+    async def validate_directoryservices(self, ds):
+        """
+        Validation currently only succeeds in case of HEALTHY AD
+        directory service. At future point, depending on demand, the
+        same convenience feature can be added for the LDAP directory
+        service.
+        """
+        ds = await self.middleware.call('directoryservices.get_state')
+        ad_status = DSStatus[ds['activedirectory']]
+        ldap_status = DSStatus[ds['ldap']]
+
+        if ldap_status != DSStatus.DISABLED:
+            raise CallError('This feature has not yet been implemented for '
+                            'the LDAP directory service.', errno=errno.ENOSYS)
+
+        if ad_status != DSStatus.HEALTHY:
+            raise CallError('Active Directory Directory Service is currently'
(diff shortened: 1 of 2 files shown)


anodos325:
Idea here is to present an API endpoint to GUI team / end-users that allows them to provide admin credentials (after joining an AD domain) and have us generate the necessary Kerberos entries in AD to allow NFS + KRB5 to work.

src/middlewared/middlewared/plugins/nfs_/krb5.py
```diff
@@ -0,0 +1,95 @@
...
+        directory service. At future point, depending on demand, the
+        same convenience feature can be added for the LDAP directory
+        service.
+        """
+        ds = await self.middleware.call('directoryservices.get_state')
+        ad_status = DSStatus[ds['activedirectory']]
+        ldap_status = DSStatus[ds['ldap']]
+
+        if ldap_status != DSStatus.DISABLED:
+            raise CallError('Automated addition of NFS SPN records has '
```
themylogin, line 30: Is this a correct error message?

src/middlewared/middlewared/plugins/nfs_/krb5.py
```diff
@@ -0,0 +1,95 @@
...
+        relying on the existing kerberos ticket / principal.
+        """
+        ad = await self.middleware.call('activedirectory.config')
+        ad['dstype'] = DSType.DS_TYPE_ACTIVEDIRECTORY.value
+        ad['bindname'] = data.get("username", "")
+        ad['bindpw'] = data.get("password", "")
+        ad['kerberos_principal'] = ''
+
+        await self.middleware.call('kerberos.do_kinit', ad)
+        return await self.middleware.call('activedirectory.add_nfs_spn', ad)
```
yocalebo, line 53: What happens if the SPN entry already exists on an object in AD? Will this return appropriately or at least hint to the end-user why it failed to add?

yocalebo (approved):

src/middlewared/middlewared/plugins/nfs_/krb5.py
anodos325: Just to document this... if we fail to set the SPN in AD, then we raise a CallError with information about why it failed. Whether the information contained in stderr from setspn will be of use to support is up for debate, but at least they will know that an error of sorts occurred.

src/middlewared/middlewared/plugins/nfs_/krb5.py
anodos325: It's unclear and pre-coffee english. In that regard, no... it's barely grammatical. I'll simplify the message.

themylogin (approved):

PR #5655, opened 2020-09-10
NAS-107519 / 20.10 / Prevent middleware from killing itself

anodos325:
Check whether the process that is being flagged for stopping / killing
is the parent middlewared process or one of our child processes. Skip
in this case.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -29,7 +29,7 @@ from middlewared.service import (
 )
 from middlewared.service_exception import ValidationError
 import middlewared.sqlalchemy as sa
-from middlewared.utils import osc, Popen, filter_list, run, start_daemon_thread
+from middlewared.utils import osc, Popen, filter_list, run, start_daemon_thread, middleware_pid
 from middlewared.utils.asyncio_ import asyncio_map
 from middlewared.utils.path import is_child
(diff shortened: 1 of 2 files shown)


anodos325:
Observed during testing:
`[2020/09/09 12:08:27] (INFO) PoolDatasetService.kill_processes():3728 - Killing process 796 ("'python3.8: middlewared'") that holds dataset 'tank'`

I'm open to more clever ways of dealing with this issue.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3746,7 +3748,13 @@ async def kill_processes(self, oid, control_services, max_tries=5):
             if not processes:
                 return
 
+            mid_procs.extend([x.pid for x in psutil.Process(middleware_pid()).children()])
```
themylogin, line 3751: We only need to keep main middleware process. If multiprocessing worker is killed, it's fine, it will be restarted. If rclone/rsync are killed, it's fine too.

themylogin (changes requested):

src/middlewared/middlewared/plugins/pool.py
anodos325: Okay. I'll only check for the main pid.

themylogin (approved):

PR #5659, opened 2020-09-10
NAS-107505 / 12.0 / Unlock replication tasks on failover pool import

themylogin:


yocalebo:
@themylogin is this needed in 11.3?

themylogin:
@yocalebo I think it should be, but there were no issues filed so far, I wonder why it works

william-gr (approved):

PR #5663, opened 2020-09-10
NAS-107521 / 20.10 / multiple fixes for network general validation

yocalebo:
This fixes a few things.

1. we were doing double computation in the `do_update` method when all of the work had already been done in the `_extend` method.
2. we were reloading the hostname service twice, even if the hostname didn't change
3. reloading `resolvconf` service actually sets the hostname too so in interest of keeping api calls to a minimal just reload that service in general as necessary
4. we were restarting NFS service unnecessarily so add another check there too


src/middlewared/middlewared/plugins/network.py
```diff
@@ -213,97 +215,85 @@ async def do_update(self, data):
...
-            data = dict(
-                data,
-                domains=json.dumps(sorted(list(set('domains'))), sort_keys=True),
-                service_announcement=None,
-                netwait_ip=json.dumps(sorted(list(set('netwait_ip'))), sort_keys=True),
-                activity=json.dumps(data['activity'], sort_keys=True),
-            )
-            return set(data.items())
-
-        if len(set_(new_config) ^ set_(config)) > 0:
```
william-gr, line 263: Not sure why we had to remove this?

yocalebo: A few reasons.

1. prone to breakage if new items are added to API
(for example, service_announcement key was being overwritten to None)
2. potentially expensive
(calling json.dump on all keys in the dict twice for each key that couldn't be flattened normally)
3. the new activity feature was tied to when we restarted hostname/resolvconf which isn't needed. So if you only changed the outbound activity, we would restart hostname/resolvconf?

src/middlewared/middlewared/plugins/network.py
```diff
@@ -68,18 +67,21 @@ class Config:
...
     @private
     def network_config_extend(self, data):
+
         # hostname_local will be used when the hostname of the current machine
-        # needs to be used so it works with either FreeNAS or TrueNAS
+        # needs to be used so it works with either TrueNAS CORE or ENTERPRISE
         data['hostname_local'] = data['hostname']
-        if self.middleware.call_sync('system.is_freenas'):
+
+        if not self.middleware.call_sync('system.product_type') in ('ENTERPRISE', 'SCALE_ENTERPRISE'):
```
themylogin, line 75: `system.is_enterprise`

themylogin (changes requested):

src/middlewared/middlewared/plugins/network.py
yocalebo: 👍

william-gr (approved):

anodos325 (approved):

themylogin (approved):

PR #5669, opened 2020-09-14
NAS-107526 / 20.10 / Do not run camcontrol reprobe for NVMe

themylogin:


william-gr (approved):
Are we sure NVMe doesn't require another special command? or using /dev/nvme* ?

themylogin:
@william-gr camcontrol does not work for /dev/nvmeX as well, I've found nothing online, and it does not seem to be necessary as I am not aware of any cases where NVMe device can change it's size.

PR #5678, opened 2020-09-16
NAS-107595 / 20.10 / Add interfaces to reporting.realtime subscription

william-gr:


src/middlewared/middlewared/plugins/reporting/events.py
```diff
@@ -160,6 +161,31 @@ def run(self):
...
+                        data['interfaces'][iface_name][name] = value
+                        traffic_stats = None
+                        if (
+                            last_interface_stats.get(iface_name) and
+                            name in last_interface_stats[iface_name]
+                        ):
+                            traffic_stats = value - last_interface_stats[iface_name][name]
+                            traffic_stats = int(
+                                traffic_stats / (
+                                    stats_time - last_interface_stats[iface_name]['stats_time']
```
themylogin, line 180: Won't it be better to use `monotime` for calculations like this?

william-gr: It would, it just copied the block/logic from the freebsd side though

PR #5681, opened 2020-09-17
NAS-107348 / 12.0 / Cache pwenc secret

themylogin:


william-gr (commented):
Dont we also change that file when uploading config fileb bundle?

themylogin:
@william-gr we upload it to `/data/pwenc_secret_uploaded` to avoid collisions with currently running configuration and apply it on reboot before middleware starts

william-gr (approved):

PR #5682, opened 2020-09-17
NAS-107516 / 11.3 / S3 max_upload_parts

themylogin:


william-gr (commented):
Is this really necessary? Could we have a workaround in code for specific more famous providers like Scaleway?

themylogin:
@william-gr I would love to do this, but we don't know for sure which providers are like that, and we are can't be sure new ones won't appear

william-gr:
@themylogin but it would work out of the box for the ones we know?

william-gr (approved):

PR #5687, opened 2020-09-18
NAS-107617 / 12.0 / Catch exceptions during ZFS property update on system dataset

anodos325:
System dataset setup may raise an exception when trying to set a
ZFS dataset quota on the .system/cores. This will result in only
partial initialization of the system dataset and break SMB authentication.

Catch and log exception in this case.


anodos325:
Maybe a smarter idea is to check `written` before setting quota, and scale up if needed.

PR #5688, opened 2020-09-18
NAS-107617 / 12.0 / Enforce 1GiB limit on cores dataset

anodos325:
Destroy the cores dataset if it has > 1GiB written to it, then re-create with a 1GiB quota.

### DIFF
--- a/src/middlewared/middlewared/plugins/sysdataset.py
+++ b/src/middlewared/middlewared/plugins/sysdataset.py
@@ -296,13 +296,23 @@ class SystemDatasetService(ConfigService):
             i['id']: i['properties'] for i in await self.middleware.call('zfs.dataset.query', [('id', 'in', datasets)])
         }
         for dataset in datasets:
-            dataset_quota = {'quota': '1G'} if dataset.endswith('/cores') else {}
+            is_cores_ds = dataset.endswith('/cores')
+            dataset_quota = {'quota': '1G'} if is_cores_ds else {}
             if dataset not in datasets_prop:
                 await self.middleware.call('zfs.dataset.create', {
                     'name': dataset,
                     'properties': {'mountpoint': 'legacy', **dataset_quota},
                 })
                 createdds = True
+            elif is_cores_ds and datasets_prop[dataset]['written']['parsed'] > 1024 ** 3:
+                try:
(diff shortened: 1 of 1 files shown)


yocalebo (changes requested):
You’re probably going to want to wrap the deletion of the cores dataset in a try/except because it routinely gets “locked” by user space applications. Particularly in the past, middlewared will try to dump 50G worth and it will go catatonic and lock it up.

anodos325:
> You’re probably going to want to wrap the deletion of the cores dataset in a try/except because it routinely gets “locked” by user space applications. Particularly in the past, middlewared will try to dump 50G worth and it will go catatonic and lock it up.

I'll add try/except, but the situation where this code is most likely to get run is during boot just after upgrading to 12.0 (system dataset setup).

yocalebo (approved):

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -296,13 +296,23 @@ async def __setup_datasets(self, pool, uuid):
...
                 createdds = True
+            elif is_cores_ds and datasets_prop[dataset]['written']['parsed'] > 1024 ** 3:
+                try:
+                    await self.middleware.call('zfs.dataset.delete', dataset)
+                    await self.middleware.call('zfs.dataset.create', {
+                        'name': dataset,
+                        'properties': {'mountpoint': 'legacy', **dataset_quota},
+                    })
+                except Exception:
+                    self.logger.warning("Failed to replace dataset [%].", dataset, exc_info=True)
```
william-gr, line 315: Missing %s

william-gr (changes requested):

william-gr (approved):

yocalebo:
> > You’re probably going to want to wrap the deletion of the cores dataset in a try/except because it routinely gets “locked” by user space applications. Particularly in the past, middlewared will try to dump 50G worth and it will go catatonic and lock it up.
> 
> I'll add try/except, but the situation where this code is most likely to get run is during boot just after upgrading to 12.0 (system dataset setup).

pretty sure this runs when (not positive):
1. fresh HA system and the 1st zpool is created (we auto migrate the system dataset to zpool)
2. restart middlewared from CLI

PR #5696, opened 2020-09-20
NAS-107639 / 21.02 / Set default compression to zstandard

Ornias1993:
This is a draft PR, investigating if there are any side effect of setting zstandard as the default compression algorithm on pool creation (which is, by default, inherited by created datasets).

Notes:
- I'm not very sure about the need to manually set the featureflag in case of zstd

*edit, after testing and checking things*
- This shouldn't apply to the boot pool, as that one isn't created using the middleware afaik (middleware specifically lists these as being mounted under /mnt, which the bootpool isn't)
- Did some performance tests on an 2c/4t intel pentium CPU, 10gbe and 3x2disk mirrors. I don't think this is going to cause any significant slowdown on less-powerfull systems
- pools created from scratch on 12, aren't compatible with 12 anyhow in a lot of cases, so this wouldn't cause significant (new) compatibility issues (you can't import a pool created in TrueNAS 12/20.10 in 11 afaik)

Some actual data:
General note:
These numbers are supposed to be bottlenecked mostly by CPU and/or RAM. However they are not made to be a representation for performance, they are supposed to be used as a **comparative** reference only, meant to give a vague idea of performance delta. If you see weird results, it mostly is due to the test being CPU bottlenecked or test environment issues.

Allan Jude has done further research on performance here:
https://docs.google.com/spreadsheets/d/1TvCAIDzFsjuLuea7124q-1UtMd0C9amTgnXm2yPtiUQ/edit#gid=0

His number are more based on a real-world hardware scenario. I highly suggest going through them.

Note about random performance:
Random has been set to an 8K "extremely low" recordsize and mixed blocksizes.

---

Sequential Read:

image

image

---

Sequential Write

---
image
image

---

Sequential Read-Write

---
image

image

---

Random Read

---

image

image

---

Random Write

---
image
image

---

Random Read-Write

---

image
image

---
RAWR DATA
---
[Compression Algorithm Performance.xlsx](https://github.com/zfsonlinux/zfs/files/4014521/Compression.Algorithm.Performance.xlsx)

[test_results_1577651989.terse.txt](https://github.com/zfsonlinux/zfs/files/4014520/test_results_1577651989.terse.txt)

Test script:
https://github.com/Ornias1993/zfs-compression-test

Note:
For the IO slides, I've been lazy and column K is actually iops instead of bandwidth ;)

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -649,7 +649,7 @@ class PoolService(CRUDService):
         formatted_disks = await self.middleware.call('pool.format_disks', job, disks)
 
         options = {
-            'feature@lz4_compress': 'enabled',
+            'feature@zstd_compress': 'enabled',
             'altroot': '/mnt',
             'cachefile': ZPOOL_CACHE_FILE,
             'failmode': 'continue',
@@ -658,7 +658,7 @@ class PoolService(CRUDService):
         }
 
         fsoptions = {
-            'compression': 'lz4',
+            'compression': 'zstd',
             'aclinherit': 'passthrough',
             'mountpoint': f'/{data["name"]}',
             **encryption_dict


william-gr:
@freqlabs I will defer to the OS team about making it the default. Do you have any thoughts? Any drawbacks or problems you can foresee?
Do you think we should being in perf team as well?

Ornias1993:
> Do you think we should being in perf team as well?

I think it's wise to let perf have a look, yes.

yocalebo:
Definitely have to get perf team involved. I know that zstd can perform better in the compression scenario but that's only half the battle. The decompression is what I'd be worried about. I think zstd is a little slower than lz4? It's been a hot minute since I last read through code and looked at other users benchmark graphs.

kmoore134:
Yea, everything I've read says that there is a performance hit to reads with zstd. This page says its 1/2 of what lz4 can do:

https://facebook.github.io/zstd/

So this may not be best as a default right now until we can confirm there is no significant hit to real-world performance.

Ornias1993:
@yocalebo I'll make sure to get my test graphs into this PR. They have almost disappeared into the disarray that is github comments.

@kmoore134 ZFS performance of ZSTD and LZ4 is not in comparable with the stats the great folks at facebook/zstd provide.
(both LZ4 and ZSTD perform significant differently when used with ZFS)

TLDR:
I'll get you guys some actual data to look at instead of blablabla.... ;)

yocalebo:
@Ornias1993 great, thanks! However, keep in mind, I come from our enterprise side of the business so I'm mostly interested in how it will effect a real-world scenario. Particularly these 2:

1. using iscsi on a large striped mirror zpool with vmware as a hypervisor running lots of VMs
2. using SMB natively with 100's of unique users slamming the box with random IO

I don't expect you to be able to reproduce the real world scenarios and will probably only do synthetic benchmarks, but that's where my head is at wrt to changing the default compression 👍

Ornias1993:
Everyone:
I added the graphs and data I made earlier, current version is slightly faster though.

@yocalebo I 100% understand and agree.
And yes I would only be able to do synthetic tests, but I did actually execute high-io-load tests, by using ramdisks. :)

yocalebo:
@Ornias1993 brilliant, thanks 👍

Ornias1993:
Ohh, almost forgot... This is the scripting we used to test, also compatible with FreeBSD:
https://github.com/Ornias1993/zfs-compression-test

freqlabs:
> @freqlabs I will defer to the OS team about making it the default. Do you have any thoughts? Any drawbacks or problems you can foresee?
> Do you think we should being in perf team as well?

I would definitely like to give this more time for extensive testing and performance evaluation.

Ornias1993:
Update:
Some additional testing byond the previously executed performance tests, has been done by Allan Jude (who was also involved with the previous tests). This time with slightly varying test setups to verify any gains:

https://docs.google.com/spreadsheets/d/1TvCAIDzFsjuLuea7124q-1UtMd0C9amTgnXm2yPtiUQ/edit#gid=0

Ornias1993:
Slight update about performance:
Updating ZSTD to 1.4.7/1.4.8 (1.4.8 being an hotfix) would give a significant increase when doing smaller blocksize writes.

PR for it is up:
https://github.com/openzfs/zfs/pull/11367

adamdmoss:
From the graphs I see that 'random write' and 'sequential read-write' show zstd-fast being significantly better (than compression=off) at throughput and IOPS despite achieving no compression.  This seems anomalous and worthy of investigation.

Ornias1993:
@adamdmoss Updated research has already been done by Allan Jude.
It's also not part of this PR really, testing was done half a year ago... and as said in the original testing writedown: It's just meant to give an impression, not definitive results.

Compression=off was not really something that was relevant of the tests back in the day (it was mainly meant to illustrate GZIP-ZSTD-LZ4), i'm also not happy with the results when it comes to compression=off, but as I said: Further research has already been done by others.

Ornias1993:
As there are currently minor issues with ZSTD and it can't be updated upstream in OpenZFS because they want to delay any ZSTD updates waiting on some features that might or might not get added to OpenZFS in years. So i'm going to close this issue instead of keeping it open indefinately.

PR #5704, opened 2020-09-21
NAS-107643 / 20.10 / add glusterd service api

yocalebo:


william-gr (approved):
LGTM but lets wait Vladimir review as well

themylogin (approved):

PR #5710, opened 2020-09-23
NAS-107679 / 20.10 / Allow setting zero for special_small_block_size

sonicaj:
This commit introduces a change where middleware does not accept zero as a valid value for special_small_block_size whereas zero is the default value for the field.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2766,7 +2766,7 @@ class PoolDatasetService(CRUDService):
         Int('refquota_critical', validators=[Range(0, 100)]),
         Int('reservation'),
         Int('refreservation'),
-        Int('special_small_block_size', validators=[Range(min=512)]),
+        Int('special_small_block_size'),
         Int('copies'),
         Str('snapdir', enum=['VISIBLE', 'HIDDEN']),
         Str('deduplication', enum=['ON', 'VERIFY', 'OFF']),
@@ -3101,6 +3101,15 @@ class PoolDatasetService(CRUDService):
             for i in ('force_size', 'sparse', 'volsize', 'volblocksize'):
                 if i in data:
                     verrors.add(f'{schema}.{i}', 'This field is not valid for FILESYSTEM')
+
+            c_value = data.get('special_small_block_size')
+            if 'special_small_block_size' in data and not (
+                c_value == 0 or 512 <= data['special_small_block_size'] <= 1048576 or c_value % 512 == 0
+            ):
+                verrors.add(
+                    f'{schema}.special_small_block_size',
+                    'This field can be 0 or multiple of 512, up to 1048576'
+                )
         elif data['type'] == 'VOLUME':
             if mode == 'CREATE' and 'volsize' not in data:
                 verrors.add(f'{schema}.volsize', 'This field is required for VOLUME')


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2766,7 +2766,7 @@ def transform(dataset):
         Int('refquota_critical', validators=[Range(0, 100)]),
         Int('reservation'),
         Int('refreservation'),
-        Int('special_small_block_size', validators=[Range(min=512)]),
+        Int('special_small_block_size'),
```
yocalebo, line 2769: Is there a max size?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3101,6 +3101,14 @@ async def __common_validation(self, verrors, schema, data, mode):
             for i in ('force_size', 'sparse', 'volsize', 'volblocksize'):
                 if i in data:
                     verrors.add(f'{schema}.{i}', 'This field is not valid for FILESYSTEM')
+
+            if 'special_small_block_size' in data and not (
+                data['special_small_block_size'] == 0 or 512 <= data['special_small_block_size'] <= 1000000
```
yocalebo, line 3106: I think this line will always return true for any value <= 1000000. The validation error says it can only be 0 and any value above 512?

Maybe write it like `size == 0 or size > 512 or size <= 1000000`

yocalebo: also, maybe make a special_small_block_max_size variable and use that instead of a number?

sonicaj: I am not sure if this `size > 512 <= 1000000` will work ?
We should just change the text

yocalebo: >>> a
512
>>> a == 0 or a > 512 <= 1000000
False

sonicaj: Ah that's a mistake on my end wrt text. However `512 <= size <= 1000000` is still a valid expression.

yocalebo: Yep, I understand now that you've changed text 👍

sonicaj:
```
a = 1000001
a > 512 <= 1000000
```
This says true whereas we want it to be false

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3101,6 +3101,15 @@ async def __common_validation(self, verrors, schema, data, mode):
...
+            ):
+                verrors.add(
+                    f'{schema}.special_small_block_size',
+                    'Valid values for this field are 0 or an integer value above '
```
yocalebo, line 3110: Sorry to be OCD, but there is 1 last minor issue with the validation text. Using the correct mathematical grammar will help clear this up.

`Valid values for this field are 0 or greater than or equal to 512 or less than or equal to 100000`

src/middlewared/middlewared/plugins/pool.py
yocalebo: I know, the verbiage confused me 😉 

My brain just likes writing things out explicitly (i.e. `a == 0 or a >= 512 and a <= 1000000`) but your way works too 👍

yocalebo (approved):
Thanks!

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3101,6 +3101,14 @@ async def __common_validation(self, verrors, schema, data, mode):
             for i in ('force_size', 'sparse', 'volsize', 'volblocksize'):
                 if i in data:
                     verrors.add(f'{schema}.{i}', 'This field is not valid for FILESYSTEM')
+
+            if 'special_small_block_size' in data and not (
+                data['special_small_block_size'] == 0 or 512 <= data['special_small_block_size'] <= 1048576
+            ):
+                verrors.add(
+                    f'{schema}.special_small_block_size',
+                    'Valid values for this field are 0 or greater than or equal to 512 or less than or equal to 1048576'
```
william-gr, line 3110: This does not read very well.

Maybe:

This field can be 0 or multiple of 512, up to 1M?

Thoughts @freenas/docs-team  @Mrt134

themylogin (approved):

william-gr (approved):

PR #5712, opened 2020-09-23
NAS-107682 / 20.10 / fix regression in network general on SCALE HA

yocalebo:
I fixed 2 or 3 issues and optimized the network general section in commit: 38cd173fc2beb166aa7c2d2e0b0839e22ebf021d

However, I introduced 3 subtle regressions.

1. new_config was not being updated with the passed in data
2. domains key was not normalized before being written to db
3. netwait_ip key was not normalized before being written to db

This fixes all of the above issues.

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -214,7 +214,7 @@ class NetworkConfigurationService(ConfigService):
         Discovery support.
         """
         config = await self.config()
-        new_config = config.copy()
+        new_config = data.copy()
 
         verrors = await self.validate_general_settings(data, 'global_configuration_update')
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -214,7 +214,7 @@ async def do_update(self, data):
         Discovery support.
         """
         config = await self.config()
-        new_config = config.copy()
+        new_config = data.copy()
```
sonicaj, line 217: I think we should be doing
```
new_config = config.copy()
new_config.update(data)
```
What this gives us is that during validation we will have clear expectation of keys existing which should be present in the config if maybe we wanted to see what value a field has and based on that validate another field.
Also this is how it's been done in almost all services I think so anybody working would assume we would have the complete config available in `new_config`.

sonicaj (approved):
Thank you

PR #5713, opened 2020-09-23
NAS-107684 / 20.10 / Remove random sleeps in k8s lifecycle management

sonicaj:
1) Removes random sleep which may/may not work on every machine on first initialisation of k3s after the service is started.
2) Removes random sleep waiting for kube-router routing table to appear.
3) Removes sleep after starting docker daemon as in the updated mirror, based on tests done systemd docker unit properly waits for docker daemon to initialise itself instead of us waiting explicitly.
4) Make sure that `kube-bridge` is deleted when k8s is shutdown.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
@@ -112,9 +112,6 @@ class KubernetesService(Service):
     async def status_change_internal(self):
         await self.validate_k8s_fs_setup()
         await self.middleware.call('service.start', 'docker')
-        # This is necessary because docker daemon requires a couple of seconds after starting to initialise itself
-        # properly, if we try to load images without the delay that will fail and will only correct after a restart
-        await asyncio.sleep(5)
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
```diff
@@ -116,9 +112,6 @@ def status_change(self):
     async def status_change_internal(self):
         await self.validate_k8s_fs_setup()
         await self.middleware.call('service.start', 'docker')
```
themylogin, line 114: What ensures that docker daemon is properly initialized now?

themylogin (approved):

src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
sonicaj: It seems the latest update of our mirror has fixed that with updated packages. I ran some tests and this call only returns when the daemon has properly initialised itself.

PR #5714, opened 2020-09-24
NAS-107693 / 20.10 / fix alerts running on passive node on HA

yocalebo:
Fix 3 issues:

1. use the `system.is_enterprise` API call on SCALE
2. use `!= 'MASTER'` instead of `== 'BACKUP'`
3. remove an unused `is_freenas` variable

### DIFF
--- a/src/middlewared/middlewared/plugins/alert.py
+++ b/src/middlewared/middlewared/plugins/alert.py
@@ -172,7 +172,6 @@ class AlertService(Service):
 
     @private
     async def load(self):
-        is_freenas = await self.middleware.call("system.is_freenas")
 
         main_sources_dir = os.path.join(get_middlewared_dir(), "alert", "source")
         sources_dirs = [os.path.join(overlay_dir, "alert", "source") for overlay_dir in self.middleware.overlay_dirs]
@@ -194,10 +193,9 @@ class AlertService(Service):
 
     @private
     async def initialize(self, load=True):
-        is_freenas = await self.middleware.call("system.is_freenas")
 
         self.node = "A"
-        if not is_freenas:
+        if await self.middleware.call('system.is_enterprise'):
             if await self.middleware.call("failover.node") == "B":
                 self.node = "B"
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/alert.py
```diff
@@ -519,12 +517,9 @@ async def __should_run_or_send_alerts(self):
...
 
         if (
-            not await self.middleware.call('system.is_freenas') and
+            await self.middleware.call('system.is_enterprise') and
             await self.middleware.call('failover.licensed') and
-            (
-                await self.middleware.call('failover.status') == 'BACKUP' or
-                await self.middleware.call('failover.in_progress')
-            )
+            await self.middleware.call('failover.status') != 'MASTER'
```
themylogin, line 522: What if it is `SINGLE`?

Also I wanted to do the same check but @william-gr had some serious objections against it, let's revise them once again.

themylogin (changes requested):

src/middlewared/middlewared/plugins/alert.py
yocalebo: `SINGLE` is returned if there are:
1. no HA network interfaces configured or
2. there are no zpools imported

We shouldn't run alerts in those scenarios either since HA won't be functional. I'll add @william-gr on here too so he can review 👍

william-gr: The reason for that IIRC is that we have other statuses to account for.

The bottomline is that I am very agaisnt chaging things for the sake of change. If we have to change something it needs a very strong/undeniable argument, because much of our codebase has a lot of gotchas with uncommented code and no regression/integration/unit tests for any of it. ;)

yocalebo: This is causing database corruption. The `system_alert` table is different between the controllers.

william-gr: What causes database corruption is the database sync routines, not how alerts are run

yocalebo: Maybe they do, but not in this scenario. The `flush_alerts` method is run every 5 mins and deletes all entries in the `system_alert` table and then goes through all the alerts in memory and rewrites them but only does this `if failover.status == 'BACKUP'`. If failover.status == 'ERROR', then it will run.

yocalebo: Waiting on more collateral internally before doing anything else. Will open fresh PR as needed.

yocalebo:
Waiting on more collateral internally before doing anything else. Will open fresh PR as needed.

PR #5720, opened 2020-09-26
NAS-107723 / 20.10 / errno doesn't have EBADRPC attribute so use EHOSTUNREACH

yocalebo:


themylogin (commented):
There are two EBADRPC usages in this file.

yocalebo:
> There are two EBADRPC usages in this file.

good catch, fixed

themylogin (approved):

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/remote.py
```diff
@@ -116,7 +116,7 @@ def call(self, *args, **kwargs):
         try:
             if not self.connected.wait(timeout=20):
                 if self.remote_ip is None:
-                    raise CallError('Unable to determine remote node IP', errno.EBADRPC)
+                    raise CallError('Unable to determine remote node IP', errno.EHOSTUNREACH)
```
william-gr, line 119: Just making sure we dont look for that anywhere else in either middleware or UI?

I imagine we dont have that in Linux but we do in BSD?

yocalebo: @william-gr correct. Linux doesn't have EBADRPC, FreeBSD does. I didn't see any reference in middleware side but I'll verify webUI.

yocalebo: @william-gr verified that webUI doesn't reference it either.

william-gr: thanks for verifying!

PR #5724, opened 2020-09-28
NAS-107609 / 12.0 / Restart NFS after setting up network if it is bound to specific IP

themylogin:


william-gr (commented):
Took me a big while to review this because I was unsure how to proceed.

Seems like this is too specific to NFS and we want it to be more generic.

Every service that can bind to IPs will have the same problem.
We should probably come up with a way to verify that upfront so we can warn the user a service will be restarted.

Thoughts?

This also seems like out of scope for 12.0.

themylogin:
@william-gr the problem here is that rc.conf is generated before ix-netif, so the alert was always present. NFS is the only service where we require static IP to exist explicitly. For other services we don't do this check  and that's why this problem does not exist.

themylogin:
> We should probably come up with a way to verify that upfront so we can warn the user a service will be restarted.

When do we want to warn him? Every time he specifies a static IP to listen to? :)

william-gr:
> When do we want to warn him? Every time he specifies a static IP to listen to? :)

No, when he makes changes to network interfaces, before applying them.

william-gr:
> @william-gr the problem here is that rc.conf is generated before ix-netif, so the alert was always present. NFS is the only service where we require static IP to exist explicitly. For other services we don't do this check and that's why this problem does not exist.

That might be, but we still allow other services to bind to specific IPs, no?

What will happen if that IP is changed/removed?

william-gr (approved):

PR #5726, opened 2020-09-28
NAS-107724 / 20.10 / make sure keepalived starts properly

yocalebo:
This fixes a few issues.

1. call_sync was being made in main thread
--fixed by removing `vrrp_hook_license_update` because if a system is licensed HA for the first time, then there will be no failover interfaces configured which means that portion of code is a NO-OP

2. don't reload keepalived in the `failover` pseudo service
--unnecessary on SCALE HA

3. generate etc and start/reload the keepalived service on interface configuration

4. don't generate keepalived config if no `failover_virtual_aliases` are found

5. hook_setup_ha wasn't running on SCALE HA because `failover_vhid` never gets set. Instead use `failover_virtual_aliases`

### DIFF
--- a/src/middlewared/middlewared/plugins/interface/configure.py
+++ b/src/middlewared/middlewared/plugins/interface/configure.py
@@ -141,6 +141,9 @@ class InterfaceService(Service):
 
         # Remove addresses configured and not in database
         for addr in addrs_configured:
+            # keepalived service is responsible for deleting the VIP
+            if str(addr.address) == vrrp_vip:
+                continue
             if has_ipv6 and str(addr.address).startswith('fe80::'):
                 continue
             if addr not in addrs_database:
@@ -170,10 +173,17 @@ class InterfaceService(Service):
                 iface.carp_config = [netif.CarpConfig(carp_vhid, advskew=advskew, key=carp_pass.encode())]
         else:
             if vrrp_vip:
+                if not self.middleware.call_sync('service.started', 'keepalived'):
+                    self.middleware.call_sync('service.start', 'keepalived')
+                else:
+                    self.middleware.call_sync('service.reload', 'keepalived')
                 iface.vrrp_config = self.middleware.call_sync('interfaces.vrrp_config', name)
 
         # Add addresses in database and not configured
         for addr in (addrs_database - addrs_configured):
+            # keepalived service is responsible for adding the VIP
+            if str(addr.address) == vrrp_vip:
+                continue
             self.logger.debug('{}: adding {}'.format(name, addr))
             iface.add_address(addr)
 
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
@@ -1489,7 +1489,7 @@ async def hook_setup_ha(middleware, *args, **kwargs):
     if not await middleware.call('failover.licensed'):
         return
 
-    if not await middleware.call('interface.query', [('failover_vhid', '!=', None)]):
+    if not await middleware.call('interface.query', [('failover_virtual_aliases', '!=', None)]):
         return
 
     if not await middleware.call('pool.query'):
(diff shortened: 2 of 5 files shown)


src/middlewared/middlewared/plugins/interface/configure.py
```diff
@@ -144,14 +146,17 @@ def configure(self, data, aliases, wait_dhcp=False, options=None):
             if has_ipv6 and str(addr.address).startswith('fe80::'):
                 continue
             if addr not in addrs_database:
+                # keepalived is responsible for removing the VIP on SCALE HA
+                if vrrp_vip is not None and str(addr.address) in vrrp_vip:
```
william-gr, line 150: Should that be an "==" ?
10.0.0.1 in 10.0.0.10 would be True

src/middlewared/middlewared/plugins/interface/configure.py
```diff
@@ -170,10 +175,17 @@ def configure(self, data, aliases, wait_dhcp=False, options=None):
                 iface.carp_config = [netif.CarpConfig(carp_vhid, advskew=advskew, key=carp_pass.encode())]
         else:
             if vrrp_vip:
+                if not self.middleware.call_sync('service.started', 'keepalived'):
+                    self.middleware.call_sync('service.start', 'keepalived')
+                else:
+                    self.middleware.call_sync('service.reload', 'keepalived')
```
william-gr, line 181: is that going to reload/start keepalived for every interface?

src/middlewared/middlewared/plugins/interface/configure.py
```diff
@@ -170,10 +175,17 @@ def configure(self, data, aliases, wait_dhcp=False, options=None):
...
                 iface.vrrp_config = self.middleware.call_sync('interfaces.vrrp_config', name)
 
         # Add addresses in database and not configured
         for addr in (addrs_database - addrs_configured):
+            # keepalived is responsible for adding the VIP on SCALE HA
+            if vrrp_vip is not None and str(addr.address) in vrrp_vip:
```
william-gr, line 187: Same here

src/middlewared/middlewared/plugins/interface/configure.py
```diff
@@ -210,7 +222,7 @@ def configure(self, data, aliases, wait_dhcp=False, options=None):
             self.logger.debug('Starting dhclient for {}'.format(name))
             self.middleware.call_sync('interface.dhclient_start', data['int_interface'], wait_dhcp)
 
-        if osc.IS_FREEBSD:
+        if not linux:
```
william-gr, line 225: Keeping osc.IS_FREEBSD makes it easy for us to spot places to remove code blocks later on if needed.

Just personal opinion, no need to change this time/PR

yocalebo: I agree, i'll change back since there is no reason for me to make it inconsistent.

src/middlewared/middlewared/plugins/interface/configure.py
yocalebo: Further testing shows that keepalived doesn't care if middlewared or itself adds/removes the IP so I've removed this check.

src/middlewared/middlewared/plugins/interface/configure.py
yocalebo: Changed it back to be consistent with how we do it everywhere else.

src/middlewared/middlewared/plugins/interface/configure.py
yocalebo: It shouldn't, only reloads the service if there is a change to the `vrrp_vip` attribute.

william-gr (approved):

yocalebo:
@william-gr I do need to prevent `interface.configure` from adding/deleting the VIP on SCALE HA because we will assign the IP address to each controller on boot but since keepalived didn't add the interface on startup, it won't remove it.

william-gr (approved):

PR #5733, opened 2020-09-28
NAS-107674 / 20.10 / Less garbage in logs when hardware faults lead to disk.temperatures timeout

themylogin:


william-gr (commented):
Have we figured out why it timed out? Does he has many disks and/or some of them are too slow?

Do we have a timeout per disk so not all of it fails?

themylogin:
> Have we figured out why it timed out?

It happened a few times a day at the same moment of hour, I suspect there are conflicting concurrent tasks.

> Does he has many disks and/or some of them are too slow?

No, a few disks, some of them have SMART errors

> Do we have a timeout per disk so not all of it fails?

Added

william-gr (approved):
Thanks!

PR #5742, opened 2020-09-29
NAS-107686 / 20.10 / Clean paths from trailing slashes

themylogin:
@william-gr I guess in 12.0 backport we only do `rstrip("/")` for the paths passed to the NFS share plugin itself?


william-gr (approved):
Agreed

PR #5746, opened 2020-09-30
NAS-107770 / 20.10 / fix failover.vip.check_states on SCALE HA

yocalebo:


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/virtual_ips_linux.py
```diff
@@ -95,8 +95,6 @@ async def check_states(self, local, remote):
             errors.append('There are no failover interfaces')
 
         for name in interfaces:
-            if name not in local[0] and name in remote[0]:
-                errors.append(f'Interface "{name}" is MASTER on standby node')
```
william-gr, line 99: Could you elaborate? The logic looks right to me, is that a different behavior with keepalived?

yocalebo: Using VRRP, there is no "MASTER" or "BACKUP" keyword that gets placed on the interface. Either the VIP exists on the interface or it does not.

Since this method has no notion of which node is MASTER/BACKUP, it will always return as error.

william-gr: Ah, makes sense, thanks

william-gr (approved):

PR #5749, opened 2020-09-30
NAS-107779 / 20.10 / dont run HA code on non-HA systems

yocalebo:


src/middlewared/middlewared/etc_files/keepalived.conf.mako
```diff
@@ -8,6 +8,8 @@
             ' Keepalived config not generated.'
         )
 	raise FileShouldNotExist()
+    else:
```
william-gr, line 11: This will make it never be generated...

yocalebo: Oops

william-gr (approved):

PR #5750, opened 2020-09-30
NAS-107765 / 20.10 / fix generating ganesha.conf

yocalebo:
Multiple issues with ganesha.conf generation.

1. ganesha supports a per export protocol so specifying the protocol for each export is required
2. the `Tag` parameter cannot start with a leading `/`
3. the `Tag` parameter is ignored if NFSv4 is enabled
4. the `Tag` parameter does not need to be added unless an alias was given to us
5. if NFSv4 is enabled and there are no aliases, then default to making the `Pseudo` parameter the last part of the `path`
6. the formatting was off and made the ganesha.conf almost unreadable so fix that up

### DIFF
--- a/src/middlewared/middlewared/etc_files/ganesha/ganesha.conf.mako
+++ b/src/middlewared/middlewared/etc_files/ganesha/ganesha.conf.mako
@@ -15,59 +15,44 @@
     export_id = itertools.count(1)
 %>
 
-NFS_CORE_PARAM
-{
+NFS_CORE_PARAM {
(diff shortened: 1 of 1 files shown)


themylogin (approved):
I always wonder, why do we make config file readable instead of making our source code readable?

yocalebo:
> I always wonder, why do we make config file readable instead of making our source code readable?

Because "config file readable" vs "our source code readable" is subjective. 😄

william-gr (approved):

PR #5755, opened 2020-10-01
NAS-107811 / 20.10 / Do not break logging setup on fresh install first boot

themylogin:


src/middlewared/middlewared/logger.py
```diff
@@ -279,6 +279,7 @@ def stream(self):
     def _set_output_file(self):
         """Set the output format for file log."""
         try:
+            os.makedirs("/root/syslog", mode=0o755, exist_ok=True)
```
william-gr, line 282: This is already below, just need to move code around

william-gr (changes requested):

william-gr (approved):

PR #5758, opened 2020-10-01
Do not crash sed_unlock on SCALE

themylogin:


src/middlewared/middlewared/plugins/disk.py
```diff
@@ -319,6 +319,9 @@ async def sed_unlock(self, disk_name, disk=None, _advconfig=None):
 
         rv = {'name': disk_name, 'locked': None}
 
+        if osc.IS_LINUX:
```
william-gr, line 322: Shouldnt that be after sedutil-cli?

themylogin: @william-gr do you want me to make sedutil debian package now then?

william-gr: No, I thought we already had sedutil and only camcontrol was failing

william-gr: Not sure whats the point in "fixing" this, then? if nothing is wokring anyway?

william-gr: I think Waqar already has a ticket to make SED work

themylogin: There is a report in community slack, pool.format_disks always calls sed_unlock_all and the crash prevents disks from formatting

themylogin: Sorry, ticket about what?

william-gr: Thought we had a ticket about wiping disk not working. If thats not the case, nvm me

themylogin: You have not approved the ticket, is there anything else I am missing?

william-gr (approved):
No, simply forgot :)

PR #5765, opened 2020-10-02
NAS-107831 / 20.10 / fix multiple issues with failover on SCALE HA

yocalebo:
A few issues have been fixed:
1. signal is not thread safe so use a thread to export zpools

Finally, I'm able to test failover on real hardware, so I've identified some key differences between VRRP and CARP.

2. MASTER/BACKUP events are being processed by each controller so quickly that receiving an event and then checking if the other controller is MASTER is actually returning true...when it should not. So this is a race and that check has been removed. The `vrrp_master` and `vrrp_backup` jobs already have the necessary validation on whether or not the failover event should be
processed.
3. if we receive a MASTER event but all the zpools are already imported, then there is nothing to do so ignore.
4. get rid of some variables that weren't adding any value
5. only export the zpools if the status `!= 'OFFLINE'`
6. be sure and catch and re-raise IgnoreFailoverEvent in `run_call` to prevent a failover event loop since we restart the keepalived service (which triggers another event)

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
@@ -3,7 +3,7 @@ import os
 import time
 import contextlib
 import shutil
-import signal
+import threading
 import logging
 from collections import defaultdict
 
@@ -38,16 +38,6 @@ class AllZpoolsFailedToImport(Exception):
     """
     This is raised if all zpools failed to
     import when becoming master.
-    """
-    pass
-
-
-class ZpoolExportTimeout(Exception):
-
-    """
-    This is raised if we can't export the
-    zpool(s) from the system when becoming
-    BACKUP
     """
     pass
 
@@ -141,6 +131,9 @@ class FailoverService(Service):
     def run_call(self, method, *args):
         try:
             return self.middleware.call_sync(method, *args)
+        except IgnoreFailoverEvent:
+            # `self.validate()` calls this method
+            raise
         except Exception as e:
             logger.error('Failed to run %s:%r:%s', method, args, e)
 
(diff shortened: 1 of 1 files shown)


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
```diff
@@ -157,9 +150,23 @@ def event(self, ifname, event):
...
+                    logger.info('Exported "%s"', vol)
+        except Exception as e:
+            # catch any exception that could be raised
+            # We sleep for 5 seconds here because this is
+            # in its own thread. The calling thread waits
+            # for self.ZPOOL_EXPORT_TIMEOUT and if this
+            # thread is_alive(), then we violently reboot
+            # the node
+            logger.error('Error exporting "{}" with error {}'.format(vol, e))
+            time.sleep(5)
```
william-gr, line 169: Should we sleep ZPOOL_EXPORT_TIMEOUT + 1?

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
```diff
@@ -157,9 +150,23 @@ def event(self, ifname, event):
...
+                    self.middleware.call_sync('zfs.pool.export', vol, {'force': True})
+                    logger.info('Exported "%s"', vol)
+        except Exception as e:
+            # catch any exception that could be raised
+            # We sleep for 5 seconds here because this is
+            # in its own thread. The calling thread waits
+            # for self.ZPOOL_EXPORT_TIMEOUT and if this
+            # thread is_alive(), then we violently reboot
+            # the node
+            logger.error('Error exporting "{}" with error {}'.format(vol, e))
```
william-gr, line 168: Lets use %s here as thats the proper formatting for logger

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
yocalebo: Absolutely should, I'll fix.

william-gr (approved):

PR #5767, opened 2020-10-04
Kubernetes Catalog Support

sonicaj:


src/middlewared/middlewared/plugins/catalogs_linux/utils.py
```diff
@@ -0,0 +1,61 @@
...
+        os.makedirs(parent_dir, exist_ok=True)
+        destination = os.path.join(parent_dir, convert_repository_to_path(repository_uri))
+        repo = get_repo(destination)
+        clone_repo = not bool(repo)
+        if repo:
+            # We will try to checkout branch and do a git pull, if any of these operations fail, we will
+            # clone the repository again. Why they might fail is if user has been manually playing with the repo
+            try:
+                for action, params in (('checkout', [branch]), ('pull', [])):
+                    getattr(repo.git, action)(*params)
```
william-gr, line 43: This is a good obfuscation of code IMO :P
repo.git.checkout(branch)
repo.git.pull()

Has the same amount of lines and less obfuscated

src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
```diff
@@ -0,0 +1,238 @@
...
+            Str('item', required=True),
+            Str('release_name', required=True),
+            Str('train', default='charts'),
+            Str('version', required=True),
+        )
+    )
+    async def do_create(self, data):
+        await self.middleware.call('kubernetes.validate_k8s_setup')
+        if await self.middleware.call('chart.release.query', [['id', '=', data['release_name']]]):
+            raise CallError(f'Chart release with {data["release_name"]} already exists.')
```
william-gr, line 106: Is there a errno.EEXIST?

src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
```diff
@@ -0,0 +1,238 @@
...
+
+        catalog = await self.middleware.call(
+            'catalog.query', [['id', '=', data['catalog']]], {'get': True, 'extra': {'item_details': True}}
+        )
+        if data['train'] not in catalog['trains']:
+            raise CallError(f'Unable to locate "{data["train"]}" catalog train.')
```
william-gr, line 112: ENOENT

william-gr (commented):
Didnt go into much of the specifics but looking pretty good so far

themylogin (changes requested):

src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
```diff
@@ -0,0 +1,223 @@
...
+            await self.perform_actions(context)
+
+            with tempfile.NamedTemporaryFile(mode='w+') as f:
+                f.write(yaml.dump(new_values))
+                f.flush()
+                # We will install the chart now and force the installation in an ix based namespace
+                # https://github.com/helm/helm/issues/5465#issuecomment-473942223
+                cp = await run(
+                    [
+                        'helm', 'install', data['release_name'], chart_path, '-n',
```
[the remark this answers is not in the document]
sonicaj: So this will just deploy the workload in the k8s cluster, the actual pulling of images will happen automatically when k8s accepts pods and schedules them on the node, they will remain in `pending` state until all the required images have been pulled.
I have been thinking of using events for notifications in this case and let install proceed without actually waiting for the pods to come online

src/middlewared/middlewared/plugins/chart_releases_linux/rollback.py
```diff
@@ -0,0 +1,87 @@
...
+        if rollback_version not in release['history']:
+            raise CallError(f'Unable to find {rollback_version!r} item version in {release_name!r} history')
+
+        chart_path = os.path.join(release['path'], 'charts', rollback_version)
+        if not await self.middleware.run_in_thread(lambda: os.path.exists(chart_path)):
+            raise CallError(f'Unable to locate {chart_path!r} path for rolling back')
+
+        ix_volumes_ds = os.path.join(release['dataset'], 'volumes/ix_volumes')
+        snap_name = f'{ix_volumes_ds}@{rollback_version}'
+        if not await self.middleware.call('zfs.snapshot.query', [['id', '=', snap_name]]) and not options['force']:
```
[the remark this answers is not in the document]
sonicaj: When using just `id` filter to retrieve a single snap, `py-libzfs` is faster.
```
 ⚡ root@truenas  ~  time midclt call zfs.snapshot.query '[["id", "=", "evo/test@snap10000"]]' > /dev/null
midclt call zfs.snapshot.query '[["id", "=", "evo/test@snap10000"]]' >   0.06s user 0.01s system 90% cpu 0.081 total
 ⚡ root@truenas  ~ 
 ⚡ root@truenas  ~ 
 ⚡ root@truenas  ~  time midclt call zfs.snapshot.query '[["id", "=", "evo/test@snap10000"]]' '{"select": ["name"]}' > /dev/null
midclt call zfs.snapshot.query '[["id", "=", "evo/test@snap10000"]]'  >   0.07s user 0.00s system 91% cpu 0.085 total
```

src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
```diff
@@ -0,0 +1,284 @@
...
+            job.set_progress(100, 'Chart release created')
+            return await self.get_instance(data['release_name'])
+
+    @accepts(
+        Str('chart_release'),
+        Dict(
+            'chart_release_update',
+            Dict('values', additional_attrs=True),
+        )
+    )
```
william-gr, line 188: Create and delete are jobs, why not update?

src/middlewared/middlewared/plugins/chart_releases_linux/upgrade.py
```diff
@@ -0,0 +1,119 @@
...
+        config.update(options['values'])
+
+        config, context = await self.middleware.call(
+            'chart.release.normalise_and_validate_values', catalog_item, config, False, release['dataset'],
+        )
+
+        # We have validated configuration now
+
+        chart_path = os.path.join(release['path'], 'charts', new_version)
+        await self.middleware.run_in_thread(lambda: shutil.rmtree(chart_path, ignore_errors=True))
```
william-gr, line 92: lambda is not necessary I think...

`await self.middleware.run_in_thread(shutil.rmtree, chart_path, ignore_errors=True)`

src/middlewared/middlewared/plugins/chart_releases_linux/upgrade.py
```diff
@@ -0,0 +1,119 @@
...
+        namespace = 'chart.release'
+
+    @accepts(
+        Str('release_name'),
+        Dict(
+            'upgrade_options',
+            Dict('values', additional_attrs=True),
+            Str('item_version', required=True),
+        )
+    )
```
william-gr, line 28: Seems like this should be a job?

william-gr (commented):
Great job, a few minor comments

src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
```diff
@@ -0,0 +1,284 @@
...
+    async def query(self, filters=None, options=None):
+        """
+        Query available chart releases.
+
+        `options.extra.retrieve_resources` is a boolean when set will retrieve existing kubernetes resources
+        in the chart namespace.
+
+        `options.extra.history` is a boolean when set will retrieve all chart version upgrades for a chart release.
+        """
+        return await self.middleware.call('chart.release.query_releases', filters, options)
```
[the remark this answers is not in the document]
sonicaj: For readability ? Although I have no issues having it here. We can have it here if you think that's better ?

src/middlewared/middlewared/plugins/chart_releases_linux/upgrade.py
sonicaj: We are not actually waiting for the upgrade to complete. We just tell k8s that we want the workloads to be in this state and it performs the upgrade itself in the background.
I have been thinking of using events to notify the user wrt upgrades / creation of chart releases.

src/middlewared/middlewared/plugins/catalogs_linux/sync_catalogs.py
```diff
@@ -0,0 +1,38 @@
...
+from middlewared.service import job, periodic, private, Service
+
+from .utils import pull_clone_repository
+
+
+class CatalogService(Service):
+
+    @accepts()
+    @periodic(interval=864000)
+    @job(lock='sync_catalogs')
```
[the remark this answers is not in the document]
sonicaj: I think it has a use case where users could update all catalogs with a single end point ? Thoughts ?

src/middlewared/middlewared/plugins/catalogs_linux/utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def convert_repository_to_path(git_repository_uri):
+    return git_repository_uri.split('://', 1)[-1].replace('/', '_').replace('.', '_')
+
+
+def clone_repository(repository_uri, destination, depth):
+    shutil.rmtree(destination, ignore_errors=True)
+    kwargs = {'env': os.environ.copy(), 'depth': depth}
+    return git.Repo.clone_from(repository_uri, destination, **{k: v for k, v in kwargs.items() if v})
```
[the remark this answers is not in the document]
sonicaj: The idea is to only pass it as a kwargs if we have some value for the key, though i can definitely improve / handle it better :)

src/middlewared/middlewared/plugins/chart_releases_linux/upgrade.py
william-gr: Still might make sense to make it a job, just t o be consistent between the CRUD methods

src/middlewared/middlewared/plugins/catalogs_linux/sync_catalogs.py
```diff
@@ -0,0 +1,38 @@
...
+    async def sync_catalogs(self, job):
+        """
+        Sync all available catalogs.
+        """
+        catalogs = await self.middleware.call('catalog.query')
+        for catalog in catalogs:
+            job.set_progress(100 / len(catalogs), f'Syncing {catalog["id"]} catalog')
```
william-gr, line 20: This math seems to be wrong

src/middlewared/middlewared/plugins/catalogs_linux/sync_catalogs.py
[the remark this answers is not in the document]
william-gr: You thinking a button in the UI to update all?

Is sync the right name then?

themylogin (dismissed):

william-gr (approved):

PR #5771, opened 2020-10-05
NAS-107845 / 20.10 / make device.get_disks better for SCALE HA

yocalebo:
On SCALE HA, I've noticed the following:

1. on a functional HA system, the webUI was taking 5+ seconds to present the login screen
2. the standby nodes console was being spammed with `reservation conflict` messages

I traced this all down to the `device.get_disks` method. This method was calling `lshw` which was taking 2+ seconds to run on a system with 53 SCSI disks and causing the `reservation conflict` messages. I haven't looked at that source code but it seemed to be "loading" all subsystems no matter if you chose the "DISK" type. It is also opening the underlying disks in "w" mode which is causing the reservation errors. Instead of running `lshw`, I've replaced it with `lsblk`. The only major difference is that I'm having to shell out and run `sg_vpd` to grab the rotational rate of the disk if the type is `HDD`. Even doing this, my local system is ~52% faster and if type is `SSD`, then it can be ~85%+ faster since we don't have to grab the `rotationrate` attribute.

I've removed the private `retrieve_lshw_disks_data` method as there are no public consumers of that method.

`device.get_disks` before the change = 2.3 seconds
`device.get_disks` after the change = 1.1 seconds

`failover.disabled_reasons` before the change = 2.5 seconds
`failover.disabled_reasons` after the change = 1.1 seconds

### DIFF
--- a/src/middlewared/middlewared/plugins/device_/device_info_linux.py
+++ b/src/middlewared/middlewared/plugins/device_/device_info_linux.py
@@ -3,14 +3,14 @@ import os
 import pyudev
 import re
 import subprocess
-
-from lxml import etree
+import json
 
 from .device_info_base import DeviceInfoBase
 from middlewared.service import private, Service
 from middlewared.utils import run
 
 RE_DISK_SERIAL = re.compile(r'Unit serial number:\s*(.*)')
+RE_DISK_ROTATION_RATE = re.compile(r'\s*Nominal\s*rotation\s*rate:\s*(.*)')
 RE_GPU_VENDOR = re.compile(r'description:\s*VGA compatible controller[\s\S]*vendor:\s*(.*)')
 RE_NVME_PRIVATE_NAMESPACE = re.compile(r'nvme[0-9]+c')
 RE_SERIAL = re.compile(r'state.*=\s*(\w*).*io (.*)-(\w*)\n.*', re.S | re.A)
@@ -60,7 +60,7 @@ class DeviceService(Service, DeviceInfoBase):
 
     def get_disks(self):
         disks = {}
-        lshw_disks = self.retrieve_lshw_disks_data()
+        disks_data = self.retrieve_disks_data()
 
         for block_device in pyudev.Context().list_devices(subsystem='block', DEVTYPE='disk'):
             if block_device.sys_name.startswith(('sr', 'md', 'dm-', 'loop', 'zd')):
@@ -75,7 +75,7 @@ class DeviceService(Service, DeviceInfoBase):
             # nvme drives won't have this
 
             try:
-                disks[block_device.sys_name] = self.get_disk_details(block_device, self.disk_default.copy(), lshw_disks)
+                disks[block_device.sys_name] = self.get_disk_details(block_device, self.disk_default.copy(), disks_data)
             except Exception as e:
                 self.middleware.logger.debug(
                     'Failed to retrieve disk details for %s : %s', block_device.sys_name, str(e)
@@ -84,24 +84,28 @@ class DeviceService(Service, DeviceInfoBase):
         return disks
 
     @private
-    def retrieve_lshw_disks_data(self):
-        disks_cp = subprocess.Popen(
-            ['lshw', '-xml', '-class', 'disk'], stdout=subprocess.PIPE, stderr=subprocess.PIPE
+    def retrieve_disks_data(self):
+
+        lsblk_disks = {}
+        disks_cp = subprocess.run(
+            ['lsblk', '-OJdb'],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE
         )
-        output, error = disks_cp.communicate()
-        lshw_disks = {}
-        if output:
-            xml = etree.fromstring(output.decode())
-            for child in filter(lambda c: c.get('class') == 'disk', xml.getchildren()):
-                data = {'rotationrate': None}
-                for c in child.getchildren():
-                    if not len(c.getchildren()):
-                        data[c.tag] = c.text
-                    elif c.tag == 'capabilities':
-                        for capability in filter(lambda d: d.text.endswith('rotations per minute'), c.getchildren()):
-                            data['rotationrate'] = capability.get('id')[:-3]
-                lshw_disks[data['logicalname']] = data
-        return lshw_disks
+        if not disks_cp.returncode:
+            try:
+                lsblk_disks = json.loads(disks_cp.stdout)['blockdevices']
+                lsblk_disks = {i['path']: i for i in lsblk_disks}
+            except Exception as e:
+                self.middleware.logger.error(
+                    'Failed parsing lsblk information with error: %s', e
+                )
+        else:
+            self.middleware.logger.error(
+                'Failed running lsblk command with error: %s', disks_cp.stderr.decode()
+            )
+
+        return lsblk_disks
 
     def get_disk(self, name):
         disk = self.disk_default.copy()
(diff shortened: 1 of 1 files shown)


yocalebo:
I should add that we could make this exponentially faster by caching the disk information on middlewared setup, and then listening for uevent changes for add/remove on the disks and updating/invalidating the cache as necessary depending on the uevent. That would be the "proper" fix but this was a relatively easy stop-gap for now.

william-gr (approved):
We can probably wait to merge this after 20.10-ALPHA?

yocalebo:
> We can probably wait to merge this after 20.10-ALPHA?

yes, we can wait 👍

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -133,28 +172,29 @@ def get_disk_details(self, block_device, disk, lshw_disks):
...
 
+            # get model and serial
             disk['ident'] = disk['serial'] = disk_data.get('serial', '')
-            disk['size'] = disk['mediasize'] = int(disk_data['size']) if 'size' in disk_data else None
-            disk['descr'] = disk['model'] = disk_data.get('product')
+            disk['descr'] = disk['model'] = disk_data.get('model')
+
+            # get all relevant size attributes of disk
+            disk['sectorsize'] = disk_data['log-sec']
+            disk['size'] = disk_data['size']
```
sonicaj, line 189: I think we should still set `mediasize` for backwards compatibility on consumers consuming that key. We do that for freebsd side as well.

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -175,26 +215,26 @@ def get_disk_details(self, block_device, disk, lshw_disks):
                     disk['serial'] = disk['ident'] = reg.group(1)
 
         if not disk['model'] and os.path.exists(os.path.join(disk_sys_path, 'device/model')):
-            # For nvme drives, we are unable to retrieve it via lshw
```
sonicaj, line 178: This block is not required as `lsblk` retrieves the model data automatically for nvme drives

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -133,28 +172,29 @@ def get_disk_details(self, block_device, disk, lshw_disks):
...
+            disk['descr'] = disk['model'] = disk_data.get('model')
+
+            # get all relevant size attributes of disk
+            disk['sectorsize'] = disk_data['log-sec']
+            disk['size'] = disk_data['size']
             if disk['size'] and disk['sectorsize']:
                 disk['blocks'] = int(disk['size'] / disk['sectorsize'])
 
+            # get lunid
+            disk['lunid'] = disk_data['wwn'].lstrip('0x')
```
sonicaj, line 194: I am not sure about this. `lunid` for a nvme drive i have right now on my scale system is
```
0025385201416c20
```
`sg_vpd` gives
```
sg_vpd -i --quiet /dev/nvme0n1
  0x0025385201416c20
  eui.0025385201416C20
```

With current implementation we take the first id available. Now `lsblk` reports the second id shown in `sg_vpd` as `wwn`.
What I think will happen is that middleware will probably recognise the nvme as a new disk because it won't be able to translate `serial_lunid` to nvme disk name as the id would have changed for the nvme drive.

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -111,13 +114,49 @@ def get_disk(self, name):
...
+        rotation_rate = None
+
+        vpd = subprocess.Popen(
+            ['sg_vpd', '-p', 'bdc', device_path],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        out, err = vpd.communicate()
+        if not vpd.returncode:
+            out = out.decode().split('\n')[1].strip()
```
sonicaj, line 131: How about using regex or checking key name like `Nominal rotation rate` rather then relying on `1` index as it could probably change ? We are already checking `rotation rate` in if below but we might lose it here before as it might on a different index

src/middlewared/middlewared/plugins/device_/device_info_linux.py
sonicaj: I think removing `eui.` and having it lower case should be enough ( though we should confirm if it won't change in any other way )

src/middlewared/middlewared/plugins/device_/device_info_linux.py
yocalebo: oops, missed a key. good catch, I'll fix.

src/middlewared/middlewared/plugins/device_/device_info_linux.py
yocalebo: sg_vpd has a very stable API. If this was the case, then we would also need to do a regex for the lunid section. I'd rather leave this as is.

sonicaj: Why the if check below then ? ; )
My concern was that we do a check partially, so it shows we are not sure of the output, but doing the check there means we already have potentially lost the correct line.

yocalebo: Because `Nominal rotation rate:` doesn't get returned on SSD's.

yocalebo: Even though we're only doing this on HDD types, having that extra `if` is to make sure that `lsblk` identified the type correctly and passed us a HDD.

yocalebo: I don't like it, but to get your approval I've changed it to use a REGEX 👍

sonicaj: Thank you for clarifying and using regex @yocalebo : ) I was okay with it if you think it's good :p

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -174,27 +202,21 @@ def get_disk_details(self, block_device, disk, lshw_disks):
...
-        )
-        cp_stdout, cp_stderr = lun_id_cp.communicate()
-        if not lun_id_cp.returncode and lun_id_cp.stdout:
-            lunid = cp_stdout.decode().strip()
-            if lunid:
-                disk['lunid'] = lunid.split()[0]
-            if lunid and disk['lunid'].startswith('0x'):
-                disk['lunid'] = disk['lunid'][2:]
-
-        if disk['serial'] and disk['lunid']:
```
sonicaj, line 196: Can we please just have this and remove the if block at L183 in the new diff ? As if we get into last `if` block, we will miss setting `serial_lunid`.

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -133,28 +156,33 @@ def get_disk_details(self, block_device, disk, lshw_disks):
...
+
+            # get all relevant size attributes of disk
+            disk['sectorsize'] = disk_data['log-sec']
+            disk['size'] = disk['mediasize'] = disk_data['size']
             if disk['size'] and disk['sectorsize']:
                 disk['blocks'] = int(disk['size'] / disk['sectorsize'])
 
+            # get lunid
+            if disk['tran'] == 'nvme':
+                disk['lunid'] = disk_data['wwn'].lstrip('eui').lower()
```
sonicaj, line 179: We have the prefix `eui.` which we should be removing. Looking at `lsblk` output again, i don't think we need lower, `sg_vpd` identification flag returns them capitalised, lsblk seems to be doing the right thing following what we have been doing before.

sonicaj (approved):
Thank you

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -549,14 +549,14 @@ async def mismatch_disks(self):
         local_boot_disks = await self.middleware.call('boot.get_disks')
         remote_boot_disks = await self.middleware.call('failover.call_remote', 'boot.get_disks')
         local_disks = set(
-            v['ident']
-            for k, v in (await self.middleware.call('device.get_info', 'DISK')).items()
-            if k not in local_boot_disks
+            v['serial']
+            for k, v in (await self.middleware.call('device.retrieve_disks_data')).items()
```
sonicaj, line 553: It just came to me, `retrieve_disks_data ` this does not exist in core device service, are we looking for freebsd support in failover code for master ?

yocalebo: You're correct, I don't like calling `device.get_info 'DISK'` because it gets the rotationrate of every drive (by subprocessing out) which causes slowness. However, I don't see an easy way around it without making public API changes or doing an `if osc.IS_LINUX` check which is gross. I'll just go back to using the `get_info 'DISK'` API for now.

PR #5775, opened 2020-10-05
NAS-107819 / 12.0 / Match interfaces between FreeBSD and Linux

themylogin:


src/middlewared/middlewared/plugins/interface/link_address.py
```diff
@@ -0,0 +1,136 @@
...
+                                  {"vlan_pint": name})
+
+
+async def setup(middleware):
+    real_interfaces = RealInterfaceCollection(await middleware.call("interface.query", [["fake", "!=", True]]))
+    db_interfaces = DatabaseInterfaceCollection(
+        await middleware.call("datastore.query", "network.interfaces", [], {"prefix": "int_"}),
+    )
+
+    # Migrate BSD network interfaces to Linux
```
william-gr, line 81: This is has to live here for eternity, until we support migrating databases from FreeBSD?
Impossible to be part of a data migration?

themylogin: It can be a part of data migration, but I've decided to put it here as this process is coupled with `# Update link addresses for interfaces in the database` which is executed on both BSD and Linux. (It is not necessary in Linux, but I've decided to keep that field populated correctly as it might have some use in the future)

william-gr: Seems reasonable, but the way it is, if any exception happens in this case middlewared startup will crash, no?
Can/should we make it more resilient?

themylogin: Fixed

william-gr (approved):
Thanks

PR #5788, opened 2020-10-06
NAS-104374 / 12.0 / Use Google API to send mail via GMail

themylogin:


william-gr (approved):

src/middlewared/middlewared/plugins/mail_/gmail.py
```diff
@@ -0,0 +1,51 @@
...
+
+from middlewared.service import private, Service
+
+
+class GmailService:
+    def __init__(self, config):
+        self.config = config
+
+        credentials = Credentials.from_authorized_user_info(config["oauth"])
+        self.service = build("gmail", "v1", credentials=credentials)
```
sonicaj, line 14: Should we also be closing this ? i.e `self.service.close()` when we don't require the `GmailService` object anymore to avoid leaving sockets open ?

src/middlewared/middlewared/plugins/mail_/gmail.py
```diff
@@ -0,0 +1,51 @@
...
+        if gmail_service is None:
+            raise RuntimeError("GMail service is not initialized")
+
+        gmail_service.service.users().messages().send(userId="me", body={
+            "raw": base64.urlsafe_b64encode(message.as_string().encode("ascii")).decode("ascii"),
+        }).execute()
+
+
+async def setup(middleware):
+    await middleware.call("mail.gmail_initialize")
```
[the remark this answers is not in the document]
themylogin: Yes, it does not require network

src/middlewared/middlewared/plugins/mail_/gmail.py
themylogin: There is no persistent connection, this resource only handles refreshing OAuth access token periodically (only on request so there is no background thread as well)

sonicaj: Thank you for clarifying, i was just having a peek at https://github.com/googleapis/google-api-python-client/blob/master/docs/start.md when i noticed the sockets bit.

sonicaj: So we don't need to close it ?

themylogin: @sonicaj good catch actually, thank you

sonicaj (approved):
Thank you, however on middleware restart we wouldn't be closing it. Should we maybe look at the terminate method for each service we have which is called when middleware shuts down or maybe having sort of a destructor for gmail service class ? Though i am not sure if this would be a problem on middleware shutting down.

themylogin:
> however on middleware restart we wouldn't be closing it

Are you saying that these connections will stay alive after middleware process dies?

PR #5789, opened 2020-10-06
NAS-107860 / 20.10 / replace lshw with lspci for getting GPU info on SCALE

yocalebo:


src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -231,22 +229,22 @@ def get_storage_devices_topology(self):
...
         if self.GPU:
             return self.GPU
 
         not_available = {'available': False, 'vendor': None}
-        cp = await run(['lshw', '-numeric', '-C', 'display'], check=False)
-        if cp.returncode:
-            self.logger.error('Unable to retrieve GPU details: %s', cp.stderr.decode())
+
+        try:
+            vga = subprocess.check_output('lspci | grep -i vga', shell=True).decode().strip().split('\n')
```
sonicaj, line 239: Please use `run`, this will block because of being async.
Also can we please avoid using `shell` ?

sonicaj (changes requested):

src/middlewared/middlewared/plugins/device_/device_info_linux.py
yocalebo: Fixed

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -231,22 +232,23 @@ def get_storage_devices_topology(self):
...
+        # we only support nvidia GPUs at the moment
+        cmd = ['lspci', '-d', NVIDIA_PCI_ID + VGA_CLASS_ID]
+        vga = await run(cmd, check=False)
+        if vga.returncode:
+            self.logger.error('Unable to retrieve GPU details: %s', vga.stderr.decode())
             return not_available
 
-        vendor = RE_GPU_VENDOR.findall(cp.stdout.decode())
-        if not vendor:
+        nvidia = [i for i in vga.stdout.decode().strip().split('\n') if 'nvidia' in i.lower()]
```
sonicaj, line 248: How about just doing `'nvidia' in vga.stdout.decode().lower()` ?

sonicaj (approved):

src/middlewared/middlewared/plugins/device_/device_info_linux.py
yocalebo: Yeah, that reads better. I've changed it to that.

sonicaj (approved):
Thank you

PR #5794, opened 2020-10-06
NAS-107619 / 12.0 / configure dual nvdimms on M series version 3 hardware

yocalebo:


amotin (approved):
Looks good, other then such use of the float() for version comparison is slightly unorthodox to me.

william-gr (approved):

yocalebo:
> Looks good, other then such use of the float() for version comparison is slightly unorthodox to me.

Remove the casting to a float() as it's not needed.

src/middlewared/middlewared/etc_files/loader.py
```diff
@@ -125,5 +126,19 @@ def generate_ec2_config(middleware):
...
 
 
+def generate_dual_nvdimm_config(middleware):
+    data = middleware.call_sync('system.dmidecode_info')['system-product-name']
+
+    product = data['system-product-name']
+    version = data['system-version']
+    minimum_vers = '3'
+
+    if product.startswith('TRUENAS-M') and version >= minimum_vers:
```
william-gr, line 136: Using string comparisons here is a bit odd, no?

yocalebo: Casting it to float could lead to more issues if we start stamping with '3.0.1a' (not saying we will but who knows)

Local testing showed that python string comparison handled it correctly in about every combination that I could think of.

william-gr: system version is something we write, no?

Also there is `from packaging import version`

version.parse("3.0.1a")

yocalebo: Neat little module that I didn't know existed. I've updated the PR.

src/middlewared/middlewared/etc_files/loader.py
```diff
@@ -125,5 +127,19 @@ def generate_ec2_config(middleware):
         ]
 
 
+def generate_dual_nvdimm_config(middleware):
+    data = middleware.call_sync('system.dmidecode_info')['system-product-name']
+
+    product = data['system-product-name']
+    current_vers = version.parse(data['system-version'])
```
william-gr, line 134: I dont know what all our supported platforms return in this field.

Might be wise to catch an exception here and return early

william-gr (changes requested):

src/middlewared/middlewared/etc_files/loader.py
```diff
@@ -125,5 +127,23 @@ def generate_ec2_config(middleware):
...
 
+def generate_dual_nvdimm_config(middleware):
+    data = middleware.call_sync('system.dmidecode_info')['system-product-name']
+
+    product = data['system-product-name']
+    try:
+        current_vers = version.parse(data['system-version'])
+        minimum_vers = version.Version('3.0')
+    except Exception as e:
+        middleware.logger.error('Failed determning hardware version with error: %s', e)
```
william-gr, line 138: typo "determning"

william-gr (approved):
Thanks!

PR #5797, opened 2020-10-07
NAS-107906 / 20.10 / Add workarounds for Linux ACLs

anodos325:
Default to xattr=sa on new ZFS datasets.
Use setfacl/chmod/chown for permissions management. This is inefficient, but more testing is required for "acltool" to validate correct behavior.
Separate out Linux / FreeBSD ACL-related methods. This will be cleaned up further at a future date.


src/middlewared/middlewared/plugins/filesystem_/acl_freebsd.py
```diff
@@ -0,0 +1,751 @@
...
+                'setgid': bool(st.st_mode & pystat.S_ISGID),
+                'sticky': bool(st.st_mode & pystat.S_ISVTX),
+            },
+            'acltype': ACLType.POSIX1E.name
+        }
+
+        ret['uid'] = st.st_uid
+        ret['gid'] = st.st_gid
+
+        gfacl = subprocess.run(['getfacl', '-c' if osc.IS_LINUX else '-q', '-n', path],
```
william-gr, line 457: Since this only runs on freebsd we may wabt to get rid of this if?

william-gr (commented):
Wonderf if we should have a common class here with the methods definitions so we dont repeat ourselves and also so the API doesnt diverge?

Thoughts? @themylogin

src/middlewared/middlewared/plugins/filesystem_/acl_linux.py
```diff
@@ -0,0 +1,573 @@
...
+        return self.getacl_posix1e(path, simplified)
+
+    @private
+    def setacl_nfs4(self, job, data):
+        raise CallError('NFSv4 ACLs are not yet implemented.', errno.EOPNOTSUP)
+
+    @private
+    def setacl_posix1e(self, job, data):
+        job.set_progress(0, 'Preparing to set acl.')
+        if osc.IS_FREEBSD:
```
william-gr, line 453: I think that cannot ever be true, as _linux.py is not loaded on bsd

src/middlewared/middlewared/plugins/filesystem_/acl_freebsd.py
anodos325: > Wonderf if we should have a common class here with the methods definitions so we dont repeat ourselves and also so the API doesnt diverge?
> 
> Thoughts? @themylogin

I think that's a good idea. I'll create a base with the high-level function definitions (like getacl, setacl, etc).

src/middlewared/middlewared/plugins/filesystem_/acl_linux.py
anodos325: Yeah. I'll go through and clean up the osc stuff from both sides.

src/middlewared/middlewared/plugins/filesystem_/acl_base.py
```diff
@@ -0,0 +1,409 @@
...
+        """
+
+    @accepts()
+    async def default_acl_choices(self):
+        """
+        Get list of default ACL types.
+        """
+
+    @accepts(
+        Str('acl_type', default='NFS4_OPEN' if osc.IS_FREEBSD else 'POSIX_OPEN', enum=[x.name for x in ACLDefault]),
```
william-gr, line 400: This changes the API in a non-backward compatible way.

Thinking if we could make alias for the previous values matching NFS4. Thoughts?

kmoore134:
retest this please

src/middlewared/middlewared/plugins/filesystem_/acl_base.py
anodos325: Sounds like a good idea. I'll implement in morning.

src/middlewared/middlewared/plugins/filesystem_/acl_base.py
```diff
@@ -177,6 +177,14 @@ class ACLDefault(enum.Enum):
             'perms': {"READ": False, "WRITE": False, "EXECUTE": False},
         }
     ]}
+    OPEN = NFS4_OPEN if osc.IS_FREEBSD else POSIX_OPEN
+    RESTRICTED = NFS4_RESTRICTED if osc.IS_FREEBSD else POSIX_RESTRICTED
+    HOME = NFS4_HOME if osc.IS_FREEBSD else POSIX_RESTRICTED
+
+    def options():
+        rv = [x.name for x in ACLDefault]
+        rv.extend(['OPEN', 'RESTRICTED', 'HOME'])
```
william-gr, line 186: Wont these already be in ACLDefault?

william-gr (approved):

PR #5799, opened 2020-10-08
NAS-107859 / 20.10 / Fix setting up bridges and MTU on Linux

themylogin:
@sonicaj can you please check if we also need to handle `kube-bridge` in one of these diffs


sonicaj (approved):

src/middlewared/middlewared/plugins/interface/netif_linux/interface.py
```diff
@@ -50,7 +50,11 @@ def mtu(self):
 
     @mtu.setter
     def mtu(self, mtu):
+        up = InterfaceFlags.UP in self.flags
         run(["ip", "link", "set", "dev", self.name, "mtu", str(mtu)])
+        self.down()
```
william-gr, line 55: Shouldnt you .down only if its up?

william-gr (approved):
Thanks!

PR #5810, opened 2020-10-12
NAS-107789 / 20.10 / reporting.realtime interface speeds

themylogin:


william-gr (commented):
Thoughts on having how much of the total utilization?

Say the link is 80% saturated/

themylogin:
Can't they divide two numbers themselves? :)

william-gr:
@beanpole135 thoughts? Does this work well enough for you?

beanpole135:
I am fine with it reporting the current/max speeds, and then we can calculate the percentage ourselves.
The original reason I asked for the percentage is so that this API is not returning the same static info (max speeds for each interface) for every data interval sent out. That just seemed like a waste when you could return the dynamic percentage instead (which is typically a smaller number and reduces the data transfer - albeit only a small amount)

william-gr (approved):

themylogin:
backport

PR #5811, opened 2020-10-12
NAS-107628 / 20.12 / fix IPv6 on HA systems

yocalebo:
Here is what I've noticed. It's pretty clear that IPv6 on HA was thought about but not "finished". So here are the things I've done:

- add an `int_vipv6address` column
- add an `alias_vipv6address` column
- add an `int_ipv6address_b` column
- change a lot of the database columns string size from `42` to `45` because that's the max char size of an IPv6 string
- made freeBSD ix-netif rc script aware of the new db columns as well as simplify the logic

Another subtle change that should not break backwards compatibility. If user adds an IPv4 and an IPv6 address on an interface, then the entries will be contained to the `network_interfaces` table. If a 3rd address is added to that interface, it will roll over to the `network_alias` table as expected. The previous behavior, is that if the user added an IPv4 and an IPv6 address, depending on which IP address was added first, then parts of the second IP address would roll over to the `network_alias` table which was confusing.


src/middlewared/middlewared/plugins/network.py
```diff
@@ -1148,19 +1176,33 @@ async def _common_validation(self, verrors, schema_name, data, itype, update=Non
...
+                        for v in (await self.middleware.call(
+                            'interface.scan_vrrp', data['name'], None, 5,
+                        )).values():
+                            used_vhids.update(set(v))
+                        if data['failover_vhid'] in used_vhids:
+                            used_vhids = ', '.join([str(i) for i in used_vhids])
+                            verrors.add(
+                                f'{schema_name}.failover_vhid',
+                                f'The following VHIDs are already in use: {used_vhids}.'
+                            )
```
william-gr, line 1205: Doesn't Linux keepalived also uses VRRP packets or does it have its own conflict mechanisms?

william-gr (approved):
Looks good. Lets get @themylogin review as well, since he is the migrations expert : P

src/middlewared/middlewared/plugins/network.py
yocalebo: Great question, I've made it so that VRRP doesn't use multicast address space. It uses unicast, so the VHID is irrelevant 😄

themylogin (approved):

PR #5812, opened 2020-10-13
NAS-107829 / 20.10 / reporting.realtime disk stats

themylogin:
Two considerations:
- I mocked `/proc/diskstats` with 260 disks (780 lines: disk + two partitions), average iteration time was `0.0014 sec`, with test running every two seconds it is 0.07% of one CPU core and that makes me think Cython version is not really necessary.
- Average of "DIsk IO utilization" across all disks seems does not seem to be an indicative metric (If I have 100 disks and one of them is loaded 100% while others are idling, it will show 1%), I chose to expose Maximum, please correct me if I am wrong.

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/iostat.py
+++ b/src/middlewared/middlewared/plugins/reporting/iostat.py
@@ -1,0 +1,79 @@
+from collections import namedtuple
+import time
+
+DiskStat = namedtuple("DiskStat", ["rd_ios", "rd_bytes", "wr_ios", "wr_bytes", "tot_ticks"])
+
+
+class DiskStats:
+    def __init__(self):
+        self.prev = None
+        self.prev_time = None
+        self.current = None
+        self.current_time = None
+        self._read()
+
+    def _read(self):
+        with open("/proc/diskstats") as f:
+            lines = f.readlines()
+
+        self.current = {}
+        for line in lines:
+            line = line.split()
+
+            disk = line[2]
+
+            if disk.startswith("loop"):
+                continue
+
+            parent = disk
+            while parent and (parent[-1].isdigit() or parent[-1] == "p" and parent.startswith("nvme")):
+                parent = parent[:-1]
+            if parent in self.current:
+                continue
+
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/reporting/iostat.py
```diff
@@ -0,0 +1,75 @@
...
+        for disk, current in self.current.items():
+            prev = self.prev.get(disk)
+            if prev is None:
+                continue
+
+            read_ops += current.rd_ios - prev.rd_ios
+            read_bytes += current.rd_bytes - prev.rd_bytes
+            write_ops += current.wr_ios - prev.wr_ios
+            write_bytes += current.wr_bytes - prev.wr_bytes
+            total_ticks = max(total_ticks, current.tot_ticks - prev.tot_ticks)
```
william-gr, line 65: I believe we do not want to know the most busy disk %.
We do want to know an average because odds are all the disks will be in a pool and its likely the ZIL device will be most busy than the rest, which doesn't tell us much of a story. Now if the average % busy is high then it may be a concern.

src/middlewared/middlewared/plugins/reporting/iostat.py
```diff
@@ -0,0 +1,77 @@
...
+
+            parent = disk
+            while parent and (parent[-1].isdigit() or parent[-1] == "p" and parent.startswith("nvme")):
+                parent = parent[:-1]
+            if parent in self.current:
+                continue
+
+            disk_stat = DiskStat(
+                rd_ios=int(line[3]),
+                rd_bytes=int(line[5]) * 512,
```
william-gr, line 36: Why 512?

themylogin: These are sectors which are always 512 bytes as kernel source says https://github.com/torvalds/linux/blob/master/include/linux/types.h#L120

william-gr: Thanks for enlightening me. Might be worth putting a comment explaining that since its not clear every device would be treat as such.

william-gr (approved):

PR #5814, opened 2020-10-13
NAS-107925 / 20.12 / Prevent system dataset changes while AD is enabled

anodos325:
We need some safeguards here as this can result in service disruption.


src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -100,6 +100,13 @@ async def do_update(self, job, data):
 
         verrors = ValidationErrors()
         if new['pool'] and new['pool'] != await self.middleware.call('boot.pool_name'):
```
william-gr, line 102: Seems like this check is not in the right place? For this error it should be new['pool'] != old['pool'] or something?

william-gr: @anodos325 ping. Just making sure it was not forgotten

william-gr (approved):

PR #5838, opened 2020-10-15
NAS-107954 / 20.10 / middlewared: properly close fds and report errors when changing system dataset

william-gr:


src/middlewared/middlewared/service.py
```diff
@@ -1097,15 +1104,17 @@ def reconfigure_logging(self):
...
+        self.__kill_multiprocessing()
 
         self.middleware.send_event('core.reconfigure_logging', 'CHANGED')
 
+    @private
+    def stop_logging(self):
+        stop_logging()
+        self.__kill_multiprocessing()
+
+        self.middleware.send_event('core.reconfigure_logging', 'CHANGED', fields={'stop': True})
```
themylogin, line 1116: Why don't you just make a `core.stop_logging` event?

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -416,6 +416,9 @@ async def migrate(self, _from, _to):
             path = '/tmp/system.new'
             if not os.path.exists('/tmp/system.new'):
                 os.mkdir('/tmp/system.new')
+            else:
+                # Make sure we clean up any previous attempts
+                await run('umount', '-R', path, check=False)
```
themylogin, line 421: Does FreeBSD have this option?

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -344,8 +344,8 @@ async def __umount(self, pool, uuid):
         for dataset, name in reversed(self.__get_datasets(pool, uuid)):
             try:
                 await run('umount', '-f', dataset)
-            except subprocess.CalledProcessError:
-                self.logger.warning('Unable to umount %r', dataset)
+            except subprocess.CalledProcessError as e:
+                raise CallError(f'Unable to umount {dataset}: {e.stderr.decode()}')
```
themylogin, line 348: I remember we were not able to unmount something sometimes but it was fine to ignore it, did it happen to you while working on this issue?

themylogin (changes requested):

src/middlewared/middlewared/service.py
william-gr: Thought about it but just make things harder for no reason.

src/middlewared/middlewared/plugins/sysdataset.py
william-gr: Dunno, also not relevant anymore

src/middlewared/middlewared/plugins/sysdataset.py
william-gr: It was not fine, we were just lazy to handle it correctly. In Linux -f does not really force anything.

src/middlewared/middlewared/plugins/sysdataset.py
themylogin: Does that mean we don't need FreeBSD code in `master` anymore?

PR #5844, opened 2020-10-16
NAS-107927 / 12.0 / Fix NFS bind to any IP

themylogin:


src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -62,13 +62,13 @@ async def bindip_choices(self):
         """
         return {
             d['address']: d['address'] for d in await self.middleware.call(
-                'interface.ip_in_use', {'static': True, 'any': True}
+                'interface.ip_in_use', {'static': True}
```
william-gr, line 65: I will need some clarification why we wont return 0.0.0.0 here?

william-gr: @themylogin ping

themylogin: It's "listen on all interfaces if not specified, otherwise only listen on specified". `0.0.0.0` does not fit there.

william-gr: Guess we need to start thinking more about it and try to be consistent between the services.

I know other places that we offer IN_ANYADDR as an option.

If this is a problem of the underlying service not supporting 0.0.0.0 its something we can handle under the hood in middlewared/config generation.

What is more common? 0.0.0.0 or empty for meaning all addrs?

william-gr: I should add, this looks fine for 12.0, but lets think about it in the backported change to master

PR #5854, opened 2020-10-18
NAS-107929 / 20.10 / Properly raise exception if cloning snapshot fails

sonicaj:


themylogin (changes requested):
You can also change this method usage in vm/clone.py

themylogin (approved):

themylogin (approved):

PR #5857, opened 2020-10-18
NAS-107791 / 20.10 / Send pool.dataset.query events

sonicaj:
This PR adds changes to send events for `pool.dataset.query` by relying on zfs events, this helps ensure that changes via zfs cli are not missed.


themylogin (approved):
Cool!

PR #5861, opened 2020-10-19
NAS-107823 / 20.10 / reporting.graphite event

themylogin:


src/middlewared/middlewared/plugins/reporting/graphite.py
```diff
@@ -0,0 +1,164 @@
...
+        names = []
+        if self.arg:
+            mode, names = self.arg.split(",", 1)
+
+            if mode not in ["include", "exclude"]:
+                raise ValueError(f"Invalid mode: {mode!r}")
+
+            names = [re.compile(r) for r in names.split(",")]
+
+        queue = Queue()
```
yocalebo, line 74: Should we place an upper limit on the queue size?

william-gr (approved):
Good job!

PR #5879, opened 2020-10-20
NAS-108006 / 20.12 / Gracefully handle zfs events for system dataset

sonicaj:


src/middlewared/middlewared/plugins/zfs_/zfs_events.py
```diff
@@ -187,7 +187,14 @@ async def zfs_events(middleware, data):
         event_type = data['history_internal_name']
         ds_id = data['history_dsname']
         if event_type in ('create', 'set'):
-            ds_data = await middleware.call('pool.dataset.get_instance', ds_id)
+            ds_data = await middleware.call('pool.dataset.query', [['id', '=', ds_id]])
+            if not ds_data:
+                # We should not send an event because of 2 reasons:
+                # 1) Dataset in question was system dataset which was filtered out by pool.dataset service
+                # 2) Dataset got deleted in a race condition which is still fine as destroy event will catch that
```
william-gr, line 194: Isn't #2 not covered here because ` if event_type in ('create', 'set'):`  ?

sonicaj: It is covered in `else` clause

william-gr (approved):

PR #5880, opened 2020-10-20
NAS-107994 / 20.12 / change glusterd working-directory so metadata isn't lost on OS upgrades

yocalebo:
- get rid of `is_freenas` calls in `sysdataset.py` plugin
- setup a `glusterd` dataset underneath the `.system` dataset
- stop/restart/reconfigure glusterd in the event the system dataset changes

### DIFF
--- a/src/middlewared/middlewared/etc_files/glusterd.conf.mako
+++ b/src/middlewared/middlewared/etc_files/glusterd.conf.mako
@@ -1,0 +1,20 @@
+<%
+    sysdataset_path = middleware.call_sync('systemdataset.config')['path']
+    work_dir = sysdataset_path + '/glusterd'
+%>\
+
+volume management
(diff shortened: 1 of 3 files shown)


william-gr (commented):
Wonder if we should create a dataset for gluster inside .system? Not just a dir

william-gr:
I agree with putting it inside system dataset, I just think it would probably be more future-proof if we create another dataset within it so we can have more control if needed

william-gr:
sysdataset.py:350 __get_datasets

yocalebo:
> I agree with putting it inside system dataset, I just think it would probably be more future-proof if we create another dataset within it so we can have more control if needed

No problem, I'll make appropriate changes.

william-gr (approved):

PR #5891, opened 2020-10-20
NAS-108016 / 20.12 / Improve dataset querying / creation performance

sonicaj:

1) Uses a different endpoint to retrieve serialized datasets data if `id` filter has been applied. It takes less then half the time it used to before with this approach. One thing to note is that we remove `state_options` for `id` filter and I have ensured that there are no usages for those in our codebase. This also helps us be consistent across varying filters and not have special consideration for a single filter.
2) Improves `pool.dataset.create` performance which mainly utilises the gains introduced in (1) and also tries to minimise the times it queries for the same dataset over and over again. Earlier in my environment it used to take around 19 seconds to create a dataset and with current approach takes ~3.44 seconds.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2836,7 +2836,11 @@ class PoolDatasetService(CRUDService):
         if '/' not in data['name']:
             verrors.add('pool_dataset_create.name', 'You need a full name, e.g. pool/newdataset')
         else:
-            await self.__common_validation(verrors, 'pool_dataset_create', data, 'CREATE')
+            parent_ds = await self.middleware.call('pool.dataset.query', [('id', '=', data['name'].rsplit('/')[0])])
+            await self.__common_validation(verrors, 'pool_dataset_create', data, 'CREATE', parent_ds)
+            parent_ds = parent_ds[0]
+
+        verrors.check()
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2836,7 +2836,9 @@ async def do_create(self, app, data):
         if '/' not in data['name']:
             verrors.add('pool_dataset_create.name', 'You need a full name, e.g. pool/newdataset')
```
themylogin, line 2837: This branch does not define `parent_ds` that will be used below. You should probably just raise validation errors here.

themylogin (changes requested):

themylogin (approved):

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2836,7 +2836,11 @@ async def do_create(self, app, data):
...
         else:
-            await self.__common_validation(verrors, 'pool_dataset_create', data, 'CREATE')
+            parent_ds = await self.middleware.call('pool.dataset.query', [('id', '=', data['name'].rsplit('/')[0])])
```
william-gr, line 2839: Hmm, this is not the parent

>>> 'tank/foo/bar'.rsplit('/')[0]
'tank'

PR #5911, opened 2020-10-23
Do not use a task, run until complete for initialization

william-gr:


themylogin (approved):
Can I ask why?

william-gr:
@themylogin  If an exception happens on __initialize, middleware will continue running the asyncio loop forever

PR #5914, opened 2020-10-24
NAS-108088 / 20.12 / Use py-libzfs for retrieving snapshots

sonicaj:


src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -952,47 +952,21 @@ def query(self, filters=None, options=None):
                 filter_getattrs(filters).issubset({'name', 'pool'})
             )
         ):
-            # Using zfs list -o name is dozens of times faster than py-libzfs
```
freqlabs, line 955: That's an interesting comment... 🤔

sonicaj: It certainly was with how py-libzfs retrieved snapshots ; )
But with the new stats we have
```
>>> st=time.time();s = z.snapshots_serialized(datasets=[], props=['name'], recursive=True);len(s);print(time.time()-st)
7275
5.18239951133728
```
And zfs cli
```
time zfs list -H -o name -t snapshot 1>/dev/null
zfs list -H -o name -t snapshot > /dev/null  0.25s user 9.59s system 99% cpu 9.847 total
```

sonicaj: And after retrieving this data, we were doing some post processing which also took time and which we don't have to do in py-libzfs, so much better ; )

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -952,47 +952,21 @@ def query(self, filters=None, options=None):
...
-                        raise
-            else:
-                for i in zfs.snapshots:
-                    try:
-                        snapshots.append(i.__getstate__())
-                    except libzfs.ZFSException as e:
-                        # snapshot may have been deleted while this is running
-                        if e.code != libzfs.Error.NOENT:
-                            raise
+                kwargs['datasets'] = [filters[0][2]]
```
william-gr, line 966: Why datasets? We want a specific snapshot id, isnt the old code more suitable for this?

william-gr (approved):

PR #5916, opened 2020-10-26
NAS-102208 / 20.12 / disallow journal thread from making remote db changes

yocalebo:
This prevents the journal thread from making remote db changes when the OS'es do not match between the nodes. I've cached the os version in `RemoteClient` trying to ensure that any error that could possibly occur while setting that attribute is handled to prevent that class from crashing.

This sets the attribute when the remote client is connected as well as clears it when it is disconnected.

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
@@ -1420,8 +1420,8 @@ class JournalSync:
     def _os_versions_match(self):
 
         try:
-            loc = self.middleware.call_sync('failover.get_os_version')
-            rem = self.middleware.call_sync('failover.call_remote', 'failover.get_os_version')
+            rem = self.middleware.call_sync('failover.get_remote_os_version')
+            loc = self.middleware.call_sync('system.version')
         except Exception:
             return False
 
@@ -1442,18 +1442,23 @@ async def journal_ha(middleware):
 
 
 def journal_sync(middleware):
+
+    alert = True
     while True:
         try:
             journal = Journal()
             journal_sync = JournalSync(middleware, sql_queue, journal)
             while True:
                 journal_sync.process()
-        except Exception as e:
-            if isinstance(e, OSVersionMismatch):
+                alert = True
+        except OSVersionMismatch:
+            if alert:
                 logger.warning('OS version does not match remote node. Not syncing journal')
-            else:
-                logger.warning('Failed to sync journal', exc_info=True)
-            time.sleep(5)
+                alert = False
+        except Exception:
+            logger.warning('Failed to sync journal', exc_info=True)
+
+        time.sleep(5)
 
 
 async def interface_pre_sync_hook(middleware):
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
@@ -79,6 +79,15 @@ class TruenasNodeSessionManagerCredentials(SessionManagerCredentials):
     pass
 
 
+class OSVersionMismatch(Exception):
+
+    """
+    Raised in JournalSync thread when the remote nodes OS version
+    does not match the local nodes OS version.
+    """
+    pass
+
+
 def throttle_condition(middleware, app, *args, **kwargs):
     # app is None means internal middleware call
     if app is None or (app and app.authenticated):
@@ -1322,6 +1331,10 @@ class JournalSync:
 
             self.journal.clear()
 
+        if not self._os_versions_match():
+            if self.journal:
+                raise OSVersionMismatch()
+
         had_journal_items = bool(self.journal)
         flush_succeeded = self._flush_journal()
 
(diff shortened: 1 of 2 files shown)


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1404,6 +1417,16 @@ def _handle_sql_queue_item(self, item):
     def _update_failover_status(self):
         self.failover_status = self.middleware.call_sync('failover.status')
 
+    def _os_versions_match(self):
+
+        try:
+            loc = self.middleware.call_sync('failover.get_os_version')
+            rem = self.middleware.call_sync('failover.call_remote', 'failover.get_os_version')
```
william-gr, line 1424: I am confused, get_os_version already gets the version of the remote, no?

yocalebo: No it just gets the cached version that is set in `RemoteClient`.

william-gr: What is cached on RemoteClient is the remove version. So we are calling the remote node get its cached version of the other remote node (us?)?

yocalebo: No, the `RemoteClient` is running `self.client.call('system.version')` which is the local OS version and then setting the `self._os_version` attribute.

yocalebo: The `JournalSync` thread is running `failover.get_os_version` which gets the `RemoteClient`'s cached `self._os_version` attribute.

william-gr: `self.client.call('system.version')` definitely doesnt get the local version.

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/remote.py
```diff
@@ -174,6 +186,16 @@ def sendfile(self, token, local_path, remote_path):
                     break
             time.sleep(0.5)
 
+    def get_os_version(self):
```
william-gr, line 189: Its probably better if we rename this method to avoid confusion? `get_remote_version` or something?

yocalebo: 100% agree. I've fixed.

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1425,8 +1448,11 @@ def journal_sync(middleware):
             journal_sync = JournalSync(middleware, sql_queue, journal)
             while True:
                 journal_sync.process()
-        except Exception:
-            logger.warning('Failed to sync journal', exc_info=True)
+        except Exception as e:
```
william-gr, line 1451: except OSVersionMismatch as e:
    logger...
except Exception:
    logger

william-gr: Also we will most definitely keep getting that message on log every 5 seconds, which seems less than ideal

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/remote.py
```diff
@@ -90,6 +92,13 @@ def _on_connect(self):
         """
         Called everytime connection has been established.
         """
+
+        # journal thread checks this attribute to ensure
+        # we're not trying to alter the remote db if the
+        # OS versions do not match since schema changes
+        # can (and do) change between upgrades
+        self._os_version = self.get_remote_os_version()
```
william-gr, line 100: _remote_os_version defined in init but _os_version here

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/remote.py
```diff
@@ -174,6 +186,16 @@ def sendfile(self, token, local_path, remote_path):
...
 
+    def get_remote_os_version(self):
+
+        if self._remote_os_version is None:
+            try:
+                self._remote_os_version = self.client.call('system.version')
+            except Exception:
+                logger.error('Failed to determine OS version', exc_info=True)
+
+        return self._remote__os_version
```
william-gr, line 197: typo

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
yocalebo: Yeah we will get logs every 5 seconds so I can either change the log message to:
1. logger.debug
or
2. don't raise a custom exception and do what we do when `failover.status != 'MASTER'` and clear the journal?

william-gr: or 3. use a variable to log it consecutively only once


1. wont solve it, only .trace would but then we will miss it entirely

themylogin (approved):

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1419,15 +1442,22 @@ async def journal_ha(middleware):
...
             journal_sync = JournalSync(middleware, sql_queue, journal)
             while True:
                 journal_sync.process()
+        except OSVersionMismatch:
+            if alert:
+                logger.warning('OS version does not match remote node. Not syncing journal')
+                alert = False
         except Exception:
             logger.warning('Failed to sync journal', exc_info=True)
-            time.sleep(5)
```
william-gr, line 1430: We need to set alert = True again if we had a successful loop pass with matched os version

william-gr (approved):

PR #5938, opened 2020-10-28
NAS-108075 / 20.12 / Allow power supply alert on FREENAS-CERTIFIED

themylogin:
@william-gr I am not sure if this has sense to hide this alert on CORE when hardware is not FREENAS-CERTIFIED

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/alert/source/sensors.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/alert/source/sensors.py
@@ -42,17 +42,18 @@ class PowerSupplyAlertClass(AlertClass):
 
 
 class SensorsAlertSource(AlertSource):
-    products = ("ENTERPRISE",)
-
     async def check(self):
-        baseboard_manufacturer = (
-            await self.middleware.call('system.dmidecode_info')
-        )['baseboard-manufacturer']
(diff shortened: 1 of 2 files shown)


william-gr:
@darkfiberiru can you speak to any of this? Can we assume this will work for our hardware tagged as FREENAS-CERTIFIED?

themylogin:
This was tested on fn02 (FREENAS-CERTIFIED-4U-24)

william-gr:
Also thinking forward, are we relabeling these to TRUENAS-CERTIFIED yet?

darkfiberiru:
Freenas certified is going away and r-series will replace it. I agree with @themylogin  I don't see reason to not have sensor visible if it's there. If no pmbus is available (method to talk to psu) those fields just don't populate in sensor list. If only one psu is inserted when bmc starts and scans bus then you insert second later the ps2 field will be missing until you reset bmc to rescan i2c bus. (Might be simpler method)

If we want to segment iX hardware and non-ix hardware as policy that fine but I don't see a technical reason to as were using a standard mechanism here. These sensors are defined by ipmi not supermicro.

william-gr (commented):
We should probably add R-Series support here, then?

william-gr (approved):

darkfiberiru (commented):
Looks good to me

PR #5939, opened 2020-10-28
NAS-108050 / 12.0 / Do not retrieve unnecessary sysctls in snmp-agent

themylogin:
I am not sure about the scope of this ticket, is it only to fix SNMP or also `freenas-debug/sysctl/sysctl.sh`? (it only has `sysctl -a`, I have no idea how to fix this if not in kernel)

### DIFF
--- a/src/freenas/usr/local/bin/snmp-agent.py
+++ b/src/freenas/usr/local/bin/snmp-agent.py
@@ -32,8 +32,8 @@ def get_Kstat():
 
 def get_Kstat_FreeBSD():
     Kstats = [
-        "kstat.zfs",
-        "vfs.zfs"
+        "kstat.zfs.misc.arcstats",
+        "vfs.zfs.version.spa",
     ]
 
     Kstat = {}
--- a/src/freenas/usr/local/libexec/freenas-debug/zfs/zfs.sh
+++ b/src/freenas/usr/local/libexec/freenas-debug/zfs/zfs.sh
@@ -113,4 +113,17 @@ zfs_func()
 	section_header  "pool joined to storage"
 	cat  /tmp/pool.normal | ${FREENAS_DEBUG_MODULEDIR}/zfs/join_pool.nawk
 	section_footer
+
+	section_header  "kstat"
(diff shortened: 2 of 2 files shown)


william-gr:
Can we also fix freenas debug in this PR?

freqlabs:
I suggest sticking with `sysctl -a` in sysctl.sh and adding the few skipped kstats explicitly.
```sh
sysctl -a
for kstat in fletcher_4_bench vdev_raidz_bench dbgmsg; do
    sysctl kstat.zfs.misc.${kstat}
done
for kstat in multihost misc.state txgs; do
    for pool in $(zpool list -Ho name); do
        sysctl kstat.zfs.${pool}.${kstat}
    done
done
```
The kstats might even make more sense in the ZFS section instead, but either way works.
This gives us explicit control over the CTLFLAG_SKIP kstats we are requesting. Otherwise we end up grabbing a bunch of sysctls that are normally skipped, for example some debug sysctls for cxgbe.

themylogin:
@freqlabs I do not understand, the point of original ticket was to remove `kstat.zfs.misc.dbufs`, but your code keeps it and just adds more sysctls?

freqlabs:
@themylogin in the next update I am skipping these kstats by marking them with CTLFLAG_SKIP in the kernel module, so they will not be included in `sysctl -a`:

```
kstat.zfs.${pool}.multihost
kstat.zfs.${pool}.misc.state
kstat.zfs.${pool}.txgs
kstat.zfs.misc.fletcher_4_bench
kstat.zfs.misc.vdev_raidz_bench
kstat.zfs.misc.dbufs
kstat.zfs.misc.dbgmsg
```

Planning further ahead, in FreeBSD 13 `sysctl -aN` will still give the names of those skipped sysctls and many others (otherwise there is no way to find out they exist if you don't already know). That is why I recommend sticking with `sysctl -a` and adding the handful of soon-to-be-normally-skipped kstats we are interested in for ZFS.

freqlabs (approved):

william-gr (approved):

PR #5946, opened 2020-10-29
NAS-108140 / 20.12 / make sure ix-postinit runs as one of the last services in startup

yocalebo:


debian/debian/ix-postinit.service
```diff
@@ -1,8 +1,7 @@
 [Unit]
 Description=Execute TrueNAS custom post-init tasks
 
-After=systemd-user-sessions.service plymouth-quit-wait.service getty-pre.target
-After=rc-local.service
+After=ix-zfs.service
```
william-gr, line 4: This indeed seems too early, postinit should be after boot is completed. We have preinit for other things

yocalebo: According to systemd-analyze plot, the `glusterd` service is starting after `ix-postinit`. Changing `ix-postinit` to run after `multi-user.target` should be enough to get it done.

william-gr (approved):

PR #5951, opened 2020-10-30
NAS-108110 / 20.12 / fix exporting zpools on freeBSD

yocalebo:
A few things about lsof on  freeSBD

- it's broken and will cause a CPU core to spin at 100% and also eat TONS of ram (so much ram, that it can go into swap space and ultimately cause crashes of userspace processes)
- it's man page isn't being built on freeBSD for reasons that I don't really want to troubleshoot
- it gives a warning about the fact the version of lsof was compiled for 12.0-RELEASE but we're not using that freeBSD release in our TN-12.0-RELEASE

Replace it with the in base `fstat` utility. It's orders of magnitude quicker, and doesn't cause random memory exhaustion.

With this change, I'm able to export zpools with running services on TN-12.0 without issues. Furthermore, this fixes a bug where even if `lsof` was successful, it actually didn't work the way we thought it did. So this fixes those 2 issues.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -3546,7 +3546,7 @@ class PoolDatasetService(CRUDService):
                 if q['id'] not in ['QUOTA', 'REFQUOTA']:
                     verrors.add(
                         f'quotas.{i}.id',
-                        f'id for quota_type DATASET must be either "QUOTA" or "REFQUOTA"'
+                        'id for quota_type DATASET must be either "QUOTA" or "REFQUOTA"'
                     )
                     continue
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4069,5 +4085,51 @@ def parse_lsof(lsof, dirs):
...
+    #        from the output as its not needed
+    #        and ends up shifting those lines
+    #        quite a bit since the thread names
+    #        have spaces in their names
+    #    4. remove any lines that don't have a traditional
+    #        "path" since the path component for processes
+    #        can be a stream/socket/pipe etc
+    fstat = [
+        i for i in fstat if i and len(i.split()) >= 5 and not i.split()[1].startswith(('g_eli', 'g_mirror')) and '/' in i.split()[4]
+    ]
```
freqlabs, line 4112:
```suggestion
    def keep(fields):
        if len(fields) < 5:
            return False
        if fields[1].startswith(('g_eli', 'g_mirror')):
            return False
        return '/' in fields[4]
    fstat = [ fields for fields in map(str.split, fstat) if keep(fields) ]
```

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4069,5 +4085,51 @@ def parse_lsof(lsof, dirs):
...
+
+    pids = {}
+    pid = command = None
+    for line in fstat:
+        line = line.split()
```
freqlabs, line 4117:
```suggestion
    for line in fstat:
```

freqlabs: With the other change this is simplified as well.
I find fields to be a better name than line but that's not a big difference either way.

src/middlewared/middlewared/plugins/pool.py
freqlabs: This avoids splitting each line four times and IMO is easier to read.

yocalebo: Good idea 👍

src/middlewared/middlewared/plugins/pool.py
yocalebo: 👍

freqlabs (approved):

themylogin:
@william-gr I don't recall any specific reason. If fstat is working properly, these changes look good to me. @yocalebo does it also hang on 12.0?

yocalebo:
> @william-gr I don't recall any specific reason. If fstat is working properly, these changes look good to me. @yocalebo does it also hang on 12.0?

Yeah, it was reported to be hanging on 12 originally. lsof is including the in base zfs header files but we are using openzfs so this port will probably require some changes to make it work properly. But fstat works considerably faster and allows us to not make significant changes to lsof and diverge from upstream port.

themylogin (approved):

PR #5953, opened 2020-11-02
NAS-108168 / 12.0 / Fix SMB2 AAPL extension negotiation

anodos325:
12.0 abstracted away from exposing explicit vfs module configuration
to users (instead opting to use checkboxes). In cases where we couldn't
cleanly migrate the user's VFS objects to our new feature-based config
(acls, streams, etc), we dumped the params into SMB share auxiliary
parameters. This preserves user's existing behavior (pre-12.0), but
can cause issues with time machine support if user opts to add a
time machine share at a later date (or has inconsistently-migrated
shares) since support for the SMB2 protocol extensions is determined
on the first SMB2 tree connect. This means that all shares _must_ have
vfs_fruit enabled (even ones with VFS objects in auxiliary parameters)
if users have them on any other share.
To address this potential issue, check for "vfs objects" lines when
parsing auxiliary parameters and add vfs_fruit if the global option
"aapl_extensions" is set.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb_/registry.py
+++ b/src/middlewared/middlewared/plugins/smb_/registry.py
@@ -362,7 +362,7 @@ class SharingSMBService(Service):
             conf["fruit:resource"] = "stream"
 
         if conf["path"]:
-            await self.add_multiprotocol_conf(conf, gl, data['name'])
+            await self.add_multiprotocol_conf(conf, gl, data)
 
         if data['timemachine']:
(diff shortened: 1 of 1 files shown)


anodos325:
This also fixes an issue I observed in testing where mixed-protocol SMB / AFP / NFS shares may not properly generate a config.

william-gr (approved):

william-gr:
backport

PR #5956, opened 2020-11-02
NAS-108074 / 20.12 / Bind services to 0.0.0.0 when removing IP address from interface configuration

themylogin:


sonicaj (commented):
This looks good to me, however we should be doing it for kubernetes as well. It uses node ip too.

Also I was wondering how we should accommodate ip addresses in portals on removing ip address from interface configuration. It has a different context then this PR as this is related to services but was curious if you had thought about it as well ?

themylogin:
@sonicaj is `node_ip` the only IP we need to change

@william-gr I join to the question what should we do with iSCSI portal IPs, do we set them all to `0.0.0.0` too?

sonicaj:
@themylogin yes, we should just reset `node_ip` for k8s.

william-gr:
Yes, should reset iscsi portal IPs too.

sonicaj (approved):

src/middlewared/middlewared/plugins/interface/listen.py
```diff
@@ -0,0 +1,103 @@
...
+    @accepts()
+    async def services_restarted_on_sync(self):
+        """
+        Returns which services will be set to listen on 0.0.0.0 (and, thus, restarted) on sync.
+
+        Example result:
+        [
+            // Samba service will be set ot listen on 0.0.0.0 and restarted because it was set up to listen on
+            // 192.168.0.1 which is being removed.
+            {"type": "system-service", "service": "cifs", "ips": ["192.168.0.1"]},
```
william-gr, line 32: Very minor nit, wonder if we should make that uppercase, just to be consistent with some other places, since its a kind of ENUM

william-gr (approved):

PR #5960, opened 2020-11-03
NAS-108186 / 20.12 / Correctly populate k8s node password under etc on upgrades

sonicaj:
K3s stores node password under /etc even when we want it to operate under a different directory, so on an upgrade of the system, node password is lost as BE changes and k8s fails to initialise pods. This fixes that by ensuring we have the correct node password under /etc.

### DIFF
--- a/src/middlewared/middlewared/etc_files/rancher/node/node_passwd.py
+++ b/src/middlewared/middlewared/etc_files/rancher/node/node_passwd.py
@@ -1,0 +1,21 @@
+import os
+
+
+def render(service, middleware):
+    k8s_config = middleware.call_sync('kubernetes.config')
+    if not k8s_config['dataset']:
+        return
+
+    k3s_node_passwd_file = os.path.join('/mnt', k8s_config['dataset'], 'k3s/server/cred/node-passwd')
+    if not os.path.exists(k3s_node_passwd_file):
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/etc_files/rancher/node/node_passwd.py
```diff
@@ -0,0 +1,20 @@
...
+    if not os.path.exists(k3s_node_passwd_file):
+        # The only time this will happen is the first time k8s is configured and at that time it's okay
+        # as k3s will populate the correct password under /etc but on subsequent upgrades of the system
+        # that will be lost
+        return
+
+    with open(k3s_node_passwd_file, 'r') as f:
+        passwd = f.read().strip().split(',')[0].strip()
+
+    with open('/etc/rancher/node/password', 'w') as f:
```
william-gr, line 19: Is /etc/rancher/node dir guaranteed to exist?

william-gr (approved):

src/middlewared/middlewared/etc_files/rancher/node/node_passwd.py
sonicaj: Yes it is, `k3s` initialises `/etc/rancher/node/password` automatically with a random password if the file is not there, i'll still however handle it for just in case cases. Thank you for pointing it out.

PR #5961, opened 2020-11-03
NAS-108120 / 20.12 / raise HTTPBadRequest when user/pass is malformed or .decode() fails

yocalebo:


src/middlewared/middlewared/restful.py
```diff
@@ -21,26 +21,35 @@ async def authenticate(middleware, req):
     if auth is None:
         raise web.HTTPUnauthorized()
 
+    username = password = None
+    if auth.startswith('Authorization: Basic '):
```
william-gr, line 25: Hmm, we are already getting `Authorization` header, not sure why we would check for another Authorization within that header

src/middlewared/middlewared/restful.py
```diff
@@ -21,26 +21,35 @@ async def authenticate(middleware, req):
...
+        try:
+            username, password = base64.b64decode(auth[21:]).decode('utf8').split(':', 1)
+        except binascii.Error:
+            raise web.HTTPUnauthorized()
+
     if auth.startswith('Basic '):
         try:
             username, password = base64.b64decode(auth[6:]).decode('utf8').split(':', 1)
```
william-gr, line 33: I believe the fix would have to be here, handling the exception appropriately

william-gr (approved):

PR #5966, opened 2020-11-04
NAS-108204 / 20.12 / nfsv4 shares do not export properly

manetherenio:
The nfs-ganesha config for nfsv4 that gets generated from template is invalid due to a typo in the template. This typo prevents nfsv4 shares from being exported properly.


yocalebo (approved):

yocalebo:
Thank you for the contribution!

PR #5967, opened 2020-11-04
Initialize directory service state to HEALTHY if enabled

anodos325:
Return HEALTHY for enabled directory services until after the system is ready. The state will be updated during next directory service health check.

### DIFF
--- a/src/middlewared/middlewared/plugins/ldap.py
+++ b/src/middlewared/middlewared/plugins/ldap.py
@@ -820,6 +820,13 @@ class LDAPService(ConfigService):
             raise CallError('Automatically disabling LDAP service due to invalid configuration.',
                             errno.EINVAL)
 
+        """
+        Initialize state to "JOINING" until after booted.
+        """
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -820,6 +820,13 @@ async def started(self):
             raise CallError('Automatically disabling LDAP service due to invalid configuration.',
                             errno.EINVAL)
 
+        """
+        Initialize state to "HEALTHY" until after booted.
+        """
+        if not await self.middleware.call('system.ready'):
+            await self.set_state(DSStatus['HEALTHY'])
```
william-gr, line 827: Wonder if HEALTHY is the right state? Do we have a JOINING state that would fit better?

anodos325: Yeah, that makes more sense.

william-gr (approved):

PR #5974, opened 2020-11-04
NAS-108166 / 12.0 / R-Series support in enclosure API

themylogin:
For R10 and R20 we'll get one enclosure

For R40 we'll get two enclosures: `R40, Drawers #1-2` and `R40, Drawers #3-4` (I've invented the names, please let me know if I am wrong). They are named the same (`ECStream FS2 c209`) so I guess I order them by appearance order in `/dev/ses*`?

For R50 we'll get two enclosures too: `R50, Drawer #1` and `R50, Drawer #2`, the number I am able to get from the name

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/alert/source/license_status.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/alert/source/license_status.py
@@ -106,7 +106,8 @@ class LicenseStatusAlertSource(ThreadedAlertSource):
                         ) % license['model']
                     ))
             else:
-                if hardware[0] in ('M40', 'M50', 'M60', 'X10', 'X20', 'Z20', 'Z30', 'Z35', 'Z50'):
+                if hardware[0] in ('M40', 'M50', 'M60', 'R10', 'R20', 'R40', 'R50', 'X10', 'X20', 'Z20', 'Z30', 'Z35',
+                                   'Z50'):
                     if hardware[0] != license['model']:
                         alerts.append(Alert(
                             LicenseAlertClass,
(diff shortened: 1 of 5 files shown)


william-gr:
Code looks fine to me but I am unsure about the Drawer thing, I will let @darkfiberiru comment if it has any sense.

william-gr:
@darkfiberiru ping

darkfiberiru:
Looking at this now

darkfiberiru:
Via NEP 24 @themylogin we would want to map the two "drawers" two a single virtual ses enclosure for the ui. These are contiguous slots on the system and we don't want to duplicate slot numbers on chassis. The R50 will also have a nvme enclosure like the nvme plx but no directly attached to motherboard.

A major point of NEP 24 is that a slot number matches a single place on the chassis even if there front/rear/top drives.

I can login to the r50 and that's clear that for enclosure management drawer 1/drawer 2 just need concatenated and then pcie devices after that.

R40 looks like both ses devices(24 drives each) just need concatenated into a single enclosure for the ui also

darkfiberiru (changes requested):
We should give webui a single "enclosure" for the entire chassis so that all slot number are contiguous and non repeating. See notes above.

themylogin:
@darkfiberiru what about they order? Are they guaranteed to appear in `/dev/ses*` always in the same order?

darkfiberiru:
@themylogin They should always appear in same order. If they don't would be hardware(wiring) issue not software.

themylogin:
I've made it merge all `XXX, Drawer #N` enclosures into one

darkfiberiru (approved):
I think it looks good now other then full nvme details we need from R-Series PM

william-gr (approved):

PR #5975, opened 2020-11-04
NAS-108182 / 20.12 / integrate nfs-ganesha with gluster

yocalebo:
- query middleware for gluster peer information once so we dont have to call that for every nfs share that's in the db
- validation is completely hinged on the fact that the top-level dataset is always `/mnt/tank/.glusterfs` (this is being enforced by TC so no worries there)


yocalebo:
This works for now but I predict I'll have to make changes as the performance team starts to test this feature out @william-gr

william-gr (approved):

PR #5979, opened 2020-11-05
NAS-107663 / 20.12 / iSCSI hosts API

themylogin:


src/middlewared/middlewared/plugins/iscsi_/host.py
```diff
@@ -0,0 +1,191 @@
...
+import middlewared.sqlalchemy as sa
+
+LOCK = asyncio.Lock()
+
+
+class iSCSIHostModelModel(sa.Model):
+    __tablename__ = "services_iscsihost"
+
+    id = sa.Column(sa.Integer(), primary_key=True)
+    ip = sa.Column(sa.String(42), unique=True)
```
yocalebo, line 17: This needs to be 45 chars since IPv6 can be that length.

src/middlewared/middlewared/plugins/iscsi_/host_injection.py
```diff
@@ -0,0 +1,67 @@
...
+
+    async def run(self):
+        while True:
+            try:
+                await self.middleware.call(
+                    "iscsi.host.batch_update",
+                    [
+                        host._asdict()
+                        for host in await self.middleware.call("iscsi.host.injection.collect")
+                    ],
```
william-gr, line 24: For these I think we would like an indication wether the host was added automatically or manually?

yocalebo:
Just FYI, I merged a PR (https://github.com/freenas/freenas/pull/6038) that migrates db so you'll need to change your migration.

themylogin:
@william-gr ping

Ornias1993:
@travisghansen you might want to take a look at this PR...

travisghansen:
Interesting. The referenced issue doesn’t have much info. Can someone help me understand the driving factor behind this? I see how it plays out with scst config file but not sure what the motivation is.

themylogin:
@travisghansen https://jira.ixsystems.com/browse/NAS-107663

travisghansen:
Thanks for the link...it doesn’t say *why* it only says it needs to happen. Do you have any more insight beyond that?

src/middlewared/middlewared/etc_files/scst.conf.mako
```diff
@@ -123,6 +131,11 @@ TARGET_DRIVER iscsi {
 %   for access_control in initiator_portal_access:
```
yocalebo, line 131: Do we know how this behaves if you put a network segment in the webUI? (i.e. `192.168.1.0/24`). Currently, the config will ignore this and write a config entry like so `INITIATOR *\#<host-ip>` The `*` meaning all initiators are allowed to connect.

However, we're writing another entry that includes the `iqn` of the host. Does that mean all initiators are allowed to connect as well as the specific iqn?

themylogin:
@travisghansen the idea is to track initiator IQNs as they connect (because they are random and hard to manage manually). It's more a UI convenience issue.

src/middlewared/middlewared/etc_files/scst.conf.mako
```diff
@@ -45,6 +45,14 @@
 
     # FIXME: SSD is not being reflected in the initiator, please look into it
     # FIXME: Authorized networks for initiators has not been implemented yet, please look for alternatives in SCST
+
+    target_hosts = defaultdict(list)
+    for row in middleware.call_sync('datastore.query', 'services.iscsihosttarget'):
```
william-gr, line 50: Do we need direct datastore query here?

I try to avoid datastore.query as its a "private" API and use the iscsi service interfaces

src/middlewared/middlewared/plugins/iscsi_/host_crud.py
```diff
@@ -0,0 +1,210 @@
...
+                    if host["iqn"] not in db_host["iqns"]:
+                        await self.update_unlocked(db_host["id"], {"iqns": db_host["iqns"] + [host["iqn"]]})
+
+            except Exception:
+                await self.read_cache()
+                raise
+
+
+async def setup(middleware):
+    async with LOCK:
```
william-gr, line 209: Is this LOCK needed as we are doing it on setup?

src/middlewared/middlewared/plugins/iscsi_/host_injection.py
```diff
@@ -0,0 +1,67 @@
...
+
+CollectedHost = namedtuple("CollectedHost", ["ip", "iqn"])
+
+
+class iSCSIHostsInjectionService(Service):
+    class Config:
+        namespace = "iscsi.host.injection"
+        private = True
+
+    async def run(self):
```
william-gr, line 16: Does this means this will run forever, even if iSCSI service is not running?

src/middlewared/middlewared/etc_files/scst.conf.mako
themylogin: `auth_network` is not used anywhere in `scst.conf` because
```
# FIXME: Authorized networks for initiators has not been implemented yet, please look for alternatives in SCST
```

src/middlewared/middlewared/plugins/service_/services/iscsitarget.py
```diff
@@ -14,6 +14,12 @@ class ISCSITargetService(SimpleService):
 
     systemd_unit = "scst"
 
+    async def after_start(self):
+        await self.middleware.call("iscsi.host.injection.start")
```
william-gr, line 18: I dont recall how the services are started in SCALE. Are these methods triggered when service is started by systemd?

themylogin: No, they are not. We rely on the fact that user does not use scripts or systemctl to start/stop services and always does so through the middleware.

william-gr: Ok, but what starts scst at boot?

themylogin: Systemd starts it. And we have this https://github.com/freenas/freenas/pull/5979/files#diff-15237d4d3fd65ead3be91fce6291c116db8681d11e69f8afdc52eb79a06295b2R106

william-gr: I am a bit confused, I am rusty, sorry :)

How does that diff you pasted helps? middlewared starts before scst so doing it on setup should be a no-op unless we restart middleware

themylogin: The condition there is `service["enable"] or service["state"] == "RUNNING":`

`service["enable"]` — this means that systemd will start scst shortly
`service["state"] == "RUNNING":` — this is mostly for us, developers, when we restart middleware after the system has already booted

william-gr: Ah makes sense, thanks for explaining

william-gr (approved):

william-gr:
@yocalebo  any remaining concerns?

travisghansen:
I'm still interesting in knowing what this is actually achieving that we don't already have? LUN masking is already available so what is this adding on top of that?

src/middlewared/middlewared/etc_files/scst.conf.mako
yocalebo: That's what I'm saying, because we aren't using `auth_network` we already write an `INITIATOR` entry that looks like `*\#<host-ip>`

Your change  will write another `INITIATOR` entry that has `<iqn>\#<host-ip>`. So I'm trying to understand which entry is the correct one? Yours or the `*` entry.

themylogin: @yocalebo good catch! Latest commit fixes this

yocalebo: nice!

yocalebo (approved):

william-gr (approved):

PR #5981, opened 2020-11-06
NAS-108223 / 20.12 / filter .glusterfs system dataset from pool.dataset.query

yocalebo:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2675,6 +2675,15 @@ def query(self, filters=None, options=None):
                 ['id', '!^', f'{sys_config["basename"]}/'],
             ])
 
+            # top level dataset that stores all things related to gluster config
+            # needs to be hidden from local webUI since TrueCommander is
+            # responsible for configuring/managing gluster related settings
+            gluster_dataset = sys_config['basename'].split('/')[0] + '/.glusterfs'
```
william-gr, line 2681: We could use `pool` instead of `basename`?

yocalebo: Now that you bring it up, this wasn't proper anyways since the `.glusterfs` top-level dataset always exist on the zpool. Anyways, I've simplified the logic.

PR #5984, opened 2020-11-07
Telegram alert service

dafo90:
I added the support for the Telegram alert service

### DIFF
--- a/src/middlewared/middlewared/alert/service/telegram.py
+++ b/src/middlewared/middlewared/alert/service/telegram.py
@@ -1,0 +1,32 @@
+import json
+import requests
+
+from middlewared.alert.base import ThreadedAlertService
+from middlewared.schema import Dict, Str, List, Int
+
+
+class TelegramAlertService(ThreadedAlertService):
+    title = "Telegram"
+
+    schema = Dict(
+        "telegram_attributes",
+        Str("bot_token", required=True, empty=False),
+        List("chat_ids", required=True, empty=False, items=[Int("chat_id")]),
+        strict=True,
+    )
+
+    def send_sync(self, alerts, gone_alerts, new_alerts):
+        token = self.attributes["bot_token"]
+        chat_ids = self.attributes["chat_ids"]
+        for chat_id in chat_ids:
+            r = requests.post(
+                f"https://api.telegram.org/bot{token}/sendMessage",
+                headers={"Content-type": "application/json"},
+                data=json.dumps({
+                    "chat_id": chat_id,
+                    "text": self._format_alerts(alerts, gone_alerts, new_alerts),
+                    "parse_mode": "Markdown",
+                }),
+                timeout=15,
+            )
+            r.raise_for_status()


src/middlewared/middlewared/alert/service/telegram.py
```diff
@@ -0,0 +1,32 @@
...
+    schema = Dict(
+        "telegram_attributes",
+        Str("bot_token", required=True, empty=False),
+        Str("chat_ids", required=True, empty=False),
+        strict=True,
+    )
+
+    def send_sync(self, alerts, gone_alerts, new_alerts):
+        token = self.attributes["bot_token"]
+        chatIds = [cId.strip() for cId in self.attributes["chat_ids"].split(",")]
```
themylogin, line 20: Can we use PEP8-compliant names? (i.e. `chat_ids`)

Also can we please try to be more friendly and accept more separators (i.e. simply whitespaces)
```
[chat_id.strip() for chat_id in self.attributes["chat_ids"].replace(",", " ").replace(";", " ").split() if chat_id.strip()]
```

themylogin (approved):
Great, thank you very much! Can you please address the requests and also squash all you work into a single commit?

src/middlewared/middlewared/alert/service/telegram.py
```diff
@@ -0,0 +1,33 @@
...
+    schema = Dict(
+        "telegram_attributes",
+        Str("bot_token", required=True, empty=False),
+        Str("chat_ids", required=True, empty=False),
+        strict=True,
+    )
+
+    def send_sync(self, alerts, gone_alerts, new_alerts):
+        token = self.attributes["bot_token"]
+        chat_ids = [c_id.strip() for c_id in re.findall(r"-?[\d']+", self.attributes["chat_ids"])]
```
themylogin, line 21: Looks great now, thank you! But why is `'` included? Also you won't need `.strip()` because your regex does not match any whitespace characters

src/middlewared/middlewared/alert/service/telegram.py
```diff
@@ -0,0 +1,33 @@
...
+from middlewared.schema import Dict, Str
+
+
+class TelegramAlertService(ThreadedAlertService):
+    title = "Telegram"
+
+    schema = Dict(
+        "telegram_attributes",
+        Str("bot_token", required=True, empty=False),
+        Str("chat_ids", required=True, empty=False),
```
william-gr, line 15: Shouldnt this be a List?

src/middlewared/middlewared/alert/service/telegram.py
dafo90: Yeah, you're right! I fix these two things. Thanks for the tips.

themylogin: Sorry, can you please use `List("chat_ids", required=True, empty=False, items=[Int("chat_id")]),` instead and remove all the parsing? We'll do it in the UI instead.

themylogin (approved):
LGTM, thank you!

themylogin (changes requested):
@dafo90 actually you've forgot to import `Int`

themylogin (approved):

themylogin:
@dafo90 thank you!

dafo90:
I would like to integrate this feature into the webui repository, but I need any explanation... Can someone help me?

themylogin:
@dafo90 I've opened an UI issue in our bug tracker https://jira.ixsystems.com/browse/NAS-108271

I guess if you search for something like "slack" or "victorops" in https://github.com/freenas/webui you'll see what you need to copypaste in order to add telegram service.

dafo90:
I have a doubt about this feature, the method `_format_alerts` what type of string does produce? HTML or something else? I hardcoded `Markdown` but now I'm not sure it's the right choice

themylogin:
@dafo90 markdown is the right choice

PR #5994, opened 2020-11-10
NAS-107384 / 20.12 / Do not allow ldap users to use SSH if 2fa is setup

sonicaj:


william-gr:
I am not PAM expert, leaving it up to @anodos325

anodos325:
Looks reasonable.

anodos325 (approved):

PR #6001, opened 2020-11-12
NAS-107187 / 12.0 / Allow joining pools in disk.query

themylogin:


src/middlewared/middlewared/plugins/disk.py
```diff
@@ -98,13 +98,29 @@ async def disk_extend(self, disk, context):
         else:
             disk.pop('passwd')
             disk.pop('kmip_uid')
+        disk['pool'] = context['zfs_guid_to_pool'].get(disk['zfs_guid'])
         return disk
 
     @private
     async def disk_extend_context(self, extra):
```
william-gr, line 105: Should we document these extra options in disk.query docstring?

william-gr (approved):

PR #6012, opened 2020-11-13
NAS-107694 / 20.12 / Allow to update `autotrim` pool property

william-gr:
Ticket:	NAS-107694

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -468,6 +468,7 @@ class PoolService(CRUDService):
                 'topology': self.transform_topology(zpool['groups']),
                 'healthy': zpool['healthy'],
                 'status_detail': zpool['status_detail'],
+                'autotrim': zpool['properties']['autotrim'],
             })
         else:
             pool.update({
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -476,6 +477,7 @@ def pool_extend(self, pool):
                 'topology': None,
                 'healthy': False,
                 'status_detail': None,
+                'autotrim': {},
```
themylogin, line 480: Why `{}`?, isn't it on/off?

william-gr: @themylogin no, its an dict like other properties: {"parsed": "XXX", "value": "XX", "inherit": "XXX"}

themylogin (approved):

src/middlewared/middlewared/plugins/pool.py
themylogin: But what does `{}` mean then? Does py-libzfs return values like this? IMO this will be erroneously used like `pool["autotrim"]["value"]` and it does not matter if we crash on `KeyError("autotrim")` or `KeyError("value")`

PR #6013, opened 2020-11-13
NAS-108294 / 12.0 / fix HA network validation

yocalebo:


src/middlewared/middlewared/plugins/network.py
```diff
@@ -1094,74 +1094,74 @@ async def _common_validation(self, verrors, schema_name, data, itype, update=Non
...
-                    configured_attrs = set([i for i in failover_attrs if data.get(i)])
+            # have to make sure that active, standby and virtual ip addresses are equal
+            active_node_ips = len(data.get('aliases', []))
+            standby_node_ips = len(data.get('failover_aliases', []))
+            virtual_node_ips = len(data.get('failover_virtual_aliases', []))
+            are_equal = active_node_ips == standby_node_ips == virtual_node_ips
+            if not are_equal:
+                verrors.add(
+                    f'{schema_name}.failover_aliases',
+                    'You must specify an active, standby and virtual IP address.'
```
william-gr, line 1105: Thats a bit different than reality the way I see it. Should probably better word it, noting the same number is expected between the 3 of them?

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1094,74 +1094,74 @@ async def _common_validation(self, verrors, schema_name, data, itype, update=Non
...
-                    verrors.add(
-                        f'{schema_name}.failover_aliases',
-                        'You must specify an active, standby and virtual IP address.'
-                    )
-
-                if not update:
-                    failover_attrs = set(
-                        [k for k, v in validation_attrs.items() if k not in ('mtu', 'ipv4_dhcp', 'ipv6_auto')]
-                    )
-                    configured_attrs = set([i for i in failover_attrs if data.get(i)])
```
william-gr, line 1112: I dont quite understand why this was moved out of the if block?

yocalebo: Everything had to be moved outside of the `if ha_configured and not failover['disabled']` statement or none of this validation will run.

william-gr: Ohhh, I thought it was outside `if not await self.middleware.call('failover.licensed')` block

src/middlewared/middlewared/plugins/network.py
yocalebo: Pushed a change in verbiage.

william-gr:
Thanks

PR #6015, opened 2020-11-13
NAS-108824 / 12.0 / fix issues with scan_vrrp

yocalebo:
Fix 2 main issues.

- run the `interface.scan_vrrp` method on newly created interfaces (as well as interfaces whose VHID is being updated)
- fix the verror verbiage when a duplicate VHID is detected on the network

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -1149,17 +1149,36 @@ class InterfaceService(CRUDService):
                             'A lagg interface using the "Failover" protocol is not allowed to be marked critical for failover.'
                         )
 
-            if update.get('failover_vhid') != data['failover_vhid']:
+            # We check to see if a VHID is being used on the external network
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -1149,17 +1149,36 @@ async def _common_validation(self, verrors, schema_name, data, itype, update=Non
                             'A lagg interface using the "Failover" protocol is not allowed to be marked critical for failover.'
                         )
 
-            if update.get('failover_vhid') != data['failover_vhid']:
```
william-gr, line 1152: Isnt this the same? In a new interface, data['failover_vhid'] will be null and failover_vhid will be XX?

yocalebo: A new interface will have data['failover_vhid'] as XX and update object will be NULL.

However, you're correct in the fact that the logic is the same. This commit is doing nothing but adding documentation (via comments) so I'll close.

PR #6017, opened 2020-11-14
NAS-100018 / 20.12 / pool.replace preserve_settings

themylogin:


src/middlewared/middlewared/plugins/pool_/replace_disk.py
```diff
@@ -16,6 +17,7 @@ class PoolService(Service):
         Str('disk', required=True),
         Bool('force', default=False),
         Str('passphrase', private=True),
+        Bool('preserve_settings', default=False),
```
william-gr, line 20: I think this should probably be True?

themylogin: So we change default behavior? I've thought we'll add checkbox in UI and let user decide.

william-gr: It was not a default behavior, it was a missing feature, thats the expected behavior.

william-gr (approved):

PR #6018, opened 2020-11-15
NAS-108305 / 20.12 / Allow using latest as a valid item version for catalog items

sonicaj:
This PR aims to introduce `latest` string as a valid item version string for creating / upgrading chart releases. Motivation is to make it users to deploy latest version of a catalog item without looking up what the latest version actually is.

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/items.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/items.py
@@ -65,7 +65,9 @@ class CatalogService(Service):
 
         item_data.update({k: item_data.get(k) for k in ITEM_KEYS})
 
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/catalogs_linux/items.py
```diff
@@ -65,7 +65,9 @@ def item_details(self, item_path):
 
         item_data.update({k: item_data.get(k) for k in ITEM_KEYS})
 
-        for version in filter(lambda p: os.path.isdir(os.path.join(item_path, p)), os.listdir(item_path)):
+        for version in sorted(
+            filter(lambda p: os.path.isdir(os.path.join(item_path, p)), os.listdir(item_path)), reverse=True
```
themylogin, line 69: Does default comparator sort versions right?

themylogin (approved):

PR #6020, opened 2020-11-15
NAS-108309 / 20.12 / Use safe load when reading yaml files

sonicaj:
Not specifying a loader when reading yaml is deprecated and is affected by security issues. This commit changes that behaviour to use FullLoader as a loader for loading yaml files

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/items.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/items.py
@@ -61,7 +61,7 @@ class CatalogService(Service):
         # for each version available under the item
         item_data = {'versions': {}}
         with open(os.path.join(item_path, 'item.yaml'), 'r') as f:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/catalogs_linux/items.py
```diff
@@ -61,7 +61,7 @@ def item_details(self, item_path):
         # for each version available under the item
         item_data = {'versions': {}}
         with open(os.path.join(item_path, 'item.yaml'), 'r') as f:
-            item_data.update(yaml.load(f.read()))
+            item_data.update(yaml.load(f.read(), Loader=yaml.FullLoader))
```
themylogin, line 64: Why not just use `yaml.safe_load`?

themylogin (changes requested):

themylogin (approved):

PR #6023, opened 2020-11-16
fix JournaSync thread on HA systems

yocalebo:
If a `sqlite3.IntegrityError` (wrapped in a `CallError`) raised in the `JournalSync` thread occurs, then it leads to the simple fact that we will never update the standby node with  the database changes that occur on the active node. The reason why is because we process the HEAD of the `/data/ha-journal` list so when the `process()` method circles back around and processes the entry in `/data/ha-journal` that caueses the integrity error, it simply loops back to the beginning and we begin the cycle all over. This is all the while the `/data/ha-journal` is continually being appended to at the same time.

In this certain scenario, I feel it's best that we do 2 things:
1. obviously clear out the `/data/ha-journal` file to remove any broken entries
2. and run `send_database` to copy over a fresh db to the standby node


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1373,11 +1373,41 @@ def _flush_journal(self):
...
+                        # entries get an opportunity to run so we append more entries to
+                        # `/data/ha-journal`. This is extremely painful on HA systems because,
+                        # for example, if a user disables HA (to perform maintenance) the standby
+                        # node doesn't get this request and still thinks HA is enabled between the
+                        # nodes. Ultimately, this leads to production outages/unpredictable behavior
+                        # and databases that do not match between the nodes.
+
+                        # sqlite3.IntegrityError is particulary bad so clear out the journal
+                        # and sync a fresh database to the other side.
+                        self.journal.clear()
```
themylogin, line 1398: Non-empty journal was a sign that remote node is out of sync, so we continued trying to send the database if an attempt failed, and we also were able to remember that the journal is not synchronized across reboots. How is this behavior preserved now as we clear the journal?

yocalebo: I can remove the `self.journal.clear()` call since by calling `failover.send_database` actually does that for us.

themylogin: We should still account for that behavior change and store some flag (e.g. `/data/ha-journal-not-synchronized`)

yocalebo: If I remove the `self.journal.clear()` call, I don't see how this is a behavior change. If you remove my newly added logic and check a few lines above we actually create a oneshot alert in the event of a failure that just tells the end-user to "Manually sync to peer". Instead of alerting the user in the event of a sqlite3.IntegrityError, I send the database for them since that entry in `/data/ha-journal` will hold up all other operations.

themylogin (approved):
I get it now, thank you for elaborating!

yocalebo:
Tired of this being in my stack so closing since it has rotted for too long.

PR #6024, opened 2020-11-17
NAS-107830 / 12.0 / reporting.processes event

themylogin:
@william-gr I took the liberty to implement this as an event source in order to be able to track CPU usage.

Average iteration time is `0.008` sec, so it is 0.08% of one CPU core if running every 10 seconds.

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/processes.py
+++ b/src/middlewared/middlewared/plugins/reporting/processes.py
@@ -1,0 +1,71 @@
+import json
+import time
+
+import psutil
+
+from middlewared.event import EventSource
+from middlewared.service import CallError
+from middlewared.utils import osc
+
+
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/reporting/processes.py
```diff
@@ -0,0 +1,67 @@
...
+    Retrieve currently running processes stats.
+
+    Usage: reporting.processes:{"interval": 10, "cpu_percent": 0.1, "memory_percent": 0.1}
+    """
+
+    def run(self):
+        options = {}
+        if self.arg:
+            options = json.loads(self.arg)
+        options.setdefault("interval", 10)
```
william-gr, line 21: Looks good! Lets make sure internal is not smaller than 5

william-gr (approved):

PR #6026, opened 2020-11-17
NAS-108316 / 20.12 / ensure snmp starts on passive nodes on HA systems

yocalebo:


william-gr (commented):
Not sure its a good idea to enable SNMP in the passive... What is the intent here?

PR #6030, opened 2020-11-17
NAS-108093 / 12.0 / Create VM context if libvirt fails to poweroff/shutdown vm in bhyve

sonicaj:


src/middlewared/middlewared/plugins/vm.py
```diff
@@ -270,7 +276,23 @@ def restart(self, vm_data=None, shutdown_timeout=None):
...
+        except libvirt.libvirtError:
+            if not self.destroy_vm_context():
+                raise
+
+    def destroy_vm_context(self):
+        # There is a race condition where poweroff/shutdown fail sometimes in libvirt which results in it trying
+        # to destroy a VM whose context does not exist resulting in user having to reboot their machines in order
+        # to get the VM back.
+        # This is a workaround where we create a vm context if it doesn't exist so stop/poweroff work
+        if os.path.exists('/dev/vmm') and self.libvirt_domain_name in os.listdir('/dev/vmm'):
```
william-gr, line 290: Did you mean libvirt_domain_name `not in`?

william-gr: Guess not, nvm

sonicaj: Nope, if the context is there - then the error is different which we have been seeing reported by users

sonicaj: In that case we should raise whatever the error is

themylogin (approved):

PR #6040, opened 2020-11-19
NAS-104668 / 20.12 / Support SED disks in Scale

sonicaj:


themylogin (changes requested):
Why don't you make `disk_/sed_freebsd.py` and `disk_/sed_linux.py`?

sonicaj:
Primarily because of the diff. Ideally I wanted to have sed in a separate plugin part. I can update the PR with the changes : )

themylogin:
We are not getting a good diff for this method anyway, so let's move at least this one method. For others it's probably better to stay where they are now.

src/middlewared/middlewared/plugins/disk.py
```diff
@@ -396,26 +392,11 @@ async def sed_unlock(self, disk_name, disk=None, _advconfig=None):
...
-                        locked = False
-                        unlocked = True
-                else:
-                    locked = False
+            locked, unlocked = await self.middleware.call('disk.unlock_ata_security', devname, _advconfig, password)
 
         if unlocked:
             try:
-                # Disk needs to be retasted after unlock
+                # Disk needs to be retested after unlock
```
themylogin, line 399: It was not a typo, it's really `GEOM retaste`. Is it needed on Linux by the way?

themylogin (dismissed):

src/freenas/usr/local/bin/truenas-grub.py
```diff
@@ -21,9 +21,10 @@ def dict_factory(cursor, row):
     c.execute("SELECT * FROM system_advanced")
     advanced = {k.replace("adv_", ""): v for k, v in c.fetchone().items()}
 
+    # We need to allow tpm in grub as sedutil-cli requires it
     config = [
         'GRUB_DISTRIBUTOR="TrueNAS Scale"',
-        'GRUB_CMDLINE_LINUX_DEFAULT=""',
+        'GRUB_CMDLINE_LINUX_DEFAULT="libata.allow_tpm=1"',
```
william-gr, line 27: This is also possible to control via a /proc or /sys (dont remember which). Wonder if its better to keep it in t he plugin to keep it self-contained?

william-gr (approved):

PR #6053, opened 2020-11-21
NAS-108354 / 20.12 / Add validation to ensure valid mac address is provided for VM nic device

sonicaj:
We only allow `:` as valid separators for mac address validation because libvirt only considers them as valid and fails otherwise to properly initialise the domain.

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/devices/nic.py
+++ b/src/middlewared/middlewared/plugins/vm/devices/nic.py
@@ -3,6 +3,7 @@ import random
 from middlewared.plugins.interface.netif import netif
 from middlewared.schema import Dict, Str
 from middlewared.service import CallError
+from middlewared.validators import MACAddr
 
 from .device import Device
 from .utils import create_element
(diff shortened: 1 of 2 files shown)


sonicaj:
backport

src/middlewared/middlewared/plugins/vm/vm_devices.py
```diff
@@ -363,6 +366,9 @@ async def validate_device(self, device, old=None, vm_instance=None):
                 nic_choices = await self.middleware.call('vm.device.nic_attach_choices')
                 if nic not in nic_choices:
                     verrors.add('attributes.nic_attach', 'Not a valid choice.')
+            mac = device['attributes'].get('mac')
+            if mac and not RE_MAC_CHECK.search(mac):
```
themylogin, line 370: Can't we just use `MACAddr` validator on the schema field?

themylogin (changes requested):

themylogin (approved):

william-gr (approved):

PR #6056, opened 2020-11-22
NAS-108367 / 20.12 / sharing:smb - add special behavior for CLUSTERED smb_ha_mode

anodos325:
Add registry wrappers for sharing.smb CRUD interface so that
datastore is bypassed when ctdb is used. ctdb will ensure that SMB share
configuration is consistent cluster-wide. This is an intermediate
step before transitioning the [global] section to registry as well.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -1,7 +1,7 @@
 from middlewared.common.attachment import LockableFSAttachmentDelegate
 from middlewared.common.listen import SystemServiceListenMultipleDelegate
 from middlewared.schema import Bool, Dict, IPAddr, List, Str, Int, Patch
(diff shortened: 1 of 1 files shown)


yocalebo (approved):

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -1003,6 +1044,31 @@ async def do_delete(self, id):
...
         return result
 
+    @filterable
+    async def query(self, filters=None, options=None):
+        """
+        Query shares with filters. In clustered environments, local datastore query
+        is bypassed in favor of clustered registry.
+        """
+        extra = options.get('extra', {})
+        try:
```
william-gr, line 1054: Shouldnt we make sure ha_mode exists? Letting a KeyError be raised on None almost every time seems bad

PR #6057, opened 2020-11-23
NAS-107243 / 21.04 / Allow properly isolating GPU from host for passthrough to VMs

sonicaj:

1) Allows specifying PCI ids of GPUs which the system will remove from host usage after a reboot
2) Add field to allow hiding guest from msr based discovery
3) Do not render unavailable devices when generating domain xml of a guest
4) Bug fix for locating PCI address of a PCI device for passthrough
5) Improve performance while starting VM by not looking up all PCI devices


yocalebo:
Just FYI, I merged a PR (https://github.com/freenas/freenas/pull/6038) that migrates db so you'll need to change your migration.

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -253,23 +252,48 @@ def get_storage_devices_topology(self):
...
-        if self.GPU:
-            return self.GPU
-
-        not_available = {'available': False, 'vendor': None}
+    def get_gpus(self):
+        cp = subprocess.Popen(['lspci', '-D'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        stdout, stderr = cp.communicate()
+        if cp.returncode:
+            raise CallError(f'Unable to list available gpus: {stderr.decode()}')
+
```
yocalebo, line 260: Why are we no longer caching the GPU information?

sonicaj: We would need to clear the cache each time a user updated his gpu configuration, i am not sure if the management ( or keeping it in sync ) of keeping the cache fresh is better then just not using cache at all as this method is not going to be called very often.

yocalebo: Are GPUs hot-pluggable? Not sure how a GPU configuration can be updated without shutting down the PC.

sonicaj: They could be but we are not looking to support that. I am looking to make sure we have updated information being displayed to the user if a GPU can be used by host or not. So user changing configuration means user wanting to isolate a GPU from the host and we reflecting that here.

yocalebo: If we're not going to support hot-pluggable GPUs, then caching this data seems to make more sense.

sonicaj: Yes but keeping the cache in sync for one subprocess call which is also not really time/resource intensive ? Do you think  we should be doing that ?

sonicaj: I have no issues doing it btw :p Just raising my concern, we can do that of course or maybe even handle it differently by updating `available_to_host` field each time on cache - that seems like a good compromise btw 🤔

yocalebo: I was just going to say maybe put `GPUS = None` in the `DeviceService` class. And then update that variable with the `get_gpus` method if `GPUS = None` else just return `self.GPUS`.

sonicaj: So that worked if the information did not change through out the uptime of the system, but that changes. Specifically `available_to_host` field changes.
I think a fair compromise is to update `self.GPUS` always with logic in place for `available_to_host` field and use cache.

yocalebo: I didn't realize that attribute changes after boot (seems weird btw). In that case, probably best to just not cache like you originally did and move on. Not worth trying to solve cache validation/invalidation issues especially if this method isn't called often and isn't intensive.

yocalebo: Thanks for explanation.

themylogin (changes requested):
A lot of time has passed since we discussed this. There already is a script that handles initrd configuration (/usr/local/bin/initrd.zfs.py). Let's rename and extend it and make a single entry point for all initrd updates.

src/freenas/usr/local/bin/truenas-initrd.py
```diff
@@ -0,0 +1,159 @@
...
+def update_zfs_default(root):
+    with libzfs.ZFS() as zfs:
+        existing_pools = [p.name for p in zfs.pools]
+
+    for i in ['freenas-boot', 'boot-pool']:
+        if i in existing_pools:
+            boot_pool = i
+            break
+    else:
+        logger.error('Failed to locate valid boot pool. Pools located were: %r', ', '.join(existing_pools))
```
themylogin, line 26: Why is this not a fatal error?

src/freenas/usr/local/bin/truenas-initrd.py
```diff
@@ -0,0 +1,159 @@
...
+
+def dict_factory(cursor, row):
+    d = {}
+    for idx, col in enumerate(cursor.description):
+        d[col[0]] = row[idx]
+    return d
+
+
+def get_current_gpu_pci_ids(root):
+    from middlewared.plugins.config import FREENAS_DATABASE
```
themylogin, line 76: Why late import?

src/freenas/usr/local/bin/truenas-initrd.py
```diff
@@ -0,0 +1,159 @@
...
+
+    new_config = "\n".join(lines) + "\n"
+    if new_config != original_config:
+        with open(zfs_config_path, "w") as f:
+            f.write(new_config)
+        return True
+    return False
+
+
+def dict_factory(cursor, row):
```
themylogin, line 68: Let's put this to some middleware.utils module to avoid copy-pasting to every standalone console script we write

src/freenas/usr/local/bin/truenas-initrd.py
```diff
@@ -0,0 +1,159 @@
...
+
+    for path in map(get_path, ['etc/initramfs-tools/modules', 'etc/modules']):
+        with open(path, 'w') as f:
+            f.write(f'vfio\nvfio_iommu_type1\nvfio_virqfd\nvfio_pci ids={",".join(pci_ids)}\n')
+
+    with open(get_path('etc/modprobe.d/kvm.conf'), 'w') as f:
+        f.write('options kvm ignore_msrs=1\n')
+
+    with open(get_path('etc/modprobe.d/nvidia.conf'), 'w') as f:
+        f.write('softdep nouveau pre: vfio-pci\nsoftdep nvidia pre: vfio-pci\nsoftdep nvidia* pre: vfio-pci\n')
```
themylogin, line 117:
```
    f.write(textwrap.dedend('''\
        softdep nouveau pre: vfio-pci
        softdep nvidia pre: vfio-pci
```

src/freenas/usr/local/bin/truenas-initrd.py
```diff
@@ -0,0 +1,159 @@
...
+        update_module_files(root, initramfs_config)
+        return True
+
+    return False
+
+
+if __name__ == "__main__":
+    root = sys.argv[1]
+    if root != "/":
+        sys.path.append(os.path.join(root, "usr/lib/python3/dist-packages/middlewared"))
```
themylogin, line 146: Should it be `sys.path.insert(0, ` to give it a priority if we are running from some root where we already have middlewared?

src/middlewared/middlewared/etc_files/initramfs_update.py
```diff
@@ -0,0 +1,7 @@
+from middlewared.utils import run
+
+
+async def render(service, middleware):
+    await run(
+        '/usr/local/bin/truenas-initrd.py', '/', encoding='utf8', errors='ignore', check=False
```
themylogin, line 6: I don't understand why this is an etc plugin and not a `boot` service method like it was before, what configuration are we generating here?

themylogin (changes requested):

src/middlewared/middlewared/utils/db.py
```diff
@@ -0,0 +1,5 @@
+def dict_factory(cursor, row):
```
themylogin, line 1: We can have a common function for connecting to SQLite and querying config row as well

src/freenas/usr/local/bin/truenas-initrd.py
```diff
@@ -88,36 +79,45 @@ def update_module_files(root, config):
...
+    os.makedirs(get_path("etc/initramfs-tools"), exist_ok=True)
+    os.makedirs(get_path("etc/modprobe.d"), exist_ok=True)
 
-    for path in map(get_path, ['etc/initramfs-tools/modules', 'etc/modules']):
-        with open(path, 'w') as f:
-            f.write(f'vfio\nvfio_iommu_type1\nvfio_virqfd\nvfio_pci ids={",".join(pci_ids)}\n')
+    for path in map(get_path, ["etc/initramfs-tools/modules", "etc/modules"]):
+        with open(path, "w") as f:
+            f.write(textwrap.dedent(f"""\
+            vfio
```
themylogin, line 103: One more indent level please

themylogin (changes requested):

themylogin (approved):
Thank you!

william-gr (approved):

PR #6065, opened 2020-11-24
NAS-108398 / 21.02 / NAS-108398 Remove suggestion of using Flash Media

q5sys:


freqlabs (approved):

william-gr:
I believe in SCALE we actually use another repo, truenas-installer

PR #6069, opened 2020-11-24
NAS-108401 / 20.12 / Fix migrating system dataset on SCALE

themylogin:


src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -431,6 +435,8 @@ async def migrate(self, _from, _to):
 
         if await self.middleware.call('service.started', 'cifs'):
             restart.insert(0, 'cifs')
+        if await self.middleware.call('service.started', 'webdav'):
```
sonicaj, line 438: Why do we want to restart `webdav` ?

sonicaj (approved):

src/middlewared/middlewared/plugins/sysdataset.py
themylogin: Apache keeps a log file open in `/var/log`, we need to stop it before umounting syslog dataset.

PR #6074, opened 2020-11-25
NAS-108320 / 12.0 / Create hint.isp.{i}.role=2 tunables on non-iX hardware

themylogin:


william-gr:
@amotin How does that look to you?

themylogin:
@amotin It will create `hint.isp.[0-4].role=2` tunables on non-iX hardware if they don't exist.

src/middlewared/middlewared/plugins/sysctl/hint_isp_role_freebsd.py
```diff
@@ -0,0 +1,28 @@
...
+        var = f"hint.isp.{i}.role"
+        if var in tunables:
+            continue
+        if not sysctl.filter(var):
+            continue
+
+        await middleware.call("tunable.create", {
+            "var": var,
+            "value": "2",
+            "type": "SYSCTL",
```
amotin, line 25: The "hint.*" are only tunables, not sysctl's.  Respective sysctl's are called "dev.isp...", but in this case use of tunables is preferable to make disks probed at boot.

themylogin: @amotin so they'll be `LOADER` tunables and we'll need a reboot to properly apply them?

william-gr: Ah right. Thats why I said we should prob do that at install time and on upgrade phase, so the extra reboot is not necessary

william-gr (approved):

william-gr (approved):

PR #6075, opened 2020-11-25
NAS-108411 / 20.12 / Ensure sharing task is extended in smb.query()

anodos325:


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -1065,14 +1065,26 @@ async def query(self, filters=None, options=None):
...
-                'datastore.query', self._config.datastore, filters, options
+                'datastore.query', self._config.datastore, [], datastore_options
+            )
+            return await self.middleware.run_in_thread(
+                filter_list, result, filters, options
+            )
+        else:
+            return await self.middleware.call(
+                'datastore.query', self._config.datastore, filters, options,
             )
```
william-gr, line 1087: Isnt it better to use super() here? Dont recall if this has already been fixed by anotehr PR

william-gr (changes requested):

src/middlewared/middlewared/plugins/smb.py
anodos325: Yeah, it was fixed in another PR by Waqar (happened right around thanksgiving).

PR #6078, opened 2020-11-25
NAS-108598 / 21.08 / Make idmap/smb/directory services cluster-aware

anodos325:
General change (clustered / non-clustered):
* The global portion of the smb4.conf will be moved to Samba's registry. This means that /etc/smb4.conf will now be a stub containing only two lines in non-clustered case:
```
[global]
    include = registry
```
There will be some fallout from this change. For instance, user comments will no longer appear in the smb4.conf. This also means that configuration changes from all plugins that used to impact generation of the smb4.conf file (idmap, activedirectory, ldap, idmap, smb)  will do this through `net conf setparm` and `net conf getparm`. `smb.reg_globals` queries the global portion of the smb4.conf and splits it up into parts customized for each plugin. 

In the non-clustered case, the update() methods will get this smb4.conf portion, convert the new data to an smb4.conf portion, get a diff, and apply it to the registry. idmap.delete() will also remove all smb4.conf parameters associated with the domain being deleted. Otherwise, normal datastore operations should not be impacted. 

When SMBHAMODE == CLUSTERED behavior for the above-described services will be changed as follows:
* Active Directory and LDAP configuration details will be split between the clustered secrets.tdb and registry.tdb files. The former is due to some items being unsuited to a world-readable file.
* The idmap plugin configuration details are entirely stored in the registry.tdb (as Samba's running config)
* The smb plugin configuration is likewise entirely stored in the registry.tdb file.
* The general procedure of updating configuration is same as non-clustered case with the omissions of writes via the datastore plugin. query() / config() calls for these plugins are likewise ultimately wrappers for tdb reads.

This basically ensures that SMB, Idmap, AD, and LDAP config are consistent cluster-wide. Further work will be needed for kerberos and local users and groups.

Added validation check to only permit one directory service to be enabled. This will vastly simplify configuration.


anodos325:
Fairly large changeset for shifting to base SMB service entirely off its registry (clustered / non-clustered case). I'll do more testing before merging, but initial feedback is appreciated.

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -804,3 +852,190 @@ async def unixid_to_sid(self, data):
...
+        security = global_params['raw'].get('security')
+
+        for k, v in global_params['idmap'].items():
+            if "*" in k:
+                domain = DSType.DS_TYPE_DEFAULT_DOMAIN.name
+            else:
+                domain = k.strip("idmap config ")
+                domain = domain.split(":")[0].strip()
+            entry = k.split(":", 1)[1]
+            self.logger.debug("k: %s, v %s, entry: %s", k, v, entry)
```
themylogin, line 872: Isn't it a leftover debug?

themylogin (changes requested):

src/middlewared/middlewared/plugins/idmap.py
anodos325: Absolutely. Thanks

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -380,6 +401,30 @@ async def common_validate(self, new, old, verrors):
...
+    @private
+    async def config(self):
+        ha_mode = SMBHAMODE[(await self.middleware.call('smb.get_smb_ha_mode'))]
+        if ha_mode != SMBHAMODE.CLUSTERED:
+            return await self.middleware.call(
+                'datastore.config',
+                self._config.datastore,
+                {'extend': self._config.datastore_extend,
+                 'prefix': self._config.datastore_prefix}
+            )
```
william-gr, line 413: Its probably more future-proof if we use super() here

return await super().config()

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -380,6 +401,30 @@ async def common_validate(self, new, old, verrors):
...
+        globals = (await self.middleware.call('smb.reg_globals'))['raw']
+        rv.update({
+            'domainname': globals.get('realm', ''),
+            'use_default_domain': globals.get('winbind use default domain', 'No') == 'Yes',
+            'allow_trusted_domains': globals.get('allow trusted domains', 'No') == 'Yes',
+            'allow_dns_updates': globals.get('ads dns updates', 'No') == 'Yes',
+            'nss_info': globals.get('winbind nss info', ''),
+            'enabled': globals.get('security', 'user') == 'ADS',
+        })
+        return await self.ad_extend(rv)
```
william-gr, line 426: await self.middleware.call(self._config.datastore_extend, rv) , perhaps?

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -526,14 +571,54 @@ async def do_update(self, data):
...
+        ad config to secrets.tdb.
+        """
+        if ha_mode != SMBHAMODE.CLUSTERED:
+            await self.middleware.call(
+                'datastore.update',
+                'directoryservice.activedirectory',
+                old['id'],
+                new,
+                {'prefix': 'ad_'}
+            )
```
william-gr, line 589: Same here about super()

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -724,11 +739,44 @@ async def do_update(self, id, data):
...
+        if ha_mode == SMBHAMODE.CLUSTERED:
+            result = await self.reg_query(filters, options)
+
+        else:
+            options['extend'] = self._config.datastore_extend
+            options['prefix'] = self._config.datastore_prefix
+
+            result = await self.middleware.call(
+                'datastore.query', self._config.datastore, filters, options
+            )
```
william-gr, line 777: super()

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -1061,6 +1107,7 @@ async def query(self, filters=None, options=None):
         else:
             ha_mode = SMBHAMODE[ha_mode_str]
 
+        self.logger.debug("hamode: %s", ha_mode.name)
```
william-gr, line 1110: another leftover

src/middlewared/middlewared/plugins/smb_/registry_share.py
```diff
@@ -372,6 +373,9 @@ async def share_to_smbconf(self, conf_in, globalconf=None):
         else:
             conf["nt acl support"] = "no"
 
+        if data["cluster_volname"]:
```
yocalebo, line 376: I've done an initial pass at reviewing this and syntactically, everything looks fine. I think this is where we should `cache.put SMB_HA_MODE` logic so the cache is updated making `smb_get_ha_mode` return clustered.

I'll review a lot more in depth tomorrow.

src/middlewared/middlewared/etc_files/krb5.keytab.py
```diff
@@ -59,3 +59,8 @@ async def render(service, middleware):
         db_keytabfile = base64.b64decode(keytab['file'].encode())
         db_keytabname = keytab['id']
         await write_keytab(db_keytabname, db_keytabfile)
+
+    if osc.IS_LINUX:
+        if os.path.exists(keytabfile):
```
themylogin, line 64: The best practice is
```
        with contextlib.suppress(OSError):
            os.unlink(keytabfile)
```
Because theoretically someone else might remove it between `os.path.exists` and `os.remove`

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -380,6 +400,25 @@ async def common_validate(self, new, old, verrors):
                 "AD domain name is required."
             )
 
+    @private
```
william-gr, line 403: This probably shouldnt be private?

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -649,7 +664,13 @@ async def do_update(self, id, data):
         """
         Update a domain by id.
         """
-        old = await self._get_instance(id)
+        ha_mode = SMBHAMODE[(await self.middleware.call('smb.get_smb_ha_mode'))]
+        if ha_mode == SMBHAMODE.CLUSTERED:
+            ctdb_healthy = await self.middleware.call('ctdb.general.healthy')
+            if not ctdb_healthy:
+                raise CallError("Idmap configuration may only be modified when cluster is healthy")
```
william-gr, line 671: We do that in many places, perhaps it could/should be a common method?

src/middlewared/middlewared/etc_files/krb5.keytab.py
anodos325: Thanks. I'll file that tip away.

yocalebo (approved):

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -380,6 +400,25 @@ async def common_validate(self, new, old, verrors):
...
+        rv = await self.middleware.call('directoryservices.get_conf',
+                                        self._config.datastore)
+        globals = (await self.middleware.call('smb.reg_globals'))['raw']
+        rv.update({
+            'domainname': globals.get('realm', ''),
+            'use_default_domain': globals.get('winbind use default domain', 'No') == 'Yes',
+            'allow_trusted_domains': globals.get('allow trusted domains', 'No') == 'Yes',
+            'ads_dns_updates': globals.get('ads dns updates', 'No') == 'Yes',
+            'nss_info': globals.get('winbind nss info', ''),
+            'enabled': globals.get('security', 'user') == 'ADS',
```
yocalebo, line 418: `enabled` needs to be `enable` or a `KeyError` will be raised on line 1689 as well as the `def config()` method in cache.py

src/middlewared/middlewared/plugins/directoryservices.py
```diff
@@ -344,6 +368,34 @@ def get_last_password_change(self, domain=None):
...
         return {"dbconfig": stored_ts, "secrets": passwd_ts}
 
+    @private
+    def get_conf(self, schema, ha_mode=None):
+        if ha_mode is None:
+            ha_mode = self.middleware.call_sync('smb.get_smb_ha_mode')
+        with DirectorySecrets(logger=self.logger, ha_mode=ha_mode) as s:
+            conf = s.get_conf(schema)
+
+        rv = conf.decode()
```
yocalebo, line 378: This will raise an `AttributeError` if rv is `None`.

```suggestion
        rv = conf.decode() if conf is not None else {}
```

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -724,11 +742,40 @@ async def do_update(self, id, data):
...
-        await self.middleware.call("datastore.delete", self._config.datastore, id)
+
+        ha_mode = SMBHAMODE[(await self.middleware.call('smb.get_smb_ha_mode'))]
+        if ha_mode != SMBHAMODE.CLUSTERED:
+            return await self.middleware.call("datastore.delete", self._config.datastore, id)
+
+        return await self.reg_delete(id)
+
+    @filterable
+    async def query(self, filters=None, options=None):
```
yocalebo, line 760: This will cause the following error: `ValueError: All public method default arguments should be specified in @accepts()`

```suggestion
    async def query(self, filters, options):
```

src/middlewared/middlewared/plugins/smb_/registry_global.py
```diff
@@ -0,0 +1,297 @@
...
+        reg_globals.pop("bind interfaces only", "")
+        loglevel = reg_globals.pop("log level", "1")
+        if loglevel.startswith("syslog@"):
+            loglevel = loglevel[len("syslog@")]
+
+        llevel = LOGLEVEL_MAP.get(loglevel.split()[0])
+
+        ret = {
+            "id": 1,
+            "netbiosname": reg_globals.pop("tn:netbiosname"),
```
yocalebo, line 120: On my box I'm getting multiple `KeyErrors`

```suggestion
            "netbiosname": reg_globals.pop("tn:netbiosname", "truenas"),
```

src/middlewared/middlewared/plugins/smb_/registry_global.py
```diff
@@ -0,0 +1,297 @@
...
+            "netbiosname_b": reg_globals.pop("tn:netbiosname_b"),
```
yocalebo, line 121: Same as above.
```suggestion
            "netbiosname_b": reg_globals.pop("tn:netbiosname_b", "truenas_b"),
```

src/middlewared/middlewared/plugins/smb_/registry_global.py
```diff
@@ -0,0 +1,297 @@
...
+            "netbiosname_local": reg_globals.pop("netbios name", ""),
+            "workgroup": reg_globals.pop("workgorup", "WORKGROUP"),
+            "cifs_SID": reg_globals.pop("tn:sid", None),
+            "netbiosalias": (reg_globals.pop("netbios aliases", "")).split(),
+            "description": reg_globals.pop("server string"),
```
yocalebo, line 126: Ditto here:
```suggestion
            "description": reg_globals.pop("server string", None),
```

src/middlewared/middlewared/plugins/smb_/registry_global.py
```diff
@@ -0,0 +1,297 @@
...
+            "localmaster": False,
+            "loglevel": llevel,
+            "guest": reg_globals.pop("guest account", "nobody"),
+            "admin_group": reg_globals.pop("tn:admin_group", ""),
+            "filemask": reg_globals.pop("create mask", "0775"),
+            "dirmask": reg_globals.pop("directory mask", "0775"),
+            "ntlmv1_auth": reg_globals.pop("ntlm auth", "No") == "Yes",
+            "bind ip": bind_ips,
+        }
+        reg_globals.pop('logging')
```
yocalebo, line 140: Ditto here:
```suggestion
        reg_globals.pop('logging', '')
```

yocalebo (changes requested):
Added a few comments

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -14,6 +14,7 @@
 import re
```
yocalebo, line 14: Not that you have altered the `def do_create` method in this plugin, I'm hitting a `KeyError` using the API directly, because I'm trying to create a SMB share without the `name` attribute. You call the `def clean()` method expecting that the `name` key being preset. Not sure how to handle that scenario? Do you require the name at the schema level? Or do you just `name = data.get('name', '')` in the `clean()` methods.

src/middlewared/middlewared/plugins/smb_/registry_share.py
```diff
@@ -372,6 +373,9 @@ async def share_to_smbconf(self, conf_in, globalconf=None):
...
+            conf["glusterfs: volume"] = data["cluster_volume"]
```
yocalebo, line 377: This will raise a `KeyError`
```suggestion
            conf["glusterfs: volume"] = data["cluster_volname"]
```

src/middlewared/middlewared/plugins/smb_/registry_share.py
```diff
@@ -66,7 +66,7 @@ async def netconf(self, **kwargs):
 
     @private
     async def reg_listshares(self):
```
yocalebo, line 68: When creating a fresh SMB share using the `cluster_volname` schema attribute, I was presented with this traceback. The share was created successfully, just raised that exception.

```  File "/usr/lib/python3/dist-packages/middlewared/plugins/smb.py", line 935, in do_create
    ret = await self.query([('name', '=', data['name'])],
  File "/usr/lib/python3/dist-packages/middlewared/schema.py", line 998, in nf
    return await f(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/middlewared/plugins/smb.py", line 1130, in query
    result = await self.middleware.call(
  File "/usr/lib/python3/dist-packages/middlewared/main.py", line 1263, in call
    return await self._call(
  File "/usr/lib/python3/dist-packages/middlewared/main.py", line 1220, in _call
  File "/usr/lib/python3/dist-packages/middlewared/schema.py", line 998, in nf
    return await f(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/middlewared/plugins/smb_/registry_share.py", line 287, in registry_query
    return filter_list(rv, filters, options)
  File "/usr/lib/python3/dist-packages/middlewared/utils/__init__.py", line 198, in filter_list
    raise MatchNotFound()
middlewared.service_exception.MatchNotFound

**Comment by yocalebo** [comment]
When starting the SMB service with these changes, I get the following error message from the daemon.
```
Feb 01 11:00:49 sc01.home.lan smbd[76650]: [2021/02/01 11:00:49.259677,  0] ../../source3/printing/print_standard.c:71(std_pcap_cache_reload)
Feb 01 11:00:49 sc01.home.lan smbd[76650]:   Unable to open printcap file /etc/printcap for read!
```

Seems like an easy fix is to add an entry in the `[global]` section of the config like:
```
printcap name = /dev/null
load printers = no
printing = bsd

But I have no idea as that information could be (and probably is) outdated.

src/middlewared/middlewared/plugins/smb_/registry_share.py
```diff
@@ -66,7 +66,7 @@ async def netconf(self, **kwargs):
...
-        return (await self.netconf(action='listshares')).splitlines()
+        return (await self.netconf(action='listshares')).splitlines()[1:]
```
yocalebo, line 69: Do you mean just `splitlines()`? Doing it this way will return an empty list if there is only 1 share added.

src/middlewared/middlewared/plugins/smb_/registry_share.py
yocalebo, line 68: Creating a fresh share raises a `KeyError` wrt to the `comment` key.

``` 
File "/usr/lib/python3.9/concurrent/futures/_base.py", line 389, in __get_result
    raise self._exception
  File "/usr/lib/python3/dist-packages/middlewared/plugins/smb_/registry_share.py", line 353, in share_to_smbconf
    if data['comment']:
KeyError: 'comment'
```

In the `share_to_smbconf` method, I'd change it to:
```
if data.get('comment', ''):
```

yocalebo, line 68: Another `KeyError` for the `durablehandle` key in `share_to_smbconf` method.
```
  File "/usr/lib/python3/dist-packages/middlewared/plugins/smb_/registry_share.py", line 391, in share_to_smbconf
    if data['durablehandle']:
KeyError: 'durablehandle'
```

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -428,6 +429,15 @@ async def import_conf_to_registry(self):
             self.logger.warning('failed to load share config: %s',
                                 load.stderr.decode())
 
+    @private
```
yocalebo, line 432: There is an edge-case dead lock here. So I feel like we should put a timeout on this `while True:` loop.

For example, if ctdb is started and has the `50.samba` event script enabled and, for example, the smbd service will not start, then we will deadlock here if the node that is causing the cluster to be unhealthy is the node that is running this.

A typical catch-22. So I would maybe put a timeout for 30 seconds (or whatever you think is best so that the `smb.configure` job will at least error out and give some hint on why it failed instead of deadlocking here.

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -1143,6 +1238,28 @@ async def validate_aux_params(self, data, schema_name):
...
+            verrors.add(
+                f'{schema_name}.shadowcopy',
+                'Shadow Copies are not implemented for clustered shares.'
+            )
+        if data['fsrvp']:
+            verrors.add(
+                f'{schema_name}.fsrvp',
+                'FSRVP support is not implemented for clustered shares.'
+            )
+        if not data['cluser_volname']:
```
yocalebo, line 1257: `KeyError` because of typo here

```suggestion
        if not data['cluster_volname']:
```

src/middlewared/middlewared/plugins/smb_/registry_share.py
```diff
@@ -343,6 +346,10 @@ async def share_to_smbconf(self, conf_in, globalconf=None):
 
         if osc.IS_FREEBSD:
             data['vfsobjects'] = ['aio_fbsd']
+        elif is_clustered:
+            conf["glusterfs: volume"] = data["cluster_volume"]
```
yocalebo, line 350: `KeyError` because of minor typo here:
```suggestion
            conf["glusterfs: volume"] = data["cluster_volname"]
```

src/middlewared/middlewared/plugins/smb_/registry_share.py
```diff
@@ -343,6 +346,10 @@ async def share_to_smbconf(self, conf_in, globalconf=None):
...
+            conf["glusterfs: logfile"] = f'/var/log/samba4/glusterfs-{data["cluster_volume"]}.log'
```
yocalebo, line 351: `KeyError` because of minor typo here:
```suggestion
            conf["glusterfs: logfile"] = f'/var/log/samba4/glusterfs-{data["cluster_volname"]}.log'
```

william-gr:
There is a newer PR

PR #6081, opened 2020-11-27
`cluster_volname` is not required on update

william-gr:


themylogin (approved):

william-gr:
@themylogin i think checking data dict was not the wring thing to do, re-asking for review, please analyze surrounding code to confirm too

themylogin (approved):

PR #6082, opened 2020-11-27
Fix shell for user test

william-gr:


themylogin (approved):
Just curious: why? Aren't /bin and /usr/bin the same?

william-gr:
Not sure, caleb changed recently to use /usr/bin as these are the newer paths to be used...

We had conflicting options for bash, /bin/bash and /usr/bin/bash

PR #6083, opened 2020-11-29
NAS-108351 / 20.12 / Allow different zstd algorithms to be set for datasets

sonicaj:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1942,6 +1948,13 @@ class Config:
         namespace = 'pool.dataset'
         event_send = False
 
+    @accepts()
+    async def compression_algorithm_choices(self):
```
william-gr, line 1952: Should probably be just `compression_choices` to map to the attribute name like we do with all other choices

william-gr (approved):

sonicaj:
backport

PR #6084, opened 2020-11-29
NAS-108440 / 20.12 / Fix race condition initializing catalogs on configuring pool for k8s

sonicaj:
This PR fixes a race condition where we might initialize catalogs before we create a dataset for them resulting in dataset creation failing as we already have a directory created.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
@@ -2,6 +2,7 @@ import asyncio
 import errno
 import json
 import os
+import shutil
 
 from middlewared.service import CallError, private, Service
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
```diff
@@ -129,11 +130,17 @@ async def status_change_internal(self):
...
+        # Now we would like to setup catalogs
+        await self.middleware.call('catalog.sync_all')
 
     @private
     async def create_update_k8s_datasets(self, k8s_ds):
         for dataset in await self.kubernetes_datasets(k8s_ds):
             if not await self.middleware.call('pool.dataset.query', [['id', '=', dataset]]):
+                if os.path.exists(os.path.join('/mnt', dataset)):
+                    await self.middleware.run_in_thread(
+                        lambda: shutil.rmtree(os.path.join('/mnt', dataset), ignore_errors=True)
```
william-gr, line 142: Why do we have to use rmtree?

Its a very dangerous function we should avoid if possible

william-gr (approved):

PR #6085, opened 2020-11-30
NAS-102413 / 21.02 / CLI

themylogin:


src/middlewared/middlewared/plugins/network.py
```diff
@@ -2446,6 +2453,7 @@ class NetworkGeneralService(Service):
 
     class Config:
         namespace = 'network.general'
+        cli_namespace = 'network'
```
yocalebo, line 2456: Why not `cli_namespace = 'network.general'` ?

src/middlewared/middlewared/service.py
```diff
@@ -979,6 +1044,14 @@ def get_methods(self, service=None):
                         i in method._job['pipes'] for i in ('input', 'output')
                     ),
                 }
+
+            if is_service_class(svc, CRUDService):
+                # FIXME: Fina a better solution
```
yocalebo, line 1049: nit pick but typo `Fina` -> `Find`

src/middlewared/middlewared/service.py
```diff
@@ -1002,6 +1075,7 @@ def get_events(self):
     async def call_hook(self, name, args, kwargs=None):
         kwargs = kwargs or {}
         await self.middleware.call_hook(name, *args, **kwargs)
+        await self.middleware.call_hook(name, *args, **kwargs)
```
yocalebo, line 1078: duplicate `await self.middleware.call_hook`? or am I missing something?

sonicaj (commented):
@themylogin it looks good to me generally, however i am curious why we are setting a default for `List` automatically ? We removed that explicitly in https://github.com/freenas/freenas/pull/1448

src/middlewared/middlewared/schema.py
```diff
@@ -116,6 +117,24 @@ def to_json_schema(self, parent=None):
...
+        if self.title:
+            schema['title'] = self.title
+
+        if self.description:
+            schema['description'] = self.description
+
+        if self.has_default:
+            schema['default'] = self.default
+
+        schema['_required_'] = self.required
```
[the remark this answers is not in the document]
themylogin: I think we don't, I don't see any reason to hide `_required_` attribute if we are a `Dict` key (most of our fields)

src/middlewared/middlewared/schema.py
```diff
@@ -530,32 +523,12 @@ def validate(self, value):
...
-            if self.has_default:
-                schema['default'] = self.default
-            schema['_required_'] = self.required
+        schema = self._to_json_schema_common(parent)
         if self.null:
             schema['type'] = ['array', 'null']
         else:
             schema['type'] = 'array'
-        if self.enum is not None:
-            schema['enum'] = self.enum
```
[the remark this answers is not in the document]
themylogin: I don't see how a `List` can have an `enum`

src/middlewared/middlewared/schema.py
```diff
@@ -530,32 +523,12 @@ def validate(self, value):
...
-            if isinstance(child_schema['type'], list):
-                for t in child_schema['type']:
-                    items.append({'type': t})
-                    break
-            else:
-                items.append({'type': child_schema['type']})
-        if not items:
-            items.append({'type': 'null'})
-        schema['items'] = items
+        schema['items'] = [i.to_json_schema(self) for i in self.items]
```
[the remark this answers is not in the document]
themylogin: Why do we need it? If `items` is empty then it means we accept any value for the list item. Adding `{'type': `null'}` there will mean we only accept nulls?

src/middlewared/middlewared/alert/service/telegram.py
```diff
@@ -11,7 +11,7 @@ class TelegramAlertService(ThreadedAlertService):
     schema = Dict(
         "telegram_attributes",
         Str("bot_token", required=True, empty=False),
-        List("chat_ids", required=True, empty=False, items=[Int("chat_id")]),
+        List("chat_ids", empty=False, items=[Int("chat_id")]),
```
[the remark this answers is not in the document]
themylogin: All out lists are required (except for update), I've just removed this tautology

src/freenas/usr/local/lib/middlewared_truenas/plugins/truenas.py
```diff
@@ -44,6 +44,9 @@ class TruenasCustomerInformationModel(sa.Model):
 
 class TrueNASService(Service):
 
+    class Config:
+        cli_namespace = "system.truenas"
```
william-gr, line 48: I dont see this service being useful in CLI at all

william-gr (commented):
I am on the fence about making middlewared fully aware of CLI by having namespace listed here.
Wouldnt it be more contained to have it in CLI itself? Allow it to be overridden/changed in that side?

Curious what the team thinks too

themylogin:
@sonicaj @yocalebo ^^^

@william-gr I decided to do it because I won't be happy changing CLI every time we add a new plugin. That is the only change, no more upcoming. Middleware already affects CLI much (schemas, methods documentation and fields descriptions), just feels more natural to also have it here.

src/freenas/usr/local/lib/middlewared_truenas/plugins/truenas.py
themylogin: Won't `accept_eula`, `update_customer_information` and `set_production` methods be useful for the support team?

yocalebo:
> @sonicaj @yocalebo ^^^
> 
> @william-gr I decided to do it because I won't be happy changing CLI every time we add a new plugin. That is the only change, no more upcoming. Middleware already affects CLI much (schemas, methods documentation and fields descriptions), just feels more natural to also have it here.

I agree with @themylogin. Changing CLI every time a new plugin is added would be troublesome.

william-gr:
We dont need to change CLI everytime a new plugin is added. We have sane defaults for 90% of the services, there are just a few that may need to be overridden.

sonicaj:
I think I agree with @william-gr. It would be more contained that way. Also @themylogin were you able to have a peek at not having a default for list ? ( please do see the related PR above for it too )
Btw is `cli_private` only valid for jail plugin and i am curious if we can move that too CLI as well ?

src/freenas/usr/local/lib/middlewared_truenas/plugins/truenas.py
william-gr: Not really, maybe the last two only.

There is a number that would not make much sense as a CLI interface, e.g. get_chassis_hardware, get_eula, etc

william-gr:
Nevermind, I see that actually most of the services require us to put t he service in under another namespace.

themylogin:
> We have sane defaults for 90% of the services, there are just a few that may need to be overridden.

No, `cli_namespace` is different from `namespace` for 90% of the services

william-gr:
Right, thats what I just said :P

themylogin:
@william-gr so what will be the conclusion, we leave `cli_namespace` in the middleware?

william-gr:
yes

themylogin:
@sonicaj so I've figured this List default issue. We have `List(default=[]` everywhere, I just think it's excessive. We can set `default=NOT_PROVIDED` for the cases related to https://github.com/freenas/freenas/pull/1448 but I have found none in today's code

william-gr (approved):

yocalebo (approved):

PR #6087, opened 2020-11-30
NAS-108333 / 20.12 / implement ctdb plugin

yocalebo:
Implement ctdb clustering plugin. I've tried to keep this as modular as possible since ctdb can be used with NFS (it's unknown if this is feasible right now but this is needed for SMB nonetheless).

- override ctdb service and make it start after gluster service
- create 3 sections for ctdb config generation (`ctdb.conf`, `/etc/ctdb/nodes` (private ips), `/etc/ctdb/public_addresses` (public ips))
- create 2 new CRUD APIs for ctdb config (for adding/deleting private/public IPs respectively)
- create 1 new public API for querying status information from the ctdb daemon
- create ctdb service

### DIFF
--- a/src/middlewared/middlewared/etc_files/ctdb_private.conf.mako
+++ b/src/middlewared/middlewared/etc_files/ctdb_private.conf.mako
@@ -1,0 +1,15 @@
+<%
+    from middlewared.plugins.smb import SMBHAMODE
+
+    clustered = SMBHAMODE[middleware.call_sync('smb.get_smb_ha_mode')] == SMBHAMODE.CLUSTERED
+    if not clustered:
+        return
+
+    data = middleware.call_sync('datastore.query', 'ctdb.private.ips')
+    if not data:
+        raise FileShouldNotExist()
+
+%>\
+% for i in data:
+${i['ip']}
+% endfor
--- a/src/middlewared/middlewared/etc_files/ctdb_private.conf.mako
+++ b/src/middlewared/middlewared/etc_files/ctdb_private.conf.mako
@@ -5,7 +5,7 @@
     if not clustered:
         return
 
-    data = middleware.call_sync('datastore.query', 'ctdb.private.ips')
+    data = middleware.call_sync('ctdb.private.ips.query')
     if not data:
         raise FileShouldNotExist()
 
--- a/src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
+++ b/src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
@@ -1,0 +1,138 @@
+from middlewared.schema import List, Dict, IPAddr, Int
+from middlewared.service import accepts, private, job, CRUDService, ValidationErrors
+import middlewared.sqlalchemy as sa
+from .utils import CTDBConfig, JOB_LOCK
+
+
+PRIVATE_IP_FILE = CTDBConfig.PRIVATE_IP_FILE.value
+
+
+class CtdbPrivateIpModel(sa.Model):
+    __tablename__ = 'ctdb_private_ips'
+
+    id = sa.Column(sa.Integer(), primary_key=True)
+    ip = sa.Column(sa.String(45), nullable=True)
+
+
+class CtdbPrivateIpService(CRUDService):
+
+    class Config:
+        namespace = datastore = 'ctdb.private.ips'
+
+    @private
+    async def common_validation(self, data, schema_name, verrors):
+
+        # make sure something was given to us
+        if not data['private_ips']:
+            verrors.add(
+                f'{schema_name}.no_ips',
+                'Private IP(s) must be specified.',
+            )
+
+        verrors.check()
+
+        # normalize the data and remove netmask info (if its given to us)
+        # from `ip` keys since it's not used
+        for i in data['private_ips']:
+            i['ip'] = i['ip'].split('/')[0]
+
+        if schema_name == 'node_create':
+            # get the current ips in the cluster
+            cur_data = {'private_ips': [], 'public_ips': []}
+            cur_data['private_ips'] = await self.query()
+            cur_data['public_ips'] = await self.middleware.call('datastore.query', 'ctdb.public.ips')
+
+            # need to make sure the new private ip(s) don't already exist in the cluster
+            if m := list(set(i['ip'] for i in data['private_ips']).intersection(set(i['ip'] for i in cur_data['private_ips']))):
+                verrors.add(
+                    f'{schema_name}.private_ip_exists',
+                    f'Private IP address(es): {m} already in the cluster.',
+                )
+
+            # need to make sure the new private ip(s) don't already exist in the cluster as public ip(s)
+            if m := list(set(i['ip'] for i in data['private_ips']).intersection(set(i['ip'] for i in cur_data['public_ips']))):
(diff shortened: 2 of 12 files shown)


src/middlewared/middlewared/etc_files/ctdb_private.conf.mako
```diff
@@ -0,0 +1,15 @@
+<%
+    from middlewared.plugins.smb import SMBHAMODE
+
+    clustered = SMBHAMODE[middleware.call_sync('smb.get_smb_ha_mode')] == SMBHAMODE.CLUSTERED
+    if not clustered:
+        return
+
+    data = middleware.call_sync('datastore.query', 'ctdb.private.ips')
```
william-gr, line 8: Why not `ctdb.private.ips.query` ?

Seems more future proof than directly querying database

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,138 @@
...
+
+    @private
+    async def common_validation(self, data, schema_name, verrors):
+
+        # make sure something was given to us
+        if not data['private_ips']:
+            verrors.add(
+                f'{schema_name}.no_ips',
+                'Private IP(s) must be specified.',
+            )
```
william-gr, line 30: Doesn't empty=False in the schema accomplishes the same?

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,138 @@
...
+
+        verrors.check()
+
+        # normalize the data and remove netmask info (if its given to us)
+        # from `ip` keys since it's not used
+        for i in data['private_ips']:
+            i['ip'] = i['ip'].split('/')[0]
+
+        if schema_name == 'node_create':
```
william-gr, line 39: If we have specific create validation, seems it should not be part of the common validation method?

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,138 @@
...
+            # get the current ips in the cluster
+            cur_data = {'private_ips': [], 'public_ips': []}
+            cur_data['private_ips'] = await self.query()
+            cur_data['public_ips'] = await self.middleware.call('datastore.query', 'ctdb.public.ips')
+
+            # need to make sure the new private ip(s) don't already exist in the cluster
+            if m := list(set(i['ip'] for i in data['private_ips']).intersection(set(i['ip'] for i in cur_data['private_ips']))):
+                verrors.add(
+                    f'{schema_name}.private_ip_exists',
```
william-gr, line 48: `private_ip_exists` bit should be the path to an attribute, not some arbitrary error name

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,138 @@
...
+        return data
+
+    @private
+    async def write_private_ips_to_ctdb(self):
+
+        data = {'private_ips': []}
+        data['private_ips'] = await self.query()
+
+        with open(PRIVATE_IP_FILE, 'w') as f:
+            for i in map(lambda i: i['ip'], data['private_ips']):
```
william-gr, line 70: We are writing to a file, which could be blocking I suppose? Seems we should turn this into a method, not a coroutine.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,138 @@
...
+        verrors = ValidationErrors()
+
+        data = await self.common_validation(data, schema_name, verrors)
+
+        for i in data['private_ips']:
+            await self.middleware.call(
+                'datastore.insert', 'ctdb.private.ips', i,
+            )
+
+        return await self.write_private_ips_to_ctdb()
```
william-gr, line 103: This is a weird CRUD service, since we create a batch of entries? And update a batch of entries.

I don't think it should be a CRUD service at all, or have it be a CRUD method and add extra methods for batch operation...

william-gr (changes requested):

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,104 @@
...
+                f.write(f'{i}\n')
+
+        return data
+
+    @accepts(Dict(
+        'node_create',
+        List('private_ips', unique=True, empty=False, items=[Dict('private_ip', IPAddr('ip'))]),
+    ))
+    @job(lock=JOB_LOCK)
+    def do_create(self, job, data):
```
william-gr, line 69: I am still confused by the CRUD Service.

Usually a CRUD Service is when we can create one entity, update it, delete it.

But ion here are are performing a batch create, which seems to defeat the CRUD paradigm since I cant operate over a single entity.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,89 @@
...
+        # from `ip` key since it's not used
+        data['ip'] = data['ip'].split('/')[0]
+
+        self.common_validation(data, schema_name, verrors)
+
+        self.middleware.call_sync('datastore.insert', 'ctdb.private.ips', data)
+
+        self.write_private_ips_to_ctdb()
+
+        return data
```
william-gr, line 75: We usually return get_instance(id), so its the same output as .query

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,89 @@
...
+    @private
+    def write_private_ips_to_ctdb(self):
+
+        with open(PRIVATE_IP_FILE, 'w') as f:
+            for i in map(lambda i: i['ip'], self.middleware.call_sync('ctdb.private.ips.query')):
+                f.write(f'{i}\n')
+
+    @accepts(Dict(
+        'node_create',
+        IPAddr('ip'),
```
william-gr, line 48: Seems this should be required=True since thats the only param

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+PRIVATE_IP_FILE = CTDBConfig.PRIVATE_IP_FILE.value
+
+
+class CtdbPrivateIpModel(sa.Model):
+    __tablename__ = 'ctdb_private_ips'
+
+    id = sa.Column(sa.Integer(), primary_key=True)
+    ip = sa.Column(sa.String(45), nullable=True)
```
themylogin, line 14: Why is this field nullable?

themylogin (changes requested):

yocalebo:
The `/etc/ctdb/nodes` and `/etc/ctdb/public_addresses` files have to be the same across all nodes in the cluster. Putting them in the local db isn't really conducive to that paradigm. I'm going to close this and reopen fresh PR using flat files and symlinks since the man pages even suggest this can be done.

PR #6095, opened 2020-12-03
NAS-108511 / 20.12 / Update Text Strings

q5sys:
Spoke with Tim and we decided it's best to make this message more generic in case there are other BMC issues in the future, that we may want to flag for.


william-gr (changes requested):
+src/middlewared/middlewared/alert/source/freenas_bmc.py:16:93: E999 SyntaxError: invalid syntax

william-gr (approved):

PR #6107, opened 2020-12-04
NAS-108334 / 21.02 / auto-create and mount ctdb shared volume

yocalebo:
If gluster is configured/started/enabled etc, it's imperative that we create the `ctdb_shared_vol` clustered filesystem and then mount it locally, on all peers in the TSP. This will be used to host various cluster wide files to provide active-active sharing services.


src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        return payload
+
+    async def shared_volume_exists_and_started(self):
+
+        exists = started = False
+        v = await self.middleware.call('gluster.volume.status', self.CTDB_VOL_NAME)
+        await v.wait()
+        if v.error:
```
themylogin, line 46: You can use `result = v.wait(raise_error=True)`

src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
```diff
@@ -0,0 +1,232 @@
...
+        # try to mount it locally
+        mount = await self.middleware.call('ctdb.shared.volume.mount')
+        await mount.wait()
+        if mount.error:
+            raise CallError(f'{mount.error}')
+
+        return 'SUCCESS'
+
+    @accepts()
+    @job(lock='ctdb_delete_lock')
```
themylogin, line 129: Shouldn't it be the same lock everywhere?

src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
```diff
@@ -0,0 +1,232 @@
...
+
+        # try to mount it
+        if not path.is_mount():
+            try:
+                command = [
+                    'mount', '-t', 'glusterfs',
+                    'localhost:/' + path.name, path.as_posix(),
+                ]
+                cp = await run(command)
+                if cp.returncode:
```
themylogin, line 206: `check=True` is the default, we won't reach here

src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
```diff
@@ -0,0 +1,232 @@
...
+                    raise CallError(f'{cp.stderr.decode().strip()}')
+            except subprocess.CalledProcessError as e:
+                raise CallError(f'{e.stderr.decode().strip()}')
```
themylogin, line 209: It will be re-raised as `CallError` automatically in `adapt_exception`, no need to handle it every time

src/middlewared/middlewared/plugins/gluster_linux/eventsd.py
```diff
@@ -147,3 +167,67 @@ def sync(self, job):
...
+            try:
+                # make sure glusterd_dir/events subdir exists
+                webhook_file.parent.mkdir(exist_ok=True)
+
+                # now create the webhook file
+                if not webhook_file.exists():
+                    webhook_file.touch()
+                    new_file = True
+            except Exception as e:
+                raise CallError('%r', e)
```
themylogin, line 195: You've probably meant something else here?

src/middlewared/middlewared/plugins/gluster_linux/peer.py
```diff
@@ -38,6 +44,32 @@ def __peer_wrapper(self, method, host=None):
...
+    def _parse_peer(self, p):
+
+        data = {
+            'uuid': p.find('uuid').text,
+            'hostname': p.find('hostname').text,
+            'connected': p.find('connected').text,
+        }
+
+        if data['connected'] == '1':
+            data['connected'] = 'Connected'
```
themylogin, line 56: Why do you use string typing for a boolean value?

themylogin (changes requested):

src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
yocalebo: The create/delete methods need to share the same lock but we need to use a separate lock for mount/umount methods since those are called from the create/delete methods respectively.

src/middlewared/middlewared/plugins/gluster_linux/peer.py
yocalebo: The subprocessed command that gets run returns XML data so it gets returned as a string.

src/middlewared/middlewared/plugins/gluster_linux/eventsd.py
```diff
@@ -147,3 +167,71 @@ def sync(self, job):
             return json.loads(out.strip())['output']
         else:
             raise CallError(err.strip())
+
+    @job(lock=EVENTSD_LOCK)
```
william-gr, line 171: Isnt this supposed to be a private method?

src/middlewared/middlewared/plugins/gluster_linux/eventsd.py
```diff
@@ -147,3 +167,71 @@ def sync(self, job):
...
+                # make sure glusterd_dir/events subdir exists
+                webhook_file.parent.mkdir(exist_ok=True)
+
+                # now create the webhook file
+                if not webhook_file.exists():
+                    webhook_file.touch()
+                    new_file = True
+            except Exception as e:
+                raise CallError(
+                    'Failed creating %s with error: %r', str(webhook_file), e
```
william-gr, line 196: CallError doesnt accept args like that, we need to format ourselves

src/middlewared/middlewared/plugins/gluster_linux/eventsd.py
```diff
@@ -147,3 +167,71 @@ def sync(self, job):
...
+
+        init_data = {}
+        if not new_file:
+            # need to know if webhooks have already been added
+            # to the file so act accordingly
+            with webhook_file.open('r') as f:
+                init_data = json.load(f)
+
+        # dont add local URL if it's already there
+        if not init_data.get(f'{LOCAL_WEBHOOK_URL}'):
```
william-gr, line 211: f-string not necessary here, just LOCAL_WEBHOOK_URL

src/middlewared/middlewared/plugins/gluster_linux/eventsd.py
```diff
@@ -147,3 +167,71 @@ def sync(self, job):
...
+            # make sure to add local api endpoint to file
+            local_url = {f'{LOCAL_WEBHOOK_URL}': {'token': '', 'secret': ''}}
```
william-gr, line 213: f-string not necessary here, just LOCAL_WEBHOOK_URL

src/middlewared/middlewared/plugins/gluster_linux/eventsd.py
```diff
@@ -147,3 +167,71 @@ def sync(self, job):
...
+            init_data.update(local_url)
+
+        # finally write it out
+        try:
+            with webhook_file.open('w') as f:
+                f.write(json.dumps(init_data))
+        except Exception as e:
+            raise CallError(
+                'Failed writing to %s with error: %r', str(webhook_file), e
```
william-gr, line 222: CallError doesnt accept args like that, we need to format ourselves

src/middlewared/middlewared/plugins/gluster_linux/peer.py
```diff
@@ -38,6 +44,32 @@ def __peer_wrapper(self, method, host=None):
...
+        return data
+
+    def _parse_peer_status_xml(self, data):
+
+        peers = []
+        for _ in data.findall('peerStatus/peer'):
+            try:
+                peers.append(self._parse_peer(_))
+            except Exception as e:
+                raise CallError(f'{e}')
```
william-gr, line 69: raise CallError(f'Failed to whatever: {e}')

src/middlewared/middlewared/plugins/gluster_linux/peer.py
```diff
@@ -125,4 +157,60 @@ def pool(self, job):
         including localhost.
         """
 
-        return self.__peer_wrapper(peer.pool)
+        # get the local viewpoint of the remote peers in the TSP
+        local_view = self.__peer_wrapper(peer.status)
+        if not local_view:
+            return local_view
```
william-gr, line 163: Good use of assignemnt expression

if not local_view := self.__peer_wrapper(peer.status):
   return

:P

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -344,9 +350,35 @@ async def __mount(self, pool, uuid, path=SYSDATASET_PATH):
...
         for dataset, name in reversed(self.__get_datasets(pool, uuid)):
             try:
+                # if this is the ctdb dataset, then we have to
+                # unmount the local glusterfuse mount first before
+                # unmounting the underlying zfs dataset
+                if osc.IS_LINUX and name == CTDBConfig.CTDB_VOL_NAME.value:
+                    c = await self.middleware.call('ctdb.shared.volume.umount')
+                    await c.wait()
+                    if c.error:
+                        raise CallError(f'{c.error}')
```
william-gr, line 381: Just error may give no context, we may say "Failed to umount ctdb shared volume: {e}"

src/middlewared/middlewared/webhooks/cluster_events.py
```diff
@@ -0,0 +1,81 @@
+from middlewared.plugins.cluster_linux.utils import CTDBConfig
+
+from aiohttp import web
+
+
+class ClusterEventsApplication(object):
+
+    def __init__(self, middleware):
+        self.middleware = middleware
+        self.ctdb_shared_vol = CTDBConfig.CTDB_VOL_NAME.value
```
william-gr, line 10: When overriding a method like this dont forget to call super()

```
        super().__init__(*args, **kwargs)
        self._ctdb_shared_vol = CTDBConfig.CTDB_VOL_NAME.value
```

src/middlewared/middlewared/webhooks/cluster_events.py
```diff
@@ -0,0 +1,81 @@
...
+                    umount_it = True
+
+            if event == 'AFR_SUBVOLS_DOWN':
+                if msg.get('subvol', '') in self.ctdb_shared_vol:
+                    umount_it = True
+
+        if mount_it:
+            mount = await(
+                await self.middleware.call('ctdb.shared.volume.mount')
+            ).wait()
```
william-gr, line 38: Is it OK to block process_event like we are waiting it to mount/umount?

src/middlewared/middlewared/plugins/gluster_linux/eventsd.py
yocalebo: I've made it private

src/middlewared/middlewared/plugins/sysdataset.py
yocalebo: I've actually changed it so that it's logging an error and continuing on so we don't block other mount operations in the sysdataset.setup method.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
```diff
@@ -0,0 +1,215 @@
...
+                                                     CRE_OR_DEL_LOCK)
+
+import os
+import pathlib
+
+
+class CtdbSharedVolumeService(Service):
+
+    class Config:
+        namespace = 'ctdb.shared.volume'
```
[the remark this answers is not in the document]
yocalebo: No, I wrote it for public consumption and I also expect that TC will need to use it at some point.

src/middlewared/middlewared/webhooks/cluster_events.py
yocalebo: Shouldn't matter, but the `volume.{mount/umount}` methods share the same lock so we should be protected anyways.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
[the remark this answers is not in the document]
william-gr: In that case seems like the both methods below need to be `@private`?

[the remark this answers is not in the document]
yocalebo: I thought that these weren't available even if the `@private` decorator wasn't given?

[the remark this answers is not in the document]
yocalebo: I've added the decorator.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
```diff
@@ -0,0 +1,213 @@
...
+            # stop the volume
+            force = {'force': True}
+            await(
+                await self.middleware.call('gluster.volume.stop', self.CTDB_VOL_NAME, force)
+            ).wait(raise_error=True)
+
+        # now delete it
+        data = {'name': self.CTDB_VOL_NAME}
+        await(
+            await self.middleware.call('gluster.volume.delete', data)
```
william-gr, line 149: Is gluster.volume a CRUD or a simple Service?

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -344,9 +350,50 @@ async def __mount(self, pool, uuid, path=SYSDATASET_PATH):
                 os.mkdir(mountpoint)
             await run('mount', '-t', 'zfs', dataset, mountpoint, check=True)
 
+        if osc.IS_LINUX:
+
+            # make sure the glustereventsd webhook dir and
+            # config file exist and start the glustereventsd
+            # service (if appropriate)
```
william-gr, line 357: Just realized something... We will *always* start gluster and gluster eventsd?
Even when gluster has never been set up?

src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
yocalebo: It's a CRUD service.

src/middlewared/middlewared/plugins/sysdataset.py
yocalebo: I've changed the `eventsd.init` method to only setup the `/var/db/system/glusterd/events/webhooks.json` path if `glusterd` has been enabled.

Not sure how you want to handle `glusterd` and `ctdb_shared_vol` datasets that are setup in the sysdataset plugin always. If we don't set those up and user enables `glusterd` after system is booted, I'm assuming I'd have to run `sysdataset.setup` again?

src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
william-gr: Then I am confused. Following CRUD methods, delete method should receive an id/name as the first param, here we are providing a dict

src/middlewared/middlewared/plugins/sysdataset.py
william-gr: Creating dataset is fine, but it seems like we are doing more than that, including mounting the gluster filesystem?

We wouldnt expect to have any of that infrastructure related to gluster up and running if there is no gluster configured/enabled, especially because 90%+ of TrueNAS users won't use it

I may be misunderstanding what all these things do

src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
yocalebo: Because we're not storing cluster information in the db. I can go back and change the volume and peer plugins to SimpleService

william-gr: If these are not regular usual CRUD Services then it should not be a CRUD Service to begin with. I agree.

src/middlewared/middlewared/plugins/sysdataset.py
yocalebo: You're correct in that it was also creating a directory in the `glusterd` system dataset as well as a file and then writing certain data to that file and then starting the `glustereventsd` service. However, with my recent commit that doesn't occur unless `glusterd` is enabled in db like all other services.

william-gr: Thank you for clarifying.

Shouldn't we make a way to not even try to mount in this case then?

e.g. .init returns something that allows me to put whole block within an `if`?

src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
```diff
@@ -0,0 +1,213 @@
...
+                'The ctdb shared volume does not exist or '
+                'is not started.'
+            )
+
+        path = pathlib.Path(CTDBConfig.CTDB_LOCAL_MOUNT.value)
+        try:
+            # make sure the dirs are there
+            path.mkdir(parents=True, exist_ok=True)
+        except Exception as e:
+            raise CallError(f'{e}')
```
william-gr, line 185: Just noticed this lonely CallError without any description before the error message

src/middlewared/middlewared/plugins/sysdataset.py
yocalebo: Good point

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -344,9 +350,47 @@ async def __mount(self, pool, uuid, path=SYSDATASET_PATH):
...
+                    'Failed to initilize %s directory with error: %s',
+                    CTDBConfig.CTDB_VOL_name.value,
+                    init.error
+                )
+            elif init.result:
+                # mount the local glusterfuse mount after
+                # successfully initializing glustereventsd
+                mnt_job = await self.middleware.call('ctdb.shared.volume.mount')
+                mount = await mnt_job.wait()
+                if mount.error:
```
william-gr, line 371: mnt_job.error

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -344,9 +350,47 @@ async def __mount(self, pool, uuid, path=SYSDATASET_PATH):
...
             await run('mount', '-t', 'zfs', dataset, mountpoint, check=True)
 
+        if osc.IS_LINUX:
+
+            # make sure the glustereventsd webhook dir and
+            # config file exist and start the glustereventsd
+            # service (if appropriate)
+            init_job = await self.middleware.call('gluster.eventsd.init')
+            init = await init_job.wait()
+            if init.error:
```
william-gr, line 360: init_job.error

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -344,9 +350,47 @@ async def __mount(self, pool, uuid, path=SYSDATASET_PATH):
...
+            elif init:
+                # mount the local glusterfuse mount after
+                # successfully initializing glustereventsd
+                mnt_job = await self.middleware.call('ctdb.shared.volume.mount')
+                mount = await mnt_job.wait()
+                if mnt_job.error:
+                    self.logger.error(
+                        'Failed to mount locally %s with error: %s ',
+                        CTDBConfig.CTDB_VOL_NAME.value,
+                        mount.error
```
william-gr, line 375: mnt_job.error

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -344,9 +350,47 @@ async def __mount(self, pool, uuid, path=SYSDATASET_PATH):
...
+
         for dataset, name in reversed(self.__get_datasets(pool, uuid)):
             try:
+                # if this is the ctdb dataset, then we have to
+                # unmount the local glusterfuse mount first before
+                # unmounting the underlying zfs dataset
+                if osc.IS_LINUX and name == CTDBConfig.CTDB_VOL_NAME.value:
+                    umnt_job = await self.middleware.call('ctdb.shared.volume.umount')
+                    umount = await umnt_job.wait()
+                    if umount.error:
```
william-gr, line 388: umnt_job.error

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -344,9 +350,47 @@ async def __mount(self, pool, uuid, path=SYSDATASET_PATH):
...
+                        self.logger.error(
+                            'Failed to umount %s with error: %s',
+                            CTDBConfig.CTDB_VOL_NAME.value,
+                            umount.error
```
william-gr, line 392: umnt_job.error

yocalebo: sigh

src/middlewared/middlewared/plugins/sysdataset.py
yocalebo: sigh

william-gr (approved):
Lets wait on @themylogin review to see what he thinks as well

william-gr:
Oh also two new flake errors :P

yocalebo:
@themylogin ping

themylogin (approved):

PR #6108, opened 2020-12-05
NAS-107395 / 12.0 / Implement smart.test.disk_choices to remove NVMe disks

themylogin:


src/middlewared/middlewared/plugins/smart.py
```diff
@@ -183,6 +183,16 @@ async def validate_data(self, data, schema):
 
         return verrors
 
+    @accepts(Bool('full_disk', default=False))
+    async def disk_choices(self, full_disk):
+        """
+        Returns disk choices for S.M.A.R.T. test.
```
william-gr, line 189: A note about what full_disk is/does could be worthwhile.

Also, please remind me why NVMe are not a possible choice?

themylogin: `smartmontools` does not support NVMe drives self-tests. Neither starts tests, nor reports tests results.

william-gr (approved):
Thanks

PR #6109, opened 2020-12-05
NAS-108521 / 12.0 / Prevent race condition when generating same config at the same time

themylogin:
(cherry-picked from commit f0e88688808ebfefad7428e24db2778c8754263c)


william-gr (approved):

william-gr:
I imagine we verified the possibility of deadlock here?

themylogin:
@william-gr locks are per-group, I the group is unlikely to call `etc.generate` on itself. I've double checked, currently, the only `etc.generate` in `etc` plugin is `rc` generating `kerberos`. Additionally, these changes lived in SCALE for 5 months, and no deadlocks were notices.

PR #6120, opened 2020-12-08
NAS-108441 / 12.0 /  Do not include system dataset into quota alerts

themylogin:
I am not sure we should have a new alert in 12.0. What if an average system has a core file there? (I had one for syslog-ng for example) We'll have a ton of tickets!

### DIFF
--- a/src/middlewared/middlewared/alert/source/cores.py
+++ b/src/middlewared/middlewared/alert/source/cores.py
@@ -1,22 +1,45 @@
 import os
+import time
 
 from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, Alert, ThreadedAlertSource
+from middlewared.utils.osc import IS_FREEBSD
 
 
 class CoreFilesArePresentAlertClass(AlertClass):
     category = AlertCategory.SYSTEM
     level = AlertLevel.WARNING
     title = "Core Files Found in System Database"
-    text = ("System core files were found in /var/db/system/cores: %s. Please create a ticket at https://jira.ixsystems.com/ "
-            "and remove these files.")
+    text = ("System core files were found in /var/db/system/cores. Please create a ticket at "
+            "https://jira.ixsystems.com/ and remove these files.")
 
 
 class CoreFilesArePresentAlertSource(ThreadedAlertSource):
     def check_sync(self):
+        cores = "/var/db/system/cores"
         try:
-            files = sorted(os.listdir("/var/db/system/cores"))
+            listdir = os.listdir(cores)
         except Exception:
-            files = []
+            return
 
-        if files:
-            return Alert(CoreFilesArePresentAlertClass, ", ".join(files))
+        has_core_file = False
+        for file in listdir:
+            if not file.endswith(".core"):
+                continue
+
+            path = os.path.join(cores, file)
+            if not os.path.isfile(path):
+                continue
+
+            unlink = False
+            if IS_FREEBSD and file == "syslog-ng.core":
+                unlink = True
+            elif os.stat(path).st_mtime < time.time() - 86400 * 5:
+                unlink = True
+            else:
+                has_core_file = True
+
+            if unlink:
+                os.unlink(path)
+
+        if has_core_file:
+            return Alert(CoreFilesArePresentAlertClass)
--- a/src/middlewared/middlewared/alert/source/cores.py
+++ b/src/middlewared/middlewared/alert/source/cores.py
@@ -6,9 +6,9 @@ from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, Alert,
 class CoreFilesArePresentAlertClass(AlertClass):
     category = AlertCategory.SYSTEM
     level = AlertLevel.WARNING
-    title = "Core Files Are Present"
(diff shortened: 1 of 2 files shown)


themylogin:
@freenas/docs-team please write proper title and text for new alert, thank you.

src/middlewared/middlewared/alert/source/cores.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, Alert, ThreadedAlertSource
+
+
+class CoreFilesArePresentAlertClass(AlertClass):
+    category = AlertCategory.SYSTEM
+    level = AlertLevel.WARNING
+    title = "Core Files Are Present"
+    text = ("Core files are present in /var/db/system/cores: %s. Please, file a ticket at https://jira.ixsystems.com/ "
```
william-gr, line 10: Not sure if there is any usefulness in listing every file here.

william-gr (commented):
I agree with you.

I wonder if we should compare the creation date of these files and only account for relatively newer ones... e.g. within last 5 days?

Also syslog-ng is known to segfault a LOT for unknown reasons, may make sense to blacklist that one on BSD.

Thoughts?

src/middlewared/middlewared/alert/source/cores.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, Alert, ThreadedAlertSource
+
+
+class CoreFilesArePresentAlertClass(AlertClass):
+    category = AlertCategory.SYSTEM
+    level = AlertLevel.WARNING
+    title = "Core Files Are Present"
```
Mrt134, line 9:
```suggestion
    title = "Core Files Found in System Database"
```

src/middlewared/middlewared/alert/source/cores.py
themylogin: > I wonder if we should compare the creation date of these files and only account for relatively newer ones... e.g. within last 5 days?

I don't think that'll be right, files older than 5 days are still consuming disk space

src/middlewared/middlewared/alert/source/cores.py
```diff
@@ -0,0 +1,22 @@
...
+    text = ("Core files are present in /var/db/system/cores: %s. Please, file a ticket at https://jira.ixsystems.com/ "
+            "and then remove them.")
```
Mrt134, line 11:
```suggestion
    text = ("System core files were found in /var/db/system/cores: %s. Please create a ticket at https://jira.ixsystems.com/ "
            "and remove these files.")
```

Mrt134 (approved):
I put in a few suggestions for title and text

src/middlewared/middlewared/alert/source/cores.py
william-gr: We could delete older files?
Not sure its useful to keep old files. i have seen reports back fro.2018

src/middlewared/middlewared/alert/source/cores.py
```diff
@@ -0,0 +1,40 @@
...
+    def check_sync(self):
+        cores = "/var/db/system/cores"
+        try:
+            listdir = os.listdir(cores)
+        except Exception:
+            return
+
+        has_core_file = False
+        for file in listdir:
+            path = os.path.join(cores, file)
```
william-gr, line 26: Wonder if we should make sure its a file and not a dir, and ending with .core?

william-gr:
@anodos325 @yocalebo curious on yours opinion on this

src/middlewared/middlewared/alert/source/cores.py
```diff
@@ -0,0 +1,45 @@
...
+        except Exception:
+            return
+
+        has_core_file = False
+        for file in listdir:
+            if not file.endswith(".core"):
+                continue
+
+            path = os.path.join(cores, file)
+            if not os.path.isdir(path):
```
william-gr, line 30: you mean .isfile?

src/middlewared/middlewared/alert/source/cores.py
```diff
@@ -0,0 +1,45 @@
...
+                continue
+
+            unlink = False
+            if IS_FREEBSD and file == "syslog-ng.code":
```
william-gr, line 34: .core

yocalebo:
`syslog-ng` should definitely be blacklisted. How is this alert removed? Removing all the core files in that directory and then letting the period alert task run and detect there are none?

themylogin:
@yocalebo yes, and period is 1 minute

yocalebo (approved):
I'm okay with this change, however, I predict we will be seeing a few more tickets because of this 😄

william-gr:
Me too but I couldn't think of anything better.

The cores dataset was getting full of cores and we were getting alerts of zfs quota on that.
The alternative was ignoring cores dataset being full and unable to core dump anymore.

Open to new/other ideas

yocalebo:
This approach is better than ignoring cores dataset and preventing core dumps for sure. I agree with the change as this will probably give us some insight into common applications that are core dumping....for which we should fix obviously.

william-gr (approved):

PR #6122, opened 2020-12-09
NAS-108562 / 20.12 / Remove root from /etc/ftpusers when root login is enabled

themylogin:
@william-gr this is a blacklist for ftp users, by default it contains some of the system built-in users

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/proftpd.py
+++ b/src/middlewared/middlewared/etc_files/local/proftpd.py
@@ -1,7 +1,11 @@
 import os
 
+from middlewared.utils import osc
+
 
 def setup(middleware):
+    ftp = middleware.call_sync("ftp.config")
+
     os.makedirs("/var/log/proftpd", exist_ok=True)
     os.makedirs("/var/run/proftpd", exist_ok=True)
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/etc_files/local/proftpd.py
```diff
@@ -1,18 +1,31 @@
...
     open("/var/run/proftpd/proftpd.delay", "w+").close()
     open("/etc/hosts.allow", "w+").close()
     open("/etc/hosts.deny", "w+").close()
 
+    if osc.IS_LINUX:
+        filters = [["builtin", "=", True]]
+        if ftp["rootlogin"]:
+            filters.append(["username", "!=", "root"])
+
+        ftpusers = [user["username"] for user in middleware.call_sync("user.query", filters, )]
```
william-gr, line 21: `("user.query", filters, )` looks odd

`("user.query", filters, )`

william-gr (commented):
/etc/ftpusers is only for builtin users?

themylogin:
> /etc/ftpusers is only for builtin users?

In addition, it only allows logins with "ftp" group, so this file is a protection against a case where built-in user accidentally gets a password and ftp group membership.

william-gr (approved):

PR #6130, opened 2020-12-10
NAS-108601 / 20.12 / minor tweaks to HA setup method and sync ssh state to standby on HA setup

yocalebo:
- `failover.setup` event was being called but had never been registered
- log appropriate messages in the `hook_setup_ha` method if tracebacks occur
- sync the SSH state to the standby node in `hook_setup_ha` if it's enabled but not started on standby node

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
@@ -1479,19 +1479,40 @@ async def hook_restart_devd(middleware, *args, **kwargs):
 
 async def hook_license_update(middleware, *args, **kwargs):
     FailoverService.HA_MODE = None
-    if await middleware.call('failover.licensed'):
+
+    if not await middleware.call('failover.licensed'):
+        return
+
+    etc_generate = ['rc']
+    if await middleware.call('system.feature_enabled', 'FIBRECHANNEL'):
+        await middleware.call('etc.generate', 'loader')
+        etc_generate += ['loader']
+
+    # setup the local heartbeat interface
+    heartbeat = True
+    try:
         await middleware.call('failover.ensure_remote_client')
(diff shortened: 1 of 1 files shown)


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1479,19 +1479,33 @@ async def hook_restart_devd(middleware, *args, **kwargs):
...
+        await middleware.call('failover.send_small_file', '/data/license')
+    except Exception:
+        middleware.logger.warning('Failed to sync db to standby')
+
+    try:
+        await middleware.call('failover.call_remote', 'failover.ensure_remote_client')
+    except Exception:
+        middleware.logger.warning('Failed to ensure remote client on standby')
+
+    try:
```
william-gr, line 1504: The way they were placed wasn't a mistake. If `failover.ensure_remote_client` there is no point in trying to call failover.call_remote next

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1537,6 +1551,17 @@ async def hook_setup_ha(middleware, *args, **kwargs):
     except Exception:
         ha_configured = False
 
+    # on brand new HA systems, we don't synchronize service states to the
+    # standby controller. Minimally, however, it's nice to synchronize ssh
+    # (if appropriate, of course)
+    filters = [('srv_service', '=', 'ssh')]
+    ssh_enabled = remote_ssh_started = False
```
william-gr, line 1558: Shouldnt this whole block be in the `if not ha_configured:` ?

I dont want to bother to keep trying to start ssh all the other times, only if its really the first time we are setting up HA

william-gr (approved):

PR #6132, opened 2020-12-10
NAS-108611 / 12.0 / Move aio_fbsd to end of list of VFS modules

anodos325:
Certain VFS modules may alter targets of vfs_pread_send() / vfs_pwrite_send() requests depending on the target file type. For instance, in a non-default configuration (for us) configuration, vfs_fruit will intercept stream reads / writes for afp_info and afp_resource streams and convert them to writes to special adouble files. This needs to happen before the async pread / pwrite hits vfs_aio_fbsd, which means vfs_aio_fbsd needs to be at the end of the module list.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb_/registry.py
+++ b/src/middlewared/middlewared/plugins/smb_/registry.py
@@ -156,7 +156,7 @@ class SharingSMBService(Service):
     @private
     async def order_vfs_objects(self, vfs_objects):
         vfs_objects_special = ('catia', 'zfs_space', 'fruit', 'streams_xattr', 'shadow_copy_zfs',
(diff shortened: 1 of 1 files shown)


william-gr (approved):

william-gr:
@anodos325 I am only confused why you're saying move, if you're only adding the vfs module here?

anodos325:
This change is in order_vfs_objects(), which has a prestigious lineage originating in generate_smb4_conf.py. Basically, once we resolve all the "vfs objects" required to provide whatever behavior is chosen through the user-provided combination of parameters and auxiliary parameters, positions of birds, etc, we re-order them into something that should work correctly. Having aio_fbsd at the end of the list ensures that it will always be the last in the list of vfs objects in samba (if it is in the list provided to this function).

PR #6136, opened 2020-12-11
NAS-106689 / 12.0 / Ensure HBA update is to latest approved hba version and not just that…

themylogin:
… it has correct major version

### DIFF
--- a/src/freenas/usr/local/sbin/firmware_update.py
+++ b/src/freenas/usr/local/sbin/firmware_update.py
@@ -10,6 +10,7 @@ import logging
 import logging.config
 import logging.handlers
 import os
+from packaging import version
 import re
 import sys
 import subprocess
@@ -21,9 +22,9 @@ FWPATH = '/usr/local/share/firmware/'
 FAILED_UPDATE_SENTINEL = '/data/.hba_firmware_update_fail'
 UPDATE_SENTINEL = '/data/.hba_firmware_update'
 
-SAS92_VERSION = 20
-SAS93_VERSION = 16
-HBA94_VERSION = 15
+SAS92_VERSION = version.parse("20.0.0.0")
+SAS93_VERSION = version.parse("16.0.0.0")
+HBA94_VERSION = version.parse("15.0.0.0")
 
 UPDATE_SUCCESS = []
 UPDATE_FAIL = []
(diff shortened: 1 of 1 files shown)


src/freenas/usr/local/sbin/firmware_update.py
```diff
@@ -21,9 +22,9 @@
 FAILED_UPDATE_SENTINEL = '/data/.hba_firmware_update_fail'
 UPDATE_SENTINEL = '/data/.hba_firmware_update'
 
-SAS92_VERSION = 20
-SAS93_VERSION = 16
-HBA94_VERSION = 15
+SAS92_VERSION = version.parse("20.0.0.0")
```
themylogin, line 25: @darkfiberiru will you be able to help me put the current firmware versions here?

william-gr:
@darkfiberiru  ping

src/freenas/usr/local/sbin/firmware_update.py
darkfiberiru: Sorry github notifications are  a mess for me and have been sick. Put current firmware how? Like how to get it or how list of them or???

amotin: It will not work like this, since we have different minor versions for different board models.  It has to be encoded in file name probably.  For 92xx and 93xx the value can be easily seen just as a string in the binary.  For 94xx something has changed.

themylogin: Are there multiple firmwares in each file?

darkfiberiru: Each card has a individual firmware that is fixed. With each family of cards being same major version(atleast so far). We have one file per card.... issue is we previously didn't do a good job of document what that version is. According to mav all but 9400 cards you can find string in binary firmware file.

darkfiberiru: https://github.com/freenas/firmware

william-gr (approved):

amotin (approved):
This part looks good to me.  Make sure the package with firmware images is in sync.  Because before this IIRC it was not last time I've checked.

PR #6148, opened 2020-12-13
NAS-108637 / 20.12 / Set 120 character pep8 line limit

sonicaj:
This commit changes max character limit to 120 keeping in line with what we have had before.

### DIFF
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -3,6 +3,7 @@ scanner:
     linter: flake8
 
 flake8:
+    max-line-length: 120
     ignore:
       - E402
       - E501


yocalebo (commented):
This looks good to me, however, do we know why pep3speaks is still complaining about line length when it's clearly in the `ignore:` section?

sonicaj:
Not sure, if we search to see how to disable the 80 char limit, this is what comes up for pep3speaks. So we probably need these 2 to make sure it's respected.

yocalebo (approved):
Just makes me wonder if the other items in the ignore list are correct.

william-gr (approved):

william-gr:
Try using the other checker, instead of flake8

william-gr:
@sonicaj

PR #6150, opened 2020-12-14
NAS-108575 / 21.02 / Make checking container image updates optional

sonicaj:
Dockerhub has recently introduced rate limits and we are affected by the rate limit. How it happens is that each time when we check if we have an update for an image available, we consume the rate limit which can potentially hinder operations involving docker/k8s, so this PR makes checking container image updates optional.

### DIFF
--- a/src/middlewared/middlewared/plugins/docker_linux/update.py
+++ b/src/middlewared/middlewared/plugins/docker_linux/update.py
@@ -1,0 +1,33 @@
+import middlewared.sqlalchemy as sa
+
+from middlewared.schema import Bool, Dict
+from middlewared.service import accepts, ConfigService
+
+
+class ContainerModel(sa.Model):
+    __tablename__ = 'services_container'
+
+    id = sa.Column(sa.Integer(), primary_key=True)
+    enable_image_updates = sa.Column(sa.Boolean(), default=True)
+
+
+class ContainerService(ConfigService):
+
+    class Config:
+        datastore = 'services.kubernetes'
+
+    @accepts(
+        Dict(
+            'container_update',
+            Bool('enable_image_updates'),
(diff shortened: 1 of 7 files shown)


src/middlewared/middlewared/plugins/docker_linux/update.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @accepts(
+        Dict(
+            'container_update',
+            Bool('enable_image_updates'),
+            update=True,
+        )
+    )
+    async def do_update(self, data):
+        old_config = await self.config()
```
yocalebo, line 27: Should we create a docstring here specifying the potential for rate throttling? Otherwise I see end-users just enabling this by default to check for latest updates for the images running their containers.

sonicaj: We definitely should, thank you for pointing it out ;)

yocalebo (approved):

themylogin (approved):

Ornias1993:
@sonicaj Watchtower hit the same issue for the same reason...
you might want to check their issue, because it seems they manage to fix it:
https://github.com/containrrr/watchtower/issues/669

william-gr:
@sonicaj is this ready to merge?

sonicaj:
Thank you for the links @Ornias1993, they have been helpful but in my limited experiments i found that i ended up doing a get request anyways with the same logic they were using with headers and HEAD request and they also document this too. I am not sure about the scale of this with respect to different images but i think what we can at least do is still try to HEAD request like they do and if that fails do a GET one but still allow this to be configurable as the HEAD method is not going to work every time.

themylogin (approved):

PR #6154, opened 2020-12-14
Fix check for AD enabled

anodos325:


william-gr (approved):

william-gr:
@anodos325 this is not required in 12.0, I imagine?

PR #6162, opened 2020-12-15
NAS-108512 / 21.02 / Do not start a VM which has unavailable resources

sonicaj:
This PR introduces changes to ensure that before we start a VM we ensure that the VM in question has resources it requires available for starting/executing.


william-gr (approved):

sonicaj:
backport

PR #6165, opened 2020-12-16
NAS-108624 / 21.02 / Handle race condition when retrieving data from netif

sonicaj:


william-gr (commented):
I didnt understand why change code structure instead of just adding exception handling

william-gr (approved):

PR #6166, opened 2020-12-16
NAS-108678 / 12.0 / Fix possibility for middleware plugin to leak passwords to debug

themylogin:
@sonicaj is it possible that ACME / KMIP jobs might be returning something we want to hide from debug?


sonicaj:
@themylogin certificate creation job is sensitive where we return the complete created certificate instance, we don't return anything sensitive from kmip sync jobs.

william-gr (approved):

PR #6170, opened 2020-12-16
NAS-108659 / 12.0 / fix loader.conf.local being lost on upgrade

yocalebo:
There was a singular issue. `/boot/loader.conf.local` was being "lost" on upgrades....kind of. `kenv` didn't report the values set by us but the file had the correct contents. I spun up a local VM and delved into the problem. Slowly but inexorably crawling upon my consciousness and rising above every other impression, came a dizzying dissonance between the upgrade code and my personal fear. I stared into the abyss, and it stared back. I've barely made it out alive (thankful to the ones that helped me through this trying time 😄 )

I've found a few issues that I've tried to address with this PR.

- `freenas-update` creates `/update-scritps` every time which means `ix-update-scripts` runs and as far as I can comprehend, `ix-update` needs to run after `ix-update-scripts`
- `ix-update` runs `BEFORE: middlewared` which means the `midclt` call will fail
- partially revert 7d5f773ffa4338ebab1bc11ef605b1ee12eb046e because it introduced the fact that we were creating a `/boot/loader.conf.local` file at _BUILD_ time so when an upgrade occurred it would overwrite the current `/boot/loader.conf.local` file.
- etc config and loader files are executable

### DIFF
--- a/nas_ports/freenas/freenas-files/files/pkg-install.in
+++ b/nas_ports/freenas/freenas-files/files/pkg-install.in
@@ -36,12 +36,6 @@ post-install()
         fi
 
         cp /data/freenas-v1.db /data/factory-v1.db
-
-        if ! dmidecode -s system-product-name | egrep "^(FREE|TRUE)NAS" > /dev/null; then
-                for i in `seq 0 3`; do
-                        echo "hint.isp.$i.role=2" >> /boot/loader.conf.local
-                done
-        fi
 }
 
 backupwarning()
--- a/src/freenas/etc/ix.rc.d/ix-update
+++ b/src/freenas/etc/ix.rc.d/ix-update
@@ -4,6 +4,7 @@
 #
 
 # PROVIDE: ix-update
+# REQUIRE: ix-update-scripts
 # BEFORE: middlewared earlykld
 
 . /etc/rc.freenas
@@ -88,9 +89,21 @@ db_update()
 		if [ -f /usr/local/sbin/firmware_update.py ]; then
 			LD_LIBRARY_PATH=/usr/local/lib /usr/local/bin/python /usr/local/sbin/firmware_update.py
 			if [ $? -eq 0 ]; then
-				# A reboot will be required, but we'll hold off
-				# on that and do database upgrades if they need
-				# to be done as well to avoid a double reboot.
+				# Users often wonder why a 3rd reboot occurs on their system
+				# when they upgrade their box. This is the reason why.
+				# To keep the long, drawn out story as short as possible
+				# and to maintain the fickle state that my mind is in
+				# after troubleshooting upgrade code for an entire day,
+				# /update-scripts gets created on upgrade by default
+				# in the Installer.py code (freenas-update).
+				# ix-update-scripts runs before this script and reboots
+				# the box before this rc script gets called.
+				# So a 2nd reboot occurs, and if the system that is being
+				# upgraded also has a SAS HBA card that received a
+				# firmware upgrade, then we set this variable to 1. As
+				# you may already guess, a 3rd reboot occurs. This will
+				# make you drink, if you don't already and/or step away
+				# from the computer and make you question your life choices.
 				REBOOT_REQUIRED=1
 			fi
 		else
@@ -143,8 +156,6 @@ db_update()
 
 	rm -f $NEED_UPDATE_SENTINEL
 	rm -f $CD_UPGRADE_SENTINEL
-
-	/usr/local/bin/midclt call etc.generate loader > /dev/null
 
 	echo "Database upgrade complete.  Rebooting."
 	cd /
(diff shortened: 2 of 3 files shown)


src/middlewared/middlewared/utils/io.py
```diff
@@ -8,7 +8,7 @@ def write_if_changed(path, data):
 
     changed = False
 
-    with open(os.open(path, os.O_CREAT | os.O_RDWR), 'wb+') as f:
+    with open(path, 'wb+') as f:
```
william-gr, line 11: I remember there was a good reason to use os.open but now I don't remember why. Yay for comments in the code.

Have you tested this extensively?

william-gr: I think the reason was because without os.open we couldnt actually READ *and* WRITE.
But could be a python bug at the time.

william-gr: @themylogin do you remember the exact reason?

yocalebo: Yeah, I think it was probably a python bug at the time. `wb+` should open for read and write.

yocalebo: I'm actually waiting to build a custom image with these changes (and the PR from the build repo) so I can test more thoroughly.

chris3torek: > @themylogin do you remember the exact reason?

open(path, "wb+") includes `os.O_TRUNC`.

themylogin, line 11: Every doc says the contrary, and this still does not work
```
root@truenas[~]# echo "long text" >> file
root@truenas[~]# python                  
Python 3.8.5 (default, Dec  3 2020, 23:15:58) 
[Clang 10.0.1 (git@github.com:llvm/llvm-project.git llvmorg-10.0.1-0-gef32c611a on freebsd12
>>> open('file').read()
'long text\n'
>>> f = open('file', 'wb+')
>>> f.seek(0)
0
>>> f.read()
b''
```

themylogin (changes requested):

src/middlewared/middlewared/utils/io.py
yocalebo: It was truncated because of the `w` option.

yocalebo: `>>> f = open('testing.txt', 'wb+')
>>> f.write(b'neat')
4
>>> f.flush()
>>> f.seek(0)
0
>>> f.read()
b'neat'`

william-gr: Thanks @chris3torek ! Glad you're still around checking our mistakes, hope you + family are doing OK ; )

william-gr: @yocalebo that wont work on our use case because the file already exists. So the issue was it was getting truncated, not read+write

yocalebo: The real problem, actually, is that `os.open` is missing the `mode` option.

yocalebo: Fixed 2 issues.
- specified mode
- changed to `rb+` to not cause confusion on what we're actually trying to do

src/middlewared/middlewared/utils/io.py
```diff
@@ -8,7 +8,7 @@ def write_if_changed(path, data):
 
     changed = False
 
-    with open(os.open(path, os.O_CREAT | os.O_RDWR), 'wb+') as f:
+    with open(os.open(path, os.O_CREAT | os.O_RDWR, 0o644), 'rb+') as f:
```
william-gr, line 11: Not all files are 644, we have mode specified in the plugin for some files, don't we?

yocalebo: I honestly can't see why any of the config files for the various services would need to be executable. Only possibility I can think of is if we're using the `etc` plugin to generate scripts that are run by various services which is quite plausible I guess. Nonetheless, you're right. We have the `mode` key in `etc` plugin. I don't really feel like testing changing the default mode for all consumers of this method so I'll change in `etc` plugin.

src/middlewared/middlewared/plugins/etc.py
```diff
@@ -112,7 +112,7 @@ class EtcService(Service):
             {'type': 'mako', 'path': 'local/openldap/ldap.conf'},
         ],
         'loader': [
-            {'type': 'py', 'path': 'loader', 'platform': 'FreeBSD'},
+            {'type': 'py', 'path': 'loader', 'platform': 'FreeBSD', 'mode': 0o0644},
```
william-gr, line 115: `mode` here without any other changes is a no-op AFAIK

william-gr (approved):

themylogin (approved):

PR #6171, opened 2020-12-16
NAS-108560 / 20.12 / Fix zfs-zed

themylogin:


debian/debian/postinst
```diff
@@ -42,8 +40,7 @@ systemctl disable smbd
 systemctl disable winbind
 systemctl disable wsdd
 
-# kubelet and nvidia-persistenced are not respecting vendor preset file so we disable them explicitly
-systemctl disable kubelet
```
william-gr, line 46: kubelet started to respect preset?

themylogin: kubelet does not exist anymore

william-gr (approved):
Joe reported zed alone didn't fix the problem so there is more to the problem

PR #6173, opened 2020-12-17
NAS-108609 / 12.0 / Do not use fake serial numbers for multipath

themylogin:


src/middlewared/middlewared/plugins/disk.py
```diff
@@ -534,6 +534,9 @@ async def multipath_sync(self):
             serial = ''
             v = g.provider.config.get('ident')
             if v:
+                # Exclude fake serial numbers e.g. `000000000000` reported by FreeBSD 12.2 USB stack
+                if not v.replace("0", ""):
+                    continue
```
william-gr, line 539: How hard is it to also exclude disks sitting on top of USB "controller" like mav suggested?

william-gr (approved):
Thanks!

src/middlewared/middlewared/plugins/disk.py
amotin: `000000000000` is reported not by FreeBSD 12.2 USB stack, but by ugly USB bridge firmware.  We saw many different variants there, so attempt to blacklist all of them is probably hopeless.

PR #6174, opened 2020-12-17
NAS-108685 / 21.02 / Fix traceback trying to export an offline pool

themylogin:


src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -698,6 +698,7 @@ async def extent_extend_context(self, extra):
             'pools': {
                 p['name']: {'disks': await self.middleware.call('zfs.pool.get_disks', p['name']), 'all_flash': False}
                 for p in await self.middleware.call('pool.query')
```
william-gr, line 700: Shall we use pool.query to filter?

william-gr (approved):
Thanks!

PR #6175, opened 2020-12-17
NAS-105062 / 21.02 / Allow empty specific attributes for NIC

william-gr:
Ticket:	NAS-105062

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -895,7 +895,7 @@ class InterfaceService(CRUDService):
             required_attrs = ('vlan_parent_interface', 'vlan_tag')
 
         for i in required_attrs:
-            if not data.get(i):
+            if i not in data:
                 verrors.add(f'interface_create.{i}', 'This field is required')
 
         verrors.check()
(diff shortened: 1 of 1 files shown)


themylogin (commented):
It works well for memberless bridge, but how will it work for VLAN without a parent interface or LAGG without a protocol?

william-gr:
@themylogin isn't there code making sure vlan parent iface/lagg proto/etc others are valid?

themylogin:
@william-gr it will pass VLAN without parent https://github.com/freenas/freenas/blob/master/src/middlewared/middlewared/plugins/network.py#L1143 but is it possible to have a VLAN without parent?

william-gr:
Good catch, I will add an else

themylogin:
@william-gr for an empty lagg protocol it will have a poor error message. It will also permit LAGG without interfaces and allow VLAN with empty tag.

william-gr:
Dont think its a poor error message. Its accurate.

There is no null=True for vlan tag

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1110,9 +1110,12 @@ async def _common_validation(self, verrors, schema_name, data, itype, update=Non
             if data['lag_protocol'] not in await self.middleware.call('interface.lag_supported_protocols'):
                 verrors.add(
                     f'{schema_name}.lag_protocol',
-                    f'{platform.system()} does not support LAGG protocol {data["lag_protocol"]}',
+                    f'{platform.system()} does not support LAG protocol {data["lag_protocol"]}',
```
themylogin, line 1113: if `data["lag_protocol"]` is empty, error message will be `"FreeBSD does not support LAG protocol "`

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1110,9 +1110,12 @@ async def _common_validation(self, verrors, schema_name, data, itype, update=Non
...
                 )
-            for i, member in enumerate(data.get('lag_ports') or []):
+            lag_ports = data.get('lag_ports') or []
```
themylogin, line 1115: What is the point of `or []` now?

themylogin (changes requested):

src/middlewared/middlewared/plugins/network.py
william-gr: Can it be empty if we have enum in the schema?

themylogin: Good point, sorry

themylogin (approved):

src/middlewared/middlewared/plugins/network.py
william-gr: Removed

themylogin (approved):

william-gr:
Thanks for not allowing me screw up :)

william-gr:
backport

PR #6179, opened 2020-12-17
NAS-108714 / 20.12 / installer: Clarify preferred media for install

freqlabs:
"flash media" makes it sound like USB flash is OK to install to.
We do not want to recommend it.

### DIFF
--- a/src/freenas-installer/etc/install.sh
+++ b/src/freenas-installer/etc/install.sh
@@ -271,10 +271,10 @@ EOD
 
     cat << EOD >> "${_tmpfile}"
 - You can't use ${_disks} for sharing data.
-
+ 
 NOTE:
-- Installing on SATA, SAS, or NVMe flash media is preferred to
-  installing on hard drives or USB flash sticks.
+- Installing on SATA, SAS, or NVMe flash media is recommended.
+  USB flash sticks are discouraged.
 
 Proceed with the ${_type}?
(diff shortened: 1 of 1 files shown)


yocalebo:
Being nit-picky but maybe say SATA, SAS or NVMe flash media.

freqlabs:
Should I discourage SATA DOMs as well?

william-gr:
SATA DOMs are not bad, the specific ones we used were, IMO

yocalebo (approved):

yocalebo:
@Mrt134 We want to make it obvious that we don't recommend installing flash usb sticks. I think this verbiage makes sense and gets the point across but would like you to give final review since you're docs 😄

kmoore134 (changes requested):
Recommend the following wording:

Installing on SATA, SAS, or NVMe flash media is preferred. USB flash sticks are unsupported.

kmoore134:
Ok, slight wording change after review with William:

"Installing on SATA, SAS, or NVMe flash media is recommended. USB flash sticks are discouraged."

Mrt134 (commented):
Suggest we explicitly discourage USB installs as it only appears to lead to trouble and a poor customer experience. Kris' suggestion should do the trick.

william-gr (approved):

PR #6181, opened 2020-12-18
NAS-108720 / 21.02 / Do not background Samba path creation during sysdataset setup

anodos325:
Other tasks may be dependent on these paths actually existing because samba libraries may not be able to generate a messaging context.

### DIFF
--- a/src/middlewared/middlewared/plugins/sysdataset.py
+++ b/src/middlewared/middlewared/plugins/sysdataset.py
@@ -295,7 +295,10 @@ class SystemDatasetService(ConfigService):
             if osc.IS_LINUX:
                 await self.middleware.call('etc.generate', 'glusterd')
 
-            await self.middleware.call('smb.configure')
+            await self.middleware.call('smb.setup_directories')
+            # The following should be backgrounded since they may be quite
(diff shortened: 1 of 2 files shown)


themylogin (approved):

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -295,7 +295,10 @@ async def setup(self, mount, exclude_pool=None):
             if osc.IS_LINUX:
                 await self.middleware.call('etc.generate', 'glusterd')
 
-            await self.middleware.call('smb.configure')
+            await self.middleware.call('smb.setup_directories')
+            # The following should be backgrounded since they may be quite
+            # long-running.
```
william-gr, line 300: A bit confused, why dont we do that, then?

william-gr: Ah, they are jobs already, nevermind

themylogin: Because it's a job and we don't wait for it to complete

william-gr (approved):

PR #6184, opened 2020-12-18
NAS-108724 / 21.02 / if serial input is enabled, read the input buffer to clear any garbage

dhiltonp:
On my hardware, `serial` *always* has some garbage after initialization.

Grub reads that garbage input and disables the timeout.

If we wait 1 second with `--interruptible` set, grub reads the buffer (`ESC` early aborts the sleep). 

In order for this to *not* be sufficient, ESC (`10`(STOP/START)+`00011011`+`1`(STOP)) would have to be read on the byte boundary. This is possible, but it seems very unlikely.


william-gr (approved):

william-gr:
ok to test

PR #6204, opened 2020-12-22
NAS-108762 / 12.0 / Create VLAN network interface database record from netcli

themylogin:


yocalebo (commented):
Am I correct in understanding that this is creating the parent and vlan interface in the `network_interfaces` table if they don't already exist?

william-gr (approved):

william-gr:
Thats what I am gathering from it. @themylogin ?

themylogin:
@yocalebo it was only creating parent interface, not it's also creating VLAN interface itself. It was only creating `network_vlan` table row previously, which led to an inconsistent database state.

PR #6207, opened 2020-12-28
NAS-108822 / 21.02 / return IPv6 addresses in failover.get_ips

yocalebo:


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -346,17 +346,29 @@ def in_progress(self):
...
+            for i in await self.middleware.call('interface.query', [('failover_vhid', '!=', None)]):
+                # user can bind to a single v4 address but all v6 addresses
+                # or vice versa
+                addrs.extend([
+                    x['address'] for x in i.get('failover_virtual_aliases', [])
+                    if x['type'] == 'INET' and all_ip4
+                ])
+                addrs.extend([
+                    x['address'] for x in i.get('failover_virtual_aliases', [])
+                    if x['type'] == 'INET6' and all_ip6
```
william-gr, line 367: Would it make sense to use 

`if (x['type'] == 'INET' and all_ip4) or (x['type'] == 'INET6' and all_ip6)`

Just so we dont iterate twice?

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -346,17 +346,29 @@ def in_progress(self):
...
+                ])
+            return addrs
```
william-gr, line 369: So if we have all ipv4 or all ipv6 we will only return these? and not selected ipv4 or ipv6?

themylogin (approved):

william-gr (approved):

PR #6209, opened 2020-12-29
NAS-108666 / 21.02 / prevent netcli from hanging on sockstat calls by redoing get_ui_urls

yocalebo:
This fixes a few issues.

- `sockstat` can block on FreeBSD `getpwuid()` calls if joined to a directory service and the connection to said service is not "healthy". I've since patched `sockstat` to actually ignore trying to resolve numeric UIDs to user names here: [freenas/os repo](https://github.com/freenas/os/pull/235/commits/447470473e51597ff6c028d3435ffff34a628be4) which was pushed upstream here: [upstream commit](https://cgit.freebsd.org/src/commit/?id=ccdd2b2b3cc2f86cd08a355be2fb58e435ec8ff8)
- However, I've removed the call to `sockstat` since we can get the same information by running `psutil.net_connections()`
- Furthermore, (based on a previous discussion in this PR) there is absolutely no reason to run `requests.head()` for each IP address that returned from the `psutil.net_connections()` and/or `interface.ip_in_use`. There is only 1 consumer of this method (`netcli`) and all this function does is return a list of formatted URLs to be displayed on the console.
- Running `requests.head()` doesn't scale either since we have customer systems with 50+ IP addresses on their system so this could potentially do a HEAD request for each http IP and https IP with timeouts for each set at 10 and 15 seconds respectively. This is unacceptable. Furthermore, trying to run all these requests in parallel would solve 1 problem but would introduce another one altogether. I.E. what happens if the total time given to run all the HEAD requests expires? Do we take only the completed parallel tasks or do we not include any of them? What if a system has 50+ IPs on their system? Do we increase the total time for all parallel tasks 3*total_ips? Do we have to increase the total timeout if the system has above a certain threshold of IPs? 100? 200? An unnecessary problem to give ourselves when this is simply returning formatted URLs to the console.
- before changes, the method didn't account for HA systems (VIPs are the only functioning IP that can be used when accessing the webUI)

To summarize:
- remove `sockstat` call
- no longer perform any type of HTTP HEAD request on the IPs dramatically speeding up the time it takes for this method to run
- make it work with HA VIPs

### DIFF
--- a/src/freenas/etc/netcli
+++ b/src/freenas/etc/netcli
@@ -1482,8 +1482,7 @@ def show_ip():
         print()
         print(_("The web user interface is at:"))
         print()
-
-        for url in sorted(urls):
+        for url in urls:
             print(url)
     else:
         print()
--- a/src/middlewared/middlewared/plugins/system.py
+++ b/src/middlewared/middlewared/plugins/system.py
@@ -1414,101 +1414,86 @@ class SystemGeneralService(ConfigService):
 
         raise CallError('Unable to connect to any of the specified UI addresses:\n' + '\n'.join(errors))
 
-    def __get_urls(self, aliases, addrs, ipv6=False):
-
-        skip_internal = False
-        if self.middleware.call_sync('system.product_type') == 'ENTERPRISE':
-            skip_internal = True
-
-        urls = []
-        for addr in addrs:
-            ip, port = addr.split(':')
-
-            if ip == '*':
-                ips = [
-                    i["address"]
-                    for i in aliases
-                    if i['type'] == ('INET6' if ipv6 else 'INET')
-                ]
-            else:
-                ips = [ip]
-
-            for o in ips:
-                if skip_internal and o in (
-                    '169.254.10.1',
-                    '169.254.10.2',
-                    '169.254.10.20',
-                    '169.254.10.80',
-                ):
-                    continue
-
-                if ipv6 and '%' in o:
-                    o = o.split('%')[0]
-
-                if ipv6:
-                    url = f'http://[{o}]'
-                else:
-                    url = f'http://{o}'
-                if port != '80':
-                    url = f'{url}:{port}'
-                try:
-                    r = requests.head(url, timeout=10)
-                    assert r.status_code in (200, 302, 301)
-                    urls.append(url)
-                    continue
-                except Exception:
-                    pass
-
-                if ipv6:
-                    url = f'https://[{o}]'
-                else:
-                    url = f'https://{o}'
-                if port != '443':
-                    url = f'{url}:{port}'
-                try:
-                    r = requests.head(url, timeout=15, verify=False)
-                    assert r.status_code in (200, 302)
-                    urls.append(url)
-                    continue
-                except Exception:
-                    pass
-        return urls
-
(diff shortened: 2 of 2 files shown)


themylogin:
I don't see any DNS lookups performed in https://github.com/freebsd/freebsd/blob/master/usr.bin/sockstat/sockstat.c but I feel it might be worth investigating and fixing, because the proposed method of determining the listening ports seems a little bit crazy for me :) Or am I missing something?

yocalebo:
What specifically is crazy?

All I'm doing is:
1. `interface.query` and get VIPs (if HA system) other wise aliases
2. pass that to a function that runs `interface.ip_in_use`
3. generate a list of addresses that are configured and show up in `interface.ip_in_use`
4. then I simply try to make a (http/https) `HEAD` request to each IP based on what the end-user has configured in the webUI

themylogin:
@yocalebo retrieving listening ports should be as simple as reading some kernel data structures, I don't like step 4 especially, you shouldn't need to connect to local port to check if we are listening on it, it's like:
```
 - How do you determine if a database exists?
 - if Query.Exec('drop database "name"') then showmsg('database existed').
```
If you want to stick to that approach for any other reason, let's make this `for` loop parallel (so in worst case it hangs for 5 seconds instead of 5*N seconds)

yocalebo:
@themylogin Before my change, we were serializing `request.head()` requests after running `sockstat` output and filtering for what IP's `nginx` is listening on. We were doing that for each detected IPv4 and IPv6 addresses in separate loops and specifying different timeouts whether or not it was `http` or `https`.

anodos325:
I do see a call to getpwuid() in sockstat output. I have seen some pathological cases with directory services where this can hang (often because DNS). It would be nice if there was a switch for sockstat to not try this (just display numeric ids).

yocalebo:
> I do see a call to getpwuid() in sockstat output. I have seen some pathological cases with directory services where this can hang (often because DNS). It would be nice if there was a switch for sockstat to not try this (just display numeric ids).

Yeah, I should clarify a little more. I don't see a flat to disable this behavior (unless you see one)

themylogin:
> I don't see a flag to disable this behavior (unless you see one)

That should be pretty simple to patch

> Before my change, we were serializing request.head()

Sorry, didn't see it. My proposal to make these requests parallel is still valid then :) Also why don't you remove `__get_urls` method if no one is using it anymore?

yocalebo:
> That should be pretty simple to patch

So instead of simply working around the issue, you'd rather patch an in-base utility that will almost never get up-streamed further diverging our freeBSD OS port? I'm not sure which method seems more crazy 😄

> Sorry, didn't see it. My proposal to make these requests parallel is still valid then :) Also why don't you remove `__get_urls` method if no one is using it anymore?

Yep, seems I've forgotten to remove the unused `__get_urls` method. I agree, and will investigate parrallelizing the requests method.

themylogin:
> you'd rather patch an in-base utility that will almost never get up-streamed further diverging our freeBSD OS port?

If we are doing HTTP requests anyway, I am alright with not patching sockstat if we are not using it :)

themylogin (approved):

PR #6210, opened 2020-12-29
NAS-108837 / 21.02 / fix _register_virthostname in ad plugin

yocalebo:
- flake8 fixes
- `failover.get_ips` does not return anything related to HA IPs (despite the horrible name), it returns IP addresses that the webUI (`nginx`) is bound to. So, instead, call `interface.ip_in_use` which returns known working VIP addresses on HA systems.


yocalebo:
retest this please

william-gr:
@anodos325 ping

anodos325 (approved):

PR #6219, opened 2021-01-05
NAS-108900 / 12.0 / Add exception handling for setting NTP in AD plugin

anodos325:
Setting the NTP server may fail (ValidationError). In general case,
failure in this method should not be fatal for domain join process.
Wrap in try/except and log it.

### DIFF
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -1138,7 +1138,12 @@ class ActiveDirectoryService(ConfigService):
                                 "Failed to automatically set time source.")
             return
 
-        self.middleware.call_sync('system.ntpserver.create', {'address': pdc, 'prefer': True})
+        try:
+            self.middleware.call_sync('system.ntpserver.create', {'address': pdc, 'prefer': True})
+        except Exception:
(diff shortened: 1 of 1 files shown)


william-gr (approved):

src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1138,7 +1138,12 @@ def set_ntp_servers(self):
                                 "Failed to automatically set time source.")
             return
 
-        self.middleware.call_sync('system.ntpserver.create', {'address': pdc, 'prefer': True})
+        try:
+            self.middleware.call_sync('system.ntpserver.create', {'address': pdc, 'prefer': True})
```
william-gr, line 1142: Hmm, wonder, what if for some reason ntp server is already there?

william-gr: @anodos325 ping

anodos325: Before we try to add PDC emulator we first check for whether the NTP server list is in the default out-of-box state.

william-gr: Ah, I see it now. For the master PR we have to change the default server url

PR #6227, opened 2021-01-06
NAS-108796 / 21.02 / Do not receive sharenfs and sharesmb properties when replicating

themylogin:


src/middlewared/middlewared/plugins/zettarepl.py
```diff
@@ -554,6 +554,13 @@ async def get_definition(self):
                                           for periodic_snapshot_task in replication_task["periodic_snapshot_tasks"]]
             my_schedule = replication_task["schedule"]
 
+            properties_exclude = replication_task["properties_exclude"].copy()
+            properties_override = replication_task["properties_override"].copy()
+            for property in ["sharenfs", "sharesmb"]:
```
william-gr, line 559: Guess based on what you said in the other ticket we can also just add mountpoint here and call it a day?

william-gr (approved):

PR #6230, opened 2021-01-06
NAS-108749 / 21.02 / Remove HTML from alert services that do not support it. Allow HTML-only e-mails

themylogin:


william-gr (commented):
Does that mean text emails will no longer be sent?

Thats not the expectation, some dinosaurs prefer to see it in plain text

anodos325:
> Thats not the expectation, some dinosaurs prefer to see it in plain text

Hey, I resemble that remark.

themylogin:
@william-gr 

> Does that mean text emails will no longer be sent?

No, it means that it's now possible to specify `html` and omit `text` for an e-mail, and text part will be an automatically generated markdown-ish version of HTML.
All other alert services (slack, telegram, etc.) will also receive this markdown version of alert message.

william-gr (approved):
Ah, I see it now. Thanks!

PR #6247, opened 2021-01-08
NAS-108691 / 21.02 / Add ability to retrieve and validate used/unused ports in Apps

sonicaj:
This PR introduces changes to retrieve used ports by Applications and also adds validation to ensure that we prohibit using a used port.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/validation.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/validation.py
@@ -8,7 +8,9 @@ from .schema import get_schema, get_list_item_from_value, update_conditional_def
 from .utils import RESERVED_NAMES
 
 
-validation_mapping = {}
+validation_mapping = {
+    'validations/nodePort': 'port_available_on_node',
+}
 
 
 class ChartReleaseService(Service):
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/chart_releases_linux/validation.py
```diff
@@ -75,20 +79,31 @@ async def validate_question(self, verrors, value, question, var_attr, schema_nam
...
         return verrors
+
+    @private
+    async def validate_port_available_on_node(self, verrors, value, question, schema_name, release_data):
+        if release_data and value in [p['port'] for p in release_data['used_ports']]:
+            # TODO: This still leaves a case where user has multiple ports in a single app and mixes
+            #  them to the same value however in this case we will still get an error raised by k8s.
+            return
+
+        if value in await self.middleware.call('chart.release.used_ports'):
```
themylogin, line 108: What about 80, 135-139, 443, 2049 and all other ports that can be busy?

sonicaj: The minimum limit right now is 9000 for Apps. However now that check requires a different approach and we can certainly do that. Please advise if you think we should look into that ?

sonicaj: ( to clarify where we are looking at ports in use other then Apps )

themylogin: Forgot about that, it's ok then

themylogin (approved):

PR #6249, opened 2021-01-08
NAS-108889 / 12.0 / Split middleware and winbind netlogon traffic between DCs

anodos325:
In some edge cases it appears that middleware performing a netlogon
connection can trample winbindd's netlogon session causing winbind
operations to fail with STATUS_RPC_SEC_PKG_ERROR. This commit splits
up two things that are vital for checking overall status of our
domain join. (1) health of winbindd's netlogon connection to the
current DC it's using and (2) the ability to fail over to a working
DC if the one winbind is currently connected to becomes inaccessible.

Checks for (2) are rolled into the general periodic AD domain health
check - once per hour, and (1) is switched back to a simple "wbinfo -P" command.

### DIFF
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -12,7 +12,7 @@ import threading
 import time
 
 from dns import resolver
-from middlewared.plugins.smb import SMBCmd, SMBPath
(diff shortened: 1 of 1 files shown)


anodos325:
get_n_working_servers() will only return DCs to which we can successfully connect. This will catch errors more specifically related to our account in AD (getting locked, broken replication, etc).

william-gr (approved):

PR #6250, opened 2021-01-09
NAS-108604 / 21.02 / Make restarting attachments optional on dataset unlock

sonicaj:


themylogin (approved):

yocalebo (changes requested):
This breaks HA systems.

`unlock_zfs_datasets` uses `toggle_attachments` in `pool.dataset.unlock`

sonicaj:
@yocalebo thank you for pointing that out. I'll fix it.

yocalebo (approved):

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2248,7 +2246,7 @@ async def detach(delegate):
             'unlock_options',
             Bool('key_file', default=False),
             Bool('recursive', default=False),
-            Bool('toggle_attachments', default=True),
+            List('services_restart', default=[]),
```
william-gr, line 2249: Was the API change necessary? Backward incompatible API change... @sonicaj

PR #6260, opened 2021-01-12
NAS-108951 / 21.02 / dont try to fuse mount ctdb shared volume if peers are disconnected

yocalebo:


src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
```diff
@@ -186,7 +210,7 @@ def umount(self, job):
         cmd = ['umount', '-R', CTDB_LOCAL_MOUNT]
         cp = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if cp.returncode:
-            if b'not mounted' in cp.stderr:
+            if b'not mounted' or b'ctdb_shared_vol: not found' in cp.stderr:
```
themylogin, line 213: It is always true-ish

themylogin (changes requested):

src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
yocalebo: Thanks

themylogin (approved):

PR #6268, opened 2021-01-13
NAS-108967 / 21.02 / fix mount/umount event loop when gluster volume is self-healing

yocalebo:
This fixes 2 issues:

- fix a mount/umount event loop when the ctdb shared volume is not healthy (1 of the peers is down)
- mount the ctdb shared volume when the system is booting

### DIFF
--- a/src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
+++ b/src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
@@ -46,10 +46,9 @@ class CtdbSharedVolumeService(Service):
     @private
     def shared_volume_exists_and_started(self):
 
-        exists = started = vol = False
+        exists = started = False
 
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
```diff
@@ -196,6 +193,7 @@ def mount(self, job):
         else:
             mounted = True
 
+        self.middleware.call_sync('service.start', 'glustereventsd')
```
themylogin, line 196: Will we benefit from try/finally block to ensure that glustereventsd is always started again despite any (now seemingly impossible, but still) error?

yocalebo: Good idea!

themylogin (approved):

PR #6269, opened 2021-01-13
NAS-108970 / 21.02 / fix 2 issues with ganesha mako template

yocalebo:
Fixes 2 issues:

- running `gluster.peer.status` will traceback if glusterd isn't started which is guaranteed to be the case when the system is freshly booting
- if there are no NFS shares that are enabled, then don't continue and bail out early

### DIFF
--- a/src/middlewared/middlewared/etc_files/ganesha/ganesha.conf.mako
+++ b/src/middlewared/middlewared/etc_files/ganesha/ganesha.conf.mako
@@ -15,8 +15,7 @@
     # n times (n being the number of shares in db)
     peers = []
     if middleware.call_sync("service.started", "glusterd"):
-        peer_job = middleware.call_sync("gluster.peer.status")
-        peers = peer_job.wait_sync()
+        peers = middleware.call_sync("gluster.peer.status")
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/etc_files/ganesha/ganesha.conf.mako
```diff
@@ -4,15 +4,19 @@
...
     gc = middleware.call_sync("network.configuration.config")
 
     # call this here so that we don't have to call this
     # n times (n being the number of shares in db)
-    peer_job = middleware.call_sync("gluster.peer.status")
-    peers = peer_job.wait_sync()
+    peers = []
+    if middleware.call_sync("service.started", "glusterd"):
+        peer_job = middleware.call_sync("gluster.peer.status")
+        peers = peer_job.wait_sync()
```
themylogin, line 19: Is it supposed to always return fast? We are waiting for a job in a non-job.

yocalebo: Just reminded me this isn't a job anymore with recent changes.

themylogin (approved):

PR #6271, opened 2021-01-14
NAS-108979 / 21.02 / Properly setup k3s agent

sonicaj:
This PR adds changes accounting for upstream k3s changes where registered worker nodes passwords are not available in a file anymore and they are only available now on worker node's `/etc/rancher/node/password` path. However this is lost on system upgrades, so in order to have it persist, we use a static constant value for the node worker password. ( https://github.com/k3s-io/k3s/issues/2813 )

For existing users already using k8s, we delete the worker node which forces k3s to re-register the node and that works nicely without us losing data.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/utils.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/utils.py
@@ -1,4 +1,5 @@
 BACKUP_NAME_PREFIX = 'ix-applications-backup-'
 KUBECONFIG_FILE = '/etc/rancher/k3s/k3s.yaml'
+KUBERNETES_WORKER_NODE_PASSWORD = 'e3d26cefbdf2f81eff5181e68a02372f'
 NODE_NAME = 'ix-truenas'
 UPDATE_BACKUP_PREFIX = 'system-update-'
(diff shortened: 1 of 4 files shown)


themylogin (commented):
Can you just set it to constant value before cluster initialization?

sonicaj:
@themylogin it should, `secrets.token_hex(16)` should do the job nicely i think, however this behavior is not documented. I'll test and make changes and we can see if there are issues with it down the road.

themylogin:
@sonicaj can't we just have constant value for all TrueNAS installations?

themylogin (approved):

src/middlewared/middlewared/plugins/kubernetes_linux/utils.py
```diff
@@ -1,4 +1,5 @@
 BACKUP_NAME_PREFIX = 'ix-applications-backup-'
 KUBECONFIG_FILE = '/etc/rancher/k3s/k3s.yaml'
+KUBERNETES_WORKER_NODE_PASSWORD = 'e3d26cefbdf2f81eff5181e68a02372f'
```
william-gr, line 3: Same password for everybody? Doesn't this open possibility for a local exploit?

sonicaj: We are going to disallow any more nodes from joining the k3s cluster in another PR ( this is not happening right now ), which means that we won't have more nodes joining, so this does not matter ?

william-gr (approved):

PR #6275, opened 2021-01-14
NAS-108985 / 12.0 / Only output last 1000 lines of zpool history ioctls

themylogin:


src/freenas/usr/local/libexec/freenas-debug/zfs/zfs.sh
```diff
@@ -60,8 +60,12 @@ zfs_func()
 	zpool status -v
 	section_footer
 
-	section_header "zpool history -i"
-	zpool history -i
+	section_header "zpool history"
+	zpool history
+	section_footer
+
+	section_header "zpool history -i | tail -n 1000"
```
yocalebo, line 67: I'd rather just remove the `zpool history -i` command since that produces duplicate output from the `zpool history` command. If we need to get the detailed logs (`-i`) of zpool history, then it can be asked for explicitly.

@william-gr but that's really up to you I guess. The other changes look fine.

yocalebo (changes requested):

src/freenas/usr/local/libexec/freenas-debug/zfs/zfs.sh
william-gr: I dont recall the reason -i was added. I guess there are things that play directly with libzfs, e.g. py-libzfs and samba.
That said I think its probably worthwhile to still keep -i?

yocalebo: Fair enough

yocalebo (approved):

william-gr (approved):

PR #6278, opened 2021-01-14
NAS-108956 / 12.0 / Always terminate rsync if pool import disk fails

themylogin:


william-gr:
@themylogin user states he kept seeing it even after a reboot? Should we make sure to remove any files before starting?

themylogin:
@william-gr he sees that because there is an exception somewhere in this code, which we can't see because it is masked by an exception in mount context manager __exit__. There are leftover rsync processes which prevent unmounting and removing mountpoints.

william-gr (approved):

PR #6280, opened 2021-01-14
NAS-108154 / 12.0 / max_parallel_replication_tasks

themylogin:
@william-gr I am very sorry, I pushed this over a month ago but somehow forgot to create a PR. UI counterpart is not documented and implemented too, although it's only a form of one integer field.


william-gr:
@themylogin we will still be missing UI side, right?

themylogin:
Yes, that's what I've written in the first comment :(

william-gr:
Ah missed that, sorry

william-gr:
Do you know if there is a ticket for them?

themylogin:
@william-gr I usually create them after we merge so they din't need to patch middleware

william-gr (approved):

william-gr:
@themylogin is this OK to merge? Mostly worried about migration revs that may have changed

themylogin:
@william-gr it's ok, we did not have any migrations since then

PR #6281, opened 2021-01-15
NAS-108987 / 21.06 / Unable to import pool - pool shows up in dialog import list, but fails at zfs.find_import

dhiltonp:
This is just a quick hack to look for importable pools on whole disks, too.

See discussion here: https://jira.ixsystems.com/browse/NAS-108987

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/disk_info_linux.py
+++ b/src/middlewared/middlewared/plugins/disk_/disk_info_linux.py
@@ -117,7 +117,14 @@ class DiskService(Service, DiskInfoBase):
 
     def label_to_disk(self, label, *args):
         part_disk = self.label_to_dev(label)
-        return self.get_disk_from_partition(part_disk) if part_disk else None
+        if not part_disk:
+            # the label is already a disk
+            return label
+        elif os.path.exists(os.path.join('/sys/block', part_disk)):
+            # it's imported with a label, but it's a whole disk
+            return part_disk
+        else:
+            return self.get_disk_from_partition(part_disk)
 
     def get_disk_from_partition(self, part_name):
         if not os.path.exists(os.path.join('/dev', part_name)):
--- a/src/middlewared/middlewared/plugins/zfs.py
+++ b/src/middlewared/middlewared/plugins/zfs.py
@@ -369,7 +369,7 @@ class ZFSPoolService(CRUDService):
         found = False
         with libzfs.ZFS() as zfs:
             for pool in zfs.find_import(
-                cachefile=cachefile, search_paths=['/dev/disk/by-partuuid', 'dev/disk/by-uuid'] if osc.IS_LINUX else None
+                cachefile=cachefile, search_paths=['/dev/disk/by-partuuid', '/dev'] if osc.IS_LINUX else None
             ):
                 if pool.name == name_or_guid or str(pool.guid) == name_or_guid:
                     found = pool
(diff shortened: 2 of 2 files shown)


dhiltonp:
Tested locally, works for me.

dhiltonp:
Hold that, I bumped in to another issue.

dhiltonp:
I initially was passing `/dev/disk/by-uuid` as the 2nd paramater, zpool status showed:

	NAME                     STATE     READ WRITE CKSUM
	tank                     ONLINE       0     0     0
	  mirror-0               ONLINE       0     0     0
	    1824969785541290475  ONLINE       0     0     0
	    sdb                  ONLINE       0     0     0

The issue is that TrueNAS doesn't realize that `1824969785541290475` is `/dev/sda`, and so `/dev/sda` shows up as available for use in other vdevs.

I've updated the code to pass `/dev`...

I tested on the command line with multiple -d switches and the path found first (left to right) is the one that is used, so uuids are used for partitions and disk names are used for whole disks.

dhiltonp:
With the current version, I do get several "Error getting pool data" dialogs appearing over the screen; I don't know what they mean. I searched for the text and was unable to find it.

image

Also note that `tank` is listed multiple times. It's listed 4 times. The pool has 2 disks, and the error dialog appears 8 times.

On subsequent refreshes the pool only shows up once.

william-gr:
You probably get those because whole disk ZFS pools are not supported by our middleware.

It is very likely many operations would fail, including replacing a disk because the lack of space.

dhiltonp:
It's clear that importing pools with vdevs that use whole disks is not currently supported. That's an excellent thing to discuss.

The previous behavior is wrong - just getting a stack trace with no explanation.

At a minimum, the pool could be shown in the import list, but the import could fail with an error like `Cannot import pools with vdevs that use whole disks; vdev devices must be partitions.` 

This seems wrong; only a fraction of what TrueNAS provides is vdev management. If nothing else, the user can manage the vdev via the command line, using TrueNAS for other operations.

That's what this PR allows.

Another option beyond what I have in this PR is to allow the import, but warn the user `Disk management for vdevs that use whole disks instead of partitions is not fully supported.`

I don't know what operations don't work. I suspect that all operations would work, so long as the new device is larger than the old (to make room for the overhead of the partition table).

william-gr:
This sounds like a bigger feature request than just allowing a pool to be imported.

There are many assumptions that are made in middleware about how the pool is set up.
If we allow pools like that to be imported than we are automatically in the hook for fixing all the issues it will imply along the way.

dhiltonp:
If that's true, then the only importable ZFS pools would be ones created by TrueNAS.

That seems short-sighted.

It's ok to have soft boundaries where you say "we don't support this, but we accept contributions." 

In this case "we do our testing on pools created by TrueNAS. You can import other pools but you may run in to issues with volume management. PRs improving TrueNAS are welcome :)"

william-gr:
Sorry, where did I say anything different?

dhiltonp:
I interpreted "If we allow pools like that to be imported than we are automatically in the hook for fixing all the issues it will imply along the way" as you giving a reason to not accept the PR.

william-gr:
No. I just meant there may be implications and before accepting the PR we need to make sure at least some basic functionalities are working. Its no good if we change this to allow importing that kind of pool and a lot of other places are failing (based on your screenshot that may be the case).

sonicaj:
Closing this PR as changes introduced here will be catered for in https://github.com/truenas/middleware/pull/6856.

PR #6282, opened 2021-01-15
NAS-108991 / 21.02 / Add ability to update container images in a chart release

sonicaj:


src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
```diff
@@ -33,10 +33,16 @@ async def query(self, filters=None, options=None):
         `options.extra.history` is a boolean when set will retrieve all chart version upgrades for a chart release.
         """
         if not await self.middleware.call('service.started', 'kubernetes'):
-            return []
+            return filter_list([], filters, options)
```
william-gr, line 36: Why?

sonicaj: Basically `get: true` fails where consumer would expect an error to be raised if there is no chart release but is instead presented with an empty list

william-gr: May I suggest a comment? Anyone looking at the code would probably wonder why we are filtering an empty list

sonicaj: Of course, adding it right away

sonicaj: Added :)

william-gr (approved):

PR #6287, opened 2021-01-17
NAS-109008 / 21.02 / Reserve library name in catalogs

sonicaj:
This commit introdues a change to not consider 'library' as a catalog train and this folder then instead can be used to place helm library charts.

Closes truenas/charts#25


themylogin (approved):
What's wrong with good old `!=`?

PR #6289, opened 2021-01-18
NAS-108333 / 21.02 / add initial ctdb clustering plugin (phase 1)

yocalebo:


src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,241 @@
...
+PRI_LOCK = CTDBConfig.PRI_LOCK.value
+
+
+class CtdbIpService(Service):
+
+    class Config:
+        namespace = 'ctdb.private.ips'
+
+    @private
+    def query(self):
```
william-gr, line 21: It should be nice, even if being private, that it could accept the same arguments as every .query method we have. Or rename this method so its not confusing

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,241 @@
...
+        #       exist then assume this is the first peer being
+        #       added to the cluster and skip the cluster health
+        #       check.
+        ctdb_file = pathlib.Path(CTDBConfig.GM_PRI_IP_FILE.value)
+        etc_file = pathlib.Path(CTDBConfig.ETC_PRI_IP_FILE.value)
+
+        if ctdb_file.exists():
+            if self.middleware.call_sync('service.started', 'ctdb'):
+                if not self.middleware.call_sync('ctdb.general.healthy'):
+                    self.logger.error('ctdb cluster is not healthy')
```
william-gr, line 131: might be worth adding a note on what it implies: `ctdb cluster is not healthy, not updating private ip files`

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,241 @@
...
+                    return False
+
+        # make sure the private ip file exists
+        try:
+            ctdb_file.touch(exist_ok=True)
+        except Exception:
+            self.logger.error('Failed creating %s', str(ctdb_file), exc_info=True)
```
william-gr, line 138: str() is not necessary, %s already does the job

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,241 @@
...
+                return False
+
+        # ctdb documentation is _VERY_ explicit about
+        # how this file should be modified
+        if not delete:
+            # in the case of adding a node, it _MUST_ be
+            # added to the end of the file always
+            with open(PRI_IP_FILE, 'a') as f:
+                f.write(pri_ip + '\n')
+                f.flush()
```
william-gr, line 177: flush() is probably not necessary, since a close() from the context manager should imply it

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,241 @@
...
+            # the file and the other lines _MUST_ be
+            # kept in the same order when writing it
+            # back out or ctdb will barf and fail to
+            # reload...
+            index = lines.index(pri_ip)
+            lines[index] = '#' + pri_ip
+
+            with open(PRI_IP_FILE, 'w') as f:
+                f.writelines(map(lambda x: x + '\n', lines))
+                f.flush()
```
william-gr, line 193: Same about flush

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,241 @@
...
+import pathlib
+
+
+SHARED_VOL = CTDBConfig.CTDB_LOCAL_MOUNT.value
+PRI_IP_FILE = CTDBConfig.GM_PRI_IP_FILE.value
+ETC_IP_FILE = CTDBConfig.ETC_PRI_IP_FILE.value
+PRI_LOCK = CTDBConfig.PRI_LOCK.value
+
+
+class CtdbIpService(Service):
```
william-gr, line 15: I would probably make this a CRUDService

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,241 @@
...
+            raise CallError(f'Failed updating {PRI_IP_FILE}. Please check logs.')
+
+        return data['ip']
+
+    @accepts(Dict(
+        'node_delete',
+        IPAddr('ip'),
+    ))
+    @job(lock=PRI_LOCK)
+    async def delete(self, job, data):
```
william-gr, line 225: If this is a Crid, this would simply be: `do_delete(self, job, id):`

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,241 @@
...
+        `ip` is an IP v4/v6 address
+        """
+
+        schema_name = 'node_delete'
+        verrors = ValidationErrors()
+
+        await self.middleware.call(
+            'ctdb.private.ips.common_validation', data['ip'], schema_name, verrors, True
+        )
+        if not await self.middleware.call('ctdb.private.ips.update_file', data['ip'], verrors, True):
```
william-gr, line 238: Wouldn't it be better to make ctdb.private.ips.update_file raise an error show we could potentially know the error instead of telling user to check logs?

src/middlewared/middlewared/plugins/cluster_linux/ctdb_public.py
```diff
@@ -0,0 +1,246 @@
...
+import pathlib
+
+
+SHARED_VOL = CTDBConfig.CTDB_LOCAL_MOUNT.value
+PUB_IP_FILE = CTDBConfig.GM_PUB_IP_FILE.value
+ETC_IP_FILE = CTDBConfig.ETC_PUB_IP_FILE.value
+PUB_LOCK = CTDBConfig.PUB_LOCK.value
+
+
+class CtdbIpService(Service):
```
william-gr, line 15: Would probably make it a CRUDService too

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
yocalebo: I didn't make it a CRUDService because I'm not writing anything to the db?

william-gr: Its not about writing to the db or not, but consistent API.

Also, are these really supposed to be both public APIs?

yocalebo: Okay well I'm being told 2 different things then. I've been told not to make it CRUD if it doesn't touch our dB. That's the entire approach I took when writing the gluster api

william-gr: Was I the one who told you that? I think there was some miscommunication. CRUD was never related to DB.

yocalebo: The questions I have now are:

1. When do I not use CRUDService?
2. Do you want me to back and fix gluster plugin?

william-gr: You do not use CRUDService when it does not make sense to have a query method, nor a create/delete method.

CRUDService is for when you're operating over a single entity:

CRUD - Create/Retrieve/Update/Delete
Create - A single entity (e.g. an ip, an interface, a user)
Retrieve - A collection (a list of objects)
Update - A single entity (e.g. in the interface X, set attribute Y to Z)
Delete - A single entity (e.g. delete an interface X)

william-gr: About fixing gluster, if that would make things consistent, I am +1 for that, as that API is not considered stable yet

yocalebo: Okay thanks, I'll take care of it.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,241 @@
...
+                mounted = False
+
+            if mounted:
+                pri_ip_file = pathlib.Path(PRI_IP_FILE)
+                etc_ip_file = pathlib.Path(ETC_IP_FILE)
+                if pri_ip_file.exists():
+                    if etc_ip_file.is_symlink() and etc_ip_file.resolve() == pri_ip_file:
+                        with open(PRI_IP_FILE) as f:
+                            ips = f.read().splitlines()
+        return ips
```
anodos325, line 54: To expand on William's comment, you can include `filter_list` from middlewared.utils and then do something like:
```
return filter_list(ips, filters, options)
```

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,232 @@
...
+                etc_ip_file = pathlib.Path(ETC_IP_FILE)
+                if pri_ip_file.exists():
+                    if etc_ip_file.is_symlink() and etc_ip_file.resolve() == pri_ip_file:
+                        with open(PRI_IP_FILE) as f:
+                            for i in f.read().splitlines():
+                                ips.append({
+                                    'id': i,
+                                    'address': i,
+                                })
+
```
anodos325, line 60: What is layout of expected dict here? `ctdb listnodes -j` should return something like this for the nodelist:
`[ { "pnn": 0, "address": "<ip address as string>"} ]` (no`id` key). If we read it from the nodes file, we add an `id` key. These should probably be consistent. I'll look up logic for how we determine `pnn` in ctdb. It might be based on order of list (skipping disabled/commented-out entries).

yocalebo: I add an `id` key to what the `ctdb listnodes -j` returns as well as what we read from the private IP file. This is to make `get_instance` work in the `do_delete` method.

anodos325:
```
/* Read a nodes file into a node map */
static struct ctdb_node_map *ctdb_read_nodes_file(TALLOC_CTX *mem_ctx,
                                                  const char *nlist)
{
        char **lines;
        int nlines;
        int i;
        struct ctdb_node_map *nodemap;

        nodemap = talloc_zero(mem_ctx, struct ctdb_node_map);
        if (nodemap == NULL) {
                return NULL;
        }

        lines = file_lines_load(nlist, &nlines, 0, mem_ctx);
        if (lines == NULL) {
                return NULL;
        }

        while (nlines > 0 && strcmp(lines[nlines-1], "") == 0) {
                nlines--;
        }

        for (i=0; i<nlines; i++) {
                char *node;
                uint32_t flags;
                size_t len;

                node = lines[i];
                /* strip leading spaces */
                while((*node == ' ') || (*node == '\t')) {
                        node++;
                }

                len = strlen(node);

                /* strip trailing spaces */
                while ((len > 1) &&
                       ((node[len-1] == ' ') || (node[len-1] == '\t')))
                {
                        node[len-1] = '\0';
                        len--;
                }

                if (len == 0) {
                        continue;
                }
                if (*node == '#') {
                        /* A "deleted" node is a node that is
                           commented out in the nodes file.  This is
                           used instead of removing a line, which
                           would cause subsequent nodes to change
                           their PNN. */
                        flags = NODE_FLAGS_DELETED;
                        node = discard_const("0.0.0.0");
                } else {
                        flags = 0;
                }
                if (! node_map_add(nodemap, node, flags)) {
                        talloc_free(lines);
                        TALLOC_FREE(nodemap);
                        return NULL;
                }
        }

        talloc_free(lines);
        return nodemap;
}
```
That's the logic for generating the `nodes` mapping in ctdb. It reads the `nodes` file and `pnn` is based on order in file. Disabled nodes (ones beginning with a `#` character) are present to preserve pnn value, but ip address set to 0.0.0.0 and a flag set NODE_FLAGS_DELETED (internally in ctdb).

anodos325: TL;DR. `pnn` and `id` in this case should end up being the same. Rename either the one or the other (maybe easier to use `pnn` since that's referenced in other places in ctdb output.

anodos325: Do you think it's worthwhile to add an `enabled` key to this and make it correspond to the ip string starting with `#`?

anodos325: Scratch that. I just realized that `control_listnodes()` skips entries with NODE_FLAGS_DELETED.

yocalebo: Adding an `enabled` key to `ctdb listnodes -j` output would be beneficial.

anodos325: For that matter, I can add deactivated nodes to the control_listnodes() JSON output and add a boolean key `enabled` to differentiate the two if you want to add an extra key to this. Not sure if there's a benefit to allowing users to deactivate entries in the nodes file in this way.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,232 @@
...
+        """
+
+        schema_name = 'node_delete'
+        verrors = ValidationErrors()
+
+        address = (await self.get_instance(id))['id']
+        await self.middleware.call('ctdb.private.ips.common_validation', address, schema_name, verrors, True)
+        await self.middleware.call('ctdb.private.ips.update_file', address, verrors, True)
+
+        return address
```
anodos325, line 232: For consistency with CRUD API, we will probably also want an `update()` method that accepts an `id` / `pnn`. This can be a wrapper around your existing update_file() method.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_public.py
```diff
@@ -0,0 +1,238 @@
...
+                if pub_ip_file.exists():
+                    if etc_ip_file.is_symlink() and etc_ip_file.resolve() == pub_ip_file:
+                        with open(PUB_IP_FILE) as f:
+                            lines = f.read().splitlines()
+                            for i in lines:
+                                ips.append({
+                                    'id': i.split('/')[0],
+                                    'public_ip': i.split('/')[0],
+                                    'netmask': int(i.split('/')[1].split()[0]),
+                                    'interfaces': list(i.split()[-1]),
```
anodos325, line 59: `ctdb -j ip` output entries here (for `interfaces`) are of form: `{"name": <string>, "active": <bool>, "available": <bool>"}`.

So we'll want to mirror that structure when we read from files. Since ctdb isn't started, we can assume `active` is probably `False`. I'd have to look closer about `available`, which is based on ctdb_iface->link_state value. Most likely though hard-coding `False` is acceptable if ctdb isn't running.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_public.py
```diff
@@ -0,0 +1,238 @@
...
+            if mounted:
+                pub_ip_file = pathlib.Path(PUB_IP_FILE)
+                etc_ip_file = pathlib.Path(ETC_IP_FILE)
+                if pub_ip_file.exists():
+                    if etc_ip_file.is_symlink() and etc_ip_file.resolve() == pub_ip_file:
+                        with open(PUB_IP_FILE) as f:
+                            lines = f.read().splitlines()
+                            for i in lines:
+                                ips.append({
+                                    'id': i.split('/')[0],
```
anodos325, line 56: `id` from `ctdb -j ip` is sorted based on mildly complex rules specified in ctdb/tools/ctdb.c and ctdb/protocol/protocol_util.c `ctdb_sock_addr_cmp()`. Not sure if consistency in id allocation is important here (probably not).

yocalebo: public IPs are not strictly handled like the private IPs are in ctdb. However, the `pnn` output from `ctdb ip -j` still seems to mirror the position in which the entry is written in the file.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_public.py
```diff
@@ -0,0 +1,238 @@
...
+        `interface` a network interface to have assigned the `ip`
+        """
+
+        schema_name = 'node_create'
+        verrors = ValidationErrors()
+
+        await self.middleware.call('ctdb.public.ips.common_validation', data, schema_name, verrors)
+        await self.middleware.call('ctdb.public.ips.update_file', data, verrors)
+
+        return data
```
anodos325, line 220: May need to add the `id` to the returned dictionary. Will probably have to go through self.query() to ensure that id is consistent.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_public.py
```diff
@@ -0,0 +1,238 @@
...
+            if data['ip'] not in cur_ips:
+                verrors.add(
+                    f'{schema_name}.{data["ip"]}',
+                    f'"{data["ip"]}" does not exist in the cluster.'
+                )
+
+        verrors.check()
+
+    @private
+    def update_file(self, data, verrors, delete=False):
```
anodos325, line 145: Not sure if this is important, may want to consult with William, but we may need an update() method that wraps around this.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
yocalebo: I've changed `id` to be an alias for `pnn` in the `query()` methods. I've also just defaulted to returning `active = False` and `available = False` if we're  not able to query the ctdb daemon directly.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_public.py
yocalebo: `do_update` methods don't make sense in this context.

For example. `update()` in the private IP API would be change `#192.168.1.150` to `192.168.1.150` so it wouldn't be clean to do it that way since I'd have to by pass `query()` and read directly from the file.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
yocalebo: Spoke in DM and you're going to add a new `enabled` keyword to the `listnodes` ctdb command so that implementing a `do_update()` method makes a little more sense.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,107 @@
...
+        schema_name = 'private_update'
+        verrors = ValidationErrors()
+
+        data = await self.get_instance(id)
+        data['enable'] = option['enable']
+
+        await self.middleware.call('ctdb.ips.common_validation', data, schema_name, verrors)
+        await self.middleware.call('ctdb.ips.update_file', data, schema_name)
+
+        return id
```
william-gr, line 107: do_update is suppose to return the new instance, get_instance(id)

src/middlewared/middlewared/plugins/cluster_linux/ctdb_private.py
```diff
@@ -0,0 +1,107 @@
...
+        `ip` string representing an IP v4/v6 address
+        """
+
+        schema_name = 'private_create'
+        verrors = ValidationErrors()
+
+        await self.middleware.call('ctdb.ips.common_validation', data, schema_name, verrors)
+        await self.middleware.call('ctdb.ips.update_file', data, schema_name)
+
+        return data['ip']
```
william-gr, line 80: do_create is suppose to return the new instance, get_instance(id)

src/middlewared/middlewared/plugins/cluster_linux/ctdb_public.py
```diff
@@ -0,0 +1,125 @@
...
+        schema_name = 'public_update'
+        verrors = ValidationErrors()
+
+        data = await self.get_instance(id)
+        data['enable'] = option['enable']
+
+        await self.middleware.call('ctdb.ips.common_validation', data, schema_name, verrors)
+        await self.middleware.call('ctdb.ips.update_file', data, schema_name)
+
+        return id
```
william-gr, line 125: do_update is suppose to return the new instance, get_instance(id)

src/middlewared/middlewared/plugins/cluster_linux/ctdb_ips.py
```diff
@@ -0,0 +1,193 @@
...
+                address = data['address']
+            else:
+                address = data["public_ip"]
+
+            find_entry = address if not enable else '#' + address
+
+            # no matter what we're doing at this point,
+            # we're going to have to read, truncate,
+            # and rewrite the contents.
+            with open(ctdb_file, 'r+') as f:
```
william-gr, line 161: Didnt we had a python issue where it couldnt open a file for read and write this way? and we had to use os.open()?

yocalebo: The `write_if_changed` method has this problem but only because the `etc` plugin doesn't guarantee the file is created when you call it.

At the beginning of this method, we always `file.touch(exist_ok=True)` before we get here. So this is the way of opening the file for reading and writing and not truncating if it exists (which it should always).

william-gr: Ah, makes sense, thanks!

william-gr (approved):

anodos325 (approved):

PR #6291, opened 2021-01-20
NAS-109064 / 21.02 / Allow to not retrieve children of a dataset

sonicaj:
UI right now needs top level dataset space usage so that it can show that information on the dashboard. However if we have around 700 datasets, UI can take around 10 seconds to just load the dashboard. Mostly this is UI being blocked by middleware call where it tries to retrieve information for these root datasets, doing a comparison of zfs cli and py-libzfs, py-libzfs is actually faster in this case - so the main lag comes from middleware retrieving all the children information.

This PR introduces a change where this can be configurable and for such cases, consumer can choose not to get children's data.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2749,6 +2749,10 @@ class PoolDatasetService(CRUDService):
         The second type is hierarchical, where only top level datasets are returned in the list. They contain all the
         children in the `children` key. This retrieval type is slightly faster.
         These options are controlled by the `query-options.extra.flat` attribute (default true).
+
+        In some cases it might be desirable to only retrieve details of a dataset itself and not it's children, in this
+        case `query-options.extra.retrieve_children` should be explicitly specified and set to `false` which will
+        result in children not being retrieved.
         """
         # Optimization for cases in which they can be filtered at zfs.dataset.query
         zfsfilters = []
@@ -2774,13 +2778,18 @@ class PoolDatasetService(CRUDService):
             if k8s_config['dataset']:
                 filters.append(['id', '!^', f'{k8s_config["dataset"]}/'])
 
+        extra = copy.deepcopy(options.get('extra', {}))
+        retrieve_children = extra.get('retrieve_children', True)
         return filter_list(
             self.__transform(self.middleware.call_sync(
-                'zfs.dataset.query', zfsfilters, {'extra': {'flat': options.get('extra', {}).get('flat', True)}})
+                'zfs.dataset.query', zfsfilters, {
+                    'extra': {'flat': extra.get('flat', True), 'retrieve_children': retrieve_children}
+                }
+            ), retrieve_children,
             ), filters, options
         )
 
-    def __transform(self, datasets):
+    def __transform(self, datasets, retrieve_children):
         """
         We need to transform the data zfs gives us to make it consistent/user-friendly,
         making it match whatever pool.dataset.{create,update} uses as input.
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -455,6 +455,10 @@ def query(self, filters=None, options=None):
         `query-options.extra.snapshots` is a boolean which when set will retrieve snapshots for the dataset in question
         by adding a snapshots key to the dataset data.
 
+        `query-options.extra.retrieve_children` is a boolean which when set which is the default will retrieve child
+        datasets of a dataset. Motivation for providing a knob to configure this is performance concerns where getting
+        children sometimes is not required and in that case this user can be explicitly unset.
+
```
yocalebo, line 461: Just so I'm understanding correctly, this will not retrieve grandchildren datasets properties? `/mnt/tank/parent/child/grandchild`?

sonicaj: It will not traverse children at all which means yes it won't retrieve properties for them as well as the traversal would never happen and we remove `children` key altogether to not give the impression that this dataset does not has children when it might have.

yocalebo: So it will only retrieve top-level datasets? `/mnt/tank/a` and `/mnt/tank/b` and `/mnt/tank/b` etc?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2774,13 +2778,18 @@ def query(self, filters=None, options=None):
             if k8s_config['dataset']:
                 filters.append(['id', '!^', f'{k8s_config["dataset"]}/'])
 
+        extra = copy.deepcopy(options.get('extra', {}))
+        retrieve_children = extra.get('retrieve_children', True)
         return filter_list(
             self.__transform(self.middleware.call_sync(
-                'zfs.dataset.query', zfsfilters, {'extra': {'flat': options.get('extra', {}).get('flat', True)}})
+                'zfs.dataset.query', zfsfilters, {
+                    'extra': {'flat': extra.get('flat', True), 'retrieve_children': retrieve_children}
```
yocalebo, line 2786: To be consistent, why not create another variable called `flat = extra.get('flat', True)` like you did with retrieve_children?

sonicaj: I can do that but that can grow :) Motivation was to not have an extra line due to line being too long ( 120 > ). If you think i should, i can :p

yocalebo: It's me just being OCD mostly, but I like the consistency of pulling out the extra arguments in their own variables. Looks cleaner, but obviously not needed. I'll leave it up to you. 😄

src/middlewared/middlewared/plugins/zfs.py
yocalebo: If it only retrieves the top-level datasets then may I suggest a better doc-string?

`query-options.extra.retrieve_children is a boolean. Set to True by default. When set to True, will retrieve all child datasets which can cause a performance penalty. When set to False, will explicitly ignore all child datasets and only return the top-level datasets.`

sonicaj: Sure, so the bit about `top-level datasets` is not accurate as it retrieves the datasets passed and that can be `pool/someds/anotherchild` which is not a top level dataset. Can you please advise on that ? Thank you

yocalebo: Sure,

`query-options.extra.retrieve_children is a boolean. Set to True by default. When set to True, will retrieve all children datasets which can cause a performance penalty. When set to False, will not retrieve children datasets which does not incur the performance penalty.`

yocalebo (approved):

PR #6293, opened 2021-01-20
NAS-108813 / 21.02 / Fake HyperThreading CPU thread temperatures in Linux

themylogin:


themylogin:
@william-gr ping

william-gr:
@themylogin hadnt we discussed faking was bad and didnt make sense?

william-gr:
Also not every CPU has core count uniform. I remember seeing some CPUs where Virtual Core 1 and 4 and for real core #1

themylogin:
> @themylogin hadnt we discussed faking was bad and didnt make sense?

@william-gr no, we didn't, the only discussion was here https://jira.ixsystems.com/browse/NAS-108813

What should we do then?

> Also not every CPU has core count uniform. I remember seeing some CPUs where Virtual Core 1 and 4 and for real core #1

This will only fake for most common HyperThreading CPUs with 2 threads per core

william-gr:
> This will only fake for most common HyperThreading CPUs with 2 threads per core

I dont think you can fake it like that, there is no guarantee Virtual Core 1 and Virtual Core 2 are consecutive and belong to the same Real Core.

william-gr:
> @william-gr no, we didn't, the only discussion was here https://jira.ixsystems.com/browse/NAS-108813
> 
> What should we do then?

You said its only reported per core, and I replied that we have no choice. I think it means we cannot fake anything and keep it real.

themylogin:
@william-gr we should probably ask UI team to redesign that widget for SCALE?

william-gr:
@themylogin Thats the only thing I can think of

PR #6295, opened 2021-01-20
NAS-108519 / 21.02 / Use SQLite AUTOINCREMENT to make integer primary keys unique

themylogin:


william-gr (approved):
This sounds risky but I have no better answer, we will need a lot of testing.

src/middlewared/middlewared/alembic/versions/12.1/2021-01-20_10-19_autoincrement.py
```diff
@@ -0,0 +1,65 @@
...
+down_revision = 'c68c71c34771'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    conn = op.get_bind()
+    conn.execute("PRAGMA legacy_alter_table = TRUE")
+    for name, sql in conn.execute("SELECT name, sql FROM sqlite_master WHERE type = 'table'").fetchall():
+        if m := re.match(r'CREATE TABLE "(.+)" \(\s*"?id"? integer (NOT NULL |)PRIMARY KEY,', sql, flags=re.IGNORECASE):
```
william-gr, line 24: Wonder if there is any benefit in changing all these regex to be non-greedy (`.+?`)

themylogin: This is ran against a fixed set of strings every time and it already works, so I think there is not :)

PR #6298, opened 2021-01-21
NAS-109033 / 21.02 / Bug fix for correctly retrieving DN value from a cert

sonicaj:
There are cases where simple decode can fail when trying to read cert attributes. We should rely on openssl module instead to give us normalized values instead of trying to decode ourselves.


yocalebo (approved):

sonicaj:
backport

PR #6299, opened 2021-01-21
NAS-109047 / 21.02 / use iptables to allow/disallow network traffic on failover event for SCALE HA

yocalebo:
- I'm writing to a file to prevent unnecessary `subprocess` calls on SCALE HA systems that have a large amount of VIPs.
- SCALE HA uses VRRP so the VIPs **ONLY** exist on the MASTER controller. However, I'll still block the VIP traffic on the BACKUP node to help speed the process up by disallowing any NEW connections that come in while any of the remaining VIPs exist before the event fully completes.

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
@@ -491,11 +491,10 @@ class FailoverService(Service):
                     self.run_call('service.restart', i, self.HA_PROPAGATE)
 
         logger.info('Allowing network traffic.')
-        try:
-            self.run_call('failover.firewall.accept_all')
-        except Exception:
-            # don't crash here
-            logger.error('Failed to allow network traffic', exc_info=True)
+        fw_accept_job = self.run_call('failover.firewall.accept_all')
+        fw_accept_job.wait_sync()
+        if fw_accept_job.error:
+            logger.error(f'Error allowing network traffic: {fw_accept_job.error}')
 
         logger.info('Critical portion of failover is now complete')
 
@@ -577,18 +576,17 @@ class FailoverService(Service):
         logger.warning('Stopping fenced')
         self.run_call('failover.fenced.stop')
 
+        logger.info('Blocking network traffic.')
+        fw_drop_job = self.run_call('failover.firewall.drop_all')
+        fw_drop_job.wait_sync()
+        if fw_drop_job.error:
+            logger.error(f'Error allowing network traffic: {fw_drop_job.error}')
+
         # restarting keepalived sends a priority 0 advertisement
         # which means any VIP that is on this controller will be
         # migrated to the other controller
         logger.info('Transitioning all VIPs off this node')
         self.run_call('service.restart', 'keepalived')
-
-        logger.info('Blocking network traffic')
-        try:
-            self.run_call('failover.firewall.drop_all')
-        except Exception:
-            # don't crash here
-            logger.error('Failed to block network traffic', exc_info=True)
 
         # ticket 23361 enabled a feature to send email alerts when an unclean reboot occurrs.
         # TrueNAS HA, by design, has a triggered unclean shutdown.
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
@@ -490,9 +490,12 @@ class FailoverService(Service):
                     logger.info('Restarting critical service "%s"', i)
                     self.run_call('service.restart', i, self.HA_PROPAGATE)
 
-        # TODO: look at nftables
-        # logger.info('Allowing network traffic.')
-        # run('/sbin/pfctl -d')
+        logger.info('Allowing network traffic.')
+        try:
+            self.run_call('failover.firewall.accept_all')
+        except Exception:
+            # don't crash here
+            logger.error('Failed to allow network traffic', exc_info=True)
 
         logger.info('Critical portion of failover is now complete')
 
@@ -580,9 +583,12 @@ class FailoverService(Service):
         logger.info('Transitioning all VIPs off this node')
         self.run_call('service.restart', 'keepalived')
 
-        # TODO: look at nftables
-        # logger.info('Enabling firewall')
-        # run('/sbin/pfctl -ef /etc/pf.conf.block')
+        logger.info('Blocking network traffic')
+        try:
+            self.run_call('failover.firewall.drop_all')
+        except Exception:
+            # don't crash here
+            logger.error('Failed to block network traffic', exc_info=True)
 
         # ticket 23361 enabled a feature to send email alerts when an unclean reboot occurrs.
         # TrueNAS HA, by design, has a triggered unclean shutdown.
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/iptables_linux.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/iptables_linux.py
@@ -1,7 +1,7 @@
-from middlewared.service import Service, accepts, job, CallError
-
 import subprocess
 import copy
+
+from middlewared.service import Service, accepts, job, CallError
 
 
 V4_FILE = '/tmp/v4-fw.rules'
@@ -20,13 +20,14 @@ class IptablesService(Service):
         Generate a list of default firewall rules.
         """
 
-        # this is always the first rule
-        rules = ['*filter']
-
-        # the positions of these are important
-        rules.insert(1, ':INPUT ACCEPT [0:0]')
-        rules.insert(2, ':FORWARD ACCEPT [0:0]')
-        rules.insert(3, ':OUTPUT ACCEPT [0:0]')
+        # this is always the beginning of the rules
+        # the order of these are important
+        rules = [
+            '*filter',
+            ':INPUT ACCEPT [0:0]',
+            ':FORWARD ACCEPT [0:0]',
+            ':OUTPUT ACCEPT [0:0]',
+        ]
 
         if data['drop']:
             # we always allow ssh and webUI access when limiting inbound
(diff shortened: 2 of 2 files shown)


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/iptables_linux.py
```diff
@@ -0,0 +1,159 @@
...
+        # the positions of these are important
+        # v4 default section
+        v4rules.insert(1, ':INPUT ACCEPT [0:0]\n')
+        v4rules.insert(2, ':FORWARD ACCEPT [0:0]\n')
+        v4rules.insert(3, ':OUTPUT ACCEPT [0:0]\n')
+
+        # v6 default section
+        v6rules.insert(1, ':INPUT ACCEPT [0:0]\n')
+        v6rules.insert(2, ':FORWARD ACCEPT [0:0]\n')
+        v6rules.insert(3, ':OUTPUT ACCEPT [0:0]\n')
```
william-gr, line 38: Would it make sense/clearer to do not have \n to every entry and join them at the end with \n?

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/iptables_linux.py
```diff
@@ -0,0 +1,159 @@
...
+
+            # v4 ssh/webUI rules
+            v4rules.append(f'-A INPUT -p tcp -m tcp --dport {sshport} -j ACCEPT\n')
+            v4rules.append(f'-A INPUT -p tcp -m tcp --dport {web["ui_port"]} -j ACCEPT\n')
+            v4rules.append(f'-A INPUT -p tcp -m tcp --dport {web["ui_httpsport"]} -j ACCEPT\n')
+
+            # v6 ssh/webUI rules
+            v6rules.append(f'-A INPUT -p tcp -m tcp --dport {sshport} -j ACCEPT\n')
+            v6rules.append(f'-A INPUT -p tcp -m tcp --dport {web["ui_port"]} -j ACCEPT\n')
+            v6rules.append(f'-A INPUT -p tcp -m tcp --dport {web["ui_httpsport"]} -j ACCEPT\n')
```
william-gr, line 54: Up to this point v4 and v6 rules are exactly the same, perhaps we can come up with a way to avoid duplicating that code?

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/iptables_linux.py
```diff
@@ -0,0 +1,159 @@
...
+        except Exception as e:
+            raise CallError(f'Failed writing {V4_FILE} with error {e}')
+
+        try:
+            with open(V6_FILE, 'w+') as f:
+                f.writelines(v6rules)
+        except Exception as e:
+            raise CallError(f'Failed writing {V6_FILE} with error {e}')
+
+    def restore_files(self, v4rules, v6rules):
```
william-gr, line 89: You dont need v4rules, v6rules args here it seems

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
```diff
@@ -490,9 +490,12 @@ def vrrp_master(self, job, fobj, ifname, event):
                     logger.info('Restarting critical service "%s"', i)
                     self.run_call('service.restart', i, self.HA_PROPAGATE)
 
-        # TODO: look at nftables
-        # logger.info('Allowing network traffic.')
-        # run('/sbin/pfctl -d')
+        logger.info('Allowing network traffic.')
+        try:
+            self.run_call('failover.firewall.accept_all')
```
sonicaj, line 495: `run_call` already logs exceptions. We don't need another try/catch.

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
```diff
@@ -580,9 +583,12 @@ def vrrp_backup(self, job, fobj, ifname, event):
         logger.info('Transitioning all VIPs off this node')
         self.run_call('service.restart', 'keepalived')
 
-        # TODO: look at nftables
-        # logger.info('Enabling firewall')
-        # run('/sbin/pfctl -ef /etc/pf.conf.block')
+        logger.info('Blocking network traffic')
+        try:
+            self.run_call('failover.firewall.drop_all')
```
sonicaj, line 588: Same as above

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
sonicaj: `accept_all` is a job, so we should wait for it to complete here ?

sonicaj: Also do we want it to be a job ? I think it should return before 60 secs ?

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/iptables_linux.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def write_files(self, v4rules, v6rules):
+        """
+        Write the firewall rules to the appropriate file(s).
+        """
+
+        try:
+            with open(V4_FILE, 'w+') as f:
+                # there must be a trailing newline in the file
+                f.write(''.join(f"{i}\n" for i in v4rules))
```
william-gr, line 76: `'\n'.join(v4rules) + '\n'`

thats simpler and less time/cpu consuming

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/iptables_linux.py
```diff
@@ -0,0 +1,157 @@
...
+    async def generate_default_rules(self, data):
+        """
+        Generate a list of default firewall rules.
+        """
+
+        # this is always the first rule
+        rules = ['*filter']
+
+        # the positions of these are important
+        rules.insert(1, ':INPUT ACCEPT [0:0]')
```
sonicaj, line 27: Why do we need to use `insert` here and not just declare it at `L24` ?

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/iptables_linux.py
```diff
@@ -0,0 +1,157 @@
+from middlewared.service import Service, accepts, job, CallError
+
+import subprocess
+import copy
+
+
+V4_FILE = '/data/v4-fw.rules'
```
william-gr, line 7: Just noticed, does it need to be in /data, why not /tmp?

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/iptables_linux.py
```diff
@@ -0,0 +1,157 @@
...
+
+        try:
+            with open(V6_FILE, 'w+') as f:
+                # there must be a trailing newline in the file
+                f.write('\n'.join(v6rules) + '\n')
+        except Exception as e:
+            raise CallError(f'Failed writing {V6_FILE} with error {e}')
+
+    def restore_files(self):
+
```
sonicaj, line 88: extra line

william-gr (approved):
Thanks!

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/iptables_linux.py
yocalebo: I used insert to be explicit about their positions in the list.

sonicaj: We can just add a comment for that perhaps ? In my opinion, it just looks better that way as there is no dynamic change happening here otherwise.

sonicaj: Also iirc `ACCEPT` is a terminating target, meaning that we having this at the start, will have all the packets not move forward to the other rules. Please correct me otherwise ?

yocalebo: `ACCEPT` is the default behavior for the `INPUT` chain. It means if no rules match in the `INPUT` chain, allow the traffic.

It's the path of least resistance since changing the default behavior for the `INPUT` chain to `DROP` would be system wide and would effect everything running service on the system.

sonicaj: I agree on that, i am concerned about the order. Having that rule as the first one means that every packet is going to be accepted and rest of the rules in the chain are not going to be evaluated. If i am correct, it means we should have this at the end so that other rules are processed first.

yocalebo: No that line in the list is for the default `INPUT` chain not a rule. Rules only get generated if `drop_all` is run.

So the rules look like this:
`-P INPUT ACCEPT
-P FORWARD ACCEPT
-P OUTPUT ACCEPT
-A INPUT -p tcp -m tcp --dport 22 -j ACCEPT
-A INPUT -p tcp -m tcp --dport 80 -j ACCEPT
-A INPUT -p tcp -m tcp --dport 443 -j ACCEPT
-A INPUT -s 192.168.1.184/32 -j DROP
-A INPUT -s 2.2.2.2/32 -j DROP
-A INPUT -s 10.10.10.10/32 -j DROP`

These rules mean allow any traffic to dst ports 22, 80, and 443 no matter what. If the traffic doesn't match those rules, then go to next rule which means drop any traffic to the IP of `192.168.1.184` (which is a VIP address on HA systems)

yocalebo: Finally if the traffic doesn't match any of the `-A INPUT` rules, then allow the traffic since `-P INPUT ACCEPT`

sonicaj (approved):
Thanks

PR #6303, opened 2021-01-22
NAS-109081 / 21.02 / Make retrieving single chart release faster

sonicaj:
This commit introduces changes to make retrieving single chart release considerably faster. With these changes we get chart release details in about 0.7 seconds and without this we get it in around 2 seconds for about 25 chart releases. And when chart release chart schema is required, it is 0.85 secs and the latter around 4.8 seconds.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
@@ -32,8 +32,8 @@ class ChartReleaseService(CRUDService):
 
         `options.extra.history` is a boolean when set will retrieve all chart version upgrades for a chart release.
 
-        `options.extra.include_chart_schema` is a boolean when set will retrieve the schema being used by the chart release
-        in question.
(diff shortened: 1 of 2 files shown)


yocalebo (approved):

src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
```diff
@@ -67,9 +67,15 @@ async def query(self, filters=None, options=None):
         get_resources = extra.get('retrieve_resources')
         get_history = extra.get('history')
 
+        if filters and len(filters) == 1 and filters[0][:2] == ['id', '=']:
+            extra['namespace_filter'] = ['metadata.namespace', '=', f'{CHART_NAMESPACE_PREFIX}{filters[0][-1]}']
+            service_filters = resources_filters = [extra['namespace_filter']]
+        else:
+            service_filters = resources_filters = [['metadata.namespace', '^', CHART_NAMESPACE_PREFIX]]
```
william-gr, line 74: I dont understand why you need two different vars for the same thing?

william-gr (approved):

PR #6304, opened 2021-01-22
NAS-109037 / 21.02 / set comconsole,efi on UEFI booted mseries systems

yocalebo:
Gen3 M-series systems come with m.2 nvme boot drives. This means they're UEFI booted. The current code only enables `comconsole` if we're booted UEFI and the efi firmware provides a serial device.

If only setting `comconsole`, then the BSD boot loader doesn't appear on the convenient iKVM/HTML5 website for the m-series devices. This is undesired behavior since logging into a website as opposed to running `ipmitool sol` is subjectively easier.

This change will set the kernel console setting to `comconsole,efi` if it's UEFI booted and an m-series device.

### DIFF
--- a/src/freenas-installer/etc/install.sh
+++ b/src/freenas-installer/etc/install.sh
@@ -394,7 +394,7 @@ videoconsole()
 
 mseries()
 {
-    dmidecode -s system-product-name | grep -i "TRUENAS-M" | grep -v "MINI"
+    dmidecode -s system-product-name | grep -iv "MINI" | grep -iq "TRUENAS-M"
 }
 
 save_serial_settings()
--- a/src/freenas-installer/etc/install.sh
+++ b/src/freenas-installer/etc/install.sh
@@ -394,7 +394,7 @@ videoconsole()
 
 mseries()
 {
-    dmidecode -s system-product-name | grep -i "TRUENAS-M"
+    dmidecode -s system-product-name | grep -i "TRUENAS-M" | grep -v "MINI"
 }
 
 save_serial_settings()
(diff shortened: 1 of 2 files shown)


william-gr (commented):
Just so I understand, we want this for all M-Series, not just Gen3?

yocalebo:
> Just so I understand, we want this for all M-Series, not just Gen3?

It should only happen for Gen3 series at the moment since the older generation m-series are booted using SATA DOMs which means they're not UEFI booted.

william-gr:
I thought earlier Gens were not configured to boot UEFI by default, but they could

yocalebo:
> I thought earlier Gens were not configured to boot UEFI by default, but they could

That's correct, the earlier ones are not configured to boot UEFI. However, if they do change to boot UEFI this change will apply to them as well. Which is the desired behavior.

src/freenas-installer/etc/install.sh
```diff
@@ -392,6 +392,11 @@ videoconsole()
     fi
 }
 
+mseries()
+{
+    dmidecode -s system-product-name | grep -i "TRUENAS-M" | grep -v "MINI"
```
freqlabs, line 397:
```suggestion
    dmidecode -s system-product-name | grep -iv "MINI" | grep -iq "TRUENAS-M"
```

freqlabs (approved):

src/freenas-installer/etc/install.sh
```diff
@@ -392,6 +392,11 @@ videoconsole()
     fi
 }
 
+mseries()
+{
+    dmidecode -s system-product-name | grep -iv "MINI" | grep -qv "TRUENAS-M"
```
freqlabs, line 397: Er, that wasn't what I suggested.

freqlabs (approved):

PR #6306, opened 2021-01-22
NAS-108532 / 21.02 / Add support for INTEL GPU in k8s

sonicaj:

1) Ability to recognize INTEL gpus in middleware.
2) Configure INTEL gpu automatically for k8s if the system has one.

### DIFF
--- a/src/middlewared/middlewared/etc_files/docker.py
+++ b/src/middlewared/middlewared/etc_files/docker.py
@@ -23,8 +23,8 @@ def nvidia_configuration(middleware):
 
 
 def gpu_configuration(middleware):
-    available_gpu = middleware.call_sync('device.get_info', 'GPU')
-    if available_gpu['vendor'] == 'NVIDIA':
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/etc_files/docker.py
```diff
@@ -23,8 +23,8 @@ def nvidia_configuration(middleware):
 
 
 def gpu_configuration(middleware):
-    available_gpu = middleware.call_sync('device.get_info', 'GPU')
-    if available_gpu['vendor'] == 'NVIDIA':
+    available_gpus = middleware.call_sync('device.get_info', 'GPU')
+    if any(gpu['vendor'] == 'NVIDIA' for gpu in available_gpus):
```
william-gr, line 27: Curious why no Intel here?

sonicaj: Because this patch is only required for `nvidia` and docker, for intel gpus, there is no other configuration required on docker end.

william-gr (approved):

PR #6307, opened 2021-01-22
fix traceback when rotation rate fails inside qemu/kvm vms

yocalebo:
https://github.com/truenas/py-sgio/pull/6 was merged which properly raises a `RuntimeError` when the `ioctl()` fails. This is expected behavior in qemu/kvm guest vms.

When I changed this code to use `libsgio`, I forgot to account for this fact when the previous code did.

### DIFF
--- a/src/middlewared/middlewared/plugins/device_/device_info_linux.py
+++ b/src/middlewared/middlewared/plugins/device_/device_info_linux.py
@@ -125,6 +125,10 @@ class DeviceService(Service, DeviceInfoBase):
         try:
             disk = libsgio.SCSIDevice(device_path)
             rotation_rate = disk.rotation_rate()
+        except RuntimeError:
+            # this means the ioctl failed which is
+            # expected on qemu/kvm guests
+            return
         except Exception as e:
             self.logger.error(
                 'Failed to retrieve rotational rate '


src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -125,6 +125,10 @@ def get_rotational_rate(self, device_path):
         try:
             disk = libsgio.SCSIDevice(device_path)
             rotation_rate = disk.rotation_rate()
+        except RuntimeError:
```
sonicaj, line 128: Can this exception only be raised for `qemu/kvm` guests or can it be raised otherwise as well ? Because in that case we would like to log the error ?

yocalebo: Any system could fail here, but I chose not to log it since that's what we did previously. Do you feel it's necessary to log an error here considering it's expected on qemu/kvm systems?

sonicaj: Nope, i am just seeing we are logging below but are choosing not to do so here. If you feel this is good, I am okay with that.

sonicaj (approved):

src/middlewared/middlewared/plugins/device_/device_info_linux.py
yocalebo: I've added a little more logic. Let me know if this is better.

william-gr (commented):
Even though dmidecode is now cached it will go through all the call stack to retrieve this. If we have 100 disks it seems its an unnecessary call for every disk. I wonder if there is a way to cache that before iterating through the disk list

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -122,14 +122,21 @@ def get_disk(self, name):
...
             disk = libsgio.SCSIDevice(device_path)
             rotation_rate = disk.rotation_rate()
-        except Exception as e:
-            self.logger.error(
-                'Failed to retrieve rotational rate '
-                'for disk %s with error: %r', device_path, e
-            )
+        except RuntimeError:
+            # RuntimeError() means the ioctl failed
+            if host_type == 'QEMU':
```
sonicaj, line 132: Let's log with `host_type != QEMU` and just have return once ? Agreed with William, we can allow passing the host_type perhaps ?

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -122,14 +122,21 @@ def get_disk(self, name):
...
+                # this is expected behavior on qemu/kvm guests
+                return
+            else:
+                self.logger.warning('Ioctl failed for %s', device_path)
+                return
+        except Exception:
```
sonicaj, line 138: Now that we are logging this, how about having it here `(Exception, RuntimeError)` and the same logging would work with the if condition for host type plus exception type ? We avoid duplication this way

yocalebo:
The called function in `libsgio` can only ever raise `RuntimeError`

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -88,6 +89,16 @@ def get_disks(self):
...
     def retrieve_disks_data(self):
 
+        # some disk information will fail to be retrived
+        # based on what type of guest this is. For example,
+        # if this a qemu/kvm guest, then rotation_rate is
+        # expected to fail since th ioctl to query that
+        # information is invalid. So cache the result here
+        # so that we don't have to continually call this
+        # method for every disk on the system
+        if self.HOST_TYPE is not None:
```
sonicaj, line 99: `if self.HOST_TYPE is None` ?

william-gr (commented):
Question, if we know its going to fail on kvm, why even try?

Or are there situations where it can succeed?

src/middlewared/middlewared/plugins/device_/device_info_linux.py
yocalebo: Oof, typo. Good catch, I'll fix.

yocalebo:
> Question, if we know its going to fail on kvm, why even try?
> 
> Or are there situations where it can succeed?

Valid point, I've only gotten reports of it failing on qemu/kvm. I've tested on vbox, m-series and z-series. I'll make it so that we always return `None` in that situation without even calling `get_rotational_rate()`

william-gr (approved):

PR #6318, opened 2021-01-26
NAS-109138 / 21.02 / fix up gluster.volume/peer and ctdb.shared.volume plugins

yocalebo:
Since the gluster and ctdb related APIs aren't stable, yet. I've gone ahead and done a few things:

- make gluster.volume/peer plugins follow our standard for CRUD apis
- add a new plugin that is simply a wrapper for calling external library for gluster related information
- make ctdb.shared.volume and gluster.volume/peer plugins use new wrapper
- convert ctdb.shared.volume and gluster.volume/peer to `async`
- removed unnecessary methods in gluster.volume/peer plugins respectively
- updated doc strings to be uniform between the plugins
- along the way I've fixed a few various small bugs

### DIFF
--- a/src/middlewared/middlewared/plugins/gluster_linux/peer.py
+++ b/src/middlewared/middlewared/plugins/gluster_linux/peer.py
@@ -1,26 +1,32 @@
-from glustercli.cli import peer
-from middlewared.async_validators import resolve_hostname
-from middlewared.schema import Dict, Str
-from middlewared.service import (accepts, private, job,
-                                 CallError, Service,
-                                 ValidationErrors)
-from .utils import GlusterConfig, run_method
-
-import subprocess
 import xml.etree.ElementTree as ET
+from glustercli.cli import peer
+
+from middlewared.utils import filter_list, run
+from middlewared.schema import Dict, Str, Bool
+from middlewared.service import (accepts, private, job, filterable,
+                                 CallError, CRUDService)
+from .utils import GlusterConfig
 
 
 GLUSTER_JOB_LOCK = GlusterConfig.CLI_LOCK.value
 
 
-class GlusterPeerService(Service):
+class GlusterPeerService(CRUDService):
 
     class Config:
         namespace = 'gluster.peer'
 
-    @private
-    def _parse_peer(self, p):
+    @filterable
+    async def query(self, filters=None, options=None):
+        peers = []
+        if await self.middleware.call('service.started', 'glusterd'):
+            peers = await self.middleware.call('gluster.peer.status')
+            peers = list(map(lambda i: dict(i, id=i['uuid']), peers))
 
+        return filter_list(peers, filters, options)
+
+    @private
+    async def _parse_peer(self, p):
         data = {
             'uuid': p.find('uuid').text,
             'hostname': p.find('hostname').text,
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/gluster_linux/peer.py
```diff
@@ -1,26 +1,32 @@
...
-class GlusterPeerService(Service):
+class GlusterPeerService(CRUDService):
 
     class Config:
         namespace = 'gluster.peer'
 
-    @private
-    def _parse_peer(self, p):
+    @filterable
+    async def query(self, filters=None, options=None):
```
william-gr, line 20: From your CLI changes, we should not use =None here, right @themylogin ?

src/middlewared/middlewared/plugins/gluster_linux/peer.py
```diff
@@ -35,161 +41,102 @@ def _parse_peer(self, p):
...
+        # running it on the `remote_node` so that we can get 2 different
+        # "views" of the peers in the TSP.
+        command = ['gluster', f'--remote-host={remote_node}', 'peer', 'status', '--xml']
+        cp = await run(command, check=False)
+        if cp.returncode:
+            # the gluster cli utility will return stderr
+            # to stdout and vice versa on certain failures.
+            # account for this and decode appropriately
+            err = cp.stderr if cp.stderr else cp.stdout
+            if isinstance(err, bytes):
```
william-gr, line 113: Why would it not be bytes?

src/middlewared/middlewared/plugins/gluster_linux/peer.py
```diff
@@ -35,161 +41,102 @@ def _parse_peer(self, p):
...
+        await self.middleware.call(
+            'gluster.method.run', peer.detach, {'args': ((await self.get_instance(id))['hostname'],)})
 
-    @accepts()
-    def pool(self):
+    @accepts(Dict(
+        'peer_status',
+        Bool('localhost', default=True),
+    ))
+    async def status(self, data):
```
william-gr, line 87: We seem to be doing xml things here like parsing and finding. I wonder if it would make sense to keep it in a method and not a coroutine so the event loop is not potentially blocked?

src/middlewared/middlewared/plugins/gluster_linux/run.py
```diff
@@ -0,0 +1,39 @@
...
+        try:
+            result = func(*args, **kwargs)
+        except GlusterCmdException as e:
+            # gluster cli binary will return stderr to stdout
+            # and vice versa depending on the failure.
+            rc, out, err = e.args[0]
+            err = err if err else out
+            if isinstance(err, bytes):
+                err = err.decode()
+            raise CallError(f'{err.strip()}')
```
william-gr, line 32: We probably dont need f-string here

src/middlewared/middlewared/plugins/gluster_linux/peer.py
themylogin: Yes, let's get rid of this, thank you, @william-gr

yocalebo: Do I need to wait until the CLI changes are merged? @themylogin @william-gr

themylogin: No, you can merge it now

william-gr (approved):

PR #6323, opened 2021-01-27
Fix /var/log.20210127150453 garbage directories accumulating

themylogin:


sonicaj (approved):

william-gr:
@themylogin  So we will lose the previous contents? Thats not what we were after...

themylogin:
@william-gr there is a code above that merges existing /var/log contents into syslog dataset

PR #6329, opened 2021-01-28
NAS-102413 / 21.02 / CLI followups

themylogin:


src/middlewared/middlewared/service.py
```diff
@@ -1019,7 +1019,12 @@ def get_methods(self, service, cli):
                             for j in i['properties'].values():
                                 names.add(j['_name_'])
 
-                    args_descriptions = self._cli_args_descriptions(doc, names)
+                    args_descriptions_doc = doc
+                    if attr == 'update':
+                        if do_create := getattr(svc, 'do_create', None):
+                            args_descriptions_doc += "\n" + inspect.getdoc(do_create)
```
william-gr, line 1025: Can you explain a bit further why use do_create desc for update?

themylogin: @william-gr we parse arguments descriptions (to have some help in CLI) from method's documentation. `update` arguments are same as `create` fields, that's why we copy documentation (otherwise `create` will have help but `update` will not)

william-gr: There are a few cases where this is not true, e.g. pool.update

william-gr: There are cases where we patch do_create schema enough to grant a new/different do_update

themylogin: @william-gr docstring is only used to get individual descriptions for the fields that appear in schema. If we remove e.g. `volblocksize` from update method schema for `pool.dataset`, the presence of `volblocksize` documentation in the update method doc won't change anything, it'll be just ignored.

It might be an issue if `volblocksize` was present in both create and update method schemas, and in update method schema it had a different meaning; however, I don't think we have such fields.

william-gr: What if update introduces a new attribute?

themylogin: That's why we are concatenating so both docstrings are searched for attributes help

william-gr: Ah, missed the +=, thanks.

william-gr (approved):

william-gr:
retest this please

PR #6330, opened 2021-01-29
NAS-109164 / 21.02 / dont traceback in ctdb.general.healthy

yocalebo:
If `ctdb.general.healthy` is called and the `ctdbd` service isn't started, then this will raise a traceback.

- check to make sure `ctdbd` service is started
- make sure the ctdb shared volume exists and is started
- make sure the ctdb shared volume is mounted locally

### DIFF
--- a/src/middlewared/middlewared/plugins/cluster_linux/ctdb_general.py
+++ b/src/middlewared/middlewared/plugins/cluster_linux/ctdb_general.py
@@ -79,6 +79,15 @@ class CtdbGeneralService(Service):
         Returns a boolean if the ctdb cluster is healthy.
         """
 
+        # without ctdbd running, nothing to do
+        if not await self.middleware.call('service.started', 'ctdb'):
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/cluster_linux/ctdb_general.py
```diff
@@ -79,6 +79,16 @@ async def healthy(self):
...
         """
 
+        # glusterd has to be started before anything else is
+        # possible
+        if not await self.middleware.call('service.started', 'glusterd'):
+            return False
+
+        # if the ctdb service isn't started then nothing else
+        # to do
+        if not await self.middleware.call('service.started', 'ctdb'):
```
sonicaj, line 89: Curious if this check alone is enough ?

sonicaj (approved):

william-gr (approved):

PR #6338, opened 2021-01-31
NAS-109098 / 21.02 / Start/stop s3 if underlying dataset is locked/unlocked

sonicaj:

1) Introduces attachment delegate for s3
2) Allows unsetting storage path for s3 service

### DIFF
--- a/src/middlewared/middlewared/plugins/s3_/attachments.py
+++ b/src/middlewared/middlewared/plugins/s3_/attachments.py
@@ -27,6 +27,7 @@ class MinioFSAttachmentDelegate(FSAttachmentDelegate):
         return attachment['id']
 
     async def delete(self, attachments):
+        await self.stop(attachments)
         await self.middleware.call('s3.update', {'storage_path': ''})
 
     async def toggle(self, attachments, enabled):
--- a/src/middlewared/middlewared/plugins/s3_/attachments.py
+++ b/src/middlewared/middlewared/plugins/s3_/attachments.py
@@ -1,0 +1,49 @@
+import os
+
+from middlewared.common.attachment import FSAttachmentDelegate
+
+
+class MinioFSAttachmentDelegate(FSAttachmentDelegate):
+    name = 's3'
+    title = 'S3'
+    service = 's3'
+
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/service_/services/s3.py
```diff
@@ -10,3 +12,7 @@ class S3Service(SimpleService):
     freebsd_pidfile = "/var/run/minio.pid"
 
     systemd_unit = "minio"
+
+    async def before_start(self):
+        if not (await self.middleware.call('s3.config'))['storage_path']:
+            raise CallError('Storage path must bet set to start S3 service')
```
yocalebo, line 18:
```suggestion
            raise CallError('Storage path must be set to start S3 service')
```

src/middlewared/middlewared/plugins/s3_/attachments.py
```diff
@@ -0,0 +1,50 @@
...
+        try:
+            await self.middleware.call('service.stop', 's3')
+        except Exception as e:
+            self.middleware.logger.error('Failed to stop s3: %s', e)
+
+    async def start(self, attachments):
+        try:
+            await self.middleware.call('service.start', 's3')
+        except Exception:
+            self.middleware.logger.error('Failed to start s3')
```
themylogin, line 46: Please log this with exc_info=True, same for above

themylogin (changes requested):

src/middlewared/middlewared/plugins/s3_/attachments.py
```diff
@@ -37,13 +37,13 @@ async def stop(self, attachments):
         try:
             await self.middleware.call('service.stop', 's3')
         except Exception as e:
-            self.middleware.logger.error('Failed to stop s3: %s', e)
+            self.middleware.logger.error('Failed to stop s3: %s', e, exc_info=True)
```
themylogin, line 40: There is no need for `'%s' % e` anymore

themylogin (changes requested):

william-gr (approved):

themylogin (approved):

PR #6339, opened 2021-01-31
NAS-108842 / 21.02 / Custom Catalogs Support

sonicaj:


src/middlewared/middlewared/plugins/catalogs_linux/update.py
```diff
@@ -1,47 +1,117 @@
...
+        await self.middleware.call('datastore.insert', self._config.datastore, data)
+
+        asyncio.ensure_future(self.middleware.call('catalog.sync', data['label']))
+
+        return await self.get_instance(data['label'])
+
+    @accepts(
+        Str('id'),
+    )
+    def do_delete(self, id):
```
william-gr, line 106: So you're allowing the OFFICIAL catalog to be deleted?

Seems we should have another attribute in the db (`builtin`) and do not allow builtin to be changed?

src/middlewared/middlewared/alert/source/catalogs.py
```diff
@@ -0,0 +1,18 @@
+from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, Alert, OneShotAlertClass
+
+
+class CatalogNotHealthyAlertClass(AlertClass, OneShotAlertClass):
+    deleted_automatically = False
+    level = AlertLevel.WARNING
+    category = AlertCategory.APPLICATIONS
+    title = 'Catalog Not Healthy'
+    text = '%(apps)s Applications in %(catalog)s Catalog are not healthy'
```
william-gr, line 9: Think our text usually ends with a dot?

src/middlewared/middlewared/plugins/catalogs_linux/items.py
```diff
@@ -38,9 +38,11 @@ def items(self, label, options):
         elif not os.path.exists(catalog['location']):
             self.middleware.call_sync('catalog.update_git_repository', catalog, True)
 
+        self.middleware.call_sync('alert.oneshot_delete', 'CatalogNotHealthy', f'"{label}"')
```
william-gr, line 41: Why a f-string?

william-gr: Ah, the quotes, nevermind, but why quotes?

sonicaj: So basically when we create an alert specifying `key`, it json dumps the value which results in the key having quotes and thus we need to do the same during deletion. But if we use `_key` instead, that does not happen and i changed that. Thank you for pointing it out :)

william-gr (approved):

PR #6340, opened 2021-02-01
NAS-109194 / 21.02 / handle k3s and VMs on SCALE HA appropriately

yocalebo:
- don't start VMs or k3s automatically on system event `ready` on SCALE HA systems since the failover event plugin is responsible for starting those services
- on failover master event start VMs and k3s as necessary
- fix minor `logger.info` typo when disallowing network traffic on failover backup event
- when a VM fails to start, don't log as `logger.debug` but instead use `logger.error`

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
@@ -512,10 +512,10 @@ class FailoverService(Service):
         # restart the non-critical services in the background
         self.run_call('failover.events.restart_background', fobj['services'])
 
-        # restart any k3s applications
+        # restart any kubernetes applications
         if self.run_call('kubernetes.config')['dataset']:
             logger.info('Restarting applications')
-            self.run_call('service.start', 'k3s')
+            self.run_call('service.start', 'kubernetes')
 
         # start any VMs (this will log errors if the vm(s) fail to start)
         self.run_call('vm.start_on_boot')
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
@@ -512,8 +512,12 @@ class FailoverService(Service):
         # restart the non-critical services in the background
         self.run_call('failover.events.restart_background', fobj['services'])
 
-        # TODO: jails don't exist on SCALE (yet)
-        # self.run_call('jail.start_on_boot')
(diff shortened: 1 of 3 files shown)


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event_linux.py
```diff
@@ -512,8 +512,12 @@ def vrrp_master(self, job, fobj, ifname, event):
         # restart the non-critical services in the background
         self.run_call('failover.events.restart_background', fobj['services'])
 
-        # TODO: jails don't exist on SCALE (yet)
-        # self.run_call('jail.start_on_boot')
+        # restart any k3s applications
+        if self.run_call('kubernetes.config')['dataset']:
+            logger.info('Restarting applications')
+            self.run_call('service.start', 'k3s')
```
sonicaj, line 518: Please replace `k3s` with `kubernetes`

src/middlewared/middlewared/plugins/vm/lifecycle.py
```diff
@@ -142,13 +142,16 @@ async def __event_system_ready(middleware, event_type, args):
...
     async def stop_vm(mw, vm):
         stop_job = await mw.call('vm.stop', vm['id'], {'force_after_timeout': True})
         await stop_job.wait()
         if stop_job.error:
             mw.logger.error(f'Stopping VM {vm["name"]} failed: {stop_job.error}')
 
-    if args['id'] == 'ready':
+    # we ignore the 'ready' event on an HA system since the failover event plugin
+    # is responsible for starting this service
+    if args['id'] == 'ready' and not await middleware.call('failover.licensed'):
```
sonicaj, line 154: We still need to initialize vms, let's please remove this condition as it's already being done below at L159

sonicaj (changes requested):
We should also stop `kubernetes` service on MASTER node when it's going to become BACKUP ?

yocalebo:
> We should also stop `kubernetes` service on MASTER node when it's going to become BACKUP ?

No, I'd rather not. The BACKUP event needs to be as quick as possible. Waiting on those would cause unnecessary time to the overall failover event.

yocalebo:
> > We should also stop `kubernetes` service on MASTER node when it's going to become BACKUP ?
> 
> No, I'd rather not. The BACKUP event needs to be as quick as possible. Waiting on those would cause unnecessary time to the overall failover event.

Derp, you mean if we're able to export the zpools cleanly. (On unclean export, we forcefully reboot). So you're correct, I'll fix.

sonicaj (approved):

PR #6345, opened 2021-02-02
NAS-109196 / 12.0 / Issue alert for incorrect NVDIMM firmware version on M-Series

themylogin:


src/middlewared/middlewared/alert/source/m_series_nvdimm_firmware_freebsd.py
```diff
@@ -0,0 +1,40 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class TrueNASMNVDIMMFirmwareVersionAlertClass(AlertClass):
+    category = AlertCategory.HARDWARE
+    level = AlertLevel.EMERGENCY
+    title = "Invalid NVDIMM Firmware Version"
+    text = (
+        "You are using an invalid NVDIMM firmware version %(version)s which can produce data loss. "
```
yocalebo, line 15:
```suggestion
        "You are using an invalid NVDIMM firmware version %(version)s which can cause data loss. "
```

src/middlewared/middlewared/plugins/enterprise/m_series_nvdimm_firmware_freebsd.py
```diff
@@ -0,0 +1,69 @@
...
+    @private
+    async def m_series_nvdimm_firmware_version(self):
+        if self.VERSION is None:
+            raise CallError(self.VERSION_ERROR)
+
+        return self.VERSION
+
+
+async def setup(middleware):
+    if (await middleware.call("truenas.get_chassis_hardware")).startswith("TRUENAS-M"):
```
yocalebo, line 68: Same comment as above.
```suggestion
    if await middleware.call("failover.hardware") == "ECHOWARP":
```

src/middlewared/middlewared/alert/source/m_series_nvdimm_firmware_freebsd.py
```diff
@@ -0,0 +1,40 @@
...
+
+    products = ("ENTERPRISE",)
+    hardware = True
+
+
+class TrueNASMNVDIMMFirmwareVersionAlertSource(AlertSource):
+    products = ("ENTERPRISE",)
+
+    async def check(self):
+        if (await self.middleware.call("truenas.get_chassis_hardware")).startswith("TRUENAS-M"):
```
yocalebo, line 27: I'd rather use `failover.hardware == 'ECHOWARP'` instead considering we now sell truenas minis which can be flashed `truenas-mini` which do not have nvdimms. Also, failover.hardware runs `getencstat` on the ses devices and checks the name returned in the firmware, which should be more predictable.

```suggestion
        if await self.middleware.call("failover.hardware") == "ECHOWARP":
```

yocalebo (changes requested):

src/middlewared/middlewared/plugins/enterprise/m_series_nvdimm_firmware_freebsd.py
```diff
@@ -0,0 +1,69 @@
...
+    SIZE = None
+    SIZE_ERROR = "Size not retrieved yet"
+
+    VERSION = None
+    VERSION_ERROR = "Version not retrieved yet"
+
+    @private
+    def setup_m_series_nvdimm(self):
+        try:
+            self.SIZE = int(re.match("NVDIMM region ([0-9]+)GB", sysctl.filter("dev.nvdimm.0.%desc")[0].value).group(1))
```
william-gr, line 24: As far as I know nvdimm may not exist, so first we need to make sure the device is there.

yocalebo: If it's an `ECHOWARP` platform, then it should have an NVDIMM as they come standard with the m-series.

src/middlewared/middlewared/plugins/enterprise/m_series_nvdimm_firmware_freebsd.py
```diff
@@ -0,0 +1,69 @@
...
+            self.SIZE_ERROR = str(e)
+
+        try:
+            # This should be ran before pool import so we cache the value obtained on first middlewared startup
+            cache_path = "/tmp/dev-nvdimm-0-dump_i2c"
+            if os.path.exists(cache_path):
+                with open(cache_path) as f:
+                    i2c_dump = f.read()
+            else:
+                i2c_dump = sysctl.filter("dev.nvdimm.0.dump_i2c")[0].value
```
amotin, line 35: M60 is going to have two NVDIMMs per controller, so I guess we should check both?  Or we just assume that all new M60 are good enough and never going to have wrong version?

I also don't like the idea of reading full dev.nvdimm.0.dump_i2c, since it takes seconds for a run.  There is a method to read specific byte(s) I've pointed to Nick Wolff via setting dev.nvdimm.0.arg3 to ((offset << 8) | page) and then read the binary output from dev.nvdimm.0.func27, where first 4 bytes are status and following byte is the data.  I don't have NVDIMM inserted into my system right now to check, but it was working for me before, and that is what dump_i2c does inside for each byte.

Also I don't see any check for NVDIMM maker/model.  It makes me think it will fail if we switch to different one.  I am not sure right now what and how to check, or what models we are using, but IMO that definitely needs investigation.

src/middlewared/middlewared/alert/source/m_series_nvdimm_firmware_freebsd.py
themylogin: I don't like the idea of hardcoding magic words like `ECHOWARP`. Will truenas-mini identify like TRUENAS-MINI? Then we probably have other failing bits of code that use `startswith("TRUENAS-M")` like:
```
    if product.startswith('TRUENAS-M') and current_vers.major == minimum_vers.major:
        return [
            'hint.ntb_hw.0.split=1',
            'hint.ntb_hw.0.config="ntb_pmem:1:4:0,ntb_pmem:1:4:0,ntb_transport"'
        ]
```

themylogin: Also I found this
```
is_m_series = baseboard_manufacturer == "Supermicro" and failover_hardware == "ECHOWARP"
```

yocalebo: `ECHOWARP` is not a magic string. It's the product name we use to describe the m-series head-unit chassis. As a matter of fact, if that detection fails, then HA will not work. The Z and X series platforms have their own names as well.

Yes, the new truenas minis will be tagged with `TRUENAS-MINI` so the other code bits will have an opportunity to fail. I'll open a ticket about it.

themylogin:
@yocalebo I've introduces `truenas.get_chassis_series`, please let me know what you think

amotin:
I still see unacceptable to not check NVDIMM maker/model, since we are already thinking about alternative suppliers.  The BAEBI specification tells that first thing that should be done when updating firmware is to compare that image matches that module identification.  Trying to investigate this area I've found curios fact that the information is stored in the DIMM SPD chip, and neither duplicated nor passed through through the separate NVDIMM i2c device, exposed by the ACPI DSM interface we use to read the version, so not that simple.  So far I've found this information available either via ACPI NFIT table, or AgigA Tech's proprietary driver pulls it from DMI.  I'll need to tune my NVDIMM driver's parsing of the NFIT table to expose this information in some way.  I'll work on it tomorrow.

src/middlewared/middlewared/plugins/enterprise/m_series_nvdimm_freebsd.py
```diff
@@ -0,0 +1,80 @@
...
+
+            result = []
+
+            for i in itertools.count(0):
+                try:
+                    pnpinfo = sysctl.filter(f"dev.nvdimm.{i}.%pnpinfo")[0].value
+                except IndexError:
+                    break
+
+                if "vendor=0x2c80 device=0x4e32" not in pnpinfo:
```
themylogin, line 39: @amotin please invite here someone who can check this

amotin: I was right, after discussing with Nick Wolff it should be: "vendor=0x2c80 device=0x4e36" for 16GB and "vendor=0x2c80 device=0x4e33" for 32GB.  I don't think you need to check capacity after that.

src/middlewared/middlewared/worker.py
```diff
@@ -99,7 +101,13 @@ def main_worker(*call_args):
 
 
 def receive_events():
-    c = Client('ws+unix:///var/run/middlewared-internal.sock', py_exceptions=True)
+    while True:
+        try:
+            c = Client('ws+unix:///var/run/middlewared-internal.sock', py_exceptions=True)
+            break
+        except socket.timeout:
+            MIDDLEWARE.logger.warning("Timeout connecting to middlewared internal socket, retrying...")
```
william-gr, line 109: Why would it be unable to connect or timeout?

src/middlewared/middlewared/plugins/enterprise/m_series_nvdimm_freebsd.py
```diff
@@ -0,0 +1,80 @@
...
+import json
+import logging
+import os
+import struct
+
+import sysctl
+
+from middlewared.service import CallError, private, Service
+
+logger = logging.getLogger(__name__)
```
william-gr, line 12: Not used, can be removed

src/middlewared/middlewared/plugins/enterprise/m_series_nvdimm_freebsd.py
```diff
@@ -0,0 +1,80 @@
...
+            if os.path.exists(cache_path):
+                with open(cache_path) as f:
+                    self.DATA = json.load(f)
+                    return
+
+            result = []
+
+            for i in itertools.count(0):
+                try:
+                    pnpinfo = sysctl.filter(f"dev.nvdimm.{i}.%pnpinfo")[0].value
```
william-gr, line 34: Over talking with mav, its probably better if we use his flash tool to retrieve firmware version instead of sysctl.

Also he says we dont necessarily need to do it before pool import and it should work.

src/middlewared/middlewared/alert/source/m_series_nvdimm_firmware_freebsd.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding=utf-8 -*-
+import logging
+
+from middlewared.alert.base import Alert, AlertClass, AlertCategory, AlertLevel, AlertSource
+from middlewared.service_exception import CallError
+
+logger = logging.getLogger(__name__)
```
william-gr, line 7: Can be removed

src/freenas/usr/local/lib/middlewared_truenas/plugins/truenas.py
```diff
@@ -119,6 +119,15 @@ async def get_chassis_hardware(self):
         # Give up
         return 'TRUENAS-UNKNOWN'
 
+    @accepts()
+    async def get_chassis_series(self):
```
william-gr, line 123: I agree with Caleb. I dont see the need for this.

We can just do truenas.get_chassis_hardware == TRUENAS-M

Fix get_chassis_hardware to not return MINI

src/middlewared/middlewared/alert/source/m_series_nvdimm_firmware_freebsd.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class TrueNASMNVDIMMFirmwareVersionAlertClass(AlertClass):
+    category = AlertCategory.HARDWARE
+    level = AlertLevel.EMERGENCY
+    title = "Invalid NVDIMM Firmware Version"
+    text = (
+        "Your NVDIMM controller %(index)d is using an invalid firmware version %(version)s which can cause data loss."
+        "Please contact support immediately."
```
william-gr, line 16: @Mrt134 are we OK with this text?

@wmohanlon Thoughts on this one? You OK with that wording? And the level of this alert being emergency?

src/middlewared/middlewared/worker.py
themylogin: Due to race condition between main process listening on socket and worker startup

william-gr: Is it hard to fix that race? Do not start process pool until after we start listening

src/middlewared/middlewared/alert/source/m_series_nvdimm_firmware_freebsd.py
```diff
@@ -0,0 +1,38 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class TrueNASMNVDIMMFirmwareVersionAlertClass(AlertClass):
+    category = AlertCategory.HARDWARE
+    level = AlertLevel.EMERGENCY
+    title = "Invalid NVDIMM Firmware Version"
+    text = (
+        "Your NVDIMM controller %(index)d is using an invalid firmware version %(version)s which can cause data loss. "
```
yocalebo, line 15: The text needs to be changed. NVDIMM controller is confusing since there is an A controller and B controller. We'll probably need to specify which controller has the incorrect NVDIMM firmware.

(i.e. NVDIMM device on controller (A/B) is using invalid firmware....)

themylogin: This refers to the device number in /dev. The A/B controller is displayed for each alert in the UI and in the E-Mail.

src/freenas/usr/local/libexec/freenas-debug/hardware/hardware.sh
```diff
@@ -211,6 +211,14 @@ hardware_freebsd()
 		sas3flash -listall
 		section_footer
 	fi
+
+	if case "$(midclt call truenas.get_chassis_hardware)" in TRUENAS-M*) true;; *) false;; esac; then
```
william-gr, line 215: Why this complicated if case and not `if ["$(midclt call truenas.get_chassis_hardware)" = "TRUENAS-M" ]; then`

src/middlewared/middlewared/alert/source/m_series_nvdimm_firmware_freebsd.py
```diff
@@ -0,0 +1,38 @@
...
+
+    products = ("ENTERPRISE",)
+    hardware = True
+
+
+class TrueNASMNVDIMMFirmwareVersionAlertSource(AlertSource):
+    products = ("ENTERPRISE",)
+
+    async def check(self):
+        if (await self.middleware.call("truenas.get_chassis_hardware")).startswith("TRUENAS-M"):
```
william-gr, line 27: Why startswith and not "=="

src/middlewared/middlewared/alert/source/m_series_nvdimm_firmware_freebsd.py
william-gr: @Mrt134 @wmohanlon ping

src/middlewared/middlewared/alert/source/m_series_old_bios_version_freebsd.py
```diff
@@ -0,0 +1,28 @@
...
+logger = logging.getLogger(__name__)
+
+
+class TrueNASMOldBIOSVersionAlertClass(AlertClass):
+    category = AlertCategory.HARDWARE
+    level = AlertLevel.EMERGENCY
+    title = "Old BIOS Version"
+    text = (
+        "Your M-Series TrueNAS has an old BIOS version. "
+        "Please contact support immediately."
```
william-gr, line 15: @Mrt134 @wmohanlon same question here, please

src/middlewared/middlewared/alert/source/m_series_old_bios_version_freebsd.py
```diff
@@ -0,0 +1,28 @@
...
+    )
+
+    products = ("ENTERPRISE",)
+    hardware = True
```
william-gr, line 19: @wmohanlon Are we OK with it being part of the proactive support?

src/middlewared/middlewared/alert/source/m_series_old_bios_version_freebsd.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class TrueNASMNVDIMMFirmwareVersionAlertSource(AlertSource):
+    products = ("ENTERPRISE",)
+
+    async def check(self):
+        if (await self.middleware.call("truenas.get_chassis_hardware")).startswith("TRUENAS-M"):
```
william-gr, line 26: `==` ?

src/freenas/usr/local/lib/middlewared_truenas/plugins/truenas.py
```diff
@@ -90,6 +90,8 @@ async def get_chassis_hardware(self):
 
         chassis = await run('dmidecode', '-s', 'system-product-name', check=False)
         chassis = chassis.stdout.decode(errors='ignore').split('\n', 1)[0].strip()
+        if chassis == 'TRUENAS-MINI':
+            return 'TRUENAS-UNKNOWN'
         if chassis.startswith(('TRUENAS-M', 'TRUENAS-X', 'TRUENAS-Z')):
```
william-gr, line 95: Do we use the exact version M40 X10, X20 anywhere?

yocalebo: I still don't know why we wouldn't just use `failover.hardware == 'ECHOWARP'`

src/middlewared/middlewared/alert/source/m_series_nvdimm_firmware_freebsd.py
yocalebo: `NVDIMM device (/dev/%(index)d is using...`

src/freenas/usr/local/lib/middlewared_truenas/plugins/truenas.py
william-gr: I can understand that ECHOWARP may make it more difficult for people reading the code, its not directly clear it references M-series, its "obscure"

src/middlewared/middlewared/alert/source/m_series_old_bios_version_freebsd.py
wmohanlon: Can we lose the word "immediately", to soften it a bit?  I don't want to downplay the importance, but I don't want support to get hammered, either.

src/middlewared/middlewared/alert/source/m_series_old_bios_version_freebsd.py
wmohanlon: Yes.

src/freenas/usr/local/lib/middlewared_truenas/plugins/truenas.py
yocalebo: But it's the proper check. Write another method in `failover` plugin that returns True/False.

def is_mseries(self):
    if failover.hardware == 'ECHOWARP':
        return True
    chassis = truenas.get_chassis_hardware
    if chassis.startswith('TRUENAS-M') and 'MINI' not in chassis:
        return True
    return False

src/middlewared/middlewared/alert/source/m_series_nvdimm_firmware_freebsd.py
william-gr: Agreed, perhaps loose the /dev/ and just use nvdimm%(index)d

src/freenas/usr/local/lib/middlewared_truenas/plugins/truenas.py
william-gr: Its redundant, we already have the chassis hardware for this. We dont need is_xseries, is_zseries and is_mseries methods. One is enough

themylogin: @william-gr what have we decided then? `==` or `startswith`?

themylogin: I am not a fan of `failover.hardware == 'ECHOWARP'` either

william-gr: @themylogin do we check for specific hardware anything using that method? If we do then we keep with startswith

themylogin: @william-gr yes, we do

src/middlewared/middlewared/alert/source/m_series_nvdimm_firmware_freebsd.py
```diff
@@ -0,0 +1,38 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class TrueNASMNVDIMMFirmwareVersionAlertClass(AlertClass):
+    category = AlertCategory.HARDWARE
+    level = AlertLevel.EMERGENCY
+    title = "Invalid NVDIMM Firmware Version"
+    text = (
+        "Your NVDIMM controller nvdimm%(index)d is using an invalid firmware version %(version)s which can cause data "
```
william-gr, line 15: NVDIMM device (nvdimm%(index)d)

I believe device is a better word than controller

src/middlewared/middlewared/alert/source/m_series_nvdimm_firmware_freebsd.py
```diff
@@ -0,0 +1,38 @@
...
+
+class TrueNASMNVDIMMFirmwareVersionAlertSource(AlertSource):
+    products = ("ENTERPRISE",)
+
+    async def check(self):
+        if (await self.middleware.call("truenas.get_chassis_hardware")).startswith("TRUENAS-M"):
+            for nvdimm in await self.middleware.call("enterprise.m_series_nvdimm"):
+                size_to_version = {16: "2.2", 32: "2.4"}
+                if nvdimm["size"] not in size_to_version:
+                    raise CallError(f"Unknown NVDIMM size: {nvdimm['size']}")
```
william-gr, line 31: Does it make any sense to raise a CallError here? Wouldnt just a log be better?

themylogin: It will create an alert and user will know immediately that his hardware is weird. Log will just get lost.

src/middlewared/middlewared/alert/source/m_series_nvdimm_firmware_freebsd.py
wmohanlon: Same thought as other comment -- could we leave out "immediately" ?

src/middlewared/middlewared/alert/source/m_series_old_bios_version_freebsd.py
wmohanlon: Yes

yocalebo (approved):
Looks good with the exception that we have `logger` objects not being used.

src/middlewared/middlewared/alert/source/m_series_old_bios_version_freebsd.py
```diff
@@ -0,0 +1,28 @@
...
+logger = logging.getLogger(__name__)
+
+
+class TrueNASMOldBIOSVersionAlertClass(AlertClass):
+    category = AlertCategory.HARDWARE
+    level = AlertLevel.EMERGENCY
+    title = "Old BIOS Version"
+    text = (
+        "Your M-Series TrueNAS has an old BIOS version. "
+        "Please contact support."
```
Mrt134, line 15:
```suggestion
        "Please contact iXsystems Support using the form in System > Support."
```

src/middlewared/middlewared/alert/source/m_series_nvdimm_firmware_freebsd.py
```diff
@@ -0,0 +1,38 @@
...
+logger = logging.getLogger(__name__)
+
+
+class TrueNASMNVDIMMFirmwareVersionAlertClass(AlertClass):
+    category = AlertCategory.HARDWARE
+    level = AlertLevel.EMERGENCY
+    title = "Invalid NVDIMM Firmware Version"
+    text = (
+        "Your NVDIMM device (nvdimm%(index)d) is using an invalid firmware version %(version)s which can cause data "
+        "loss. Please contact support."
```
Mrt134, line 16:
```suggestion
        "loss. Please contact iXsystems Support using the form in System > Support."
```

src/middlewared/middlewared/alert/source/m_series_nvdimm_firmware_freebsd.py
```diff
@@ -0,0 +1,38 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class TrueNASMNVDIMMFirmwareVersionAlertClass(AlertClass):
+    category = AlertCategory.HARDWARE
+    level = AlertLevel.EMERGENCY
+    title = "Invalid NVDIMM Firmware Version"
+    text = (
+        "Your NVDIMM device (nvdimm%(index)d) is using an invalid firmware version %(version)s which can cause data "
```
Mrt134, line 15:
```suggestion
        "NVDIMM device (nvdimm%(index)d) is using invalid firmware version %(version)s which can cause data "
```

src/middlewared/middlewared/alert/source/m_series_old_bios_version_freebsd.py
```diff
@@ -0,0 +1,28 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class TrueNASMOldBIOSVersionAlertClass(AlertClass):
+    category = AlertCategory.HARDWARE
+    level = AlertLevel.EMERGENCY
+    title = "Old BIOS Version"
+    text = (
+        "Your M-Series TrueNAS has an old BIOS version. "
```
Mrt134, line 14:
```suggestion
        "This TrueNAS M-Series has an old BIOS version. "
```

Mrt134 (commented):
I put a couple suggestions in:
- Minor text rework to take the you/your out.
- Updated the "contact support" to also provide a method of contacting Support. I suggested the form in System > Support as that seems to be the most resilient option (phone numbers change), but am generally trying to provide the reader with a path to resolve the alert.

william-gr (approved):
Thanks!

PR #6346, opened 2021-02-02
NAS-109228 / 21.02 / Allow running a migration for chart release upgrades

sonicaj:
This commit adds changes allowing chart devs to have a migration in a chart release version which will allow them to change format of user specified input to a new format if they have made breaking changes in the configuration format while extending the chart or fixing some issues.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/upgrade.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/upgrade.py
@@ -1,10 +1,14 @@
 import asyncio
+import copy
 import errno
+import importlib.util
 import os
 import shutil
 import tempfile
 import yaml
 
+from collections import Callable
+from inspect import signature
 from pkg_resources import parse_version
 
 from middlewared.schema import Bool, Dict, Str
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/chart_releases_linux/upgrade.py
```diff
@@ -156,6 +162,33 @@ async def upgrade(self, job, release_name, options):
 
         return chart_release
 
+    @private
+    def upgrade_values(self, release, new_version_path):
```
william-gr, line 166: I know we talked about a migration using python, but i would if its possible to make it language agnostic, passing the config as a temporary file location in json format?

Thoughts?

src/middlewared/middlewared/plugins/chart_releases_linux/upgrade.py
```diff
@@ -156,6 +161,29 @@ async def upgrade(self, job, release_name, options):
...
+        config = copy.deepcopy(release['config'])
+        chart_version = release['chart_metadata']['version']
+        migration_path = os.path.join(new_version_path, 'migrations')
+        migration_files = [os.path.join(migration_path, k) for k in (f'migrate_from_{chart_version}', 'migrate')]
+        if not os.path.exists(migration_path) or all(not os.access(p, os.X_OK) for p in migration_files):
+            return config
+
+        # This is guaranteed to exist based on above check
+        file_path = next(f for f in migration_files if os.access(f, os.X_OK))
+        cp = subprocess.Popen([file_path, json.dumps(config)], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
william-gr, line 175: The idea was to pass the json as a temporary filename parameter.

Since this may be really big on some situations we may exceed limits and also such a big arg is not recommended.

william-gr (approved):

PR #6352, opened 2021-02-03
NAS-109234 / 21.02 / Fix potential division-by-zero error

anodos325:
Ensure that we don't call core.bulk with empty list of params from
restart_services_after_unlock(). Also return early in core.bulk
if no params are passed in.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/unlock.py
+++ b/src/middlewared/middlewared/plugins/pool_/unlock.py
@@ -91,7 +91,14 @@ class PoolDatasetService(Service):
             if not to_restart:
                 return
 
-            await self.middleware.call('core.bulk', 'service.restart', to_restart)
+            restart_job = await self.middleware.call('core.bulk', 'service.restart', to_restart)
+            statuses = await restart_job.wait()
+            for idx, srv_status in enumerate(statuses):
+                if srv_status['error']:
+                    self.logger.error(
+                        'Failed to restart %r service after %r unlock: %s',
+                        to_restart[idx], dataset_name, srv_status['error']
+                    )
             if 'jails' in services_to_restart:
                 await self.middleware.call('jail.rc_action', ['RESTART'])
             if 'vms' in services_to_restart:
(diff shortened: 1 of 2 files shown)


anodos325:
Observed in API test logs
```
    "error": "division by zero",
    "exception": "Traceback (most recent call last):\n  File \"/usr/lib/python3/dist-packages/middlewared/job.py\", line 378, in run\n    await self.future\n  File \"/usr/lib/python3/dist-packages/middlewared/job.py\", line 414, in __run_body\n    rv = await self.method(*([self] + args))\n  File \"/usr/lib/python3/dist-packages/middlewared/schema.py\", line 998, in nf\n    return await f(*args, **kwargs)\n  File \"/usr/lib/python3/dist-packages/middlewared/service.py\", line 1321, in bulk\n    progress_step = 100 / len(params)\nZeroDivisionError: division by zero\n",
```

themylogin (approved):

src/middlewared/middlewared/plugins/pool_/unlock.py
```diff
@@ -87,9 +87,11 @@ async def restart_vms_after_unlock(self, dataset_name):
...
     async def restart_services_after_unlock(self, dataset_name, services_to_restart):
         try:
-            await self.middleware.call('core.bulk', 'service.restart', [
-                [i] for i in set(services_to_restart) - {'jails', 'vms'}
-            ])
+            to_restart = [[i] for i in set(services_to_restart) - {'jails', 'vms'}]
+            if not to_restart:
+                return
+
+            await self.middleware.call('core.bulk', 'service.restart', to_restart)
```
william-gr, line 94: @sonicaj Should we wait core.bulk to finish?

sonicaj: I think we should, this would ensure that once the unlock job completes we have necessary services restarted/reloaded.

sonicaj (approved):
Thanks

PR #6364, opened 2021-02-04
NAS-109078 / 12.0 / Explain system.info license addhw

themylogin:


src/middlewared/middlewared/plugins/system.py
```diff
@@ -485,7 +486,12 @@ def _get_license():
             "customer_name": licenseobj.customer_name,
             "expired": licenseobj.expired,
             "features": [],
-            "addhw": licenseobj.addhw,
+            "addhw_raw": licenseobj.addhw,
```
william-gr, line 489: Thats backward incompatible.
The new one should probably be "addhw_detail" or something?

william-gr (approved):

PR #6378, opened 2021-02-05
NAS-109286 / 21.02 / Fix validation for kerberos default_ccache_type

anodos325:
This fixes validation for the parameter. Does not guarantee the parameter works in any comprehensible way on the server.

### DIFF
--- a/src/middlewared/middlewared/plugins/kerberos.py
+++ b/src/middlewared/middlewared/plugins/kerberos.py
@@ -187,9 +187,16 @@ class KerberosService(ConfigService):
                     raise CallError(f'[{e}] is not a supported encryption type')
 
         if data['ptype'] == 'cctype':
-            for cctype in ['DIR', 'FILE', 'MEMORY']:
-                if cctype not in data['value']:
-                    raise CallError(f'{data["value"]} is an unsupported cctype')
+            available_types = ['FILE', 'MEMORY']
+            if osc.IS_FREEBSD:
+                available_types.append('SCC')
+            else:
+                available_types.append('DIR')
+
+            if data['value'] not in available_types:
+                raise CallError(f'[{data["value"]}] is an unsupported cctype. '
+                                f'Available types are {", ".join(available_types)}. '
+                                'This parameter is case-sensitive')
 
         if data['ptype'] == 'keytab':
             try:


src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -187,9 +187,8 @@ async def _validate_param_type(self, data):
                     raise CallError(f'[{e}] is not a supported encryption type')
 
         if data['ptype'] == 'cctype':
-            for cctype in ['DIR', 'FILE', 'MEMORY']:
-                if cctype not in data['value']:
-                    raise CallError(f'{data["value"]} is an unsupported cctype')
+            if data['value'] not in ['DIR', 'FILE', 'MEMORY']:
```
yocalebo, line 190: Few lines above you normalize `data['value']` by calling `.upper()`. Why not do that here?

yocalebo (approved):

src/middlewared/middlewared/plugins/kerberos.py
anodos325: Underlying kerberos behavior for parsing sting parameters is case-sensitive. I expanded the error message to be a little more intuitive. I'm trying to strike a balance here between a total free-for-all for these aux parameters (leading to more broken configs) and having something more user-friendly (encouraging users to change these parameters - leading to more broken configs).

src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -187,9 +187,11 @@ async def _validate_param_type(self, data):
                     raise CallError(f'[{e}] is not a supported encryption type')
 
         if data['ptype'] == 'cctype':
-            for cctype in ['DIR', 'FILE', 'MEMORY']:
-                if cctype not in data['value']:
-                    raise CallError(f'{data["value"]} is an unsupported cctype')
+            available_types = ['DIR', 'FILE', 'MEMORY']
+            if data['value'] not in available_types:
+                raise CallError(f'{data["value"]} is an unsupported cctype. '
+                                f'Available types are {", ".join(available_types)} '
```
william-gr, line 193: Missing dot at the end of sentence here

william-gr (approved):

PR #6380, opened 2021-02-05
NAS-109283 / 21.02 / Better handle system managed internal datasets

sonicaj:

1) We make sure to remove system managed internal datasets from children attribute of a dataset.
2) We make sure that we are not able to create a dataset under system managed internal datasets using `pool.dataset.create` api.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -3218,6 +3218,12 @@ class PoolDatasetService(CRUDService):
                 [('id', '=', data['name'].rsplit('/', 1)[0])]
             )
 
+        if await self.is_internal_dataset(data['name']):
+            verrors.add(
+                f'{schema}.name',
+                f'{data["name"]!r} is using system internal managed dataset. Please specify a different parent.'
+            )
+
         if not parent:
             verrors.add(
                 f'{schema}.name',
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2765,28 +2765,30 @@ def query(self, filters, options):
...
-        filters.extend([
+        children_filters.extend([
             ['id', 'rnin', '.glusterfs'],
         ])
 
         if osc.IS_LINUX:
             k8s_config = self.middleware.call_sync('kubernetes.config')
             if k8s_config['dataset']:
-                filters.append(['id', '!^', f'{k8s_config["dataset"]}/'])
+                children_filters.append(['id', '!^', f'{k8s_config["dataset"]}/'])
```
themylogin, line 2789: Do we exclude `k8s_config["dataset"]` itself?

sonicaj: Nope, we still show it

sonicaj: The attachment delegate should handle that case if the user intends to delete the dataset

sonicaj: Good point :)

themylogin (approved):
Great, thank you!

william-gr (approved):

PR #6382, opened 2021-02-05
NAS-109292 / 21.02 / Bug fix for chart release name regex

sonicaj:
This commit makes sure that we don't match string partially for chart release name

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
@@ -263,7 +263,7 @@ class ChartReleaseService(CRUDService):
             Dict('values', additional_attrs=True),
             Str('catalog', required=True),
             Str('item', required=True),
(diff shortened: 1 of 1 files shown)


themylogin (approved):

src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
```diff
@@ -263,7 +263,7 @@ async def perform_actions(self, context):
             Dict('values', additional_attrs=True),
             Str('catalog', required=True),
             Str('item', required=True),
-            Str('release_name', required=True, validators=[Match(r'[a-z0-9]([-a-z0-9]*[a-z0-9])?')]),
+            Str('release_name', required=True, validators=[Match(r'[a-z0-9]([-a-z0-9]*[a-z0-9])?$')]),
```
william-gr, line 266: Shouldnt you have `^` at the beginning too, then?

william-gr (approved):

PR #6383, opened 2021-02-06
NAS-108574 / 21.04 / Allow using TN certs/ca's in chart releases

sonicaj:


themylogin (approved):

src/middlewared/middlewared/plugins/catalogs_linux/features.py
```diff
@@ -8,6 +8,8 @@
 SUPPORTED_FEATURES = {
     'normalize/interfaceConfiguration',
     'normalize/ixVolume',
+    'definitions/certificate',
+    'definitions/certificate_authority',
```
themylogin, line 12: Why everything else is camelCase but here we have underscore?

themylogin (changes requested):

src/middlewared/middlewared/plugins/catalogs_linux/features.py
sonicaj: Thank you for catching that :) It should be camel case here as well

themylogin (approved):

PR #6384, opened 2021-02-06
NAS-108692 / 21.04 / Allow retrieving pod / container logs from a chart release

sonicaj:


themylogin:
@sonicaj can they be multi-megabyte files that will have zero sense to be displayed in browser? If so, we can use our job streaming facilities.

src/middlewared/middlewared/plugins/chart_releases_linux/resources.py
```diff
@@ -27,6 +28,68 @@ async def pod_console_choices(self, release_name):
...
+
+    @accepts(
+        Str('release_name'),
+        Dict(
+            'options',
+            Str('pod_name', required=True, empty=False),
+            Str('container_name', required=True, empty=False),
+        )
+    )
+    @job(lock='dataset_export_keys', pipes=['output'])
```
themylogin, line 66: Why `lock='dataset_export_keys'`?

src/middlewared/middlewared/plugins/chart_releases_linux/resources.py
```diff
@@ -27,6 +28,68 @@ async def pod_console_choices(self, release_name):
...
+        )
+        temp_path = None
+        try:
+            with tempfile.NamedTemporaryFile(mode='w', delete=False) as f:
+                temp_path = f.name
+                os.chmod(temp_path, 0o600)
+                f.write(logs)
+
+            with open(temp_path, 'rb') as f:
+                shutil.copyfileobj(f, job.pipes.output.w)
```
themylogin, line 88: You may just write directly to the pipe.

There is little point in piping `k8s.pod.get_logs` result since it is already loaded to the RAM entirely

themylogin (changes requested):

src/middlewared/middlewared/plugins/kubernetes_linux/pods.py
```diff
@@ -25,3 +28,52 @@ async def query(self, filters, options):
                 pod['events'] = events[pod['metadata']['uid']]
 
         return filter_list(pods, filters, options)
+
+    async def get_logs(self, pod, container, namespace):
+        async with api_client() as (api, context):
+            return await context['core_api'].read_namespaced_pod_log(
```
themylogin, line 34: Don't we have any sort of streaming API here? I don't like the entire log being loaded into RAM

src/middlewared/middlewared/plugins/kubernetes_linux/pods.py
```diff
@@ -25,3 +28,52 @@ async def query(self, filters, options):
...
+        release_data = await self.middleware.call('chart.release.get_instance', release)
+
+        async with api_client() as (api, context):
+            self.watch = Watch()
+            async with self.watch.stream(
+                context['core_api'].read_namespaced_pod_log, name=pod, container=container,
+                namespace=release_data['namespace'],
+            ) as stream:
+                async for event in stream:
+                    self.send_event('ADDED', fields={'data': event})
```
themylogin, line 67: How much of the existing logs will be streamed here?

themylogin (changes requested):

william-gr (approved):

src/middlewared/middlewared/plugins/kubernetes_linux/pods.py
```diff
@@ -25,3 +31,73 @@ async def query(self, filters, options):
...
         return filter_list(pods, filters, options)
+
+    async def get_logs(self, pod, container, namespace, tail_lines=500, limit_bytes=None):
+        async with api_client() as (api, context):
+            return await context['core_api'].read_namespaced_pod_log(
+                name=pod, container=container, namespace=namespace, tail_lines=tail_lines, limit_bytes=limit_bytes,
+            )
+
+
+class KubernetesPodLogsFileFollowTailEventSource(EventSource):
```
themylogin, line 42: What `FileFollowTail` has to deal with it?

themylogin (approved):

PR #6387, opened 2021-02-08
NAS-109304 / 21.02 / set xattr=sa by default in zfs.dataset.create API

yocalebo:
The public API `pool.dataset.create` will set the `xattr` zfs property to 'sa' if it's not provided by end-user. This mitigates, potentially, quite severe performance problems wrt to EA's. The private API `zfs.dataset.create` needs to do the same.

### DIFF
--- a/src/middlewared/middlewared/plugins/zfs.py
+++ b/src/middlewared/middlewared/plugins/zfs.py
@@ -781,6 +781,13 @@ class ZFSDatasetService(CRUDService):
         for k, v in data['properties'].items():
             params[k] = v
 
+        # it's important that we set xattr=sa for various
+        # performance reasons related to ea handling
+        # pool.dataset.create already sets this by default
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -781,6 +781,13 @@ def do_create(self, data):
         for k, v in data['properties'].items():
             params[k] = v
 
+        # it's important that we set xattr=sa for various
+        # performance reasons related to ea handling
+        # pool.dataset.create already sets this by default
+        # so mirror the behavior here
+        if 'xattr' not in params.keys():
```
william-gr, line 788: You dont need keys here, in fact that may only make things slower

anodos325 (approved):

william-gr (approved):

PR #6394, opened 2021-02-09
NAS-109305 / 21.04 / Read values from a yaml file as well if present for chart release

sonicaj:
This commit adds changes to give chart devs more control on specifying values for a chart. Right now values are constructed only from questions.yaml which works but there are cases where chart dev would not like to have something modified by the user and then has to specify that value in questions.yaml as hidden and editable=false. This will allow the chart dev to specify such values separately and then change them as desired.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
@@ -5,6 +5,7 @@ import errno
 import itertools
 import os
 import shutil
+import subprocess
 import tempfile
 import yaml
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
```diff
@@ -367,13 +367,19 @@ async def do_create(self, job, data):
 
             job.set_progress(75, 'Installing Catalog Item')
 
+            file_args = []
+            if os.path.exists(os.path.join(chart_path, 'ix_values.yaml')):
+                file_args = ['-f', os.path.join(chart_path, 'ix_values.yaml')]
```
william-gr, line 372: Repeating same code in three places, wonder if there is something to do about it

src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
```diff
@@ -486,6 +490,13 @@ async def do_delete(self, job, release_name):
         job.set_progress(100, f'{release_name!r} chart release deleted')
         return True
 
+    @private
+    def helm_file_args(self, chart_path):
+        file_args = ['-f']
+        if os.path.exists(os.path.join(chart_path, 'ix_values.yaml')):
+            file_args = [os.path.join(chart_path, 'ix_values.yaml'), '-f']
+        return file_args
```
william-gr, line 498: I meant more, all between context manager for tempfile, write config and then run command seem to be the same. Only changing the helm parameter (install vs upgrade)

william-gr (approved):
Sweet, thanks!

william-gr:
```
  File "/usr/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 786, in exec_module
  File "<frozen importlib._bootstrap_external>", line 923, in get_code
  File "<frozen importlib._bootstrap_external>", line 853, in source_to_code
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/dpkg-src/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py", line 486
    cp = await run(
         ^
SyntaxError: 'await' outside async function
```

PR #6395, opened 2021-02-09
NAS-109325 / 21.04 / ignore check_path_resides_within_volume when clustered

yocalebo:
`check_path_resides_within_volume` doesn't apply when a sharing service is configured for clustering. For example, SMB takes a relative path to the gluster volume. (i.e. `/`) so ignore the checks entirely when `gluster_bypass=True`


yocalebo:
At the time of writing this, [githubstatus](https://www.githubstatus.com/) shows yellow. I have pushed a 2nd commit to only make the `.glusterfs` checks occur on linux....but it's not showing up here.

william-gr (approved):

PR #6398, opened 2021-02-09
NAS-109314 / 21.04 / Save preferred trains for a user

sonicaj:
This PR introduces changes where we save preferred train for a user which UI will use to display catalog items. This helps the user to not have to each time go and change/select train for a catalog in the UI for installing some application.

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/update.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/update.py
@@ -60,10 +60,24 @@ class CatalogService(CRUDService):
             )
         return catalog
 
+    @private
+    async def common_validation(self, catalog, schema, data):
+        found_trains = set(catalog['trains'])
+        diff = set(data['preferred_trains']) - found_trains
+        verrors = ValidationErrors()
+        if diff:
+            verrors.add(
+                f'{schema}.preferred_trains',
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/catalogs_linux/update.py
```diff
@@ -22,6 +22,7 @@ class CatalogModel(sa.Model):
     repository = sa.Column(sa.Text(), nullable=False)
     branch = sa.Column(sa.String(255), nullable=False)
     builtin = sa.Column(sa.Boolean(), nullable=False, default=False)
+    preferred_trains = sa.Column(sa.JSON(type=list))
```
themylogin, line 25: Why don't we make it `nullable=False`?

themylogin (changes requested):

src/middlewared/middlewared/plugins/catalogs_linux/update.py
sonicaj: It's `nullable=False` by default

themylogin: Then your migration lacks code to set NULL column to NOT NULL

themylogin (approved):
Thank you!

william-gr (approved):

PR #6399, opened 2021-02-09
NAS-109313 / 21.04 / Skip reading invalid trains

sonicaj:


themylogin (changes requested):
`train.startswith('.') or train != 'library'` is always true

themylogin (approved):

william-gr (approved):

PR #6401, opened 2021-02-09
NAS-109336 / 21.04 / Delete snapshots in descendant filesystems as well on chart release upgrade

sonicaj:
This PR adds changes to delete snapshots of descendant chart release filesystems as well if they exist with the same name which can happen on failed upgrades.


themylogin (approved):

Ornias1993 (approved):
Nice clean solution! :)

william-gr (approved):

PR #6402, opened 2021-02-09
NAS-109302 / 21.04 / Force termination of namespace for misbehaving apps on deletion

sonicaj:


src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
```diff
@@ -458,14 +459,23 @@ async def do_update(self, job, chart_release, data):
         job.set_progress(100, 'Update completed for chart release')
         return await self.get_instance(chart_release)
 
-    @accepts(Str('release_name'))
+    @accepts(
+        Str('release_name'),
+        Dict(
+            'options',
+            Int('wait_secs', default=300)
```
themylogin, line 466: Can `timeout` be a better commonly used name?

sonicaj: Agreed

Ornias1993 (approved):

william-gr (approved):

sonicaj:
Closing this PR as the changes introduced in this PR are mostly cosmetic where k8s stops owning the underlying resources and this results in the underlying resources ( in our case containers possibly running indefinitely ) remaining in the same state.

Ornias1993:
> Closing this PR as the changes introduced in this PR are mostly cosmetic where k8s stops owning the underlying resources and this results in the underlying resources ( in our case containers possibly running indefinitely ) remaining in the same state.

Shoot, indeed... which would lead to "just another error"

Ornias1993:
@sonicaj 
I've been thinking about this and got some more feedback from users.
It seems this is caused by underlaying resources not being removed correctly.

So you're right, deleting just the namespace would be cosmetic.
However every app has it's own namespace, so we could just nuke everything in the namespace:
`kubectl delete all --all -n ix-applicationname`

sonicaj:
@Ornias1993 okay let me clarify, from my tests we have unkillable containers with non-existent PID, so trying to nuke everything does not help.
What happens if we try to nuke everything is that it won't work and we will get stuck in terminating state as k8s waits for the CRI to confirm deletion of resources which are not going to happen because of the above issue and issue only goes away after a restart of docker or in our terms the `kubernetes` service. So k8s will remain in that state indefinitely waiting for confirmation from CRI that underlying resources have been released.

Now what this PR did was that it instructed k8s to not wait for confirmation from the CRI and just consider the resources cleaned ( sort of ), the issue with this approach is that say a pod is using host path volumes ( which is very common in our case ), it will continue doing so and if they are ix volumes, we won't be able to delete the chart release datasets which will have us in an inconsistent state.

I would definitely consider this an upstream issue wrt CRI which in our case is docker. There are issues in k8s as well for this but they have been closed with the same feedback.

Ornias1993:
@sonicaj
Ahh so: 
- The namespace gets stuck because the service isn't deleting
- The service is stuck because the pod isn't deleting
- The pod isn't deleting because the PID doesn't exist

Yeah in that case the underlaying issue is not solved by above command.

Would a reboot solve this cleanly? I guess so?

sonicaj:
Restart of kubernetes service would as well. However we would have datasets leftover of chart release which wouldn’t be deleted so they would have to be deleted manually.

Ornias1993:
@sonicaj 
Yeah had that issue a few times too...
But that is actually a problem that can be solved by the middleware, making sure there are no orphaned releases in `ix-applications/releases` when starting k8s

sonicaj:
@Ornias1993 thats a nice suggestion, only concern would be accidentally deleting user data if we get in an inconsistent state wrt k3s db which can happen ( unlikely but possible ). 

Can you please create a jira ticket about it ? I think we can perhaps rename release dataset to counter for this.

Ornias1993:
@sonicaj 
Will do :)
Renaming is a nice idea, i've had some undeletable datasets in the past and almost always deleting works like a charm 👍 

Another solution (besides referencing the k3s db) is keeping track of installed releases in the SCALE database and just crossreferencing the database with the datasets.

Ornias1993:
Done:
https://jira.ixsystems.com/browse/NAS-109609

PR #6404, opened 2021-02-10
NAS-109307 / 21.04 / Reload/Restart dependent services/chart releases after renewing ACME certificate

sonicaj:


themylogin (approved):

src/middlewared/middlewared/plugins/crypto_/acme_certs_renewal.py
```diff
@@ -0,0 +1,69 @@
+from middlewared.service import private, Service
+
+
+class CertificateService(Service):
+
+    class Config:
+        cli_namespace = 'system.certificate'
+
+    @private
+    async def services_dependent_on_cert(self, id):
```
william-gr, line 10: @sonicaj This seems a *very* weird method and going against how we do things in middleware.
We are making cyrpto plugin aware of external services, instead of providing an interface so external services can register.

Relying in core.get_services and comparing services name seems to be very easy to break without nobody noticing.

PR #6418, opened 2021-02-12
NAS-109169 / 21.04 / HA config upload / factory reset

themylogin:


src/middlewared/middlewared/plugins/config.py
```diff
@@ -206,6 +206,8 @@ def move(src, dst):
 
         self.middleware.call_hook_sync('config.on_upload', UPLOADED_DB_PATH)
 
+        # make sure we includ the uploaded db to be sent to the standby controller
+        upload.insert(0, UPLOADED_DB_PATH)
```
themylogin, line 210: Weird that I did not do that, just replace `shutil.move` with `move` in line 186 above

yocalebo (approved):

yocalebo:
I've QA'ed this PR on SCALE HA system and it works.

PR #6420, opened 2021-02-12
NAS-100503 / 21.04 / Avoid N event subscriptions to run the same code

themylogin:


src/middlewared/middlewared/main.py
```diff
@@ -192,32 +192,14 @@ async def __crash_reporting(self, exc_info):
...
         if ':' in name:
             shortname, arg = name.split(':', 1)
         else:
             shortname = name
             arg = None
-        event_source = self.middleware.get_event_source(shortname)
-        if event_source:
-            for v in self.__event_sources.values():
-                # Do not allow an event source to be subscribed again
-                if v['name'] == name:
```
sonicaj, line 205: Would we now allow users to subscribe to same event source multiple times ?

src/middlewared/middlewared/main.py
```diff
@@ -229,15 +211,13 @@ async def subscribe(self, ident, name):
     async def unsubscribe(self, ident):
         if ident in self.__subscribed:
             self.__subscribed.pop(ident)
-        elif ident in self.__event_sources:
-            event_source = self.__event_sources[ident]['event_source']
-            await self.middleware.run_in_thread(event_source.cancel)
-            self.__event_sources.pop(ident)
+        elif self.session_id + ident in self.middleware.event_source_manager.idents:
```
sonicaj, line 214: Do you think we should perhaps have a method which determines the unique identifier so that we would not have to do `self.session_id + ident` everywhere ?

src/middlewared/middlewared/main.py
themylogin: Yes, as we allow to subscribe to the same event multiple times.

sonicaj (approved):

PR #6426, opened 2021-02-16
NAS-109430 / 12.0 / Improve error handling for directory services

anodos325:
This PR covers a few edge cases that may trip up some users.

First off, we prevent users from simultaneously enabling AD and LDAP directory services. There are very few cases where users actually need to do this. One common example may be FreeIPA + AD in the same environment, but the typical resolution to this configuration need is to create a cross-realm trust between FreeIPA and AD and enable / configure trusted domains in the AD plugin.

Kerberos libraries will let us kinit even if clock offset is larger than 3 minutes, but services behave badly in this situation. This PR makes us try a little harder. If CLDAP ping fails to get us a DC, then we switch to performing normal DNS lookup for a DC. If time offset is too large (or our service account can't be used for netlogon connection), then destroy the service account's kerberos ticket to prevent it from being used by middleware or other processes.

### DIFF
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -521,6 +521,7 @@ class ActiveDirectoryService(ConfigService):
                 self.logger.warning("NTP request to Domain Controller failed.",
                                     exc_info=True)
             except Exception as e:
+                await self.middleware.call("kerberos.stop")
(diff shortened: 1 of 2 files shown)


william-gr (approved):

william-gr:
@anodos325 can I go ahead and merge this?

anodos325:
Yes, let's go ahead and merge

PR #6433, opened 2021-02-17
NAS-109441 / 21.04 / Properly send events for event sources when arguments are specified

sonicaj:
We should properly send events for event sources when arguments are specified. The subscriber relies on complete name instead of short name ( which is event source name ) for actually recieving/handling the event as there are arguments involved and the subscriber will keep track of callbacks based on the complete name ( like our midclt client does ).

### DIFF
--- a/src/middlewared/middlewared/common/event_source/manager.py
+++ b/src/middlewared/middlewared/common/event_source/manager.py
@@ -16,6 +16,20 @@ class EventSourceManager:
         self.idents = {}
         self.subscriptions = defaultdict(lambda: defaultdict(set))
 
+    def short_name_arg(self, name):
+        if ':' in name:
+            shortname, arg = name.split(':', 1)
+        else:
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/common/event_source/manager.py
```diff
@@ -64,7 +64,7 @@ def _send_event(self, name, arg, event_type, **kwargs):
                 self.middleware.logger.trace("Ident %r is gone", ident)
                 continue
 
-            ident_data.app.send_event(ident_data.name, event_type, **kwargs)
+            ident_data.app.send_event(ident_data.full_name, event_type, **kwargs)
```
themylogin, line 67: You can reconstruct it from `name` and `arg`

themylogin (changes requested):

themylogin (approved):

PR #6436, opened 2021-02-17
NAS-109454 / 12.0 / Don't generate corefile alerts on enterprise product

anodos325:
We seem to be generating a fair number of these early on in the free version. We can fix the actionable ones there and avoid creating extra load for support team.

### DIFF
--- a/src/middlewared/middlewared/alert/source/cores.py
+++ b/src/middlewared/middlewared/alert/source/cores.py
@@ -37,6 +37,8 @@ class CoreFilesArePresentAlertSource(ThreadedAlertSource):
             unlink = False
             if IS_FREEBSD and file == "syslog-ng.core":
                 unlink = True
+            elif file == "su.core":
+                unlink = True
             elif os.stat(path).st_mtime < time.time() - 86400 * 5:
                 unlink = True
             else:
--- a/src/middlewared/middlewared/alert/source/cores.py
+++ b/src/middlewared/middlewared/alert/source/cores.py
@@ -16,6 +16,9 @@ class CoreFilesArePresentAlertClass(AlertClass):
 class CoreFilesArePresentAlertSource(ThreadedAlertSource):
     def check_sync(self):
         cores = "/var/db/system/cores"
+        if self.middleware.call_sync("system.is_enterprise"):
+            return
+
         try:
             listdir = os.listdir(cores)
         except Exception:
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/alert/source/cores.py
```diff
@@ -16,6 +16,9 @@ class CoreFilesArePresentAlertClass(AlertClass):
 class CoreFilesArePresentAlertSource(ThreadedAlertSource):
     def check_sync(self):
         cores = "/var/db/system/cores"
+        if self.middleware.call_sync("system.is_enterprise"):
```
themylogin, line 19: If you specify `products = ("CORE", "SCALE")` in the class, this alert source won't be shown in enterprise product at all (and won't be ran too)

themylogin (changes requested):

src/middlewared/middlewared/alert/source/cores.py
anodos325: Thanks. I wasn't aware of that.

anodos325:
I tweaked alert text a little to add a list of corefiles. I also changed up the PR so that old corefiles will be removed post-upgrade so that there's less chance of confusion about whether an issue is new.

anodos325:
Having an alert text with shell command instructions doesn't sit 100% well with me. Perhaps we need to provide a link to documentation on how to tar up and submit core files in a bug ticket.

src/middlewared/middlewared/alert/source/cores.py
```diff
@@ -36,10 +40,10 @@ def check_sync(self):
...
-                has_core_file = True
+                corefiles.append(file)
 
             if unlink:
                 os.unlink(path)
 
-        if has_core_file:
-            return Alert(CoreFilesArePresentAlertClass)
+        if corefiles:
+            return Alert(CoreFilesArePresentAlertClass, {"corefiles": ', '.join(corefiles)})
```
themylogin, line 49: @william-gr 's opinion was: "Not sure if there is any usefulness in listing every file here."

https://github.com/truenas/middleware/pull/6120/files/72c308edfeed2f289fae614f6b6a610cada1f60b#diff-af469e38245d7ba3120168e3b197e205fdfcd036e12b79d85af3cbc7a5f18771

Or has anything changed?

themylogin (changes requested):

src/middlewared/middlewared/alert/source/cores.py
anodos325: I honestly wasn't aware that we had gone down this road before. Big picture, I think having the filenames in the alert will help with triaging.

william-gr: Yes, it seems I was wrong and having them may be useful after all

william-gr:
@themylogin do you mind taking this over and adding su.core to the blacklist? And also automatically adding them to the debug if they are small enough? e.g. < 500KB

william-gr (approved):

PR #6437, opened 2021-02-17
NAS-109456 / 21.04 / deprecate internal usage of system.is_freenas

yocalebo:
Keep the public facing `system.is_freenas` endpoint for backwards compatibility, however remove all the internal usages of it and use the new (correct) `system.is_enterprise` endpoint.

### DIFF
--- a/src/autotune/files/autotune.py
+++ b/src/autotune/files/autotune.py
@@ -21,10 +21,10 @@ import sys
 from middlewared.client import Client
 
 c = Client('ws+unix:///var/run/middlewared-internal.sock')
-if c.call('system.is_freenas'):
-    TRUENAS = False
+if not c.call('system.is_enterprise'):
+    ENTERPRISE = False
 else:
-    TRUENAS = True
+    ENTERPRISE = True
     hardware = c.call('truenas.get_chassis_hardware')
     hardware = hardware.replace('TRUENAS-', '')
     hardware = hardware.split('-')
@@ -111,7 +111,7 @@ DEF_KNOBS = {
 
 
 def guess_vfs_zfs_dirty_data_max_max():
-    if TRUENAS and hardware[0].startswith("M"):
+    if ENTERPRISE and hardware[0].startswith("M"):
         return 12 * GiB
     else:
         return None
@@ -161,7 +161,7 @@ def guess_vfs_zfs_arc_max():
 
     - See comments for USERLAND_RESERVED_MEM.
     """
-    if HW_PHYSMEM_GiB > 200 and TRUENAS:
+    if HW_PHYSMEM_GiB > 200 and ENTERPRISE:
         return int(max(min(int(HW_PHYSMEM * .92),
                        HW_PHYSMEM - (USERLAND_RESERVED_MEM + KERNEL_RESERVED_MEM)),
                        MIN_ZFS_RESERVED_MEM))
@@ -218,28 +218,28 @@ def guess_net_inet_tcp_recvbuf_inc():
 
 
 def guess_vfs_zfs_vdev_async_read_max_active():
-    if TRUENAS and hardware[0] == "Z50":
+    if ENTERPRISE and hardware[0] == "Z50":
         return 64
     else:
         return None
 
 
 def guess_vfs_zfs_vdev_sync_read_max_active():
-    if TRUENAS and hardware[0] == "Z50":
+    if ENTERPRISE and hardware[0] == "Z50":
         return 64
     else:
         return None
 
 
 def guess_vfs_zfs_vdev_async_write_max_active():
-    if TRUENAS and hardware[0] == "Z50":
+    if ENTERPRISE and hardware[0] == "Z50":
         return 64
     else:
         return None
 
 
 def guess_vfs_zfs_vdev_sync_write_max_active():
-    if TRUENAS and hardware[0] == "Z50":
+    if ENTERPRISE and hardware[0] == "Z50":
         return 64
     else:
         return None
--- a/src/middlewared/middlewared/alert/source/smartd.py
+++ b/src/middlewared/middlewared/alert/source/smartd.py
@@ -15,7 +15,7 @@ class SmartdAlertSource(ThreadedAlertSource):
             if self.middleware.call_sync("system.vm"):
                 return
 
-            if not self.middleware.call_sync("system.is_freenas"):
+            if self.middleware.call_sync("system.is_enterprise"):
                 if self.middleware.call_sync("failover.status") != "MASTER":
                     return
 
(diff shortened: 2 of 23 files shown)


src/autotune/files/autotune.py
```diff
@@ -21,7 +21,7 @@
 from middlewared.client import Client
 
 c = Client('ws+unix:///var/run/middlewared-internal.sock')
-if c.call('system.is_freenas'):
+if not c.call('system.is_enterprise'):
     TRUENAS = False
```
william-gr, line 25: Can we also replace TRUENAS with ENTERPRISE? Just to not cause future confusion

src/middlewared/middlewared/alert/source/smartd.py
```diff
@@ -15,7 +15,7 @@ def check_sync(self):
             if self.middleware.call_sync("system.vm"):
                 return
 
-            if not self.middleware.call_sync("system.is_freenas"):
+            if self.middleware.call_sync("system.is_enterprise"):
```
william-gr, line 18: Perhaps failover.licensed is best suited here?

src/middlewared/middlewared/migration/0003_iscsi_vendor.py
```diff
@@ -1,12 +1,11 @@
 async def migrate(middleware):
-    is_freenas = await middleware.call('system.is_freenas')
     extents = await middleware.call('iscsi.extent.query', [['vendor', '=', None]])
     for extent in extents:
         await middleware.call(
             'datastore.update',
             'services.iscsitargetextent',
             extent['id'], {
-                'iscsi_target_extent_vendor': 'FreeNAS' if is_freenas else 'TrueNAS'
+                'iscsi_target_extent_vendor': 'TrueNAS'
```
william-gr, line 8: This may need more thought, we want older systems to keep using FreeNAS for upgrades

src/middlewared/middlewared/alert/source/smartd.py
```diff
@@ -15,9 +15,10 @@ def check_sync(self):
             if self.middleware.call_sync("system.vm"):
                 return
 
-            if not self.middleware.call_sync("system.is_freenas"):
-                if self.middleware.call_sync("failover.status") != "MASTER":
-                    return
+            is_ha = self.middleware.call_sync('failover.licensed')
+            is_master = self.middleware.call_sync('failover.status') == 'MASTER'
+            if is_ha and not is_master:
```
william-gr, line 20: The problem with this change is that now we are making two calls instead of one, if its not HA licensed

william-gr:
Lets get one more review since its very easy to make unnoticeable mistakes in things like this

william-gr (approved):

src/middlewared/middlewared/alert/source/smartd.py
```diff
@@ -15,9 +15,9 @@ def check_sync(self):
             if self.middleware.call_sync("system.vm"):
                 return
 
-            if not self.middleware.call_sync("system.is_freenas"):
-                if self.middleware.call_sync("failover.status") != "MASTER":
-                    return
+            is_ha = self.middleware.call_sync('failover.licensed')
+            if is_ha and not self.middleware.call_sync('failover.status') == 'MASTER':
```
themylogin, line 19: `not self.middleware.call_sync('failover.status') == 'MASTER':` seems weird

Also this file uses `"` so let's keep that consistent

src/middlewared/middlewared/plugins/disk_/sync.py
```diff
@@ -17,11 +17,10 @@ async def sync(self, name):
...
         """
-        if (
-            not await self.middleware.call('system.is_freenas') and
-            await self.middleware.call('failover.licensed') and
-            await self.middleware.call('failover.status') == 'BACKUP'
-        ):
+        is_enterprise = await self.middleware.call('system.is_enterprise')
+        is_ha = await self.middleware.call('failover.licensed')
+        is_backup = await self.middleware.call('failover.status') == 'BACKUP'
+        if is_enterprise and is_ha and is_backup:
```
themylogin, line 23: You are trying to get failover status even on non-enterprise systems or systems without failover licensed

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4147,9 +4140,8 @@ async def __run(self, name, threshold):
         if name == await self.middleware.call('boot.pool_name'):
             pool = await self.middleware.call('zfs.pool.query', [['name', '=', name]], {'get': True})
         else:
-            if not await self.middleware.call('system.is_freenas'):
-                if await self.middleware.call('failover.status') == 'BACKUP':
-                    return
+            if await self.middleware.call('failover.status') == 'BACKUP':
```
themylogin, line 4143: Again, failover status on non-HA system?

themylogin (changes requested):

src/middlewared/middlewared/plugins/disk_/sync.py
william-gr: It should be fine, we return SINGLE on that case

themylogin: @william-gr but that takes an `interface.query`

Maybe we can use this chance to make `failover.status` immediately return `SINGLE` if we are not enterprise or failover is not licensed? And simplify many logic expressions in this diff.

yocalebo: Since code base is merged, there is no reason to make 3 calls here and I can simplify this down to 1 call.

william-gr: Agreed

yocalebo: @themylogin I like your idea, however, after looking at this a little more closely it will cause further complications for certain scenarios. For example, if we don't want to run code on the standby node, we have a check like:
`if failover.status == 'BACKUP': return` which will now have to become `failover.status not in ('MASTER', 'SINGLE'): return`. I'll change this PR to only replace the `is_freenas` calls and I'll open a separate ticket about unifying the `failover` checks in general since it needs to be cleaned up now that all the code is merged into 1 repo.

themylogin: @yocalebo Sorry, I don't understand how `if failover.status == 'BACKUP': return` will change its behavior if we still will be returning `SINGLE`, just doing it faster.

yocalebo: I'm going to keep the changes in this PR to what it was originally set out to do. Simply replace `is_freenas` with `is_enterprise`. I'll open a separate PR for condensing the `failover.*` api calls.

themylogin (approved):

PR #6439, opened 2021-02-17
Fix reporting.realtime crash introduced in 3ff60f8fd876cac83538fe6b15…

themylogin:
Fix reporting.realtime crash introduced in 3ff60f8fd876cac83538fe6b15b58db7cb0bc17e

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/events.py
+++ b/src/middlewared/middlewared/plugins/reporting/events.py
@@ -211,8 +211,6 @@ class RealtimeEventSource(EventSource):
                     'time': time.monotonic(),
                     'speeds': self.get_interface_speeds(),
                 })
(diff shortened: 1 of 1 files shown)


william-gr (approved):

src/middlewared/middlewared/plugins/reporting/events.py
```diff
@@ -211,8 +211,6 @@ def run(self):
                     'time': time.monotonic(),
                     'speeds': self.get_interface_speeds(),
                 })
-            data['interfaces'] = defaultdict(dict)
```
william-gr, line 214: Wait, hoes does this fix a crash?

themylogin: @william-gr was crashing because `defaultdict` is not imported in 12.0

PR #6444, opened 2021-02-17
NAS-109458 / 21.04 / try smartctl SAS to SATA translation in debug

yocalebo:
- on freeBSD, `sysctl -n kern.disks` returned `pmem` and `nvd` devices for which `smartctl` would fail
- convert the `nvd` devices to their respective `nvme` namespace devices using `nvmecontrol nsid` on freeBSD
- on linux, use `lsblk` and a specific inclusion list to, hopefully, only pull the devices that we want
- use the `timeout` command when using the SAS to SATA translation command since certain USB devices can take a long time to respond when sent those commands
- use a `case` statement and only run the `smartctl` commands on the devices that start with the expected driver names

### DIFF
--- a/src/freenas/usr/local/libexec/freenas-debug/smart/smart.sh
+++ b/src/freenas/usr/local/libexec/freenas-debug/smart/smart.sh
@@ -113,7 +113,7 @@ smart_func()
 		fi
 		# double-quotes are important here to
 		# maintain original formatting
-		echo "/dev/$i msg" >> /tmp/smart.out
+		echo "/dev/$i $msg" >> /tmp/smart.out
 		echo "$output" >> /tmp/smart.out
 		echo "" >> /tmp/smart.out
 	done
--- a/src/freenas/usr/local/libexec/freenas-debug/smart/smart.sh
+++ b/src/freenas/usr/local/libexec/freenas-debug/smart/smart.sh
@@ -97,7 +97,7 @@ smart_func()
 	# one involved without doing some extravagant reading of
 	# specific VPD pages from the device itself. Even doing
 	# that is fraught with errors because the interposer could
-	# not trnalsate those pages appropriately.
+	# mistranslate those pages.
 	# So, instead, we'll try to tell smartctl to do the translation
 	# and if it fails (which it will on proper SAS devices) then
 	# we'll try to run it without translation
--- a/src/freenas/usr/local/libexec/freenas-debug/smart/smart.sh
+++ b/src/freenas/usr/local/libexec/freenas-debug/smart/smart.sh
@@ -103,15 +103,17 @@ smart_func()
 	# we'll try to run it without translation
 	for i in $disks
 	do
-    		echo /dev/$i >> /tmp/smart.out
 		# try with translation first
 		output=$(smartctl -a -d sat /dev/$i)
+		msg="(USING TRANSLATION)"
 		if [ $? -ne 0 ]; then
 			# oops try without translation
 			output=$(smartctl -a /dev/$i)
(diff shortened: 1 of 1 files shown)


themylogin (commented):
Can we also write to debug the fact that translation was used?

yocalebo:
> Can we also write to debug the fact that translation was used?

great idea

src/freenas/usr/local/libexec/freenas-debug/smart/smart.sh
```diff
@@ -92,10 +92,30 @@ smart_func()
 		disks=$(sysctl -n kern.disks)
 	fi
 
+	# SAS to SATA interposers could be involed. Unfortunately,
+	# there is no "easy" way of identifying that there is
+	# one involved without doing some extravagant reading of
+	# specific VPD pages from the device itself. Even doing
+	# that is fraught with errors because the interposer could
+	# not trnalsate those pages appropriately.
```
anodos325, line 100: tpyo here

yocalebo: thanks

themylogin (approved):

william-gr:
@amotin do you foresee any problem with this?

william-gr:
@amotin ping

src/freenas/usr/local/libexec/freenas-debug/smart/smart.sh
```diff
@@ -92,10 +92,30 @@ smart_func()
...
+	# mistranslate those pages.
+	# So, instead, we'll try to tell smartctl to do the translation
+	# and if it fails (which it will on proper SAS devices) then
+	# we'll try to run it without translation
 	for i in $disks
 	do
-    		echo /dev/$i >> /tmp/smart.out
-		smartctl -a /dev/$i >> /tmp/smart.out
+		# try with translation first
+		output=$(smartctl -a -d sat /dev/$i)
```
amotin, line 107: What happen if `-d sat` is used with non-SCSI disk like SATA or NVMe?  Probably nothing bad, but still.  Also I worry about using it on USB sticks, since I remember some ugly USB devices hanging from unexpected commands.

yocalebo: In my testing `-a -d sat /dev/*` on a SATA HDD/SSD completed without problems. It failed when talking to SAS or interposer device for which `-a /dev/*` worked.

To answer your question about NVMe, it seems that's never worked quite right because `sysctl -n kern.disks` is returning `nvd*` devices so `-a -d sat /dev/nvd*` or `-a /dev/nvd*` fails because it's expecting the `/dev/nvme*` device. So that will need to be fixed.

src/freenas/usr/local/libexec/freenas-debug/smart/smart.sh
```diff
@@ -81,21 +81,83 @@ smart_func()
...
+	# and if it fails (which it will on proper SAS devices) then
+	# we'll try to run it without translation
+	for i in $disks; do
+		case "$i" in
+		    nvd*)
+			# freebsd returns nvd* devices for the
+			# nvme drives so map to their appropriate
+			# namespace device
+			ns=$(nvmecontrol nsid $i | awk '{print $1}')
+			rc=$?
```
william-gr, line 115: william@DESKTOP-OV98RES[~]: foobar | awk '{print $1}'
zsh: command not found: foobar
william@DESKTOP-OV98RES[~]: echo $?
0

william-gr (approved):

PR #6445, opened 2021-02-18
NAS-109273 / 21.04 / Do not remove encryption keys if pool is not found

sonicaj:
It is possible we have a pool configured but for some mistake/reason the pool did not import like during repair disks were not plugged in and system was booted, in such cases we would like to not to remove the pool's datasets encryption keys from the database.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2021,7 +2021,7 @@ class PoolDatasetService(CRUDService):
         # If we are unable to find the key in database, we see if we have it in memory with the KMIP server, if not,
         # there are 2 ways this can go, we don't retrieve the key or the user can sync KMIP keys and we will have it
         # with the KMIP service again through which we can retrieve them
-        datasets = self.middleware.call_sync('datastore.query', self.dataset_store, filters)
+        datasets = filter_list(self.middleware.call_sync('datastore.query', self.dataset_store), filters)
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2113,8 +2113,22 @@ def sync_db_keys(self, job, name=None):
...
+
+        # It is possible we have a pool configured but for some mistake/reason the pool did not import like
+        # during repair disks were not plugged in and system was booted, in such cases we would like to not
+        # remove the encryption keys from the database.
+        for root_ds in {pool['name'] for pool in self.middleware.call_sync('pool.query')} - {
+            ds['id'] for ds in self.middleware.call_sync(
+                'pool.dataset.query', [], {'extra': {'retrieve_children': False, 'properties': []}}
+            )
+        }:
+            filters.extend([['name', '!=', root_ds], ['name', '!=', f'{root_ds}/']])
```
william-gr, line 2125: Easier/better to have a single filter with `['name', 'nin', list_after_all_this_loop]`' ?

sonicaj: That would mean that we should have the list of encrypted datasets in that specific pool which did not import at the time ?

william-gr: I am not suggesting any logic change. Isn;t what I proposed exactly the same but using a single filter instead of multiple?

william-gr (approved):

PR #6447, opened 2021-02-18
NAS-109435 / 21.04 / Restart UI after some X seconds

sonicaj:
We should restart UI after some X seconds and return from the ui_restart endpoint as if we don't do this, this can result in consumer trying to repeatedly try the same endpoint ( wget ) or curl complaining ( CURL_GOT_NOTHING ) error.

### DIFF
--- a/src/middlewared/middlewared/plugins/system.py
+++ b/src/middlewared/middlewared/plugins/system.py
@@ -1385,12 +1385,15 @@ class SystemGeneralService(ConfigService):
 
         return await self.config()
 
-    @accepts()
-    async def ui_restart(self):
+    @accepts(Int('delay', default=3, validators=[Range(min=0)]))
+    async def ui_restart(self, delay):
         """
         Restart HTTP server to use latest UI settings.
+
+        HTTP server will be restarted after `delay` seconds.
         """
-        await self.middleware.call('service.restart', 'http')
+        event_loop = asyncio.get_event_loop()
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/system.py
```diff
@@ -1385,12 +1387,15 @@ async def do_update(self, data):
...
-    async def ui_restart(self):
+    @accepts(Int('delay', default=3, validators=[Range(min=1)]))
+    async def ui_restart(self, delay):
         """
         Restart HTTP server to use latest UI settings.
+
+        HTTP server will be restarted after `delay` seconds.
         """
-        await self.middleware.call('service.restart', 'http')
+        asyncio.ensure_future(run_after(functools.partial(self.middleware.call, 'service.restart', 'http'), delay))
```
themylogin, line 1397: `loop.call_later`

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1385,12 +1387,15 @@ async def do_update(self, data):
 
         return await self.config()
 
-    @accepts()
-    async def ui_restart(self):
+    @accepts(Int('delay', default=3, validators=[Range(min=1)]))
```
themylogin, line 1390: Why can't we restart UI without a delay if we really want to?

william-gr (approved):

themylogin (approved):

PR #6455, opened 2021-02-19
NAS-109491 / 21.04 / cache if the system is licensed for HA

yocalebo:
Adding/changing a license is an explicit event done by end-user. On an HA system, we should cache the license event to prevent an unnecessary call to read a file from disk (after the 1st time it's read).

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
@@ -200,9 +200,10 @@ class FailoverService(ConfigService):
         """
         if self.HA_LICENSED is None:
             info = self.middleware.call_sync('system.info')
-            if not info['license'] or not info['license']['system_serial_ha']:
+            if info['license'] and info['license']['system_serial_ha']:
+                self.HA_LICENSED = True
+            else:
                 self.HA_LICENSED = False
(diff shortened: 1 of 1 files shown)


william-gr (approved):

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -197,10 +198,13 @@ def licensed(self):
...
         Checks whether this instance is licensed as a HA unit.
         """
-        info = self.middleware.call_sync('system.info')
-        if not info['license'] or not info['license']['system_serial_ha']:
-            return False
-        return True
+        if self.HA_LICENSED is None:
+            info = self.middleware.call_sync('system.info')
+            if not info['license'] or not info['license']['system_serial_ha']:
+                self.HA_LICENSED = False
```
william-gr, line 204: else:

william-gr: Its usually less brainfuck if we check for positive

```
 if info['license']and info['license']['system_serial_ha']:
     self.HA_LICENSED = True
else:
     self.HA_LICENSED = False
````

william-gr (approved):

PR #6456, opened 2021-02-19
NAS-109468 / 21.04 / Change Application Update Alerts

sonicaj:

1) Raise alert that an update is available for a chart release if any of it's container images are outdated.
2) Remove docker image alerts as they cause confusion for the end user i.e alert says plex docker image is outdated but updating docker image only will not mean that the plex chart release is at latest version as well.
3) Do not validate values from `ix_values.yaml` as we don't have their schema and that's something we allow chart devs to control.
4) Concatenate app version and chart version which will be displayed in the UI.
5) Retrieve more information about container images in use by chart releases.

### DIFF
--- a/src/middlewared/middlewared/alert/source/applications_linux.py
+++ b/src/middlewared/middlewared/alert/source/applications_linux.py
@@ -34,8 +34,8 @@ class ChartReleaseUpdateAlertClass(AlertClass, OneShotAlertClass):
 
     category = AlertCategory.APPLICATIONS
     level = AlertLevel.INFO
-    title = 'Chart Release Update Available'
-    text = 'An update is available for "%(name)s" chart release.'
+    title = 'Application Update Available'
+    text = 'An update is available for "%(name)s" application.'
 
     async def create(self, args):
         return Alert(ChartReleaseUpdateAlertClass, args, key=args['id'])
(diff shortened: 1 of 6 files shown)


Ornias1993:
@sonicaj How does this work out with Multi version tags?
Like nextcloud which has tags for:
20 (includes anything 20.*.*)
20.1 (includes anything 20.1.*)
20.1.1 (includes only 20.1.1)

Or apps running "latest", do users still get a way of grabbing the latest update to the multi-version-tag?

sonicaj:
@Ornias1993 this is not specific to what tags are being used. How the update alerts work is that we don't check if there's a newer tag available but rather if the image id changed for the same tag. Why we can't do the former is because there are no rules defined for what versioning scheme a docker image may be using so we can't predict / control the versioning.

What this PR basically does is that it just combines chart release upgrades and chart release container image updates in a single endpoint ( it was there before as well but it refines the process assuming that chart release is not considered to be on the latest version unless the images being used and the chart release version are both not updated ) and introduces a human friendly version identifier.

About using `latest`, that is not good practice and at least for OFFICIAL apps we are introducing changes moving away from that so that we can have sane versioning. If there is a newer image hash available against a specific image tag, the user will be alerted about it.

One more good point about not using `latest` would be that if we are using a tag like `20.1`, it is highly unlikely that the image against that tag will change as that release is out already. So we won't be having container image update involved in such cases and users would be alerted only when the chart version changes with the app version.

Ornias1993:
@sonicaj Hash based, awesome.
I was not 100% sure if it's hash based or tag based, thanks for clearing that up.

And yes, I know the best practices, as you know ;-)

william-gr (approved):

PR #6473, opened 2021-02-22
NAS-107219 / 12.0 / Allow passwordless sudo

themylogin:


william-gr (changes requested):
We also want to allow specific commands to be specified

themylogin:
@william-gr so it'll be a list that can be `["ALL"]`?

william-gr:
How about a list and if its empty means ALL?

src/middlewared/middlewared/etc_files/local/sudoers
```diff
@@ -1,16 +1,35 @@
...
 
 %>\
 % if IS_LINUX:
 root ALL=(ALL:ALL) ALL
 % endif
 % for user in users:
+% if user['sudo_nopasswd']:
+${user['username']} ALL=(ALL) NOPASSWD: ${sudo_commands(user['sudo_commands'])}
+% else:
 ${user['username']} ALL=(ALL) ALL
```
william-gr, line 24: Shouldnt the commands also take effect even without nopasswd?

src/middlewared/middlewared/plugins/account.py
```diff
@@ -180,6 +182,8 @@ async def query(self, filters=None, options=None):
         Bool('microsoft_account', default=False),
         Bool('smb', default=True),
         Bool('sudo', default=False),
+        Bool('sudo_nopasswd', default=False),
+        List('sudo_commands', items=[Str('command', empty=False, validators=[Match('^/[a-zA-Z0-9]')])]),
```
william-gr, line 186: This regex doesnt seem quite right:

```
>>> re.match('^/[a-zA-Z0-9]', '/foo/bar *?@#!')
<re.Match object; span=(0, 2), match='/f'>
```

themylogin: `/foo/bar *?@#!` is a valid `sudoers` file entry

First "world" should be an absolute path, and a few characters are escaped (in mako file)

william-gr: Do you realize its not matching the first entry?

It matched /f, not /foo/bar

themylogin: @william-gr I only want to ensure that command starts with `/` and that the second character is `[a-zA-Z0-9]`. That's the only thing `sudo` wants.

william-gr: The manpage seems to disagree:

A command name is a fully qualified file name which may include shell-style wildcards (see the Wildcards section below). A simple file name allows the user to run the command with any arguments he/she wishes. However, you may also specify command line arguments (including wildcards). Alternately, you can specify "" to indicate that the command may only be run without command line arguments. A directory is a fully qualified path name ending in a '/'. When you specify a directory in a Cmnd_List, the user will be able to run any file within that directory (but not in any sub-directories therein).

themylogin: Do you want to make it also accept `/*`? I don't find any other disagreements.

william-gr: I am looking for sane validation to prevent foot shooting. Making sure it starts with a slash and a letter doesnt help one bit.

If its a full path, I want to make sure that path exists. If its a wildcard, i want to make sure there are matches. If its a directory, I want to make sure that directory exists.

themylogin: > Making sure it starts with a slash and a letter doesnt help one bit.

It helps to prevent `sudo` syntax errors (that will lead to breaking `sudo` for everyone), nothing more

I didn't want a user to be uneditable just because he is allowed to sudo an executable that is on the pool that is not decrypted at this moment, that's why I did not introduce any validations you're proposing. Should I?

william-gr: I believe we should. We will likely have less problems if we properly validate it then let people write anything and end up with non-working configs/

william-gr (approved):
Thanks!

PR #6481, opened 2021-02-23
Send timestamp separately for pod logs

sonicaj:


src/middlewared/middlewared/plugins/kubernetes_linux/pods.py
```diff
@@ -88,7 +90,17 @@ async def run(self):
...
                 ) as stream:
                     async for event in stream:
-                        self.send_event('ADDED', fields={'data': event})
+                        # Event should contain a timestamp in RFC3339 format, we should parse it and supply it
+                        # separately so UI can highlight the timestamp giving us a cleaner view of the logs
+                        timestamp = event.split(maxsplit=1)[0].strip()
+                        try:
+                            timestamp = str(parse(timestamp))
+                        except (TypeError, ParserError):
+                            timestamp = str(datetime.now())
```
themylogin, line 99: Why misinform user? Shouldn't it be `None`?

themylogin (changes requested):

src/middlewared/middlewared/plugins/kubernetes_linux/pods.py
sonicaj: I have not seen this case happen and k8s api should always add a timestamp. If you think we should, I can do that, I was torn as to what should be done in this case.

william-gr (approved):

PR #6483, opened 2021-02-24
NAS-109442 / 21.04 / Properly cleanup/perform ACME challenges

sonicaj:
1) Defines a common interface for all ACME authenticators.
2) Properly validate authenticator schemas.
3) Cleanup TXT records after completing DNS challenges.
4) Add support for both API token and Global API key for Cloudflare.


william-gr (approved):
I didnt review this in too much detail but looks generally OK

Ornias1993:
Bit of a Necro, but non-the-less relevant:

When in the future clustering fully rolls out, be aware of either possible race conditions when 2 nodes try to fetch/update certificates for the same domain or certificate fetching relying on a single node.

It's one of the reasons the non-enterprise version of Traefik doesn't support HA deployments when also managing the certificates.

sonicaj:
@Ornias1993 yes that is accurate. However we would at that time need to redefine how we approach this problem as there are other concerns as well like not using ix volumes. Thanks for the heads up :)

Ornias1993:
@sonicaj 
Mounting certs as secrets and/or configmaps isn't that much of an issue, even if we stop using ix_volumes.
But yes, everything at their time

PR #6493, opened 2021-02-25
NAS-109448 / 21.04 / Properly restart/reload services dependent on ACME certificates

sonicaj:


src/middlewared/middlewared/plugins/chart_releases_linux/cert_attachments.py
```diff
@@ -0,0 +1,25 @@
...
+    async def attachments(self, cert_id):
+        return await self.middleware.call(
+            f'{self.NAMESPACE}.query', await self.get_filters(cert_id), {'extra': {'retrieve_resources': True}}
+        )
+
+    async def redeploy(self, cert_id):
+        chart_releases = [r['name'] for r in await self.attachments(cert_id)]
+        bulk_job = await self.middleware.call(
+            'core.bulk', 'chart.release.redeploy', [[r] for r in await self.attachments(cert_id)]
+        )
```
Ornias1993, line 20: Ahh this is what you meant in our chat. 
Missed it, looking good :)

Ornias1993 (approved):
As I mostly just work with this from the k3s/apps point of view, my approval is on just those codeportions.

src/middlewared/middlewared/common/attachment/certificate.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class CertificateServiceAttachmentDelegate(CertificateAttachmentDelegate, ServiceChangeMixin):
+
+    CERT_FIELD = 'certificate'
+    SERVICE = NotImplementedError
+    SERVICE_VERB = 'reload'
+
+    async def get_namespace(self):
+        return self.SERVICE if self.NAMESPACE == NotImplementedError else self.NAMESPACE
```
william-gr, line 25: I think in theory the correct comparison here would be `is` and not `==`

src/middlewared/middlewared/common/attachment/certificate.py
```diff
@@ -0,0 +1,50 @@
...
+
+    async def state(self, cert_id):
+        config = await self.middleware.call(f'{await self.get_namespace()}.config')
```
william-gr, line 28: When is .config not a dict?

william-gr (approved):

themylogin (approved):
That looks great, we need to do more things like this!

PR #6497, opened 2021-02-25
NAS-109545 / 21.04 / middlewared/iscsi: skip reload after extent creation

rkojedzinszky:
As it is still not referenced, there is no need to reload the service.
With large configurations, this can save time.


william-gr:
ok to test

william-gr:
backport

william-gr (approved):

PR #6503, opened 2021-02-26
NAS-109381 / 21.04 / Send firstboot usage stats

sonicaj:


src/middlewared/middlewared/plugins/usage_/firstboot.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class UsageService(Service):
+
+    FAILED_RETRIES = 3
+
+    class Config:
+        private = True
+
+    async def submit_stats(self, data):
```
william-gr, line 15: Why submit_stats is within the firstboot plugin?

william-gr (approved):

PR #6505, opened 2021-02-26
change app-readme.md from plaintext into markdown

Ornias1993:
app-readme isn't actually a markdown file, as it's loaded as `str.strip`, not `markdown.markdown`

This leads to confusion when creation Charts because chart creators will assume it's rendered as markdown, while it's just plaintext (hence txt).

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/items.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/items.py
@@ -151,7 +151,7 @@ class CatalogService(Service):
         for key, filename, parser in (
             ('chart_metadata', 'Chart.yaml', yaml.safe_load),
             ('schema', 'questions.yaml', yaml.safe_load),
-            ('app_readme', 'app-readme.txt', str.strip),
+            ('app_readme', 'app-readme.md', markdown.markdown),
             ('detailed_readme', 'README.md', markdown.markdown),
         ):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/catalogs_linux/items.py
```diff
@@ -151,7 +151,7 @@ def item_version_details(self, version_path):
         for key, filename, parser in (
             ('chart_metadata', 'Chart.yaml', yaml.safe_load),
             ('schema', 'questions.yaml', yaml.safe_load),
-            ('app_readme', 'app-readme.md', str.strip),
+            ('app_readme', 'app-readme.txt', str.strip),
```
sonicaj, line 154: @Ornias1993 can we please use `markdown.markdown` instead ? We will be making relevant changes on the UI side to accommodate this - why it would be nicer is because it gives us more leverage wrt formatting which the chart devs can control and we can standardize like on chart release upgrade summary as well.

Ornias1993: @sonicaj Ofcoarse thats the logical alternative :)

Ornias1993, line 154:
```suggestion
            ('app_readme', 'app-readme.md', markdown.markdown),
```

sonicaj (approved):
Sweet, thanks!

PR #6521, opened 2021-03-03
NAS-109700 / 21.04 / Support mailcfg[fromname]

MarkBarbieri:


william-gr:
ok to test

themylogin (changes requested):
@MarkBarbieri I think it'll be better to remove that `if mailcfg['fromemail'] != '':` altogether. `mail.send` puts correct `From` header automatically, it just seems to be overwritten with incorrect `From` supplied in `extra_headers`. Can you change that and test on your test case please?

MarkBarbieri:
> @MarkBarbieri I think it'll be better to remove that `if mailcfg['fromemail'] != '':` altogether. `mail.send` puts correct `From` header automatically, it just seems to be overwritten with incorrect `From` supplied in `extra_headers`. Can you change that and test on your test case please?

Good, call.  That works well.

So all this can go.  I'll add a commit to this PR
```
53d52
<         mailcfg = c.call('mail.config')
63,67d61
<
<         if mailcfg['fromemail'] != '':
<             margs['extra_headers'].update({
<                 'From': mailcfg['fromemail']
<             })
```

themylogin (approved):
LGTM, thank you!

PR #6522, opened 2021-03-03
NAS-109638 / 21.04 / call fenced on zpool events

yocalebo:


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/fenced_linux.py
```diff
@@ -1,35 +1,44 @@
...
         except Exception:
-            self.middleware.logger.warning(
-                'Failed to get boot disks for fenced', exc_info=True
-            )
+            self.logger.warning('Failed to get boot disks', exc_info=True)
             # just because we can't grab the boot disks from middleware
             # doesn't mean we should fail to start fenced since it
             # (ultimately) prevents data corruption on HA systems
+            boot_disks = ''
             pass
```
william-gr, line 32: can remove pass now

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/fenced_linux.py
```diff
@@ -40,6 +49,75 @@ def start(self, force=False):
...
+            try:
+                if options.get('reload', False):
+                    os.kill(res['pid'], signal.SIGHUP)
+                if options.get('log_info', False):
+                    os.kill(res['pid'], signal.SIGUSR1)
+            except OSError as e:
+                raise CallError(f'Failed to signal fenced: {e}')
+
+
+async def hook_pool_event(middleware, *args, **kwargs):
```
william-gr, line 101: Just something to consider, its just a matter of personal preference, so no need to change, but for myself I like to return first to avoid deep indentation levels, e.g.:

```
# We only care for SCALE Enterprise
if await middleware.call('system.product_type') != 'SCALE_ENTERPRISE':
    return
    
# HA Licensed handle fenced on their own, we only want mixed Controllers
 if await middleware.call('failover.licensed'):
     return

...
```

yocalebo: Yeah, I like that too. I've changed it.

william-gr (approved):
Thanks

PR #6525, opened 2021-03-03
DOCS-1838 added --dev-test to run only test for development 

ericbsd:
The idea is to create an option --dev-test that will skip all test that is not needed for development.
With pytestmark = pytest.mark.skipif(dev_test, reason=reason) when dev_test is true, all test files that are marked with pytestmark will be skipped for development. If we want to work on a specific file comment the pytestmark line.


q5sys (approved):
LGTM

PR #6537, opened 2021-03-04
DOCS-1833 / 21.04 / Added default testing for SCALE  Kubernetes API

ericbsd:


ericbsd (commented):
LGTM

q5sys (approved):
LGTM

tests/api2/test_273_kubernetes.py
```diff
@@ -0,0 +1,59 @@
...
+    payload = {
+        'pool': pool_name,
+        'route_v4_interface': interface,
+        'route_v4_gateway': gateway,
+        'node_ip': '0.0.0.0'
+    }
+    results = PUT('/kubernetes/', payload)
+    assert results.status_code == 200, results.text
+    job_id = results.json()
+    job_status = wait_on_job(job_id, 180)
```
sonicaj, line 37: This can take quite some as k8s would be initialized from scratch - let's please change the time to 5 minutes as this is just going to be a VM where this would be running i think.

tests/api2/test_273_kubernetes.py
```diff
@@ -0,0 +1,59 @@
...
+    assert results.status_code == 200, results.text
+    assert isinstance(results.json(), dict), results.text
+    assert results.json()[data] == payload[data], results.text
+
+
+def test_04_get_kubernetes_node_ip():
+    results = GET('/kubernetes/node_ip/')
+    assert results.status_code == 200, results.text
+    assert isinstance(results.json(), str), results.text
+    assert results.json() == ip, results.text
```
sonicaj, line 53: How do we ensure `node_ip` is going to equal to `ip` ?
We are binding to `0.0.0.0` so it is free to choose any ip it wants. We should probably select a valid static ip and bind k8s to it to perform this check.

tests/api2/test_273_kubernetes.py
q5sys: I can update that.

tests/api2/test_273_kubernetes.py
q5sys: I'll leave this for Eric to address.

PR #6538, opened 2021-03-04
update wait time

q5sys:
set to 5 min per suggestion.  I know we had an issue with this before with the init time of k8s.

### DIFF
--- a/tests/api2/test_273_kubernetes.py
+++ b/tests/api2/test_273_kubernetes.py
@@ -34,7 +34,7 @@ def test_02_setup_kubernetes(request):
     results = PUT('/kubernetes/', payload)
     assert results.status_code == 200, results.text
     job_id = results.json()
-    job_status = wait_on_job(job_id, 180)
+    job_status = wait_on_job(job_id, 300)
     assert job_status['state'] == 'SUCCESS', str(job_status['results'])
 
 


sonicaj (commented):
Let's please use a static ip for k8s as well if we are going to have the test below for checking node ip

ericbsd:
> Let's please use a static ip for k8s as well if we are going to have the test below for checking node ip

The test already uses a static IP.

sonicaj:
But what if there are multiple ips configured moving on and this test breaks as k8s selects a different IP itself ? So isn't it better to not use `0.0.0.0` in this case ?

ericbsd:
> But what if there are multiple ips configured moving on and this test breaks as k8s selects a different IP itself ? So isn't it better to not use `0.0.0.0` in this case ?

I have battle-tested and that did not happen.

Also, in my opinion, that should be on a ticket and separate PR.

sonicaj:
Your call, but that's just something which is easy to break :)

sonicaj (approved):

sonicaj:
I wonder how / why it would not happen though if there are multiple ips available on the node or are you saying tests only just configure 1 ip and if they at some point do configure 2, this will break sometimes and work sometimes ?

Or are you saying that even with multiple ips it always is going to select one @ericbsd ?

ericbsd:
The systems interface gets setup as static with the IP that ixautomation gave to runtest.py.

As for Kubernetes it only gives me the choice of the vtnet0 interface IP, IP vlan, and 0.0.0.0.

I figured that 0.0.0.0 was fine since I set up the interface with vtnet.

If I am wrong on that logic I will make a ticket for that.

ericbsd (approved):
LGTM

sonicaj:
@ericbsd with `0.0.0.0` selected, it's up to k8s to select any IP available on the node it wants. Which means that at some point if ixautomation configures multiple IPs, the test will work some times and fail some times. So we should not use `0.0.0.0` or not test this. I would go with the former as we should test node ip.

PR #6543, opened 2021-03-04
NAS-109694 / 12.0 / Initial protocol testing for SMB using pylibsmb

anodos325:
Pylibsmb gives more fine-grained control over the SMB operations
performed on the server. Initial commit sets framework for doing these
additional tests as well as implements setting DOS modes through SMB1/2 CREATE.

### DIFF
--- a/tests/api2/smb_protocol.py
+++ b/tests/api2/smb_protocol.py
@@ -199,6 +199,19 @@ def test_010_check_dosmode_create(request, dm):
         assert (to_check & dm.value) != 0, f
 
 
+def test_011_check_dos_ro_cred_handling(request):
+    """
+    This test creates a file with readonly attribute set, then
+    uses the open fd to write data to the file.
+    """
+    depends(request, ["SHARE_IS_WRITABLE"])
+    c = SMB()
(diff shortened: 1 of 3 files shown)


anodos325:
Note: this change will require net/samba412 (or some recent samba) to be installed on the server performing the tests
Minimal smb.conf (or smb4.conf) on the API testing server will be:
```
[global]
client min protocol = NT1
```
To allow python binding to perform SMB1 connections.

anodos325:
Future commits will expand functionality for SMB1/2 tests to include shadow copies and alternate data streams.

william-gr (approved):

tests/api2/smb_protocol.py
```diff
@@ -0,0 +1,299 @@
+#!/usr/bin/env python3
+
+# Author: Eric Turgeon
```
ericbsd, line 3: Should remove this or change the name I did not write the code.

ericbsd (changes requested):

ericbsd (approved):
LGTM

PR #6545, opened 2021-03-04
Add deprecation warning to docs

sonicaj:
This commit adds a deprecation warning to vm service docs outlining that manipulating vm devices via vm service is deprecated and this will be removed in future release as doing this introduces a dependency between vm and vm.device service which opens up weird edge cases especially for display devices.

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/vms.py
+++ b/src/middlewared/middlewared/plugins/vm/vms.py
@@ -1,6 +1,7 @@
 import asyncio
 import errno
 import re
+import warnings
 
 import middlewared.sqlalchemy as sa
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/vm/vms.py
```diff
@@ -101,6 +101,8 @@ async def do_create(self, data):
         `shutdown_timeout` indicates the time in seconds the system waits for the VM to cleanly shutdown. During system
         shutdown, if the VM hasn't exited after a hardware shutdown signal has been sent by the system within
         `shutdown_timeout` seconds, system initiates poweroff for the VM to stop it.
+
+        Specifying `devices` is deprecated and will be removed in coming future release.
```
william-gr, line 105: `SCALE Angelfish: Specifying `devices` is deprecated and will be removed in next major release.`


Also wonder if we should have a deprecation warning in logs everytime that is used.

sonicaj: Logs are for us, i don't think we should have it in the logs ? Thoughts ?

sonicaj: Or mostly us :p ?

william-gr: Do you have a better idea how people will realized an API call is deprecated? Only in docs doesn't sound too good.

Developers often look at middleware log when using the API, I would imagine.

sonicaj: True, i'll make necessary changes. Thank you

src/middlewared/middlewared/plugins/vm/vms.py
```diff
@@ -101,10 +101,18 @@ async def do_create(self, data):
...
         """
         async with LIBVIRT_LOCK:
             await self.middleware.run_in_thread(self._check_setup_connection)
 
+        if data.get('devices'):
+            self.logger.warning(
+                'SCALE Angelfish: Specifying "devices" in "vm.create" is deprecated and will be '
+                'removed in next major release.'
+            )
+
```
william-gr, line 115: Thats usually done in python using warnings module, wonder if it makes sense to use that too

william-gr (approved):

PR #6547, opened 2021-03-05
DOCS-1843 / 21.04 / Changed the setup kubernetes test to use the system static ip.

ericbsd:
Added assert to make sure that the system static ip is present in kubernetes test 01

### DIFF
--- a/tests/api2/test_273_kubernetes.py
+++ b/tests/api2/test_273_kubernetes.py
@@ -37,7 +37,7 @@ def test_02_setup_kubernetes(request):
     results = PUT('/kubernetes/', payload)
     assert results.status_code == 200, results.text
     job_id = results.json()
-    job_status = wait_on_job(job_id, 180)
+    job_status = wait_on_job(job_id, 300)
     assert job_status['state'] == 'SUCCESS', str(job_status['results'])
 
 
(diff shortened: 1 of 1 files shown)


tests/api2/test_273_kubernetes.py
```diff
@@ -31,12 +32,12 @@ def test_02_setup_kubernetes(request):
...
         'route_v4_interface': interface,
         'route_v4_gateway': gateway,
-        'node_ip': '0.0.0.0'
+        'node_ip': ip
     }
     results = PUT('/kubernetes/', payload)
     assert results.status_code == 200, results.text
     job_id = results.json()
-    job_status = wait_on_job(job_id, 300)
+    job_status = wait_on_job(job_id, 180)
```
sonicaj, line 40: Let's please keep this to 300

ericbsd: oh, my master was up-to-date tanks.

ericbsd: done

tests/api2/test_273_kubernetes.py
```diff
@@ -20,6 +20,7 @@ def test_01_get_kubernetes_bindip_choices():
     assert results.status_code == 200, results.text
     assert isinstance(results.json(), dict), results.text
     assert results.json()['0.0.0.0'], results.text
+    assert results.json()[ip], results.text
```
sonicaj, line 23: If ip does not exist, we will have a keyerror. I think we should do “ip in results.json()” to check if it exists

ericbsd: good idea

ericbsd: I will do the same for 0.0.0.0

ericbsd: Done

sonicaj (approved):
Thanks

PR #6572, opened 2021-03-09
DOCS-1853 / 12.1 / SCALE / added pytest.mark.timeout(10) to all mount test

ericbsd:
To avoid mount hang and job timeout.


q5sys (approved):
LGTM

PR #6573, opened 2021-03-09
NAS-109732 / 12.0 / Clear various AD-related caches when service explicitly stopped

anodos325:
A not-insignificant number of users will first try deploying TrueNAS
in a lab / test domain before migrating to the final production domain.
This may be done haphazardly without cleanly leaving the old AD domain.
To ease the migration process, we should clear out middleware user
cache when the AD service is explicitly stopped and flush samba caches.

### DIFF
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -750,6 +750,18 @@ class ActiveDirectoryService(ConfigService):
         await self.middleware.call('etc.generate', 'pam')
         await self.middleware.call('etc.generate', 'nss')
         await self.set_state(DSStatus['DISABLED'])
+        try:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -750,6 +750,18 @@ async def stop(self):
         await self.middleware.call('etc.generate', 'pam')
         await self.middleware.call('etc.generate', 'nss')
         await self.set_state(DSStatus['DISABLED'])
+        try:
+            os.unlink('/var/db/system/.AD_cache_backup')
+        except FileNotFoundError:
+            pass
+        except Exception:
+            self.logger.debug("Failed to remove AD cache backup. Depending on the circumstances, this may result "
```
yocalebo, line 758: Would you not consider this an error? `self.logger.error`

PR #6577, opened 2021-03-09
Add check for IS_FREEBSD in hosts mako config

anodos325:


yocalebo (commented):
Can you access `IS_FREEBSD` if it's not set in the `<% %>` section?

william-gr (approved):

william-gr:
@yocalebo its a global var set in the mako config in etc plugin

yocalebo (approved):

PR #6597, opened 2021-03-10
Ensure clean state before POSIX execute test

anodos325:


ericbsd (approved):
LGTM

sonicaj (commented):
Will it always exist or should we add `-f` to be safe ?

anodos325:
> Will it always exist or should we add `-f` to be safe ?

Sorry. Got click-happy. We're not explicitly checking results from this SSH_TEST and so it should just go on to next test. I'll add a -f when I get to the next failing test.

PR #6600, opened 2021-03-11
NAS-109762 / 21.04 / prevent the ability to share raw disks via iscsi on SCALE

yocalebo:
- defeats the purpose of our product to allow the sharing of raw disks (we expect ZFS to be used underneath the hood)
- SCST has a build time warning about data corruption in certain circumstances. If a raw disk is shared via SCST and if the `sd`, `sg` or `st` drivers are used for accessing/modifying disk state then it can lead to data corruption. On SCALE Enterprise, we use the `sd` and `sg` drivers.

### DIFF
--- a/src/middlewared/middlewared/alembic/versions/12.1/2021-03-15_14-22_remove_rawdisks_from_iscsi.py
+++ b/src/middlewared/middlewared/alembic/versions/12.1/2021-03-15_14-22_remove_rawdisks_from_iscsi.py
@@ -1,0 +1,48 @@
+"""remove raw disks from iscsi
+
+Revision ID: 370ff38939fd
+Revises: 382b7ca9bb51
+Create Date: 2021-03-15 14:22:19.009163+00:00
+
+"""
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = '370ff38939fd'
+down_revision = '382b7ca9bb51'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    conn = op.get_bind()
+    main_table = 'services_iscsitargetextent'
+    main_col = 'iscsi_target_extent_type'
+    tartoext_table = 'services_iscsitargettoextent'
+
+    tar_ids = set()
+    # query db for entries that point to raw disks
+    for disk_id, in conn.execute(f'SELECT id FROM {main_table} WHERE {main_col} = "Disk"').fetchall():
+        # delete the target to extent associations first
+        for id, tar_id in conn.execute(
+            f'SELECT id, iscsi_target_id FROM {tartoext_table} WHERE iscsi_extent_id = {disk_id}'
+        ).fetchall():
+            tar_ids.add(tar_id)
+            conn.execute(f'DELETE FROM {tartoext_table} WHERE id = {id}')
+
+        # delete extent next
+        conn.execute(f'DELETE FROM {main_table} WHERE id = {disk_id}')
+
+    # delete target(s) information last
+    for tar_id in tar_ids:
+        try:
+            conn.execute(f'DELETE FROM services_iscsitargetgroups WHERE iscsi_target_id = {tar_id}')
+            conn.execute(f'DELETE FROM services_iscsitarget WHERE id = {tar_id}')
+        except Exception:
+            # dont fail the transaction for all targets since this can fail on
+            # on foreign key constraint since targets can have multiple assignments
+            continue
+
+
+def downgrade():
+    pass
--- a/src/middlewared/middlewared/plugins/iscsi.py
+++ b/src/middlewared/middlewared/plugins/iscsi.py
@@ -843,7 +843,7 @@ class iSCSITargetExtentService(SharingService):
             return naa
 
     @accepts(List('ignore'))
-    async def disk_choices(self, exclude):
+    async def disk_choices(self, ignore):
         """
         Return a dict of available zvols that can be used
         when creating an extent.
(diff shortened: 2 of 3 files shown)


sonicaj (commented):
I have following concerns:
1) What will happen to users who have already configured disks ?
2) We have more disks logic in extent service, can we remove that as well please because we are not using it anymore ?

src/middlewared/middlewared/alembic/versions/12.1/2021-03-15_14-22_remove_rawdisks_from_iscsi.py
```diff
@@ -0,0 +1,48 @@
...
+
+    tar_ids = set()
+    # query db for entries that point to raw disks
+    for disk_id, in conn.execute(f'SELECT id FROM {main_table} WHERE {main_col} == "Disk"').fetchall():
+        # delete the target to extent associations first
+        for id, tar_id in conn.execute(
+            f'SELECT id, iscsi_target_id FROM {tartoext_table} WHERE iscsi_extent_id == {disk_id}'
+        ).fetchall():
+            tar_ids.add(tar_id)
+            conn.execute(f'DELETE FROM {tartoext_table} WHERE id == {id}')
```
william-gr, line 31: Hmm I never saw SQL queries with `==`

yocalebo: > Hmm I never saw SQL queries with `==`

sqlite> SELECT * FROM services_iscsitargetextent where id == 2;
                                 id = 2
           iscsi_target_extent_name = disk01
         iscsi_target_extent_serial = 5624ba25976f000
           iscsi_target_extent_type = Disk
           iscsi_target_extent_path = {serial}drive-scsi1
       iscsi_target_extent_filesize = 0
      iscsi_target_extent_blocksize = 512
     iscsi_target_extent_pblocksize = 0
iscsi_target_extent_avail_threshold =
        iscsi_target_extent_comment = disk01
            iscsi_target_extent_naa = 0x6589cfc000000857b205f610a35fb3bd
   iscsi_target_extent_insecure_tpc = 1
            iscsi_target_extent_xen = 0
            iscsi_target_extent_rpm = SSD
             iscsi_target_extent_ro = 0
        iscsi_target_extent_enabled = 1
         iscsi_target_extent_vendor = TrueNAS

src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -800,7 +798,8 @@ async def clean_type_and_path(self, data, schema_name, verrors):
         if extent_type is None:
             return data
 
-        if extent_type == 'Disk':
+        # raw disks being shared via iscsi is only supported on freeBSD
+        if osc.IS_FREEBSD and extent_type == 'Disk':
```
sonicaj, line 802: How would this happen even on FreeBSD ?

src/middlewared/middlewared/alembic/versions/12.1/2021-03-15_14-22_remove_rawdisks_from_iscsi.py
themylogin: I did not know that works either. We don't use `==` in SQL queries

src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -819,10 +818,10 @@ async def clean_type_and_path(self, data, schema_name, verrors):
                 raise verrors  # They need this for anything else
 
             if '/iocage' in path:
```
themylogin, line 820: @william-gr can we start removing FreeBSD stuff?

src/middlewared/middlewared/plugins/iscsi.py
yocalebo: Not sure I understand. This would happen if we release a core image based off master repo?

src/middlewared/middlewared/alembic/versions/12.1/2021-03-15_14-22_remove_rawdisks_from_iscsi.py
yocalebo: So are you all asking me to change to `=`?

william-gr: Yes, `==` is not SQL standard as far as I can understand, maybe only works in sqlite3, its not consistent

```
postgres=# select 1=1;
 ?column? 
----------
 t
(1 row)

postgres=# select 1==1;
ERROR:  operator does not exist: integer == integer
LINE 1: select 1==1;
                ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```

src/middlewared/middlewared/plugins/iscsi.py
william-gr: You just removed all Disk from database so this would never be true and its easier/better to remove the code too?

src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -910,8 +852,7 @@ async def disk_choices(self, exclude):
 
         zvol_query_filters = [('type', '=', 'ZVOL')]
         for e in exclude:
-            if e:
-                zvol_query_filters.append(('path', '!=', e))
+            zvol_query_filters.append(('path', '!=', e))
```
william-gr, line 855: Since we are changing this, could probable be
`zvol_query_filters.append(('path', 'nin', exclude))`

thoughts?

william-gr (commented):
Shouldn't we still remove the old DISK entries from database?

yocalebo:
> Shouldn't we still remove the old DISK entries from database?

@william-gr yeah, when I pushed rebased PR I forgot to include migration file 😁  I've fixed. Also, I've updated the docstring for the `disk_choices` method. I really don't like the whole `ignore` parameter and as best as I can tell nothing in middleware (or webui) is using that so I'm inclined to remove it.

william-gr (approved):

themylogin (approved):

PR #6603, opened 2021-03-12
fix _event_system_ready for HA upgrades

yocalebo:
send_event takes 3 positional arguments, not 4.

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
@@ -1864,7 +1864,7 @@ async def _event_system_ready(middleware, event_type, args):
         return
 
     if await middleware.call('keyvalue.get', 'HA_UPGRADE', False):
-        middleware.send_event('failover.upgrade_pending', 'ADDED', id='BACKUP', fields={'pending': True})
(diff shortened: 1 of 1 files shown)


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover.py
```diff
@@ -1864,7 +1864,7 @@ async def _event_system_ready(middleware, event_type, args):
         return
 
     if await middleware.call('keyvalue.get', 'HA_UPGRADE', False):
-        middleware.send_event('failover.upgrade_pending', 'ADDED', id='BACKUP', fields={'pending': True})
+        middleware.send_event('failover.upgrade_pending', 'ADDED', fields={'id': 'BACKUP', 'pending': True})
```
william-gr, line 1867: Thats not direct, I will need some more explanation. What are the consumers of this event?

william-gr: If you mean 

```
  File "/usr/local/lib/python3.8/site-packages/middlewared/main.py", line 1343, in wrap
    await handler(self, event_type, kwargs)
  File "/usr/local/lib/middlewared_truenas/plugins/failover.py", line 1865, in _event_system_ready
    middleware.send_event('failover.upgrade_pending', 'ADDED', {
```

this has already been fixed in 963c3ae442b996a9378c774271a6c3d81503bad2

PR #6608, opened 2021-03-12
NAS-109469 / 21.04 / Add support for rollbacks of PV/PVCs

sonicaj:

1) Make sure all zfs-localpv CRDs are created.
2) Take snapshots of PV/PVCs on chart release upgrades.
3) Rollback PV/PVCs snapshots on chart release rollbacks.
4) Add support for doing a complete recursive rollback of a snapshot.
5) Fixes a bug where we would not rollback chart release volumes on chart release rollback.
6) Fixes an issue where we would not wait for pods to terminate before initiating a rollback which could potentially result in failed rollback.
7) Make sure any allocated PV's are removed on chart release deletion.
8) Explicitly set `Delete` reclaim policy for default storage class to ensure we don't have leftover datasets if users don't use chart release storage class or have deployed manual helm charts / k8s resources.
9) Backup PV's information which are based on a PVC on k8s cluster backup.
10) Restore PV's on k8s cluster restore which were backed up before.
11) Adds validation to ensure we only allow rollback by recreating resources if the resources being consumed by the chart release can be recreated.
12) Fixes an issue where storage classes for each chart release were not created on kubernetes cluster restore.
13) Fixes an issue where complete recursive rollback of `ix-applications` dataset was not happening on kubernetes cluster restore.

To clarify, only those PV/PVC's are restored which are consuming the storage class for their respective namespaces and not the default storage class. How the k8s cluster backup / restore process works wrt PVC restoration is that we keep information in the backup with which PVCs were consuming their namespace storage class and create the relevant storage resources before restoring the chart release themselves and bind the newly created PV on restore to the chart release's PVCs ( although at this time, the PVC has not been created yet because the chart release has not been restored but we are still able to bind a PV to the PVC ). When the chart release's resources are restored, PVC and PV bind themselves correctly which results in correct volume being made available to kubernetes workloads when they start executing.


src/middlewared/middlewared/plugins/chart_releases_linux/rollback.py
```diff
@@ -31,15 +33,22 @@ async def rollback(self, job, release_name, options):
...
 
-        It should be noted that rollback is not possible if a chart release is using persistent volume claims
-        as they are immutable.
-        Rollback is only functional for the actual configuration of the release at the `item_version` specified and
-        any associated `ix_volumes`.
+        `recreate_resources` is a boolean which will delete and then create the kubernetes resources on rollback
+        of chart release. This should be used with caution as if chart release is consuming immutable objects like
+        a PVC, the rollback operation can't be performed and will fail as helm tries to do a 3 way patch for rollback.
+
+        `force` is a boolean which when set will override values provided for `force_rollback` and `recreate_resources`
```
themylogin, line 46: Why are we already having deprecated options in early ALPHA releases?

src/middlewared/middlewared/plugins/chart_releases_linux/rollback.py
```diff
@@ -87,10 +117,14 @@ async def rollback(self, job, release_name, options):
         # TODO: Upstream helm does not have ability to force stop a release, until we have that ability
         #  let's just try to do a best effort to scale down scaleable workloads and then scale them back up
         scale_stats = await self.middleware.call('chart.release.scale', release_name, {'replica_count': 0})
-        job.set_progress(45, 'Scaled down workloads')
+        job.set_progress(45, 'Scaling down workloads')
+
+        await self.middleware.call('chart.release.wait_for_pods_to_terminate', release['namespace'])
+
+        job.set_progress(50, 'Scaled down workloads')
```
themylogin, line 124: Won't `Rolling back chart release` be a better progress indication message?

themylogin (changes requested):

src/middlewared/middlewared/plugins/chart_releases_linux/rollback.py
sonicaj: Yes because we should separate rollback/recreate_resources options. Earlier `force` did both which is not desired.

src/middlewared/middlewared/plugins/chart_releases_linux/rollback.py
sonicaj: That would be wrong as we wouldn't have rolled back chart release. We do that after scaling them down ?

sonicaj: Ah actually, `Rolling` is good :p I'll make the change asap.

src/middlewared/middlewared/plugins/chart_releases_linux/rollback.py
themylogin: Why do we keep it if we have not releases any stable API yet?

themylogin (approved):

PR #6610, opened 2021-03-12
NAS-109790 / 12.0 / Add initial tests for alternate datastream support

anodos325:
Verify that streams can be opened and written over SMB protocol.
In case of large offset test 128KiB, we are validating that the
kernel is allowing us to write large streams as userspace xattrs, which is
not currently supported on linux due to 64KiB size limit in VFS.


ericbsd (approved):
LGTM

william-gr (approved):

PR #6611, opened 2021-03-14
NAS-109728 / 21.04 / Allow running inadyn in daemon mode

sonicaj:
Inadyn init script loads environment variables from /etc/default/inadyn before attempting to start the service and if the env variable dictates not to start it as a daemon, it does not do that. There is no way of overriding this default except for changing it in the default file itself. Correct fix would be to have this fixed upstream and we have an upstream debian bug for this as well ( https://bugs.debian.org/cgi-bin/bugreport.cgi\?bug\=813347 )

### DIFF
--- a/debian/debian/postinst
+++ b/debian/debian/postinst
@@ -1,6 +1,6 @@
 #!/bin/sh -ex
 
-mkdir /var/trash
+mkdir -p /var/trash
 for file in \
     /etc/cron.d/certbot \
     /etc/cron.d/e2scrub_all \
--- a/src/freenas/etc/default/inadyn
+++ b/src/freenas/etc/default/inadyn
@@ -1,0 +1,14 @@
+# Please note, /etc/inadyn.conf should be properly configured before
+# running inadyn
+
+# Set to "yes" if inadyn should run in daemon mode
+# If this is changed to "yes", RUN_IPUP must be set to "no".
+RUN_DAEMON="yes"
+
+# Set to "yes" if inadyn should be run every time a new ppp connection is
+# established. This might be useful, if you are using dial-on-demand.
+RUN_IPUP="no"
+
+# The user and group that inadyn should be run as.
+USER="debian-inadyn"
+GROUP="debian-inadyn"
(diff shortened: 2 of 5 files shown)


themylogin (changes requested):
Just add this file to truenas-files package and the list of overrides in `debian/debian/postinst`

debian/debian/postinst
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh -ex
 
-mkdir /var/trash
+mkdir -p /var/trash
```
themylogin, line 3: Why?

themylogin (approved):

debian/debian/postinst
sonicaj: Because it already exists when `postinst` is called for `truenas` package.
```
+ mkdir /var/trash
mkdir: cannot create directory '/var/trash': File exists
dpkg: error processing package truenas (--configure):
 installed truenas package post-installation script subprocess returned error exit status 1
Errors were encountered while processing:
 truenas
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

src/freenas/etc/default/inadyn
```diff
@@ -0,0 +1,14 @@
...
+# Set to "yes" if inadyn should run in daemon mode
+# If this is changed to "yes", RUN_IPUP must be set to "no".
+RUN_DAEMON="yes"
+
+# Set to "yes" if inadyn should be run every time a new ppp connection is
+# established. This might be useful, if you are using dial-on-demand.
+RUN_IPUP="no"
+
+# The user and group that inadyn should be run as.
+USER="debian-inadyn"
```
william-gr, line 13: Do we have that user?

debian/debian/postinst
themylogin: Thank you, I forgot about that feature of this flag :)

src/freenas/etc/default/inadyn
sonicaj: Yup

sonicaj:
```
❯ cat /etc/passwd | grep debian
debian-inadyn:x:111:118:inadyn dyndns client,,,:/run/inadyn:/bin/false
❯
```

william-gr (approved):

PR #6614, opened 2021-03-15
NAS-109803 / 12.0 / Ensure that domain sid component of groupmap entries updated properly

anodos325:
Wen Samba's netbiosname changes, the server's local SID is automatically
regenerated and reflected in Samba's passdb. Groupmap entries are not
updated in a similar way. This commit simplifies and fixes smb.fix_sid
so that the domain component of each group's SID matches our current
system SID. If it doesn't, then the group_mapping.tdb is removed and
regenerated. The system SID stored in the TrueNAS configuration database
is also updated at this time. A backgrounded groupmap synchronization
job is started when our netbios name is changed in an SMB service update.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb_/sid.py
+++ b/src/middlewared/middlewared/plugins/smb_/sid.py
@@ -68,33 +68,37 @@ class SMBService(Service):
 
     @private
     async def fixsid(self, groupmap=None):
+        """
+        Samba generates a new domain sid when its netbios name changes or if samba's secrets.tdb
+        has been deleted. passdb.tdb will automatically reflect the new mappings, but the groupmap
+        database is not automatically updated in these circumstances. This check is performed when
+        synchronizing group mapping database. In case there entries that no longer match our local
+        system sid, group_mapping.tdb will be removed and re-generated.
+        """
         if groupmap is None:
-            groupmap = (await self.middleware.call('smb.groupmap_list').values())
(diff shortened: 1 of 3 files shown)


anodos325:
regression tests will be added in separate PR linked in Jira.

src/middlewared/middlewared/plugins/smb_/sid.py
```diff
@@ -68,33 +68,37 @@ async def set_system_sid(self, SID):
...
-            self.logger.debug(f"Domain SID in group_mapping.tdb ({domain_SID}) is not SID in nas config ({db_SID}). Updating db")
-            ret = await self.set_database_sid(domain_SID)
-            if not ret:
-                return ret
-            ret = await self.set_system_sid("setlocalsid", domain_SID)
-
-        return ret
+            domain_SID = str(group['SID']).rsplit("-", 1)[0]
+            if not domain_SID.startswith(well_known_SID_prefix) and domain_SID != system_sid:
+                self.logger.error(
```
yocalebo, line 89: This probably isn't an `error` right? Just a warning notice would be sufficient?

yocalebo (approved):

src/middlewared/middlewared/plugins/smb_/sid.py
anodos325: Warning is a good idea. May cause a slight service impact while we resync.

PR #6627, opened 2021-03-16
NAS-109824 / 21.04 / Ensure aclmode set to DISCARD if not using NFSv4 ACLs on SCALE

anodos325:
Having a discard aclmode ensures that users can "undo" an existing
native ZFS ACL through chmod or setfacl. The ZoL POSIX ACL implementation
writes the POSIX ACL in an xattr separate from the native ZFS ACL and
zfs_acl_chmod_setattr() is used to set POSIX mode during setacl and chmod
requests. If aclmode is set to DISCARD, then existing native ZFS ACL is
replaced with one equivalent to specified mode.

This commit also updates aclmode property names to correct, final ones
returned by openzfs.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -691,7 +691,8 @@ class PoolService(CRUDService):
             fsoptions['aclmode'] = 'passthrough'
 
         if osc.IS_LINUX:
-            fsoptions['acltype'] = 'posixacl'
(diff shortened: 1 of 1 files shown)


freqlabs (approved):
This makes sense to me. I will try to push this into ZFS so it is automatic in the future. In the meantime, better safe than sorry.

sonicaj (approved):

PR #6634, opened 2021-03-17
NAS-109601 / 21.04 / Move consolemsg to system.general

themylogin:


src/middlewared/middlewared/plugins/system.py
```diff
@@ -1332,6 +1344,7 @@ async def ui_certificate_choices(self):
             Str('timezone', empty=False),
             Bool('crash_reporting', null=True),
             Bool('usage_collection', null=True),
+            Bool('consolemsg'),
```
william-gr, line 1347: Can we use ui_consolemsg?

src/middlewared/middlewared/plugins/system.py
```diff
@@ -240,9 +240,17 @@ async def do_update(self, data):
         hardware.
 
         When `syslogserver` is defined, logs of `sysloglevel` or above are sent.
+
+        `consolemsg` is a deprecated attribute and will be removed in further releases. Please, use `consolemsg`
+        attribute in the `system.general` plugin.
         """
+        consolemsg = None
+        if 'consolemsg' in data:
```
william-gr, line 248: Wonder if it makes sense to generate a DeprecateWarning

themylogin: To the middleware log? What for, if it is a client issue?

william-gr: We do that elsewhere. Developers may look at our logs. It sounds better than nothing.

william-gr (approved):
Thanks

PR #6636, opened 2021-03-17
NAS-109846 / 12.0 / restart services on failover event concurrently

yocalebo:
- `nfsd` service is known to go catatonic when being restarted if `lockd` and/or `statd` logging is turned on
- we restart the services on failover synchronously and it just so happens that we restart nfs first so all other services can fail to get restarted in this scenario
- restart the services concurrently (since they're self-contained) and specify a timeout (`15` seconds by default)
- log **WAY** more information during the failover event so that I can troubleshoot this a bit easier
- general code clean up

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
@@ -1,24 +1,20 @@
-# Copyright (c) 2020 iXsystems, Inc.
-# All rights reserved.
-# This file is a part of TrueNAS
-# and may not be copied and/or distributed
-# without the express permission of iXsystems.
-from middlewared.utils import filter_list
-from middlewared.service import Service, private
-
 from lockfile import LockFile, AlreadyLocked
 from collections import defaultdict
 import multiprocessing
 import os
-import sqlite3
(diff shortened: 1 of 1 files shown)


william-gr (approved):

rick-mesta:
@themylogin: Hey Vladimir, you think you may have some time to review this PR (as per @william-gr 's request) ?

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
```diff
@@ -124,6 +120,52 @@ def event(self, ifname, vhid, event):
             if refresh:
                 self.middleware.call_sync('failover.status_refresh')
 
+    @private
+    async def restart_service(self, service, timeout, ha=False):
```
themylogin, line 124: Where is `ha` parameter used?

themylogin (approved):

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
yocalebo: Good point, it's not changed anywhere by callers. I could remove that.

PR #6637, opened 2021-03-18
NAS-109849 / 12.0 / prevent failover log spam when remote node goes offline

yocalebo:
- m-series devices can take upwards of 15mins to reboot so spamming `exc_info` every 5 seconds for 15mins (or more depending on what the state of the node is in that's offline) is too noisy and does nothing but muddy the log file
- clarify the verbiage when the `JournalSync` thread is sync'ing db queries to the other node.

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/remote.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/remote.py
@@ -197,6 +197,15 @@ class RemoteClient(object):
         if self._remote_os_version is None:
             try:
                 self._remote_os_version = self.client.call('system.version')
+            except AttributeError:
+                # happens when other node goes offline
+                # (upgrade, reboot, etc) self.client gets
+                # set to None which is expected. However,
+                # the JournalSync thread calls this method
+                # every 5 seconds which is particularly
+                # painful on m-series devices since those
+                # systems can take upwards of 15mins to reboot.
+                pass
             except Exception:
                 logger.error('Failed to determine OS version', exc_info=True)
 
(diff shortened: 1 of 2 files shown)


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/remote.py
```diff
@@ -197,6 +197,15 @@ def get_remote_os_version(self):
         if self._remote_os_version is None:
             try:
                 self._remote_os_version = self.client.call('system.version')
+            except AttributeError:
```
william-gr, line 200: Why its an AttributeError?

yocalebo: `self.client` gets set to `None` when the remote client disconnects.

william-gr: Shoudnt we check for that then first? Instead of catching exception

yocalebo: `self._remote_os_version` gets set to None as well on remote client disconnect but `self.client` gets set to None first so I could move the check for that first.

yocalebo: err, actually no `self._remote_os_version` gets set to `None` first. That's why I'm catching the exception @william-gr

william-gr: I still don't get it, shouldnt matter, as long as check for self.client

william-gr (approved):

PR #6639, opened 2021-03-18
NAS-109780 / 21.04 / Improve error message when creating ACME cert

sonicaj:
This commit adds changes to improve error message so it's clear that one needs to configure root user's email address before trying to create an ACME cert and also raises a validation error if an email is not configured.


yocalebo (approved):

Ornias1993 (approved):
It's still not the best user experience, but thats for another time.
For now this at least (hopefully) prevents more bug reports like NAS-109780 👍

PR #6644, opened 2021-03-18
NAS-109861 / 21.04 / Selectively add locked key

sonicaj:
This commit introduces a change where we only add `locked` key to `pool.dataset.query` output if encryption properties are retrieved.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2867,7 +2867,8 @@ class PoolDatasetService(CRUDService):
                     dataset[i]['value'] = method(dataset[i]['value'])
             del dataset['properties']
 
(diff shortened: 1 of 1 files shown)


themylogin:
What if we don't want any encryption information, why are we still forced to retrieve that?

themylogin (approved):

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2805,6 +2805,10 @@ def query(self, filters, options):
         extra = copy.deepcopy(options.get('extra', {}))
         retrieve_children = extra.get('retrieve_children', True)
         props = extra.get('properties')
+        if props is not None:
```
william-gr, line 2808: This looks odd. Why is this needed?

william-gr (approved):

PR #6648, opened 2021-03-18
Optimize API response times

rkojedzinszky:
Avoid using filtering in code and using expensive extend operations, instead, let sql do the filtering, and extend only the result.

### DIFF
--- a/src/middlewared/middlewared/etc_files/scst.conf.mako
+++ b/src/middlewared/middlewared/etc_files/scst.conf.mako
@@ -5,7 +5,7 @@
 
     global_config = middleware.call_sync('iscsi.global.config')
     targets = middleware.call_sync('iscsi.target.query')
-    extents = {d['id']: d for d in middleware.call_sync('iscsi.extent.query', [['enabled', '=', True]])}
+    extents = {d['id']: d for d in middleware.call_sync('iscsi.extent.query') if d['enabled']}
     portals = {d['id']: d for d in middleware.call_sync('iscsi.portal.query')}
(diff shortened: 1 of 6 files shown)


william-gr:
I am not sure what to think about this approach.

Sometimes we need to filter after we transform the data, having post extend would make things potentially behave differently, two different code paths for the same functionality.

What is the exact performance improvement are we getting here? I am imagining this only starts to play a whole with hundreds of items?

themylogin:
@william-gr I think we can just add an option that will force `CRUDService.query` to use SQL filters even if we have `extend` specified and use it for all `iscsi.extent.query` calls that this PR references.

This:
```
        extent_type = data['type'].upper()
        extent_rpm = data['rpm'].upper()
        data['rpm'] = extent_rpm
        data['type'] = extent_type
```
has to be done in DB migration in order to avoid changing queries.

william-gr:
@themylogin that makes total sense, since these are actually legacy things that were necessary while we still had the django UI.

rkojedzinszky:
Hi, that sounds good. Shall I do it?

rkojedzinszky:
> I am not sure what to think about this approach.
> 
> Sometimes we need to filter after we transform the data, having post extend would make things potentially behave differently, two different code paths for the same functionality.
> 
> What is the exact performance improvement are we getting here? I am imagining this only starts to play a whole with hundreds of items?

In our environment tens of items is enough.

But, we plan to use the nas with hundreds of items, so yes, it is a critical part.

william-gr:
I imagine @themylogin will handle it, unless he is OK with it and you can work together. I will let him respond.

PR #6649, opened 2021-03-19
NAS-109727 / 21.04 / Set replication state error on abnormal zettarepl process termination

themylogin:


src/middlewared/middlewared/plugins/zettarepl.py
```diff
@@ -290,9 +292,11 @@ def stop(self):
         with self.lock:
             if self.process:
                 self.process.terminate()
-                self.process.join(5)
-
-                if self.process.is_alive():
+                for i in range(50):
```
william-gr, line 295: Does linux have a facility to wait for a process? Potentially sending 50 OS signals and another 50 os sleep syscalls could be avoided

william-gr (approved):

PR #6650, opened 2021-03-19
NAS-109545 / 21.04 / iSCSI API speed improvements

themylogin:


src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -672,28 +671,7 @@ async def compress(self, data):
...
-        }
-        disks_names = {}
-        for disk in await self.middleware.call('disk.query'):
-            disks_names[disk['devname']] = disk
-            context['disks'][disk['identifier']] = disk
-
-        for pool in filter(lambda p: context['pools'][p]['disks'], context['pools']):
-            data = context['pools'][pool]
-            if all(disks_names.get(d, {}).get('type') == 'SSD' for d in data['disks']):
-                data['all_flash'] = True
```
william-gr, line 692: I remember the all_flash being useful to something, probably in ctl.conf

william-gr: Meaning, there is probably something similar we will have to do in scst someday

yocalebo (commented):
I left the extent_extend_context in for the reason why CORE uses it. `ctld` has a config option if the zpool is all flash. I left it here just in case we needed it in future on SCST.

william-gr:
Right we should probably leave that, at worse case comment it for now. And/or come up with a way to cache that value

sonicaj:
Well if we are not consuming it, we can just have it freebsd specific

william-gr:
Its just a matter of time of making use of it in SCALE

Also the reporter is targeting CORE

william-gr (approved):

PR #6651, opened 2021-03-19
ZFSDatasetService: use id optimization in more cases

rkojedzinszky:
For example, if called from PoolDatasetService.query, where filter list contains expressions to exclude system datasets.

### DIFF
--- a/src/middlewared/middlewared/plugins/zfs.py
+++ b/src/middlewared/middlewared/plugins/zfs.py
@@ -504,8 +504,17 @@ class ZFSDatasetService(CRUDService):
                 props=props, top_level_props=top_level_props, user_props=user_properties, snapshots=snapshots,
                 retrieve_children=retrieve_children,
             )
-            if filters and len(filters) == 1 and list(filters[0][:2]) == ['id', '=']:
-                kwargs['datasets'] = [filters[0][2]]
+            # Process all filter elements, and allow one distinct value for `id`.
+            dataset = None
+            for filter in filters:
+                if len(filter) == 3 and list(filter[:2]) == ['id', '=']:
+                    if dataset is not None and dataset != filter[2]:
+                        return []
+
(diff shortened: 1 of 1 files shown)


william-gr:
Thanks for the patch but I am curious where you saw an usage like that.

Proper usage would probably be `["id", "in", ["dataset1", "dataset2"]]`

rkojedzinszky:
@william-gr i did not mean such invocations, and even, my solution is broken. The intent is to use `datasets` filter even if more than one filter is passed to speed up lookups. I'll force-push a more correct fix.

rkojedzinszky:
For example, this is called when an iSCSIextent is created, and in some cases, extra filters may be added to exclude system datasets. In that case, no optimization was made.

sonicaj (commented):
@rkojedzinszky can you please clarify your use case ? I am curious how you are planning on consuming this change especially since `zfs.dataset` is a private service ?

rkojedzinszky:
> @rkojedzinszky can you please clarify your use case ? I am curious how you are planning on consuming this change especially since `zfs.dataset` is a private service ?

You are right, in master branch this might not be needed anymore. In 12-stable, pool.dataset.query adds more filter elements, which is handled different, that may cause all datasets to be retrieved. Then, I'll close this PR, but in 12-stable, it might be relevant.

sonicaj:
That shouldn't be the case anymore in `truenas/12.0-stable` branch, please let me know otherwise. Thank you

rkojedzinszky:
> That shouldn't be the case anymore in `truenas/12.0-stable` branch, please let me know otherwise. Thank you

Sure, I probably missed https://github.com/truenas/middleware/commit/8b5e322fa1792b1744a821056dc20b3bb2953fda. I was working on a tree pulled from a working truenas-12-u2, and there it was still needed.

PR #6654, opened 2021-03-20
NAS-109855 / 21.04 / Add support for creating wildcard ACME certs

sonicaj:
This commit fixes issues with the implementation allowing wildcard certificate to be created.


Ornias1993 (approved):
Thanks for picking this up this fast.
Interesting how simple some of the problems/changes turned out to be :)

william-gr (approved):

PR #6655, opened 2021-03-20
NAS-109898 / 21.04 / PoolDatasetService.do_create: skip children lookups on parent datasets

rkojedzinszky:
This will speed up create operations with high number of sibling datasets.


william-gr:
ok to test

sonicaj (approved):

PR #6656, opened 2021-03-21
NAS-109192 / 21.04 / Properly load nvidia kernel modules

sonicaj:
This commit adds changes to properly load and configure nvidia modules so it can be consumed by applications.


Ornias1993 (approved):
Ahh, yes the missing modprobe...
Pretty KISS fix :)

william-gr (approved):

PR #6657, opened 2021-03-22
NAS-109545 / 12.0 / Prevent nested loop queries in iscsi.target

themylogin:


src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -1128,6 +1125,15 @@ async def extend(self, data):
             )
         return data
 
+    @private
+    async def extend_context(self, extra):
+        return {
+            'groups': await self.middleware.call(
```
sonicaj, line 1131: Do you think we should maybe have this normalized here so that we don't pop entries from groups each time at L1113 ?

themylogin: Nice catch, thank you!

sonicaj (approved):

themylogin:
backport

PR #6664, opened 2021-03-23
NAS-109910 / 12.0 / Add unique constraints

rkojedzinszky:
Add missing unique constraints


src/middlewared/middlewared/alembic/versions/12.0/2021-03-21_09-49_iscsi_target_extent_name_unique.py
```diff
@@ -0,0 +1,32 @@
...
+revision = 'd015631e2929'
+down_revision = '90b815426c10'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('services_iscsitargetextent', schema=None) as batch_op:
+        batch_op.create_unique_constraint('iscsi_target_extent_name_unique', ['iscsi_target_extent_name'])
```
themylogin, line 22: Thank you for your contribution! This migration will fail if existing table rows do not match the constraint. Although I understand that due to existing middleware validations it is highly unlikely, we must nevertheless at least rename such rows as failure here will prevent TrueNAS from upgrading and leave user with a broken system.

themylogin (changes requested):

src/middlewared/middlewared/alembic/versions/12.0/2021-03-21_09-49_iscsi_target_extent_name_unique.py
rkojedzinszky: You are totally right. But then, this should be a longer discussion. If for whatever reason two records exist with the same name, what is the expected operation now, and after the change? I saw uniqueness check in the code, that is why I added it. Is there a way in the upgrade procedure to do some pre-validation before the actual upgrade process? Perhaps, that would be the most reasonable way. So, for example if I rename iscsi targets, then it could affect/break services.

themylogin: Services probably never worked correctly with duplicate names, our goal here is only to avoid crashes on update.

rkojedzinszky: Ok, got it.

PR #6671, opened 2021-03-24
NAS-109907 / 21.04 / Add listen directive for wireguard ip for TC

sonicaj:
This commit adds changes where we add a listen directive in nginx for TC, motivation for this change is that when nginx is listening on some specific ip and not a wildcard one, TC instance is unable to connect to nginx as it uses the wireguard ip of the NAS to do so and nginx refuses connections on that.

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
+++ b/src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
@@ -1,5 +1,6 @@
 <%
     import contextlib
+    import ipaddress
     import os
 
     from cryptography.hazmat.backends import default_backend
@@ -55,6 +56,10 @@
                 'wildcard': '::',
             })
     ip6_list = [f'[{ip}]' for ip in ip6_list]
+
+    wg_config = middleware.call_sync('datastore.config', 'system.truecommand')
+    if wg_config['enabled'] and wg_config['wg_address']:
+        ip4_list.append(ipaddress.ip_network(wg_config['wg_address'], False).network_address)
 
     ip_list = ip4_list + ip6_list
 
--- a/src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
+++ b/src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
@@ -58,7 +58,7 @@
     ip6_list = [f'[{ip}]' for ip in ip6_list]
 
     wg_config = middleware.call_sync('datastore.config', 'system.truecommand')
-    if wg_config['enabled'] and wg_config['wg_address']:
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
```diff
@@ -56,6 +57,10 @@
             })
     ip6_list = [f'[{ip}]' for ip in ip6_list]
 
+    wg_config = middleware.call_sync('datastore.config', 'system.truecommand')
+    if wg_config['enabled'] and wg_config['wg_address']:
+        ip4_list.append(ipaddress.ip_network(wg_config['wg_address'], False).network_address)
```
william-gr, line 62: Dont you have to reload nginx after wireguard starts?

What happens if you start nginx and the ip is not there yet in the wireguard interface?

william-gr: Also for wireguard, we probably always want to listen on 80/443

william-gr (approved):

src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
```diff
@@ -56,6 +57,10 @@
             })
     ip6_list = [f'[{ip}]' for ip in ip6_list]
 
+    wg_config = middleware.call_sync('datastore.config', 'system.truecommand')
```
william-gr, line 60: why datastore and not truecommand.config?

src/middlewared/middlewared/plugins/truecommand/wireguard.py
```diff
@@ -135,5 +137,6 @@ async def start_truecommand_service(self):
 
     @private
     async def stop_truecommand_service(self):
+        await self.middleware.call('service.reload', 'http')
```
william-gr, line 140: Dont you need to do that after you stop truecommand?

src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
sonicaj: Because we don't expose `wg_address` in `truecommand.config`

src/middlewared/middlewared/plugins/truecommand/wireguard.py
sonicaj: At this point, TC connection status has already shifted, so it's safe to do it here. We can do afterwards as well but it won't change much ?

src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
william-gr: Why would it be in datastore and not it .config?

PR #6677, opened 2021-03-25
NAS-109927 / 21.04 / Do not have catalog query fail

sonicaj:
This commit adds changes to ensure we do not have catalog.query failing if we have a repo configured which does not exist or we are unable to clone for some reason.


themylogin (approved):

Ornias1993 (approved):
Awesome, definately hit this issue before during testing.
Neat fix!
LGTM!

PR #6681, opened 2021-03-25
NAS-109782 / 21.04 / Run DHCP in parallel

themylogin:


src/middlewared/middlewared/plugins/network.py
```diff
@@ -1925,13 +1925,13 @@ async def sync(self, wait_dhcp=False):
...
-                await self.sync_interface(interface, wait_dhcp, sync_interface_opts[interface])
-            except Exception:
-                self.logger.error('Failed to configure {}'.format(interface), exc_info=True)
+        vlans, non_vlans = bisect(lambda x: x.startswith('vlan'), filter(lambda i: not i.startswith('br'), interfaces))
+        if non_vlans:
+            await asyncio.wait([self._sync_interface_wrapper(interface, wait_dhcp, sync_interface_opts[interface])
+                                for interface in non_vlans])
+        if vlans:
+            await asyncio.wait([self._sync_interface_wrapper(interface, wait_dhcp, sync_interface_opts[interface])
+                                for interface in vlans])
```
william-gr, line 1934: Wont we remain with the same problem here? waiting dhcp on an unconfigured interface before we set up bridges?

themylogin: @william-gr all this code is executed only for configured interfaces. The interface with no carrier was explicitly configured for DHCP.

william-gr: Same question applies

william-gr: If a DHCP times out for an interface that is not related to bridge, bridge will only be configured after timeout of that unrelated interface, no?

william-gr (approved):

PR #6684, opened 2021-03-26
NAS-109944 / 21.04 / Allow specifying default identifier field for tables

sonicaj:
This commit adds changes to allow specifying custom default identifier field. Before even if id attribute was added in extend, that would work but with recent changes that does not work as we are forcing the use of sql filters to speed up retrieval.

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/update.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/update.py
@@ -31,6 +31,7 @@ class CatalogService(CRUDService):
         datastore = 'services.catalog'
         datastore_extend = 'catalog.catalog_extend'
         datastore_extend_context = 'catalog.catalog_extend_context'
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/catalogs_linux/update.py
```diff
@@ -31,6 +31,7 @@ class Config:
         datastore = 'services.catalog'
         datastore_extend = 'catalog.catalog_extend'
         datastore_extend_context = 'catalog.catalog_extend_context'
+        datastore_default_identifier = 'label'
```
themylogin, line 34: What's a `default identifier`? What other identifiers do we have? Can `datastore_primary_key` be a better name?

themylogin (changes requested):

themylogin (approved):

PR #6688, opened 2021-03-28
NAS-109949 / 21.04 / Bring back cgroupsv2

sonicaj:
This commit adds changes to have scale use cgroups2. With newer upstream version of k3s, we have cgroupsv2 support fixed and working so we bring cgroupsv2 back.


themylogin (approved):

sonicaj:
Nvidia does not like cgroupsv2, made an issue with them. So let's keep using it for now.

PR #6689, opened 2021-03-29
Only query related datasets locked status in SharingTaskService

themylogin:


src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -474,24 +494,40 @@ def query(self, filters=None, options=None):
 
         with libzfs.ZFS() as zfs:
             # Handle `id` filter specially to avoiding getting all datasets
-            if filters and len(filters) == 1 and list(filters[0][:2]) == ['id', '=']:
+            if filters and filters[0][0] == 'id' and filters[0][1] in ['=', 'in']:
```
william-gr, line 497: `['id', 'in']`

william-gr: No, nvm

william-gr (commented):
This seems way more intrusive than what I was hoping for 12.0, unfortunately

themylogin:
@william-gr unfortunately, this is even more intrusive that it seems, libzfs behavior is different for `zfs.get_dataset(...).__getstate__())` and `zfs.datasets_serialized`, we are now fixing at least one difference we've found. I was thinking maybe we backport this to SCALE (performance improvement is really very good) and in 12.0 fix one specific iscsi extent endpoint @rkojedzinszky is having performance issues with?

william-gr:
@themylogin I agree

rkojedzinszky:
@themylogin @william-gr thank you for your great efforts.

PR #6690, opened 2021-03-29
NAS-109914 / 21.04 / FUSE mnt/umnt gluster volumes on start/stop/delete

yocalebo:
Adding localhost to the `glustereventsd` daemon proved worthless. So this commit does a few things.

- remove the localhost:6000/_clusterevents from the `glustereventsd` daemon (if it's been added)
- repurpose localhost:6000/_clusterevents endpoint to receive custom events sent by middleware explicitly
- protect localhost:6000/_clusterevents endpoint by using JWT encoding. The JWT will be encoded using a secret key that is sent to each peer in the trusted storage pool by TC. This will be done "out of band" and stored on each peer in the gluster metadata directory.
- on gluster volume start/stop or delete, FUSE mnt/umnt the associated mountpoint and then send a request to each peer in the trusted storage pool to do the same. The reason why we do this is because when volumes were stopped/deleted previously, glusterd only sent a local event (not a cluster event) for these events. So this meant that the peer who issued the command would FUSE mnt/umnt appropriately, but the other peers would not. In the case of a delete/stop, leaving behind a stale FUSE mountpoint proved to be troublesome.
- add a public api for uploading a JWT secret key to be used for sending events to each peer in the TSP. This is sitting behind our normal authentication.

### DIFF
--- a/src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
+++ b/src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
@@ -93,8 +93,8 @@ class CtdbSharedVolumeService(Service):
         # to all the other peers in the TSP to also
         # FUSE mount it
         data = {'event': 'VOLUME_START', 'name': CTDB_VOL_NAME, 'forward': True}
-        j = await self.middleware.call('gluster.localevents.send', data)
-        await j.wait(raise_error=True)
+        job = await self.middleware.call('gluster.localevents.send', data)
+        await job.wait(raise_error=True)
 
         return await self.middleware.call('gluster.volume.query', [('name', '=', CTDB_VOL_NAME)])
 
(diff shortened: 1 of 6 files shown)


src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
```diff
@@ -89,11 +89,12 @@ async def create(self, job):
...
-        # try to mount it locally
-        mount_job = await self.middleware.call(
-            'gluster.fuse.mount', {'name': CTDB_VOL_NAME, 'raise': True}
-        )
-        await mount_job.wait(raise_error=True)
+        # try to mount it locally and send a request
+        # to all the other peers in the TSP to also
+        # FUSE mount it
+        data = {'event': 'VOLUME_START', 'name': CTDB_VOL_NAME, 'forward': True}
+        j = await self.middleware.call('gluster.localevents.send', data)
```
william-gr, line 96: Its pretty minor, but `job` instead of `j` probably make the code just a little bit more readable :P

william-gr (approved):

PR #6705, opened 2021-03-31
NAS-109973 / 21.04 / Fix Linux disk LUN scanning

SuperQ:
* Fix nvme detection, use correct variable for detecting transport.
* Fix USB detection, use the GPT UUID as the LUN.

Signed-off-by: Ben Kochie <superq@gmail.com>

### DIFF
--- a/src/middlewared/middlewared/plugins/device_/device_info_linux.py
+++ b/src/middlewared/middlewared/plugins/device_/device_info_linux.py
@@ -192,10 +192,15 @@ class DeviceService(Service, DeviceInfoBase):
 
             # get lunid
             if disk_data['wwn']:
-                if disk.get('tran') == 'nvme':
+                if disk_data['tran'] == 'nvme':
                     disk['lunid'] = disk_data['wwn'].lstrip('eui.')
                 else:
                     disk['lunid'] = disk_data['wwn'].lstrip('0x')
+            # USB disks can use the GUID from the partition table as a uniqe ID.
+            if disk_data['tran'] == 'usb' and disk_data['pttype'] == 'gpt':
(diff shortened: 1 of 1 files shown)


william-gr:
ok to test

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -192,10 +192,15 @@ def get_disk_details(self, block_device, disk, disks_data):
 
             # get lunid
             if disk_data['wwn']:
-                if disk.get('tran') == 'nvme':
+                if disk_data['tran'] == 'nvme':
                     disk['lunid'] = disk_data['wwn'].lstrip('eui.')
                 else:
                     disk['lunid'] = disk_data['wwn'].lstrip('0x')
+            # USB disks can use the GUID from the partition table as a uniqe ID.
+            if disk_data['tran'] == 'usb' and disk_data['pttype'] == 'gpt':
```
william-gr, line 200: That will only work if the disk is partitioned.

What if you add 4 new unformatted disks to your USB enclosure?

SuperQ: This is a intentional, since TrueNAS doesn't have a physical disk adoption process compared to other more standard OS distributions. You have to prep disks with GPT ahead of time.

william-gr: We don't think this is a reasonable solution.

GPT header gets wiped as soon as a pool is created with the disks.

How about we use the long serial if its USB?

SuperQ:
I'll leave this to you to figure out a solution. I'm just going to install regular Debian and manage it with Ansible.

PR #6707, opened 2021-03-31
NAS-109545 / 12.0 / Improve performance when extending zero or one iSCSI extent

themylogin:


sonicaj (approved):

src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -495,6 +496,52 @@ async def sharing_task_determine_locked(self, data, locked_datasets):
...
+                about_to_lock_dataset = None
+
+            if len(result) == 1 and result[0]['type'] == 'DISK' and about_to_lock_dataset is None:
+                # Special optimized case
+                try:
+                    dataset = await self.middleware.call(
+                        'zfs.dataset.query',
+                        [['id', '=', result[0]['disk'][5:]]],
+                        {
+                            'extra': {'properties': ['encryption', 'keystatus']},
```
sonicaj, line 526: Specifying `properties` has no effect in 12.0 because for single `id` case this is not respected as we are not using `datasets_serialized` endpoint.

PR #6717, opened 2021-04-01
NAS-109522 / 21.04 / Remove minio configuration directory

sonicaj:
Remove minio configuration directory each time minio is started. We need to do this so that given minio credentials are accepted as otherwise we would need to maintain old/new credentials so minio can encrypt it's config using new credentials. After some discussion, we decided it's best that we remove minio configuration directory for now, drawback with this is that policies/users of minio won't stick.


src/middlewared/middlewared/etc_files/local/minio/configure.py
```diff
@@ -41,3 +40,15 @@ def render(service, middleware):
             f.write(cert['privatekey'])
         os.chown(minio_privatekey, minio_uid, minio_gid)
         os.chmod(minio_privatekey, 0o600)
+
+
+def configure_minio_sys_dir(s3):
+    storage_path = s3['storage_path']
+    minio_dir = os.path.join(storage_path, '.minio.sys')
+    shutil.rmtree(minio_dir, ignore_errors=True)
```
william-gr, line 48: So we remove that dir everytime? Cant that cause trouble?

sonicaj: Nope. Minio recognises buckets by the directories and does not save them in the `.minio.sys` configuration directory.

william-gr (approved):

PR #6718, opened 2021-04-02
NAS-109903 / 21.04 / Correctly show type of identifier in rest api docs

sonicaj:
This PR fixes an issue where we rendered the acceptable type of a CRUD endpoint as `array`. For example, the endpoint
`pool/dataset/{id}/` would show that the accepted parameter type of `id` is an `array`.

### DIFF
--- a/src/middlewared/middlewared/plugins/jail_freebsd.py
+++ b/src/middlewared/middlewared/plugins/jail_freebsd.py
@@ -143,6 +143,7 @@ def common_validation(middleware, options, update=False, jail=None, schema='opti
 class PluginService(CRUDService):
 
     class Config:
+        datastore_primary_key_type = 'string'
         cli_private = True
 
     @accepts()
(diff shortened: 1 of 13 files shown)


src/middlewared/middlewared/plugins/jail_freebsd.py
```diff
@@ -143,6 +143,7 @@ def common_validation(middleware, options, update=False, jail=None, schema='opti
 class PluginService(CRUDService):
```
themylogin, line 143: Let's just remove this file altogether

themylogin (approved):

src/middlewared/middlewared/plugins/jail_freebsd.py
sonicaj: Agreed, i can do that in a separate PR.

PR #6719, opened 2021-04-02
NAS-109925 / 21.06 / Allow specifying extra arguments with id keyword in rest api

sonicaj:

1) Updates documentation to not specify filterable arguments like `sort` when using `id` keyword in rest api endpoint.
2) Allows specifying `query-options.extra` arguments when using `id` keyword.
3) Fixes an issue which correctly outlines that `query-options.extra` is supported in request body with `id` keyword in rest api endpoint and not the complete `filterable` request body.

### DIFF
--- a/src/middlewared/middlewared/restful.py
+++ b/src/middlewared/middlewared/restful.py
@@ -497,9 +497,30 @@ class Resource(object):
                     filterid = kwargs['id']
                     if filterid.isdigit():
                         filterid = int(filterid)
+                    extra = {}
+                    text = await req.text()
+                    if text:
+                        data = await req.json()
+                        if not isinstance(data, dict):
+                            resp.set_status(400)
+                            resp.body = json.dumps({'message': 'Endpoint expects object/dict.',})
+                            return resp
+                        if data:
+                            if (
+                                'query-options' not in data or len(data['query-options']) > 1
+                                or (data['query-options'] and 'extra' not in data['query-options'])
+                            ):
+                                resp.set_status(400)
+                                resp.body = json.dumps({
+                                    'message': 'Only "query-options.extra" supported with "id" parameter.',
+                                })
+                                return resp
+                            elif data.get('query-options', {}).get('extra'):
+                                extra = data['query-options']['extra']
+
                     method_args = [
                         [(self.service_config['datastore_primary_key'], '=', filterid)],
-                        {'get': True, 'force_sql_filters': True}
+                        {'get': True, 'force_sql_filters': True, 'extra': extra}
                     ]
                 else:
                     method_args = self._filterable_args(req)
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/restful.py
```diff
@@ -497,9 +511,31 @@ def get_post_args(self, req, resp, **kwargs):
                     filterid = kwargs['id']
                     if filterid.isdigit():
                         filterid = int(filterid)
+                    extra = {}
+                    text = await req.text()
```
themylogin, line 515: Are we reading request body of a GET method?

themylogin (changes requested):

src/middlewared/middlewared/restful.py
sonicaj:
```
 if not method_args:
                # RFC 7231 specifies that a GET request can accept a payload body
                # This means that all the http methods now ( delete, get, post, put ) accept a payload body
```
Yes

themylogin: That does not make it a good request semantic. You can accept any JSON variable just fine using query parameters `/items/query-options.extra.include_chart_schema=true&query-options.extra.retrieve_resources=false`

themylogin (changes requested):

src/middlewared/middlewared/restful.py
```diff
@@ -227,7 +227,12 @@ def add_path(self, path, operation, methodname, service_config):
                         'schema': {'type': 'string'},
                     },
                 ]
-            elif accepts and not (operation == 'delete' and method['item_method'] and len(accepts) == 1):
+            elif '{id}' in path and method['filterable']:
+                desc = f'{desc}\n\n' if desc else ''
+                opobject['description'] = desc + '`query-options.extra` can be specified as query parameters.'
```
themylogin, line 232: Can't they be specified for e.g. `GET /storage/pool/`, why a special case?

themylogin (approved):
Thank you!

src/middlewared/middlewared/restful.py
```diff
@@ -497,9 +518,13 @@ def get_post_args(self, req, resp, **kwargs):
                     filterid = kwargs['id']
                     if filterid.isdigit():
                         filterid = int(filterid)
+                    extra = {}
+                    for key, val in list(req.query.items()):
```
themylogin, line 522: Should it be `extra.` here too?

themylogin (changes requested):

src/middlewared/middlewared/restful.py
sonicaj: I don't think so because we are not expecting anything else here ?

themylogin (approved):

PR #6726, opened 2021-04-05
NAS-110029 / 21.04 / add system hash to public api

yocalebo:


src/middlewared/middlewared/plugins/system_/hostid.py
```diff
@@ -0,0 +1,20 @@
...
+
+class SystemHostIdService(Service):
+    hostid = None
+
+    class Config:
+        namespace = 'system.host'
+
+    @accepts()
+    def id(self):
+        if self.hostid is None:
```
sonicaj, line 14: Let's please add a docstring.

sonicaj (commented):
Would we be extending `system.host` service ? If not, how about just having it under `system` namespace ?

src/middlewared/middlewared/plugins/system_/hostid.py
yocalebo: sure, no problem

sonicaj (approved):
Thanks!

PR #6727, opened 2021-04-05
NAS-110028 / 21.06 / Do not use cached version of refs 

sonicaj:
This PR adds changes to not use cached version of refs. This fixes issues where if system state changed like certificates were added, we would still be having old/cached version, we normalize the questions again to make sure this is not the case.


Ornias1993 (approved):
Can confirm the underlaying issues....
It looks as a good fix for 21.06 :)

Definately can't go into 21.04, because the chance of it accidentally breaking on of the refs is significant.

william-gr (approved):

PR #6728, opened 2021-04-05
NAS-110031 / 12.0 / make fallback to ix-netif work

yocalebo:
It's come to my attention that we've had multiple customers fail the upgrade from 11.3 to 12.0 on their HA systems. This was able to be reproduced in-house. Upon investigation of the internal system, I found that we failed back to using `ix-netif` rc script but that failed because the `ha_*` related rc routines in `rc.freenas` have been removed in 12. This commit does the following:

- fix `hadetect.py` to work since it was broken in the first place
- add `ha_hardware`, `ha_node`, and `ha_licensed` to `rc.freenas` to be used by `ix-netif`
- fix multiple issues in `ix-netif` rc script which is called as a fallback if the `interface.sync` fails

### DIFF
--- a/src/freenas/etc/ix.rc.d/ix-netif
+++ b/src/freenas/etc/ix.rc.d/ix-netif
@@ -12,8 +12,6 @@
 
 _interface_config()
 {
-	local failover_status="" failover_licensed=0
-
 	local saved_ifs="${IFS}"
 	local IFS=\|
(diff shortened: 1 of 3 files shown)


yocalebo:
I used the wrong branch name so bugclerk isn't picking it up 🤦 Closing to open a fresh PR using correct branch name.

PR #6729, opened 2021-04-05
NAS-110032 / 12.0 / make fallback to ix-netif work

yocalebo:
It's come to my attention that we've had multiple customers fail the upgrade from 11.3 to 12.0 on their HA systems. This was able to be reproduced in-house. Upon investigation of the internal system, I found that `interface.sync` failed so we failed back to using `ix-netif` rc script but that failed because the `ha_*` related rc routines in `rc.freenas` have been removed in 12. This commit does the following:

- fix `hadetect.py` to work since it was broken in the first place
- make `hadetect.py` standalone and not require middlewared to be running
- add `ha_hardware`, `ha_node`, and `ha_licensed` to `rc.freenas` to be used by `ix-netif`
- fix multiple issues in `ix-netif` rc script

### DIFF
--- a/src/freenas/etc/rc.freenas
+++ b/src/freenas/etc/rc.freenas
@@ -90,3 +90,18 @@ ro_sqlite()
 	cp ${FREENAS_CONFIG} ${ret}
 	echo ${ret}
 }
+
+ha_hardware()
+{
+	/usr/local/bin/hadetect | jq -r '.hardware'
+}
+
+ha_node()
+{
+	/usr/local/bin/hadetect | jq -r '.node'
+}
+
+ha_licensed()
+{
+	/usr/local/bin/hadetect | jq -r '.licensed'
+}
--- a/src/middlewared/middlewared/scripts/hadetect.py
+++ b/src/middlewared/middlewared/scripts/hadetect.py
@@ -1,16 +1,97 @@
-#!/usr/bin/env python
-import importlib
+#!/usr/bin/env python3
 
-spec = importlib.util.spec_from_file_location(
-    'failover', '/usr/local/lib/middlewared_truenas/plugins/failover.py',
-)
-failover = importlib.util.module_from_spec(spec)
-spec.loader.exec_module(failover)
+import subprocess
+import re
+import json
+import glob
+
+from licenselib.license import License
+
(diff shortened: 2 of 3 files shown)


src/middlewared/middlewared/scripts/hadetect.py
```diff
@@ -1,16 +1,104 @@
...
+                        result['node'] = node.group(1)
+                        break
+
+                elif re.search(XSERIES, encstat, re.M):
+                    # puma (X-series)
+                    result['hardware'] = 'PUMA'
+
+                    # now get node position
+                    smp = subprocess.run(
+                        ['/sbin/camcontrol', 'smpphylist', enclosure, '-q'],
```
william-gr, line 46: you mean enclosure.name?

src/middlewared/middlewared/scripts/hadetect.py
```diff
@@ -1,16 +1,104 @@
...
+XSERIES = 'Enclosure Name: CELESTIC (P3215-O|P3217-B)'
+MSERIES = 'Enclosure Name: (ECStream|iX) 4024S([ps])'
+LICENSE = '/data/license'
 
 
 def main():
-    print(':'.join(str(i) for i in failover.FailoverService._ha_mode()))
+    result = {'hardware': 'MANUAL', 'node': '', 'licensed': False}
+    try:
+        for enclosure in pathlib.Path('/dev').iterdir():
```
william-gr, line 18: Might be simpler to use glob... glob.glob('/dev/ses*')

william-gr (approved):

src/freenas/etc/rc.freenas
```diff
@@ -90,3 +90,18 @@ ro_sqlite()
 	cp ${FREENAS_CONFIG} ${ret}
 	echo ${ret}
 }
+
+ha_hardware()
+{
+	/usr/local/bin/python3 $(find /usr/local -name hadetect.py -type f) | jq -r '.hardware'
```
william-gr, line 96: hadetect is an entry point in middlewared setup.py... It should be just available in /usr/local/bin instad of requiring us to use find every time?

yocalebo: true

william-gr (approved):

PR #6731, opened 2021-04-05
NAS-110039 / 21.04 / Ignore encoding errors in stdout/stderr

darkfiberiru:
When calling lsblk apparently disks can have non utf-8 data in at least labels and potentially more spots. By adding error=ignore to subprocess.run we can decode stdout earlier and toss junk data instead of causing exception on decoding it during json.loads later. Maybe needed in other spots in middlewared

### DIFF
--- a/src/middlewared/middlewared/plugins/device_/device_info_linux.py
+++ b/src/middlewared/middlewared/plugins/device_/device_info_linux.py
@@ -100,7 +100,8 @@ class DeviceService(Service, DeviceInfoBase):
         disks_cp = subprocess.run(
             ['lsblk', '-OJdb'],
             stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE
+            stderr=subprocess.PIPE,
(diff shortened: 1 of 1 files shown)


darkfiberiru:
Without this patch a bad label on a disk can make TrueNAS scale believe a system has no disks.

yocalebo (approved):

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -100,7 +100,8 @@ def retrieve_disks_data(self):
         disks_cp = subprocess.run(
             ['lsblk', '-OJdb'],
             stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE
+            stderr=subprocess.PIPE,
+            errors="ignore"
```
william-gr, line 104:
```suggestion
            errors='ignore'
```

We use single quotes in this whole file

PR #6737, opened 2021-04-06
NAS-110053 / 21.04 / fix nvme drive detection on SCALE

yocalebo:


themylogin (approved):

themylogin:
@yocalebo  how does this fix anything?

yocalebo:
> @yocalebo how does this fix anything?

`disk` object is defined in `DeviceInfoBase` which does not have a `tran` key. The `disk_data` is the object built from `lsblk`.

themylogin:
@yocalebo sorry, I did not see that variable name was changed too

PR #6745, opened 2021-04-07
NAS-110046 / 21.04 / Make sure we keep polling udev

sonicaj:
This commit fixes an issue where in case udev errored out, we stopped polling udev. We want to keep polling udev to continue monitoring the system state. About the current issue, it was an error where system failed to poll the file descriptor.

### DIFF
--- a/src/middlewared/middlewared/plugins/device_/udev_events_linux.py
+++ b/src/middlewared/middlewared/plugins/device_/udev_events_linux.py
@@ -1,5 +1,6 @@
 import pyudev
 import subprocess
+import time
 
 from middlewared.service import private, Service
 from middlewared.utils import run, start_daemon_thread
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/device_/udev_events_linux.py
```diff
@@ -19,8 +19,16 @@ def udev_events(middleware):
...
+    while True:
+        # We always want to keep polling udev, let's log what error we are
+        # seeing and fix them accordingly as we see them
+        try:
+            for device in iter(monitor.poll, None):
+                middleware.call_hook_sync(
+                    f'udev.{device.subsystem}', data={**dict(device), 'SYS_NAME': device.sys_name}
+                )
+        except Exception:
+            middleware.logger.error('Polling udev failed', exc_info=True)
```
william-gr, line 31: Could we run into an infinite busy loop here?

e.g. monitor.poll for some reason always raise an exception like EOFError or StopIteration

Perhaps we should add a sleep here

william-gr (approved):

PR #6749, opened 2021-04-07
NAS-110099 / 21.04 / Make sure required crds are setup before considering k8s node to be ready

sonicaj:
This PR fixes an issue where in some cases kubernetes can take a few seconds to create/apply the CRDs we use and this results in issues as we consider the crds to be available.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
@@ -73,6 +73,23 @@ class KubernetesService(Service):
         ]
 
     @private
+    async def ensure_k8s_crd_are_available(self):
+        required_crds = [
+            'volumesnapshots.snapshot.storage.k8s.io',
+            'volumesnapshotcontents.snapshot.storage.k8s.io',
+            'volumesnapshotclasses.snapshot.storage.k8s.io',
+            'zfsrestores.zfs.openebs.io',
+            'zfsbackups.zfs.openebs.io',
+            'zfssnapshots.zfs.openebs.io',
+            'zfsvolumes.zfs.openebs.io',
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
```diff
@@ -72,6 +72,23 @@ async def iptable_rules(self):
...
+            'zfsrestores.zfs.openebs.io',
+            'zfsbackups.zfs.openebs.io',
+            'zfssnapshots.zfs.openebs.io',
+            'zfsvolumes.zfs.openebs.io',
+            'network-attachment-definitions.k8s.cni.cncf.io',
+        ]
+        while len(
+            await self.middleware.call('k8s.crd.query', [['metadata.name', 'in', required_crds]])
+        ) < len(required_crds):
+            await asyncio.sleep(5)
```
william-gr, line 90: I assume apply_yaml_file would ensure these?

But I wonder whether anything does wrong here, we will loop infinitely?
Should we add a max retry count?
Also, perhaps a TODO to make it react on an event? poll like this makes me twitch :D

sonicaj: `apply_yaml` should have taken care of this but I saw a case where it did not in time take care of it which resulted in `post_start` failing and we didn't consider the node to be healthy because of `post_start`'s early exit.
I think i can add a max retry count here for now.
The upstream k3s issue has also been diagnosed due to which we actually started using `apply_yaml` and it's an issue with both k3s and zfs-localpv. I'll be fixing that appropriately and we would be probably removing these altogether but that would not be for 21.04.

william-gr (approved):

PR #6750, opened 2021-04-08
NAS-109961 / 21.04 / Do not start collectd immediately during boot because it can hang

themylogin:
It happened only for Rehaf Yosaf, I was only able to repeat it once out of many attempts. Also yesterday I've seen one indirect evidence of this happening in an unrelated user debug  If I trace it, it does not hang. Harmless sleep is the only sane option IMO.


src/freenas/etc/systemd/system/collectd.service.d/override.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
```
william-gr, line 1: Does this file accept comments? Due to nature of this it probably grants a couple lines explaining why this was needed, and link to a couple tickets

william-gr (approved):

PR #6755, opened 2021-04-09
NAS-110133 / 21.04 / Do not show a successful status for cloud sync that was not executed …

themylogin:
…due to the dataset being locked


themylogin:
@william-gr I think that should probably be merged for 21.04

william-gr (approved):

PR #6763, opened 2021-04-10
Ignore files and some folders in trains folder

Ornias1993:
As an addition to:
https://github.com/truenas/catalog_validation/pull/19

Currently a train folder can only contain folder and no files.
However: it would be nice ignoring files in the train folder completely, because they are totally irrelevant.

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/items.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/items.py
@@ -68,8 +68,11 @@ class CatalogService(Service):
 
         for train in filter(lambda c: os.path.exists(os.path.join(location, c)), trains):
             category_path = os.path.join(location, train)
+            
             for item in filter(lambda p: os.path.isdir(os.path.join(category_path, p)), os.listdir(category_path)):
                 item_location = os.path.join(category_path, item)
+                if (not os.path.isdir(item_location) or train.startswith('.')):
+                    continue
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/catalogs_linux/items.py
```diff
@@ -68,8 +68,11 @@ def get_trains(self, location, options=None):
 
         for train in filter(lambda c: os.path.exists(os.path.join(location, c)), trains):
             category_path = os.path.join(location, train)
+
```
Ornias1993, line 71:
```suggestion

```

src/middlewared/middlewared/plugins/catalogs_linux/items.py
```diff
@@ -70,6 +70,8 @@ def get_trains(self, location, options=None):
             category_path = os.path.join(location, train)
             for item in filter(lambda p: os.path.isdir(os.path.join(category_path, p)), os.listdir(category_path)):
                 item_location = os.path.join(category_path, item)
+                if (not os.path.isdir(item_location) or train.startswith('.')):
```
sonicaj, line 73: This is not required as the `filter` above at L71 already ensures that at this stage `item_location` is a directory. Please let me know otherwise.

sonicaj: ping @Ornias1993

PR #6764, opened 2021-04-11
NAS-110159 / 21.04 / Connect POSIX ACL logic to nfs4xdr_winacl

anodos325:
This commit brings in proper behavior for recursively applying POSIX ACLs and also adds no-xdev capabilities for associated permissions changes.

Since this more-or-less finishes POSIX ACL implementation, more thorough validation is being added with this commit.

### DIFF
--- a/src/middlewared/middlewared/plugins/filesystem_/acl_linux.py
+++ b/src/middlewared/middlewared/plugins/filesystem_/acl_linux.py
@@ -2,8 +2,9 @@ import errno
 import os
 import subprocess
 import stat as pystat
+from pathlib import Path
 
-from middlewared.service import private, CallError, Service
+from middlewared.service import private, CallError, ValidationErrors, Service
 from middlewared.plugins.smb import SMBBuiltin
 from .acl_base import ACLBase, ACLDefault, ACLType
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/filesystem_/acl_linux.py
```diff
@@ -12,82 +13,102 @@ class FilesystemService(Service, ACLBase):
...
+        if p.is_file() and recursive:
+            verrors.add(f'{schema}.path', 'Recursive operations on a file are invalid.')
+            return
 
         if not os.path.realpath(path).startswith('/mnt/'):
-            raise CallError(f"Changing permissions on paths outside of /mnt is not permitted: {path}",
-                            errno.EPERM)
+            verrors.add(
+                f'{schema}.path',
+                f"Changing permissions on paths outside of /mnt is not permitted: {path}"
```
yocalebo, line 46:
```suggestion
                f"Changing permissions on paths outside of /mnt are not permitted: {path}"
```

src/middlewared/middlewared/plugins/filesystem_/acl_linux.py
```diff
@@ -12,82 +13,102 @@ class FilesystemService(Service, ACLBase):
...
-                                       check=False, capture_output=True)
-            else:
-                chown = subprocess.run(['chown', '-R', f'{uid}:{gid}', data['path']],
-                                       check=False, capture_output=True)
+            return
 
-            if chown.returncode != 0:
-                raise CallError(f"Failed to chown [{data['path']}]: "
-                                f"{chown.stderr.decode()}")
+        if uid == -1 and gid == -1:
```
yocalebo, line 80: What does this mean in the context of this method?

anodos325: Passing -1 as an id to chown(2) leaves owner / group same. I think it probably would make more sense in this case to raise a validation error since it doesn't make sense to leave both unchanged.

src/middlewared/middlewared/plugins/filesystem_/acl_linux.py
anodos325: I don't english real good apparently.

anodos325: Okay. Took me a second to parse what I wrote. In this case the gerund "changing" is the subject of sentence and therefore singular form of verb is correct, but the overall construction is awkwardly phrased.

yocalebo: Something something, English are hard...

yocalebo (approved):
Cool, thanks!

PR #6770, opened 2021-04-15
NAS-110161 / 21.06 / Properly stop glusterd to avoid crash on system dataset unmount

themylogin:


yocalebo (approved):
Seems to be an upstream debian/systemd issue but this should work.

src/middlewared/middlewared/plugins/service_/services/glusterd.py
```diff
@@ -27,3 +31,14 @@ async def before_stop(self):
 
     async def after_stop(self):
         await self.middleware.call('service.stop', 'glustereventsd')
+
+        # systemd[1]: glusterd.service: Unit process 1376703 (glusterfsd) remains running after unit stopped.
+        # systemd[1]: glusterd.service: Unit process 1376720 (glusterfs) remains running after unit stopped.
+        # This prevents from tank/.system/ctdb_shared_vol from being unmounted
+        futures = []
+        for process in psutil.process_iter(attrs=['cmdline']):
```
william-gr, line 39: Running synchronous code within the asyncio loop. Isn't there a more direct/smart way to do this other than loop through all processes?

yocalebo: Gluster has a PID directory but now you're having to do IO on the asyncio loop which isn't any better. At this point it might just be worth it to `pkill -9 -f gluster`...

PR #6779, opened 2021-04-16
NAS-109843 / 12.0 / Report total disk statistics instead of per-disk

themylogin:


rick-mesta (approved):
This LGTM - still would like a second opinion from the more seasoned middleware folk -- to get a warm fuzzy 🤗 ... @yocalebo / @sonicaj  ... if either of you have the time

src/middlewared/middlewared/plugins/reporting/iostat.py
```diff
@@ -54,7 +56,31 @@ def _read(self):
...
     def read(self):
-        return self.stats
+        read_ops = 0
+        read_bytes = 0
+        write_ops = 0
+        write_bytes = 0
+        busy = 0
+        count = 0
+        for disk, current in self.current.items():
+            prev = self.prev.get(disk)
```
sonicaj, line 66: Why do we keep track of previous disk stats ?

sonicaj (approved):

PR #6780, opened 2021-04-17
NAS-110231 / 21.04 / fix replacebrick api endpoint

yocalebo:
Though the 3rd party python module allows `force` as true or false, gluster only allows `force` as true.

### DIFF
--- a/src/middlewared/middlewared/plugins/gluster_linux/volume.py
+++ b/src/middlewared/middlewared/plugins/gluster_linux/volume.py
@@ -394,7 +394,6 @@ class GlusterVolumeService(CRUDService):
             Str('peer_path', required=True),
             required=True,
         ),
-        Bool('force', default=False),
     ))
     async def replacebrick(self, data):
         """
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -416,7 +414,7 @@ async def replacebrick(self, data):
         new_brick = new['peer_name'] + ':' + new['peer_path']
 
         method = volume.bricks.replace_commit
-        options = {'args': (data.pop('name'), src_brick, new_brick), 'kwargs': data}
+        options = {'args': (data.pop('name'), src_brick, new_brick), 'kwargs': {'force': True}}
```
sonicaj, line 417: Do we need to specify it if it's only supposed to be `true` ?

sonicaj (approved):

src/middlewared/middlewared/plugins/gluster_linux/volume.py
yocalebo: I can make a comment

yocalebo:
I've pushed multiple bug fixes upstream that have been merged. I'm going to close this PR and open a single PR with all the fixes.

PR #6783, opened 2021-04-19
NAS-110222 / 21.04 / system.info physical_cores

themylogin:


yocalebo (approved):

yocalebo (changes requested):
I just tested this on a dual-socket 8core CPU machine and I `cpu_count(logical=False)` is giving me a count of `2` which is the total number of CPU's I have. Each CPU is 8 physical cores. I tested this on freeBSD.

themylogin:
@yocalebo it is a documented behavior, it only works properly on Linux, we won't backport this to 12.0

yocalebo (approved):

william-gr:
Merging this as being required for UI regression work.

PR #6788, opened 2021-04-19
NAS-110267 / 21.06 / Do not consider bond interface as an internal interface

sonicaj:


themylogin (approved):

themylogin:
@sonicaj why was it considered an internal interface?

PR #6789, opened 2021-04-20
NAS-108336 / 21.06 / Fixate snapshot removal date when updating periodic snapshot tasks

themylogin:


src/middlewared/middlewared/plugins/snapshot_/removal_date.py
```diff
@@ -0,0 +1,23 @@
...
+from middlewared.service import Service, job, private
+
+
+class PeriodicSnapshotTaskService(Service):
+
+    class Config:
+        namespace = "pool.snapshottask"
+
+    @private
+    async def removal_date_property(self):
```
yocalebo, line 12: Probably should make this synchronous since it can block the event loop.

yocalebo: Also, can you use the `system.host_id` method which basically does the same thing?

yocalebo (approved):

PR #6793, opened 2021-04-21
NAS-110209 / 21.06 / Move rrdcached journal and nslcd cache from boot pool

themylogin:


ChrisCheney:
I added the above patch to my 21.02 install and noticed that /tmp/cache/nscd didn't exist after reboot but that the symlink to it did. I suspect /tmp was not yet mounted by the time the middleware code ran running the mkdir on /tmp/cache/nscd.

themylogin:
@ChrisCheney thank you! systemd was removing everything from /tmp after middleware startup, we fixed that

themylogin:
@yocalebo can you please take a look at this?

yocalebo (commented):
Do you know how large these directories can grow? We have had past problems where `tmpfs` will be filled up to 100% and then lots of other things begin to break.

themylogin:
@yocalebo I can't think of a case where `nscd` cache will grow past a few megabytes

yocalebo:
> @yocalebo I can't think of a case where `nscd` cache will grow past a few megabytes

Sure, I'm mostly wondering about rrdcached.

themylogin:
@yocalebo rrdcached journal is being moved to the system dataset

yocalebo (approved):

PR #6799, opened 2021-04-22
NAS-110156 / 21.06 / Update ttl of cloudflare/route53 txt records

sonicaj:
This commit introduces changes to follow certbot settings for setting ttl value of cloudflare/route53 to 10 minutes respectively.

### DIFF
--- a/src/middlewared/middlewared/plugins/acme_protocol_/authenticators/cloudflare.py
+++ b/src/middlewared/middlewared/plugins/acme_protocol_/authenticators/cloudflare.py
@@ -49,7 +49,7 @@ class CloudFlareAuthenticator(Authenticator):
         verrors.check()
 
     def _perform(self, domain, validation_name, validation_content):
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/acme_protocol_/authenticators/cloudflare.py
```diff
@@ -49,7 +49,7 @@ def validate_credentials(data):
         verrors.check()
 
     def _perform(self, domain, validation_name, validation_content):
-        self.get_cloudflare_object().add_txt_record(domain, validation_name, validation_content, 3600)
+        self.get_cloudflare_object().add_txt_record(domain, validation_name, validation_content, 120)
```
themylogin, line 52: Should it be 120 here? Commit message is saying 10 minutes

themylogin (approved):

PR #6800, opened 2021-04-22
NAS-110174 / 12.0 / WBC_ERR_WINBIND_NOT_AVAILABLE is not an SSH login failure

themylogin:


anodos325:
In the bug ticket case, it was application ` pam_winbind(su)` (su application using pam) logging to syslog that generated the SSH alert. Not a message from sshd. I'm okay with this if we think it will prevent SSH alerts from being generated by non-SSH sources, but it might be better to see if we can make it more explicitly related to `sshd`.

themylogin:
@anodos325 you are absolutely right, fixed

anodos325 (approved):

PR #6802, opened 2021-04-22
NAS-110243 / 21.06 / Allow endpoints to specify explicit http methods

sonicaj:
This is useful when we want to maintain backwards compatibility with endpoints which were not expecting payload before but are now and users still would like to consume them with previous method which would be GET whereas it's POST now.

### DIFF
--- a/src/middlewared/middlewared/plugins/system.py
+++ b/src/middlewared/middlewared/plugins/system.py
@@ -7,7 +7,8 @@ from middlewared.i18n import set_language
 from middlewared.logger import CrashReporting
 from middlewared.schema import accepts, Bool, Dict, Int, IPAddr, List, Str
 from middlewared.service import (
-    CallError, ConfigService, no_auth_required, job, pass_app, private, Service, throttle, ValidationErrors
+    CallError, ConfigService, no_auth_required, job, pass_app, private, rest_api_metadata,
+    Service, throttle, ValidationErrors
 )
 import middlewared.sqlalchemy as sa
 from middlewared.utils import Popen, run, start_daemon_thread, sw_buildtime, sw_version, osc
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/system.py
```diff
@@ -1470,6 +1471,7 @@ async def do_update(self, data):
 
         return await self.config()
 
+    @rest_api_metadata(explicit_methods=['get'])
```
william-gr, line 1474: Can we allow post, too?

Also maybe it should be uppercase

william-gr: We dont want in U4 for it to break again ;)

sonicaj: `post` is already allowed ? This is going to add other `explicit` methods apart from the default one which is going to be configured automatically based on the endpoint's `accepts` like other endpoints.

sonicaj: In this case the default one would be `post` and `get` is explicitly added. I can make it upper case :)

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1470,6 +1471,7 @@ async def do_update(self, data):
 
         return await self.config()
 
+    @rest_api_metadata(explicit_methods=['GET'])
```
william-gr, line 1474: That sounds a little bit confusing to me then...
We are saying explicit method is GET, but it also works for POST

From my point of view that would be the list of methods i would want this to work

additional methods, or extra methods might be a better name... or have it be a list of all methods it will accept

sonicaj: You are right, we should reword it because it is confusing otherwise. I'll make the required changes.

william-gr:
I would remove the "explicit", but thats just me :P

sonicaj:
Done :D

william-gr (approved):
Thanks

PR #6805, opened 2021-04-23
move failover files to common plugins dir

yocalebo:


yocalebo:
ping @truenas/middleware

PR #6806, opened 2021-04-23
Fix numbering in SMBHAMODE enum

anodos325:


yocalebo (approved):

anodos325:
This issue causes HA in some configurations to be mis-identified as clustered. Apparently, I don't count real good either.

PR #6813, opened 2021-04-26
NAS-108576 / 21.06 / Allow having attributes only editable on creation

sonicaj:
This PR adds changes to have attributes which are only editable once and later cannot be modified. This helps with applications where some docker images only take values on initialization which cannot be changed later.

### DIFF
--- a/src/middlewared/middlewared/schema.py
+++ b/src/middlewared/middlewared/schema.py
@@ -41,8 +41,8 @@ class EnumMixin(object):
         self.enum = kwargs.pop('enum', None)
         super(EnumMixin, self).__init__(*args, **kwargs)
 
-    def clean(self, value, old_value=NOT_PROVIDED):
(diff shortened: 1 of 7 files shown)


src/middlewared/middlewared/schema.py
```diff
@@ -41,8 +41,8 @@ def __init__(self, *args, **kwargs):
         self.enum = kwargs.pop('enum', None)
         super(EnumMixin, self).__init__(*args, **kwargs)
 
-    def clean(self, value):
-        value = super().clean(value)
+    def clean(self, value, old_value=NOT_PROVIDED):
```
themylogin, line 44: Can we find another way to implement this without messing with the `Attribute` interface?

themylogin (changes requested):

themylogin (approved):

PR #6820, opened 2021-04-26
NAS-110394 / 21.06 / Allow specifying extra kernel arguments

sonicaj:
This PR adds changes to allow specifying extra kernel arguments. This is helpful when users are using hardware which requires custom arguments to the kernel, for example we have had a user for whom `poweroff` does not work even on stock debian unless a custom parameter is specified.

### DIFF
--- a/src/middlewared/middlewared/plugins/system.py
+++ b/src/middlewared/middlewared/plugins/system.py
@@ -247,7 +247,7 @@ class SystemAdvancedService(ConfigService):
             Str('syslog_transport', enum=['UDP', 'TCP', 'TLS']),
             Int('syslog_tls_certificate', null=True),
             List('isolated_gpu_pci_ids', items=[Str('pci_id')]),
-            List('kernel_extra_options', items=[Str('pci_id')]),
+            List('kernel_extra_options', items=[Str('extra_option')]),
             update=True
         )
     )
--- a/src/middlewared/middlewared/plugins/system.py
+++ b/src/middlewared/middlewared/plugins/system.py
@@ -93,7 +93,7 @@ class SystemAdvancedModel(sa.Model):
     adv_kmip_uid = sa.Column(sa.String(255), nullable=True, default=None)
     adv_kdump_enabled = sa.Column(sa.Boolean(), default=False)
     adv_isolated_gpu_pci_ids = sa.Column(sa.JSON(), default=[])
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/system.py
```diff
@@ -246,6 +247,7 @@ async def __validate_fields(self, schema, data):
             Str('syslog_transport', enum=['UDP', 'TCP', 'TLS']),
             Int('syslog_tls_certificate', null=True),
             List('isolated_gpu_pci_ids', items=[Str('pci_id')]),
+            List('kernel_extra_options', items=[Str('extra_option')]),
```
themylogin, line 250: Why does it have to be an array?

src/middlewared/middlewared/plugins/system.py
```diff
@@ -363,8 +365,11 @@ async def do_update(self, data):
                 await self.middleware.call('etc.generate', 'kdump')
                 await self.middleware.call('etc.generate', 'grub')
 
-            if osc.IS_LINUX and original_data['isolated_gpu_pci_ids'] != config_data['isolated_gpu_pci_ids']:
-                await self.middleware.call('boot.update_initramfs')
+            if osc.IS_LINUX:
```
themylogin, line 368: We're linux-only now

src/middlewared/middlewared/plugins/system.py
william-gr: Makes the most sense to me IMO. Each entry being one option, instead of having one very long string array.

The opposite would also be true, why would it be a single string?

themylogin: Kernel options is a string, that's why I am suggesting.

How will this look like in UI? Multiple `<input type=text>` with `+` button? Why not a single textarea?

william-gr: No UI for this feature

themylogin: Strings will be easier to use in CLI too

themylogin: @sonicaj we also have to prohibit symbols like `\n` here, and handle `"` in generator (or prohibit it too)

sonicaj: I think we can have it as a string because that's also how we pass auxiliary parameters to other services. So we will be consistent that way as well.

PR #6821, opened 2021-04-26
NAS-110396 / 21.06 / Correctly retrieve host path volumes

sonicaj:
In some cases k8s api can retrieve a null value for volumes key. This commit addresses that case.


Ornias1993 (approved):
As discussed, thanks!

themylogin (approved):

PR #6823, opened 2021-04-27
NAS-110400 / 21.06 / Harden k3s

sonicaj:
This commit introduces changes to harden k3s based on rancher's recommendations. We are not setting PSP ( pod security policy ) explicitly because that will prevent us from consuming host features like host network which for example is very helpful in setting up clustered minio.

### DIFF
--- a/src/middlewared/middlewared/etc_files/rancher/k3s/flags.py
+++ b/src/middlewared/middlewared/etc_files/rancher/k3s/flags.py
@@ -16,7 +16,19 @@ def render(service, middleware):
             os.unlink(FLAGS_PATH)
         return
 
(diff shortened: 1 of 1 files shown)


Ornias1993:
> We are not setting PSP ( pod security policy ) explicitly because that will prevent us from consuming host features like host network which for example is very helpful in setting up clustered minio.

I think it's cleaner to just add RBAC's to the App charts to allow bypassing the PSP on a per-app basis....

sonicaj:
@Ornias1993 that is a good suggestion. Let me see what we can do about it :)

sonicaj:
We have decided to not implement PSP right now because they have been deprecated. A ticket has been made to keep track of the alternative to PSP and implement it once it's feature complete.

yocalebo (approved):

Ornias1993 (approved):

PR #6825, opened 2021-04-27
NAS-110417 / 21.06 / Have unique serial number for certs

sonicaj:
When creating a self generated certificate, we should have unique serial numbers. The upper limit for serial is 2^160 but in the changes i am keeping it 30 as otherwise it will just be a really long number which is hard to access and it has a low probability that we have 2 self signed certificates generated by us with same serial number.


themylogin (approved):

william-gr:
backport

PR #6828, opened 2021-04-29
NAS-110343 / 21.06 / Properly set hostname in scale

sonicaj:
On first boot, contents of `/etc/hostname` contain the host name of the builder where the iso was generated on. What happens is that when the system boots and interfaces are synced, `hostname --all-fqdn` reports the old hostname as well because it has already been advertised in the network. We should be setting the new/correct hostname on fresh install / upgrade before syncing network to ensure correct hostname is advertised.

### DIFF
--- a/src/middlewared/middlewared/etc_files/hostname.py
+++ b/src/middlewared/middlewared/etc_files/hostname.py
@@ -1,14 +1,20 @@
 import logging
+import subprocess
 
-from middlewared.utils import run
+from middlewared.service import CallError
 
 logger = logging.getLogger(__name__)
 
 
-async def render(service, middleware):
-    config = await middleware.call("network.configuration.config")
-    hostname = f"{config['hostname_local']}.{config['domain']}"
-    with open('/etc/hostname', 'w') as f:
+def render(service, middleware):
+    config = middleware.call_sync("network.configuration.config")
+    hostname = config['hostname_local']
+    if config['domain']:
+        hostname += f'.{config["domain"]}'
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/etc_files/hostname.py
```diff
@@ -7,4 +7,8 @@
 
 async def render(service, middleware):
     config = await middleware.call("network.configuration.config")
-    await run(["hostname", f'{config["hostname_local"]}.{config["domain"]}'])
+    hostname = f"{config['hostname_local']}.{config['domain']}"
+    with open('/etc/hostname', 'w') as f:
```
yocalebo, line 11: Since we're doing IO now, I guess this can block event loop so probably should make it synchronous or run in thread?

yocalebo (changes requested):

src/middlewared/middlewared/etc_files/hostname.py
```diff
@@ -7,4 +7,8 @@
 
 async def render(service, middleware):
     config = await middleware.call("network.configuration.config")
-    await run(["hostname", f'{config["hostname_local"]}.{config["domain"]}'])
+    hostname = f"{config['hostname_local']}.{config['domain']}"
```
yocalebo, line 10: As discussed in slack, `/etc/hostname` should be the short-hand name (`hostname_local`) and `/etc/resolv.conf` has the domain name portion.

yocalebo (approved):
Awesome, thanks!

PR #6836, opened 2021-05-05
NAS-110510 / 21.06 / Only delete jobs log dir if it exists

sonicaj:
This commit fixes an issue where middleware will not start if the jobs log directory did not exist and we tried to delete it.

### DIFF
--- a/src/middlewared/middlewared/job.py
+++ b/src/middlewared/middlewared/job.py
@@ -1,4 +1,5 @@
 import asyncio
+import contextlib
 from collections import OrderedDict
 import copy
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/job.py
```diff
@@ -183,7 +183,8 @@ def __init__(self, maxlen=1000):
         self.maxlen = maxlen
         self.count = 0
         self.__dict = OrderedDict()
-        shutil.rmtree(LOGS_DIR)
+        if os.path.exists(LOGS_DIR):
```
themylogin, line 186: `with contextlib.suppress(FileNotFoundError)` otherwise it's a potential race condition

themylogin (changes requested):

themylogin (approved):

PR #6837, opened 2021-05-05
NAS-110511 / 21.06 / Bug fixes for setting system birthday

sonicaj:

1) We were trusting that system time would be the correct one, but that may not be the case if `ntpd` has not synced. Changes have been introduced to make sure that we trust the time returned by ntp request if only the clock has synchronized.

2) If we were unable to determine system time, we were waiting for 24 hours before checking again for system birthday which means that we would be storing the incorrect system birthday even when we could potentially save the correct time if we just checked sooner.

3) Do not unnecessarily have a hook for `interface.post_sync` when setting system birthday is a one time thing. It will already be ensured that we do set it. So if the first pass failed and interfaces were synced, we will be having another pass after 5 minutes and we will catch up with the correct synced time.

### DIFF
--- a/src/middlewared/middlewared/plugins/system.py
+++ b/src/middlewared/middlewared/plugins/system.py
@@ -817,9 +817,9 @@ class SystemService(Service):
             response = client.request('localhost')
         except Exception:
             # Cannot connect to NTP server
-            self.logger.debug('Error while connecting to NTP server', exc_info=True)
+            self.logger.error('Error while connecting to NTP server', exc_info=True)
         else:
-            if response.leap != 3:
+            if response.version and response.leap != 3:
                 # https://github.com/darkhelmet/ntpstat/blob/11f1d49cf4041169e1f741f331f65645b67680d8/ntpstat.c#L172
                 # if leap second indicator is 3, it means that the clock has not been synchronized
                 return datetime.fromtimestamp(response.tx_time, timezone.utc)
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/system.py
```diff
@@ -1618,41 +1621,18 @@ def set_crash_reporting(self):
...
+    while True:
+        birthday = await middleware.call('system.get_synced_clock_time')
+        if birthday:
+            middleware.logger.debug('Updating birthday data')
+            # update db with new birthday
+            settings = await middleware.call('datastore.config', 'system.settings')
+            await middleware.call('datastore.update', 'system.settings', settings['id'], {'stg_birthday': birthday})
+            break
+        else:
+            middleware.logger.debug('Waiting for system clock to be synchronized')
```
themylogin, line 1634: Will it flood middleware logs if clock are never synchronized?

themylogin (changes requested):

themylogin (approved):

src/middlewared/middlewared/job.py
```diff
@@ -183,7 +183,8 @@ def __init__(self, maxlen=1000):
         self.maxlen = maxlen
         self.count = 0
         self.__dict = OrderedDict()
-        shutil.rmtree(LOGS_DIR)
+        if os.path.exists(LOGS_DIR):
```
yocalebo, line 186: Looks like the most recent commit might have removed the `with contextlib` context manager?

yocalebo: Oh wait, looks like this should be part of the changes for this commit?

sonicaj: It would catch up as soon as the related PR is merged - a sec ; )

yocalebo (approved):

PR #6839, opened 2021-05-05
NAS-103891 / 21.06 / ipmi.clear_sel

themylogin:


william-gr:
How can we read SEL ?

yocalebo (changes requested):
This will not work on X series based devices since IPMI device is not accessible from the host OS.

yocalebo:
> This will not work on X series based devices since IPMI device is not accessible from the host OS.

Meh, I see this is just adding a method to be called.

themylogin:
@william-gr alert source creates alerts for new events. Do we need to show them all somewhere?

william-gr:
@themylogin Yes, if I get an alert saying SEL is full, I want a way to retrieve them all first

william-gr (approved):

yocalebo (approved):

bvasilenko:
@themylogin Could you pls show an example of back-end response from `ipmi.query_sel` ? I'm integrating this into WebUI

themylogin:
@bvasilenko please ask QE team for a hardware access to a machine that has IPMI SEL events present and SCALE installed.

PR #6845, opened 2021-05-05
IPMI Battery Voltage Alert

themylogin:


yocalebo (approved):

yocalebo (changes requested):
Actually wait, it can be `is_ix_hardware` and `is_gigabyte` which means `_produce_sensor_list_alert()` will run twice.

yocalebo:
> @yocalebo I think it will not because we only produce alerts for CPU and SYSFAN sensors when `is_gigabyte is True`

This is in scope of `is_gigabyte` ?
```
if alert := self._produce_sensor_alert(sensor):
    alerts.append(alert)
```

themylogin:
@yocalebo we have
```
                if not (RE_CPUTEMP.match(sensor["name"]) or RE_SYSFAN.match(sensor["name"])):
                    continue
```
above

yocalebo:
> @yocalebo we have
> 
> ```
>                 if not (RE_CPUTEMP.match(sensor["name"]) or RE_SYSFAN.match(sensor["name"])):
>                     continue
> ```
> 
> above

how does that effect battery voltage alert?

yocalebo (approved):

PR #6848, opened 2021-05-06
Bug fix for starting middleware

sonicaj:
Do not use strict for lifetime dict because cron schema always has a default by default.


themylogin (changes requested):
`strict=True` should remain there

PR #6850, opened 2021-05-06
NAS-109069 / 21.06 / Sharing protocol client counts in reporting.realtime

themylogin:


src/middlewared/middlewared/plugins/nfs_/status.py
```diff
@@ -0,0 +1,32 @@
...
+
+        bus = dbus.SystemBus()
+        proxy = bus.get_object("org.ganesha.nfsd", "/org/ganesha/nfsd/ClientMgr")
+        iface = dbus.Interface(proxy, dbus_interface="org.ganesha.nfsd.clientmgr")
+        return len(iface.ShowClients()[1])
+
+        We should reconsider our approach if they ever fix this.
+        """
+
+        with open("/var/run/ganesha.pid") as f:
```
yocalebo, line 24: I know that we check if the service is running 1st, but probably best to catch FileNotFoundError and return early.

src/middlewared/middlewared/plugins/nfs_/status.py
```diff
@@ -0,0 +1,32 @@
...
+            pid = int(f.read())
```
yocalebo, line 25: What happens if ganesha fails to write the PID into the file?

themylogin: Is this a normal state for Ganesha? What do you think will be better approach, retrieve PID from systemd? List processes and match by cmdline?

src/middlewared/middlewared/plugins/nfs_/status.py
themylogin: We catch all these exceptions in event handler (and ignore them in production mode) so I thought it would be fine for this method to fail.

src/middlewared/middlewared/plugins/nfs_/status.py
yocalebo: If no PID, casting to int will fail and be caught like you said up above so no need to complicate it now.

yocalebo (approved):

yocalebo:
> @yocalebo will reading procfs/sysfs will be slow on large systems? Or calling this samba method?

For smb, running that command actually traverses (c)tdb database files on disk. Using the `fast` option avoids some of the most expensive operations but still, the performance impact is unknown. Potential for O(n) scaling issues as client count increases for SMB.

themylogin:
Well, pseudofs readings will also be O(n) :) @anodos325 what do you think, how often can we run this?

yocalebo:
I vote that we don't put this in the realtime event reporting facility until it can be more thoroughly tested by performance team on what the impact will be. I'd rather just have these methods available to be consumed whenever the consumer wants. If a consumer causes a performance impact then the onus is put on them to fix it. If the performance impact is because we did it to ourselves then the fix becomes more complicated.

PR #6853, opened 2021-05-07
NAS-109476 / 21.06 / Replication name regex

themylogin:


yocalebo (commented):
Is the migration supposed to be dropping the fiber channel table?

yocalebo:
> @yocalebo it seems that this table is not defined anymore, I've thought you dropped it while moving enterprise plugins to the main tree. Is this an error?

I can't remember but FC hasn't been touched on SCALE, yet so this is fine removing it.

yocalebo (approved):

william-gr:
@yocalebo @themylogin if we drop fiber channel table, how do we migrate from CORE/Enterprise? :)

themylogin:
@william-gr will we need data from this table later?

william-gr:
I dont know how FC works in Linux, but I imagine so?

PR #6854, opened 2021-05-07
NAS-101630 / 22.12 / Allow dumping and restoring xattr with cloud sync

themylogin:


william-gr:
We probably want explicit review from @anodos325 here ;)

anodos325:
Yeah, getfacl is not correct command in this case. On SCALE it will only grab POSIX1e ACLs. Let's hold off on finalizing / merging until the kernel / ZFS / middleware components are in place.

themylogin:
I've commented them out but I still want to merge this so UI team can add this checkbox. The JIRA issue will remain open and we'll revisit it once we have ZFS ready. Or can it be that we won't be able to have this ready by SCALE release? @anodos325

anodos325:
I'm also somewhat worried about how the size of this file will scale. We might want to test with something like a dataset with a million files or so. In theory you should be able to check with POSIX1E acltype (without setting an actual ACL).

themylogin:
@anodos325 You are correct, I need to also gzip this file

anodos325:
The way NFSv4 ACLs are designed for scale, I know they can be preserved via rsync by explicitly specifying the system.nfs4_acl_xdr xattr.

themylogin:
@anodos325 no, cloud backends only store file name, size and mtime

anodos325:
> no, cloud backends only store file name, size and mtime

Hmm... that seems like a more fundamental problem that we need to address before moving on to ACLs. The ACLs in SCALE can be read from / written to server as xattrs. If we solve the xattr storage problem, then we also store the ACL problem.

themylogin:
This problem can't be addressed, Google Drive will never store xattrs, that's why we are simply dumping them to a file.

rick-mesta:
Am I reading this right ? How are our competitors dealing with (what seems) this very basic limitation of cloud providers ? Storing ZFS ACL's as a separate object (associated with the actual file data) ? Doesn't that mean that typical tools like rsync, et. al. need to know about this "layout" when sucking the file's data (and new metadata file) down ?

themylogin:
@rick-mesta I am not aware of any other competitors that can push xattrs to generic "cloud"

themylogin:
backport

rick-mesta:
@truenas/middleware: Team, can we get Vlad some feedback on this bad boy ?

themylogin:
@rick-mesta I don’t need any feedback yet, this is not done, I need xattr test cases from @anodos325 :)

rick-mesta:
Oh... I see: thx for clarifying @themylogin 😄

anodos325:
For xattrs storing ACLs (POSIX and NFSv4) you can probably implement a test as simple as:
(1) create two datasets (POSIX acltype and NFSV4 acltype).
(2) create a directory and create a file in the dataset
(3) iterate through all valid ACL permutations for both ACL types and do the following:

a) write new ACL through our API
b) back up to cloud service
c) restore to different path (with same dataset acltype)
d) use `filesystem.getacl` to verify that ACL on source and restored is same

xattrs are:
```
"system.posix_acl_access"
"system.posix_acl_default"
```
for POSIX1E

```
"system.nfs4_acl_xdr"
```

You should also ensure that user and security namespaces are also preserved and written to correct namespace.

anodos325:
If needed I can expand the python SMB client to be able to read / write ACLs if you want that available for validating restoring the permissions on the combination of the POSIX ACL + security.ntacl.

anodos325:
Max xattr size (for single xattr) on ZFS on SCALE is 2 MiB. Clients may set a large amount of these, but I don't think I've seen more than 20 in real life, with most typical case 0-5 in count. In a typical SMB share with MacOS clients, at least some will be present on all files.

So maybe for testing have some deeply directory trees, with large numbers of files, large numbers 1-5 xattrs on each with varying sizes.

PR #6856, opened 2021-05-09
NAS-108987 / 21.06 / Allow importing pool which is consuming complete disk

sonicaj:
This PR takes changes from https://github.com/truenas/middleware/pull/6281 ( thank you @dhiltonp for your contribution ) to allow importing pools which are consuming complete disk(s). With this change, it also corrects behavior so that middleware is able to recognize that a complete disk is being used by a pool which middleware was not able to translate before.

### DIFF
--- a/src/middlewared/middlewared/plugins/zfs.py
+++ b/src/middlewared/middlewared/plugins/zfs.py
@@ -369,7 +369,7 @@ class ZFSPoolService(CRUDService):
         found = False
         with libzfs.ZFS() as zfs:
             for pool in zfs.find_import(
-                cachefile=cachefile, search_paths=['/dev/disk/by-partuuid'] if osc.IS_LINUX else None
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -369,7 +369,7 @@ def import_pool(self, name_or_guid, options, any_host, cachefile, new_name):
         found = False
         with libzfs.ZFS() as zfs:
             for pool in zfs.find_import(
-                cachefile=cachefile, search_paths=['/dev/disk/by-partuuid'] if osc.IS_LINUX else None
+                cachefile=cachefile, search_paths=['/dev/disk/by-partuuid', '/dev'] if osc.IS_LINUX else None
```
themylogin, line 372: `osc.IS_LINUX else` is always true

themylogin (changes requested):

themylogin:
Do all our pool/vdev operations work with such pools?

PR #6864, opened 2021-05-10
NAS-109910 / 12.0 / Bring back iscsi_target_extent_name unique constraint

themylogin:


themylogin:
@rick-mesta I am adding a unique constraint to a database column (it will also serve as an index which will speed up the search by this column). That's what
```
    with op.batch_alter_table('services_iscsitargetextent', schema=None) as batch_op:
        batch_op.create_unique_constraint(batch_op.f('uq_services_iscsitargetextent_iscsi_target_extent_name'), ['iscsi_target_extent_name'])
```
is doing. It has such a complicated form because SQLite does not allow almost any table structure modifications, so every time we want to modify table structure, we must recreate the table. And so alembic always groups these operations in a batch to do them all at once (even if we have only one operation to do).

I added a custom `ensure_unique_string` function to make sure we don't have any rows that have the same column value. It's highly unlikely to have such rows (we don't allow them with middleware), but we can never allow migrations to fail (it will prevent user from upgrading).

rick-mesta:
@themylogin: Thx for the detailed explanation 👍🏼

rick-mesta (approved):

PR #6872, opened 2021-05-11
Convert dscache.refresh into a middleware job

anodos325:
This commit serializes cache refresh requests into a job queue and
ensures that cache is refreshed before writing to file.

API is private and so converting to a job shouldn't be a significant change.


src/middlewared/middlewared/plugins/cache.py
```diff
@@ -223,7 +224,8 @@ async def refresh(self):
         for ds in available_ds:
             ds_state = await self.middleware.call(f'{ds}.get_state')
             if ds_state == 'HEALTHY':
-                await self.middleware.call(f'{ds}.fill_cache', True)
+                fill_job = await self.middleware.call(f'{ds}.fill_cache', True)
+                await fill_job.wait()
```
sonicaj, line 228: If it errors out, would we silently ignore ? Should we perhaps use `wrap` here ?

anodos325:
I'd say "yes" and "yes". Turtles all the way down.
I'll fix it.

PR #6877, opened 2021-05-11
NAS-109989 / 12.0 / Convert dscache.refresh into a middleware job (#6…

anodos325:
…872)

This commit serializes cache refresh requests into a job queue and
ensures that cache is refreshed before writing to file.

Add docstring to directoryservices.cache_refresh


anodos325:
Forgot to bugclerk this one.

PR #6882, opened 2021-05-12
NAS-110794 / 21.06 / Inherit special_small_blocks by default

Ornias1993:
Users would expect that special_small_blocks on child datasets inherit the setting of their parent, just like compression.
Not having this inherited leads to costly mistakes, requiring completely rewriting the data to get the smallblocks loaded into the metadata device.

Note:
While the UI isn't able to show "inherited" it does show inherited small block values when set.
However: We might want to also use a dropdown for small-block sizes, as the same limitations apply as for recordsizes (power of 2)


themylogin (approved):

sonicaj:
This will be handled in https://github.com/truenas/middleware/pull/6922

PR #6888, opened 2021-05-17
NAS-110600 / 21.06 / Correctly retrieve s3 attachment

sonicaj:
This commit fixes an issue if s3 path did not exist for some reason ( got deleted or was an old path ), retrieving s3 attachments would error out as the path would not exist. Secondly, if the path does not exist but has been configured - we create the path before starting the s3 service.

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/minio/configure.py
+++ b/src/middlewared/middlewared/etc_files/local/minio/configure.py
@@ -44,11 +44,16 @@ def render_certificates(s3, middleware):
 
 def configure_minio_sys_dir(s3):
     storage_path = s3['storage_path']
+    # Create storage path if it does not exist
+    os.makedirs(storage_path, exist_ok=True)
     minio_dir = os.path.join(storage_path, '.minio.sys')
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/etc_files/local/minio/configure.py
```diff
@@ -44,11 +44,16 @@ def render_certificates(s3, middleware):
 
 def configure_minio_sys_dir(s3):
     storage_path = s3['storage_path']
+    # Create storage path if it does not exist
+    os.makedirs(storage_path, exist_ok=True)
```
themylogin, line 48: Why should we do that? Do we do that for other share types?

sonicaj: @themylogin it's already being done in `s3.update` - similarly it's happening for webdav/smb as well ( Here it's more of a sanity check to ensure it exists complying with `s3.update` ) . We should probably create a ticket to remove this everywhere but that would be change of behavior, so not sure.

themylogin (approved):

sonicaj:
backport

PR #6889, opened 2021-05-17
NAS-110667 / 21.06 / Add testparm output to debug

anodos325:


anodos325:
testparm is most reliable way to get running configuration, and performs some basic validation.

yocalebo (approved):

PR #6892, opened 2021-05-18
NAS-110689 / 21.06 / Only consider interfaces in database for dhcp/ipv6 autoconfiguration

sonicaj:
This commit  fixes an issue where it's impossible to configure aliases with dhcp if we have for example 4 NICs and on a fresh install all interfaces will have DHCP running, then if for 1 interface we want to configure aliases with dhcp, we won't be able to do that as it will error out saying that DHCP can only be enabled for 1 interface.


sonicaj:
retest this please

themylogin (approved):

PR #6896, opened 2021-05-19
NAS-110712 / 12.0 / add --trace-malloc arg to middlewared

yocalebo:
This allows one to start `middlewared` in the foreground and pass `--trace-malloc` as an argument to print memory allocations for the middlewared process.

This tracks memory allocations and not memory leaks. However, the idea is that you monitor the output of the logs and if you see a module continually allocate memory and never release it, then you might be able to glean some useful information or at least get pointed somewhere.

This allowed me to spot some low-hanging fruit with middlewared start-up where we were allocating significant memory unnecessarily. See https://github.com/truenas/iocage/pull/345 for details.

### DIFF
--- a/src/middlewared/middlewared/main.py
+++ b/src/middlewared/middlewared/main.py
@@ -51,6 +51,7 @@ import traceback
 import types
 import urllib.parse
 import uuid
+import tracemalloc
 
 from . import logger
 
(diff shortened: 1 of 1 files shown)


yocalebo:
@themylogin I've asked you to review this one too since it's the same as https://github.com/truenas/middleware/pull/6930

themylogin (approved):
Excellent, thank you!

PR #6897, opened 2021-05-20
NAS-110714 / 12.0 / fix memory allocation/leak issue in multipath.query()

yocalebo:
For reasons that aren't very clear, `multipath.query()` is allocating (leaking) memory that is never freed in the io_thread pool. We, eventually, end up in `_ParserDictionaryContext` in `lxml`. That particular function does some fancy stuff for having a shared global dictionary per thread and "deduplicates" name entries when processing xml data. The idea is that the names of the attributes being processed in the xml document aren't going to change very often. Once everything is "cached" it doesn't grow anymore. This was added to `lxml` for the obvious speed improvements, however, it's causing us problems in our thread pool. I suspect it's because we're using daemon=True for all our threads which means the resources for said thread isn't cleaned up properly (safely) but I digress....

This leak is particularly painful because it shows up even if the user doesn't have multipath disks. The leak is actually happening by this line `doc = etree.fromstring(sysctl.filter("kern.geom.confxml")[0].value)`. Even though there are no multipath objects in the xml document and the method would return an empty `[]`, it still caused memory to be allocated that was never released back to OS.

My fix does 2 things:
1. Instead of using `lxml` in the multipath module, use the built-in xml library since we're not using anything in lxml that isn't already present in the shipped xml library with base python.
2. Get rid of the `Multipath` and `Consumer` classes. I didn't spot any modules importing these modules anywhere and, in my opinion, they made this file entirely overly complicated for no real benefit.

Also, since I'm on my soap-box, I suspect this will (hopefully) fix some of the `middlewared` core dumps that we're seeing. On _every_ _single_ core dump that I analyzed, I saw `_ParserDictionaryContext` somewhere on CPU. Either in the thread that actually crashed, or on another thread.

Finally I can confirm that by simply switching to the base `xml` library, resident memory growth on initial `middlewared` service startup reduced from growing 5-8MB per 60 seconds to not growing at all after the initial memory allocation stabilized.

### DIFF
--- a/src/middlewared/middlewared/plugins/multipath.py
+++ b/src/middlewared/middlewared/plugins/multipath.py
@@ -7,6 +7,9 @@ from middlewared.utils import filter_list
 
 class MultipathService(CRUDService):
 
+    class Config:
+        datastore_primary_key_type = 'string'
+
     @filterable
     def query(self, filters, options):
         """
--- a/src/middlewared/middlewared/plugins/multipath.py
+++ b/src/middlewared/middlewared/plugins/multipath.py
@@ -83,6 +83,5 @@ class MultipathService(CRUDService):
                 'status': g.find("./config/State").text,
                 'children': children,
             }),
-        else:
-            doc = None
-            return result
+
+        return result
--- a/src/middlewared/middlewared/plugins/multipath.py
+++ b/src/middlewared/middlewared/plugins/multipath.py
@@ -1,18 +1,11 @@
-try:
-    import sysctl
-except ImportError:
-    sysctl = None
+from xml.etree import ElementTree as ET
 
-from lxml import etree
-
+import sysctl
 from middlewared.service import CRUDService, filterable
-from middlewared.utils import filter_list, osc
+from middlewared.utils import filter_list
 
 
 class MultipathService(CRUDService):
-
-    class Config:
-        datastore_primary_key_type = 'string'
 
     @filterable
     def query(self, filters, options):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/multipath.py
```diff
@@ -1,19 +1,12 @@
...
 
+import sysctl
 from middlewared.service import CRUDService, filterable
-from middlewared.utils import filter_list, osc
+from middlewared.utils import filter_list
 
 
 class MultipathService(CRUDService):
 
-    class Config:
```
sonicaj, line 14: Why are we removing this ?

src/middlewared/middlewared/plugins/multipath.py
```diff
@@ -56,86 +49,37 @@ def query(self, filters, options):
...
-            self.lunid = ""
-        self.__xml = xmlnode
+            result.append({
+                'type': 'root',
+                'name': 'multipath/' + g.find("./name").text,
+                'status': g.find("./config/State").text,
+                'children': children,
+            }),
+        else:
+            doc = None
```
sonicaj, line 84: Is this required ?

src/middlewared/middlewared/plugins/multipath.py
```diff
@@ -56,86 +49,37 @@ def query(self, filters, options):
...
+            return result
```
sonicaj, line 85: Why do we return in `else` ?

src/middlewared/middlewared/plugins/multipath.py
```diff
@@ -56,86 +49,37 @@ def query(self, filters, options):
...
-        except Exception:
-            self.lunid = ""
-        self.__xml = xmlnode
+            result.append({
+                'type': 'root',
+                'name': 'multipath/' + g.find("./name").text,
+                'status': g.find("./config/State").text,
+                'children': children,
+            }),
+        else:
```
sonicaj, line 83: When would we never reach this block ?

yocalebo: `else:` is reached if you have a system with no multipath disks.

sonicaj: I am not sure i understand why we need the `else` block ? Actions performed in `else` need to be performed in either cases. This could even potentially introduce a bug if later a break is added in the loop.

yocalebo: It's old debugging code I was  using on live system. I've removed it.

sonicaj (approved):

sonicaj:
Thanks

src/middlewared/middlewared/plugins/multipath.py
```diff
@@ -56,86 +52,36 @@ def query(self, filters, options):
...
 
-        if osc.IS_LINUX:
-            return []
-
-        multipaths = self.__get_multipaths()
-
-        items = []
-        for mp in multipaths:
+    def __get_multipaths(self):
+        doc = ET.fromstring(sysctl.filter("kern.geom.confxml")[0].value)
```
william-gr, line 58: We may want to leave a comment somewhere of why xml.etree and not lxml for anyone with bright ideas in the future.

yocalebo: Kind of ashamed I didn't do that initially. Thanks for pointing it out. I've added a comment.

rick-mesta (approved):
ship it !!! 🚢

PR #6902, opened 2021-05-21
NAS-110635 / 12.0 / Do not allow comma in cdrom path

sonicaj:
This commit adds changes to ensure user does not specify a comma in cdrom path as that would result in bhyve not able to parse the argument correctly.


yocalebo (approved):

vampyren:
Thank you for the quick fix.

rick-mesta (approved):

PR #6903, opened 2021-05-21
NAS-110724 / 12.0 / optimize certificate alert source

yocalebo:
Anytime `certificate.query()` or `certificateauthority.query()` is made the following calls are made for *EACH* cert and/or ca: `cryptokey.load_certificate` 2x `cryptokey.load_private_key` 1x.

In the `certificates` alert source, we do not specify a schedule so it defaults to running every 60 seconds. Furthermore, between the `certificate.query()` and `certificateauthority.query()` api calls, we were calling them13 times total.

This means the formula for API calls being made every 60 seconds is `39 * n`. `n` being the number of certificates and/or cas on the system. So for 2 certificates on the system, (potentially) 78 API calls were made. That's especially painful because the `cert**.query()` api calls return quite a bit of data but they also return quite a bit of duplicate data. Part of this data is the full contents (in string form) the public/private key contents. This becomes quite a bit of data to throw around. With my change the formula for calculating API calls is `3 * n`.

My changes do a few things:

1. don't run the certificates alert source every 60 seconds but instead run it once a day (discussed w/ @sonicaj)
2. restructure the certificates alert source so that we only call `certificate.query()` and `certificateauthority.query()` once
3. this alert does not need to be run on the standby controller since on failover we reconfigure the certs and restart webUI

### DIFF
--- a/src/middlewared/middlewared/alert/source/certificates.py
+++ b/src/middlewared/middlewared/alert/source/certificates.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 
 from middlewared.alert.base import AlertClass, SimpleOneShotAlertClass, AlertCategory, AlertLevel, Alert, AlertSource
+from middlewared.alert.schedule import CrontabSchedule
 
 
 class CertificateIsExpiringAlertClass(AlertClass):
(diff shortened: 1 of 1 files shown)


sonicaj (commented):
@yocalebo there are 2 issues at it's core:
1) Alert running every 60sec.
2) Retrieving same certificate data in each separate alert source

I propose that for 12.0 we only tackle (1) as (2) is only going to run once daily to keep the diff small for stable branch. Please let me know what you think about it. Thank you

yocalebo:
@sonicaj I've made a separate PR here: https://github.com/truenas/middleware/pull/6906 to keep 12 changes minimal.

yocalebo:
backport

PR #6904, opened 2021-05-23
NAS-110641 / 21.06 / Allow specifying user properties with pool.dataset service

sonicaj:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2912,6 +2912,15 @@ def transform(dataset):
         Ref('encryption_options'),
         Bool('encryption', default=False),
         Bool('inherit_encryption', default=True),
+        List(
```
themylogin, line 2915: It's pretty unobvious API. The natural use case here is to replace dataset's user properties with the ones provided. If you want to avoid re-sending properties every time, add a second field `user_properties_update` into `do_update` method.

themylogin (changes requested):

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3120,6 +3148,7 @@ def add(attr):
         ('edit', _add_inherit('readonly')),
         ('edit', _add_inherit('recordsize')),
         ('edit', _add_inherit('snapdir')),
+        ('edit', _user_props_rename('user_properties')),
```
themylogin, line 3151: This will not allow to replace dataset's properties with ones specified in the request, won't it?

themylogin (changes requested):

src/middlewared/middlewared/plugins/pool.py
sonicaj: It would ?

themylogin: How? Your `edit` seems to be renaming `user_properties` into `user_properties_update`

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3120,10 +3162,19 @@ def add(attr):
...
         ('add', Inheritable('quota_warning', value=Int('quota_warning', validators=[Range(0, 100)]))),
         ('add', Inheritable('quota_critical', value=Int('quota_critical', validators=[Range(0, 100)]))),
         ('add', Inheritable('refquota_warning', value=Int('refquota_warning', validators=[Range(0, 100)]))),
         ('add', Inheritable('refquota_critical', value=Int('refquota_critical', validators=[Range(0, 100)]))),
+        ('add', List(
```
themylogin, line 3170: Won't it be simpler to have `user_properties` schema without the `remove` key and add for update method just add `user_properties_update` schema that has the `remove` key?

themylogin (approved):

PR #6905, opened 2021-05-23
Allow creating ancestors when creating a dataset

sonicaj:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3276,10 +3292,14 @@ async def __common_validation(self, verrors, schema, data, mode, parent=None, cu
             )
 
         if not parent:
-            verrors.add(
-                f'{schema}.name',
-                'Please specify a pool which exists for the dataset/volume to be created'
-            )
+            # This will only be true on dataset creation
+            if data.get('create_ancestors'):
```
themylogin, line 3296: Why `.get` if we have a default?

themylogin (changes requested):

PR #6909, opened 2021-05-24
NAS-110704 / 21.06 / Do not remove snapshot fields before running annotate_snapshots on them

themylogin:


src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -1002,12 +1002,20 @@ def query(self, filters, options):
...
-            result = self.middleware.call_sync('zettarepl.annotate_snapshots', [result])[0]
+        if not select or 'retention' in select:
+            if isinstance(result, list):
+                result = self.middleware.call_sync('zettarepl.annotate_snapshots', result)
+            elif isinstance(result, dict):
+                result = self.middleware.call_sync('zettarepl.annotate_snapshots', [result])[0]
+
+        if select:
+            if isinstance(result, list):
+                result = [{k: v for k, v in item.items() if k in select} for item in result]
```
sonicaj, line 1016: Why don't we do this before doing `filter_list` ? If a user has thousands of snapshots, we would be iterating twice ?

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -1002,12 +1002,20 @@ def query(self, filters, options):
...
+            elif isinstance(result, dict):
```
sonicaj, line 1017: Should we also be doing this if a snapshot is queried directly ? ( the `id` case )

src/middlewared/middlewared/plugins/zfs.py
themylogin: Do what exactly?

src/middlewared/middlewared/plugins/zfs.py
themylogin: Sorry, do what?

sonicaj (approved):

src/middlewared/middlewared/plugins/zfs.py
sonicaj: Ah, my bad - the above return is only limited to `name` selection. Thank you for pointing it out

PR #6911, opened 2021-05-24
NAS-110766 / 12.0 / remove lxml from global_freebsd plugin

yocalebo:


src/middlewared/middlewared/plugins/iscsi_/global_freebsd.py
```diff
@@ -17,12 +14,12 @@ class Config:
...
         namespace = 'iscsi.global'
 
-    def sessions(self, filters, options):
+    async def sessions(self, filters, options):
         """
         Get a list of currently running iSCSI sessions. This includes initiator and target names
         and the unique connection IDs.
         """
-        def transform(tag, text):
+        async def transform(tag, text):
```
themylogin, line 22: Why do you make it an async function?

themylogin (changes requested):

src/middlewared/middlewared/plugins/iscsi_/global_freebsd.py
yocalebo: Because I converted the outer function to an `async` function since I got rid of the `subprocess` module and used our `await run` wrapper method.

themylogin: You don't need to make `transform` asynchronous, you don't do any I/O inside

yocalebo: Is it not bad practice to have an `async` method call a synchronous method?

themylogin: It's a bad practice if you call a synchronous method that performs IO or heavy computing. Converting string to int or adding up a few numbers on the main loop is totally fine.

themylogin (approved):

PR #6912, opened 2021-05-25
NAS-110769 / 12.0 / remove lxml from fcport and fix initators bug

yocalebo:
before my change the `initiators` key would return this on an HA system:
```
    "initiators": [
      "naa.21000024ff19a90a",
      "naa.21000024ff19a90b",
      "naa.21000024ff19a90b (TrueNAS Controller 1)",
      "naa.21000024ff19a90a (TrueNAS Controller 1)"
    ]
```

after my change the `initiators` key returns this on HA systems:
```
    "initiators": [
      "naa.21000024ff19a90a (Active Controller)",
      "naa.21000024ff19a90b (Active Controller)",
      "naa.21000024ff19a90b (Standby Controller)",
      "naa.21000024ff19a90a (Standby Controller)"
    ]

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/fcport.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/fcport.py
@@ -1,18 +1,8 @@
-# Copyright (c) 2019 iXsystems, Inc.
-# All rights reserved.
-# This file is a part of TrueNAS
-# and may not be copied and/or distributed
-# without the express permission of iXsystems.
-
 import re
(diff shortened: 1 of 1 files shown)


themylogin:
I don't understand if this behavior change is correct or not :(

yocalebo:
> I don't understand if this behavior change is correct or not :(

Yeah, it's not obvious but the duplicate entries are actually different initiators so I've corrected the verbiage on HA systems.

themylogin (approved):

PR #6913, opened 2021-05-25
Fix middleware startup

themylogin:


yocalebo (approved):

src/middlewared/middlewared/plugins/ipmi.py
```diff
@@ -174,14 +174,14 @@ async def identify(self, options):
...
+    async def query_sel(self, filters, options):
         """
         Query IPMI System Event Log
         """
-        return [
+        return filter_list([
             record._asdict()
             for record in parse_ipmitool_output(await run('ipmitool', '-c', 'sel', 'elist'))
-        ]
+        ])
```
sonicaj, line 184: We should pass `filters/options` ?

PR #6919, opened 2021-05-25
NAS-110264 / 21.08 / Migrate AFP shares to Samba

themylogin:


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -1388,6 +1390,8 @@ async def pool_post_import(middleware, pool):
         By the time the post-import hook is called, the smb.configure should have
         already completed and initialized the SMB service.
         """
+        await middleware.call('sharing.smb.disable_acl_if_trivial')
```
anodos325, line 1393: Is this always disabling ACL if the ACL on share is trivial? If so, this is wrong.

src/middlewared/middlewared/alembic/versions/21.MM/2021-05-25_22-00_afp_to_smb.py
```diff
@@ -0,0 +1,162 @@
...
+        cifs_auxsmbconf = [
+            "fruit:locking = netatalk",
+            "fruit:metadata = netatalk",
+            "fruit:resource = file",
+            "streams_xattr:prefix = user.",
+            "streams_xattr:store_stream_type = no",
+            "oplocks = no",
+            "level2 oplocks = no",
+            "strict locking = auto",
+        ]
```
anodos325, line 45: Rather than adding auxiliary parameters, perhaps we should add a column named something like `cifs_legacy_afp_compabilty` and then alter smb.conf generation based on that.

src/middlewared/middlewared/alembic/versions/21.MM/2021-05-25_22-00_afp_to_smb.py
```diff
@@ -0,0 +1,162 @@
...
+        if share["afp_rw"].strip():
+            cifs_auxsmbconf.append(f"rwlist = {share['afp_rw'].strip()}")
+            share_disable_acl_if_trivial = True
+        umask_inv = "%o" % (~int(share["afp_umask"], 8) & 0o777)
+        cifs_auxsmbconf.append(f"create mask = 0{umask_inv}")
+        cifs_auxsmbconf.append(f"directory mode = 0{umask_inv}")
+        cifs_auxsmbconf.append(f"force create mask = 0{share['afp_fperm']}")
+        cifs_auxsmbconf.append(f"force directory mode = 0{share['afp_dperm']}")
+        if share["afp_auxparams"].strip():
+            cifs_auxsmbconf.append(textwrap.indent(f"\nNetatalk Auxiliary Parameters:\n\n{share['afp_auxparams']}", "; "))
```
anodos325, line 65: As we move to clustered registry config for samba, comments like this will be stripped. Maybe we can just drop them since they probably don't apply to SMB shares, or write to a one-off alert that appears after upgrade.

src/middlewared/middlewared/alembic/versions/21.MM/2021-05-25_22-00_afp_to_smb.py
```diff
@@ -0,0 +1,162 @@
...
+
+        cifs_share = {
+            "cifs_purpose": "MULTI_PROTOCOL_AFP",
```
anodos325, line 68: Let's do NO_PRESET here and use the legacy_afp_compatibility checkbox instead.

src/middlewared/middlewared/etc_files/local/avahi/avahi_services.py
```diff
@@ -67,19 +67,16 @@ def _is_running(self):
...
             ))
-            if afp_is_running or smb_is_running:
+            if smb_is_running:
                 return True
             else:
                 return False
 
         if self.service == 'AFPOVERTCP':
-            return any(filter_list(self.service_info, [('service', '=', 'afp'), GENERATE_SERVICE_FILTER]))
+            return any(filter_list(self.service_info, [('service', '=', 'cifs'), GENERATE_SERVICE_FILTER]))
```
anodos325, line 79: We should remove AFPOVERTCP advertisements entirely

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -1398,6 +1402,8 @@ async def pool_post_import(middleware, pool):
             ('path', '^', f'{path}/'),
         ])
     ]):
+        await middleware.call('sharing.smb.disable_acl_if_trivial')
+        await middleware.call('sharing.smb.set_timemachine_quotas')
```
anodos325, line 1406: Have we validated whether this is a correct / supported way of doing time machine quotas over SMB?

src/middlewared/middlewared/alembic/versions/21.MM/2021-05-25_22-00_afp_to_smb.py
themylogin: These are pulled from `MULTI_PROTOCOL_AFP` purpose so migrated shares are identical to `MULTI_PROTOCOL_AFP` samba being created manually, is this a wrong approach?

src/middlewared/middlewared/plugins/smb.py
themylogin: Yes, reading this plist file is protocol agnostic

src/middlewared/middlewared/plugins/smb.py
themylogin: This method only processes migrated shares on first corresponding pool import / dataset unlock and is no-op at other times.

themylogin:
> I'm not sure about adding a time machine quota column. plist file from AFP will still be in share and should be interpreted by MacOS.

What about setting quota for newly created shares?

> I believe adVF in this case needs to be 0x82 otherwise we're advertising that this share is both AFP and SMB.

Won't it be both AFP and SMB?

src/middlewared/middlewared/alembic/versions/21.MM/2021-05-25_22-00_afp_to_smb.py
anodos325: We don't want the auxiliary parameters to be written to the database. The preset only changes the running configuration. This allows us flexibility to fix without doing a migration and continues to move us away from auxiliary parameters.

I think that creating a new column in the cifs table with a name like `legacy_afp_compatibility` is better. Since we no longer have AFP, we can also migrate users away from having that preset as well (i.e. preset -> legacy_afp_compatibillity). Goal is to not have this option presented prominently and not write things to auxiliary parameters.

anodos325:
> > I'm not sure about adding a time machine quota column. plist file from AFP will still be in share and should be interpreted by MacOS.
> 
> What about setting quota for newly created shares?
> 
> > I believe adVF in this case needs to be 0x82 otherwise we're advertising that this share is both AFP and SMB.
> 
> Won't it be both AFP and SMB?

No. Because we're removing AFP. This value is used by time machine to determine what protocols are supported by the particular time machine backup target.

```
        dk(n)=adVF=
        0xa1, 0x81 - AFP support
        0xa2, 0x82 - SMB support
        0xa3, 0x83 - AFP and SMB support
```

src/middlewared/middlewared/plugins/smb.py
anodos325: Good. Just verifying.

src/middlewared/middlewared/plugins/smb_/registry.py
```diff
@@ -249,18 +246,6 @@ async def add_multiprotocol_conf(self, conf, gl, data):
                                            data['id'], {'cifs_durablehandle': False})
                 data['durablehandle'] = False
 
-        if any(filter(lambda x: f"{x['path']}/" in f"{conf['path']}/" or f"{conf['path']}/" in f"{x['path']}/", gl['afp_shares'])):
```
anodos325, line 252: Rather than removing this. We can simply change this check to see whether this is a legacy_afp_share. This will keep logic of only applying parameters to running configuration / not write to DB.

src/middlewared/middlewared/alembic/versions/21.MM/2021-05-25_22-00_afp_to_smb.py
themylogin: So if I get you right you want to remove `MULTI_PROTOCOL_AFP` purpose and add a checkbox that does the same thing instead? Won't this introduce unnecessary conflicts (e.g. checkbox enabled and some non-trivial purpose selected)?

src/middlewared/middlewared/plugins/smb_/registry.py
```diff
@@ -249,18 +246,6 @@ async def add_multiprotocol_conf(self, conf, gl, data):
...
-            self.logger.debug("SMB share [%s] is also an AFP share. "
-                              "Applying parameters for mixed-protocol share.", data['name'])
-            conf.update({
-                "fruit:locking": "netatalk",
-                "fruit:metadata": "netatalk",
-                "fruit:resource": "file",
-                "strict locking": "auto",
-                "streams_xattr:prefix": "user.",
```
anodos325, line 260: remove:
```
fruit:locking = netalk
strict locking = auto
```

As we no longer care about cross-protocol locking.

src/middlewared/middlewared/plugins/smb.py
anodos325: I'm still not fond of the plist. Is this actually documented by Apple as a method of managing time machine size? I'm asking because there is a documented and supported way of setting quotas on time machine backups in samba:

```
fruit:time machine max size = SIZE [K|M|G|T|P]
```

src/middlewared/middlewared/alembic/versions/21.MM/2021-05-25_22-00_afp_to_smb.py
themylogin: I thought we can still have them in database, and user can see them and investigate corresponding samba options if he suddenly finds out something is not working.

Or are we planning on removing samba shares from databases and reading/writing from samba registry directly?

src/middlewared/middlewared/alembic/versions/21.MM/2021-05-25_22-00_afp_to_smb.py
william-gr: @themylogin are we addressing this review too?

themylogin: @william-gr sorry, discussion still pending in Slack

anodos325:
We're still doing quotas in what I think is the wrong way. vfs_fruit has a parameter `fruit:time machine max size = SIZE [K|M|G|T|P]` specifically for this purpose.

src/middlewared/middlewared/alembic/versions/21.MM/2021-05-25_22-00_afp_to_smb.py
anodos325: My recommendation was to remove the MULTI_PROTOCOL_AFP preset and migrate shares with this as well to the new column for cifs_legacy_afp_compat. Then apply the relevant parameters to registry based on this config.

anodos325: Keeping / using a purpose that's explicitly related to AFP support seems to be odd.

src/middlewared/middlewared/alembic/versions/21.MM/2021-05-25_22-00_afp_to_smb.py
```diff
@@ -0,0 +1,163 @@
...
+        if cifs_auxsmbconf:
+            # They will not be automatically appended by the preset in this case
+            cifs_auxsmbconf.extend([
+                "fruit:metadata = netatalk",
+                "fruit:resource = file",
+                "streams_xattr:prefix = user.",
+                "streams_xattr:store_stream_type = no",
+                "oplocks = no",
+                "level2 oplocks = no",
+            ])
```
anodos325, line 66: If we are using new column, then there's no need for aux parameters here.

anodos325:
```
streams_xattr:prefix = user.
streams_xattr:store_stream_type = no
oplocks = no
level2 oplocks = no
```
^^^ migration is still adding these to auxiliary parameters. oplocks related stuff is not required because no AFP.

anodos325:
Migration needs to set `aapl_extensions` to True in Services->SMB if any AFP shares are migrated to SMB.

src/middlewared/middlewared/plugins/smb_/registry.py
```diff
@@ -435,6 +420,15 @@ async def share_to_smbconf(self, conf_in, globalconf=None):
             conf["fruit:time machine"] = "yes"
             conf["fruit:locking"] = "none"
 
+            if data['timemachine_quota']:
+                conf['fruit:time machine max size'] = f'{data["timemachine_quota"]}G'
+
+        if data['afp']:
+            conf['fruit:metadata'] = 'netatalk'
+            conf['fruit:resource'] = 'file'
+            conf['streams_xattr:prefix'] = 'user'
```
anodos325, line 429: That needs to be `user.` with trailing period.

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -923,7 +911,7 @@ async def do_update(self, id, data):
                 except OSError as e:
                     raise CallError(f'Failed to create {path}: {e}')
 
-        if old['purpose'] != new['purpose']:
+        if old['purpose'] != new['purpose'] or old['afp'] != new['afp']:
```
anodos325, line 914: Back out this change now that we're handling in registry.py

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -1313,7 +1301,16 @@ async def apply_presets(self, data):
...
+        if data["afp"]:
+            params = {
+                'acl': True,
+                'aapl_name_mangling': True,
+                'streams': True,
+                'durablehandle': False,
+                'auxsmbconf': '',
+            }
+        else:
+            params = (SMBSharePreset[data["purpose"]].value)["params"].copy()
```
anodos325, line 1313: Back out this change now that we're handling in registry.py

themylogin: @anodos325 do you mean that we need to apply
```
                'acl': True,
                'aapl_name_mangling': True,
                'streams': True,
                'durablehandle': False,
``` in registry.py too ignoring databse options? But we disable `acl` for some of the migrated shares, that will break it.

**Comment by anodos325** [comment]
Since correct behavior of legacy / migrated AFP shares is dependent on SMB service parameter `aapl_extensions`, we must raise a ValidationError if user tries to unset this while this the `afp` box is set on any share.

File: src/middlewared/middlewared/plugins/smb_/registry.py [hunk 13]
```diff
@@ -435,6 +420,15 @@ async def share_to_smbconf(self, conf_in, globalconf=None):
             conf["fruit:time machine"] = "yes"
             conf["fruit:locking"] = "none"
 
+            if data['timemachine_quota']:
+                conf['fruit:time machine max size'] = f'{data["timemachine_quota"]}G'
+
+        if data['afp']:
+            conf['fruit:metadata'] = 'netatalk'
+            conf['fruit:resource'] = 'file'
+            conf['streams_xattr:prefix'] = 'user.'
```
anodos325, line 429: add
```
conf['fruit:encoding'] = 'native'
```

src/middlewared/middlewared/plugins/smb.py
anodos325: No, sorry, my eyes glazed over for a minute. But we do not need to overlay these items via a preset. Setting them to during initial migration is enough.

The only item that must enabled is the encoding. This we set via `fruit:encoding = native` in registry as I suggested elsewhere.

themylogin: > Setting them to during initial migration is enough.

How will these options be forced for newly created AFP shares then?

anodos325: There won't be any newly created AFP shares. If user creates an SMB share with `afp` checked, then the only minor deviation from above will be status of `durablehandle`, which based on some issues users have seen, I may change default of.  `durablehandle = false` is not a requirement for compatibility with legacy AFP shares. It only affects whether our server advertises support for durable handles for that share (ability to survive network disruptions).

Setting `afp` and having `aapl_extensions` is sufficient to force the correct character encoding settings for the share.

themylogin:
backport

themylogin:
backport

themylogin:
backport

themylogin:
backport

themylogin:
backport

src/middlewared/middlewared/alembic/versions/21.MM/2021-05-25_22-00_afp_to_smb.py
```diff
@@ -0,0 +1,167 @@
...
+        cifs_auxsmbconf.append(f"directory mode = 0{umask_inv}")
+        cifs_auxsmbconf.append(f"force create mode = 0{share['afp_fperm']}")
+        cifs_auxsmbconf.append(f"force directory mode = 0{share['afp_dperm']}")
+        if share["afp_auxparams"].strip():
+            cifs_auxsmbconf.append(textwrap.indent(f"\nNetatalk Auxiliary Parameters:\n\n{share['afp_auxparams']}", "; "))
+
+        cifs_share = {
+            "cifs_purpose": "NO_PRESET",
+            "cifs_path": share["afp_path"],
+            "cifs_path_suffix": "",
```
anodos325, line 72: We should set `%U` as a path suffix if `has_cifs_home` and `share["afp_home"]`

This is to approximately preserve behavior of `basedir regex` from afp.conf.

src/middlewared/middlewared/plugins/smb_/registry.py
```diff
@@ -466,6 +451,16 @@ async def share_to_smbconf(self, conf_in, globalconf=None):
...
 
+            if data['timemachine_quota']:
+                conf['fruit:time machine max size'] = f'{data["timemachine_quota"]}G'
+
+        if data['afp']:
+            conf['fruit:encoding'] = 'native'
+            conf['fruit:metadata'] = 'netatalk'
+            conf['fruit:resource'] = 'file'
+            conf['streams_xattr:prefix'] = 'user.'
+            conf['streams_xattr:store_stream_type'] = 'no'
```
anodos325, line 462: need `conf['streams_xattr:xattr_compat'] = 'true'` per NAS-110888

src/middlewared/middlewared/alembic/versions/21.MM/2021-05-25_22-00_afp_to_smb.py
```diff
@@ -0,0 +1,174 @@
...
+            cifs_auxsmbconf.append(f"rolist = {share['afp_ro'].strip()}")
+            share_disable_acl_if_trivial = True
+        if share["afp_rw"].strip():
+            cifs_auxsmbconf.append(f"rwlist = {share['afp_rw'].strip()}")
+            share_disable_acl_if_trivial = True
+        umask_inv = "%o" % (~int(share["afp_umask"], 8) & 0o777)
+        cifs_auxsmbconf.append(f"create mask = 0{umask_inv}")
+        cifs_auxsmbconf.append(f"directory mode = 0{umask_inv}")
+        cifs_auxsmbconf.append(f"force create mode = 0{share['afp_fperm']}")
+        cifs_auxsmbconf.append(f"force directory mode = 0{share['afp_dperm']}")
```
yocalebo, line 65: This doesn't look right. @anodos325 can you confirm?

yocalebo: Confirmed with @anodos325 this looks correct.

anodos325 (approved):
@william-gr, I'm approving this but things may have changed significantly since you first looked at it. You may wish to give it another once-over.

src/middlewared/middlewared/alembic/versions/21.MM/2021-05-25_22-00_afp_to_smb.py
themylogin: @yocalebo yeah, it was tested, samba share behaves as intended

yocalebo (approved):

william-gr (approved):

PR #6921, opened 2021-05-26
NAS-110793 / 12.0 / remove lxml and xml.etree from identify_freebsd

yocalebo:
- remove `lxml` and `xml` from `identify_freebsd` and use the built-in `geom.class_by_name.xml` attribute
- while here, remove unused `identify_linux` and `device_info_linux` modules (1 of which actually imports `lxml`)
- if a `{serial}` was requested to be identified, we were calling `geom.class_by_name.xml` twice for which I've made it so that we only call it once


src/middlewared/middlewared/plugins/disk_/identify_freebsd.py
```diff
@@ -68,32 +64,32 @@ def identifier_to_device(self, ident, disks=None):
...
-                './/provider/config['
-                f'normalize-space(ident) = normalize-space("{value}")'
-                ']/../../name'
-            )
-            if len(search) > 0:
-                return search[0].text
+
+            # remove leading/trailing and more than single whitespace char(s)
+            _value = ' '.join(value.split())
+            search = xml.find(f'.//provider/config[ident = "{_value}"]/../../name')
```
themylogin, line 74: Are we sure that XML document values are always whitespace-normalized?

yocalebo: It's probably not, so I've fixed the logic.

rick-mesta:
@themylogin / @yocalebo : What's the status on this one, gents ? Are all issues resolved ?

yocalebo:
> @themylogin / @yocalebo : What's the status on this one, gents ? Are all issues resolved ?

I've addressed the review, just waiting on approval/re-review from @themylogin

themylogin (approved):

rick-mesta:
Noice ! Thx @themylogin 😁

PR #6922, opened 2021-05-26
NAS-110794 / 21.06 / Allow inheriting values explicitly during dataset creation in API

sonicaj:


Ornias1993 (approved):
Yes, this is what I meant initially, thanks for fixing it up @sonicaj :)

Ornias1993:
@themylogin You can check this one instead of my PR, my PR definately has some issues that are fixed here ^^

themylogin (approved):

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3079,36 +3081,36 @@ async def do_create(self, app, data):
...
+            ('snapdir', None, str.lower, False),
+            ('sparse', None, None, False),
+            ('sync', None, str.lower, False),
+            ('volblocksize', None, None, False),
+            ('volsize', None, lambda x: str(x), False),
+            ('xattr', None, str.lower, False),
+            ('special_small_block_size', 'special_small_blocks', None, True),
         ):
-            if i not in data:
+            if i not in data or (inheritable and data[i] == 'INHERIT'):
```
themylogin, line 3113: Can't you just remove all inheritable with `INHERIT` value from `data` instead of introducing such a large diff?

How is `special_small_block_size` different from any other dataset properties in this case? Why do we allow it to be explicitly inherited while others are inherited when they are not specified?

themylogin (approved):

PR #6924, opened 2021-05-26
NAS-110783 / 12.0 / only unlock SED disks on active controller

yocalebo:


src/freenas/etc/ix.rc.d/ix-sed
```diff
@@ -11,7 +11,13 @@
 
 unlock_drives()
 {
-	LD_LIBRARY_PATH=/usr/local/lib /usr/local/bin/midclt call disk.sed_unlock_all > /dev/null
+	# carp event script in the failover plugin has to call this
+	# since running this method on both controllers at the same
+	# time (on HA systems) can cause, ultimately, the disks to
+	# not get unlocked
+	if [ "$(ha_licensed)" = "false" ]; then
```
william-gr, line 18: will ha_license even be available here?

yocalebo: I'd assume so since `ha_licensed` just calls the `hadetect` standalone script which exists on `/usr/local/bin` and we're calling `jq` which also exists on `/usr/local/bin`

william-gr: Oh right, I confused that with the shell script function, nevermind me

yocalebo: Do I need to `. /etc/rc.freenas` in this rc script or will `. /etc/rc.subr` somehow source that function?

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
```diff
@@ -519,6 +519,14 @@ def carp_master(self, fobj, ifname, vhid, event, user_override, forcetakeover):
                 if attach_all_job.error:
                     logger.error('Failed to attach geli providers: %s', attach_all_job.error)
 
+                try:
+                    self.middleware.call_sync('disk.sed_unlock_all')
```
william-gr, line 523: I wonder if there may be any implications here.

In theory this only needs to be called once, on boot. There is really no reason to call that when actually taking over, since disks will already be unlocked.

yocalebo: We can't call this at boot time because we don't know which controller is `MASTER` or `BACKUP`. That's why I'm calling it in the event script.

william-gr: You must have understood me. Yes, I got that, but I mean we will call it every time a takeover happens, not only on boot.

yocalebo: I don't see a way around that. `sed_unlock_all` will call `sedutil --query` to see if the disk(s) is locked before trying to unlock it? There will be a significant delay in failover event if this is called and the disks end up being locked. (i.e. both controllers are powered on and HA is configured). But again, I don't see a way around that?

william-gr: The way around that would be somehow have a flag around to only call it if we are in the boot process, prior to enabling CARP.

william-gr (approved):

william-gr:
Thanks

PR #6927, opened 2021-05-27
NAS-110803 / 21.06 / Do not treat mountpoint specially

sonicaj:
This PR introduces changes to make sure we are consistent in our usage wrt consuming `mountpoint` property.
Before py-libzfs used to return `mountpoint` key/value always even if it did or did not return the actual value of the property. This can be confusing and lead to errors/mistakes - the changes introduced here expect the key to not be present anymore and instead rely on the property to be retrieved explicitly and not make an implicit expectation that it's going to be there.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2828,6 +2828,7 @@ class PoolDatasetService(CRUDService):
                 ('org.freenas:refquota_critical', 'refquota_critical', None),
                 ('org.truenas:managedby', 'managedby', None),
                 ('dedup', 'deduplication', str.upper),
+                ('mountpoint', None, _null),
                 ('aclmode', None, str.upper),
                 ('acltype', None, str.upper),
                 ('xattr', None, str.upper),
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2864,6 +2864,12 @@ class PoolDatasetService(CRUDService):
                 if method:
                     dataset[i]['value'] = method(dataset[i]['value'])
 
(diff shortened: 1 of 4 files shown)


themylogin (approved):

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2863,6 +2864,12 @@ def transform(dataset):
                 if method:
                     dataset[i]['value'] = method(dataset[i]['value'])
 
+            if 'mountpoint' in dataset:
+                # This is treated specially to keep backwards compatibility with API
+                dataset['mountpoint'] = dataset['mountpoint']['value']
```
william-gr, line 2869: I think mountpoint should always be there, we cannot have top level properties that may or may not be there, its very misleading and backward incompatible.

sonicaj: So if mountpoint property was not retrieved explicitly, what should we say for `mountpoint` property ?

sonicaj: Ability to specify custom properties was introduced recently - which means that for old users/releases, `mountpoint` property would continue to exist always

william-gr: Recently as in SCALE? not 12?

sonicaj: @william-gr, yes that is accurate - for `pool.dataset.query`, retrieving specific properties was introduced in SCALE

william-gr: Ok, then sounds reasonable, thanks for clarifying!

william-gr (approved):

PR #6929, opened 2021-05-28
Bug fix for creating a volume

sonicaj:


src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -717,7 +717,9 @@ def check_key(self, id, options):
     )
     def unload_key(self, id, options):
         force = options.pop('force_umount')
-        if options.pop('umount') and self.middleware.call_sync('zfs.dataset.get_instance', id)['mountpoint']:
+        if options.pop('umount') and self.middleware.call_sync(
+            'zfs.dataset.query', [['id', '=', id]], {'extra': {'retrieve_children': False}}
```
themylogin, line 721: Can we allow passing extra options to `get_instance` instead?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2905,18 +2903,18 @@ def transform(dataset):
         ]),
         Bool('sparse'),
         Bool('force_size'),
-        Inheritable(Str('comments')),
-        Inheritable(Str('sync', enum=['STANDARD', 'ALWAYS', 'DISABLED'])),
-        Inheritable(Str('compression', enum=ZFS_COMPRESSION_ALGORITHM_CHOICES)),
+        Inheritable(Str('comments'), default='INHERIT'),
```
themylogin, line 2906: How was it crashing without `default`?

themylogin (changes requested):

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2905,18 +2903,18 @@ def transform(dataset):
...
+        Inheritable(Int('quota_warning', validators=[Range(0, 100)]), default='INHERIT'),
+        Inheritable(Int('quota_critical', validators=[Range(0, 100)]), default='INHERIT'),
         Int('refquota', null=True, validators=[Or(Range(min=1024**3), Exact(0))]),
-        Inheritable(Int('refquota_warning', validators=[Range(0, 100)])),
-        Inheritable(Int('refquota_critical', validators=[Range(0, 100)])),
+        Inheritable(Int('refquota_warning', validators=[Range(0, 100)]), default='INHERIT'),
+        Inheritable(Int('refquota_critical', validators=[Range(0, 100)]), default='INHERIT'),
         Int('reservation'),
         Int('refreservation'),
         Inheritable(Int('special_small_block_size')),
```
themylogin, line 2920: Don't you also need `default` here?

themylogin (changes requested):

src/middlewared/middlewared/plugins/pool.py
sonicaj: No, this property is not valid for volumes

themylogin: So it works without `default`? Why do we need it then?

sonicaj: `INHERIT` is a no-op, zfs create api will inherit if it's not specified explicitly. It's just been defined to be explicit rather then implicit.

themylogin (approved):

PR #6937, opened 2021-06-02
NAS-110616 / 21.08 / Print "Press Enter to bring Console setup menu." after Linux console …

themylogin:
…gets spammed with kernel messages

### DIFF
--- a/src/freenas/etc/netcli
+++ b/src/freenas/etc/netcli
@@ -35,6 +35,7 @@ import re
 import signal
 import sys
 import subprocess
+import threading
 import time
 import traceback
 
@@ -1510,7 +1511,13 @@ def netcli_title():
     return f'{product} | {version}{serial}'
 
 
+is_in_main_menu_prompt = False
+main_menu_initialized = False
+
+
 def main_menu():
+    global is_in_main_menu_prompt, main_menu_initialized
+
     menu = [
         [_("Configure Network Interfaces"), configure_interface_stub],
         [_("Configure Link Aggregation"), configure_lagg_interface],
@@ -1594,6 +1601,9 @@ def main_menu():
                     time.sleep(5)
                     continue
 
+        is_in_main_menu_prompt = True
+        main_menu_initialized = True
+
         if is_enterprise:
             print(netcli_title())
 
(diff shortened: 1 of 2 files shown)


src/freenas/etc/netcli
```diff
@@ -1619,17 +1629,43 @@ def main_menu():
...
+                last_kernel_message = time.monotonic()
+
+
+def press_enter_to_bring_console_setup_menu():
+    global last_kernel_message
+    while True:
+        if last_kernel_message is not None and time.monotonic() - last_kernel_message >= 3:
+            last_kernel_message = None
+            if is_in_main_menu_prompt:
+                print("Press Enter to bring Console setup menu.")
```
yocalebo, line 1658: Am I understanding correctly that this now gets printed to the console every 3 seconds?

src/freenas/etc/netcli
```diff
@@ -1619,17 +1629,43 @@ def main_menu():
...
 
 
+last_kernel_message = None
+
+
+def read_kmsg():
+    global last_kernel_message, main_menu_initialized
+    with open("/dev/kmsg") as f:
+        while True:
+            k = f.readline()
```
yocalebo, line 1647: why `readline()` if we don't do anything with it?

william-gr:
Just realize we really need to do this in CLI. By 21.08 netcli will be gone and replaced with midcli.

src/freenas/etc/netcli
themylogin: No, it only gets printed after we receive kernel messages (in 3 seconds after last message). Then `last_kernel_message` will be `None` until a new message is received

rick-mesta:
@yocalebo: has Vlad addressed all issues to your satisfaction ?

yocalebo:
> @yocalebo: has Vlad addressed all issues to your satisfaction ?

No, @william-gr chimed in and said this needs to be done in `midcli` and not via `netcli` since it's expected that `netcli` will be removed in 21.08 so this PR can more than likely be closed.

rick-mesta:
oh... thx for the ⚡ fast reply, Caleb. @william-gr / @themylogin : you guys okay with closing this PR so the fix can be done in `midcli` ?

william-gr:
yup

rick-mesta:
As mentioned by both @william-gr and @yocalebo, this work should be delivered as part of `midcli`. Closing this PR.

PR #6940, opened 2021-06-03
NAS-110877 / 21.08 / Create storage class before doing helm action

sonicaj:


Ornias1993 (approved):
Good catch, clean fix.
LGTM

themylogin (approved):

PR #6941, opened 2021-06-03
NAS-110515 / 21.08 / Use Tdie for some AMD CPU models

themylogin:


rick-mesta (approved):
Looks like these changes will definitely take into account the offending AMD CPU models, while leaving the rest of the behavior unscathed 👍🏼 . @yocalebo or @sonicaj -- would either/both of you gents also please take a once-over of this PR ? 🙏🏼

src/middlewared/middlewared/plugins/reporting/events.py
```diff
@@ -366,6 +333,61 @@ def _cpu_temperature(self):
...
+        "AMD Ryzen Threadripper 19",
+        "AMD Ryzen Threadripper 29",
+    )
+
+    def _amd_cpu_temperature(self, amd_sensor):
+        if self.AMD_CPU_MODEL is None:
+            self.AMD_CPU_MODEL = self.middleware.call_sync('system.info')['model']
+
+        if self.AMD_CORE_COUNT is None:
+            self.AMD_CORE_COUNT = len([
```
yocalebo, line 354: We already cache the CPU core count and the CPU model in `system.cpu_info` so I'd rather use that instead since it removes a subprocess call.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/reporting/events.py
```diff
@@ -366,6 +333,54 @@ def _cpu_temperature(self):
...
+        "AMD Ryzen Threadripper 29",
+    )
+    AMD_SYSTEM_INFO = None
+
+    def _amd_cpu_temperature(self, amd_sensor):
+        if self.AMD_SYSTEM_INFO is None:
+            self.AMD_SYSTEM_INFO = self.middleware.call_sync('system.info')
+
+        cpu_model = self.AMD_SYSTEM_INFO['model']
+        core_count = self.AMD_SYSTEM_INFO['physical_core_count']
```
yocalebo, line 353: keyerror here `physical_cores`

yocalebo (changes requested):

yocalebo (approved):

PR #6945, opened 2021-06-03
NAS-110887 / 21.08 / enable STP by default when creating bridges on SCALE

yocalebo:
FreeBSD bridges have STP/RSTP enabled by default for obvious reasons. Mirror the behavior on SCALE.


rick-mesta (approved):

Ornias1993:
Does this fix also apply itself on update if the bridge existed before the update?

PR #6950, opened 2021-06-04
NAS-110792 / 21.08 / Fix active directory user token generation

anodos325:
*Background*
In a Windows / AD environment, all objects are identified by
SID. SIDs for user / group objects take the form
`S-1-5-21-<domain>-<rid>`

The <domain> component for a SID will vary depending on
whether the account is a local one or one from an external domain.
A domain sid is of the form `S-1-5-21-<domain>`, and every
TrueNAS server has a unique domain sid, which is randomly
generated by samba libraries when it is first needed.

Each relative id (rid) uniquely identifies an object in the domain.
Certain rids are present in every domain. Of particular note
for this commit are the following:

```
S-1-5-21-<domain>-501 - Guest
S-1-5-21-<domain>-512 - Domain Admins
S-1-5-21-<domain>-513 - Domain Users
S-1-5-21-<domain>-514 - Domain Guests
```

In addition to domain (S-1-5-21) sids, every Windows computer
and Samba server has sids that are identical on every machine
that are prefixed with S-1-5-32 (built-in groups). Of particular note
for this commit are the following:
```
S-1-5-32-544 - Adminstrators
S-1-5-32-545 - Users
S-1-5-32-546 - Guests
```

Since Unix-like servers use uids / gids (xids) to identify users and
groups rather than sids, sids must be mapped into xids and vice-versa.
This task falls on Samba's passdb, groupdb, and winbindd's idmapping
facilities.

During samba startup, if samba's groupdb lacks entries for
Administrators, Users, and Guests, then they will be automatically
added by allocating new gids for each of them from winbindd's idmap
backend that has been configured to provide mappings for built-in
sids. This allocation increments the xid high-water mark in
winbindd_idmap.tdb, (but does not write the explicit mapping in
the key-value store) and then writes the explicit mapping in
group_mapping.tdb.

Windows has the concept of nested groups. Groups in Windows may
have members that are either users or groups. Accordingly, each
groupmap entry in group_mapping.tdb may have zero or more
foreign memberships in it. The following is a sample tdb entry:

```
{
key(23) = "UNIXGROUP/S-1-5-32-546\00"
data(32) = "\83J]\05\04\00\00\00Guests\00Local Unix group\00"
}
{
key(54) = "MEMBEROF/S-1-5-21-944110568-1438105595-1944063070-514\00"
data(13) = "S-1-5-32-546\00"
}
```

In this case, S-1-5-32-546 is mapped to gid 90,000,002 and has
a foreign member of S-1-5-21-944110568-1438105595-1944063070-514.

During the domain-join process, libads adds domain sids as members
of the above built-in groups:
```
S-1-5-21-<new domain>-512 --> S-1-5-32-544
S-1-5-21-<new domain>-513 --> S-1-5-32-545
S-1-5-21-<new domain>-514 --> S-1-5-32-546
```

Which means that when nss_winbind generates a passwd struct for
a domain user, BUILTIN\Users is added to the grouplist with the
gid listed in the group_mapping.tdb.

*Problem*
There are various situations that can occur where original
mapping for builtins is lost or remapped to different ids
foreign memberships are lost, or id collisions are generated
with other groups allocated in winbindd_idmap.tdb.

Although these built-in groups are not exposed via middleware and
the webui, they are exposed via the SMB protocol and Samba's
RPC endpoints. The most common reason for them to be used is
when robocopy is used to migrate data from a Windows share on
Windows server where the system administrator is using
built-in groups rather than AD groups to share data.

*Impact*
Impact of potential indetermenancy with the mapping potentially
profound. User tokens may be generated with incorrect ids,
and filesystem ACLs may cease to grant expected access.
If foreign group membership is dropped, then built-in groups
will not appear in passwd entries for AD and local users.

*Resolution*
When built-in groups are handled by idmap_tdb (default),
ensure that Administrators, Users, and Guests are mapped
explicitly to the lowest three gids in the range that is
specified for the default domain (*). Use newly-added
json-based batch operations for groupmap to achieve this.
net_groupmap text variant does not allow direct manipulation
of gids in the groupmap file, which creates a chicken-and-egg
problem for mapping built-ins explicitly when there is no
corresponding winbindd_idmap.tdb entry.

Adjust high-water-mark in idmap_tdb to never allocate gids
in this reserved area.

Ensure that local builtin_users, builtin_admins, builtin_guests
groups are mapped to the respective local domain (TrueNAS)
domain users, domain admins, domain guests SIDs.

Ensure that foreign mappings for Administrators, Users, and Guests
always exist.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb_/groupmap.py
+++ b/src/middlewared/middlewared/plugins/smb_/groupmap.py
@@ -4,9 +4,14 @@ from middlewared.utils import run
 from middlewared.plugins.smb import SMBCmd, SMBBuiltin, SMBPath
 
 import os
-import re
+import json
+import tdb
+import struct
 
-RE_NETGROUPMAP = re.compile(r"^(?P<ntgroup>.+) \((?P<SID>S-[0-9\-]+)\) -> (?P<unixgroup>.+)$")
+# This follows JSON output version for net_groupmap.c
+# Output format may change between this and final version accepted
+# upstream, but Samba project has standardized on following version format
+GROUPMAP_JSON_VERSION = {"major": 0, "minor": 1}
 
 
 class SMBService(Service):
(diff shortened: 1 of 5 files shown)


anodos325:
This fix is dependent on some pending changes to `source3/utils/net_groupmap.c` in Samba. This will need to wait until those changes are finalized and merged.

src/middlewared/middlewared/plugins/smb_/groupmap.py
```diff
@@ -5,144 +5,348 @@
...
-    async def groupmap_add(self, group, passdb_backend=None):
+    async def sync_foreign_groups(self):
         """
-        Map Unix group to NT group. This is required for group members to be
-        able to access the SMB share. Name collisions with well-known and
-        builtin groups must be avoided. Mapping groups with the same
-        names as users should also be avoided.
+        Domain Users, and Domain Admins must have S-1-5-32-545 and S-1-5-32-544
+        added to their respective Unix tokens for correct behavior in AD domain.
+        This are added by making them foreign members in the group_mapping for
```
yocalebo, line 66:
```suggestion
        These are added by making them foreign members in the group_mapping for
```

src/middlewared/middlewared/plugins/smb_/groupmap.py
```diff
@@ -15,134 +20,334 @@ class Config:
...
-        gm_add = await run(
-            [SMBCmd.NET.value, '-d', '0', 'groupmap', 'add', 'type=local', f'rid={next_rid}', f'unixgroup={group}', f'ntgroup={group}'],
-            check=False
-        )
-        if gm_add.returncode != 0:
-            raise CallError(
-                f'Failed to generate groupmap for [{group}]: ({gm_add.stderr.decode()})'
+        if ad_state == 'FAULTED':
+            self.logger.debug(
+                "Unable to validate foreign group members for builting groups"
```
yocalebo, line 111:
```suggestion
                "Unable to validate foreign group members for built-in groups"
```

src/middlewared/middlewared/plugins/smb_/groupmap.py
```diff
@@ -15,134 +20,334 @@ class Config:
...
+        return must_reload
+
+    @private
+    async def batch_groupmap(self, data):
+        for op in ["ADD", "MOD", "DEL"]:
+            if data.get(op) is not None and len(data[op]) == 0:
+                data.pop(op)
+
+        payload = json.dumps(data)
+        out = await run([SMBCmd.NET.value, '--json', 'groupmap', 'batch_json', f'data={payload}'], check=False)
```
yocalebo, line 285: I'm assuming that if you have a situation where there is a `DEL` and an `ADD` operation, the underlying `batch_json` operation will process those in order? Or does order not matter?

yocalebo (commented):
Minor nit-picks and a general question.

src/middlewared/middlewared/plugins/smb_/groupmap.py
anodos325: Right. We iterate through a lookup table with function pointers:
```
const struct {
	const char *name;
	int (*fn)(struct net_context *c, int argc, const char **argv);
} optable[] = {
	{ "DEL", net_groupmap_delete },
	{ "ADD", net_groupmap_add },
	{ "MOD", net_groupmap_modify },
};
```

Then use `name` to attempt to retrieve a json array of data to iterate through and pass to `fn`. So we get "DEL", then "ADD", then "MOD".

rick-mesta:
Kudos on the most excellent write-up I've seen in years, @anodos325 👍🏼 🎉

rick-mesta (commented):
Other than the obvious `if`/`else` clause pattern that I'm having trouble grokking, this is 👍🏼

src/middlewared/middlewared/plugins/smb_/groupmap.py
```diff
@@ -15,134 +20,334 @@ class Config:
...
 
-        if passdb_backend != 'tdbsam':
-            return
+        # First add our local users, admins, and guests to the builtins
+        admins = await self.groupmap_listmem("S-1-5-32-544")
+        if groupmap['local_builtins'][544]['sid'] not in admins:
+            to_add = groupmap['local_builtins'][544]['sid']
+            await self.groupmap_addmem("S-1-5-32-544", to_add)
+        else:
+            admins.remove(groupmap['local_builtins'][544]['sid'])
```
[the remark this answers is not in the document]
anodos325: Post-lunch crazy pills. Fixed.

src/middlewared/middlewared/plugins/smb_/groupmap.py
```diff
@@ -15,134 +20,367 @@ class Config:
...
+                new_hwm_bytes = struct.pack("<L", group_hwm_bytes)
+                tdb_handle.store(b'GROUP HWM\00', new_hwm_bytes)
+                tdb_handle.transaction_commit()
+                must_reload = True
+
+            for key in tdb_handle.keys():
+                if key[:3] == b'GID' and int(key.decode()[4:-3]) < (low_range + 2):
+                    reverse = tdb_handle.get(key)
+                    tdb_handle.transaction_start()
+                    tdb_handle.delete(key)
```
anodos325, line 200: May want to snapshot underlying samba4 dataset before touching tdb file.

rick-mesta (approved):

yocalebo (approved):

PR #6957, opened 2021-06-06
NAS-110929 / 21.08 / Allow retrieving catalog data partially

sonicaj:
As community usage has grown, we are seeing lots of trains with a decent number of applications with decent number of versions in a catalog, this results in middleware taking considerable amount of time to reading the data and sending it to the UI. Even with caching, we get slow at json dumping/loading which impacts UI performance.

This PR intends to introduce the following changes:
1) Allow retrieving information of some specific trains only
2) Add `latest_version` key to item level payload so that consumer is able to know the latest healthy item version without going through the payload
3) Allow skipping retrieving versions of an item which helps improve performance for json dump/load operations
4) Introduce a new endpoint to retrieve version details of an item specifically without querying up everything

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/items.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/items.py
@@ -24,7 +24,7 @@ class CatalogService(Service):
             'options',
             Bool('cache', default=True),
             Bool('retrieve_all_trains', default=True),
-            Bool('skip_retrieving_versions', default=False),
+            Bool('retrieve_versions', default=True),
             List('trains', items=[Str('train_name')]),
         )
     )
@@ -41,7 +41,7 @@ class CatalogService(Service):
         `options.trains` is a list of train name(s) which will allow selective filtering to retrieve only information
         of desired trains in a catalog. If `options.retrieve_all_trains` is set, it has precedence over `options.train`.
 
-        `options.skip_retrieving_versions` can be set to skip retrieving version details of each catalog item. This
+        `options.retrieve_versions` can be unset to skip retrieving version details of each catalog item. This
         can help in cases to optimize performance.
         """
         catalog = self.middleware.call_sync('catalog.get_instance', label)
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/catalogs_linux/items.py
```diff
@@ -23,6 +23,9 @@ class Config:
         Dict(
             'options',
             Bool('cache', default=True),
+            Bool('retrieve_all_trains', default=True),
+            Bool('skip_retrieving_versions', default=False),
```
themylogin, line 27: Can we make it `Bool('retrieve_versions', default=True),` to avoid double negative?

themylogin (changes requested):

Ornias1993:
On top of `4.` it might be worthwhile to do the same for catalogs:
- Store a basic summary for each train (apps, last versions, icon urls, maybe health), storing on refresh
- Fetch this basic summary when loading the Apps UI, instead of actually requesting data for individual Apps.
- Maybe cache this on the UI client, with a incremental and purge on refresh.

Afaik thats a more efficient/quicker way of getting this basic information into the apps page.

sonicaj:
@Ornias1993 it's already happening like that right now :)

Ornias1993:
> @Ornias1993 it's already happening like that right now :)

You're right... I could've just spend the time to look at the code and not just your summary >.<
```
    @accepts(
        Str('item_name'),
        Dict(
            'item_version_details',
            Bool('cache', default=True),
            Str('catalog', required=True),
            Str('train', required=True),
        )
    )
```

#FaceHitsWall

rick-mesta:
@themylogin: can't quite decipher if the issue you raised got resolved... would you mind taking another look and give a 👍🏼 or 👎🏼 ?

themylogin (approved):

rick-mesta:
Thx Vlad 🎉

PR #6970, opened 2021-06-07
NAS-110949 / 21.08 / Improve retrieving installed application(s) performance

sonicaj:
This PR adds changes to improve the speed for retrieving installed applications. With these changes we see around 30% improvement in the speed to installed chart releases.


themylogin (approved):

Ornias1993 (approved):
In this case these improvements are enough to ensure relatively quick loading of installed Apps. As the amount of expected Apps users are expected to Install is no-where close to the amounts that is going to be available in the Catalogs themselves (so we can get away with this simpler solution) :)

Nice catch 👍

PR #6972, opened 2021-06-07
NAS-110950 / 21.08 / remove unused files referencing lxml

yocalebo:
Remove files that reference `lxml` that aren't used on SCALE but on CORE only.


rick-mesta (approved):
This is great... but definitely get a second opinion from either @themylogin , @sonicaj or @anodos325 😁

themylogin (approved):

PR #6983, opened 2021-06-07
NAS-110955 / 21.08 / Add regression tests for AFP/SMB migration param

anodos325:
These tests ensure that the `afp` key is being
properly handled in smb.conf generation so that
no xattr truncation occurs and Finder metadata
is properly preserved. This is accomplished via
a series of tests using a selection of xattrs
from MacOS client files.

First, the selection of xattrs is iterated through
and written to a test file.

Then, the xattrs are read through the SMB protocol
with the relevant name transformations applied. In
this case the Netatalk metadata xattr is renamed to
AFP_AFPInfo over SMB protocol, and `:` character is
replaced with `U\F022`. Netatalk metadata xattr
contents are transformed into a 60-byte AFP_AFPInfo
xattr as well.

Then the xattrs are unlinked.

THen the xattrs are rewritten via SMB protocol and
read via SSH (and resulting length / contents verified).

### DIFF
--- a/tests/api2/test_425_smb_protocol.py
+++ b/tests/api2/test_425_smb_protocol.py
@@ -5,6 +5,7 @@ import sys
 import os
 import enum
 from time import sleep
+from base64 import b64decode, b64encode
 apifolder = os.getcwd()
 sys.path.append(apifolder)
 from functions import PUT, POST, GET, DELETE, wait_on_job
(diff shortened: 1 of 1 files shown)


anodos325:
`@pytest.mark.parametrize('xat', AFPXattr.keys())`

Means that the method will be repeated with each of the AFPXattr keys passed in as an argument.

themylogin (approved):

rick-mesta (approved):

ericbsd (approved):
LGTM

PR #6984, opened 2021-06-08
NAS-110116 / 21.08 / Document return type(s) for api

sonicaj:


src/middlewared/middlewared/main.py
```diff
@@ -1643,6 +1644,8 @@ def main():
     parser.add_argument('--loop-debug', action='store_true')
     parser.add_argument('--trace-malloc', '-tm', action='store', nargs=2, type=int, default=False)
     parser.add_argument('--overlay-dirs', '-o', action='append')
+    parser.add_argument(f'--enable-debug-mode', dest='debug_mode', action='store_true')
```
themylogin, line 1647: Unnecessary `f` here. And do we need `--enable-debug-mode` option?

src/middlewared/middlewared/plugins/account.py
```diff
@@ -381,7 +397,7 @@ async def do_create(self, data):
                 self.logger.warn('Failed to update authorized keys', exc_info=True)
                 raise CallError(f'Failed to update authorized keys: {e}')
 
-        return pk
+        return await self.get_instance(pk)
```
themylogin, line 400: As much as I wish I could do this too, this is a breaking of API 2.0

Now I think that with the appearance of documentation, we may introduce API 2.1 where we'll fix all piled up things like this. We'll only document return schemas for API 2.1, and for API 2.0 we'll continue returning integers.

I am not sure how much of a change that will be, I am writing this only having reviewed the code so far, so it's just a proposal, let's investigate further.

src/middlewared/middlewared/plugins/account.py
```diff
@@ -588,9 +604,10 @@ async def do_delete(self, pk, options):
         await self.middleware.call('datastore.delete', 'account.bsdusers', pk)
         await self.middleware.call('service.reload', 'user')
 
-        return pk
+        return True
 
     @accepts(Int('user_id', default=None, null=True))
+    @returns(Dict('shell_choices', additional_attrs=True))
```
themylogin, line 610: For dictionaries with `additional_attrs=True` we might need something like `example={"bash": "/bin/bash", "zsh": "/usr/bin/zsh"}`, otherwise it's not much informative.

src/middlewared/middlewared/plugins/account.py
```diff
@@ -694,6 +727,7 @@ async def get_next_uid(self):
 
     @no_auth_required
     @accepts()
+    @returns(Bool('root_has_password'))
```
themylogin, line 730: It might be boring to always give names to return values, especially in cases like this when they just rephrase the method name, should we try to skip that?

themylogin (changes requested):

src/middlewared/middlewared/plugins/account.py
```diff
@@ -132,6 +132,22 @@ class Config:
...
+        ('rm', {'name': 'group_create'}),
+        ('rm', {'name': 'home_mode'}),
+        ('rm', {'name': 'password'}),
+        ('add', Dict('group', additional_attrs=True)),
+        ('add', Int('id')),
+        ('add', Bool('builtin')),
+        ('add', Bool('id_type_both')),
+        ('add', Bool('local')),
+        ('add', Str('unixhash')),
+        ('add', Str('smbhash')),
```
anodos325, line 148: Directory services users don't return `unixhash` or `smbhash` for obvious reasons.

sonicaj: @anodos325 that is good as they are not marked as `required`, i am although unsure if more fields can be added to the payload under certain flags/circumstances when ad is involved

themylogin (changes requested):

themylogin (approved):

PR #6996, opened 2021-06-08
NAS-108383 / 21.08 / API key allowlist

themylogin:


rick-mesta:
@yocalebo / @anodos325 / @sonicaj: can I get any of you guys to give Vlad some feedback on this bad boy ?

tests/api2/test_012_api_key.py
```diff
@@ -0,0 +1,135 @@
...
+    key = results.json()
+
+    try:
+        yield key["key"]
+    finally:
+        results = DELETE(f"/api_key/id/{key['id']}/")
+        assert results.status_code == 200, results.text
+
+
+def test_root_api_key_websocket():
```
anodos325, line 34: I believe convention here is to number API tests. Short docstring blurb explaining what's being tested may be helpful for QE.

tests/api2/test_012_api_key.py
```diff
@@ -0,0 +1,135 @@
...
+    with api_key([]) as key:
+        cmd = f"midclt -u ws://{ip}/websocket --api-key {key} call system.info"
+        results = SSH_TEST(cmd, user, password, ip)
+        assert results['result'] is True, str(results['output'])
+        assert 'uptime' in str(results['output'])
```
anodos325, line 39: May want to print output.

tests/api2/test_012_api_key.py
```diff
@@ -0,0 +1,135 @@
...
+import os
+import sys
+
+import requests
+
+apifolder = os.getcwd()
+sys.path.append(apifolder)
+from functions import POST, GET, DELETE, PUT, SSH_TEST, wait_on_job
+from auto_config import pool_name, scale, ha, password, user, ip
+
```
anodos325, line 19: Is this a SCALE-only feature? If so, you may want to skip this test if not SCALE

tests/api2/test_012_api_key.py
themylogin: Why? Assert prints both sides of the `in` expression if it fails
```
______________________________________________________________________________________ test_a ______________________________________________________________________________________

    def test_a():
        x = 'aaa'
        y = 'bbb'
>       assert x in y
E       AssertionError: assert 'aaa' in 'bbb'

1.py:4: AssertionError
```

tests/api2/test_012_api_key.py
themylogin: We don't run this branch on CORE

tests/api2/test_012_api_key.py
anodos325: Sorry. Didn't think this one through.

tests/api2/test_012_api_key.py
themylogin: I believe they number API tests because their execution order matters (not a good practice IMO). This module's tests can be executed in any order.

Docstrings added.

william-gr:
Sorry, I completely forgot about this PR. Please wait for me to review it in the next few days too.

src/middlewared/middlewared/plugins/api_key.py
```diff
@@ -36,7 +45,14 @@ async def item_extend(self, item):
...
-            register=True
+            List("allowlist", items=[
+                Dict(
+                    "allowlist_item",
+                    Str("method", required=True, enum=["GET", "POST", "PUT", "DELETE", "CALL", "SUBSCRIBE"]),
+                    Str("resource", required=True),
+                ),
+            ]),
+            register=True,
         )
```
anodos325, line 56: Won't this be O(N) lookups on each API request? What if we use dictionary `<resource>: {"GET": True, "POST": False, "PUT": False, "DELETE": True, "CALL": False, "SUBSCRIBE": True}`?

Though maybe ACL for this would be better defined through groups (separate db table) rather than a per-key basis (from a sysadmin standpoint). Then you could define permissions for a group and grant individual keys membership to that group.

anodos325: Having a separate table where API ACL is defined may be useful for situation where you decide to add additional authentication methods. It would also allow us to pre-populate with some sensible permissions / roles templates, like "backup operator".

themylogin: @anodos325 this would be nanoseconds

@william-gr do we want separate roles table now?

rick-mesta:
ping: @william-gr / @anodos325

src/middlewared/middlewared/plugins/api_key.py
william-gr: A few things I dont quite understand.

We are only matching exact methods and not wildcards? It may be impractical to require a list of every single if not

Are we reading in the database for every call? That should probably be kept in memory for performance.

I agree with Andrew, we should have a forward thinking, the more we can do to make this future proof and better performance the better. We could use some of that for Kerberos auth in the future.

Also I am wondering how this pave the road for usage with the UI in the future. Someday we will want to restrict the UI using same backend.

rick-mesta:
@william-gr / @anodos325: Fellas, does Vlad need to take any action on this one ? Or have all issues/questions been resolved ?

william-gr:
@rick-mesta he spoke with me this morning. He is going to address a couple things still ; )

rick-mesta:
Thx for the update, @william-gr 👍🏼

src/middlewared/middlewared/plugins/api_key.py
```diff
@@ -58,6 +99,8 @@ async def do_create(self, data):
             data
         )
 
+        await self.load_keys()
```
william-gr, line 102: Why load everything again if you can just add/update/remove the key in question?

rick-mesta:
@anodos325 / @william-gr: there have been additional commits after all y'all's comments... would you kindly revisit 🙏🏼 ?

william-gr (approved):

anodos325 (approved):

PR #7008, opened 2021-06-10
NAS-110987 / 21.08 / Dataset checksum

themylogin:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2935,6 +2951,7 @@ def transform(dataset):
         Inheritable(Int('copies')),
         Inheritable(Str('snapdir', enum=['VISIBLE', 'HIDDEN'])),
         Inheritable(Str('deduplication', enum=['ON', 'VERIFY', 'OFF'])),
+        Inheritable(Str('checksum', enum=ZFS_CHECKSUM_CHOICES)),
         Inheritable(Str('readonly', enum=['ON', 'OFF'])),
```
anodos325, line 2955: Are we sure we want to present all of these choices to users?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -43,6 +43,9 @@
 
 RE_HISTORY_ZPOOL_SCRUB = re.compile(r'^([0-9\.\:\-]{19})\s+zpool scrub', re.MULTILINE)
 RE_HISTORY_ZPOOL_CREATE = re.compile(r'^([0-9\.\:\-]{19})\s+zpool create', re.MULTILINE)
+ZFS_CHECKSUM_CHOICES = [
+    'ON', 'OFF', 'FLETCHER2', 'FLETCHER4', 'SHA256', 'SHA512', 'SKEIN', 'EDONR',
```
themylogin, line 47: @amotin which of these should we leave?

anodos325: Is EDONR on FreeBSD?

anodos325: Are we sure we want to give the user the ability to turn off checksum? (Because the number one rule about dangerous options is that they will be picked up by a youtube influencer somewhere as a performance recommendation) :)

freqlabs: EDONR is not available on FreeBSD.

themylogin: @freqlabs so what should be and what should not be present on FreeBSD and on Linux?

freqlabs: I'm not entirely familiar with what's going on in this code. I'd probably present all the choices. But if an API consumer is able to enumerate the choices at run time, it might make sense to only offer the available choices on the platform, which is all of them on Linux, and all but edonr on FreeBSD. I don't think it's prudent to hide valid options.

themylogin: @anodos325 do you think it'll be fine if we keep `OFF` here but remove it from `pool.dataset.checksum_choices` so it won't be selectable in the UI?

william-gr: We cannot break on things that can be recklessly be done from a shell but we can prevent them from doing stupid things from UI, so remove the bad choices from the _choices API call seems reasonable.

themylogin:
I'll remove `EDONR` in FreeBSD backport.

rick-mesta:
@william-gr / @freqlabs / @anodos325: Gents, has @themylogin addressed your outstanding issues on this PR ?

freqlabs (approved):
Removing EDONR for the backport to FreeBSD should be fine, as long as that plays out correctly for 13.

william-gr:
lgtm too

rick-mesta:
Thx for the quick feedback @freqlabs & @william-gr 👋🏼

PR #7009, opened 2021-06-10
NAS-107662 / 21.08 / Remote syslog server is overloaded alert

themylogin:


src/middlewared/middlewared/plugins/alert.py
```diff
@@ -328,17 +328,17 @@ async def list_sources(self):
...
+        Text('key', null=True),
         Datetime('datetime'),
         Datetime('last_occurrence'),
         Bool('dismissed'),
         Any('mail', null=True),
         Str('text'),
         Str('id'),
         Str('level'),
-        Str('formatted', null=True),
+        Text('formatted', null=True),
```
themylogin, line 341: @sonicaj I suppose we now have middleware failing to return strings with length > 1024.

yocalebo (changes requested):
1. This doesn't take into account the remote syslog connection on HA systems.
2. This request originated in September of 2020 by @anodos325. I'd like to know why this request came in the first place and if this is still even a valid request.

themylogin:
@yocalebo you're right, I forgot about HA syslog.

I'll wait with fixing this until the idea itself gets reviewed.

yocalebo:
Spoke with @anodos325 offline and this can be closed as the request is no longer valid.

PR #7022, opened 2021-06-11
Remove AFP tests

anodos325:


themylogin:
@anodos325 what if we adapt these tests to create AFP shares using new Samba API and continue using OS X to test if AFP protocol still works?

anodos325:
AFP protocol is gone with netatalk removal.

anodos325:
But you're right that we need some tests for functionality specific to the `afp` tag.

anodos325:
Closing PR because this master runs tests against 13-stable.

PR #7028, opened 2021-06-11
NAS-111022 / 21.08 / Add validation related to legacy AFP shares

anodos325:
Since SMB shares will traverse mountpoints verify that new
share is not a child or parent of a share with different AFP
compatibility settings.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -5,6 +5,7 @@ from middlewared.service import accepts, job, private, SharingService, SystemSer
 from middlewared.service_exception import CallError
 import middlewared.sqlalchemy as sa
 from middlewared.utils import osc, Popen, run
+from pathlib import Path
 
 import asyncio
 import codecs
@@ -839,6 +840,7 @@ class SharingSMBService(SharingService):
 
         await self.clean(data, 'sharingsmb_create', verrors)
         await self.validate(data, 'sharingsmb_create', verrors)
+        await self.legacy_afp_check(data, 'sharingsmb_create', verrors)
 
         verrors.check()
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -905,6 +907,7 @@ async def do_update(self, id, data):
         new['vuid'] = await self.generate_vuid(new['timemachine'], new['vuid'])
         await self.clean(new, 'sharingsmb_update', verrors, id=id)
         await self.validate(new, 'sharingsmb_update', verrors, old=old)
+        await self.legacy_afp_check(new, 'sharingsmb_verrors', verrors)
```
william-gr, line 910: _update?

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -1073,6 +1076,28 @@ async def query(self, filters, options):
             return await super().query(filters, options)
         return result
 
+    @private
+    async def legacy_afp_check(self, data, schema, verrors):
```
william-gr, line 1080: You could just return if data['afp'] is False?

anodos325: Not really, because an existing share on the same path may be `true`

william-gr: Sorry, if True

src/middlewared/middlewared/plugins/smb.py
anodos325: The number of typos I make increases exponentially as the number of lines in diff decreases :(

rick-mesta (approved):
ship it 🚢

src/middlewared/middlewared/plugins/smb.py
anodos325: In that case, we may have a share on a parent or child directory that is False. :)

PR #7034, opened 2021-06-14
NAS-111039 / 21.08 / Remove strict check for path existence in AFP validation

anodos325:
SMB share API currently allows creating directories by passing
assing non-existent paths to the sharing.smb.create endpoint.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -1078,7 +1078,7 @@ class SharingSMBService(SharingService):
 
     @private
     async def legacy_afp_check(self, data, schema, verrors):
-        to_check = Path(data['path']).resolve(strict=True)
+        to_check = Path(data['path']).resolve(strict=False)
         legacy_afp = await self.query([
             ("afp", "=", True),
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -1078,7 +1078,7 @@ async def query(self, filters, options):
 
     @private
     async def legacy_afp_check(self, data, schema, verrors):
-        to_check = Path(data['path']).resolve(strict=True)
+        to_check = Path(data['path']).resolve(strict=False)
```
sonicaj, line 1081: Is this to keep backwards compatibility with those cases where the path/dataset might not exist anymore but we still have it in database ?

sonicaj (approved):

src/middlewared/middlewared/plugins/smb.py
anodos325: I wish.... there's ancient SMB plugin behavior that I forgot about where we create the share path if it doesn't exist. I'm probably going to make this some sort of non-default option once we get BETA out the door.

PR #7054, opened 2021-06-17
NAS-110623 / 21.08 / Add explicit route for cluster service network to kube router config

siku2:
This PR updates the generated "kube router config" such that it includes a route for the cluster service network to the cluster. Currently, the config only specifies the pod network and relies on the default route to route requests to the service network correctly.
The primary use case for this is that it allows users to override the default route for individual containers without breaking said container's connection to the service network.

Additionally, the config now uses the "ranges" array instead of the deprecated<sup>[[1]](https://www.cni.dev/plugins/current/ipam/host-local/#network-configuration-reference)</sup> top-level "subnet" key.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/cni.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/cni.py
@@ -65,8 +65,13 @@ class KubernetesCNIService(ConfigService):
                 {
                     'bridge': 'kube-bridge',
                     'ipam': {
-                        'subnet': config['cluster_cidr'],
                         'type': 'host-local',
+                        "ranges": [[{
+                            "subnet": config['cluster_cidr'],
+                        }]],
+                        "routes": [{
+                            "dst": config['service_cidr'],
+                        }],
                     },
                     'isDefaultGateway': True,
                     'name': 'kubernetes',


sonicaj (commented):
@siku2 we use kuberouter for service proxy and load balancing and it uses IPVS. So i am not sure how adding a route which btw still uses the same gateway helps, can you please clarify ?

siku2:
> which btw still uses the same gateway

That's exactly the point though. The goal here is to ensure that the cluster traffic always uses this gateway.
Normally when trying to reach the `service_cidr` network, the default route is used. Normally this is fine because the default route is part of the `cluster_cidr`.

Imagine, however, what happens when the user changes the container's default route using a cni annotation:

```yaml
k8s.v1.cni.cncf.io/networks: |
  [{
    "name": "macvlan", // attach a macvlan interface which is directly connected to the LAN
    "default-route": ["10.0.0.1"] // override the default route so that the container uses the macvlan interface to communicate with the world
  }]
```

Now the container can no longer communicate with the `service_cidr`, because all the traffic flows through the macvlan interface instead of the host-local one.


By adding this simple route for the `service_cidr`, this issue can be avoided altogether.

sonicaj:
@siku2 how about adding this route in the application itself ? The launch docker image wizard exposes that and wouldn't that suffice this use case instead of adding this route for all workloads ?

siku2:
@sonicaj that's possible, but IMO it's much cleaner to have Scale take care of it. If the user isn't aware of this issue it's very hard to debug (speaking from experience). The container can still connect to IP addresses, but can't reach any Cluster services, which means that DNS doesn't work either.
Doing it manually also means hard-coding the `service_cidr` which results in the workload breaking if it is ever changed through the advanced settings.

On the other hand, I don't see any downsides to adding this. It doesn't break existing configurations and adding a route shouldn't add a measurable overhead.
Is there something I'm missing?

sonicaj:
@siku2 how about users who don't want this ? I don't this this is a very common case where you have external interfaces in a pod and you want to use that for a default route and then want to be able to access the service cidr as well directly from it. Please correct me if i am not following you correctly ?

sonicaj:
If you still think this would be useful for the majority of the user(s), i think we should wait on the votes for the suggestion ticket then. Thoughts ?

siku2:
> I don't think this is a very common case where you have external interfaces in a pod and you want to use that for a default route and then want to be able to access the service cidr as well directly from it.

Obviously I'm biased but I think this use-case *should* be more common. It allows you to run applications which communicate using protocols that aren't easily routable (like raw UDP or TCP) without much hassle.
One example that comes to mind is Plex. Using this method you can easily run multiple Plex servers using their own (LAN) IP and MAC addresses, which means both instances are discoverable on the LAN without having to change a single port or even opening the necessary ports on the host itself. It's even possible to use UPnP for opening a port on the router instead of hard-coding them.
All this while still having the ability to make use of the Kubernetes networking, so you could use ingresses like plex1.example.com, plex2.example.com to access them.

> If you still think this would be useful for the majority of the user(s), i think we should wait on the votes for the suggestion ticket then. Thoughts ?

Absolutely, I have no problem with that. I was honestly expecting that this wouldn't be a controversial change so I thought I'd put the implementation out in the open.
Given that this is a very advanced use case most users probably don't know about I suspect the issue won't get much traction though.

Honestly, given that you don't seem to like the idea I'm perfectly fine with closing this too. Once Scale switches from Klipper to  MetalLB (which seems to have been accepted a while ago) my main use case for this will no longer be an issue.

sonicaj:
Thank you @siku2 for understanding. Let's wait for the suggestion ticket and we can move with this change later then. Thanks again for your contribution, it's much appreciated.

sonicaj:
Just to be clear, you can still have the desired affect by adding that route for the application in question via UI. Good luck!

PR #7056, opened 2021-06-17
NAS-111080 / 12.0 /  Fix Enclosure Mapping for early version Mini X 3.0 Systems

darkfiberiru:


darkfiberiru:
@themylogin I believe you wrote this code I patched :D if you can review

PR #7061, opened 2021-06-18
NAS-111096 / 21.08 / Synchronously sync catalog on creation

sonicaj:
This commit adds changes to sync a catalog on creation synchronously as if the catalog is huge and not cached, it will mean that user would still expct it to be available instantaneously which won't be true. As catalog creation is a job, we can safely sync the catalog after inserting it into database which will ensure that the system is ready to serve the catalog.

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/update.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/update.py
@@ -1,4 +1,3 @@
-import asyncio
 import errno
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/catalogs_linux/update.py
```diff
@@ -173,10 +172,12 @@ async def do_create(self, job, data):
         job.set_progress(60, 'Completed Validation')
 
         await self.middleware.call('datastore.insert', self._config.datastore, data)
+        job.set_progress(70, f'Successfully added {data["label"]!r} catalog')
 
-        asyncio.ensure_future(self.middleware.call('catalog.sync', data['label']))
+        job.set_progress(80, f'Syncing {data["label"]} catalog')
+        await self.middleware.call('catalog.sync', data['label'])
```
yocalebo, line 178: This method seems to go to the internet to update git repository information. We create a system alert `CatalogSyncFailed` if any exception occurs but I'm curious if we're going to run into a situation where we have a `CallTimeout` since it gets raised after 60 seconds by default. A slow connection or some weird DNS issue could easily lapse 60 seconds. Thoughts on that situation or do you just middlewared to raise it?

yocalebo (changes requested):

src/middlewared/middlewared/plugins/catalogs_linux/update.py
sonicaj: So this is fixed in https://github.com/truenas/middleware/pull/7063 where this has been converted to a job, but this change needs to go in first before https://github.com/truenas/middleware/pull/7063 can be merged

sonicaj: Also internally we don't have a `CallTimeout`, so if that method takes potentially more then a minute, it wouldn't error out.

yocalebo (approved):

PR #7062, opened 2021-06-18
NAS-108598 / 21.08 / Make idmap/smb/directory services cluster-aware

anodos325:
General change (clustered / non-clustered):
* The global portion of the smb4.conf will be moved to Samba's registry. This means that /etc/smb4.conf will now be a stub containing only two lines in non-clustered case:
```
[global]
    include = registry
```
There will be some fallout from this change. For instance, user comments will no longer appear in the smb4.conf.  There are several middleware plugins that directly impact smb.conf generation (activedirectory, idmap, ldap, smb).

In the non-clustered case, the update() methods will get this smb4.conf portion, convert the new data to an smb4.conf portion, get a diff, and apply it to the registry. idmap.delete() will also remove all smb4.conf parameters associated with the domain being deleted. Otherwise, normal datastore operations should not be impacted.

When SMBHAMODE == CLUSTERED behavior for the above-described services will be changed as follows:
* Active Directory, LDAP, SMB, and idmap, and kerberos-related plugins are backed by `tdb` rather than `datastore`
* SMB shares are layered on top of the samba registry (local datastore skipped entirely).

A clustered cache  (clustercache) plugin was added to middleware. API is largely same as non-clustered cache plugin. This can be used for shared cluster-wide state. In case of directory service start / stop, payload is inserted into clustercache and 
cluster_reload method is triggered on all nodes (to force kerberos / service reload).

This basically ensures that SMB, Idmap, AD, and LDAP config are consistent cluster-wide. Further work will be needed local users and groups.


anodos325:
Made PR to give some visibility. Not quite ready for review yet. Refactoring for smbconf output changes (JSON).

yocalebo:
@anodos325 I know this is currently WIP, but I've pushed flake8 fixes for convenience.

src/middlewared/middlewared/plugins/gluster_linux/local_events.py
```diff
@@ -62,7 +63,9 @@ async def send(self, data):
         token = jwt.encode({'dummy': 'data'}, secret, algorithm='HS256')
         headers = {'JWTOKEN': token.decode('utf-8'), 'content-type': 'application/json'}
         async with aiohttp.ClientSession() as sess:
-            await sess.post(LOCAL_WEBHOOK_URL, headers=headers, json=data, timeout=5)
+            resp = await sess.post(LOCAL_WEBHOOK_URL, headers=headers, json=data, timeout=30)
+
+        return str(resp)
```
anodos325, line 68: Forgot I had added this here. Will remove.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_job.py
```diff
@@ -0,0 +1,136 @@
...
+
+    @private
+    @periodic(3600)
+    @job(lock="queue_lock")
+    async def process_queue(self, job):
+        gl_enabled = (await self.middleware.call('service.query', [('service', '=', 'glusterd')], {'get': True}))['enable']
+        if not gl_enabled:
+            return
+
+        node = (await self.middleware.call('ctdb.general.status', {'all_nodes': False}))[0]
```
yocalebo, line 61: This will raise a `CallError` if the ctdb service isn't running so not sure if you want that or if you want to add another check up above.

anodos325: I think raising a CallError in this case is fine.

yocalebo (approved):
I've tested the clustering portion of active-active SMB shares as much as I know how to 😄 As far as I'm concerned, works for me. SHIP IT

PR #7063, opened 2021-06-20
NAS-111102 / 21.08 / Improve handling of large catalogs

sonicaj:

1) Make `catalog.items` a job as reading large catalog(s) can potentially take quite a few minutes.
2) Properly report progress when reading large catalog(s) with `catalog.items` so that consumer(s) can have a clear estimate.
3) Only retrieve cached content of catalog item(s) from `catalog.query`. Motivation for this change is that querying catalog(s) should be fast whereas reading a single large catalog can potentially take quite some time. However, `OFFICIAL` catalog is treated specially where we try to use cache but if the content's not been cached yet, we will read it from disk. This is with the assumption that `OFFICIAL` catalog would not go into thousands of item version(s) which is what which takes lots of time to read.
4) Do not retrieve catalog item versions with `catalog.query` as catalog(s) with thousand(s) of catalog items would increase the payload which can increase the time to load/dump the catalog version(s) data.
5) Make `catalog.sync` a job as it does 2 functions, one is updating the git repository and the second is calling `catalog.items` and both can take potentially quite some time to complete.

With these changes, UI can load catalog(s) faster and for catalog(s) which are still being synced, UI would report progress for them that they are being synced and display them when they are.

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/items.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/items.py
@@ -1,5 +1,4 @@
-import contextlib
-import functools
+import copy
 import itertools
 import json
 import markdown
@@ -21,6 +20,15 @@ class CatalogService(Service):
     class Config:
         cli_namespace = 'app.catalog'
 
+    @private
+    def cached(self, label, retrieve_versions):
+        return self.middleware.call_sync('cache.has_key', self.cache_key(label, retrieve_versions))
+
+    @private
+    def cache_key(self, label, retrieve_versions):
+        suffix = '' if retrieve_versions else '_without_versions'
+        return f'catalog_{label}_train_details{suffix}'
+
     @accepts(
         Str('label'),
         Dict(
@@ -28,7 +36,7 @@ class CatalogService(Service):
             Bool('cache', default=True),
             Bool('cache_only', default=False),
             Bool('retrieve_all_trains', default=True),
-            Bool('retrieve_versions', default=True),
+            Bool('retrieve_versions', default=False),
             List('trains', items=[Str('train_name')]),
         )
     )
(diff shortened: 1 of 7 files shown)


src/middlewared/middlewared/plugins/catalogs_linux/update.py
```diff
@@ -176,7 +176,9 @@ async def do_create(self, job, data):
         job.set_progress(70, f'Successfully added {data["label"]!r} catalog')
 
         job.set_progress(80, f'Syncing {data["label"]} catalog')
-        await self.middleware.call('catalog.sync', data['label'])
+        sync_job = await self.middleware.call('catalog.sync', data['label'])
```
yocalebo, line 224: Before checking for `error`, you have to `wait()` right?

src/middlewared/middlewared/plugins/catalogs_linux/items.py
```diff
@@ -130,52 +130,68 @@ def items(self, label, options):
...
+            train = items[item]
             trains[train] = {}
-            category_path = os.path.join(location, train)
-            for item in filter(lambda p: os.path.isdir(os.path.join(category_path, p)), os.listdir(category_path)):
-                item_location = os.path.join(category_path, item)
-                if not os.path.isdir(item_location):
-                    continue
+            item_location = os.path.join(location, train, item)
+            job.set_progress(
+                ((index / total_items) * (90 - 10)) + 10,
```
yocalebo, line 205: 1. Why `(90 - 10)`? Just make it `* 80`
2. You probably want to cast this to an `int()` or `round()` it to 2 decimal places so you don't have a `33.3333334%` or something like that.

src/middlewared/middlewared/plugins/catalogs_linux/items.py
```diff
@@ -93,81 +121,106 @@ def items(self, label, options):
...
-        if all_trains and options['retrieve_versions']:
+        if all_trains:
             # We will only update cache if we are retrieving data of all trains for a catalog
             # which happens when we sync catalog(s) periodically or manually
-            self.middleware.call_sync('cache.put', f'catalog_{label}_train_details', trains, 86400)
+            # We cache for 90000 seconds giving system an extra 1 hour to refresh it's cache which
+            # happens after 24h - which means that for a small amount of time it's possible that user
+            # come with a case where system is trying to access cached data but it has expired and it's
+            # reading again from disk hence the extra 1 hour.
+            if options['retrieve_versions']:
```
yocalebo, line 143: How much memory does this eat when we cache everything?

src/middlewared/middlewared/plugins/catalogs_linux/sync_catalogs.py
```diff
@@ -20,31 +20,44 @@ async def sync_all(self, job):
         catalog_len = len(catalogs)
         for index, catalog in enumerate(catalogs):
             job.set_progress((index / catalog_len) * 100, f'Syncing {catalog["id"]} catalog')
-            try:
-                await self.middleware.call('catalog.sync', catalog['id'])
-            except Exception as e:
-                self.logger.error('Failed to sync %r catalog: %s', catalog['id'], e)
+            sync_job = await self.middleware.call('catalog.sync', catalog['id'])
```
yocalebo, line 23: Before this change you were logging an error, if one was raised. Now you're not. Is that intended?

yocalebo (changes requested):

src/middlewared/middlewared/plugins/catalogs_linux/sync_catalogs.py
sonicaj: Yes that's intended

src/middlewared/middlewared/plugins/catalogs_linux/items.py
sonicaj: @yocalebo that would depend on the size of catalog, we don't have a catalog big enough which i think would affect memory right now or in the coming year i hope. However i think we shouldn't be retrieving version(s) from this endpoint at all and i have updated the docstring to point out that it has been deprecated.
I tested with a catalog with 250 items and it took approximately 6 MB of memory to cache and we are nowhere close that figure now.

yocalebo (approved):

PR #5474, opened 2020-08-13
NAS-107214 / 12.1 / cache dmidecode results to reduce CPU time

yocalebo:

PR #7065, opened 2021-06-22
NAS-111124 / 12.0 / cache static info and optimize system.info on CORE

yocalebo:
I profiled the middlewared process a few months ago and found lots of areas of improvement. The gist of it is that we're spending an exorbitant amount of CPU time simply retrieving static information that never changes after boot.

The details of the performance penalty for doing this can be found [here](https://github.com/truenas/middleware/pull/5474), [here](https://github.com/truenas/middleware/pull/5612), and [here](https://github.com/truenas/middleware/pull/5619). To add insult to injury, the NAS-109709 ticket has shown that 75% of the core dumps show `system.info` was on CPU at the time of the crash. The fact that it's on CPU doesn't mean it's related to the core dumps but it does show the fact that we're wasting lots of CPU time on static data that never changes.

These changes have been in SCALE for many months and they were actually supposed to be backported to TN Core "nightlies" at the time, but that was never done because of the `truenas/13.0-stable` branch creation so it was missed.

This just backports the above mentioned PR to `truenas/12.0-stable`.

### DIFF
--- a/src/middlewared/middlewared/plugins/system.py
+++ b/src/middlewared/middlewared/plugins/system.py
@@ -1,5 +1,5 @@
 import asyncio
-from datetime import datetime, date, timezone
+from datetime import datetime, date, timezone, timedelta
 from middlewared.event import EventSource
 from middlewared.i18n import set_language
 from middlewared.logger import CrashReporting
@@ -356,7 +356,6 @@ class SystemAdvancedService(ConfigService):
 
 class SystemService(Service):
 
-    HOST_ID = None
     DMIDECODE_CACHE = {
         'ecc-memory': None,
         'baseboard-manufacturer': None,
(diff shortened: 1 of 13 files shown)


yocalebo:
@rick-mesta and @sonicaj  these changes are cherry-picked from the above mentioned PRs. I've only made slight alterations (flake8 fixes) mostly.

rick-mesta (approved):
LGTM; certainly,  best to wait for @sonicaj's review 👍🏼

src/middlewared/middlewared/plugins/system.py
```diff
@@ -356,12 +356,133 @@ def autotune(self, conf='loader'):
...
+        'cpu_model': None,
+        'core_count': None,
+    }
+
+    MEM_INFO = {
+        'physmem_size': None,
+    }
+
+    BIRTHDAY_DATE = {
+        'date': None,
```
sonicaj, line 379: Can we have something else here apart from `date` key ?

src/middlewared/middlewared/plugins/system.py
```diff
@@ -1442,39 +1527,36 @@ def set_crash_reporting(self):
...
     timeout = 3600 * 24
 
     middleware.register_hook('interface.post_sync', _update_birthday_data)
 
+    middleware.logger.debug('Waiting for clock sync to update system birthday')
     while birthday is None:
         birthday = await middleware.call('system.sync_clock')
 
-        # Wait until be able to sync the clock
+        # sleep for 1 day and try again
```
sonicaj, line 1559: We should not wait for 1 day, i think a delay of 15 min or 30 min should be enough to let ntp sync clock

src/middlewared/middlewared/plugins/system.py
```diff
@@ -356,12 +356,133 @@ def autotune(self, conf='loader'):
...
+        'cpu_model': None,
+        'core_count': None,
+    }
+
+    MEM_INFO = {
+        'physmem_size': None,
+    }
+
+    BIRTHDAY_DATE = {
+        'birth': None,
```
sonicaj, line 379: I was basically thinking that do we require this to be a dict ? How about storing it directly ?

yocalebo: Oh yeah, makes sense. I'll fix

sonicaj (approved):
Thanks!

PR #7068, opened 2021-06-22
NAS-111122 / 21.08 / fix BHYVE SCALE HA VM detection

yocalebo:
Internally, we're testing SCALE HA functionality by configuring BHYVE hypervisors a specific way. However, before this change, the code actually caused any SCALE VM that's installed on a BHYVE hypervisor to return `SCALE_ENTERPRISE` which is incorrect. (i.e. a TN Core user installs a SCALE VM)

Change it so that we only return `BHYVE` based on how we configure our internal BHYVE hypervisors.


sonicaj (approved):
Thanks!

PR #7069, opened 2021-06-22
NAS-111132 / 21.08 / Fix r10 enclosures

darkfiberiru:
For iX branded hardware, it's expected that the drive mapping matches a standard pattern (left to right, then top to bottom). The R10 has the inverse of this, however, we have never had to map anything other than TrueNAS-Mini's. This fixes and improves a few things. This was tested on R10 and Mini-X+ to confirm mini mapping still works.

1. update r-series regex to catch R10 model
2. add a `MINI_REGEX` so that the `self.model` in `class Enclosure` doesn't interfere with R10 product mapping
3. future-proof the code by adding back the expansion shelf information to the product mapping

### DIFF
--- a/src/middlewared/middlewared/plugins/enclosure_/map.py
+++ b/src/middlewared/middlewared/plugins/enclosure_/map.py
@@ -139,16 +139,11 @@ class EnclosureService(Service):
 
             elements.append(element)
 
-        info = await self.middleware.call("system.info")
-        if '-MINI-' in info["system_product"]:
-            model = info["system_product"]
-        else:
-            model = info["system_product"].replace("TRUENAS-", "")
-        return [
+        mapped = [
             {
                 "id": "mapped_enclosure_0",
                 "name": "Drive Bays",
-                "model": model,
+                "model": enclosures[0].model,
                 "controller": True,
                 "elements": [
                     {
@@ -161,3 +156,8 @@ class EnclosureService(Service):
                 ],
             }
         ]
+        #Add shelves back in
+        for enclosure in enclosures:
+            if enclosure.controller == false:
+                mapped.append(enclosure)
+        return mapped
--- a/src/middlewared/middlewared/plugins/enclosure_/map.py
+++ b/src/middlewared/middlewared/plugins/enclosure_/map.py
@@ -143,7 +143,7 @@ class EnclosureService(Service):
             {
                 "id": "mapped_enclosure_0",
                 "name": "Drive Bays",
-                "model": enclosures[0].model,
+                "model": enclosures[0]['model'],
                 "controller": True,
                 "elements": [
                     {
@@ -161,7 +161,7 @@ class EnclosureService(Service):
         # ability to support expansion shelves, then we need to add them
         # back in here so drive identification works
         for enclosure in enclosures:
-            if enclosure.controller == false:
+            if enclosure['controller'] == False:
                 mapped.append(enclosure)
 
         return mapped
--- a/src/middlewared/middlewared/plugins/enclosure_/map.py
+++ b/src/middlewared/middlewared/plugins/enclosure_/map.py
@@ -55,6 +55,26 @@ MAPPINGS = [
             MappingSlot(1, 3, False),
         ]),
     ]),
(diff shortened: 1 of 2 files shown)


darkfiberiru:
@themylogin Please review

src/middlewared/middlewared/plugins/enclosure_/map.py
```diff
@@ -120,11 +140,15 @@ async def _map_enclosures(self, enclosures, slots):
             elements.append(element)
 
         info = await self.middleware.call("system.info")
+        if '-MINI-' in ["system_product"]:
```
yocalebo, line 143: This is wrong and will always evaluate to `False`

yocalebo (changes requested):

src/middlewared/middlewared/plugins/enclosure_/map.py
yocalebo: Also, is there a jira ticket for this?

darkfiberiru: I was gonna ask some to add jira label as I'm not in truenas org

yocalebo: Ahh okay sounds good, I'm going to close the PR against `truenas/12.0-stable` since I can just add the `backport-12.0` label as well which will handle it for you.

src/middlewared/middlewared/plugins/enclosure_/map.py
```diff
@@ -120,11 +140,15 @@ async def _map_enclosures(self, enclosures, slots):
...
+            model = info["system_product"]
+        else:
```
yocalebo, line 145: Looking at this a bit closer, I'm curious why you're stripping `TRUENAS-` from the model name for all platforms except the mini?

yocalebo (changes requested):

src/middlewared/middlewared/plugins/enclosure_/map.py
darkfiberiru: It's to deal with webui implementation.

darkfiberiru: fixed

darkfiberiru: Also previously some assumptions we would only need to map mini's. That's also why I removed comment on line 231 as I believe it's not accurate.

darkfiberiru: Please resolve this if your happy

yocalebo: Please make the following changes
```
-        info = await self.middleware.call("system.info")
+        mode = enclosures[0]["model"]
+        if not model:
+            model = await self.middleware.call("system.dmidecode_info")["system-product-name"]
+
         return [
             {
                 "id": "mapped_enclosure_0",
                 "name": "Drive Bays",
-                "model": info["system_product"],
+                "model": model,
                 "controller": True,
                 "elements": [
                     {

Comment (yocalebo) [reply src/middlewared/middlewared/plugins/enclosure_/map.py]: Discussed with @darkfiberiru privately in slack. Changes were implemented here. Resolving this conversation.

**Review by yocalebo — APPROVED** [review]
I conditionally approve this so long as @themylogin gives the okay.

**Comment by darkfiberiru** [comment]
Sounds good. I have tested hot patched version on r10 and mini-x

File: src/middlewared/middlewared/plugins/enclosure_/map.py [hunk 3]
```diff
@@ -119,12 +139,11 @@ async def _map_enclosures(self, enclosures, slots):
...
             elements.append(element)
 
-        info = await self.middleware.call("system.info")
-        return [
+        mapped = [
             {
                 "id": "mapped_enclosure_0",
                 "name": "Drive Bays",
-                "model": info["system_product"],
+                "model": enclosures[0]["model"],
```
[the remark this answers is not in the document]
yocalebo: Yes, we will always have at least 1 item in `enclosures` if we end up getting to this code path.

sonicaj (approved):

PR #7070, opened 2021-06-22
Fix r10 enclosure

darkfiberiru:
Including mapping drives bays to match standard pattern(left to right then top to bottom) instead of inverse. Tested on R10 and Mini-X+ to confirm mini mapping still works

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
@@ -55,6 +55,26 @@ MAPPINGS = [
             MappingSlot(1, 3, False),
         ]),
     ]),
+    ProductMapping(re.compile(r"TRUENAS-R10$"), [
+        VersionMapping(re.compile(".*"), [
+            MappingSlot(0, 0, False),
+            MappingSlot(0, 4, False),
+            MappingSlot(0, 8, False),
+            MappingSlot(0, 12, False),
+            MappingSlot(0, 1, False),
+            MappingSlot(0, 5, False),
+            MappingSlot(0, 9, False),
+            MappingSlot(0, 13, False),
+            MappingSlot(0, 2, False),
+            MappingSlot(0, 6, False),
+            MappingSlot(0, 10, False),
+            MappingSlot(0, 14, False),
+            MappingSlot(0, 3, False),
+            MappingSlot(0, 7, False),
+            MappingSlot(0, 11, False),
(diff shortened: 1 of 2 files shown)


darkfiberiru:
@themylogin Please review

src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
```diff
@@ -120,11 +140,15 @@ async def _map_enclosures(self, enclosures, slots):
             elements.append(element)
 
         info = await self.middleware.call("system.info")
+        if '-MINI-' in ["system_product"]:
```
yocalebo, line 143: This is wrong and will always evaluate to `False`

yocalebo (changes requested):

src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
darkfiberiru: Thanks  you mean that I messed up what should of been info["system_product"] right?

src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
```diff
@@ -120,11 +140,15 @@ async def _map_enclosures(self, enclosures, slots):
             elements.append(element)
 
         info = await self.middleware.call("system.info")
```
yocalebo, line 142: `system.info` actually runs a bunch of unnecessary/unrelated API calls that are irrelevant for this. `system.dmidecode_info` is much better suited for this particular scenario.

yocalebo (changes requested):

src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
darkfiberiru: That's probably a good fix but don't want to mix unrelated fixes

src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
yocalebo: Yeah, that's what it looks like from here.

yocalebo:
Closing this PR in favor of the `backport-12.0` label on PR: https://github.com/truenas/middleware/pull/7069

PR #7077, opened 2021-06-25
NAS-111173 / 21.08 / add system.hostname endpoint and change callers to use it

yocalebo:


yocalebo:
Note: rebased on https://github.com/truenas/middleware/pull/7073 so merge that one first before this one.

yocalebo:
I tried to be clever with rebasing and don't really know what I'm doing. I don't like what this is doing so I'll close and re-open fresh.

src/middlewared/middlewared/plugins/system.py
```diff
@@ -524,27 +524,39 @@ async def dmidecode_info(self):
             self.DMIDECODE_CACHE['ecc-memory'] = ecc
 
         if self.DMIDECODE_CACHE['baseboard-manufacturer'] is None:
-            bm = ((await run(["dmidecode", "-s", "baseboard-manufacturer"], check=False)).stdout.decode(errors='ignore')).strip()
+            bm = (
+                (await run(["dmidecode", "-s", "baseboard-manufacturer"], check=False)).stdout.decode(errors='ignore')
```
sonicaj, line 528: Can we please be consistent with quotes ? Like use single or double but not both ?

src/middlewared/middlewared/plugins/system.py
```diff
@@ -524,27 +524,39 @@ async def dmidecode_info(self):
...
+            ).strip()
             self.DMIDECODE_CACHE['baseboard-manufacturer'] = bm
 
         if self.DMIDECODE_CACHE['baseboard-product-name'] is None:
-            bpn = ((await run(["dmidecode", "-s", "baseboard-product-name"], check=False)).stdout.decode(errors='ignore')).strip()
+            bpn = (
```
sonicaj, line 533: Is it useful to set all of these variables separately and have a separate `if` condition for each variable ? In other words, won't this only be true once ?

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1091,10 +1093,18 @@ def _get_filters(key):
             return [[key, '!=', update['id']]] if update else []
 
         validation_attrs = {
-            'aliases': ['Active node IP address', ' cannot be changed.', ' is required when configuring HA'],
-            'failover_aliases': ['Standby node IP address', ' cannot be changed.', ' is required when configuring HA'],
-            'failover_virtual_aliases': ['Virtual IP address', ' cannot be changed.', ' is required when configuring HA'],
-            'failover_group': ['Failover group number', ' cannot be changed.' ' is required when configuring HA'],
+            'aliases': [
```
sonicaj, line 1096: Can you please clarify why we have changed this ?

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1325,10 +1335,9 @@ def _get_filters(key):
                 # newly created laggs.
                 if not update:
                     if data.get('failover_critical') and data.get('lag_protocol') == 'FAILOVER':
-                        verrors.add(
-                            f'{schema_name}.failover_critical',
-                            'A lagg interface using the "Failover" protocol is not allowed to be marked critical for failover.'
-                        )
+                        msg = 'A lagg interface using the "Failover" protocol '
```
sonicaj, line 1338: Same here

src/middlewared/middlewared/plugins/network.py
```diff
@@ -682,7 +682,9 @@ def iface_extend(self, iface_state, configs, ha_hardware, fake=False):
                     'netmask': int(config['int_v6netmaskbit']),
                 })
 
-        for alias in self.middleware.call_sync('datastore.query', 'network.alias', [('alias_interface', '=', config['id'])]):
+        for alias in self.middleware.call_sync(
```
sonicaj, line 685: Same as below

src/middlewared/middlewared/plugins/network.py
```diff
@@ -96,14 +96,16 @@ class Config:
             strict=True
         ),
         Str('hostname_local', required=True, validators=[Hostname()]),
+        Str('hostname_b', validators=[Hostname()], null=True),
```
sonicaj, line 99: Can this be empty string ?

src/middlewared/middlewared/plugins/mail.py
```diff
@@ -406,9 +405,13 @@ def read_json():
             msg['Cc'] = ', '.join(message.get('cc'))
         msg['Date'] = formatdate()
 
-        local_hostname = socket.gethostname()
-
-        msg['Message-ID'] = "<%s-%s.%s@%s>" % (sw_name.lower(), datetime.utcnow().strftime("%Y%m%d.%H%M%S.%f"), base64.urlsafe_b64encode(os.urandom(3)), local_hostname)
+        local_hostname = self.middleware.call_sync('system.hostname')
+        msg['Message-ID'] = "<%s-%s.%s@%s>" % (
```
sonicaj, line 409: Can we use f-string please ?

src/middlewared/middlewared/plugins/system.py
yocalebo: That's how it was before but I agree, I will go back and fix it.

src/middlewared/middlewared/plugins/system.py
yocalebo: Correct, this should only be set once for each item. This is static data that never changes after boot or after middlewared starts up for the first time.

src/middlewared/middlewared/plugins/network.py
yocalebo: flake8 fix one of the lines were too long but I changed all of them so they were all consistent.

src/middlewared/middlewared/plugins/network.py
yocalebo: flake8 fix for line being too long

src/middlewared/middlewared/plugins/system.py
sonicaj: Yes but that would not happen here ;) This if would never be true twice

src/middlewared/middlewared/plugins/network.py
yocalebo: This was the weird rebase issue I had, but this change has already been merged into master

src/middlewared/middlewared/plugins/network.py
sonicaj: We have merged this, but i think we should have flake8 changes in a different PR to limit the scope of a PR.

src/middlewared/middlewared/plugins/network.py
yocalebo: flake8 fix for line being too long

PR #7079, opened 2021-06-25
NAS-111173 / 21.08 / add system.hostname endpoint and change callers to use it

yocalebo:
We're `import socket` in multiple modules and call `socket.gethostname()`. This is minor code cleanup to make a `system.hostname` endpoint and change all the callers to use it instead.


rick-mesta (approved):
ship it 🚢 !

PR #7081, opened 2021-06-26
NAS-111146 / 21.08 / add interfaces marked critical for failover in network debug

yocalebo:
Support ream requests that the interfaces that are marked critical for failover be added to the debug archive. While I'm here, remove references to freebsd since we have a separate repo for that code.

### DIFF
--- a/src/freenas/usr/local/libexec/freenas-debug/network/network.sh
+++ b/src/freenas/usr/local/libexec/freenas-debug/network/network.sh
@@ -45,79 +45,38 @@ network_func()
 	hostname
 	section_footer
 
-	#
-	#	Dump hosts configuration
-	#
 	section_header "Hosts File (/etc/hosts)"
 	sc "/etc/hosts"
 	section_footer
 
-	#
-	#	Dump resolver information
-	#
 	section_header "/etc/resolv.conf"
 	sc /etc/resolv.conf 2>/dev/null
 	section_footer
 
 	section_header "Interfaces"
-	if is_linux; then
-		interfaces=$(ls /sys/class/net)
-	else
-		interfaces=$(ifconfig -l)
-	fi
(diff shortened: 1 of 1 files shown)


src/freenas/usr/local/libexec/freenas-debug/network/network.sh
```diff
@@ -126,59 +85,71 @@ network_func()
...
-		ip route show default | awk '/default/ {print $3}'
-	else
-		route -n show default|grep gateway|awk '{ print $2 }'
+	# grab the interfaces marked critical for failover
+	# if this is an HA system
+	is_ha=$(midclt call failover.licensed)
+	if [ "$is_ha" = "True" ]; then
+		section_header "Interfaces marked critical for failover"
+		ints=$(${FREENAS_SQLITE_CMD} ${FREENAS_CONFIG} -line "
+		SELECT
```
sonicaj, line 94: Should we use `midclt` to retrieve this information please like we are doing above perhaps ? I think that would mean consistent results as well like via midclt and what we expect in debug causing less confusion between field names etc as well ( not saying this is confusing but just saying that would be consistent with the API as well ) ?

yocalebo: Using midclt makes it impossible to read since 1 interface is 111 lines long after passing it to jq.

sonicaj: We can use `select` to only retrieve values we are interested in ?

yocalebo: I'd rather not. Parsing becomes a headache and the debug scripts don't really have a standard that is followed. 😄

sonicaj (approved):

PR #7090, opened 2021-06-28
NAS-111215 / 21.08 / Bug fix for validating acltype when parent ds does not exist

sonicaj:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3429,6 +3429,9 @@ async def __common_validation(self, verrors, schema, data, mode, parent=None, cu
         else:
             parent = parent[0]
 
+        # We raise validation errors here as parent could be used down to validate other aspects of the dataset
+        verrors.check()
+
```
rick-mesta, line 3434: Shouldn't this be inside the `if not parent` conditional ? Otherwise, it'll be executed whether the parent dataset is there or not

sonicaj: @rick-mesta it will raise any validation errors which have been added to `verrors` at this point. So if none have been added, no validation error will be raised.

rick-mesta: Ah... copy that, and thx for the clarification/education 😁

rick-mesta (approved):

PR #7095, opened 2021-06-29
NAS-111212 / 21.08 / Add POSIX1E ACL API regression tests

anodos325:
This covers basic endpoint for getting / setting. Correct OS /FS
behavior is not validated at this point (permissions enforcment).


ericbsd (approved):
LGTM

PR #7096, opened 2021-06-29
NAS-111240 / 21.08 / Add `who` string to ACL output

anodos325:
Add optional (non-default) `resolve_ids` argument to `filesystem.getacl`
so that API users have an efficient way to get names from ACLs. This is
non-default because it exposes us to broken DNS or other environmental
issues. Should only be specified when _really_ needed.

### DIFF
--- a/src/middlewared/middlewared/plugins/filesystem_/acl_base.py
+++ b/src/middlewared/middlewared/plugins/filesystem_/acl_base.py
@@ -299,33 +299,33 @@ class ACLBase(ServicePartBase):
         Return ACL of a given path. This may return a POSIX1e ACL or a NFSv4 ACL. The acl type is indicated
         by the `acltype` key.
 
-        `simplified` NFSv4 - returns a shortened form of the ACL permset, POSIX1E - this is a no-op.
+        `simplified` - effect of this depends on ACL type on underlying filesystem. In the case of
+        NFSv4 ACLs simplified permissions and flags are returned for ACL entries where applicable.
+        NFSv4 errata below. In the case of POSIX1E ACls, this setting has no impact on returned ACL.
 
         `resolve_ids` - adds additional `who` key to each ACL entry, that converts the numeric id to
         a user name or group name. In the case of owner@ and group@ (NFSv4) or USER_OBJ and GROUP_OBJ
-        (POSIX1E), st_uid or st_gid will be converted from stat() return for file. In all other cases.
+        (POSIX1E), st_uid or st_gid will be converted from stat() return for file. In the case of
         MASK (POSIX1E), OTHER (POSIX1E), everyone@ (NFSv4), key `who` will be included, but set to null.
         In case of failure to resolve the id to a name, `who` will be set to null. This option should
         only be used if resolving ids to names is required.
 
         Errata about ACLType NFSv4:
 
-        `simplified` returns a shortened form of the ACL permset and flags.
+        `simplified` returns a shortened form of the ACL permset and flags where applicable. If permissions
+        have been simplified, then the `perms` object will contain only a single `BASIC` key with a string
+        describing the underlying permissions set.
 
         `TRAVERSE` sufficient rights to traverse a directory, but not read contents.
 
         `READ` sufficient rights to traverse a directory, and read file contents.
 
-        `MODIFIY` sufficient rights to traverse, read, write, and modify a file. Equivalent to modify_set.
+        `MODIFIY` sufficient rights to traverse, read, write, and modify a file.
 
         `FULL_CONTROL` all permissions.
 
         If the permisssions do not fit within one of the pre-defined simplified permissions types, then
         the full ACL entry will be returned.
-
-        An inheriting empty everyone@ ACE is appended to non-trivial ACLs in order to enforce Windows
-        expectations regarding permissions inheritance. This entry is removed from NT ACL returned
-        to SMB clients when 'ixnas' samba VFS module is enabled. We also remove it here to avoid confusion.
         """
 
     @accepts(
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/filesystem_/acl_base.py
```diff
@@ -292,11 +292,21 @@ def setacl(self, job, data):
...
+        by the `acltype` key.
+
+        `simplified` NFSv4 - returns a shortened form of the ACL permset, POSIX1E - this is a no-op.
+
+        `resolve_ids` - adds additional `who` key to each ACL entry, that converts the numeric id to
+        a user name or group name. In the case of owner@ and group@ (NFSv4) or USER_OBJ and GROUP_OBJ
+        (POSIX1E), st_uid or st_gid will be converted from stat() return for file. In all other cases.
+        MASK (POSIX1E), OTHER (POSIX1E), everyone@ (NFSv4), key `who` will be included, but set to null.
+        In case of failure to resolve the id to a name, `who` will be set to null. This option should
+        only be used if resolving ids to names is required.
```
rick-mesta, line 309: Query: would it be useful to convert an unknown id to its string representation ? say uid 12345 is unable to be translated into its `username`, would it be useful to return the string "12345" as the `who`  instead of simply returning null ?

rick-mesta (approved):

src/middlewared/middlewared/plugins/filesystem_/acl_base.py
```diff
@@ -292,11 +292,21 @@ def setacl(self, job, data):
     @accepts(
         Str('path'),
         Bool('simplified', default=True),
+        Bool('resolve_ids', default=False),
     )
-    def getacl(self, path, simplified):
+    def getacl(self, path, simplified, resolve_ids):
```
yocalebo, line 297: Do we need to make this method a job since, historically, resolving uid/gids to their respective names can take quite awhile?

yocalebo (changes requested):

src/middlewared/middlewared/plugins/filesystem_/acl_linux.py
```diff
@@ -254,22 +281,33 @@ def getacl_posix1e(self, path, simplified):
...
 
         ret['trivial'] = (len(ret['acl']) == 3)
         return ret
 
-    def getacl(self, path, simplified):
+    def getacl(self, path, simplified, resolve_ids):
         if not os.path.exists(path):
             raise CallError('Path not found.', errno.ENOENT)
         # Add explicit check for ACL type
+        self.logger.debug("resolve_ids: %s", resolve_ids)
```
anodos325, line 306: leftover debug. Derp

src/middlewared/middlewared/plugins/filesystem_/acl_base.py
```diff
@@ -292,11 +292,21 @@ def setacl(self, job, data):
...
         """
         Return ACL of a given path. This may return a POSIX1e ACL or a NFSv4 ACL. The acl type is indicated
-        by the `ACLType` key.
+        by the `acltype` key.
+
+        `simplified` NFSv4 - returns a shortened form of the ACL permset, POSIX1E - this is a no-op.
```
anodos325, line 302: Take two lines to write something sensible and correct.

src/middlewared/middlewared/plugins/filesystem_/acl_base.py
```diff
@@ -292,11 +292,21 @@ def setacl(self, job, data):
...
+
+        `resolve_ids` - adds additional `who` key to each ACL entry, that converts the numeric id to
+        a user name or group name. In the case of owner@ and group@ (NFSv4) or USER_OBJ and GROUP_OBJ
+        (POSIX1E), st_uid or st_gid will be converted from stat() return for file. In all other cases.
+        MASK (POSIX1E), OTHER (POSIX1E), everyone@ (NFSv4), key `who` will be included, but set to null.
```
anodos325, line 307: Change to "In the case of MASK (POSIX1E), OTHER (POSIX1E), everyone@ (NFSv4), key `who` will be included, but set to null."

src/middlewared/middlewared/plugins/filesystem_/acl_base.py
anodos325: I'm on the fence about that one honestly. In JSON output variant of nfs4xdr_getfacl, I do just that (return numeric ID as string). Our webui team though puts the name in red text in case of name resolution failure and I thought this might be a good way to indicate that there is an issue to them. 

I was also thinking of cases where id and name might be identical. uid = 12345 and user gives name "12345".

rick-mesta: Solid point ! I know we had to do something similar in Solaris, but I realize this can open up Pandora's box. Only making a suggestion for visibility purposes (ie. it may be more useful to see a stringified uid than a null), but I feel your pain. 😬

src/middlewared/middlewared/plugins/filesystem_/acl_base.py
anodos325: Changing to a job is a hugely intrusive change IMHO (this is used lots of places), but there are pathological edge cases in some environments. Perhaps a better option is to specify timeout in the `id_to_name()` method and return None if it takes something like more than 5 seconds to complete the request. In theory this should be enough to get winbind to cache response.

yocalebo: It's definitely intrusive 👀 . I agree, a timeout value would be good.

yocalebo (approved):

PR #7099, opened 2021-06-29
NAS-111175 / 21.08 / add xmit_hash_policy and lacpdu_rate for lagg interfaces

yocalebo:
While I'm in this horribly messy code, fix a few issues.

1. we were calling `next(iter(iface.port))` without specifying a default to the `next()` function so it could raise a `StopIteration` in an edge-case which causes a nasty `asyncio` warning
2. `lag_setup` method was written in a way that would force me to `down()` and `up()` the interface for any of the 2 newly added options (if they were specified). This is bad but more importantly, some switches will detect this as "link flaps" and actually disable the associated switch port(s). Change this so when a lagg is setup for the first time and/or updated, only take it down as necessary.
3. `disable_capabilities` was kept for API backwards capability and while it's `False` always on SCALE, if someone were to specify it as `True` via the API manually, then we would fail because that method doesn't even exist on SCALE.
4. This also fixes a race condition with `attach_interface` and a udev event. Since we were `down()`'ing and then `up()'ing` the interface the moment we changed the protocol, `attach_interface` would call `interface.query()` which would run `__getstate__` and `self.protocol.name` would return `'NoneType' object has no attrbute 'name'` because the protocol had not been set by the time the udev event was triggered.
5. various flake8 fixes and general code clean up

### DIFF
--- a/src/middlewared/middlewared/plugins/interface/lag.py
+++ b/src/middlewared/middlewared/plugins/interface/lag.py
@@ -11,16 +11,16 @@ class InterfaceService(Service):
     @private
     def lag_setup(self, lagg, members, disable_capabilities, parent_interfaces, sync_interface_opts):
         name = lagg['lagg_interface']['int_interface']
-        self.logger.info('Setting up {}'.format(name))
+        self.logger.info('Setting up %s', name)
 
         try:
             iface = netif.get_interface(name)
         except KeyError:
             iface = None
         else:
-            first_port = next(iter(iface.ports))
+            first_port = next(iter(iface.ports), None)
             if first_port is None or first_port[0] != members[0]['lagg_physnic']:
-                self.logger.info('Destroying existing %s as its first port has changed', name)
+                self.logger.info('Destroying %s because its first port has changed', name)
                 netif.destroy_interface(name)
                 iface = None
 
@@ -28,13 +28,50 @@ class InterfaceService(Service):
             netif.create_interface(name)
             iface = netif.get_interface(name)
 
+        # FIXME: this will fail on SCALE since we don't even have
+        # a "disable_capabilities" method
+        """
         if disable_capabilities:
             self.middleware.call_sync('interface.disable_capabilities', name)
+        """
 
+        info = {'protocol': None, 'xmit_hash_policy': None, 'lacpdu_rate': None}
         protocol = getattr(netif.AggregationProtocol, lagg['lagg_protocol'].upper())
         if iface.protocol != protocol:
-            self.logger.info('{}: changing protocol to {}'.format(name, protocol))
-            iface.protocol = protocol
+            info['protocol'] = protocol
+
+        xmit_hash = lagg['lagg_xmit_hash_policy']
+        if iface.xmit_hash_policy != xmit_hash:
+            info['xmit_hash_policy'] = xmit_hash
+
+        lacpdu_rate = lagg['lagg_lacpdu_rate']
+        if iface.lacpdu_rate != lacpdu_rate:
+            info['lacpdu_rate'] = lacpdu_rate
+
+        if any(i is not None for i in info.values()):
+            # means one of the lagg options changed or is being
+            # setup for the first time so we have to down the
+            # interface before performing any of the actions
+            iface.down()
+
+            if info['protocol'] is not None:
+                # we _always_ have to start with the protocol
+                # information first since it deletes members
+                # (if any) of the current lagg and then changes
+                # the protocol
+                self.logger.info('Changing protocol on %r to %s', name, info['protocol'].name)
+                iface.protocol = info['protocol']
+
+            if info['xmit_hash_policy'] is not None:
+                self.logger.info('Changing xmit_hash_policy on %r to %s', name, info['xmit_hash_policy'])
+                iface.xmit_hash_policy = info['xmit_hash_policy']
+
+            if info['lacpdu_rate'] is not None:
+                self.logger.info('Changing lacpdu_rate on %r to %s', name, info['lacpdu_rate'])
+                iface.lacpdu_rate = info['lacpdu_rate']
+
+            # be sure and bring the lagg back up after making changes
+            iface.up()
 
         members_database = []
         members_configured = {p[0] for p in iface.ports}
@@ -58,7 +95,8 @@ class InterfaceService(Service):
             try:
                 port_iface = netif.get_interface(port[0])
             except KeyError:
-                self.logger.warn('Could not find {} from {}'.format(port[0], name))
+                self.logger.warning('Could not find %s from %s', port[0], name)
                 continue
+
             parent_interfaces.append(port[0])
             port_iface.up()
--- a/src/middlewared/middlewared/plugins/interface/netif_linux/lagg.py
+++ b/src/middlewared/middlewared/plugins/interface/netif_linux/lagg.py
@@ -1,9 +1,8 @@
-# -*- coding=utf-8 -*-
 import enum
 import logging
+import pathlib
 
 import middlewared.plugins.interface.netif_linux.interface as interface
-
 from .utils import run
 
 logger = logging.getLogger(__name__)
@@ -23,6 +22,7 @@ def create_lagg(name):
 
 
 class LaggMixin:
+
     @property
     def protocol(self):
         value = self._sysfs_read(f"/sys/devices/virtual/net/{self.name}/bonding/mode").split()[0]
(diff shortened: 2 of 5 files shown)


rick-mesta:
Hey @yocalebo... how do you go about testing these deltas to ensure we minimize any regression ?

yocalebo:
@rick-mesta for this particular scenario, I used a local VM and tested the creation, updating, retrieval, and deletion of `bond` interfaces.

rick-mesta:
Thx for the info, @yocalebo; this LGTM

rick-mesta (approved):
How are these changes getting tested to ensure no regression

yocalebo:
@rick-mesta I'm going to flag this as `Pending QA` for now. See if maybe I can update and/or write some unit tests.

rick-mesta:
Wow... perhaps I had previously typed the "_How are these changes getting tested to ensure no regression_" bit in a window and forgot about it. I genuinely meant the LGTM. Having said that, I think you're doing the right thing (long term) by trying to put unit tests in place that can be automated in the future 👍🏼

src/middlewared/middlewared/plugins/interface/lag.py
```diff
@@ -11,30 +11,75 @@ class Config:
...
         try:
             iface = netif.get_interface(name)
         except KeyError:
             iface = None
         else:
-            first_port = next(iter(iface.ports))
+            first_port = next(iter(iface.ports), None)
             if first_port is None or first_port[0] != members[0]['lagg_physnic']:
-                self.logger.info('Destroying existing %s as its first port has changed', name)
+                self.logger.info(f'Destroying {name} because its first port has changed')
```
sonicaj, line 23: For logging, we intentionally don't use f-strings as William/Vlad have pointed out earlier that if we don't use f-strings, the logging messages depending on the logging facility can be used to aggregate messages

src/middlewared/middlewared/plugins/interface/lag.py
```diff
@@ -11,30 +11,75 @@ class Config:
...
                 netif.destroy_interface(name)
                 iface = None
 
         if iface is None:
             netif.create_interface(name)
             iface = netif.get_interface(name)
 
+        # FIXME: this will fail on SCALE since we don't even have
```
sonicaj, line 31: Let's please create a ticket for this perhaps so we can test if this is required for SCALE and we can remove this entirely if that's not the case.

src/middlewared/middlewared/plugins/interface/lag.py
```diff
@@ -11,30 +11,75 @@ class Config:
...
+            'changed': False,
+            'protocol': None,
+            'xmit_hash_policy': None,
+            'lacpdu_rate': None,
+        }
         protocol = getattr(netif.AggregationProtocol, lagg['lagg_protocol'].upper())
         if iface.protocol != protocol:
-            self.logger.info('{}: changing protocol to {}'.format(name, protocol))
-            iface.protocol = protocol
+            info['changed'] = True
```
sonicaj, line 46: Can we maybe just check `changed` at the end of this if clauses and see if any value has a value differnet then `None`, we can consider this `changed` ?

src/middlewared/middlewared/plugins/interface/lag.py
```diff
@@ -11,30 +11,75 @@ class Config:
...
+            if info['xmit_hash_policy'] is not None:
+                self.logger.info(f'Changing xmit_hash_policy on "{name}" to {info["xmit_hash_policy"]}')
+                iface.xmit_hash_policy = info['xmit_hash_policy']
+
+            if info['lacpdu_rate'] is not None:
+                self.logger.info(f'Changing lacpdu_rate on "{name}" to {info["lacpdu_rate"]}')
+                iface.lacpdu_rate = info['lacpdu_rate']
+
+        # up the interface if options changed
+        iface.up() if info['changed'] else None
```
sonicaj, line 82: is this a simple statement ? We can do this in the above if clause ? I am not sure why we are using an inline if statement ?

src/middlewared/middlewared/plugins/interface/lag.py
```diff
@@ -58,7 +103,8 @@ def lag_setup(self, lagg, members, disable_capabilities, parent_interfaces, sync
             try:
                 port_iface = netif.get_interface(port[0])
             except KeyError:
-                self.logger.warn('Could not find {} from {}'.format(port[0], name))
+                self.logger.warning(f'Could not find {port[0]} from {name}')
```
sonicaj, line 106: Same as above

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1019,17 +1023,26 @@ async def do_create(self, data):
...
-                    'network.lagginterface',
-                    {'lagg_interface': interface_id, 'lagg_protocol': data['lag_protocol'].lower()},
-                )
-                lagports_ids += await self.__set_lag_ports(lag_id, data['lag_ports'])
+                async for interface_id in self.__create_interface_datastore(data, {'interface': name}):
+                    lag_proto = data['lag_protocol'].lower()
+                    xmit = lacpdu_rate = None
+                    if lag_proto in ('lacp', 'loadbalance'):
+                        xmit = data['xmit_hash_policy'].lower() if data['xmit_hash_policy'] else None
+                        if xmit is None:
```
sonicaj, line 1031: Is this required, we can just perhaps add the default value above in the inline if ?

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1019,17 +1023,26 @@ async def do_create(self, data):
...
+                            xmit = 'layer2+3'
+
+                        if lag_proto == 'lacp':
+                            lacpdu_rate = data['lacpdu_rate'].lower() if data['lacpdu_rate'] else None
+                            if lacpdu_rate is None:
```
sonicaj, line 1036: Same as above ?

sonicaj (changes requested):

src/middlewared/middlewared/plugins/interface/lag.py
yocalebo: Ticket has already been created: NAS-111238

src/middlewared/middlewared/plugins/interface/lag.py
```diff
@@ -11,30 +11,67 @@ class Config:
...
+            # setup for the first time so we have to down the
+            # interface before performing any of the actions
+            iface.down()
+
+            if info['protocol'] is not None:
+                # we _always_ have to start with the protocol
+                # information first since it deletes members
+                # (if any) of the current lagg and then changes
+                # the protocol
+                self.logger.info('Changing protocol on "%s" to %s', name, info['protocol'].name)
```
sonicaj, line 62: Let's please use `%r` instead of `"%s"`

src/middlewared/middlewared/plugins/interface/netif_linux/lagg.py
```diff
@@ -23,29 +22,50 @@ def create_lagg(name):
...
     def protocol(self, value):
-        run(["ip", "link", "set", self.name, "down"])
         for port in self.ports:
             self.delete_port(port[0])
         run(["ip", "link", "set", self.name, "type", "bond", "mode", value.value])
-        run(["ip", "link", "set", self.name, "up"])
+
+    @property
+    def xmit_hash_policy(self):
+        if self.protocol is not None:
```
sonicaj, line 41: Do we require this conditional ? The conditional below should be sufficient ?

src/middlewared/middlewared/plugins/interface/netif_linux/lagg.py
```diff
@@ -23,29 +22,50 @@ def create_lagg(name):
...
+                # this option only applies to 802.3ad and/or balance-xor
+                return self._sysfs_read(self.get_options_path("xmit_hash_policy")).split()[0]
+
+    @xmit_hash_policy.setter
+    def xmit_hash_policy(self, value):
+        run(["ip", "link", "set", self.name, "type", "bond", "xmit_hash_policy", value])
+
+    @property
+    def lacpdu_rate(self):
+        if self.protocol is not None and self.protocol == AggregationProtocol.LACP:
```
sonicaj, line 52: We can skip the first condition check as above here ?

sonicaj (commented):
LGTM, just some minor nits

src/middlewared/middlewared/plugins/interface/lag.py
yocalebo: Why? I'm printing the value of the enum object, which is a string?

src/middlewared/middlewared/plugins/network.py
```diff
@@ -948,6 +950,8 @@ async def commit(self, options):
         ]),
         List('bridge_members'),
         Str('lag_protocol', enum=['LACP', 'FAILOVER', 'LOADBALANCE', 'ROUNDROBIN', 'NONE']),
+        Str('xmit_hash_policy', enum=['LAYER2', 'LAYER2+3', 'LAYER3+4'], default=None, null=True),
```
sonicaj, line 953: I forgot :p Can we document the new columns in the return type entry please ?

sonicaj (approved):

src/middlewared/middlewared/plugins/interface/lag.py
sonicaj: I recommended it because `%r` uses `repr` which handles the quotes automatically and we don't have to add them manually each time.

yocalebo: Well that's convenient.

PR #7102, opened 2021-06-30
NAS-111259 / 21.08 / Fix R20/R20A to comply to enclosure management NEP

darkfiberiru:
Specifically returning single mapped enclosure for Head of system.
Previous method was not NEP compliant and did not work

### DIFF
--- a/src/middlewared/middlewared/plugins/enclosure.py
+++ b/src/middlewared/middlewared/plugins/enclosure.py
@@ -493,7 +493,7 @@ class Enclosure(object):
             if self.product_name in ["TRUENAS-R20", "TRUENAS-R20A"]:
                 self.model = self.product_name.replace("TRUENAS-", "")
                 self.controller = True
-            elif MINI_REGEX.match(self.encname):
+            elif MINI_REGEX.match(self.product_name):
                 # TrueNAS Mini's do not have their product name stripped
                 self.model = self.product_name
                 self.controller = True
--- a/src/middlewared/middlewared/plugins/enclosure.py
+++ b/src/middlewared/middlewared/plugins/enclosure.py
@@ -369,14 +369,6 @@ class Enclosures(object):
             enclosure = Enclosure(num, data, stat, product_name)
             if any(s in enclosure.encname for s in blacklist):
                 continue
-            if (
-                product_name and product_name in ["TRUENAS-R20", "TRUENAS-R20A"] and
-                enclosure.encname == "AHCI SGPIO Enclosure 2.00"
-            ):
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/enclosure.py
```diff
@@ -497,23 +489,21 @@ def _set_model(self, data):
         elif R_SERIES_REGEX.match(self.encname) or R20A_REGEX.match(self.encname):
             self.model = self.product_name.replace("TRUENAS-", "")
             self.controller = True
-            if self.model in ["R20", "R20A"]:
-                self.model = f"{self.model}, Drawer #1"
         elif (
-            self.product_name in ["TRUENAS-R20", "TRUENAS-R20A"] and
             self.encname == "AHCI SGPIO Enclosure 2.00" and
-            len(data.splitlines()) == 6
+            self.system_info["system_product"] in ["TRUENAS-R20", "TRUENAS-R20A"]
```
yocalebo, line 494: `self.system_info` does not exist so this will fail.
```suggestion
            self.product_name in ["TRUENAS-R20", "TRUENAS-R20A"]
```

Furthermore, we have a bug on line 541 and 543 because it also makes a reference to `system_info["system_product"]` which need to both be changed to `self.product_name`

yocalebo (changes requested):

src/middlewared/middlewared/plugins/enclosure.py
```diff
@@ -497,23 +489,21 @@ def _set_model(self, data):
...
         ):
-            self.model = f"{self.product_name.replace('TRUENAS-', '')}, Drawer #2"
+            self.model = self.product_name.replace("TRUENAS-", "")
+            self.controller = True
+        elif self.encname.startswith("iX TrueNAS R20p"):
```
yocalebo, line 498: Let's make this its own `R20_REGEX` and then move the check inline to line 489.

```
elif R_SERIES_REGEX.match(self.encname) or R20A_REGEX.match(self.encname) or R20_REGEX.match(self.encname):

yocalebo (changes requested):

src/middlewared/middlewared/plugins/enclosure.py
darkfiberiru: Should be resolved now

src/middlewared/middlewared/plugins/enclosure.py
darkfiberiru: Should be resolved now

yocalebo (approved):

PR #7106, opened 2021-07-01
NAS-111273 / 21.08 / Fix R50 to comply with enclosure NEP

darkfiberiru:
-- Removes enclosure.concatanate as we have replaced that by mapping
instead in all cases
-- On R50 Make sure 1st expander is always first irrelavant of wiring
-- Ensure no jbods are enumerated when mapping drives as additional
safety
--Remove enclosures_tail as were handling with mapping


src/middlewared/middlewared/plugins/enclosure_/map.py
```diff
@@ -155,11 +210,13 @@ async def map_enclosures(self, enclosures):
         return enclosures
 
     async def _map_enclosures(self, enclosures, slots):
+        #Ensure jbods can't messup ordering by filtering them out
```
yocalebo, line 213:
```suggestion
        # Ensure JBODs don't effect ordering by filtering them out
```

src/middlewared/middlewared/plugins/enclosure_/map.py
```diff
@@ -155,11 +210,13 @@ async def map_enclosures(self, enclosures):
...
+        controller_enclosures = list(filter(lambda x: x['controller'],enclosures))
```
yocalebo, line 214:
```suggestion
        controller_enclosures = list(filter(lambda x: x['controller'], enclosures))
```

yocalebo (changes requested):

src/middlewared/middlewared/plugins/enclosure.py
```diff
@@ -104,14 +104,15 @@ def query(self, filters, options):
                         "elements": elements,
                         "has_slot_status": has_slot_status
                     })
-
-            enclosures.append(enclosure)
+            #Ensure first R-series expander is in order independent of cabling
+            if "eDrawer4048S1" in enclosure['name']:
```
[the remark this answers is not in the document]
darkfiberiru: Not valid need specifically S1

yocalebo (approved):

PR #7108, opened 2021-07-02
NAS-111283 / 21.08 / Bug fix for recommending zvol blocksize

sonicaj:
This commit fixes an issue where we are not sending 'K' suffix with the recommended blocksize which we were doing before.


yocalebo (approved):

yocalebo:
backport

PR #7115, opened 2021-07-06
NAS-111317 / 21.08 / fix traceback when creating bond type interfaces on SCALE

yocalebo:
This fixes a race when adding bond type interfaces. In some situations, when adding a bond, a udevd event would be generated as expected. The race occurs between `interface.query()` calling `__getstate__()` and the time it takes to setup the lagg. In the failure condition, `__getstate__()` would set `protocol: self.protocol.name` but `self.protocol` would be `None` in this situation causing a `AttributeError: 'NoneType' object has no attribute 'name'`. While I'm here I've done some research and I've found that this method was added back in 2018 [here](https://github.com/truenas/middleware/pull/1368/files) which also fixes a race condition when adding laggs on freeBSD. I've made the following changes based on historical research as well as local testing.

1. remove the unused TN CORE code
2. import `CLONED_PREFIXES` since it's relevant to SCALE and use it in the `udevd_ifnet_hook` function so we can return earlier on the udevd event which saves us from the race condition mentioned above.
3. `epair` is a TN CORE only interface
4. fix `self.logger.info` call in `sync_interface` to be Sentry friendly
5. add docstring to `udevd_ifnet_hook`
6. do not crash `interface.query` in `__getstate__` when trying to set `protocol` information

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -2157,7 +2157,7 @@ class InterfaceService(CRUDService):
                 [('int_interface', '=', name)], {'get': True}
             )
         except IndexError:
-            self.logger.info('{} is not in interfaces database'.format(name))
+            self.logger.info('%s is not in interfaces database', name)
             return
 
         aliases = await self.middleware.call(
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -2735,40 +2735,26 @@ async def configure_http_proxy(middleware, *args, **kwargs):
...
+    """
+    if data.get('SUBSYSTEM', '') != 'net' and data.get('ACTION', '') != 'add':
         return
 
-    await attach_interface(middleware, data.get('subsystem'))
-
-
-async def udevd_ifnet_hook(middleware, data):
-    if data.get('SUBSYSTEM') != 'net' and data.get('ACTION') != 'add':
+    int_name = data.get('INTERFACE', None)
```
sonicaj, line 2750: It would be `None` already if `INTERFACE` is not present

src/middlewared/middlewared/plugins/network.py
```diff
@@ -2735,40 +2735,26 @@ async def configure_http_proxy(middleware, *args, **kwargs):
...
-
-    iface = await middleware.call('interface.query', [('name', '=', iface)])
-    if not iface:
-        return
-
-    iface = iface[0]
-    # We only want to sync physical interfaces that are hot-plugged,
-    # not cloned interfaces with might be a race condition with original devd.
-    # See #33294 as an example.
-    if iface['state']['cloned']:
```
sonicaj, line 2753: Would this be `false` for a cloned interface ?

sonicaj (changes requested):
I am not sure i follow, we are just fixing a usage if i follow correctly when an interface is attached,  why don't we actually fix `interface.query` as well so that the error described in the description does not occur ? An example is `interface.query` being called by UI maybe or some other client while the race condition is in play and it failing and we having the error still being logged ?

src/middlewared/middlewared/plugins/network.py
yocalebo: Explicitly specifying the default value is proper.

src/middlewared/middlewared/plugins/network.py
yocalebo: Yes it would be `false`. As a matter of fact, anything in `CLONED_PREFIXES` is tagged as `iface['state']['cloned']`. However, we're calling `interface.query()` before the bond is fully setup which leads to the race condition that I mention in the description.

sonicaj: Exactly my point, we don't need to have this complete diff. We should fix the root cause which is `interface.query` failing.

yocalebo: Why though? This avoids calling `interface.query` entirely since we don't even care about these interfaces in the first place?

yocalebo: And fixing `interface.query` is just suppressing a known failure scenario. It's not fixing the root cause by any means.

sonicaj: To be sure i follow, are you suggesting that we should not fix `interface.query` and have a traceback ? Which means regardless of the changes any client consuming `interface.query` would error out including UI ?
Also i am not i follow the root cause in this case, isn't it `interface.query` because it's not reporting the protocol correctly as it's still being set and errors out ? From my understanding, we are just fixing a side effect of the original problem and not the problem itself ?
I think the best way forward would be able to report that the protocol is not set yet in `interface.query` and that should cover the current diff as well and any other client consuming this endpoint won't error out as well.

yocalebo: You're right, if we have a 3rd party API consumer calling `interface.query` timed properly then it will still traceback if another consumer (i.e. webUI) is setting up a bond interface.

The changes here fix the traceback just from a webUI consumer standpoint since it triggers a `udevd` hook. So I'll ask you 3 follow-up questions.

1. Would you want the `protocol` key in `__getstate__()` to return null or an empty string?
2. Keeping the current diff still fixes the race condition that causes the `NoneType` specifically with the `udevd_ifnet_hook` function so, imo, I believe the fix is still correct would you agree?
3. if you agree to question number 2, how about I change the `__gestate__()` method as described in this PR too?

sonicaj: For (1), i think we can return `None` and perhaps another flag `protocol_set` which can be a boolean value of `protocol`.
I don't follow why we would need (2) if we fix (1) ?

sonicaj (approved):
LGTM, thanks!

PR #7116, opened 2021-07-07
NAS-111324 / 21.08 / catch traceback when configuring default routes

yocalebo:
When all configured interfaces have been deleted from the system, `dhclient` is called for all detected interfaces. In this scenario, we have a race that can occur. An example scenario:
1. delete all configured interfaces on system
2. interface.sync() gets called and starts dhcp on all interfaces detected on the system
3. route.sync() gets called which eventually calls dhclient_leases which reads a file on disk to see if we have any previously defined default gateways from dhclient. However, by the time we read this file, dhclient could still be requesting an address from the DHCP server
4. so when we try to install our own default gateway manually (even though dhclient will
 do this for us) it will fail expectedly here.

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -10,8 +10,9 @@ from middlewared.utils import osc
 from middlewared.utils.generate import random_string
 from middlewared.validators import Hostname, Match, Range
 
-import asyncio
 from collections import defaultdict
+from pyroute2.netlink.exceptions import NetlinkError
+import asyncio
 import contextlib
 import ipaddress
 import itertools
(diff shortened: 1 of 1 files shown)


rick-mesta (approved):
you fooled me 😁

src/middlewared/middlewared/plugins/network.py
```diff
@@ -2338,9 +2338,15 @@ async def sync(self):
                     leases = await self.middleware.call('interface.dhclient_leases', interface)
                     reg_routers = re.search(r'option routers (.+);', leases or '')
                     if reg_routers:
-                        # Make sure to get first route only
                         ipv4_gateway = reg_routers.group(1).split(' ')[0]
-                        break
+                        if ipv4_gateway:
```
sonicaj, line 2342: If we do not want to proceed further, is the complete if conditional required starting at L2322 ?

src/middlewared/middlewared/plugins/network.py
```diff
@@ -2338,9 +2338,15 @@ async def sync(self):
...
+                            # This means that the end-user hasn't specified
+                            # a default gateway in the webUI, however, we
+                            # have one from dhclient running on one of the
+                            # interfaces. In this scenario, there is nothing
+                            # to do so return.
+                            return
```
sonicaj, line 2348: We will not remove a gateway which is not in database but is configured, so this would be change of behavior ?

sonicaj: Also ipv6 gateway would be ignored completely

sonicaj (changes requested):

yocalebo:
@sonicaj I truly do not want to figure out all the race conditions since we seem to be completely ignoring the default gateway that gets installed by `dhclient`. Instead I'll just catch this specific failure condition and make a comment on why this can occur.

src/middlewared/middlewared/plugins/network.py
```diff
@@ -2350,7 +2351,26 @@ async def sync(self):
...
+                        #       calls dhclient_leases which reads a file on
+                        #       disk to see if we have any previously
+                        #       defined default gateways from dhclient.
+                        #       However, by the time we read this file,
+                        #       dhclient could still be requesting an
+                        #       address from the DHCP server
+                        #   4. so when we try to install our own default
+                        #       gateway manually (even though dhclient will
+                        #       do this for us) it will fail expectedly here.
+                        pass
```
sonicaj, line 2373: Let's please log this at least for scenarios where the race condition does not hold and we have a gateway configured already which cannot be added in the current system network state.

src/middlewared/middlewared/plugins/network.py
```diff
@@ -2350,7 +2351,27 @@ async def sync(self):
...
+                    #       disk to see if we have any previously
+                    #       defined default gateways from dhclient.
+                    #       However, by the time we read this file,
+                    #       dhclient could still be requesting an
+                    #       address from the DHCP server
+                    #   4. so when we try to install our own default
+                    #       gateway manually (even though dhclient will
+                    #       do this for us) it will fail expectedly here.
+                    # Either way, let's log the error.
+                    self.logger.warning('Failed adding %s as default gateway: %r', ipv4_gateway, e)
```
sonicaj, line 2374: `logger.error` ?

sonicaj (approved):
Thanks

PR #7117, opened 2021-07-07
NAS-111011 / 21.08 / Wait for vdev removal after a device is removed from zfs pool

sonicaj:
This PR adds changes to wait for a vdev to be removed before moving on to wiping disks. This is required for cases where we have top level vdevs being removed and `zpool remove` is not a synchronous operation in this context which results in middleware trying to wipe disks while removal of vdev is still in place.

### DIFF
--- a/src/middlewared/middlewared/plugins/zfs_/pool_wait.py
+++ b/src/middlewared/middlewared/plugins/zfs_/pool_wait.py
@@ -1,0 +1,30 @@
+import libzfs
+
+from middlewared.schema import accepts, Dict, Str
+from middlewared.service import CallError, Service
+
+
+POOL_ACTIVITY_TYPES = [a.name for a in libzfs.ZpoolWaitActivity]
+
+
+class ZFSPoolService(Service):
+
+    class Config:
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/zfs_/pool_wait.py
```diff
@@ -0,0 +1,30 @@
...
+        process_pool = True
+
+    @accepts(
+        Str('pool_name'),
+        Dict(
+            'options',
+            Str('activity_type', enum=POOL_ACTIVITY_TYPES, required=True),
+        )
+    )
+    def wait(self, pool_name, options):
```
yocalebo, line 24: Since this is public, do we not need a `returns()` ?

sonicaj: Nope, the service is private ;)

yocalebo: Ahh right

yocalebo (approved):

PR #7120, opened 2021-07-08
NAS-111339 / 21.08 / Add iptables rules with counters to debug

sonicaj:
This commit adds changes to add iptables rules with counters to debug. It can be helpful in situations when we want to debug if there are concerns with some iptable rule(s) and we can look up to counters to extract the information and make an analysis.

### DIFF
--- a/src/freenas/usr/local/libexec/freenas-debug/network/network.sh
+++ b/src/freenas/usr/local/libexec/freenas-debug/network/network.sh
@@ -130,8 +130,8 @@ network_func()
 	ip rule list
 	section_footer
 
-	section_header "Iptables Rules (iptables-save)"
-	iptables-save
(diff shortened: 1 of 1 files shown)


src/freenas/usr/local/libexec/freenas-debug/network/network.sh
```diff
@@ -134,6 +134,10 @@ network_func()
 	iptables-save
 	section_footer
 
+	section_header "Iptables Rules with counters (iptables -L -v -n)"
```
yocalebo, line 137: We call `iptables-save` right above this which is the same information just without the counters. Instead of calling this multiple times, you can probably just remove `iptables-save`

yocalebo (changes requested):

yocalebo (approved):

PR #7142, opened 2021-07-13
NAS-111326 / 21.08 / Core dumps on SCALE

themylogin:


anodos325:
The problem with simply removing the coredump files is that references to it will remain.
```
truenas# coredumpctl
TIME                            PID   UID   GID SIG COREFILE  EXE
Mon 2021-07-12 12:21:22 PDT  154313     0     0  11 missing   /usr/bin/sockstat
```

I think rather than removing coredumps, we should use one of the following config options for coredump.conf:
```
       MaxUse=, KeepFree=
           Enforce limits on the disk space taken up by externally stored core dumps.  MaxUse= makes sure that old core dumps are removed as soon as the total disk space taken up by core dumps grows beyond this limit (defaults to 10% of the total disk size).
           KeepFree= controls how much disk space to keep free at least (defaults to 15% of the total disk size). Note that the disk space used by core dumps might temporarily exceed these limits while core dumps are processed. Note that old core dumps are also
           removed based on time via systemd-tmpfiles(8). Set either value to 0 to turn off size-based clean-up.
```

anodos325:
That said, the default 15% minimum free is probably good enough. In this case, then we can simply remove the code where we delete core files.

src/middlewared/middlewared/alert/source/cores.py
```diff
@@ -30,19 +29,15 @@ def check_sync(self):
...
             if not os.path.isfile(path):
                 continue
 
             unlink = False
-            if IS_FREEBSD and file == "syslog-ng.core":
-                unlink = True
-            elif file == "su.core":
-                unlink = True
-            elif os.stat(path).st_mtime < time.time() - 86400 * 5:
+            if os.stat(path).st_mtime < time.time() - 86400 * 5:
```
anodos325, line 40: `coredumpctl --list` output contains a timestamp.

anodos325:
For maximum scope-creep, maybe we could use a middleware plugin that wraps around `coredumpctl`. No ponies required... yet.

rick-mesta:
@anodos325: Has @themylogin properly addressed your concerns on this bad boy ? or are there additional deltas you'd like to see ?

anodos325:
comments above still apply.

themylogin:
@anodos325 sorry, git push failed due to rebase

src/middlewared/middlewared/alert/source/cores.py
```diff
@@ -1,54 +1,31 @@
...
-            else:
-                corefiles.append(file)
-
-            if unlink:
-                os.unlink(path)
+    async def check(self):
+        corefiles = []
+        for coredump in await self.middleware.call("system.coredumps"):
+            if coredump["corefile"] == "present":
+                corefiles.append(f"{coredump['exe']} ({coredump['time']})")
```
yocalebo, line 28: We need to ignore `syslog-ng` core dumps here as well. Someone else has already opened up a ticket for it core dumping on SCALE today. NAS-111566

yocalebo (changes requested):

yocalebo (approved):

anodos325 (approved):

PR #7147, opened 2021-07-14
NAS-111395 / 21.08 / Add support for policy based routing for kubernetes

sonicaj:
This PR adds changes to add support for policy based routing for kubernetes. This was being provided before by kube-router itself when we were using overlay for pod networking but it was disabled due to another issue https://github.com/truenas/kube-router/pull/4 where due to ip in ip tunneling packets with fragmentation bit unset could potentially get dropped. Talking upstream about it, kube-router dev advised that server/clients should make sure that they conform to PMTU and kube-router should allow such connections, however with latest version from upstream the issue mentioned could still easily be reproduced with `curl`.

This PR adds following changes to be specific:
1) We have a bug upstream where routes are not properly cleaned, we don't make cleaning that fatal as other routes which are to be cleaned by us cannot be cleaned - so temporarily making the call non fatal.
2) Do not wait for `kube-router` table to be created and create it ourselves if it does not exist.
3) Add a rule to make sure that pod traffic gives preference to the rule(s) defined in `kube-router` table to enforce policy based routing.
4) Properly cleanup custom rule(s) added.
5) Add ability to add/remove/query ip rules.

### DIFF
--- a/src/middlewared/middlewared/plugins/interface/netif_linux/routing.py
+++ b/src/middlewared/middlewared/plugins/interface/netif_linux/routing.py
@@ -233,3 +233,23 @@ class RoutingTable:
             kwargs[key] = value
 
         ip.route(op, **kwargs)
+
+
+class RuleTable:
+
+    @property
+    def rules(self):
+        rules = []
+        tables = {t.table_id: t for t in RoutingTable().routing_tables.values()}
+        for rule in ip.get_rules():
+            attrs = dict(rule['attrs'])
+            if not all(k in attrs for k in ('FRA_TABLE', 'FRA_PRIORITY')):
+                continue
+
+            rules.append({
+                'table': tables[attrs['FRA_TABLE']],
+                'priority': attrs['FRA_PRIORITY'],
+                'source_addr': attrs.get('FRA_SRC'),
+            })
+
+        return rules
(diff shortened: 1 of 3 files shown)


themylogin (approved):

src/middlewared/middlewared/plugins/interface/netif_linux/routing.py
```diff
@@ -225,3 +233,35 @@ def _op(self, op, route):
...
+
+
+class RuleTable:
+
+    @property
+    def rules(self):
+        rules = []
+        tables = {t.table_id: t for t in RoutingTable().routing_tables.values()}
+        for rule in ip.get_rules():
+            attrs = dict(rule['attrs'])
```
yocalebo, line 245: For most of the IPRoute examples that I've found, checking if the `attrs` key in the `rule` object should be done first.

```
if 'attrs' not in rule:
     continue
 ```
 
 Furthermore, I'm not sure it's necessary to convert the `attrs` attribute to a dict since the `rule` object has built-in methods for getting the necessary items. Examples below.
 
 ```
 table = rule.get_attr('FRA_TABLE')
 if table is None:
     continue
     
prio = rule.get_attr('FRA_PRIORITY')
if prio is None:
    continue

src = rule.get_attr('FRA_SRC')
...
```

sonicaj: We have been using `attrs` directly for routes and it always exist so far ? I have found the same to be true for ip rules as well.

yocalebo (approved):

PR #7150, opened 2021-07-14
NAS-111333 / 21.08 / Filter "number of Error Log entries increased" for nvme devices until upstream bugs are fixed.

themylogin:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=900244
https://bugs.launchpad.net/ubuntu/+source/smartmontools/+bug/1878264


sonicaj (approved):
Should we maybe force the case to be same on both sides just to be on the safe side ?

PR #7156, opened 2021-07-14
NAS-108723 / 21.08 / Implement NVMe resize in disk_resize for FreeBSD.

amotin:


src/freenas/usr/local/sbin/disk_resize
```diff
@@ -45,9 +45,13 @@ else
 	echo "Resizing ${dev} to full capacity"
 fi
 
-# ATA and SCSI have completely different resize commands.
+# ATA, SCSI and NVMe have completely different resize commands.
 # SAT we handle as ATA via pass-through, but with some complications.
-devtype=`camcontrol devtype ${dev} 2>/dev/null`
+if nvmecontrol nsid ${dev} >/dev/null 2>/dev/null; then
+	devtype="nvme";
```
yocalebo, line 51: Minor nit but seems to be a trailing semicolon here when not needed.

yocalebo (approved):
Minor bit, but looks good

src/freenas/usr/local/sbin/disk_resize
amotin: Fixed.

PR #7164, opened 2021-07-15
NAS-111194 / 21.08 / document return types in failover plugin(s)

yocalebo:
- document return types in failover plugin(s)
- while here, remove the `_linux.py` suffix for the failover related files


src/middlewared/middlewared/plugins/failover_/remote.py
```diff
@@ -224,6 +224,7 @@ async def remote_ip(self):
             Any('callback'),
         ),
     )
+    @returns(Any())
```
sonicaj, line 227: `null=False` by default, can you please set that to `true` ?

src/middlewared/middlewared/plugins/failover.py
```diff
@@ -90,6 +90,12 @@ async def failover_extend(self, data):
         Bool('master', null=True),
         update=True,
     ))
+    @returns(Dict(
+        Int('id'),
+        Bool('disabled'),
+        Int('timeout'),
```
sonicaj, line 96: I suppose these are always going to be present ? We should add `required=True` in that case

sonicaj (changes requested):
Have we documented `failover.config` ?

src/middlewared/middlewared/plugins/failover.py
```diff
@@ -78,6 +78,14 @@ class Config:
         datastore_extend = 'failover.failover_extend'
         cli_namespace = 'system.failover'
 
+    ENTRY = Dict(
+        'failover_entry',
+        Int('id', required=True),
+        Bool('disabled', required=True),
+        Int('timeout', required=True),
+        Bool('master', required=True),
```
sonicaj, line 86: `master` can be `null` i assume ?

sonicaj (commented):
Have we deleted all the _linux files ?

src/middlewared/middlewared/plugins/failover.py
yocalebo: No, it always exists in the db as a boolean.

sonicaj: Why do we allow it to be null on update method ?

yocalebo: If it's null, then we figure out which node is to become (or stay) the master node.

sonicaj (approved):
Thanks!

PR #7170, opened 2021-07-16
NAS-111375 / 21.08 / remove sockstat calls on SCALE

yocalebo:
`sockstat` on linux is buggy and unmaintained. Use linux native tool (`lsof`) to accomplish the task.


themylogin (commented):
Have you tested if it works well with IPv6?

themylogin (approved):
LGTM!

PR #7175, opened 2021-07-19
NAS-110816 / 12.0 / Set bootloader property so zectl can update grub for core

sonicaj:
This commit adds changes to set bootloader property for BE's so that zectl is able to consume our grub plugin to update grub after making changes to boot environments.


src/middlewared/middlewared/plugins/update_/install.py
```diff
@@ -70,6 +68,14 @@ def install_scale(self, mounted, progress_callback):
             else:
                 raise CallError(stderr)
 
+        cp = subprocess.Popen(
```
themylogin, line 71: Doesn't this belong to `truenas_install`?

PR #7177, opened 2021-07-19
NAS-110970 / 12.0 / Refer to multipaths instead of multipath members for disk temperatures and S.M.A.R.T. alerts

themylogin:


themylogin:
backport

PR #7184, opened 2021-07-20
NAS-111453 / 21.08 / if no webdav shares are configured, dont reload

yocalebo:
We call `service.reload <service>` for _all_ services on the system (whether they're enabled or not) after a zpool has been imported via the `post_import` hook.

To fix this, instead of going through the process of `etc.generate webdav`, we return early if there are no webdav shares configured on the system. This prevents the `apache2` daemon from being started when a zpool is imported.

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/apache24/Includes/webdav.conf.mako
+++ b/src/middlewared/middlewared/etc_files/local/apache24/Includes/webdav.conf.mako
@@ -7,13 +7,14 @@
 	import subprocess
 
 	from string import digits, ascii_uppercase, ascii_lowercase
-	from middlewared.utils import osc
+
+	web_shares = middleware.call_sync('sharing.webdav.query', [['enabled', '=', True]])
+	if not web_shares:
+		raise FileShouldNotExist()
 
 	# Check to see if there is a webdav lock databse directory, if not create
 	# one. Take care of necessary permissions whilst creating it!
-	apache_dir = service.APACHE_DIR
-	if osc.IS_LINUX:
-		apache_dir = apache_dir.replace('local/', '')
+	apache_dir = service.APACHE_DIR.replace('local/', '')
 	oscmd = f'/etc/{apache_dir}/var'
 	if not os.path.isdir(oscmd):
 		os.mkdir(oscmd, 0o774)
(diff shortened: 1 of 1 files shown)


rick-mesta (approved):
LGTM: @anodos325 / @themylogin - would either/both of you guys also take a look ? 🙏🏼

src/middlewared/middlewared/etc_files/local/apache24/Includes/webdav.conf.mako
```diff
@@ -7,13 +7,14 @@
...
+	web_shares = middleware.call_sync('sharing.webdav.query', [['enabled', '=', True]])
+	if not web_shares:
+		raise FileShouldNotExist()
 
 	# Check to see if there is a webdav lock databse directory, if not create
 	# one. Take care of necessary permissions whilst creating it!
-	apache_dir = service.APACHE_DIR
-	if osc.IS_LINUX:
-		apache_dir = apache_dir.replace('local/', '')
+	apache_dir = service.APACHE_DIR.replace('local/', '')
```
themylogin, line 17: Can we just change the original constant?

themylogin (changes requested):
So `service.reload` starts apache2 even if it was not started yet? Does this happen for all services of just for apache? Can we have a more explicit way of preventing this than just providing invalid configuration? (also why does apache not start when this file is not present? Is it included explicitly?)

yocalebo:
> So `service.reload` starts apache2 even if it was not started yet?
Yes, the service is started since we create an empty `webdav.conf` file.

> Does this happen for all services of just for apache?
We were doing the same thing for proftpd in 11.3 but we also fixed it there. I've only looked at `apache2` service to see if it's the only one with this behavior.

> Can we have a more explicit way of preventing this than just providing invalid configuration?
I can change it so that the hook that gets called on pool import (`post_import`) for the webdav actually checks to see if there are any configured webdav shares and then `reload` it appropriately.

> (also why does apache not start when this file is not present? Is it included explicitly?)
See my first answer

themylogin:
@yocalebo won't it be cleaner to reload service only if it is enabled or started? Current code will start WebDAV if shares are present but service is disable which I believe is an incorrect behavior.

PR #7188, opened 2021-07-20
NAS-111447 / 21.08 / Add POSIX_HOME preset

anodos325:


anodos325:
This time with the correct bug number

rick-mesta (approved):

undsoft:
Bump.

anodos325:
Will address tomorrow. Today was spent putting out fires with clustering. May rework so that we have a proper CRUD API for ACL templates and keep around the default acl stuff as a legacy wrapper. So a little more work, but a little more standard and better for future.

rick-mesta:
@anodos325: ping... `21.08` train looking awfully close 😬

PR #7197, opened 2021-07-21
NAS-111436 / 21.08 / Fix core.get_methods crash

themylogin:


rick-mesta (approved):
LGTM: I'd feel more comfortable with a shout out from @anodos325 and/or @yocalebo (if possible)

yocalebo (approved):
This looks good, however, I'm curious if this was failing while middlewared was started normally (i.e. daemonized in background) or if this is only failing when starting middlewared in foreground mode?

themylogin:
@yocalebo it was only failing at JSON serialization; internal method calls worked normally

PR #7199, opened 2021-07-21
NAS-111496 / 21.08 / fix and improve webdav API tests

yocalebo:
With this commit https://github.com/truenas/middleware/pull/7184, a singular API test fails because the particular API test just sends a `service.reload webdav` call expecting the service to be started. This happens even if there are no webdav shares. Before my change, webdav service would start because we had an empty webdav.conf file. After my change, the service will not start unless a webdav share has been created. This resolves a long-standing issue where the webdav service gets automatically started _always_ on a zpool create/import even if there are no webdav shares.

Many fixes/improvements:
- fixes the failing API test by removing that test because we have a fairly extensive webdav test suite that tests the functionality by itself
- remove the `test_560_webdav_osx.py` as it is, literally, a duplicate file to `test_560_webdav_bsd.py`. As a matter of fact, the only thing different between those 2 files is the name of the dataset that we create.
- rename `test_560_webdav_bsd.py` to `test_560_webdav.py` since this only runs on SCALE (which is not BSD)....
- add quite a few more webdav tests to verify certain changes


themylogin (approved):

ericbsd (approved):
LGTM

PR #7202, opened 2021-07-22
NAS-111238 / 21.08 / interface capabilities (known as features) on SCALE

yocalebo:
Add `capabilities` to SCALE. While I'm here get rid of `subprocess`'ing out to `ethtool` to get media information. `pyroute2.ethtool` provides API backed by native ioctl that gets the information for us. Also, allows us to fill in the `supported_media` attribute as well.

### DIFF
--- a/src/middlewared/middlewared/plugins/interface/netif_linux/ethernet_settings.py
+++ b/src/middlewared/middlewared/plugins/interface/netif_linux/ethernet_settings.py
@@ -1,0 +1,107 @@
+from logging import getLogger
+from pyroute2.ethtool import Ethtool
+from pyroute2.ethtool.ioctl import NotSupportedError
+
+
+logger = getLogger(__name__)
+
+
+class EthernetHardwareSettings:
+
+    def __init__(self, interface):
+        self._name = interface
+        self._eth = Ethtool()
+        self._caps = self.__capabilities__()
+        self._media = self.__mediainfo__()
+
+    def __capabilities__(self):
+        result = {'enabled': [], 'disabled': [], 'supported': []}
+        try:
+            for i in self._eth.get_features(self._name):
+                for feature, value in i.items():
+                    if value.enable:
+                        result['enabled'].append(feature)
+                    else:
+                        result['disabled'].append(feature)
+                    result['supported'].append(feature)
+        except Exception:
+            logger.error('Failed to get capabilities for %s', self._name, exc_info=True)
+
+        return result
+
+    @property
+    def enabled_capabilities(self):
+        return self._caps['enabled']
+
+    @property
+    def disabled_capabilties(self):
+        return self._caps['disabled']
+
+    @property
+    def supported_capabilities(self):
+        return self._caps['supported']
+
+    def __mediainfo__(self):
+        result = {
+            'media_type': '',
+            'media_subtype': '',
+            'active_media_type': '',
+            'active_media_subtype': '',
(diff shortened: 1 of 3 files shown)


yocalebo:
This is to be merged for 21.08, however, another PR will follow to actually provide the ability to enable and/or disable capabilities for a given interface. Needs quite a bit more investigation since enabling/disabling revolves around our HA code.

src/middlewared/middlewared/plugins/interface/netif_linux/ethernet_settings.py
```diff
@@ -0,0 +1,107 @@
...
+            'active_media_subtype': '',
+            'supported_media': [],
+        }
+
+        try:
+            attrs = self._eth.get_link_mode(self._name)
+            mst = 'Unknown'
+            if attrs.speed is not None:
+                # looks like 1000Mb/s, 10000Mb/s, etc
+                mst = ''.join((str(attrs.speed), 'Mb/s'))
```
themylogin, line 58: Why not a format string?

src/middlewared/middlewared/plugins/interface/netif_linux/ethernet_settings.py
```diff
@@ -0,0 +1,107 @@
...
+            result['active_media_subtype'] = mst  # just matches media_subtype....gross
+            result['supported_media'].extend(attrs.supported_modes)
+        except NotSupportedError:
+            # saw this on a VM running inside xen where the
+            # nic driver being used doesnt report any type
+            # of media info (ethtool binary didnt report anything either)
+            # so ignore these errors
+            pass
+        except Exception as e:
+            logger.error('Failed to get media info for %s: %r', self._name, e)
```
themylogin, line 76: We have a non-trivial code above, won't we be interested in a stack trace?

src/middlewared/middlewared/plugins/interface/netif_linux/interface.py
```diff
@@ -118,26 +114,18 @@ def __getstate__(self, address_stats=False, media=False, vrrp_config=None):
...
-                    state.update({
-                        "media_type": "Ethernet",
-                        "media_subtype": "autoselect" if ethtool.get("Auto-negotiation") == "on" else media_subtype,
-                        "active_media_type": "Ethernet",
-                        "active_media_subtype": media_subtype,
-                    })
+        with EthernetHardwareSettings(self.name) as dev:
+            state.update({
+                'capabilities': dev.enabled_capabilities,
+                'supported_media': dev.supported_media,
```
themylogin, line 120: Do we need this if `media` is `False`?

And we'll still call `__mediainfo__` if `media` is `False`

themylogin (changes requested):

src/middlewared/middlewared/plugins/interface/netif_linux/interface.py
yocalebo: Seems we always populate `capabilities` attribute on CORE so I tried to keep the same behavior. However, I thought getting the media info was cheap enough to always get it but I changed it anyways. Or are you saying that we shouldn't populate `capabilities` or any of the media attributes unles `medi == True`?

themylogin: Well, it's really cheap enough, we can drop `media` argument then.

themylogin (approved):
Thank you!

PR #7203, opened 2021-07-22
NAS-111505 / 21.08 / Add support for fuse-mounted paths to filesystem plugin

anodos325:
Paths that have prefix "CLUSTER:" receive special handling in
filesystem plugin. Rather than evaluating a normal path under
/mnt, checks are performed under /cluster. CallError is raised
with errno ENXIO if the gluster volume is not mounted.

This allows modification of permissions on mounted gluster volumes.
Explicitly prevent filesystem access to ctdb shared volume to
prevent alterations that may break clustering. In this case
raise CallError with errno EPERM.

Sample path:
`CLUSTER:smb01/data`
Which resolves to /data inside the smb01 volume.

### DIFF
--- a/src/middlewared/middlewared/plugins/filesystem_/acl_linux.py
+++ b/src/middlewared/middlewared/plugins/filesystem_/acl_linux.py
@@ -31,14 +31,29 @@ class FilesystemService(Service, ACLBase):
         if acltool.returncode != 0:
             raise CallError(f"acltool [{action}] on path {path} failed with error: [{acltool.stderr.decode().strip()}]")
 
-    def _common_perm_path_validate(self, schema, path, recursive, verrors):
+    def _common_perm_path_validate(self, schema, data, verrors):
+        is_cluster = data['path'].startswith('CLUSTER')
+        try:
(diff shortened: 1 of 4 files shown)


yocalebo (approved):

src/middlewared/middlewared/plugins/filesystem_/acl_linux.py
```diff
@@ -31,16 +31,31 @@ def acltool(self, path, action, uid, gid, options):
         if acltool.returncode != 0:
             raise CallError(f"acltool [{action}] on path {path} failed with error: [{acltool.stderr.decode().strip()}]")
 
-    def _common_perm_path_validate(self, schema, path, recursive, verrors):
+    def _common_perm_path_validate(self, schema, data, verrors):
+        is_cluster = data['path'].startswith('CLUSTER')
```
rick-mesta, line 35: other places specifically look for `CLUSTER:` (colon suffix)... will this be a problem ?

anodos325: I'll fix for consistency.

rick-mesta (approved):

PR #7213, opened 2021-07-27
NAS-111444 / 21.08 / Improvements to Chart Release(s) event handling

sonicaj:

1) If an app was abusive in the sense that it consistently crashed or for other various reasons was emitting constant events, middleware was not able to cope with the flow of the events as `chart.release.query` is an expensive call and lots of events happening ended up slowing down middleware. With the current changes this has been addressed and optimized to only retrieve what we want which is basically monitoring updated status of the application.
2) If an app had probes defined, when the probe considered the pod/deployment to be ready, kubernetes would not be emitting an event in this case which resulted in UI constantly saying `DEPLOYING` for the said application. The same is true when there are multiple init containers defined and the startup can take some time but we get an asynchronous event from kubernetes after container start where it does not really wait for the container / pod to be up, this results in the same `DEPLOYING` problem. This PR handles this by polling kubernetes API at intervals until the `DEPLOYING` status changes. We would poll in intervals `2, 4, 8, 16, 32, 64, 128, 256, 512, 1024` - so after 2 seconds, we would try 4 then 8 then 16 and so on. The last try would happen at 1024 seconds since the event ( the large gap is to allow large images to be pulled ).
3) Improves performance querying for k8s resources in a specified namespace.
4) Optionally return each resource individual events.
5) Fixes a potential race condition where we might try to access cached chart release when it has been deleted already.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/events.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/events.py
@@ -1,5 +1,6 @@
 import asyncio
 import collections
+import copy
 import functools
 
 from middlewared.schema import Dict, List, Str, returns
@@ -8,6 +9,7 @@ from middlewared.service import accepts, private, Service
 from .utils import get_chart_release_from_namespace, get_namespace, is_ix_namespace
 
 
+EVENT_LOCKS = collections.defaultdict(asyncio.Lock)
 LOCKS = collections.defaultdict(asyncio.Lock)
 
 
@@ -50,12 +52,13 @@ class ChartReleaseService(Service):
     async def refresh_events_state(self, chart_release_name=None):
         filters = [['id', '=', chart_release_name]] if chart_release_name else []
         for chart_release in await self.middleware.call('chart.release.query', filters):
-            ChartReleaseService.CHART_RELEASES[chart_release['name']] = {
-                'lock': asyncio.Lock(),
-                'data': chart_release,
-                'queue': self.get_queue(),
-                'poll': False,
-            }
+            async with LOCKS[chart_release['name']]:
+                ChartReleaseService.CHART_RELEASES[chart_release['name']] = {
+                    'lock': asyncio.Lock(),
+                    'data': chart_release,
+                    'queue': copy.deepcopy(self.get_queue()),
+                    'poll': False,
+                }
 
     @private
     @functools.cache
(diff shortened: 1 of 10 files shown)


src/middlewared/middlewared/plugins/chart_releases_linux/events.py
```diff
@@ -49,30 +52,74 @@ async def events(self, release_name):
...
-            ChartReleaseService.CHART_RELEASES[chart_release['name']] = chart_release
+            async with LOCKS[chart_release['name']]:
+                ChartReleaseService.CHART_RELEASES[chart_release['name']] = {
+                    'data': chart_release,
+                    'poll': False,
+                }
+
+    @private
+    @functools.cache
+    def get_queue(self):
```
themylogin, line 63: Shouldn't we retry every 1024 seconds indefinitely?

src/middlewared/middlewared/plugins/kubernetes_linux/deployments.py
```diff
@@ -16,12 +18,20 @@ class Config:
...
     async def query(self, filters, options):
         async with api_client() as (api, context):
-            deployments = [d.to_dict() for d in (await context['apps_api'].list_deployment_for_all_namespaces()).items]
-            events = await self.middleware.call(
-                'kubernetes.get_events_of_resource_type', 'Deployment', [d['metadata']['uid'] for d in deployments]
-            )
-            for deployment in deployments:
-                deployment['events'] = events[deployment['metadata']['uid']]
+            if len(filters) == 1 and len(filters[0]) == 3 and all(
+                a == d for a, d in zip(filters[0][:2], ['metadata.namespace', '='])
```
themylogin, line 22: `filters[0][:2] == ['metadata.namespace', '=']`

themylogin (changes requested):

src/middlewared/middlewared/plugins/kubernetes_linux/deployments.py
sonicaj: That does not cover the case when the filter is a tuple

sonicaj: I should typecast it :)

src/middlewared/middlewared/plugins/chart_releases_linux/events.py
```diff
@@ -49,30 +53,74 @@ async def events(self, release_name):
...
+        iterable = []
+        cur = 2
+        while cur <= self.MAX_SECONDS:
+            iterable.append(cur)
+            cur *= 2
+        return iterable
```
themylogin, line 70: You can make this a generator that yields 2, 4, ..., 1024, 1024, 1024... and then just iterate in `poll_chart_release_status`. Additionally, you can put that to `middlewared.utils.itertools` as other packages might need exponential back-off too.

themylogin (approved):

themylogin (approved):

PR #7215, opened 2021-07-27
NAS-111542 / 21.08 / fix traceback in network.py

yocalebo:


rick-mesta (approved):

src/middlewared/middlewared/plugins/network.py
```diff
@@ -655,9 +655,9 @@ def iface_extend(self, iface_state, configs, ha_hardware, fake=False):
             if lag:
                 lag = lag[0]
                 if lag['protocol'] in ('lacp', 'loadbalance'):
-                    iface.update({'xmit_hash_policy': lag.get('xmit_hash_policy', 'layer2+3').upper()})
+                    iface.update({'xmit_hash_policy': lag.get('xmit_hash_policy') or 'LAYER2+3'})
```
sonicaj, line 658: Would it always be upper case if we do have that key in `lag` dictionary ?

yocalebo: Nope, this is the joy of how our wonderful networking API was written. I added `upper()` to match `lag['protocol']` behavior.

sonicaj (commented):
This would fail as dictionary does not has an `upper` attribute ?

sonicaj (approved):

PR #7217, opened 2021-07-27
NAS-111546 / 21.08 / initialize eventsd webhook file correctly

yocalebo:


src/middlewared/middlewared/plugins/gluster_linux/eventsd.py
```diff
@@ -187,6 +187,10 @@ def init(self, job):
                 webhook_file.parent.mkdir(exist_ok=True)
                 # create the webhooks file if it doesnt exist
                 webhook_file.touch(exist_ok=True)
+                # make sure the file is initialized to empty
+                # json formatted file since 3rd party binary
+                # expects it to be
+                webhook_file.write_text('{}') if not webhook_file.read_text() else None
```
themylogin, line 193: Do you like these Yoda-style expressions?

themylogin (approved):

src/middlewared/middlewared/plugins/gluster_linux/eventsd.py
yocalebo: Yeah, sometimes I like them for simple cases like this. They tend to quickly get out of hand if the checks are more involved, however. 😄

PR #7223, opened 2021-07-28
NAS-111564 / 21.08 / Correct vm.query event payload to match vm_entry schema

sonicaj:

1) Correctly refer to the CRUD entry schema for query events.
2) Correct vm.query event payload structure.


src/middlewared/middlewared/service.py
```diff
@@ -628,7 +628,7 @@ def __init__(self, middleware):
                 f'{self._config.namespace}.query',
                 f'Sent on {self._config.namespace} changes.',
                 self._config.private,
-                returns=Ref(self.ENTRY.name),
+                returns=Ref(self.ENTRY.newname if isinstance(self.ENTRY, Patch) else self.ENTRY.name),
```
themylogin, line 631: Should we just change `Patch` so it has proper `name`?

themylogin (approved):

PR #7225, opened 2021-07-28
NAS-111570 / 21.08 / Wait when adding/removing iptables rules

sonicaj:
1) Wait when adding/deleting iptables rules as another app might be holding the xtables-lock
2) Do not continue adding iptables rules if we are unable to add even one
3) Remove drop iptables rule first and then proceed to the accept ones
4) Remove iptables rules before stopping k3s as if user is using wildcard ip address for k8s, k8s would be selecting a node ip on it's own and after k8s has stopped we can't determine which ip was chosen by k8s.


themylogin (approved):

yocalebo:
I also am adding/removing iptables rules in `failover_/iptables.py` which I've protected with a `@job` decorator. I do this for other reasons related to failover, but can you see if I also need to be `--wait` in there too?

sonicaj:
@yocalebo we should be doing that there as well as the other app in this case is kube-router and it could potentially be holding the lock. I'll make the change, thanks for pointing it out.

PR #7232, opened 2021-07-29
NAS-111573 / 12.0 / fix race between `dd` and `gpart destroy`

yocalebo:


src/middlewared/middlewared/plugins/disk_/wipe.py
```diff
@@ -68,6 +68,14 @@ async def wipe(self, job, dev, mode, sync, options=None):
             for part in await self.middleware.call('disk.list_partitions', dev):
                 await self.wipe_quick(part['name'], part['size'])
 
+        # we have to force GEOM to retaste the disks since if this is a QUICK
+        # format and the disks had previous partition tables on them, we run
+        # like 4 instances of `dd`. Using regular GEOM utilities, they will
+        # actually wait for the related thread to become idle before returning
+        # from the syscall. `dd` doesn't do this so we force GEOM to retaste
+        # all the disks here before calling "gpart destroy"
+        await run(['sysct', 'kern.geom.confxml'], check=False)
```
sonicaj, line 77: Is it possible to retaste the disk in question only ?

sonicaj: Also typo - it should be `sysctl`

themylogin (approved):

src/middlewared/middlewared/plugins/disk_/wipe.py
yocalebo: Fixed typo. calling `kern.geom.conftxt` is cheap enough.

sonicaj (approved):

src/middlewared/middlewared/plugins/disk_/wipe.py
```diff
@@ -68,6 +68,14 @@ async def wipe(self, job, dev, mode, sync, options=None):
             for part in await self.middleware.call('disk.list_partitions', dev):
                 await self.wipe_quick(part['name'], part['size'])
 
+        # we have to force GEOM to retaste the disks since if this is a QUICK
+        # format and the disks had previous partition tables on them, we run
+        # like 4 instances of `dd`. Using regular GEOM utilities, they will
+        # actually wait for the related thread to become idle before returning
+        # from the syscall. `dd` doesn't do this so we force GEOM to retaste
+        # all the disks here before calling "gpart destroy"
+        await run(['sysctl', 'kern.geom.conftxt'], check=False)
```
william-gr, line 77: Hmm, it should be as simple enough as opening the disk as read and then closing it, no?
At least thats how i thought we did it at some point in the past. @amotin

amotin: Opening for write and closing we were using to trigger the retaste.  Read should not wait for anything.  But thinking now, closing device after write should send the retaste event and respectively should wait for event thread getting idle before returning to user space.  So It is not completely clear why would we need anything more, but may be there are some factors.  Reading that sysctl is the easiest way I found to wait for even thread getting idle.

PR #7233, opened 2021-07-29
NAS-111585 / 21.08 / Only override aclinherit if acltype explicitly nfsv4

anodos325:
For cases where acltype is omitted or set to 'INHERIT' we
should just let aclinherit stay at its default settings.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -3176,7 +3176,7 @@ class PoolDatasetService(CRUDService):
 
         parent_ds = parent_ds[0]
         mountpoint = os.path.join('/mnt', data['name'])
-        if data.get('acltype') == 'INHERIT' and len(data['name'].split('/')) == 2:
+        if data.get('acltype', 'INHERIT') == 'INHERIT' and len(data['name'].split('/')) == 2:
             data['acltype'] = 'POSIX'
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3187,7 +3187,7 @@ async def do_create(self, app, data):
             data['acltype'] = 'NFSV4'
             data['aclmode'] = 'RESTRICTED'
 
-        if data['type'] == 'FILESYSTEM':
+        if data['type'] == 'FILESYSTEM' and data.get('acltype', 'INHERIT') != 'INHERIT':
```
rick-mesta, line 3190: I may be missing something, but I can only see the code working for when `acltype` is omitted. If it's set to `INHERIT`, won't the check fail and _*not*_ go into the conditional ? 🤔

anodos325: Right, in this case we're skipping check if acltype is omitted (we're inheriting from parent implicitly or not changing it). Behavior is as follows:
* SKIP if ZVOL
* SKIP if acltype omitted from payload
* SKIP if acltype is set to INHERIT

rick-mesta: 👍🏼

rick-mesta (approved):

PR #7237, opened 2021-07-29
NAS-111269 / 21.08 / Rework directory services user cache backend

anodos325:
Switch to using tdb files for the directory services cache.
This simplifies some interactions with the cache (retrieval
while building it out) and also makes persistence simpler.

Rework backend behavior so that if API user specifies an
explict id or name, and only has one filter doing this,
then on cache lookup failure, we directly query nss for
user or group, create a synthetic result, insert into cache
and return it to API user. This requires having deterministic
method of generating fake "id" keys for directory services
users. Algorithm used is to take low-range of idmap backend,
add the rid value of the user/group to it, and add a default offset
of 100,000. This should avoid collisions with ids assigned from
sqlite (in a sane environment, 100K users or groups won't have
been added to the server). RID values in AD increment as well and
so this algorithm ensures that no `id` keys are ever re-used.

### DIFF
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -1484,7 +1484,9 @@ class ActiveDirectoryService(ConfigService):
 
     @private
     async def get_cache(self):
-        return await self.middleware.call('dscache.entries', self._config.namespace)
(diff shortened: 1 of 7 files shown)


anodos325:
Looks like this needs a slight bit more time in the oven. WIll fix flake8 issues etc tomorrow. Throwing it out so that I can get any feedback that you want to dish out :)

rick-mesta (approved):

PR #7240, opened 2021-07-30
NAS-111594 / 13.0 / improve the process for formatting disks

yocalebo:
This dramatically improves the process for "wiping" a disk. Before my changes, we were doing the following:
1. `dd`'ing first 32MB of disk
2. `dd`'ing any partition table offsets (if they were detected)
3. `dd`'ing the last 32MB of disk
4. `gpart destroy -f` the disk (after `dd`)
5. `gpart create -s gpt`
6. `gpart destroy -f` again...
(because of our process we were opening the underlying disk(s) many times unnecessarily)

To add insult to injury, we are doing this all using `subprocess` which is expensive and inefficient. Furthermore, there is a race between when `dd` is done and the GEOM thread goes idle before we call `gpart destroy`. See https://github.com/truenas/middleware/pull/7232 for details and work-around for the `truenas/12.0-stable` branch.

To resolve all of these issues in 1 fell swoop, I've added the ability to wipe a disk using python's native low-level `os` module. Overwriting the first and last 32MB of a disk is more than enough to remove all partition table information. There is no reason to run `gpart` related commands after this has been done. Lastly, this gets run in a native thread so, in theory, the speed at which the data gets written to disk will be as fast as the underlying disk. So, in summary, with the changes this fixes:
1. the elevated usage of `subprocess`
2. the unnecessary `dd`'ing and subsequent `gpart destroy/create` calls
3. the potential for race conditions
4. open the underlying disk only once

(added benefit of being more efficient and getting rid of regular expression usage as well)

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/wipe.py
+++ b/src/middlewared/middlewared/plugins/disk_/wipe.py
@@ -1,47 +1,76 @@
-import asyncio
 import os
-import re
-import signal
-import subprocess
+import array
+import struct
+import fcntl
 
 from middlewared.schema import accepts, Bool, Ref, Str
 from middlewared.service import job, private, Service
-from middlewared.utils import osc, Popen, run
 
 
-if osc.IS_LINUX:
-    RE_DD = re.compile(r'^(\d+).*bytes.*copied.*, ([\d\.]+)\s*(GB|MB|KB|B)/s')
-else:
-    RE_DD = re.compile(r'^(\d+) bytes transferred .*\((\d+) bytes')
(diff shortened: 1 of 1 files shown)


rick-mesta (approved):

rick-mesta:
I'd still like to see reviews from @themylogin and @sonicaj before we merge this bad boy 🙏🏼

themylogin (approved):
Wow!

sonicaj (approved):

yocalebo:
backport

PR #7241, opened 2021-08-01
NAS-111371 / 21.08 / Add endpoint to generate a SSH Connection

sonicaj:
This PR adds an endpoint to generate a SSH connection which the UI can consume to make a single call which will generate a SSH Key pair ( if required ) and setup a SSH Credentials object based on the connection type selected by the consumer. It is helpful for cases so UI does not has to manually rollback generated key pair if a request fails.


themylogin (changes requested):
LGTM, except that the rest of the file uses double quotes, so let's be consistent with that. Another proposal is to create keychain_/ssh.py and move this method there is order to stop this file from growing.

PR #7242, opened 2021-08-02
NAS-111420 / 21.08 / Make adding default video device optional for VMs

sonicaj:
This PR adds changes to allow adding default video device optional for VMs. Motivation for adding this device always earlier was for headless installations like ubuntu server would still require a video device to work, however when a user has passed GPU(s), some guests can be picky about the video device they would like to consume which results in GPU passthrough working for the guest but the guest not consuming the GPU for it's primary display.


themylogin (changes requested):
LGTM, but please also document what this field does in the method docstring.

themylogin (approved):

PR #7246, opened 2021-08-02
NAS-111567 / 21.10 / Improvements to middleware schema/events

sonicaj:

1) Normalizes `Patch` class so that `name` points to the new name of the schema consumer wants.
2) Send a `REMOVED` event when resources are cleared/removed and deprecate sending `CHANGED` events on these instances.

### DIFF
--- a/src/middlewared/middlewared/plugins/alert.py
+++ b/src/middlewared/middlewared/plugins/alert.py
@@ -439,7 +439,7 @@ class AlertService(Service):
 
     def _send_alert_deleted_event(self, alert):
         self.middleware.send_event("alert.list", "CHANGED", id=alert.uuid, cleared=True)
-        self.middleware.send_event("alert.list", "REMOVED", id=alert.uuid, cleared=True)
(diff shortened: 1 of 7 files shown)


sonicaj:
( will be merged after 21.08 )

src/middlewared/middlewared/plugins/alert.py
```diff
@@ -439,6 +439,7 @@ async def _send_alert_changed_event(self, alert):
 
     def _send_alert_deleted_event(self, alert):
         self.middleware.send_event("alert.list", "CHANGED", id=alert.uuid, cleared=True)
+        self.middleware.send_event("alert.list", "REMOVED", id=alert.uuid, cleared=True)
```
themylogin, line 442: Do we need `cleared=True` here?

sonicaj: I think no, `REMOVED` should be enough to clarify that.

themylogin (approved):

PR #7250, opened 2021-08-03
NAS-111655 / 21.08 / Fix check for passdb backend type

anodos325:
Since we're now explicitly specifying a non-default passdb
location we need to adjust the check for ldapsam slightly.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb_/passdb.py
+++ b/src/middlewared/middlewared/plugins/smb_/passdb.py
@@ -25,13 +25,15 @@ class SMBService(Service):
         if not os.path.exists(f'{private_dir}/passdb.tdb'):
             return pdbentries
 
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/smb_/passdb.py
```diff
@@ -75,7 +77,8 @@ async def update_passdb_user(self, username, passdb_backend=None):
                                                         'passdb backend',
                                                         'global')
 
-        if passdb_backend != 'tdbsam':
+        passdb_backend = await self.middleware.call('smb.getparm', 'passdb backend', 'global')
```
themylogin, line 80: Why this addition here?

anodos325: Because copy-paste error :)

rick-mesta (approved):

themylogin (approved):

PR #7255, opened 2021-08-03
NAS-111590 / 21.08 / Validate route_v4_gateway specified is accurate

sonicaj:
This commit adds changes to make sure that the gateway specified is accurate and is actually accessible by the system, as we have had cases with users trying to use 8.8.8.8 for the gateway which is obviously going to fail.


rick-mesta (approved):

themylogin (approved):

william-gr:
Are we waiting for something here?

PR #7262, opened 2021-08-05
NAS-111685 / 12.0 / make get_ses_enclosures scale on large systems

yocalebo:
On an M60 HA system with 417 disks, we're seeing some scalability problems. We're `subprocess`'ing out `D * E` (D being the number of disks and `E` being the number of enclosures on `disk.sync_all`) when we can instead run a single `subprocess` to get the output. So on a 417 disk system with 4 expansion shelves, we're subprocessing out 1668 times just in this method alone.

Talk is cheap so here are the numbers directly from the test system.

Before my changes, `get_ses_enclosures` profiled at:
```
1150 function calls in 0.604 seconds
```

After my changes it profiled at:
```
15399 function calls in 0.168 seconds
```

So while many more function calls are made with this change, the time spent processing the data _overall_ is significantly lower. On average it takes ~72% less time to run that method. It's important this method be as fast as possible since our `disk.*` api calls it extensively.


themylogin (approved):

rick-mesta (approved):

themylogin:
backport

PR #7265, opened 2021-08-05
NAS-111690 / 12.0 / call enclosure.query once in disk.sync_all

yocalebo:
1. Dont immediately sleep in `disk.sync_all` for 1 second without checking to see if `devd` is connected. Instead see if it's connected first and then sleep appropriately.
2. Only call `enclosure.query` once so that we don't call it N times (N being the number of disks on the system)

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure.py
@@ -148,8 +148,11 @@ class EnclosureService(CRUDService):
 
         return await self._get_instance(id)
 
-    def _get_slot(self, slot_filter, enclosure_query=None):
-        for enclosure in self.middleware.call_sync("enclosure.query", enclosure_query or []):
+    def _get_slot(self, slot_filter, enclosure_query=None, enclosure_info=None):
+        if enclosure_info is None:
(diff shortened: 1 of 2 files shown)


src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure.py
```diff
@@ -160,8 +163,8 @@ def _get_slot(self, slot_filter, enclosure_query=None):
 
         raise MatchNotFound()
 
-    def _get_slot_for_disk(self, disk):
-        return self._get_slot(lambda element: element["data"]["Device"] == disk)
+    def _get_slot_for_disk(self, disk, enclosure_info=None):
+        return self._get_slot(lambda element: element["data"]["Device"] == disk, enclosure_info)
```
themylogin, line 167: Shouldn't it be `enclosure_info=enclosure_info`?

themylogin (changes requested):

src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure.py
yocalebo: Yeah, I fixed

themylogin (approved):

themylogin:
backport

PR #7266, opened 2021-08-05
NAS-111693 / 12.0 / fix enclosure.sync_zpool

yocalebo:
1. No reason to call `truenas.get_chassis_hardware` for every zpool on system so call it outside for loop.
2. `disk.label_to_disk` is abysmally slow when it has to run `geom.scan()`, instead use new `label_to_dev_disk_cache` method to essentially make this portion of the method instant.
3. `_get_ses_slot_for_disk` calls `enclosure.query` but we already enumerate the enclosures at the top of this method in the `encs` variable. Use the built-in `find_device_slot` instead to prevent unnecessary `enclosure.query` calls for every disk on every zpool.

Before changes `enclosure.sync_zpool` was taking ~5mins+ on M60 with 417 disks.
After changes `enclosure.sync_zpool` is taking ~21seconds on M60 with 417 disks.

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure.py
@@ -1,9 +1,3 @@
-# Copyright (c) 2019 iXsystems, Inc.
-# All rights reserved.
-# This file is a part of TrueNAS
-# and may not be copied and/or distributed
-# without the express permission of iXsystems.
-
 from collections import OrderedDict
 import logging
 import os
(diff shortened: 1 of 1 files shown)


rick-mesta (approved):
LGTM; I'll wait on @themylogin 's review before merge 👍🏼

rick-mesta:
@themylogin: ping

themylogin (approved):

rick-mesta:
@themylogin: Thx Vlad 👍🏼

themylogin:
backport

PR #7278, opened 2021-08-06
NAS-111710 / 21.08 / Fix groupmap tests

anodos325:
The groupmap validation portion of the group API tests needed to be updated for internal API changes. There were also some edge cases where we had backgrounded the groupmap synchronization during group add (tests were immediately testing while job was in progress). This PR un-backgrounds, as well as fixes a case where we weren't modifying an existing groupmap on nt name change.


ericbsd (approved):
LGTM

PR #7282, opened 2021-08-07
NAS-111720 / 12.0 / fix numerous issues with disk.sync_all

yocalebo:
`disk.sync_all`, in my humble opinion, needs to be scrapped and rewritten. It's almost impossible to follow (at least for me) and it's terribly inefficient.

Problems discovered:
1. for each disk we called `geom.scan()` 3 times (maybe more)
2. after `geom.scan` we followed up that with a bunch of more `geom.*` module calls.
3. we subprocess'ed out for the enclosures for every disk
4. we then iterated over every enclosure's slot for every disk
5. we call `disk.query` multiple times per disk
6. we call `datastore.query storage.disk` multiple times per disk
7. `device.identifier_to_device` accepted a `disks` keyword, which as far as I can tell, was supposed to be used to "speed" that method up and yet, the keyword argument was completely ignored and overwritten when you called that method
8. `device.device_to_identifier` accepts a `disks` keyword as well, but it was never working because we ran `disk.get('name')` instead of `disk.get(name)` (Notice the single-quotes around the `name` variable)

I really want to scrap it and rewrite it but alas time rules indiscriminately so here's what I've done to keep changes as minimal as possible.

1. stop calling geom.scan and geom* related api calls more than once for every disk. Instead call it once outside the for loop and pass it around to the various methods appropriately.
2. make the `device.identifier_to_device` accept said geom xml instantiation from a caller
3. fix `device.device_to_identifier` method to properly `.get(name)` so it's not a NO-OP
4. Nested deep down in `disk.identifier_to_device` there was a chance to crash on a `'NoneType' object has no attribute 'split'` so fix that while I'm here
5. generate a very manageable dictionary of the disks to slots outside the for loop and then once we're done updating the database for all the disks, check the enclosure slot information all at once.

Before my changes, on an M60 HA with 417 disks (and no other changes) `disk.sync_all` took ~6.5hrs....

After my changes, on an M60 HA with 417 disks `disk.sync_all` takes ~30 seconds

While I'm still not happy with 30 seconds, it will have to do for now.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/sync.py
+++ b/src/middlewared/middlewared/plugins/disk_/sync.py
@@ -83,7 +83,8 @@ class DiskService(Service, ServiceChangeMixin):
         sys_disks = await self.middleware.call('device.get_disks')
         geom_xml = geom.class_by_name('DISK').xml
 
-        if len(sys_disks) <= 25:
+        number_of_disks = len(sys_disks)
+        if 0 > number_of_disks <= 25:
             # output logging information to middlewared.log in case we sync disks
             # when not all the disks have been resolved
             log_info = {
@@ -92,6 +93,8 @@ class DiskService(Service, ServiceChangeMixin):
                 } for ok, ov in sys_disks.items()
             }
             self.logger.info('Found disks: %r', log_info)
+        else:
+            self.logger.info('Found %d disks', number_of_disks)
 
         seen_disks = {}
         serials = []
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/disk_/sync.py
```diff
@@ -83,30 +81,31 @@ async def sync_all(self, job):
...
+
+        if len(sys_disks) <= 25:
+            # output logging information to middlewared.log in case we sync disks
+            # when not all the disks have been resolved
+            log_info = {
+                ok: {
+                    ik: iv for ik, iv in ov.items() if ik in ('name', 'ident', 'lunid', 'serial')
+                } for ok, ov in sys_disks.items()
+            }
+            self.logger.info('Found disks: %r', log_info)
```
themylogin, line 94: Can you log `Found 400 disks` in `else` branch please?

src/middlewared/middlewared/plugins/disk_/sync.py
```diff
@@ -186,7 +183,16 @@ async def sync_all(self, job):
                     await self.middleware.call('datastore.insert', 'storage.disk', disk)
                     changed = True
 
-                await self.middleware.call('enclosure.sync_disk', disk['disk_identifier'], encs)
+        # now we need to check the enclosure mapping db info for each disk
```
themylogin, line 186: Should we better have `sync_disks` method in enclosure plugin which will handle an array of disks in one place?

themylogin (changes requested):

themylogin (approved):
Thank you!

rick-mesta (approved):

PR #7287, opened 2021-08-08
NAS-111744 / 21.08 / Fix SSHd IPV6 link local ListenAddress

bmhughes:
[Jira link](https://jira.ixsystems.com/browse/NAS-109095) - not sure why it was closed as this issue still persists

If SSH is restricted to listen on specific interface(s) and those interfaces are configured for IPv6 then for a IPv6 link local address the interface name must also be appended to the address.


yocalebo:
@bmhughes Thanks for your contribution! I can confirm that without your change, binding SSH to an interface with an IPv6 link-local address, it will fail without the interface name.

yocalebo (approved):

PR #7296, opened 2021-08-09
NAS-111763 / 21.08 / Fix homedir copy on SCALE

anodos325:
We were copying the old home directory (rather than just contents) into the new home directory.


ericbsd (approved):
LGTM

PR #7298, opened 2021-08-09
Cleanly reset AD parameters in activedirectory.leave

anodos325:
Avoid calling activedirectory.stop() in activedirectory.leave().
Have AD test wait for the activedirectory.stop jobs to complete.
Avoid importing dcerpc bindings.

### DIFF
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -1566,7 +1566,7 @@ class WBStatusThread(threading.Thread):
         self.middleware = kwargs.get('middleware')
         self.logger = self.middleware.logger
         self.finished = threading.Event()
-        self.state = 1027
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -1563,7 +1566,7 @@ def __init__(self, **kwargs):
         self.middleware = kwargs.get('middleware')
         self.logger = self.middleware.logger
         self.finished = threading.Event()
-        self.state = MSG_WINBIND_ONLINE
+        self.state = 1027
```
rick-mesta, line 1569: nit: would it make sense to append `# MSG_WINBIND_ONLINE` here, as you did below on `directoryservices.py` line 32 ?

rick-mesta (approved):
lgtm

PR #7302, opened 2021-08-10
NAS-111772 / 21.08 / Fix multiple issues with krb nfs4 config in SCALE

anodos325:
- Only enable krb5 security when our keytab has nfs spn.

- Convert tests to examine /etc/ganesha/ganesha.conf

- Convert activedirectory.add_nfs_spn to a job since experimentation
  indicated the call may take longer than 60 seconds to complete.

- Convert active directory configuration changes in NFS plugin to
  use the direct_update method (clustering agnostic) and sync with
  registry tdb.

- Update regression test to create a ZFS dataset and NFS share
  because these are now requirements for generating NFS config.

### DIFF
--- a/src/middlewared/middlewared/etc_files/ganesha/ganesha.conf.mako
+++ b/src/middlewared/middlewared/etc_files/ganesha/ganesha.conf.mako
@@ -7,7 +7,7 @@
     if not shares:
         raise FileShouldNotExist()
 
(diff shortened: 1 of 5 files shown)


anodos325:
Just realized I need to add step to test to remove the dataset we're now creating.

yocalebo (approved):

PR #7305, opened 2021-08-11
NAS-111769 / 12.0 / fix disk identification with vmware nvme disks

yocalebo:
When CORE is running as a VM in VMware and NVMe disks are provided to the VM, the disks have more than 1 `_` delimiter which is causing us to raise a `ValueError: too many values to unpack (expected 2)`. This fixes that particular scenario, and hopefully more, if we get a disk that has more than 1 `_` in the string passed to us.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/identify_freebsd.py
+++ b/src/middlewared/middlewared/plugins/disk_/identify_freebsd.py
@@ -108,7 +108,20 @@ class DiskService(Service, DiskIdentifyBase):
             else:
                 xml = geom.class_by_name('DISK').xml
 
-            _ident, _lunid = value.split('_')
+            info = value.split('_')
+            info_len = len(info)
+            if info_len < 2:
+                # nothing to do return
+                return
+            elif info_len == 2:
+                _ident = info[0]
+                _lunid = info[1]
+            else:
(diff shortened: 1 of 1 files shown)


rick-mesta (approved):

src/middlewared/middlewared/plugins/disk_/identify_freebsd.py
```diff
@@ -108,7 +108,20 @@ def identifier_to_device(self, ident, geom_scan=True, geom_xml=None):
...
+                # nothing to do return
+                return
+            elif info_len == 2:
+                _ident = info[0]
+                _lunid = info[1]
+            else:
+                # vmware nvme disks look like `VMware NVME_0000_a9d1a9a7feaf1d66000c296f092d9204`
+                # so we need to account for it
+                _lunid = info[-1]
+                _ident = info[:-info_len].rstrip('_')
```
sonicaj, line 123: @yocalebo i am not sure i follow the change, it errors out ?
```
>> value = 'VMware NVME_0000_a9d1a9a7feaf1d66000c296f092d9204'
>> info = value.split('_')
>> info_len = len(info)
>> info[:-info_len].rstrip('_')
  File "<input>", line 1, in <module>
AttributeError: 'list' object has no attribute 'rstrip'
```

yocalebo: A typo which I'll fix

PR #7307, opened 2021-08-11
NAS-110003 / 21.08 / Fix nsswitch.conf usage in SCALE

anodos325:
Only generate nsswitch.conf and always include ldap and winbind
entries. This file will only be read once the first time pwd or
group modules are used and used for the life of the middlewared
process.

Move nslcd config to ldap plugin and remove extra calls to generate
nss.

Fix a couple of bugs in ldap plugin as well:
- wrong string used for checking for start_tls usage. Switch to constant
- synchronize registry regardless of whether samba schema is in payload.

### DIFF
--- a/src/middlewared/middlewared/etc_files/nsswitch.conf.mako
+++ b/src/middlewared/middlewared/etc_files/nsswitch.conf.mako
@@ -1,42 +1,11 @@
 #
 # nsswitch.conf(5) - name service switch configuration file
-# $FreeBSD$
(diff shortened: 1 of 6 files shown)


rick-mesta (approved):
LGTM: I'll let @ericbsd comment on the tests themselves 😁

yocalebo (approved):

ericbsd (approved):
LGTM

PR #7308, opened 2021-08-11
NAS-111708 / 21.08 / Add endpoint to destroy specified snapshots of a dataset

sonicaj:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3607,6 +3607,49 @@ async def do_delete(self, id, options):
...
+                    'snapshot_spec',
+                    Str('start'),
+                    Str('end'),
+                )]
+            ),
+        ),
+    )
+    @returns(List('deleted_snapshots', items=[Str('deleted_snapshot')]))
+    @job(lock=lambda args: f'destroy_snapshots_{args[0]}')
+    async def destroy_snapshots(self, job, name, snapshots_spec):
```
themylogin, line 3627: I think we should either name it `destroy_snapshots_ranges` or allow specifying single snapshots too.

themylogin (approved):

PR #7319, opened 2021-08-12
NAS-111814 / 21.10 / Do not report coredumps generated by apps in containers

sonicaj:
This commit adds changes to not report coredumps generated by any specific app in a container as users are free to run whatever they would like to in containers and we don't officially support all the apps themselves.

### DIFF
--- a/src/middlewared/middlewared/alert/source/cores.py
+++ b/src/middlewared/middlewared/alert/source/cores.py
@@ -23,6 +23,7 @@ class CoreFilesArePresentAlertClass(AlertClass):
 
 class CoreFilesArePresentAlertSource(AlertSource):
     dumps = {}
+    dump_error_logged = []
     schedule = IntervalSchedule(timedelta(minutes=5))
 
     products = ("SCALE",)
@@ -30,7 +31,7 @@ class CoreFilesArePresentAlertSource(AlertSource):
     async def check(self):
         corefiles = []
         for coredump in filter(lambda c: c["corefile"] == "present", await self.middleware.call("system.coredumps")):
-            if coredump["exe"] == "/usr/sbin/syslog-ng":
+            if coredump["exe"] == "/usr/sbin/syslog-ng" or coredump["pid"] in self.dump_error_logged:
                 continue
 
             if coredump["pid"] not in self.dumps:
@@ -41,6 +42,7 @@ class CoreFilesArePresentAlertSource(AlertSource):
                     self.middleware.logger.debug(
                         "Unable to retrieve coredump information of %r process", coredump["pid"]
                     )
+                    self.dump_error_logged.append(coredump["pid"])
                     continue
 
                 self.dumps[coredump["pid"]] = {
--- a/src/middlewared/middlewared/alert/source/cores.py
+++ b/src/middlewared/middlewared/alert/source/cores.py
@@ -1,6 +1,12 @@
+import re
+
 from datetime import timedelta
 
 from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, AlertSource, Alert, IntervalSchedule
+from middlewared.utils import run
+
+
+SKIP_REGEX = re.compile(r'Unit:\s+containerd.service')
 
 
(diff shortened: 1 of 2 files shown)


yocalebo (changes requested):
We already skip `syslog-ng` core dumps in the core dump alert plugin by looking at the `exe` value. Can we not add this logic there?

src/middlewared/middlewared/plugins/system_/coredump.py
```diff
@@ -21,6 +24,15 @@ async def coredumps(self):
         for line in lines:
             exe = line[exe_pos:]
             time, pid, uid, gid, sig, corefile = line[:exe_pos].rsplit(maxsplit=5)
+            cp = await run("coredumpctl", "info", pid, check=False, encoding="utf-8", errors="ignore")
```
themylogin, line 27: Move this code to the alert please, Also please cache coredumpctl result to avoid running N coredumpctl subprocesses every minute.

themylogin (changes requested):

sonicaj:
@yocalebo no we cannot as `EXE` value does not give us context as to where the coredump happened

src/middlewared/middlewared/alert/source/cores.py
```diff
@@ -16,18 +22,37 @@ class CoreFilesArePresentAlertClass(AlertClass):
...
+        for coredump in filter(lambda c: c["corefile"] == "present", await self.middleware.call("system.coredumps")):
+            if coredump["exe"] == "/usr/sbin/syslog-ng":
+                continue
+
+            if coredump["pid"] not in self.dumps:
+                cp = await run(
+                    "coredumpctl", "info", str(coredump["pid"]), check=False, encoding="utf-8", errors="ignore"
+                )
+                if cp.returncode:
+                    self.middleware.logger.debug(
```
themylogin, line 41: This will log the same message every minute. We should probably cache the error state too to avoid calling `coredumpctl info` for failing pid every minute.

themylogin (changes requested):

src/middlewared/middlewared/alert/source/cores.py
```diff
@@ -16,18 +22,39 @@ class CoreFilesArePresentAlertClass(AlertClass):
...
         corefiles = []
-        for coredump in await self.middleware.call("system.coredumps"):
-            if coredump["corefile"] == "present":
-                if coredump["exe"] == "/usr/sbin/syslog-ng":
+        for coredump in filter(lambda c: c["corefile"] == "present", await self.middleware.call("system.coredumps")):
+            if coredump["exe"] == "/usr/sbin/syslog-ng":
+                continue
+
+            if coredump["pid"] not in self.dumps:
+                cp = await run(
```
themylogin, line 38: We need to skip running `coredumpctl info` if `coredump["pid"]` is in `self.dump_error_logged`, otherwise we will be running it every minute for each malformed coredump.

themylogin (changes requested):

src/middlewared/middlewared/alert/source/cores.py
```diff
@@ -16,18 +22,39 @@ class CoreFilesArePresentAlertClass(AlertClass):
...
+                    self.dump_error_logged.append(coredump["pid"])
                     continue
 
-                corefiles.append(f"{coredump['exe']} ({coredump['time']})")
+                self.dumps[coredump["pid"]] = {
+                    **coredump,
+                    "happened_in_container": bool(SKIP_REGEX.findall(cp.stdout))
+                }
+
+            if self.dumps[coredump["pid"]]["happened_in_container"]:
```
themylogin, line 53: `KeyError` here if `coredump["pid"]` is in `self.dump_error_logged`

themylogin (changes requested):

themylogin (approved):
Now LGTM 👍😁 Thank you!

rick-mesta:
ping @yocalebo: does @sonicaj's new deltas address your concerns, Caleb ?

yocalebo (approved):

PR #7320, opened 2021-08-12
NAS-111816 / 21.08 / Improve acltype retrieval based on path

anodos325:
Retrieve acltype through maximum of two getxattr calls. Use this
to inform how we read ACLs. Add proper returns for case where
ZFS acltype is off.


rick-mesta (approved):
LGTM

PR #7321, opened 2021-08-13
NAS-111819 / 12.0 / Fix taking vmware-aware manual snapshots

themylogin:
This was broken by acef89b7413859a9bbe70226139756be3551b784

### DIFF
--- a/src/middlewared/middlewared/plugins/vmware.py
+++ b/src/middlewared/middlewared/plugins/vmware.py
@@ -393,6 +393,11 @@ class VMWareService(CRUDService):
         return self.middleware.call_sync("vmware.query", [f])
 
     @private
+    def snapshot_begin(self, dataset, recursive):
+        qs = self._dataset_get_vms(dataset, recursive)
+        return self.snapshot_proceed(dataset, qs)
+
+    @private
     def snapshot_proceed(self, dataset, qs):
         # Generate a unique snapshot name that won't collide with anything that exists on the VMWare side.
         vmsnapname = str(uuid.uuid4())


sonicaj (approved):

src/middlewared/middlewared/plugins/vmware.py
```diff
@@ -392,6 +392,11 @@ def _dataset_get_vms(self, dataset, recursive):
             ]
         return self.middleware.call_sync("vmware.query", [f])
 
+    @private
+    def snapshot_begin(self, dataset, recursive):
+        qs = self._dataset_get_vms(dataset, recursive)
+        return self.snapshot_proceed(dataset, qs)
+
```
rick-mesta, line 399: Q: in the original commit, `snapshot_begin()` only called `_dataset_get_vms()`; now we _*also*_ trigger `snapshot_proceed()` -- I assume this was intentional or did the underlying functionality change, so that now `snapshot_proceed()` should be called in this context ?

themylogin: @rick-mesta actual `snapshot_begin` was `_dataset_get_vms` + all the body of `snapshot_proceed`, and `snapshot_proceed` did not exist. This type of diff looks confusing sometimes.

rick-mesta: Thx for the explanation @themylogin ... just wanted to make sure it wasn't an easy oversight 👍🏼

rick-mesta (approved):

PR #7322, opened 2021-08-13
NAS-111820 / 21.08 / Extend validation of ACL entry IDs

anodos325:
Entries with tags associated with file User, Group, or Other /Everyone
may not have a numeric id set. For backwards compatibility, the
following values are permitted for these special entries: -1, null/None.


rick-mesta (approved):
LGTM: @sonicaj / @themylogin / @yocalebo - would any of you guys please put another pair of eyes on this ? 🙏🏼

PR #7325, opened 2021-08-13
NAS-111837 / 21.08 / Make sure we start nfs when user requests it

sonicaj:
Right now if no nfs shares are configured and if user attempts to start NFS, it does not start as the configuration file is missing for ganesha. This leads to violation of POLA where consumer adding shares later thinks/assumes that NFS would be running but that's not the case and it needs to be started again after configuring a share.

### DIFF
--- a/src/middlewared/middlewared/etc_files/ganesha/ganesha.conf.mako
+++ b/src/middlewared/middlewared/etc_files/ganesha/ganesha.conf.mako
@@ -4,9 +4,6 @@
     config = middleware.call_sync("nfs.config")
 
     shares = middleware.call_sync("sharing.nfs.query", [["enabled", "=", True]])
-    if not shares:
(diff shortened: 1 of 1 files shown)


yocalebo:
This is not a POLA violation.

sonicaj:
User starts a service, we allow it to be started and consumer would expect it started or should've started but in reality it did not :p Why have different behavior then for iscsi ?

Also we allow the same operation to succeed for CORE, why is SCALE different ? Users already doing this in CORE would expect it to work in SCALE as well as otherwise we are changing behavior.

Either ways, we should either be consistent about this with each service and let the user know appropriately why the service can't be started or we shouldn't change behavior and let it start.

yocalebo (approved):
I have to redo all of this when I integrate nfs into gluster anyways.

PR #7328, opened 2021-08-13
NAS-111599 / 12.0 / Prevent crashes when AHCI SGPIO Enclosure 2.00 is present on unknown …

themylogin:
…product


yocalebo (commented):
I'm assuming this method is being called before dmidecode is cached in system plugin?

PR #7329, opened 2021-08-13
NAS-111838 / 21.10 / Move LDAP client code into explicit service plugin

anodos325:
This helps to standardize LDAP access and provides a somewhat
more abstract mechanism for interacting with remote LDAP servers
(if needed for AD plugin or other parts of middleware).

### DIFF
--- a/src/middlewared/middlewared/plugins/ldap.py
+++ b/src/middlewared/middlewared/plugins/ldap.py
@@ -322,7 +322,7 @@ class LDAPClient(Service):
             self._handle = None
             self.ldap_parameters = None
 
-    def _search(self, ldap_config, basedn='', scope=pyldap.SCOPE_SUBTREE, filter='', timeout=-1, sizelimit=0):
+    def _search(self, ldap_config, basedn='', scope=pyldap.SCOPE_SUBTREE, filter='', sizelimit=0):
         self._open(ldap_config)
         result = []
         serverctrls = None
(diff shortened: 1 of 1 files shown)


yocalebo (changes requested):
`_search` doesn't have `timeout` set which means this can block forever if the called method never returns.

yocalebo (approved):
My only concern is the main interpreter hanging during restart/shutdown if this thread is created and running before the timeout has elapsed.

Granted, it's probably not appropriate for you to solve that problem here since it requires adding some type of mechanism (threading event) that's shared between this ldap thread and parent thread. Anyways, other than that looks fine to me.

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -348,7 +348,7 @@ def _search(self, ldap_config, basedn='', scope=pyldap.SCOPE_SUBTREE, filter='',
                     attrsonly=0,
                     serverctrls=serverctrls,
                     clientctrls=clientctrls,
-                    timeout=timeout,
+                    timeout=ldap_config['options']['timeout'],
```
yocalebo, line 351: Actually, wondering if we should make sure this is some sane value (instead of like 9999999).

yocalebo (changes requested):

rick-mesta:
@anodos325: ping -- is there any of @yocalebo 's comments that may require additional changes ?

yocalebo (approved):

PR #7336, opened 2021-08-15
NAS-111852 / 21.10 / Improvements to retrieving catalog(s)

sonicaj:

1) Do not automatically retrieve catalogs data. We should only do that in any of the following cases:

    a) User explicitly went to Apps UI and requested to see the catalogs
    b) User has configured catalogs which are not already builtin
    c) User has configured kubernetes

2) Make sure that we always retrieve caching job if the catalog data is not readily available ( is not cached )


yocalebo (changes requested):
This doesn't take into account if the end-user has checked the deny outbound network activity option in network general plugin.

sonicaj:
@yocalebo good point, i'll make required changes. Thanks for pointing it out

yocalebo (approved):

PR #7339, opened 2021-08-16
NAS-111864 / 21.10 / generate proper empty scst.conf

yocalebo:
Because we allow to start services with "empty" configs, we're running into a situation where `scst` doesn't load the `scst_vdisk` kernel module. This module is needed, literally, for everything that we use iscsi for so this changes how we generate the config to always include an empty `HANDLER vdisk_blockio` and `HANDLER vdisk_fileio` entry in `scst.conf`. This is enough for the `scst` service to load the `scst_vdisk` kernel module.

### DIFF
--- a/src/middlewared/middlewared/etc_files/scst.conf.mako
+++ b/src/middlewared/middlewared/etc_files/scst.conf.mako
@@ -20,7 +20,7 @@
         associated_targets[a_tgt['target']].append(a_tgt)
 
     # Let's map extents to respective ios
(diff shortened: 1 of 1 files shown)


sonicaj (changes requested):
Can we avoid duplication please ?

yocalebo:
> Can we avoid duplication please ?

Removed unused key in `extents_io` and reused existing code.

sonicaj (approved):

PR #7342, opened 2021-08-17
NAS-111881 / 21.10 / catch NoSuchDevice in ethtool

yocalebo:
Using the `pyroute2.ethtool.Ethtool` module as well as using the `ethtool` CLI tool produced the same output so document the situation and catch the exception accordingly so we don't spam the logs unnecessarily.

### DIFF
--- a/src/middlewared/middlewared/plugins/interface/netif_linux/ethernet_settings.py
+++ b/src/middlewared/middlewared/plugins/interface/netif_linux/ethernet_settings.py
@@ -1,6 +1,6 @@
 from logging import getLogger
 from pyroute2.ethtool import Ethtool
-from pyroute2.ethtool.ioctl import NotSupportedError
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/interface/netif_linux/ethernet_settings.py
```diff
@@ -72,6 +72,19 @@ def __mediainfo__(self):
             # of media info (ethtool binary didnt report anything either)
             # so ignore these errors
             pass
+        except NoSuchDevice:
```
sonicaj, line 75: Do you think it's better to do this on L69 perhaps like `(NotSupportedError, NoSuchDevice)` ? This way we can ignore specific exceptions and not have their own `pass` / block for them ?

sonicaj (approved):
Thanks!

PR #7345, opened 2021-08-17
NAS-111808 / 21.10 / User serial port name instead of I/O address in grub configuration

sonicaj:
This PR adds changes to use serial port name instead of serial port i/o address as with the latter console redirection does not work and we are left with a blank console with ipmi sol.

### DIFF
--- a/src/freenas/usr/local/bin/truenas-grub.py
+++ b/src/freenas/usr/local/bin/truenas-grub.py
@@ -7,9 +7,19 @@ from middlewared.utils.db import query_config_table
 
 
 def get_serial_ports():
-    return {
-        v: k.replace('uart', 'ttyS') if osc.IS_FREEBSD else k for k, v in osc.system.serial_port_choices().items()
-    }
+    choices = osc.system.serial_port_choices()
+    if osc.IS_FREEBSD:
+        return {v: k.replace('uart', 'ttyS') for k, v in choices.items()}
+    else:
+        options = {}
+        # We also add i/o address to the dict here because when users are upgrading and their db still references to
+        # i/o address instead of port name, we won't be generating a valid configuration which is not desired
+        for k, v in choices.items():
+            options.update({
+                k: k,
+                v: k,
+            })
+        return options
 
 
 if __name__ == "__main__":
--- a/src/freenas/usr/local/bin/truenas-grub.py
+++ b/src/freenas/usr/local/bin/truenas-grub.py
@@ -7,19 +7,7 @@ from middlewared.utils.db import query_config_table
 
 
 def get_serial_ports():
-    choices = osc.system.serial_port_choices()
-    if osc.IS_FREEBSD:
-        return {v: k.replace('uart', 'ttyS') for k, v in choices.items()}
-    else:
-        options = {}
-        # We also add i/o address to the dict here because when users are upgrading and their db still references to
-        # i/o address instead of port name, we won't be generating a valid configuration which is not desired
-        for k, v in choices.items():
-            options.update({
-                k: k,
-                v: k,
-            })
-        return options
+    return {e['start']: e['name'].replace('uart', 'ttyS') for e in osc.system.serial_port_choices()}
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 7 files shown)


src/middlewared/middlewared/utils/osc/linux/system.py
```diff
@@ -1,15 +1,23 @@
...
 
 
 def get_cpu_model():
     with open('/proc/cpuinfo', 'r') as f:
         model = RE_CPU_MODEL.search(f.read())
         return model.group(1) if model else None
+
+
+def serial_port_choices():
+    cp = subprocess.Popen('dmesg | grep ttyS', stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
```
themylogin, line 21: Using dmesg to retrieve list of available serial port seems weird. Can we use `setserial` for example? https://www.cyberciti.biz/faq/find-out-linux-serial-ports-with-setserial/

themylogin (changes requested):
With this changes do we need I/O addresses on Linux at all?

src/middlewared/middlewared/alembic/versions/21.MM/2021-08-20_17-20_serial_ports.py
```diff
@@ -0,0 +1,37 @@
...
+User serial port(s) name instead of I/O address
+
+Revision ID: 725b7264abe6
+Revises: 29abd3dce632
+Create Date: 2021-20-08 17:20:42.818433+00:00
+"""
+from alembic import op
+import sqlalchemy as sa
+
+from middlewared.utils import osc
```
themylogin, line 11: Please do not import anything from middleware in migrations and just copy the code here.

src/middlewared/middlewared/alembic/versions/21.MM/2021-08-20_17-20_serial_ports.py
```diff
@@ -0,0 +1,37 @@
...
+    io_choices = {e['start']: e['name'] for e in osc.system.serial_port_choices()}
+    sys_config = [dict(row) for row in conn.execute("SELECT * FROM system_advanced").fetchall()]
+    if not sys_config:
+        return
+
+    sys_config = sys_config[0]
+    if sys_config['adv_serialport'] not in io_choices and not sys_config['adv_serialport'].startswith('ttyS'):
+        new_val = 'ttyS0'
+    else:
+        new_val = io_choices[sys_config['adv_serialport']]
```
themylogin, line 31: It will crash if `(sys_config['adv_serialport'] not in io_choices) == True` and `not sys_config['adv_serialport'].startswith('ttyS') == False`

src/freenas/usr/local/bin/truenas-grub.py
```diff
@@ -6,6 +6,22 @@
 from middlewared.utils.db import query_config_table
 
 
+def get_serial_ports():
+    choices = osc.system.serial_port_choices()
+    if osc.IS_FREEBSD:
+        return {e['start']: e['name'].replace('uart', 'ttyS') for e in choices}
+    else:
+        options = {}
+        # We also add i/o address to the dict here because when users are upgrading and their db still references to
```
themylogin, line 15: Aren't we migrating db so that it does not reference to I/O address?

themylogin (changes requested):

src/freenas/usr/local/bin/truenas-grub.py
```diff
@@ -7,9 +7,19 @@
...
+    choices = osc.system.serial_port_choices()
+    if osc.IS_FREEBSD:
+        return {v: k.replace('uart', 'ttyS') for k, v in choices.items()}
+    else:
+        options = {}
+        # We also add i/o address to the dict here because when users are upgrading and their db still references to
+        # i/o address instead of port name, we won't be generating a valid configuration which is not desired
+        for k, v in choices.items():
+            options.update({
+                k: k,
```
themylogin: You don't need `k: k` here because it is used like `port = get_serial_ports().get(advanced['serialport'], advanced['serialport'])`

themylogin (changes requested):

themylogin (approved):

PR #7346, opened 2021-08-18
DOCS-2484 / 12.0  Replaced ssh to OSX and BSD by smbclient command

ericbsd:


q5sys (approved):

tests/api2/activedirectory.py
```diff
@@ -8,29 +8,25 @@
...
+    PUT,
+    DELETE,
+    SSH_TEST,
+    cmd_test,
+    wait_on_job
+)
 
 try:
     from config import AD_DOMAIN, ADPASSWORD, ADUSERNAME, ADNameServer
+    ad_user = fr"AD01\{ADUSERNAME.lower()}"
```
yocalebo, line 24: There is no difference between these 2 variables from what I can tell. Also since these are global variables, capitalize them.

yocalebo (changes requested):

tests/api2/activedirectory.py
```diff
@@ -60,12 +56,10 @@
     "enable"
 ]
 
-MOUNTPOINT = f"/tmp/ad-{hostname}"
 dataset = f"{pool_name}/ad-bsd"
 dataset_url = dataset.replace('/', '%2F')
-SMB_NAME = "TestShare"
-SMB_PATH = f"/mnt/{dataset}"
-group = 'root' if scale else 'wheel'
+smb_name = "TestShare"
```
yocalebo, line 61: Same here for global variables. Best practice is to capitalize these.

tests/api2/activedirectory.py
ericbsd: I had to make one `fr"AD01\\{ADUSERNAME.lower()}"` or the smbclient command fails.

yocalebo: Sure but there are 2 variables that are the same? They're both using `fr'` syntax

ericbsd: Oh and fr"AD01\\{ADUSERNAME.lower()}" fails the for the API call

yocalebo: Oh, I see now the `cmd_ad_user` has double `\\` instead of just one.

yocalebo: So `ad_user` is used for changing the perms and `cmd_ad_user` is for running the `smbclient` command.

ericbsd: Yes but one contains `//`.

tests/api2/activedirectory.py
```diff
@@ -291,134 +277,94 @@ def test_23_creating_a_smb_share_on_smb_path(request):
...
-@bsd_host_cfg
-def test_35_moving_SMB_file(request):
+def test_29_create_a_file_and_put_on_the_active_directory_share(request):
     depends(request, ["ad_01", "ad_02", "ad_07", "ad_10"])
-    cmd = 'mv "%s/testfile" "%s/testfile2"' % (MOUNTPOINT, MOUNTPOINT)
-    results = SSH_TEST(cmd, BSD_USERNAME, BSD_PASSWORD, BSD_HOST)
+    cmd_test('touch testfile.txt')
+    print()
+    command = fr'smbclient //{ip}/{smb_name} -U {cmd_ad_user}%{ADPASSWORD}' \
+        r' -m NT1 -c "put testfile.txt testfile.txt"'
```
anodos325, line 329: Out of curiosity, is there a particular reason we're using SMB1 for this test? If not it's better to just leave default protocol in place, and explicitly test SMB1 vs SMB3 in other SMB protocol-specific tests e.g. test_425.

tests/api2/activedirectory.py
ericbsd: > So `ad_user` is used for changing the perms and `cmd_ad_user` is for running the `smbclient` command.
Yes

tests/api2/activedirectory.py
```diff
@@ -257,31 +243,31 @@ def test_19_setting_up_smb(request):
...
+def test_21_creating_a_smb_share_on_smb_path(request):
     depends(request, ["ad_01", "ad_02", "ad_07", "ad_10"])
     global payload, results, smb_id
     payload = {
         "comment": "My Test SMB Share",
-        "path": SMB_PATH,
-        "name": SMB_NAME,
+        "path": smb_path,
+        "name": smb_name,
         "guestok": True,
```
anodos325: Probably need to clearly differentiate between tests where we're explicitly testing guest access and not testing guest access. Creating a share with `guestok == True` alters global SMB server configuration. Maybe start with False, then update the share immediately prior to testing guest access in AD environment.

ericbsd: actually that should be remove or False thanks

tests/api2/activedirectory.py
ericbsd: If you think it is too much testing for SMB1, I can remove that.

tests/api2/activedirectory.py
```diff
@@ -462,129 +408,96 @@ def test_45_re_enable_activedirectory(request):
...
 
 
-# Clean up mounted SMB share
-@osx_host_cfg
-def test_58_Unmount_SMB_share(request):
+def test_49_delete_testdir_on_the_active_directory_share(request):
     depends(request, ["ad_01", "ad_02", "ad_07", "ad_10"])
-    results = SSH_TEST('umount -f "%s"' % MOUNTPOINT,
-                       OSX_USERNAME, OSX_PASSWORD, OSX_HOST)
+    command = fr'smbclient //{ip}/{SMB_NAME} -U {CMD_AD_USER}%{ADPASSWORD}' \
```
yocalebo, line 488: Why `fr` here when in `test_016_copy_testfile_in_testdir_on_the_active_directory_share` you're just doing `f`?

ericbsd: oh, I forgot to remove it.

ericbsd: That was when I was testing the smbclient failure.

tests/api2/activedirectory.py
```diff
@@ -8,29 +8,25 @@
...
+    PUT,
+    DELETE,
+    SSH_TEST,
+    cmd_test,
+    wait_on_job
+)
 
 try:
     from config import AD_DOMAIN, ADPASSWORD, ADUSERNAME, ADNameServer
+    AD_USER = fr"AD01\{ADUSERNAME.lower()}"
```
yocalebo, line 24: You can probably remove `fr` from these too

yocalebo (approved):
Thanks, 1 last minor nit

tests/api2/activedirectory.py
anodos325: Not a matter of too much testing. Just a philosophical point. :) 
We should probably stick to our default settings unless we're explicitly testing a non-default configuration.

tests/api2/activedirectory.py
ericbsd: flake8 will be annoying if I remove `r`. My editor gives me W605  https://www.flake8rules.com/rules/W605.html

ericbsd: Also, pytest complain about that.

tests/api2/activedirectory.py
ericbsd: Done

pkgdemon (approved):

PR #7347, opened 2021-08-18
NAS-111719 / 21.10 / Add safeguards to prevent usage of FreeBSD VFS objects on SCALE

anodos325:
There are two parts to this:
1) we raise ValidationError if user tries to create or update share with
   invalid VFS objects for SCALE. This should prevent new damage.
2) We refuse to add shares that already have invalid parameters to
   our SMB server configuration.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -1295,6 +1295,12 @@ class SharingSMBService(SharingService):
             'socket options',
             'include',
         ]
+        freebsd_vfs_objects = [
+            'zfsacl',
+            'zfs_space',
+            'noacl',
+            'ixnas',
+        ]
         for entry in data.splitlines():
             if entry == '' or entry.startswith(('#', ';')):
                 continue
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -1318,6 +1324,14 @@ async def validate_aux_params(self, data, schema_name):
                     'are not permitted.'
                 )
 
+            if kv[0].strip() == "vfs objects":
```
yocalebo, line 1327: Minor nit but you're using single-quotes everywhere except here.

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -1610,6 +1624,8 @@ async def sync_registry(self, job):
 
             try:
                 await self.middleware.call('sharing.smb.reg_addshare', share_conf[0])
+            except ValueError:
+                self.logger.warning("Share [%s] has invalid configuration.", share, exc_info=True)
```
yocalebo, line 1628: Same here with using double-quotes instead of single.

yocalebo (approved):

PR #7350, opened 2021-08-18
NAS-111795 / 21.10 / Bug fix for fix_unique migration

sonicaj:
Row proxy object's values cannot be changed, so we should convert it to a dict for temporary changes instead.


rick-mesta (approved):

william-gr:
The target for this ticket is 21.08

sonicaj:
backport

PR #7351, opened 2021-08-18
NAS-111902 / 21.10 / Remove dependency on samba3 loadparm context

anodos325:
Initializing a samba3 loadparm context requires that middleware
open registry.tdb and initialize a full s3 context. This PR
switches us to use a different samba python module that is
lighter-weight and doesn't require the tdb open. As a side-effect
it's utility is more limited in scope. It can be used to
look up default values for parameters (which may change between
samba releases), and it can be used to validate SMB auxiliary
parameters. The key limitation to it is that the limited
loadparm context can't be used to read parameters that are stored
in the static configuration file (/etc/smb4.conf) and so we
need an alternate method to read non-default parameters in that
file. In this case, we simply read and convert into JSON types.
There should only be a limited number of well-known parameters
inserted here (unless users are manually editing the files) and
so this doesn't need to be perfect at typing all the different
parameters in the smb4.conf. If the requested parameter isn't
in the smb4.conf file, then we use loadparm context to look up
the default value and return that.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -20,15 +20,9 @@ import uuid
 import shutil
 import struct
 
-try:
-    from samba.samba3 import param
-except ImportError:
-    param = None
-
+from samba import param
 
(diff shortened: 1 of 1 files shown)


rick-mesta (approved):
Noice 👌🏼

PR #7365, opened 2021-08-19
NAS-111938 / 21.10 / Create idmap service to wrap around winbind

anodos325:
Avoid touching winbindd unless it's absolutely necessary. NetBIOS names and workgroups cannot be changed while AD is
enabled and so we should be protected from having inconsistent config on winbindd and smbd.


themylogin (approved):

ericbsd (approved):
LGTM

PR #7367, opened 2021-08-19
NAS-111779 / 21.10 / improve IoThreadPoolExecutor

yocalebo:
QE team and myself were able to reproduce python core dumps via the webUI. The core dumps occurred when smartd was configured to monitor drives for temperature alerts as well as being configured to send email alerts when the drives hit the configured threshold. The problem is that the `gmail` plugin is using non thread-safe objects `httplib2.Http()` and we were specifically caching said object in our plugin. See https://github.com/truenas/middleware/pull/7377 for details.

However, the changes in this PR are a significant improvement as it allows us to remove the burden of managing significant amount of complex code, unnecessarily.

My change does a few things.

- derive our class from the upstream `ThreadPoolExecutor` class which unties us from the underlying implementation
- instead of "shrinking" "idle" threads, we will simply spin-up single-use threads when all the threads in the threadpool are busy
- when the threadpool gets an idle thread, we'll go back to using the threadpool

There should be no behavioral change.

### DIFF
--- a/src/middlewared/middlewared/utils/io_thread_pool_executor.py
+++ b/src/middlewared/middlewared/utils/io_thread_pool_executor.py
@@ -1,131 +1,30 @@
-from concurrent.futures import _base
-import itertools
-import logging
-import queue
-import random
-import threading
+from os import cpu_count
+from concurrent.futures import ThreadPoolExecutor
 
-import middlewared.utils.osc as osc
-
-logger = logging.getLogger(__name__)
+from middlewared.utils.osc import set_thread_name
 
 
-class WorkItem(object):
-    def __init__(self, future, fn, args, kwargs):
-        self.future = future
-        self.fn = fn
-        self.args = args
-        self.kwargs = kwargs
+class IoThreadPoolExecutor(ThreadPoolExecutor):
 
-    def run(self):
-        if not self.future.set_running_or_notify_cancel():
-            return
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._initializer = set_thread_name('IoThread')
 
-        try:
-            result = self.fn(*self.args, **self.kwargs)
-        except BaseException as exc:
-            self.future.set_exception(exc)
-            # Break a reference cycle with the exception 'exc'
-            self = None
-        else:
-            self.future.set_result(result)
+        # we set these to 21 or 33 respectively so that we
+        # always have a 1 idle thread buffer when we check
+        # the semaphore which should help prevent a non-fatal
+        # race condition with the caller of this method
+        # minimally we have 21 - 1 thread available
+        # on large cpu count systems we set it to 33 - 1 (to match upstream)
+        self._max_workers = 21 if ((cpu_count() or 1) + 4) < 32 else 33
 
+    @property
+    def no_idle_threads(self):
+        # note, this is "technically" an implementation
+        # detail of the threading.Semaphore class so upstream
+        # can change this variable at any time so I'm noting
+        # it here so my future self doesn't pull their hair
+        # out when this occurs :)
 
-class Worker:
-    def __init__(self, name, executor):
-        self.name = name
-        self.executor = executor
-
-        self.busy = False
-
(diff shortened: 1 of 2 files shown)


themylogin:
I am glad that it prevents crashes but then it's still a mystery how come that if we shut down a thread then we get a segfault, but when "disposable" `concurrent.futures.ThreadPoolExecutor(max_workers=1)` shuts down a thread everything it's fine. Have you tried crashing python when `self.run_in_thread_executor` is complete and every `run_in_thread` call spawns a new thread executor?

Should we also set constant `max_workers` for main thread pool executor?
```
Changed in version 3.8: Default value of max_workers is changed to min(32, os.cpu_count() + 4). This default value preserves at least 5 workers for I/O bound tasks. It utilizes at most 32 CPU cores for CPU bound tasks which release the GIL. And it avoids using very large resources implicitly on many-core machines.
```
And they'll change it again I am sure. We use this pool mostly for synchronous I/O, I doubt we should depend on CPU core count and get different I/O behavior on different machines.

yocalebo:
> I am glad that it prevents crashes but then it's still a mystery how come that if we shut down a thread then we get a segfault, but when "disposable" `concurrent.futures.ThreadPoolExecutor(max_workers=1)` shuts down a thread everything is fine.

There is a huge difference with how we implemented our thread pool compared to how upstream implemented it. For example, we use daemon threads, they do not. We use a `Queue`, they use a `SimpleQueue`. They use a `threading.Semaphore` object we do not. They use a global `weakref` object, we do not. Our implementation, when comparing to upstream, seems half-baked and not correct. I will admit, upstream has made changes to their implementation since we copied it but that's one of the main points on why I've made the change I've made here. It removes us from having to keep up with the implementation details since we don't have a complicated requirement (just need threads to be torn down).

> Have you tried crashing python when `self.run_in_thread_executor` is complete and every `run_in_thread` call spawns a new thread executor?
No I haven't. I'd be grateful if you know how to do this easily. I'll test it on my side no problem.

> Should we also set constant `max_workers` for main thread pool executor?
I went back and forth on this quite a bit. I will agree on one hand and say that I should probably default this to `20` like we had previously to maintain backwards functionality. However, we sell some large CPU count systems (m-series) so I'm wondering if we should use `32` when the cpu count is quite significant since, in theory, it will improve the performance of the API.

themylogin:
> I'd be grateful if you know how to do this easily. I'll test it on my side no problem.

Run some temporary method that will call `asyncio.ensure_future(self.middleware.run_in_thread(lambda: time.sleep(100000)))` 20 times

themylogin:
> However, we sell some large CPU count systems (m-series) so I'm wondering if we should use 32 when the cpu count is quite significant since, in theory, it will improve the performance of the API.

Agreed

themylogin:
backport

yocalebo:
> > I'd be grateful if you know how to do this easily. I'll test it on my side no problem.
> 
> Run some temporary method that will call `asyncio.ensure_future(self.middleware.run_in_thread(lambda: time.sleep(100000)))` 20 times

brilliant, thanks. I'll do that and test.

src/middlewared/middlewared/utils/io_thread_pool_executor.py
```diff
@@ -1,131 +1,30 @@
...
-
-                return work_item
-
-    def remove_worker(self, worker):
-        try:
-            self.workers.remove(worker)
-        except ValueError:
-            pass
+        # give ourselvs a single idle thread buffer
+        return self._idle_semaphore._value - 1 == 1
```
themylogin, line 30: Should it be `<= 1`?

themylogin (changes requested):

src/middlewared/middlewared/utils/io_thread_pool_executor.py
yocalebo: good catch, it should

themylogin (approved):
LGTM. Do you want to also backport this to 21.08-BETA1?

yocalebo:
> LGTM. Do you want to also backport this to 21.08-BETA1?

Yeah I think we should. I'll add label

PR #7369, opened 2021-08-20
NAS-111936 / 21.10 / Use libvirt to determine how much memory a VM is consuming

sonicaj:
We should use libvirt to determine how much memory a VM is consuming instead of trying to do it ourselves. With current logic, a VM memory size can sometimes go to negative with how it's being calculated which when subtracted from total actually results in addition of the vm size to memory instead of subtracting from it.

Using libvirt gives us benefit of not going into implementation specifics of how memory of a VM should be calculated and also we avoid getting into the problem stated above.

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/vm_info.py
+++ b/src/middlewared/middlewared/plugins/vm/vm_info.py
@@ -106,15 +106,12 @@ class VMService(Service):
             # the vm process is currently using and add the maximum memory its
             # supposed to have.
             for vm in await self.middleware.call('vm.query'):
-                status = await self.middleware.call('vm.status', vm['id'])
-                if status['pid']:
+                if vm['status']['state'] == 'RUNNING':
                     try:
-                        p = psutil.Process(status['pid'])
-                    except psutil.NoSuchProcess:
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/vm/vm_info.py
```diff
@@ -106,15 +106,12 @@ async def get_available_memory(self, overcommit):
...
             for vm in await self.middleware.call('vm.query'):
-                status = await self.middleware.call('vm.status', vm['id'])
-                if status['pid']:
+                if vm['status']['state'] == 'RUNNING':
                     try:
-                        p = psutil.Process(status['pid'])
-                    except psutil.NoSuchProcess:
+                        vms_memory_used += await self.middleware.call('vm.get_memory_usage_internal', vm)
+                    except Exception:
+                        self.logger.error('Unable to retrieve %r vm memory usage', exc_info=True)
```
themylogin, line 113: Missing argument for the format string

themylogin (changes requested):

themylogin (approved):

themylogin:
backport

themylogin:
backport

themylogin:
backport

themylogin:
backport

PR #7370, opened 2021-08-20
NAS-111885 / 21.10 / Update check_user method to allow checking passwords of other users

sonicaj:
This commit adds a change to check_user method adapting it so that it can be used for other users as well for checking if a specified password is accurate/correct against the specified username.

### DIFF
--- a/src/middlewared/middlewared/plugins/auth.py
+++ b/src/middlewared/middlewared/plugins/auth.py
@@ -283,8 +283,6 @@ class AuthService(Service):
         """
         Verify username and password
         """
-        if username != 'root':
-            return False
         try:
             user = await self.middleware.call('datastore.query', 'account.bsdusers',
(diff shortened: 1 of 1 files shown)


themylogin (commented):
Will this effectively allow any other non-root user to log in into Web UI and perform any administrative actions?

sonicaj:
@themylogin it would, however i was thinking UI should perhaps consume `auth.login` ? Otherwise we could revert this change and perhaps have another endpoint which would allow validating any user/pass ?

themylogin:
@sonicaj `auth.login` (and other authentication routines such as RESTful API) call `auth.check_user`.

We should probably move password check login to `user.check_password` and make `auth.check_user` call `user.check_password` after ensuring that the username is `root`.

themylogin (approved):

PR #7376, opened 2021-08-20
DOCS-2484 Replaced ssh to OSX and BSD by smbclient command

ericbsd:


anodos325 (approved):
Looks much cleaner. I like that test is more self-contained (not relying on outside infrastructure).

q5sys (approved):

PR #7384, opened 2021-08-20
NAS-111958 / 21.10 / Properly retrieve snapshots in bootenv plugin

sonicaj:


themylogin (commented):
What is the difference?

sonicaj:
@themylogin it returned a list before, now it's a dict object

themylogin (approved):

PR #7388, opened 2021-08-20
NAS-111540 / 21.10 / Always handle pool attachments when unlocking pool unless it's absolutely necessary

themylogin:


tests/api2/test_999_pool_dataset_unlock.py
```diff
@@ -0,0 +1,179 @@
...
+                            run(f"mount -t cifs -o guest //{ip}/encrypted /tmp/smb2")
+                        except subprocess.CalledProcessError as e:
+                            assert "No such file or directory" in e.stderr
+                        else:
+                            assert False
+
+                    if attachments_list_mode == "DENY":
+                        # We should be able to mount encrypted share
+                        try:
+                            run(f"mount -t cifs -o guest //{ip}/encrypted /tmp/smb2")
```
anodos325, line 176: Wrong SMB client. Use our python module see test_425

src/middlewared/middlewared/plugins/failover.py
```diff
@@ -582,10 +582,6 @@ async def unlock(self, options):
     @private
     @accepts(
         Str('pool_name'),
-        Dict(
-            'unlock_zfs_datasets',
-            Bool('restart_services', default=True),
-        )
```
william-gr, line 588: Could this make API backward compatible and somehow break failover during upgrade?
Just thinking out loud.

themylogin: @william-gr 12.0 signature is exactly the same `async def unlock_zfs_datasets(self, job, pool_name):` so this exact change should not be a failover breaker during CORE to SCALE update. Wrt 21.08 to 21.10 upgrade: is someone already using SCALE failover in production? 😄

william-gr: Nope

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2450,6 +2450,13 @@ def unlock(self, job, id, options):
 
         Uploading a json file which contains encrypted dataset keys can be specified with
         `unlock_options.key_file`. The format is similar to that used for exporting encrypted dataset keys.
+
+        `attachments_list_mode` and `attachments_list` controls which attachments (shares, tasks)
+        should be put in action after unlocking pool. By default all attachments are handled, which might
```
sonicaj, line 2455: unlocking `dataset(s)` ?

src/middlewared/middlewared/plugins/failover.py
themylogin: @william-gr do you want me to change something then?

sonicaj (approved):

src/middlewared/middlewared/plugins/failover.py
william-gr: @themylogin No, I was just asking to make sure you thought of that case.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2387,7 +2387,7 @@ async def lock(self, job, id, options):
             raise CallError(f'Please move system dataset to another pool before locking {id}')
 
         async def detach(delegate):
-            await delegate.stop((await delegate.query(self.__attachments_path(ds), True)))
+            await delegate.stop(await delegate.query(self.__attachments_path(ds), True))
```
yocalebo, line 2390: This is an async method calling a sync method. The sync method isn't doing anything that could block the event loop, but still isn't "proper". Could we convert `__attachments_path` to an async method? Or instead of changing that method, just duplicate what it's doing locally in this method?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2450,6 +2450,13 @@ def unlock(self, job, id, options):
 
         Uploading a json file which contains encrypted dataset keys can be specified with
         `unlock_options.key_file`. The format is similar to that used for exporting encrypted dataset keys.
+
+        `attachments_list_mode` and `attachments_list` controls which attachments (shares, tasks)
+        should be put in action after unlocking dataset(s). By default all attachments are handled, which might
+        theoretically lead to service interruption when daemons configurations are reloaded (this should not happen,
+        and if this happens it should be considered a bug). As TrueNAS does not have a state for resources that
```
yocalebo, line 2457: Forgive me, I've read this a few times and I still don't quite understand what this is explaining. I'm not really grok'ing the title of this PR and this comment just further confuses me. Specifically the bit about `put the system into an inconsistent state so it should really never be used.`

If we feel this strongly about it, then why do we even have this option?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2591,6 +2584,25 @@ def dataset_data(unlocked_dataset):
...
+            if options['attachments_list_mode'] == 'DENY':
+                if attachment_delegate.name in options['attachments_list']:
+                    continue
+
+            # FIXME: put this into `VMFSAttachmentDelegate`
+            if attachment_delegate.name == 'vm':
+                await self.middleware.call('pool.dataset.restart_vms_after_unlock', dataset)
+                continue
+
+            attachments = await attachment_delegate.query(self.__attachments_path(dataset), True, {'locked': False})
```
yocalebo, line 2602: Same thing as above, async calling sync method.

tests/api2/test_999_pool_dataset_unlock.py
```diff
@@ -0,0 +1,178 @@
...
+        c.disconnect()
+
+
+@pytest.mark.dependency(name="create_dataset")
+@pytest.mark.parametrize("attachments_list_mode", ["ALLOW", "DENY"])
+def test_pool_dataset_unlock_smb(request, attachments_list_mode):
+    #depends(request, ["pool_04", "smb_001"], scope="session")
+    # Prepare test SMB share
+    with dataset("normal") as normal:
+        os.chmod(f"/mnt/{normal}", 0o777)
```
anodos325, line 125: This test will be run on an external server, no?
`POST('/filesystem/setperm/', {'path': f'/mnt/{normal}', 'mode': "777"})`

src/middlewared/middlewared/plugins/pool.py
themylogin: There is nothing wrong in calling sync methods in async methods if they are not doing IO or CPU-intensive computation.

src/middlewared/middlewared/plugins/pool.py
themylogin: By "inconsistent state" I mean samba shares or cloud sync tasks or whatever that were LOCKED because underlying dataset was locked. Then, if underlying services were not reloaded when unlocking the dataset, the shares or tasks will end up being shown as non-locked, but, nevertheless, non-functional, that's what original ticket was about. There is no reason not to reload anything, because scst, NFS and Samba all can add shares without service interruption. We only use this option for failover to avoid double service reload.

tests/api2/test_999_pool_dataset_unlock.py
```diff
@@ -0,0 +1,178 @@
...
+        c.disconnect()
+
+
+@pytest.mark.dependency(name="create_dataset")
+@pytest.mark.parametrize("attachments_list_mode", ["ALLOW", "DENY"])
+def test_pool_dataset_unlock_smb(request, attachments_list_mode):
+    depends(request, ["pool_04", "smb_001"], scope="session")
+    # Prepare test SMB share
+    with dataset("normal") as normal:
+        POST("/filesystem/setperm/", {'path': f"/mnt/{normal}", "mode": "777"})
```
anodos325, line 125: Sorry, I should have been clearer. I was just giving example of what call to endpoint looks like. You should probably wait for the job to complete. For example:

```
results = POST("/filesystem/setperm/", {'path': f"/mnt/{normal}", "mode": "777"})
assert result.status_code == 200, result.text
JOB_ID = result.json()
job_status = wait_on_job(JOB_ID, 180)
assert job_status['state'] == 'SUCCESS', str(job_status['results'])
```

src/middlewared/middlewared/plugins/pool.py
yocalebo: Okay thanks, so am I understanding correctly that `ALLOW` actually means ignore the reloading/restarting of services and `DENY` means reload/restart?

yocalebo: Okay, I understand now. You've added an allow/deny list (I think...). Anyways, I do not like the verbiage change here. The previous API verbiage made more sense to me.

With your changes, we only have 2 options in the `enum=`. Why do we not simplify this to a singular boolean? Something like `restart_or_reload_services` as a boolean and default it to `True` and on failover we set it to `False`?

anodos325 (approved):

Qubad786:
@themylogin can you clarify please why we removed the ability to restart services?

PR #7392, opened 2021-08-23
Fix system dataset setup

themylogin:
* We don't need three ways to set up system dataset (`systemdataset.setup`, etc file and pseudo-service).
  First one is good enough.
* Set up system dataset in two places: pool import (both all pool import on boot and specific pool import)
  and on system dataset configuration update. Previously we were doing it from multiple places multiple
  times concurrently.
* Do not restart rrdcached, syslog, and, what's most important, samba, if system dataset setup has not
  mounted anything new (e.g. it was a pool import when everything is already set up correctly on another
  pool and we have nothing else to do besides double-checking that child datasets are mounted correctly).


src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -317,9 +282,8 @@ async def setup(self, mount, exclude_pool):
                 {'properties': {'acltype': {'value': 'off'}}},
             )
 
-        if mount:
-
-            await self.__mount(config['pool'], config['uuid'])
+        if True:
```
anodos325, line 285: This looks a little odd.

themylogin: @anodos325 this is to make review diff more clear, I'll dedent this after approval

themylogin:
@yocalebo @sonicaj ready for review

src/middlewared/middlewared/plugins/service_/services/all.py
```diff
@@ -55,10 +55,10 @@
     SysctlService,
     SyslogdService,
     SystemService,
-    SystemDatasetsService,
     TimeservicesService,
     TtysService,
     UserService,
+    WinbinddService,
```
anodos325, line 61: Why are we adding a WinbinddService when we already have IdmapService that wraps around winbindd?

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -557,18 +511,22 @@ async def migrate(self, _from, _to):
...
     """
-    if pool is None:
-        return
-
     await middleware.call('systemdataset.setup')
 
 
 async def setup(middleware):
-    middleware.register_hook('pool.post_import', pool_post_import, sync=True)
+    middleware.register_hook('pool.post_create', pool_post_create)
```
yocalebo, line 527: Do we now need to look (or add) a `pool.post_{delete/remove/export}` hook for when a pool is exported/destroyed etc and the system dataset exists on that pool? I think we already have one but wondering if it needs to be updated with these changes?

anodos325:
I think we've been bitten by enough bugs related to system dataset stuff that we probably need explicit tests for correct behavior in api2/tests so that it's in our CI. 

Not sure how much is feasible to add to scope of this PR, but if we create some tests that are dependent on `pool_04` test succeeding, then we'll at least have a single pool and can test with shifting system dataset to boot pool and exporting / importing that pool (assuming we don't have coverage here).

themylogin:
@anodos325 I had in mind creating an issue for Eric after we merge this

src/middlewared/middlewared/plugins/sysdataset.py
themylogin: @yocalebo great idea, done

themylogin:
https://jira.ixsystems.com/browse/DOCS-2529

yocalebo (approved):
Thanks! Only remaining item I have is the `if True` line that you said you would address after approval is given. Just making sure you don't forget about it 😄

PR #7395, opened 2021-08-24
NAS-111992 / 21.10 / Log parameter lookup failures due to missing config file

anodos325:


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -421,6 +421,9 @@ async def getparm(self, parm, section):
                 return await self.getparm_file(parm)
         except MatchNotFound:
             pass
+        except FileNotFoundError:
+            self.logger.debug("%s: smb.conf file not generated. Returning default value.", parm)
```
sonicaj, line 425: Just to be sure, would the return statement below be safe as the file couldn't be found ?

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -421,6 +421,9 @@ async def getparm(self, parm, section):
...
+            pass
```
sonicaj, line 426: This is not required ; )

sonicaj (approved):

src/middlewared/middlewared/plugins/smb.py
anodos325: Yeah, the loadparm context doesn't use this file.

PR #7398, opened 2021-08-24
NAS-111995 / 21.10 / Fix out of order operations in applying smb config

anodos325:
netbiosname_local is value that is applied to registry running
configuraiton. Ensure that we actually apply the new config
(passing through the extend() method) rather than the updated
dict. Shift service restart to end of smb.update() method
so that vfs_fruit changes are picked up by SMB clients as they
reconnect.


anodos325:
This PR addresses CI failure in test_420_051.

ericbsd (approved):
LGTM

PR #7400, opened 2021-08-24
NAS-111997 / 21.10 / Properly retrieve registry config and ACLs in SMB debug

anodos325:
`getfacl` is not ACLtype-agnostic. Parse sharing.smb.query output
and use `filesystem.getacl` to retrieve share ACL.

### DIFF
--- a/src/freenas/usr/local/libexec/freenas-debug/smb/smb.sh
+++ b/src/freenas/usr/local/libexec/freenas-debug/smb/smb.sh
@@ -118,8 +118,8 @@ smb_func()
 	section_header "net getdomainsid"
 	net getdomainsid
 	section_footer
-	section_header "net --json groupmap list"
-	net --json groupmap list | jq
+	section_header "middleware groupmap list"
+	midclt call smb.groupmap_list | jq
 	section_footer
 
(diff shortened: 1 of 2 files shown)


src/freenas/usr/local/libexec/freenas-debug/smb/smb.sh
```diff
@@ -124,8 +118,8 @@ smb_func()
 	section_header "net getdomainsid"
 	net getdomainsid
 	section_footer
-	section_header "net groupmap list"
-	net groupmap list | head -50
+	section_header "net --json groupmap list"
```
yocalebo, line 121: I remember this potentially barfing tons of information which is why we limited it to 50 entries. You concerned with the changes here?

anodos325: Oh, right, that's when we have ldap enabled. Will fix.

anodos325: When we use `smb.groupmap_list` and Samba's passdb backend is ldapsam, we will skip this.

yocalebo (approved):

PR #7402, opened 2021-08-24
Skip ALLOW test for NFS4 WRITE_ATTRIBUTES on SCALE

anodos325:
At this point, special handling of this permission is not
ipmlmented. Specifically, WRITE_DATA | WRITE_ATTRIBUTES are
required in order to alter timestamps. At this point no
adverse affect has been observed with SMB server and such
behavior is comparable with FreeBSD NFSv4 ACL implmenetation.
We can re-evaluate in future if it becomes a demonstrable
issue in real-life workloads, otherwise we'll keep
implementation simpler.


themylogin (commented):
We can remove all `if scale` from this test, can't we?

ericbsd:
> We can remove all `if scale` from this test, can't we?

I agree!

I am working on a PR that will remove scale from the master branch API tests

anodos325:
> We can remove all `if scale` from this test, can't we?

Sounds like Eric just volunteered for me to delegate that to him :)

ericbsd:
@anodos325  how do we process about that? We merged it and I make the changes on my branch?

ericbsd (approved):
LGTM

anodos325:
Per discussion, will `if scale` in separate PR.

PR #7416, opened 2021-08-26
NAS-112034 / 21.10 / Add smb data to account plugin

anodos325:
This PR removes dependency on winbindd being started for managing the SMB admin group. It also provides a way to expand the return of account info queries by specifying the extra option `extra: {"additional_information": ["SMB"]}`. At this point, the higher SMB info level only adds `nt_name` and `sid` info to the return (because we don't have easy method of getting full SAM account info for directory services users. These two items are also available for users and groups so some duplication of code is avoided.


themylogin (approved):
@anodos325 you can name it `join` instead of `info_level` (it essentially works like SQL JOIN)

PR #7426, opened 2021-08-27
NAS-112035 / 21.10 / Fix unit tests

themylogin:


src/middlewared/middlewared/plugins/disk.py
```diff
@@ -602,7 +601,7 @@ async def multipath_sync(self):
 
         reserved = await self.get_reserved()
 
-        devlist = await camcontrol_list()
+        devlist = []
```
sonicaj, line 604: `devlist` is expected to be a dictionary which would result in an error at L620 below

sonicaj (approved):

PR #7431, opened 2021-08-28
NAS-112058 / 21.10 / Allow intel GPU's to be used by multiple Apps

Ornias1993:
As discussed and not reported on Jira (afaik) here:
https://www.truenas.com/community/threads/intel-gpu-plugin.94955/

This allows the Intel GPU's using the Intel GPU plugin, to be used by multiple Apps at once.
As it currently throws errors when used by more than one App at the same time.

It's currently set to 5, which is more than the total amount of available GPU-supporting Apps in both Official and TrueCharts combined.


sonicaj:
Taking over this PR as discussed in DM

Ornias1993:
Yeah, totally okey with it :)

PR #7433, opened 2021-08-30
NAS-112019 / 21.10 / Check new pool available size before migrating system dataset there

themylogin:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -324,6 +324,14 @@ class Config:
         )
     )
 
+    @accepts(Str('name'))
+    @returns(Ref('pool_entry'))
+    async def get_instance_by_name(self, name):
```
yocalebo, line 329: 1. Where is this called?
2. I also thought this would raise `IndexError` if `name` doesn't exist?

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -204,6 +213,31 @@ async def do_update(self, job, data):
...
+        used = existing_dataset['properties']['used']['parsed']
+
+        try:
+            new_dataset = await self.middleware.call('zfs.dataset.get_instance', new_pool)
+        except InstanceNotFound:
+            return f'Dataset {new_pool} does not exist'
+
+        available = new_dataset['properties']['available']['parsed']
+
+        used = int(used * 1.1)
```
yocalebo, line 234: Why `1.1`?

yocalebo (changes requested):

src/middlewared/middlewared/plugins/sysdataset.py
themylogin: A safety margin because same files won't take exactly the same amount of space on a different pool.

src/middlewared/middlewared/plugins/pool.py
themylogin: I needed it, then I did not need it anymore, but I still think it might be useful in the future.

yocalebo: Will it raise a `ValidationError` though like the docstring says it will? I thought it will be `IndexError`?

themylogin: I pushed a commit that fixes this

themylogin: It raises `InstanceNotFound` which is a validation error

src/middlewared/middlewared/plugins/sysdataset.py
yocalebo: That's a great point, would you care to just add a small comment explaining why you're doing that?

yocalebo (approved):

PR #7445, opened 2021-08-31
NAS-111196 / 21.10 / Add returns to ACL template methods

anodos325:


rick-mesta (approved):

src/middlewared/middlewared/plugins/filesystem_/acl_template.py
```diff
@@ -54,10 +61,14 @@ async def validate_acl(self, data, schema, verrors):
         "acltemplate_create",
         Str("name", required=True),
         Str("acltype", required=True, enum=["NFS4", "POSIX1E"]),
-        List("acl", items=[Ref("posix1e_ace"), Ref("nfs4_ace")], required=True),
+        OROperator(Ref('nfs4_acl'), Ref('posix1e_acl'), name='acl', requried=True),
         register=True
     ))
-    async def create(self, data):
+    @returns(Ref('acltemplate_create'))
```
sonicaj, line 67: The returned object would mirror `acltemplate_entry` and not `acltemplate_create` ? If i am accurate, we don't need to add an explicit `returns` as for CRUD methods, middleware will automatically do that for us based on `ENTRY` specification

src/middlewared/middlewared/plugins/filesystem_/acl_template.py
```diff
@@ -83,8 +95,12 @@ async def create(self, data):
             ('attr', {'update': True})
         )
     )
+    @returns(Ref('acltemplate_create'))
```
sonicaj, line 98: Same as above

src/middlewared/middlewared/plugins/filesystem_/acl_template.py
```diff
@@ -116,6 +132,7 @@ async def do_update(self, id, data):
         return await self.get_instance(id)
 
     @accepts(Int('id'))
+    @returns()
```
sonicaj, line 135: This is not required and middleware will automatically add this for delete call in CRUD services

sonicaj (approved):
Thanks!

PR #7446, opened 2021-08-31
NAS-112098 / 21.10 / Fix KeyError on User / Group creation

anodos325:
Remove SMB keys prior to attempting DB insertion.


anodos325:
Checked that passes test_011.

rick-mesta (approved):

PR #7448, opened 2021-08-31
NAS-112097 / 21.10 / fix filesystem.statfs (round 2)

yocalebo:
Reading from `/proc/mounts` is not reliable in all scenarios so instead create `pathlib` object and deduce the information appropriately.

Benefit is that this is particularly fast.


anodos325:
Looks simple enough. 

Are we concerned about situations where users may have performed a bind mount over path? 
Are we concerned about the snapdirs?
Do we need to get realpath before walking back mountpoints?

anodos325 (approved):

PR #7452, opened 2021-09-01
NAS-109873 / 21.10 / Check for RRD update time in the future

themylogin:


src/middlewared/middlewared/plugins/reporting/update.py
```diff
@@ -129,19 +129,28 @@ async def do_update(self, data):
...
-            await self.middleware.call('reporting.setup')
-            await self.middleware.call('service.start', 'rrdcached')
+            await self.clear(False)
 
         await self.middleware.call('service.restart', 'collectd')
 
         return await self.config()
 
+    @accepts(Bool('start_collectd', default=True, hidden=True))
+    async def clear(self, start_collectd):
```
sonicaj, line 139: We are missing `returns` decorator

sonicaj (approved):

PR #7453, opened 2021-09-01
NAS-112111 / 21.10 / Update group api tests to use 'additional_information'

anodos325:
This commit changes a couple of minor details about how this
group of tests is written.

- SSH_CMD to run `getent group` is replaced with `get_group_obj` call
  through API. In this case the API call is a wrapper around
  the python `grp` module and therefore does the same thing as
  an explicit call to getent group.

- Replace SSH_CMD to `midclt call smb.groupmap_list` (private API)
  with calls to `group.query` with `additional_information: ["SMB"]`
  specified. This has the advantage of (1) being a public API,
  (2) adding test coverage for the `additional_information` in the
  group plugin, and (3) optimizing the test somewhat.

### DIFF
--- a/tests/api2/test_210_group.py
+++ b/tests/api2/test_210_group.py
@@ -71,10 +71,18 @@ def test_07_look_group_full_name():
 
 
 def test_08_look_for_testgroup_is_in_freenas_group(request):
-    depends(request, ["ssh_password"], scope="session")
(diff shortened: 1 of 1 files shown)


themylogin (approved):

ericbsd (approved):
LGTM

PR #7456, opened 2021-09-01
NAS-111179 / 21.10 / improve gathering DMI information

yocalebo:
On `middlewared` service start up or at boot time, 7 subprocess'es run back-to-back to fill out the DMI information. This does not scale well when the system is very busy. `dmidecode` allows you to specify all  the "types" at once so we can gather this information via 1 subprocess which is a 7x improvement in that respect. Furthermore, separate the dmi related methods to their own module in `system_/dmi.py`.

### DIFF
--- a/src/middlewared/middlewared/plugins/system.py
+++ b/src/middlewared/middlewared/plugins/system.py
@@ -53,7 +53,6 @@ FIRST_INSTALL_SENTINEL = '/data/first-boot'
 LICENSE_FILE = '/data/license'
 
 RE_KDUMP_CONFIGURED = re.compile(r'current state\s*:\s*(ready to kdump)', flags=re.M)
-RE_ECC_MEMORY = re.compile(r'Error Correction Type:\s*(.*ECC.*)')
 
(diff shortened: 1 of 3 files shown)


themylogin (changes requested):
`SystemService._parse_dmi` is a great candidate to add a few unit tests! Check out `src/middlewared/middlewared/pytest/unit/alert/source/test_ipmi_sel.py` for an example.

yocalebo:
Are the unit tests in `middlewared/pytest` run in any automated way. More importantly, if they fail is anyone notified?

themylogin:
@yocalebo @rick-mesta right now developers have to run them manually, but last week I started another attempt to integrate them into Jenkins pipeline.

themylogin (changes requested):
We use pytest instead of unittest. It is run as `pytest src/middlewared/middlewared/pytest/unit/plugins/system/test_dmi.py ` and test declarations are a little bit different.

themylogin (approved):
Thank you!

PR #7464, opened 2021-09-02
Use OAuth access token for support-proxy requests

themylogin:


rick-mesta:
@anodos325 / @sonicaj / @yocalebo <ping> 🔔

src/middlewared/middlewared/plugins/support.py
```diff
@@ -184,10 +179,10 @@ async def new_ticket(self, job, data):
         """
         Creates a new ticket for support.
         This is done using the support proxy API.
-        For FreeNAS it will be created on Redmine and for TrueNAS on SupportSuite.
+        For FreeNAS it will be created on JIRA and for TrueNAS on Salesforce.
```
yocalebo, line 182: Still a reference to `FreeNAS`. Sentence should probably be `For TrueNAS Core/Scale it will be created on JIRA, and for TrueNAS Core/Scale Enterprise on SalesForce.`

themylogin:
backport

PR #7468, opened 2021-09-02
NAS-112148 / 21.10 / Add test for complex groupmap behavior

anodos325:
Verify the following:
- Local server SID is used as prefix for local builtin accounts.
- local builtin accounts are added as foreign members to
  appropriate NT BUILTIN groups.
- NT BUILTIN group mappings exist


ericbsd (approved):
LGTM

PR #7471, opened 2021-09-03
NAS-112157 / 21.10 / Properly check if system supports virtualization

sonicaj:


src/middlewared/middlewared/plugins/vm/connection.py
```diff
@@ -27,22 +27,32 @@ def _close(self):
...
-            # Libvirt will still start even if kvm cannot be used on the machine which would falsely
-            # give the impression that virtualization can be used. We have checks in place to check if system
-            # supports virtualization but if we incorporate that check in all of the vm exposed methods which
-            # consume libvirt, it would be an expensive call as we figure that out by making a subprocess call
-            return self.LIBVIRT_CONNECTION and self.LIBVIRT_CONNECTION.isAlive() and os.path.exists('/dev/kvm')
+            return self.LIBVIRT_CONNECTION and self.LIBVIRT_CONNECTION.isAlive()
         return False
 
+    def _is_connection_alive(self):
+        return self._is_libvirt_connection_alive() and self._is_kvm_supported()
```
yocalebo, line 47: Does it make sense to check `self._is_kvm_supported()` first since `_is_libvirt_connection_alive()` will never work in that scenario and it short-circuits sooner?

sonicaj: Libvirt sadly still runs but yes we should check kvm first :)

yocalebo: > Libvirt sadly still runs but yes we should check kvm first :)

That's unfortunate 😵‍💫

yocalebo (approved):

src/middlewared/middlewared/plugins/vm/connection.py
sonicaj: Yeah, but with these changes we are not trying to start libvirt at all if system does not support virtualization - i was just pointing out that libvirt can still run without the system supporting virtualization.

PR #7472, opened 2021-09-03
NAS-112159 / 21.10 / Fix activedirectory idmap cache generation

anodos325:
Due to changes in winbind behavior over time with regard to
gencache, there is no longer a strong guarantee that
xid to sid entries will be present in gencache immediately after
pwd.getpwall and grp.getgrall calls.

This commit cleans up how we generate the AD cache. There are
two general use-cases we want to handle:
1) default - use `wbinfo` to enumerate all users and groups, then
  generate cache entries from the resulting list.

2) lower-impact cache fill (required by some environments), build
  up user/group cache from gencache.tdb contents.

To simplify some of the logic involved for joining idmaps to
AD domain info and user info, a new `extra` parameter is added
to `idmap.query` `extra: {"additional_information": ["DOMAIN_INFO"]}`
which attaches full domain information (including SID and long and
short names discovered through AD) to the idmap entries returned
to caller.

### DIFF
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -1389,6 +1389,9 @@ class ActiveDirectoryService(TDBWrapConfigService):
         finally:
             gencache.close()
 
+        if tdb_val is None:
+            return None
+
         decoded_sid = tdb_val[8:-5].decode()
         if decoded_sid == '-':
             return None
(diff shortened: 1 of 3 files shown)


ericbsd (approved):
The API test LGTM

Has for the middleware code, I do not see anything wrong at first glance.

rick-mesta (approved):
lgtm; @yocalebo / @sonicaj / @themylogin - any one of you gents care to take a second look 🙏🏼 ?

sonicaj (approved):

PR #7473, opened 2021-09-03
NAS-112163 / 21.10 / Properly apply default SMB acl on dataset creation

anodos325:
Underlying API behavior changed slightly. Legacy behavior for
compatibility with requests from old django UI was to apply a
default ACL if `mode` and `acl` were empty. New behavior
requires explicitly setting option
```
{'options': {'set_default_acl': True}}
```


ericbsd (approved):
LGTM

PR #7475, opened 2021-09-03
NAS-111702 / 21.10 / Fix microsoft_account feature

anodos325:
By default home editions of Windows will attempt to authenticate
to SMB shares using the user's microsoft account (email address
name). In legacy versions of FreeNAS the decision was made to
create a username mapping for samba to map a user's email address
to the account name. This simplifies particular edge cases
where Windows appears to "forget" the stored credentials for an
SMB server and attempts to authenticate using the "microsoft
account" credentials. During migration to a different smb.conf
backend, this behavior was broken. The mapping file was still
being created, but the smb.conf parameter to use it was not
added. We really don't want this parameter / file to be added
unilaterally. In early TrueNAS 11.3 releases it was observed
that these mappings / lookups in some cases could cause
failures to look up AD users, hence the restriction to the
case of servers that aren't joined to AD. At some future
point in time, this feature should be re-evaluated and possibly
removed.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -416,6 +416,10 @@ class UserService(CRUDService):
         if data['smb']:
             gm_job = await self.middleware.call('smb.synchronize_passdb')
             await gm_job.wait()
+            if data.get('microsoft_account'):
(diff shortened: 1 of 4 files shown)


rick-mesta (approved):

ericbsd (approved):
LGTM

PR #7476, opened 2021-09-03
NAS-111856 / 12.0 / Fix ups reporting when running in slave mode

sonicaj:
This commit fixes an issue when we would not be showing reporting for UPS when it was running in slave mode because collectd stores ups logs in a different directory when it's running in slave mode.

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/plugins_freebsd.py
+++ b/src/middlewared/middlewared/plugins/reporting/plugins_freebsd.py
@@ -212,11 +212,16 @@ class UPSBase(object):
 
     plugin = 'nut'
 
-    def get_identifiers(self):
+    @property
+    def _base_path(self):
         ups_config = self.middleware.call_sync('ups.config')
-        ups_identifier = ups_config['identifier']
         if ups_config['mode'] == 'SLAVE':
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/reporting/plugins_freebsd.py
```diff
@@ -213,7 +213,10 @@ class UPSBase(object):
     plugin = 'nut'
 
     def get_identifiers(self):
-        ups_identifier = self.middleware.call_sync('ups.config')['identifier']
+        ups_config = self.middleware.call_sync('ups.config')
+        ups_identifier = ups_config['identifier']
+        if ups_config['mode'] == 'SLAVE':
+            self._base_path = os.path.join(RRD_BASE_DIR_PATH, ups_config['remotehost'])
```
themylogin, line 219: Should we do this in the constructor instead?

Also RRDBase constructor has this
```
        self._base_path = RRD_BASE_PATH
        self.base_path = os.path.join(self._base_path, self.plugin)
```

Maybe we should just turn them into properties?

themylogin (changes requested):

rick-mesta:
@themylogin: ping 🔔 - has @sonicaj's changes addressed your concern(s) ?

themylogin (approved):

PR #7483, opened 2021-09-06
NAS-111491 / 21.10 / Tear down guest memory using libvirt events

sonicaj:
This PR adds changes to tear down guest memory using libvirt events as VMs can be shutdown from the VM itself as well which means that we would not be releasing the memory allocated back if a VM is shut down that way.

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/events.py
+++ b/src/middlewared/middlewared/plugins/vm/events.py
@@ -60,7 +60,9 @@ class VMService(Service, LibvirtConnectionMixin):
             if vm_id.isdigit() and emit_type != 'REMOVED' and dom.name() in vms:
                 vm = vms[dom.name()]
                 vm['status']['state'] = state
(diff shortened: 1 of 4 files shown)


rick-mesta (approved):
LGTM - @anodos325 / @yocalebo - can I get either of your pairs of eyes 👀 on this ? 🙏🏼

yocalebo (commented):
Is there a situation where we can have multiple vm.teardown_guest_memory for the same VM?

sonicaj:
@yocalebo how so ? Are you perhaps saying multiple events generated for a single action which is shutdown/poweroff ?

yocalebo:
> @yocalebo how so ? Are you perhaps saying multiple events generated for a single action which is shutdown/poweroff ?

Idk, that's why I'm asking you? 😄

sonicaj:
@yocalebo nope, we are safe on that end ;) The condition right now for checking the VM being stopped happens only once :p

yocalebo (approved):

PR #7484, opened 2021-09-07
NAS-112181 / 21.10 / Enforce certain properties on ix-applications dataset

sonicaj:
This PR adds changes to enforce certain properties on ix-applications dataset otherwise we get behavior which can change and result in wrong permissions and other issues inside containers like binaries not executing etc.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
@@ -137,10 +137,10 @@ class KubernetesService(Service):
     @private
     def k8s_props_default(self):
         return {
+            'acltype': 'posix',
             'exec': 'on',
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
```diff
@@ -134,6 +134,15 @@ async def post_start_internal(self):
         # because it's internal traffic will also be otherwise advertised to the default route specified
         await self.middleware.call('k8s.cni.add_routes_to_kube_router_table')
 
+    @private
+    def k8s_props_default(self):
+        return {
+            'acltype': 'posix',
```
anodos325, line 140: `'aclmode': 'discard'` should also be set.

anodos325 (approved):

PR #7494, opened 2021-09-08
NAS-112216 / 21.10 / Ensure directory service cache initialization on failover

anodos325:
Make sure that we wait for smb configuration to complete
while becoming vrrp master.

Always refresh dscache during directoryservices.initialize.

### DIFF
--- a/src/middlewared/middlewared/plugins/directoryservices.py
+++ b/src/middlewared/middlewared/plugins/directoryservices.py
@@ -429,6 +429,8 @@ class DirectoryServices(Service):
         if not ldap_enabled and not ad_enabled:
             return
 
+        health_check = 'activedirectory.started' if ad_enabled else 'ldap.started'
+
         has_secrets = await self.middleware.call("directoryservices.secrets_has_domain", workgroup)
 
         if ad_enabled and not has_secrets:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/failover_/event.py
```diff
@@ -511,6 +511,16 @@ def vrrp_master(self, job, fobj, ifname, event):
         logger.info('Configuring system dataset')
         self.run_call('systemdataset.setup')
 
+        smb_setup = self.run_call(
+            'core.get_jobs',
+            [('method', '=', 'smb.configure')],
+            {'order_by': ['-id']}
+        )
+
+        if not smb_setup:
```
yocalebo, line 520: Am I going crazy? Should this not be `if smb_setup:` ?

Also, can you add a brief comment on why we do this since `systemdataset.setup` returns before it's actually complete...

yocalebo (changes requested):

src/middlewared/middlewared/plugins/failover_/event.py
yocalebo: Adding the comment is for my feeble brain so that I know why we're doing this on failover 😄

anodos325: No, you're not going crazy, I am.

yocalebo (approved):

PR #7495, opened 2021-09-08
NAS-112123 / 21.10 / create cluster API test framework

yocalebo:
This adds a new area `middlewared/tests/cluster` that will be used for writing cluster specific API tests. This commit is the first part needed before actually writing the api tests.

1. create `run-cluster-tests.py` that will be called by an internal Jenkins CI job
2. store the various cluster related config in `config.py` so when changes are made we don't have to rewrite everything
3. write wrapper methods for making REST and/or Websocket calls in `utils.py`
4. add `--initialize-cluster` argument to `run-cluster-tests.py` that is responsible for "initializing" all the cluster nodes on each Jenkins CI job since the cluster config is wiped (by design) when the Jenkins job is run.
5. adds wrapper methods for running ssh commands on the nodes in the cluster (untested)

### DIFF
--- a/tests/cluster/config.py
+++ b/tests/cluster/config.py
@@ -1,0 +1,18 @@
+from os import environ
+
+
+CLUSTER_INFO = {
+    'CLUSTER_IP': environ.get('CLUSTER_IP'),
+    'NODE_A_IP': environ.get('NODE_A_IP'),
+    'NODE_B_IP': environ.get('NODE_B_IP'),
+    'NODE_C_IP': environ.get('NODE_C_IP'),
+    'NETMASK': int(environ.get('NETMASK')),
+    'INTERFACE': environ.get('INTERFACE'),
+    'DEFGW': environ.get('DEFGW'),
+    'DNS1': environ.get('DNS1'),
+    'APIUSER': environ.get('APIUSER'),
+    'APIPASS': environ.get('APIPASS'),
+    'ZPOOL_DISK': environ.get('ZPOOL_DISK'),
+    'ZPOOL': environ.get('ZPOOL'),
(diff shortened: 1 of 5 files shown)


rick-mesta (approved):
LGTM 😁

tests/cluster/config.py
```diff
@@ -0,0 +1,19 @@
+DOMAIN = '.tn.ixsystems.net'
+NET_INFO = {
```
ericbsd, line 2: @pkgdemon and @yocalebo should this be source somewhere else?

tests/cluster/config.py
```diff
@@ -0,0 +1,19 @@
...
+    'CLUSTER': {'ip': '10.234.64.71', 'fqdn': f'tn-kvm01{DOMAIN}'},
+    'NODE_A': {'ip': '10.234.65.10', 'fqdn': f'tn-kvm01-nodea{DOMAIN}'},
+    'NODE_B': {'ip': '10.234.65.11', 'fddn': f'tn-kvm01-nodeb{DOMAIN}'},
+    'NODE_C': {'ip': '10.234.65.12', 'fqdn': f'tn-kvm01-nodec{DOMAIN}'},
+    'NETMASK': 22,
+    'INTERFACE': 'enp1s0',
+    'DEFGW': '10.234.64.1',
+    'DNS1': '10.231.0.2',
+    'DOMAIN': DOMAIN,
+}
```
ericbsd, line 12: @yocalebo and @pkgdemon, should this be sourced privately?

yocalebo: @ericbsd I see no harm with this being public 🤷

yocalebo:
I pushed changes to our private repo for sourcing the cluster configuration information and then updated the PR accordingly.

PR #7505, opened 2021-09-10
NAS-112255 / 21.10 / add R20B to enclosure plugin

yocalebo:


darkfiberiru:
More complete and tested entry here https://github.com/truenas/middleware/pull/7506

PR #7507, opened 2021-09-10
NAS-112248 / 21.10 / add framework for calling pytest-3 for cluster API tests

yocalebo:
I have to move this to it's own base directory since nesting it inside `middlewared/tests` caused many import errors pytest related collisions and such since that directory is already setup with a `tests` subdir.

Instead, to keep it simple, I've moved the cluster api tests to `middlewared/cluster-tests`.


rick-mesta (approved):
lgtm

PR #7519, opened 2021-09-13
NAS-112299 / 12.0 / fix ctld.conf generation when sharing raw disks

yocalebo:
The changes in `12.0-U5.1` to optimize systems with 100's+ of hard drives broke an edge-case where we allow configuration of iSCSI on CORE sharing raw disks. (SCALE does not allow this, thankfully).

### DIFF
--- a/src/middlewared/middlewared/etc_files/ctld.py
+++ b/src/middlewared/middlewared/etc_files/ctld.py
@@ -1,10 +1,11 @@
-from middlewared.client.utils import Struct
 import contextlib
 import logging
 import os
 import subprocess
 import sysctl
 
+from bsd import geom
+from middlewared.client.utils import Struct
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


rick-mesta (approved):

src/middlewared/middlewared/etc_files/ctld.py
```diff
@@ -212,7 +213,7 @@ def main(middleware):
     poolthreshold = {}
     zpoollist = {i['name']: i for i in middleware.call_sync('zfs.pool.query')}
 
-    system_disks = middleware.call_sync('device.get_disks')
```
sonicaj, line 215: Just to confirm, is this not providing all the disks ?

yocalebo: It is but not really what the question is in reference to?

yocalebo: Oh I see, you're talking about sharing `multipath` raw disks

yocalebo: Hmm, no just kidding. Not sure what you're asking? 😄

PR #7522, opened 2021-09-14
NAS-112041 / 12.0 / Fix disk.identifier_to_device not working with {uuid} and {label} identifiers

themylogin:


yocalebo (changes requested):
Hmm, not sure how this fixes anything? The idea with geom_xml is to provide a potentially significant optimization. All it represents is the xml being instantiated via the caller so that we don't generate that data for every disk

yocalebo (approved):
disk.sync_all passes DISK xml while these need a different type of XML. So the users in this scenario don't get the optimization. That sucks but the way this is designed there is no easy way around it.

PR #7523, opened 2021-09-14
NAS-112038 / 12.0 / Rename non-builtin user/group if a builtin user/group with a same name should exist

themylogin:


fracai:
This PR is going to change the group name for a group that I've had for years, predating the introduction of that group as a builtin group. That seems somewhat user hostile.

themylogin:
@fracai if OS needs a builtin user, it must have it; unfortunately, we don't have any other choice.

fracai:
There are choices though. The UID/GID can be changed. It can use the UID/GID of the existing user/group that I had created previously. Instead, this is going to change the name of a user and group that I already had, so the system can use a different ID for that user and group.

themylogin:
@fracai this won't work, builtin users and groups must have specific UID/GID.

fracai:
Can you indulge me and explain why? For example, why is it important that the media user/group is 8675309?

rick-mesta:
@yocalebo / @sonicaj / @anodos325: can I get a second pair of eyes on this fix, please ?

themylogin:
@fracai root filesystem image will have system files that should belong to `media` user owned by uid `8675309`

rick-mesta (approved):

fracai:
@themylogin I'm really trying to understand this and I just ran a search and didn't find any such files. I also wouldn't expect there to be any give that it's a `nologin` user with no home directory. Sorry to be persistent, I really am just trying to educate myself about this.

sonicaj (approved):

themylogin:
@fracai it's not about the specific `media` user, it's about general policy: if OS wants specific users with specific UIDs to be present, we are going to force that. The fact that this user does not own any files now does not mean that they won't appear in the future.

themylogin:
backport

anodos325:
`media` group was initially added as a way of coordinating permissions between host and jails in FN 9.10. Discussion of merits and success of this approach are somewhat beyond scope of this pull request. This was a documented feature and so I think any changes to behavior of `media` are not acceptable in 12-stable. The change itself was introduced 5 years ago, and for better or worse people have used it and probably continue to use it.

There doesn't appear to have been any explicit handling for this particular edge case (existing media group) at that time. So at best behavior was undefined but somewhat consistent. Insofar as we now have concrete handling in this PR for conflicts between builtin-in / not built-in user and group names, it is acceptable.

I think deprecating this user / group in master branch is appropriate. It's clear that the usefulness of this group has run its course, but we have to account for users who may have used it to set on-disk permissions. This is inherently a riskier migration than simply changing the name of a group.

fracai:
Agreed with all that, though I could have sworn there was indeed handling for existing media user/groups. I even think I remember looking at the code that checked if there was an existing group. Though maybe I'm misremembering that.

In any event, would it be useful for me to open a ticket regarding deprecating the media group?

anodos325:
> In any event, would it be useful for me to open a ticket regarding deprecating the media group?

Yes, please feel free to open a ticket about it. At this point we've only had an internal discussion about doing it.

fracai:
https://jira.ixsystems.com/browse/NAS-112326

jacobblock:
This broke ssh and NFS shares for my user+group `media:media`. +1 to having this for ~6-7 years.

Curious if this made my NFS shares accessible to `media:media` instead of the newly generated `media_1:media_1`. I was using mapall permissions.

PR #7524, opened 2021-09-14
NAS-112263 / 21.10 / fix xmit_hash_policy and lacp_rate on SCALE

yocalebo:
The public API accepts only upper-case values for these arguments but the kernel expects these values to be lower-case.

These 2 properties from `LaggMixin` are returned directly from `sysfs` which stores them in lower-case. We also store them in the database in lower-case. Finally, when changing these values using `ip link` command, it expects the arguments to be lower-case or it will throw an invalid argument error.


themylogin (approved):

sonicaj:
ok to test

sonicaj:
retest this please

PR #7528, opened 2021-09-14
NAS-112328 / 12.0 / set nfsd.scope and nfsd.owner_major on Core HA

yocalebo:


src/middlewared/middlewared/plugins/nfs_/v4_freebsd.py
```diff
@@ -23,6 +23,31 @@ def setup_v4(self):
...
+                    owner_major = str(uuid.uuid4())
+                    # write this to the database since it shouldn't change after it's been set
+                    # and has to be shared between controllers on HA systems
+                    self.middleware.call_sync(
+                        "datastore.update", "services.nfs", config["id"],
+                        {"v4_owner_major": owner_major},
+                        {"prefix": self._config.datastore_prefix},
+                    )
+
+                sysctl.filter("vfs.nfsd.scope")[0].value = owner_major
```
rick-mesta, line 48: you're assigning `owner_major` to both `scope` **and** `owner_major` ?

yocalebo: Yeah they both need to be set to the same string. I'll update comments

rick-mesta: kewl -- yeah, comment should be good 👍🏼

rick-mesta (approved):

PR #7529, opened 2021-09-15
NAS-112225 / 21.10 / Use "S.M.A.R.T. extra options" when probing for S.M.A.R.T. capacity

themylogin:


src/middlewared/middlewared/plugins/smart.py
```diff
@@ -24,7 +24,7 @@ async def annotate_disk_smart_tests(middleware, devices, disk):
     if disk["disk"] is None:
         return
 
-    args = await get_smartctl_args(middleware, devices, disk["disk"])
+    args = await get_smartctl_args(middleware, devices, disk["disk"], disk["smartoptions"])
```
yocalebo, line 27: `disk["smartoptions"]` will raise a `KeyError` I think because `disk_choices` has `full_disk` set to `False` by default so it i'll will only return `{disk['identifer']: disk['name']}` instead of the full disk information.

yocalebo (changes requested):
Think this will raise `KeyError`

src/middlewared/middlewared/plugins/smart.py
themylogin: I amended the commit today, now I pass `full_disk=True`

yocalebo: Seems my browser had cached previous version of commit 😵‍💫

yocalebo (approved):

PR #7536, opened 2021-09-16
NAS-112359 / 21.10 / lessen the restriction on the minimum MTU size

yocalebo:
68 is the minimum IP MTU for IPv4 and there are valid reasons (albeit rare imo) that require using a smaller MTU than 1492. Instead of choosing an arbitrary number, just use what the RFC states (https://tools.ietf.org/html/rfc791).

Added some extra validation for the scenario where an IPv4 and IPv6 address is on an interface. In this scenario, the minimum IP MTU for IPv6 is 1280 (https://tools.ietf.org/html/rfc2460).


yocalebo:
NOTE: There is a dependent webUI change that will have to occur before the mtu validation that I added will work. Ticket is [here](https://jira.ixsystems.com/browse/NAS-112363)

themylogin (approved):

yocalebo:
Merging this for now even though the validation won't work via the webUI. (They still haven't fixed their side but should by the time 21.10-RC1 is released)

PR #7546, opened 2021-09-17
DOCS-2373 / 21.10 / Added propper timeout with exit 1 in boot.exp and install.exp

ericbsd:
Changed the install timeout to 20 min instead of 30 minutes
Changed the boot timeout to 12 min instead of 30 minutes


q5sys (approved):
LGTM

PR #7550, opened 2021-09-20
Include INTELLIGENT_TIERING as available S3 storage class

ymylei:
`INTELLIGENT_TIERING` currently not available as a selectable storage class for Cloud Sync Tasks. PR is for making it available for selection.

S3 API Docs: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#API_PutObject_RequestSyntax
S3 Storage Class Docs: https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage-class-intro.html


themylogin:
@ymylei thank you! The RC.1 will not have this change, but RC.2 will.

PR #7558, opened 2021-09-20
NAS-112446 / 21.10 / rewrite coredump code on SCALE

yocalebo:
This fixes 3 primary issues:

1. we alert the end-user to open a ticket on Jira and then tell them to remove the core file once they've uploaded it to us but the interval for running this alert is every 6 hours. This means if the user removes that core file, the alert will linger in the webUI for the remainder of the interval that's left from when we last checked. This causes time to be spent on a non-issue.
2. we're calling `subprocess` and while we're doing every 6hrs, it still isn't "proper" especially since the `systemd.journal.Reader` class allows us to bypass it completely.
3. Because we're no longer using `subprocess` we no longer cache a dict of coredump information. Before this change, we were growing that dict unbounded which means, in theory, it could accumulate lots of RAM usage.

My changes allows us to check for coredumps every 60 seconds (polling alert system default) and doesn't cache anything to forgo unnecessary complexity.

### DIFF
--- a/src/middlewared/middlewared/plugins/system_/coredump.py
+++ b/src/middlewared/middlewared/plugins/system_/coredump.py
@@ -14,16 +14,20 @@ class SystemService(Service):
             with journal.Reader() as reader:
                 reader.add_match(CODE_FUNC='submit_coredump')
                 for core in reader:
-                    coredumps.append({
+                    coredump = {
                         'time': core['COREDUMP_TIMESTAMP'].strftime('%c'),
                         'pid': core['COREDUMP_PID'],
                         'uid': core['COREDUMP_UID'],
                         'gid': core['COREDUMP_GID'],
                         'unit': core['COREDUMP_UNIT'],
                         'sig': core['COREDUMP_SIGNAL'],
-                        'corefile': 'present' if exists(core['COREDUMP_FILENAME']) else 'missing',
                         'exe': core['COREDUMP_EXE'],
-                    })
+                    }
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/system_/coredump.py
```diff
@@ -1,33 +1,30 @@
...
+                reader.add_match(CODE_FUNC='submit_coredump')
+                for core in reader:
+                    coredumps.append({
+                        'time': core['COREDUMP_TIMESTAMP'].strftime('%c'),
+                        'pid': core['COREDUMP_PID'],
+                        'uid': core['COREDUMP_UID'],
+                        'gid': core['COREDUMP_GID'],
+                        'unit': core['COREDUMP_UNIT'],
+                        'sig': core['COREDUMP_SIGNAL'],
+                        'corefile': 'present' if exists(core['COREDUMP_FILENAME']) else 'missing',
```
anodos325, line 24: `coredumpctl` can show `none` here if corefile failed to generate rather than having it simply be missing. Have we checked whether this edge cases causes `core['COREDUMP_FILENAME']` to be `None` rather than a string? If it's None type then `exists(core['COREDUMP_FILENAME'])` will raise a TypeError.

anodos325: Simple way to check is to set a 1 MiB quota on the <sysdataset>/cores and then kill -6 an unwitting victim.

anodos325 (approved):

PR #7560, opened 2021-09-21
NAS-112467 / 12.0 / fix TypeError crash in enclosure plugin on 12

yocalebo:
`product` cannot be `None` because we use built-in `str` methods on that variable as well as pass it to `re`'s various methods.  This specific use-case crashed at `MINI_REGEX.match(...` with a `TypeError: expected string or bytes-like object`.

So when we instantiate an `Enclosure` class, we'll make sure the `self.product` variable is initiated to an empty string always.


yocalebo:
This is no longer an issue in 12.0-U6 since SMBIOS information is initiated to an empty string.

PR #7561, opened 2021-09-21
NAS-112468 / 21.10 / Update middleware for samba changes in 4.15

anodos325:
CLI tools use different args for kerberos-related methods.

Multichannel support is now default in samba. Plumb though
our schema 'multichannel' parameter. Disabled by default, currently
raises ValidationError (until we have satisfactory testing in place).

My max xattr size parameter was accepted upstream with slightly different
implementation. Adjust parameter name accordingly.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb_/smbconf/reg_service.py
+++ b/src/middlewared/middlewared/plugins/smb_/smbconf/reg_service.py
@@ -352,7 +352,7 @@ class ShareSchema(RegistrySchema):
 
         data_out['vfs objects']['parsed'].append("streams_xattr")
         if not data_in.get("cluster_volname"):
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/smb_/smbconf/reg_service.py
```diff
@@ -93,6 +93,7 @@ def order_vfs_objects(vfs_objects, is_clustered, fruit_enabled, purpose):
 
         data_out['vfs objects'] = {"parsed": ["io_uring"]}
         data_out['ea support'] = {"parsed": False}
+        data_out['ea support'] = {"parsed": False}
```
yocalebo, line 96: `ea_support` is set twice?

yocalebo (approved):

rick-mesta (approved):

PR #7563, opened 2021-09-21
NAS-112444 / 13.0 / Keep string typing for dmidecode_info to avoid bugs in legacy code

themylogin:


yocalebo (changes requested):
This will break this method completely since we use the `None` value as a way to identify if we've tried to fill the cache out.

```
    def dmidecode_info(self):
        if all(v is None for k, v in SystemService.CACHE.items()):
            cp = subprocess.run(['dmidecode', '-t', '1,2,16'], encoding='utf8', capture_output=True)
            self._parse_dmi(cp.stdout.splitlines())

yocalebo (approved):

yocalebo (changes requested):
Actually, now that the cache is empty strings, inside the `_parse_dmi` method when we detect no DMI information you can remove line 35.

PR #7566, opened 2021-09-21
prevent ds/zvol creation if parent is readonly=on

yocalebo:
If the parent of a dataset or zvol is `readonly=on` then creation of the dataset is permitted via zfs. However, the mounting of the dataset (or opening of the underlying zvol) will fail with related read only errnos. There is no way to mount a dataset via the webUI so prevent the creation of said dataset/zvol in this scenario.


themylogin (approved):

yocalebo:
backport

PR #7570, opened 2021-09-22
Require method to be @private or have @accepts

themylogin:


anodos325:
Can't you just have midcli treat all methods that don't explicitly have an accepts decorator as private?

PR #7576, opened 2021-09-23
NAS-112521 / 21.10 / Generate SSH keys on startu if they are absent. Force SSH keys regene…

themylogin:
…ration if current database does not have them

### DIFF
--- a/src/middlewared/middlewared/plugins/ssh.py
+++ b/src/middlewared/middlewared/plugins/ssh.py
@@ -1,13 +1,13 @@
 import base64
 import hashlib
 import os
+import subprocess
 import syslog
 
 from middlewared.schema import accepts, Bool, Dict, Int, List, Patch, returns, Str, ValidationErrors
 from middlewared.validators import Range
 from middlewared.service import private, SystemServiceService
 import middlewared.sqlalchemy as sa
-from middlewared.utils import osc
 
 
 class SSHModel(sa.Model):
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/ssh.py
```diff
@@ -190,18 +192,40 @@ def save_keys(self):
...
+        ]
+    ]
+
+    @private
+    def cleanup_keys(self):
+        config = self.middleware.call_sync("datastore.config", "services.ssh")
+        for path, column in self.keys:
+            if not config[column] and os.path.exists(path):
+                self.middleware.logger.warning("Removing irrelevant SSH host key %r", path)
+                os.unlink(path)
```
yocalebo, line 204: Should we catch an error here and log it if `unlink` fails for whatever reason?

src/middlewared/middlewared/service.py
```diff
@@ -1463,6 +1464,11 @@ def job_update(self, id, data):
                 extra=progress.get('extra'),
             )
 
+    @private
+    def is_boot(self):
```
yocalebo, line 1468: The name of this method is kind of strange. If I'm understanding your comment correctly, maybe name this something more appropriate like `def is_starting_during_boot` or something similar.

src/middlewared/middlewared/plugins/ssh.py
themylogin: I can't imagine a reason it fails. We are writing/deleting files on rootfs all over middleware without any exception handling.

yocalebo (approved):

PR #7588, opened 2021-09-27
NAS-112442 / 21.10 / fix bridge/bond interface choices on SCALE

yocalebo:
On SCALE, the kernel network API is much more lenient on what is allowed compared to CORE.

This allows the following types of configurations.

1. adding pretty much any type of supported interface to a bridge
--interfaces that are already a member of another `br` or `bond` are not allowed
--other `br` interfaces are not allowed to be added as a member interface (kernel API prevents this)

2. lessening the restriction on what can be added as a member interface to a `bond` type
--interfaces that are already part of another `br` or `bond` are not allowed
--vlans are not allowed to be added as members to a `bond` (kernel API prevents this)

I'm not using the `interface.choices` method anymore because that's used in many places and I don't want to risk behavioral change for public API endpoint. The benefit of this is that I remove an extra `datastore.query` call.

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -1943,51 +1943,73 @@ class InterfaceService(CRUDService):
     @returns(Dict(additional_attrs=True))
     async def bridge_members_choices(self, id):
         """
-        Return available interface choices for `bridge_members` attribute.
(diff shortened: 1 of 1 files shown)


themylogin (approved):
Great job, it is much better now!

PR #7601, opened 2021-09-29
Load ntb_netdev module on boot.

amotin:
I've fixed problem with NTB modules load order, so now this should work, providing NTB network interface on X and M-series, though on X-series it may still not work, haven't checked.  It is still not complete, using one hack, and so it is not future-compatible on protocol side, but should allow basic HA integration and testing.


themylogin (approved):
Just curious, how did you fix "no such device" issue? In the kernel source itself?

amotin:
Yes, kernel patching is what I am doing: https://github.com/truenas/linux/commit/be192de086ba65e4a170f50231a8365400e4717c :)

PR #7606, opened 2021-09-29
NAS-112605 / 12.0 / Ensure we have clean state for SMB VSS tests

anodos325:
Flush out samba's gencache before starting test, also
chown datasets so that they're owned by our SMB user.


ericbsd (approved):
LGTM

rick-mesta (approved):

PR #7611, opened 2021-09-30
NAS-112627 / 12.0 / make DiskStats not suck

yocalebo:
On an M60 with `1248` disks, these are the problems I've found:

1. The way we are calling `iostat` produces `2519` lines of text. We were parsing this every 2 seconds so we were parsing `5038` lines of text and discarding over half of it because we were calling `iostat` in a way that included the `pass` devices in the output.
2. we started a daemon thread that ran `subprocess` consistently to gather this information so we were having to read large output from child process stdout which is painful

Here is how I've fixed these problems.
1. `psutils.disk_io_counters(perdisk=True, nowrap=True)` gives us the _exact_ same values as `iostat`
2. get rid of the daemon thread
3. get rid of the child process that produces > 5k lines of text
4. no more absurd amount of text parsing

Before the changes, python had a thread that was eating 100% of a cpu core 100% of the time. Now it hardly shows in `top` output.

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/events.py
+++ b/src/middlewared/middlewared/plugins/reporting/events.py
@@ -166,7 +166,7 @@ class RealtimeEventSource(EventSource):
             'time': time.monotonic(),
             'speeds': self.get_interface_speeds(),
         }
-        self.disk_stats = DiskStats(2)
+        self.disk_stats = DiskStats()
 
         while not self._cancel.is_set():
             data = {}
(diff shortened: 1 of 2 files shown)


anodos325 (approved):

themylogin (approved):

themylogin (changes requested):
Previous code was producing read/write ops/bytes count and disk busy time % for `reporting.realtime` interval (2 seconds). Current code produces total ops/bytes for the system lifetime.

yocalebo:
> Previous code was producing read/write ops/bytes count and disk busy time % for `reporting.realtime` interval (2 seconds). Current code produces total ops/bytes for the system lifetime.

Fixed the problem and also removed the `osc` related code to clean this up.

themylogin (changes requested):
This code still provides read bytes count since system startup

Previous code was providing read bytes count in the last 2 seconds

yocalebo:
> This code still provides read bytes count since system startup
> 
> Previous code was providing read bytes count in the last 2 seconds

Subtle behavior that I didn't notice, thanks!

themylogin (approved):

PR #7612, opened 2021-09-30
NAS-112608 / 21.10 / Proper HA system dataset handling

themylogin:


src/middlewared/middlewared/plugins/failover_/event.py
```diff
@@ -172,7 +172,7 @@ def run_call(self, method, *args):
             # `self.validate()` calls this method
             raise
         except Exception as e:
-            logger.error('Failed to run %s:%r:%s', method, args, e)
+            logger.error('Failed to run %s%r: %s', method, args, e)
```
yocalebo, line 175: Since you've changed this would it make sense to remove `e` and just use `exc_info=True`?

yocalebo (approved):

PR #7617, opened 2021-10-01
NAS-112619 / 12.0 / Fix groupmap synchronization

anodos325:
Due to API changes, we now need full idmap config in passive
controller smb4.conf file. We also need to ensure that
RID counter for passdb and groupmap operations is same.


rick-mesta (approved):

rick-mesta:
@ericbsd: ping 🔔

ericbsd (approved):
LGTM

PR #7622, opened 2021-10-04
NAS-112656 / 21.10 / Run midclt as non-privileged user in API key tests

themylogin:


yocalebo:
This is great usage of `psutil`. Nice job!

yocalebo (approved):

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -1,22 +1,62 @@
...
+            x_real_remote_port = int(app.request.headers["X-Real-Remote-Port"])
+        except (KeyError, ValueError):
+            pass
+        else:
+            if process := get_peer_process(remote_addr, remote_port):
+                if process.name() == "nginx":
+                    try:
+                        with open("/var/run/nginx.pid") as f:
+                            nginx_pid = int(f.read().strip())
+                    except Exception:
```
yocalebo, line 45: Actually, if getting `nginx_pid` or `ppid` raises an Exception, the check `if ppid == nginx_pid` will raise an exception because those 2 variables won't be defined.

yocalebo (changes requested):

yocalebo (approved):

themylogin:
backport

PR #7641, opened 2021-10-06
NAS-112699 / 12.0 / fix pool.get_disks

yocalebo:
This was rather ridiculous.

1. `pool.get_disks` calls `pool.get_reserved`
2. `pool.get_reserved` calls an async generator function `zfs.pool.get_disks` (this creates a child process)
3. `zfs.pool.get_disks` calls `zfs.pool.get_devices` which spawns a grandchild

Each child and grandchild has to get a copy of that xml data from the parent process which is particularly painful if there are a lot of disks.

Fix this by doing 2 major things:
1. use `with libzfs.ZFS()` directly in this method to not spin up another child process since this method is already in its own child process
2. use the `disk.label_to_dev_disk_cache` method which is a manageable dict of geom xml disk information. Generate this once in the parent and then iterate normally.

The speed difference is quite dramatic.
```
OLD:
midclt call zfs.pool.get_disks sixshelves01 > /dev/null  0.12s user 0.02s system 1% cpu 8.115 total

NEW:
midclt call zfs.pool.get_disks sixshelves01 > /dev/null  0.12s user 0.03s system 35% cpu 0.415 total
```

### DIFF
--- a/src/middlewared/middlewared/plugins/zfs_/disks_freebsd.py
+++ b/src/middlewared/middlewared/plugins/zfs_/disks_freebsd.py
@@ -1,7 +1,5 @@
-from bsd import geom
-
+from libzfs import ZFS
 from middlewared.service import Service
-
 from .disks_base import PoolDiskServiceBase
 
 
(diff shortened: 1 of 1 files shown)


rick-mesta (approved):
lgtm

PR #7662, opened 2021-10-12
NAS-112817 / 22.02-RC.1 / prevent "wg" from being modified

yocalebo:
The `wg0` interface used by TrueCommand gets deleted any time a network change is made. That's no bueno. This hasn't worked since the wireguard feature was added, as far as I can tell.


william-gr:
backport

PR #7712, opened 2021-10-18
NAS-112768 / 22.02-RC.2 / Properly show UPS driver choices

sonicaj:

1) Normalize ups driver choice properly when some other options are specified in driver string tab entry. This was not happening before as we weren't accounting for custom UPS options to be specified in the string.
2) If there are multiple entries against a single key, this PR adds changes to make sure we display all of those entries for user clarity.


wongsyrone:
@sonicaj @themylogin 

Please re-check this PR, it's incorrect fundamentally.

PR #7726, opened 2021-10-20
NAS-112351 / 12.0 / Retrieve data for all non-ONLINE disks in pool status alert

themylogin:
Besides being UNAVAIL, they can also be FAULTED, etc.


themylogin:
time 0:15

PR #7778, opened 2021-10-28
NAS-113096 / 23.10 / Add tests for LDAP + GSSAPI

anodos325:
Add regression tests for using kerberos keytabs with
LDAP plugin. Since CI infrastrucutre already has
Active Directory deployed, join the AD domain first through
libads to generate a keytab and account in AD, then
stop the AD service and use the machine account keytab for
authentication through the LDAP plugin. This provides
coverage for LDAP GSSAPI binds, and general correctness
of behavior for generating krb5.conf and keytab when
LDAP plugin is used.

### DIFF
--- a/src/middlewared/middlewared/etc_files/pam.d/pam.inc.mako
+++ b/src/middlewared/middlewared/etc_files/pam.d/pam.inc.mako
@@ -166,22 +166,22 @@
 
         def pam_account(self):
             min_uid = self.min_uid()
-            ldap_args = [
-                "try_first_pass",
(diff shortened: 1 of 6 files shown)


anodos325:
Currently fails in jenkins, will revisit later.

rick-mesta:
@anodos325: Please re-target this PR to `scale/angelfish` 🙏🏼

yocalebo:
@anodos325 ping 😄

yocalebo (approved):

PR #7798, opened 2021-10-29
NAS-113134 / 22.02-RC.2 / Allow call("pool.query", [], {"extra": {"is_upgraded": true}}) to avo…

themylogin:
…id UI calling pool.is_upgraded in a loop

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -267,6 +267,7 @@ class PoolService(CRUDService):
     class Config:
         datastore = 'storage.volume'
         datastore_extend = 'pool.pool_extend'
+        datastore_extend_context = 'pool.pool_extend_context'
         datastore_prefix = 'vol_'
         event_send = False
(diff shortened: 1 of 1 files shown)


themylogin:
time 1:00

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -301,6 +302,7 @@ class Config:
                 'total_secs_left': None,
             }
         ),
+        Bool('is_upgraded'),
```
yocalebo, line 305: Does this default to `True`? If so, does this mean that anytime `pool.query` is called, we `subprocess` for every pool on the system?

themylogin: This is `ENTRY` declaration, a schema for a returned pool object, here I declare that `is_upgraded` is boolean and is not required (might be either present or not). We only call subprocess `if context['extra'].get('is_upgraded')` and `extra` is default to an empty dict.

yocalebo (approved):

PR #7800, opened 2021-10-29
Add check for SMB service being initialized

anodos325:
When unlocking dataset or importing pool, verify
that SMB service has been properly initialized
before calling hook. If service hasn't been
initialized (most likely due to bug in system
dataset handling) then log a warning message and
turn this into a no-op.


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -564,6 +570,7 @@ async def configure(self, job, create_paths=True):
         This step is not required when underlying database is clustered (cluster node should
         just recover with info from other nodes on reboot).
         """
+        self.configured = True
```
yocalebo, line 573: You're going to want to set the class attribute here. (i.e. instead of `self.configured`, do `SMBService.configured`) so that all instantiations of this class get the same value for that attribute. You can see past tribulations that I encountered doing something similar in ticket NAS-109597. Also, when is this ever set back to `False`?

yocalebo (changes requested):

rick-mesta:
@anodos325: Please re-target this PR to `scale/angelfish` 🙏🏼

yocalebo (approved):

PR #7804, opened 2021-11-01
NAS-113157 / 22.02-RC.2 / Fix ordering of crossrename and recycle vfs objects

anodos325:


anodos325:
vfs_recycle converts SMB_VFS_UNLINK to SMB_VFS_RENAME. vfs_crossrename converts SMB_VFS_RENAME into a copy op if rename fails due to crossing device boundaries. This means that we need to ensure modules are stacked in order of recycle then crossrename.

yocalebo (approved):

PR #7834, opened 2021-11-08
NAS-111019 / 22.02-RC.2 / Allow Kubernetes ServiceLB to be turned off

Ornias1993:
NAS-111019 originally aims to add MetalLB by iX systems.

**However, I've come across a few caveats:**
- It would limit users to still only be able to use either of the 2 loadbalancers
- MetalLB also includes a lot of options that are not likely to be included (such as prometheus metrics support), where users actually do want to customise their setup

**Considering:**
- Loadbalancers are also easily rolled out as Apps
- We already have a solid solution for Apps
- iX already leaves it to the communityto build Ingresses as Apps as well

I've come to the conclusion that the simplest and cleanest solution would be to just allow users to turn off the  ServiceLB "Klipper" LoadBalancer. I've done initial research and this can cleanly be done, without complete reinstallation of k3s. (aka: Hotswap is fully supported).

This PR Adds a simple hook in the middleware to disable it. The PR aims to be as simple as possible, to prevent needless overcomplication.

**Possible Bugs/Risks/Problems**
- It basically only tells k3s to not install (or remove) the LoadBalancer deployment. The default loadbalancer works mostly by adding seperate pods to each HelmChart/Service, which get removed and that's (basically) it. It's also used by a LOT of k3s users worldwide. Hence the chance this will cause _technical_ issues is practically zero, if implemented correctly.
- As this is just the middleware "backend" portion, the chances of this causing a significant support requirement, are also quite minimal. Users using this as-is (by manually toggling the flag), would not be many and relatively experienced.
- Even if this would cause an issue, the issue would be limited to there not being a loadbalancer deployed, which would not affect the official Apps at all, as those do not use the loadbalancer type service.

**Consequences for the End user**
As requested by @sonicaj here a short list of what happens to the end-user when they disable the build-in LoadBalancer:
- `SVCLB` pods will not spawn
- Services with type "LoadBalancer", will not automatically get an external IP/Port assigned:
image
- Users can still manually set an ExternalIP on a service with type "LoadBalancer" to open on the host:
image
(this will NOT open that port cluster wide, like ServiceLB does, only on the local node like NodePort)
- Users will still be able to install Apps
- Users will still be able to connect to Apps internally

**Performed tests:**
- Manually checked if the add "disable" list in config.yaml would, in fact, disable the loadbalancer (Correct)
- Checked if the added Python scripting does, in fact, result in the loadbalancer being disabled in config.yaml
- Created a UI PR (https://github.com/truenas/webui/pull/6104) and tested if that did, in fact, cause the loadbalancer to be disabled in config.yaml

In all cases where the loadbalancer is disabled, it, in fact, does not spawn any svclb pods as expected:
image

In case anyone wants to test how this would work, an example MetalLB App is available in the TrueCharts Incubator Train:
image

image

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
@@ -25,6 +25,7 @@ class KubernetesModel(sa.Model):
     node_ip = sa.Column(sa.String(128), default='0.0.0.0')
     cni_config = sa.Column(sa.JSON(type=dict), default={})
     configure_gpus = sa.Column(sa.Boolean(), default=True, nullable=False)
+    servicelb = sa.Column(sa.Boolean(), default=True, nullable=False)
 
 
 class KubernetesService(ConfigService):
@@ -36,6 +37,7 @@ class KubernetesService(ConfigService):
 
     ENTRY = Dict(
         'kubernetes_entry',
+        Bool('servicelb', required=True),
         Bool('configure_gpus', required=True),
         Str('pool', required=True, null=True),
         IPAddr('cluster_cidr', required=True, cidr=True, empty=True),
@@ -223,6 +225,12 @@ class KubernetesService(ConfigService):
         """
         `pool` must be a valid ZFS pool configured in the system. Kubernetes service will initialise the pool by
         creating datasets under `pool_name/ix-applications`.
+
+        `configure_gpus` is a boolean to enable or disable to prevent automatically loading any GPU Support
+        into kubernetes. This includes not loading any daemonsets for Intel and NVIDIA support.
+
+        `servicelb` is a boolean to enable or disable the integrated k3s Service Loadbalancer called "Klipper".
+        This can be set to disabled to enable the user to run another LoadBalancer or no LoadBalancer at all.
 
         `cluster_cidr` is the CIDR to be used for default NAT network between workloads.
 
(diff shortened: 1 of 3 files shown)


Ornias1993:
Added the required Alembic migration as well, please check and correct if needed! :)

Ornias1993:
This PR has now been verified to add the disable-toggle with above linked webui PR

src/middlewared/middlewared/plugins/kubernetes_linux/update.py
```diff
@@ -36,6 +37,7 @@ class Config:
 
     ENTRY = Dict(
         'kubernetes_entry',
+        Bool('servicelb', required=True),
```
sonicaj, line 40: Can we please add some documentation about it as well below in `do_update` doc string ?

sonicaj (commented):
LGTM :) Just one minor nit

src/middlewared/middlewared/plugins/kubernetes_linux/update.py
Ornias1993: Thats a fair point, it's not a self-explainatory variable for people that are not into k3s 👍

Ornias1993:
```suggestion
        # Allows for easy disabling of the k3s integrated Service LoadBalancer called "Klipper"
        Bool('servicelb', required=True),
```

sonicaj (approved):
Thanks!

PR #7835, opened 2021-11-09
NAS-113244 / 22.02-RC.2 / Do not crash disk.sync_all when handling CORE identifiers not supported by SCALE `disk.identifier_to_device`

themylogin:
Some of the legacy CORE disk identifiers (e.g. `{label}xxx`) are not supported by SCALE. This leads to
`disk.sync_all` failure because we still try to decode them.

While we are here, remove all references to the `ctld` which was the only other consumer of
`disk.identifier_to_device`.


themylogin:
time 1:00

sonicaj (approved):

PR #7838, opened 2021-11-09
NAS-113272 / 22.02-RC.2 / Do not repeatedly log same error

sonicaj:
This commit adds changes to not repeatedly log an error if we failed to retrieve rotational rate for some device.

### DIFF
--- a/src/middlewared/middlewared/plugins/device_/device_info_linux.py
+++ b/src/middlewared/middlewared/plugins/device_/device_info_linux.py
@@ -22,6 +22,7 @@ RE_UART_TYPE = re.compile(r'is a\s*(\w+)')
 
 class DeviceService(Service, DeviceInfoBase):
 
+    DISK_ROTATION_ERROR_LOG_CACHE = set()
     HOST_TYPE = None
 
     def get_serials(self):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -99,13 +100,17 @@ def get_disk(self, name):
...
             disk = libsgio.SCSIDevice(device_path)
             rotation_rate = disk.rotation_rate()
         except (OSError, RuntimeError):
-            self.logger.error('Ioctl failed while retrieving rotational rate for disk %s', device_path)
+            if device_path not in self.DISK_ROTATION_LOG_CACHE:
+                self.DISK_ROTATION_LOG_CACHE.add(device_path)
+                self.logger.error('Ioctl failed while retrieving rotational rate for disk %s', device_path)
             return
+        else:
+            if device_path in self.DISK_ROTATION_LOG_CACHE:
```
themylogin, line 112: Prone to race condition, just `discard` it

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -22,6 +22,7 @@
 
 class DeviceService(Service, DeviceInfoBase):
 
+    DISK_ROTATION_LOG_CACHE = set()
```
themylogin, line 25: `DISK_ROTATION_ERROR_LOG_CACHE` so we see that this is related to the error handling?

themylogin (changes requested):

themylogin (approved):

yocalebo (approved):

PR #7842, opened 2021-11-10
NAS-113034 / 22.02-RC.2 / prevent some db queries on passive

yocalebo:
We do not need to add these database operations to the `JournalSync` thread's queue to synchronize to the standby controller.

### DIFF
--- a/src/middlewared/middlewared/plugins/keyvalue.py
+++ b/src/middlewared/middlewared/plugins/keyvalue.py
@@ -50,9 +50,9 @@ class KeyValueService(Service):
                 "datastore.insert", "system.keyvalue", {"key": key, "value": json.dumps(value)}, options
             )
         else:
-            await self.middleware.call("datastore.update", "system.keyvalue", row["id"], {
-                "value": json.dumps(value)
-            })
+            await self.middleware.call(
+                "datastore.update", "system.keyvalue", row["id"], {"value": json.dumps(value)}, options
+            )
 
         return value
 
--- a/src/middlewared/middlewared/plugins/keyvalue.py
+++ b/src/middlewared/middlewared/plugins/keyvalue.py
@@ -1,5 +1,5 @@
 from middlewared.client import ejson as json
-from middlewared.schema import Any, Str, accepts
+from middlewared.schema import Any, Str, accepts, Dict
 from middlewared.service import Service
 import middlewared.sqlalchemy as sa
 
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/migration.py
```diff
@@ -50,6 +50,6 @@ async def run(self):
                     self.middleware.logger.error("Error running migration %s", name, exc_info=True)
                     continue
 
-                await self.middleware.call("datastore.insert", "system.migration", {"name": name})
+                await self.middleware.call("datastore.insert", "system.migration", {"name": name}, {"ha_sync": False})
 
             await self.middleware.call("keyvalue.set", "run_migration", False)
```
themylogin, line 55: Also need to change it here

src/middlewared/middlewared/plugins/failover_/journal.py
```diff
@@ -118,7 +118,8 @@ def _handle_sql_queue_item(self, item):
                 self.journal.append(item)
             else:
                 query, params = item
-                logger.warning('Node status %s but executed SQL query: %s', self.failover_status, query)
+                if query and query.find('services_ssh') == -1:
```
themylogin, line 121: Can't you just use `ha_sync: false` for the update host keys query?

themylogin (changes requested):

src/middlewared/middlewared/plugins/failover_/journal.py
yocalebo: I've reverted that change and will add it to my open PR here: https://github.com/truenas/middleware/pull/7843

src/middlewared/middlewared/plugins/keyvalue.py
```diff
@@ -37,22 +37,28 @@ async def get(self, key, default):
...
         except IndexError:
-            await self.middleware.call("datastore.insert", "system.keyvalue", {
-                "key": key,
-                "value": json.dumps(value)
-            })
+            await self.middleware.call(
+                "datastore.insert", "system.keyvalue", {"key": key, "value": json.dumps(value)}, options
+            )
         else:
             await self.middleware.call("datastore.update", "system.keyvalue", row["id"], {
```
themylogin, line 53: Also need to pass options here

themylogin (changes requested):

src/middlewared/middlewared/plugins/keyvalue.py
yocalebo: oops, thanks

themylogin (approved):

PR #7845, opened 2021-11-10
NAS-113286 / 22.02-RC.2 / mark primary interface on active-backup bonds

yocalebo:
When an `active-backup` type bond is created on SCALE, we're not setting the `primary` interface setting. This isn't a problem because the kernel will automatically choose one, however, it's very common to want to choose which member of the `active-backup` bond is the "primary" underlying interface to be used.

This changes the interface that has a `lagg_ordernum` of 0 to be the "primary" interface which means the order in which the members are specified when creating this bond type are important.

### DIFF
--- a/src/middlewared/middlewared/plugins/interface/lag.py
+++ b/src/middlewared/middlewared/plugins/interface/lag.py
@@ -35,10 +35,16 @@ class InterfaceService(Service):
             self.middleware.call_sync('interface.disable_capabilities', name)
         """
 
-        info = {'protocol': None, 'xmit_hash_policy': None, 'lacpdu_rate': None}
+        info = {'protocol': None, 'xmit_hash_policy': None, 'lacpdu_rate': None, 'primary_interface': None}
         protocol = getattr(netif.AggregationProtocol, lagg['lagg_protocol'].upper())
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/interface/lag.py
```diff
@@ -76,6 +82,10 @@ def lag_setup(self, lagg, members, disable_capabilities, parent_interfaces, sync
                 self.logger.info('Changing lacpdu_rate on %r to %s', name, info['lacpdu_rate'])
                 iface.lacpdu_rate = info['lacpdu_rate']
 
+            if info['primary_interface'] is not None:
+                self.logger.info('Changing primary interface on %r to %s', name, info['primary_interface'])
+                iface.primary_interface = info['primary_interface']
```
themylogin, line 87: Do we need to check if `iface.primary_interface` is already the correct one and in this case do nothing like we do this with `lacpdu_rate`?

yocalebo: I'm doing that on lines 43 - 48 up above?

themylogin (approved):

PR #7855, opened 2021-11-11
NAS-113310 / 12.0 / Work around flaky NIS domains

anodos325:
The kernel yp client will try up to 200 seconds currently for various yp-related requests. This PR makes three significant changes related to the NIS service:

1) get list of NIS users/groups through `ypcat` subprocess before using the python pwd and grp modules to get info for webui and middleware cache. This avoids enumerating users and groups through the aforementioned modules.

2) alter nsswitch.conf generation so that `nis` is only added if the NIS service is enabled and status is not FAULTED.

3) alter NIS health check so that it regenerates the nsswitch.conf on each run (once every ten minutes).

### DIFF
--- a/src/middlewared/middlewared/plugins/nis.py
+++ b/src/middlewared/middlewared/plugins/nis.py
@@ -299,6 +299,6 @@ class NISService(ConfigService):
         if not await self.middleware.call('cache.has_key', 'NIS_cache'):
             await self.middleware.call('nis.fill_cache')
             self.logger.debug('cache fill is in progress.')
-            return {'users': [], 'groups': []}
+            return {'users': {}, 'groups': {}}
 
         return await self.middleware.call('cache.get', 'NIS_cache')
--- a/src/middlewared/middlewared/plugins/nis.py
+++ b/src/middlewared/middlewared/plugins/nis.py
@@ -274,7 +274,7 @@ class NISService(ConfigService):
 
         for nis_group in nis_groups:
             if nis_group in local_groups:
-                self.logger.warning("%s: name is also a local a local group. Omitting from group cache", nis_group)
+                self.logger.warning("%s: name is also a local group. Omitting from group cache", nis_group)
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/nis.py
```diff
@@ -216,18 +234,19 @@ def fill_cache(self, job, force=False):
...
+        local_users = list(u['username'] for u in self.middleware.call_sync('user.query'))
+        local_groups = list(g['group'] for g in self.middleware.call_sync('group.query'))
         cache_data = {'users': [], 'groups': []}
 
-        for u in pwd_list:
-            is_local_user = True if u.pw_uid in local_uid_list else False
-            if is_local_user:
+        for nis_user in nis_users:
+            if nis_user in local_users:
+                self.logger.warning("%s: name is also a local a local user. Omitting from user cache", nis_user)
```
yocalebo, line 246:
```suggestion
                self.logger.warning("%s: name is also a local user. Omitting from user cache", nis_user)
```

src/middlewared/middlewared/plugins/nis.py
```diff
@@ -253,11 +272,12 @@ def fill_cache(self, job, force=False):
             }})
             user_next_index += 1
 
-        for g in grp_list:
-            is_local_user = True if g.gr_gid in local_gid_list else False
-            if is_local_user:
+        for nis_group in nis_groups:
+            if nis_group in local_groups:
+                self.logger.warning("%s: name is also a local a local group. Omitting from group cache", nis_group)
```
yocalebo, line 277:
```suggestion
                self.logger.warning("%s: name is also a local group. Omitting from group cache", nis_group)
```

rick-mesta (approved):
Other than the changes @yocalebo found, I'm good 👍🏼

yocalebo (approved):

PR #7861, opened 2021-11-12
NAS-113308 / 22.02-RC.2 / fix race with netif.create_interface and netif.list_interfaces

yocalebo:
Because we're using `subprocess` to create and list interfaces there exists a potential for a race between the time we create an interface and the time we try to retrieve that interface. To make matters worse, if an empty vlan is created and its parent is a physical interface then on boot when we create the vlan interface we will get a `KeyError` when we try to retrieve that interface which prevents the parent interface from ever coming up. If other interfaces are dependent on that physical interface to come up, then the entire network is down until someone manually brings that physical interface up. To fix all of these scenarios I've made it so that all virtual interfaces (`vlan`, `bond`, `br`) get created using `pyroute2.NDB` which actually monitors and waits for state synchronization. This means the `.commit()` will not return until the interface has actually been created and all necessary parameters have been applied.

I also change `list_interfaces` to use `pyroute2.NDB` as well.

### DIFF
--- a/src/middlewared/middlewared/plugins/interface/netif_linux/bridge.py
+++ b/src/middlewared/middlewared/plugins/interface/netif_linux/bridge.py
@@ -1,8 +1,6 @@
-# -*- coding=utf-8 -*-
 import json
 import logging
-
-import middlewared.plugins.interface.netif_linux.interface as interface
+from pyroute2 import NDB
 
 from .utils import run
 
@@ -12,12 +10,8 @@ __all__ = ["create_bridge", "BridgeMixin"]
 
 
 def create_bridge(name):
-    cmd = [
-        "ip", "link", "add", name, "type", "bridge",
-        "stp_state", "1"  # enable stp by default 1 == on 0 == off
-    ]
-    run(cmd)
-    interface.Interface(name).up()
+    with NDB(log='off') as ndb:
+        ndb.interfaces.create(ifname=name, kind="bridge").set("br_stp_state", 1).set("state", "up").commit()
 
 
 class BridgeMixin:
--- a/src/middlewared/middlewared/plugins/interface/netif_linux/bridge.py
+++ b/src/middlewared/middlewared/plugins/interface/netif_linux/bridge.py
@@ -10,7 +10,7 @@ __all__ = ["create_bridge", "BridgeMixin"]
 
 
 def create_bridge(name):
-    with NDB(log='off') as ndb:
+    with NDB(log="off") as ndb:
         ndb.interfaces.create(ifname=name, kind="bridge").set("br_stp_state", 1).set("state", "up").commit()
 
 
--- a/src/middlewared/middlewared/plugins/interface/netif_linux/netif.py
+++ b/src/middlewared/middlewared/plugins/interface/netif_linux/netif.py
@@ -1,5 +1,4 @@
 import logging
-import os
 from pyroute2 import NDB
 
 from .bridge import create_bridge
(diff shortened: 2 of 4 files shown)


src/middlewared/middlewared/plugins/interface/netif_linux/bridge.py
```diff
@@ -1,23 +1,21 @@
...
 def create_bridge(name):
-    cmd = [
-        "ip", "link", "add", name, "type", "bridge",
-        "stp_state", "1"  # enable stp by default 1 == on 0 == off
-    ]
-    run(cmd)
-    interface.Interface(name).up()
+    ndb = None
+    try:
+        ndb = NDB(log='off')
```
themylogin, line 15: You can just move this line away from try/catch block and remove `if` condition in `finally` block. Also let's use consistent single/double quotes in a file

src/middlewared/middlewared/plugins/interface/netif_linux/netif.py
```diff
@@ -38,6 +37,5 @@ def get_interface(name):
 
 
 def list_interfaces():
-    return {name: Interface(name)
-            for name in os.listdir("/sys/class/net")
-            if os.path.isdir(os.path.join("/sys/class/net", name))}
+    with NDB(log='off') as ndb:
```
themylogin, line 40: What's the difference between `with` and `close` in the files above?

themylogin (changes requested):

src/middlewared/middlewared/plugins/interface/netif_linux/vlan.py
```diff
@@ -1,40 +1,32 @@
...
     @property
     def parent(self):
-        return os.path.basename(os.readlink(glob.glob(f"/sys/devices/virtual/net/{self.name}/lower_*")[0]))
+        return basename(readlink(glob(f"/sys/devices/virtual/net/{self.name}/lower_*")[0]))
 
     @property
     def tag(self):
         with open(f"/proc/net/vlan/{self.name}") as f:
-            return int(re.search(r"VID: ([0-9]+)", f.read()).group(1))
+            return int(search(r"VID: ([0-9]+)", f.read()).group(1))
```
themylogin, line 29: What's the point of all these changes?

yocalebo: no point in importing the entire module when only using singular functions.

themylogin: @yocalebo also no point in introducing unnecessary diffs, no point in confusing developers (what is `search`?), no point in making us change `import` line every time we want to use a new function from common library

"import os" - 24,732,052 code results - https://github.com/search?q=%22import+os%22&type=code
"from os import" - 218,819 code results - https://github.com/search?q=%22from+os+import%22&type=code

themylogin (changes requested):

src/middlewared/middlewared/plugins/interface/netif_linux/vlan.py
yocalebo: If a developer gets confused because they cant see that the `search` wasn't imported from `re` then we have bigger problems. Furthermore, linking a github search is relevant in what way exactly? Why does it matter what other people do? Our entire middlewared code base does things that aren't "standard" so those queries are superfluous.

yocalebo: But alas, I'm not really interested in arguing with you on what perceive to be the "right" way of doing things so I've removed the "unnecessary diffs".

themylogin (approved):
Thank you!

PR #7868, opened 2021-11-15
NAS-113352 / 12.0 / Make sure SED disks are unlocked on HA systems

sonicaj:
This commit adds changes to make sure SED disks are unlocked on HA systems. With the changes introduced, we make sure that we always try to unlock SED disks when backup becomes master node eliminating any race conditions which might come synchronizing if master has already unlocked sed disks or not. Secondly if user changes sed disks password and initiates a failover assuming the system would unlock SED disks, it would not happen as we don't try to unlock sed disks when backup becomes master.


yocalebo (approved):
Thank you!

rick-mesta (approved):

yocalebo:
backport

yocalebo:
backport

PR #7879, opened 2021-11-18
NAS-113395 / 12.0 / add nvme mapping support to R50B

yocalebo:


darkfiberiru (changes requested):
I don't see bits for using dmidecode designations along with dbsf from sysctl %location to map on R50B  instead of handle like R50 uses.

- 
- #1 Designation: PCH SLOT1 PCI-E 3.0 X4(IN X8)
- 
- #2 Designation: CPU SLOT4 PCI-E 3.0 X16
- 
- #3 Blank entry to allow us to use same enclosure map as R50(this drive is not touched by webui)
-

src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/r50_nvme.py
```diff
@@ -18,34 +15,41 @@ class EnclosureService(Service):
...
         slot_to_nvd = {}
-        for nvme, nvd in nvme_to_nvd.items():
+        for nvme, nvd in self.middleware.call_sync('disk.nvme_to_nvd_map').items():
             try:
                 location = sysctl.filter(f"dev.nvme.{nvme}.%location")[0].value
                 m = re.search(self.RE_HANDLE, location)
-                if not m:
-                    continue
-
-                handle = m.group(1)
```
darkfiberiru, line 34: This should be slot 0 and slot 2 should be the empty one. I believe based upon output

yocalebo:
> I don't see bits for using dmidecode designations along with dbsf from sysctl %location to map on R50B instead of handle like R50 uses.
> 
> * #1 Designation: PCH SLOT1 PCI-E 3.0 X4(IN X8)
> * #2 Designation: CPU SLOT4 PCI-E 3.0 X16
> * #3 Blank entry to allow us to use same enclosure map as R50(this drive is not touched by webui)

dmidecode shouldn't be needed. Just need to account for the fact that the nvme directly on CPU doesn' t have a `handle=` entry.

darkfiberiru (approved):
Caleb sorry I see you managed to avoid touching dmidecode.

I agree that for these systems this is enough to be consistent in all variant (network packs etc) of r50b

PR #7887, opened 2021-11-19
NAS-113426 / 22.12 / remove unneeded ha_propagate service option

yocalebo:
Take the following example:

1. node A becomes passive
2. node B becomes active and calls systemdataset.setup()
3. systemdataset.setup() calls service.stop and service.start for numerous services
4. node B forwards this request to node A because ha_propagate is True by default
5. at same time node A is still becoming passive
6. node B spams a whole bunch of CallTimeout errors (because node A is becoming passive OR rebooting by design)

Maintaining the "state" between the controllers is done completely via the system dataset plugin. The service management is completely done by the failover event plugin.  No need to have this extra layer of abstraction.

### DIFF
--- a/src/middlewared/middlewared/plugins/failover.py
+++ b/src/middlewared/middlewared/plugins/failover.py
@@ -1340,35 +1340,6 @@ async def hook_kmip_sync(middleware, *args, **kwargs):
     await middleware.call('failover.sync_keys_to_remote_node')
 
 
-async def service_remote(middleware, service, verb, options):
-    """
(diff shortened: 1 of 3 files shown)


themylogin:
How are things like `network.configuration.update '{"hostname": "xxx"}' are propagated now? What will actually set the hostname on the BACKUP now? It was previously done by propagating `hostname` service restart (I don't like this at all, but that's how it works now) or is there some other mechanism?

rick-mesta:
@yocalebo: Please re-target this PR to `scale/angelfish` 🙏🏼

PR #7890, opened 2021-11-20
NAS-113432 / 22.02-RC.2 / ignore FileNotFoundError in vrrp fifo thread

yocalebo:


themylogin (changes requested):
Can we use `errno.ENOENT` to avoid magic constants?

yocalebo:
I don't believe this is a problem anymore because of some other SCALE HA related changes that I've made. I'll close this one.

PR #7891, opened 2021-11-20
NAS-113433 / 22.02 / ignore CallError in get_remote_os_version

yocalebo:
`RemoteClient.call` already does the hard work of catching the expected exceptions on the p2p connection. This ignores `CallError` in `get_remote_os_version` and logs an exception for anything else.

### DIFF
--- a/src/middlewared/middlewared/plugins/failover_/remote.py
+++ b/src/middlewared/middlewared/plugins/failover_/remote.py
@@ -192,6 +192,9 @@ class RemoteClient(object):
         if self.client is not None and self._remote_os_version is None:
             try:
                 self._remote_os_version = self.client.call('system.version')
+            except CallError:
+                # ignore CallErrors since they're being caught in self.client.call
+                pass
             except Exception:
                 logger.error('Failed to determine OS version', exc_info=True)
 


src/middlewared/middlewared/plugins/failover_/remote.py
```diff
@@ -192,8 +192,12 @@ def get_remote_os_version(self):
         if self.client is not None and self._remote_os_version is None:
             try:
                 self._remote_os_version = self.client.call('system.version')
-            except Exception:
-                logger.error('Failed to determine OS version', exc_info=True)
+            except Exception as e:
+                if "'NoneType' object has no attribute 'text_message'" in str(e):
```
themylogin, line 196: Won't it be better to handle this in the middleware client itself and reraise a specific exception?

yocalebo: Just getting back to this and the RemoteClient was already doing the hard-work of catching these specific exceptions so I just need to do the same in `get_remote_os_version`. I've made the correct changes in latest PR.

themylogin (approved):

PR #7897, opened 2021-11-22
NAS-113275 / 22.02-RC.2 / Exclude internal interfaces from avahi config

anodos325:
Binding avahi to kubernetes bridge and other internal interfaces
can cause erratic mDNS behavior.

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/avahi/avahi-daemon.conf.mako
+++ b/src/middlewared/middlewared/etc_files/local/avahi/avahi-daemon.conf.mako
@@ -21,6 +21,10 @@
     deny_interfaces = middleware.call_sync("interface.internal_interfaces")
     failover_int = middleware.call_sync("failover.internal_interfaces")
     deny_interfaces.extend(failover_int)
+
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/etc_files/local/avahi/avahi-daemon.conf.mako
```diff
@@ -18,10 +18,7 @@
 
     ipv6_enabled = any(middleware.call_sync('interface.ip_in_use', {'ipv4': False, 'ipv6': True}))
 
-    failover_int = middleware.call_sync("failover.internal_interfaces")
```
sonicaj, line 21: I think we would still want this, @yocalebo can you confirm please ?

anodos325: I also need to bring in some changes from CORE/Enterprise to raise FileShouldNotExist() on passive.

yocalebo (approved):
Thanks, lgtm!

PR #7913, opened 2021-11-25
NAS-113487 / 22.02-RC.2 / Have humanized app version for history of chart releases

sonicaj:
This commit adds changes to add another key to have humanized app version for each history version of a chart release so that we can expose this in the UI when user wants to rollback to a previous version of the chart release.


Qubad786 (approved):
LGTM!

PR #7914, opened 2021-11-26
NAS-113447 / 22.02-RC.2 / Do not reset mountpoints of ix-applications dataset when importing pool

sonicaj:
This commit adds changes to not reset mountpoints of ix-applications dataset as that's a dataset managed by us and if we reset mountpoints, PVC's require mountpoint of legacy as otherwise the apps consuming them won't deploy because of CSI not being able to mount the PVC.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -1512,17 +1512,26 @@ class PoolService(CRUDService):
                 },
             })
 
-            try:
-                # Reset all mountpoints
-                await self.middleware.call('zfs.dataset.inherit', pool_name, 'mountpoint', True)
-            except Exception as e:
-                # Let's not make this fatal
-                self.middleware.logger.debug(
-                    f'Failed to inherit mountpoints recursively for imported {pool_name} pool: {e}'
-                )
-
-            if osc.IS_FREEBSD:
-                await self.middleware.call('pool.sync_encrypted', pool_id)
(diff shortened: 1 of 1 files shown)


themylogin (approved):

themylogin:
Some people back up ix-applications dataset using zettarepl. We skip `mountpoint` property. What will happen if we receive such a dataset and try to use it?

Ornias1993:
> Some people back up ix-applications dataset using zettarepl. We skip `mountpoint` property. What will happen if we receive such a dataset and try to use it?

That's a fair point indeed.
Whereas the ix-applications, backups, docker, catalog and release datasets themselves need to actually always be mounted, the PVC's should not be mounted.

Afaik the default behavior would be to mount on zfs recv, if no mountpoint is set.
Which requires a short oneliner after backup restore, assuming default (auto generated) PVC dataset naming:
`zfs set mountpoint=legacy $(zfs list -o name | grep ix-applications | grep pvc)`

sonicaj:
@Ornias1993 yes indeed. We are going to address that in another PR where we do a best effort to identify pvc datasets and set their mountpoints appropriately once a pool is selected where ix-applications dataset already exists.

Ornias1993:
> @Ornias1993 yes indeed. We are going to address that in another PR where we do a best effort to identify pvc datasets and set their mountpoints appropriately once a pool is selected where ix-applications dataset already exists.

Awesome work @sonicaj 
It's getting there, the last 1% of polish is always the most annoying... 👍

PR #7915, opened 2021-11-28
NAS-113282 / 22.02-RC.2 / Do not allow writing to locked dataset path

sonicaj:
This PR adds changes to not allow writing to the path of a locked dataset by recursively setting immutable flag on that path to make sure the path is not modified in any way.

### DIFF
--- a/src/middlewared/middlewared/plugins/filesystem.py
+++ b/src/middlewared/middlewared/plugins/filesystem.py
@@ -9,18 +9,26 @@ import pyinotify
 import pathlib
 
 from middlewared.event import EventSource
-from middlewared.schema import accepts, Bool, Dict, Float, Int, List, Ref, returns, Path, Str
-from middlewared.service import private, CallError, filterable_returns, Service, job
-from middlewared.utils import filter_list
 from middlewared.plugins.pwenc import PWENC_FILE_SECRET
 from middlewared.plugins.cluster_linux.utils import CTDBConfig, FuseConfig
 from middlewared.plugins.filesystem_ import stat_x
+from middlewared.schema import accepts, Bool, Dict, Float, Int, List, Ref, returns, Path, Str
+from middlewared.service import private, CallError, filterable_returns, Service, job
+from middlewared.utils import filter_list, run
 
 
 class FilesystemService(Service):
 
     class Config:
         cli_namespace = 'storage.filesystem'
+
+    @private
+    async def set_immutable(self, set_flag, path):
+        cp = await run(['chattr', f'{"+" if set_flag else "-"}i', '-V', path], check=False)
+        if cp.returncode:
+            raise CallError(
+                f'Unable to {"set" if set_flag else "remove"} immutable flag on at {path!r} path: {cp.stderr.decode()}'
+            )
 
     @private
     def resolve_cluster_path(self, path, ignore_ctdb=False):
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -2463,7 +2463,7 @@ class PoolDatasetService(CRUDService):
         Str('id'),
         Dict(
             'unlock_options',
-            Bool('force_unlock', default=False),
+            Bool('force', default=False),
             Bool('key_file', default=False),
             Bool('recursive', default=False),
             Bool('toggle_attachments', default=True),
@@ -2513,7 +2513,7 @@ class PoolDatasetService(CRUDService):
         In some cases it's possible that the provided key/passphrase is valid but the path where the dataset is
         supposed to be mounted after being unlocked already exists and is not empty. In this case, unlock operation
         would fail. This can be overridden by setting `unlock_options.datasets.X.force` boolean flag or by setting
-        `unlock_options.force_unlock` flag. When any of these flags are set, system will rename the existing
+        `unlock_options.force` flag. When any of these flags are set, system will rename the existing
         directory/file path where the dataset should be mounted resulting in successful unlock of the dataset.
         """
         verrors = ValidationErrors()
@@ -2535,7 +2535,7 @@ class PoolDatasetService(CRUDService):
                     f'Passphrase or key must be specified for {ds["name"]}'
                 )
 
-            if not options['force_unlock'] and not ds['force']:
+            if not options['force'] and not ds['force']:
                 if err := self.dataset_can_be_mounted(ds['name'], os.path.join('/mnt', ds['name'])):
                     verrors.add(f'unlock_options.datasets.{i}.force', err)
 
@@ -2672,7 +2672,7 @@ class PoolDatasetService(CRUDService):
         Dict(
             'encryption_root_summary_options',
             Bool('key_file', default=False),
-            Bool('force_unlock', default=False),
+            Bool('force', default=False),
             List(
                 'datasets', items=[
                     Dict(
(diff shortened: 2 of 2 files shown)


themylogin (changes requested):
Why don't you use attachments delegate mechanism to handle FTP?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2409,12 +2421,20 @@ async def detach(delegate):
...
                 }
             )
         finally:
             await self.middleware.call('cache.pop', 'about_to_lock_dataset')
 
+        # We would now like to remove the empty directory which would exist as the dataset had been unmounted
+        # However it's possible user might have had some data already under the same path and if that holds,
+        # we rename the directory before nuking it
+        if os.path.exists(ds['mountpoint']) and os.listdir(ds['mountpoint']):
+            shutil.move(ds['mountpoint'], f'{ds["mountpoint"]}-{str(uuid.uuid4())[:4]}-{datetime.now().isoformat()}')
```
themylogin, line 2434: Why in the name of God would we want to do that?

themylogin (changes requested):

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1603,6 +1603,12 @@ async def export(self, job, oid, options):
         be disabled before exporting the last zpool on the system.
         """
         pool = await self.get_instance(oid)
+        root_ds = await self.middleware.call('pool.dataset.get_instance', pool['name'])
+        if root_ds['locked'] and os.path.exists(root_ds['mountpoint']):
+            # We should be removing immutable flag in this case if the path exists
+            cp = await run(['chattr', '-i', '-RV', root_ds['mountpoint']], check=False)
```
themylogin, line 1609: Can we add a middleware method for adding/removing this flag?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2424,21 +2446,32 @@ async def detach(delegate):
         finally:
             await self.middleware.call('cache.pop', 'about_to_lock_dataset')
 
+        job.set_progress(80, f'Setting immutable flag on unmounted {id!r} dataset path')
```
themylogin, line 2449: Do we need this in the status update, is this time-consuming operation?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2424,21 +2446,32 @@ async def detach(delegate):
...
+        cp = await run(['chattr', '+i', '-RV', ds['mountpoint']], check=False)
+        if cp.returncode:
+            raise CallError(
+                f'Unable to set immutable flag on unmounted {id!r} dataset\'s path '
+                f'({ds["mountpoint"]!r}): {cp.stderr.decode()}'
+            )
+
         await self.middleware.call_hook('dataset.post_lock', id)
+        job.set_progress(100, f'Successfully locked {id!r} dataset')
```
themylogin, line 2458: Is this necessary every time we successfully end a job to set its progress to 100% and invent another success message?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -2424,21 +2446,32 @@ async def detach(delegate):
...
 
         return True
 
     @accepts(
         Str('id'),
         Dict(
             'unlock_options',
+            Bool('force_unlock', default=False),
```
themylogin, line 2466: Why can't we have a consistent name here too? Just `force`

themylogin (changes requested):

src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -9,19 +9,31 @@
...
 
+    @private
+    async def add_immutable_flag_to_path(self, path):
+        cp = await run(['chattr', '+i', '-RV', path], check=False)
+        if cp.returncode:
+            raise CallError(f'Unable to set immutable flag on at {path!r} path: {cp.stderr.decode()}')
+
+    @private
+    async def remove_immutable_flag_to_path(self, path):
+        cp = await run(['chattr', '-i', '-RV', path], check=False)
```
yocalebo, line 33: Why subprocess when os module has chflags?

yocalebo (changes requested):

src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -9,19 +9,31 @@
...
         cli_namespace = 'storage.filesystem'
 
+    @private
+    async def add_immutable_flag_to_path(self, path):
+        cp = await run(['chattr', '+i', '-RV', path], check=False)
+        if cp.returncode:
+            raise CallError(f'Unable to set immutable flag on at {path!r} path: {cp.stderr.decode()}')
+
+    @private
+    async def remove_immutable_flag_to_path(self, path):
```
themylogin, line 32: Can't it be one method with boolean parameter?

themylogin (changes requested):

src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -9,19 +9,31 @@
...
 
 class FilesystemService(Service):
 
     class Config:
         cli_namespace = 'storage.filesystem'
 
+    @private
+    async def add_immutable_flag_to_path(self, path):
+        cp = await run(['chattr', '+i', '-RV', path], check=False)
+        if cp.returncode:
```
anodos325, line 28: Why chattr and not `os.chflags()`? 

Why are we doing this recursively?

sonicaj: We are doing this recursively to make sure that if there are any files in the directory, they cannot be modified as well.

anodos325: If it's a dataset mountpoint, shouldn't the dir be empty? I'm somewhat leery of recursive ops because we've been burned there before.

src/middlewared/middlewared/plugins/filesystem.py
sonicaj: Because we would like to do this recursively :)

themylogin: We don't need to do this recursively, it's an empty directory. If it is not empty for some reason (already contains user data due to some bug), we don't want to touch that.

src/middlewared/middlewared/plugins/filesystem.py
sonicaj: Ideally yes but that wouldn't be true for any writes which may already have been done. The idea is to make sure that the complete path is immutable.

src/middlewared/middlewared/plugins/filesystem.py
sonicaj: @themylogin so should we just make sure that we are not able to add/delete any files/folder in the mountpoint path ? Or should we also be making sure that any existing files there are not modified. Also nested directories contents can still be added/removed if it's not done recursively.

themylogin: @sonicaj that would be messing with user data, they might already have immutable attributes applied to some subset of files, we'll wipe all that. Having data in the mountpoint already is an emergency situation that will be resolved once the dataset is unlocked back and ideally should never happen again after we merge this PR. Doing this recursively might have sense, but I doubt this is really necessary.

src/middlewared/middlewared/plugins/filesystem.py
anodos325: > Ideally yes but that wouldn't be true for any writes which may already have been done. 
That sounds like a separate error condition. Perhaps on dataset lock / unmount, we check for whether stuff is inside that dir and generate an alert so that the user can take care of it.

yocalebo: If it's a non-root user, then writes to sub-files/dirs will fail if top-level dir is immutable. If it's a root account, they will work.

src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -9,19 +9,27 @@
...
+from middlewared.utils import filter_list, run
 
 
 class FilesystemService(Service):
 
     class Config:
         cli_namespace = 'storage.filesystem'
 
+    @private
+    async def update_immutable_flag_of_path(self, set_flag, path):
```
themylogin, line 26: `set` is shorter than `update` and `path` is redundant as we are already in filesystem plugin. `set_immutable` can be a better name.

This operation seems to be performed by `FS_IOC_SETFLAGS` ioctl, can you investigate it instead of using subprocess?

themylogin (changes requested):

themylogin (approved):

PR #7917, opened 2021-11-29
NAS-113452 / 22.02-RC.2 / correctly identify fstype in filesystem.statfs

yocalebo:
Strangely, python on linux doesn't have a native method for `statx` in the `os` module so I've added our own. Uses `ctypes` and is written so we can add a more involved wrapper around it (if needed) but the use case for this PR is as simple as possible since we only need to get the `stx_dev_major:stx_dev_minor` values for the given path and then find it in `/proc/self/mountinfo`.

### DIFF
--- a/src/middlewared/middlewared/plugins/filesystem.py
+++ b/src/middlewared/middlewared/plugins/filesystem.py
@@ -14,6 +14,7 @@ from middlewared.service import private, CallError, filterable_returns, Service,
 from middlewared.utils import filter_list
 from middlewared.plugins.pwenc import PWENC_FILE_SECRET
 from middlewared.plugins.cluster_linux.utils import CTDBConfig, FuseConfig
+from middlewared.plugins.filesystem_ import stat_x
 
 
 class FilesystemService(Service):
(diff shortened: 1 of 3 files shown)


anodos325 (approved):

src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -369,9 +370,16 @@ def statfs(self, path):
...
-        fstype = 'zfs' if path.startswith('/mnt/') else 'fuse.glusterfs'
+        # get fstype for given path based on major:minor entry in mountinfo
+        stx = stat_x.statx(path)
+        maj_min = f'{stx.stx_dev_major}:{stx.stx_dev_minor}'
+        fstype = None
+        with open('/proc/self/mountinfo') as f:
+            for line in f:
+                fline = line.split()
+                if len(fline) >= 3 and fline[2] == maj_min:
+                    fstype = line[line.find('-'):].split()[1]
```
Qubad786, line 381: It would have been nice if you could include a comment containing an example that should qualify to extract fstype from..

Qubad786 (approved):

src/middlewared/middlewared/plugins/filesystem.py
yocalebo: I'll add a comment on what the line looks like in `/proc/self/mountinfo` 😄

tests/api2/test_190_filesystem.py
```diff
@@ -8,8 +8,9 @@
 import os
 apifolder = os.getcwd()
 sys.path.append(apifolder)
-from functions import POST
-from auto_config import dev_test
+from functions import POST, SSH_TEST
+from auto_config import dev_test, pool_name
+from config import BSD_HOST, BSD_USERNAME, BSD_PASSWORD
```
anodos325, line 13: I think you'll want:
```
from auto_config import dev_test, ip, pool_name, user, password
```

user, password here will be privileged so you'd do:

```
    results = SSH_TEST(cmd, user, password, ip)
```

But my memory may be fuzzy. Double-check with Eric.

PR #7926, opened 2021-12-01
NAS-113568 / 22.02-RC.2 / Add nut-snmp as a dependency

sonicaj:
This commit adds nut-snmp as a dependency to make sure we support snmp driver for nut.


themylogin (approved):

themylogin (changes requested):
Wait. why here and not in `truenas` package control file?

sonicaj:
I was following `nut`, i think we can move both of these truenas control file. Making the required changes

themylogin (approved):

PR #7928, opened 2021-12-01
NAS-113580 / 22.02-RC.2 / Reduce noisy logs

Qubad786:
- Set git client's logging level to CRITICAL
- Remove necessary logging and just leave exception raises

### DIFF
--- a/src/middlewared/middlewared/logger.py
+++ b/src/middlewared/middlewared/logger.py
@@ -20,7 +20,8 @@ logging.getLogger('botocore').setLevel(logging.WARN)
 # we dont need ws4py close debug messages
 logging.getLogger('ws4py').setLevel(logging.WARN)
 # we dont need GitPython debug messages (used in iocage)
-logging.getLogger('git.cmd').setLevel(logging.WARN)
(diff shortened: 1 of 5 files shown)


sonicaj (approved):

yocalebo (changes requested):
I don't like the idea of storing another item in memory indefinitely while middlewared is running. This has potential for someone to abuse the API and fill up ram. (by specifying random URIs that never resolves). I would rather move these logs to a separate log file instead of having them in `middlewared.log`. We do this for `zettarepl` and `failover` already. Can we do something similar here?

Qubad786:
@yocalebo can you please take another look? Its already not like how I told you in DM last night 😄  I am removing loggers and also tweaked log level for git client to CRITICAL so anything lower than that level would be skipped..

yocalebo:
> @yocalebo can you please take another look? Its already not like how I told you in DM last night 😄 I am removing loggers and also tweaked log level for git client to CRITICAL so anything lower than that level would be skipped..

oh wow, you're right. Sorry, I kept a browser tab opened up from yesterday and just looked at it this morning without refreshing. These changes look good!

yocalebo (approved):

PR #7932, opened 2021-12-02
NAS-113611 / 22.02-RC.2 / properly use missing_log kwarg zfs.import_pool

yocalebo:


Qubad786 (approved):
LGTM!

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -356,8 +356,9 @@ def import_pool(self, name_or_guid, options, any_host, cachefile, new_name):
             if not found:
                 raise CallError(f'Pool {name_or_guid} not found.', errno.ENOENT)
 
+            missing_log = options.pop('missing_log', False)
```
sonicaj, line 359: How are we planning on consuming this ?

yocalebo: I'm sorry, what do you mean? It's being passed as a kwarg directly to libzfs?

sonicaj: It already defaults to `false` in `py-libzfs`, so are we planning on specifying it `true` somewhere as `import_pool` otherwise right now would just set it to `false` right now ?

sonicaj: Also i think we shouldn't be using `options` for this as they are passed to zfs directly and this is confusing because this is not being specified as import options to libzfs. Maybe re-word and change structure of this argument to accommodate the change or add another argument ? ( this is internal api, so we can change it as long as usages are covered )

yocalebo: > It already defaults to false in py-libzfs, so are we planning on specifying it true somewhere as import_pool otherwise right now would just set it to false right now ?

This is confusing me. I consume this private API in `failover_/event.py`. I'm passing `missing_log: True` as a key in the `options` dict because it's `additional_attrs=True`. I assumed that this would be handled by `py-libzfs` but it's not because `py-libzfs` API expects `missing_log` to be it's own keyword argument. I'm fixing that usage. So, while it defaults to false I'm calling this specifying it as true.

yocalebo: > Also i think we shouldn't be using options for this as they are passed to zfs directly and this is confusing because this is not being specified as import options to libzfs. Maybe re-word and change structure of this argument to accommodate the change or add another argument ? ( this is internal api, so we can change it as long as usages are covered )

These are being passed as options when importing the zpool? I'm passing `altroot: '/mnt'` as an option and it's being handled correctly so I'm not really sure what else you'd like to call this variable?

yocalebo: @sonicaj I'll merge this since changing the structure of the API in this PR is outside the scope of what I'm fixing. I'll open a separate ticket to restructure/reword it.

PR #7933, opened 2021-12-03
Add implmentation of configuring immutable flag via ioctl

sonicaj:


yocalebo (changes requested):
This is much better than subprocess 👍 . However, can you add api tests? This would be a great way to catch breakage because hard-coding the `F_IOC_*` values can change with major kernel versions.

themylogin (changes requested):
Great job! I second on having an integration test for this.

tests/api2/test_190_filesystem.py
```diff
@@ -97,3 +100,18 @@ def test_04_test_filesystem_statfs_fstype(pool):
...
+        assert results.status_code == 200, results.text
+        result = results.json()
+        assert isinstance(result, bool) is True, results.text
+        assert result == set, 'Immutable flag is still not set' if set else 'Immutable flag is still set'
+
+    t_path = os.path.join('/mnt', pool, 'random_directory_immutable')
+    with directory(t_path) as d:
+        for v in (True, False):
+            POST('/filesystem/set_immutable/', {'set_flag': v, 'path': d})
+            ensure_result_is_sane(POST('/filesystem/is_immutable_set/', d), v)
```
themylogin, line 117: What if this flag implementation is broken and just does nothing? Won't it be easier and more reliable to check if the flag is doing the job by just trying to perform an actual write?

tests/api2/test_190_filesystem.py
```diff
@@ -6,10 +6,13 @@
...
 import os
-apifolder = os.getcwd()
-sys.path.append(apifolder)
+
 from functions import POST, SSH_TEST
 from auto_config import dev_test, pool_name, ip, user, password
+from middlewared.test.integration.assets.filesystem import directory
+
+apifolder = os.getcwd()
+sys.path.append(apifolder)
```
themylogin, line 15: `functions` and `auto_config` won't import if you move this here.

src/middlewared/middlewared/test/integration/assets/filesystem.py
```diff
@@ -0,0 +1,13 @@
...
+
+from middlewared.test.integration.utils import call
+
+
+@contextlib.contextmanager
+def directory(path):
+    call('filesystem.mkdir', path)
+
+    try:
+        path
```
themylogin, line 11: This statement has no effect

themylogin (changes requested):

src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -52,6 +62,24 @@ def resolve_cluster_path(self, path, ignore_ctdb=False):
         cluster_path = path.replace(FuseConfig.FUSE_PATH_SUBST.value, f'{FuseConfig.FUSE_PATH_BASE.value}/')
         return cluster_path
 
+    @accepts(Str('path'))
+    @returns()
+    def remove_directory(self, path):
```
yocalebo, line 67: This will be abused and potentially cause problems. (`rmtree` is not removing just the directory). Since you added this for the API tests, I'd say remove this method and just use the `SSH_TEST(` method to remove the directory in the integration tests.

src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -52,6 +62,24 @@ def resolve_cluster_path(self, path, ignore_ctdb=False):
...
+        if not p.is_absolute():
+            raise CallError(f'{path}: not an absolute path.', errno.EINVAL)
+
+        if not p.exists():
+            raise CallError(f'{path}: path does not exist.', errno.ENOENT)
+
+        if not os.path.realpath(path).startswith('/mnt/'):
+            raise CallError(f'{path}: path not permitted', errno.EPERM)
+
+        shutil.rmtree(path)
```
anodos325, line 81: This seems like a rather sharp knife. Is this only for doing regression tests? If so, why not just use a SSH_TEST?

yocalebo (changes requested):

src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -11,24 +11,34 @@
...
-    @private
-    async def set_immutable(self, set_flag, path):
-        cp = await run(['chattr', f'{"+" if set_flag else "-"}i', '-V', path], check=False)
-        if cp.returncode:
-            raise CallError(
-                f'Unable to {"set" if set_flag else "remove"} immutable flag on at {path!r} path: {cp.stderr.decode()}'
-            )
+    @accepts(Str('path'))
+    @returns(Bool())
+    def is_immutable_set(self, path):
```
themylogin, line 27: Just `is_immutable`

themylogin (approved):

yocalebo (approved):

PR #7934, opened 2021-12-03
NAS-113612 / 22.02-RC.2 / retry to import zpool on SCALE HA

yocalebo:
Since I fixed the ability to specify a zfs `cachefile` with this previous commit: 561a728f8fa268f5e077a3d9610e6cdd3f32169d, it's common for an import to fail because the `cachefile` is state/invalid/doesnt exist etc, etc so we need to retry to import the zpool without it if this occurs.

### DIFF
--- a/src/middlewared/middlewared/plugins/failover_/event.py
+++ b/src/middlewared/middlewared/plugins/failover_/event.py
@@ -6,6 +6,7 @@ import contextlib
 import shutil
 import threading
 import logging
+import errno
 from collections import defaultdict
 
 from middlewared.utils import filter_list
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/failover_/event.py
```diff
@@ -425,20 +426,32 @@ def vrrp_master(self, job, fobj, ifname, event):
...
             # import the zpool(s)
+            try_again = False
             try:
                 self.run_call('zfs.pool.import_pool', vol['guid'], options, any_host, cachefile, new_name)
             except Exception as e:
-                vol['error'] = str(e)
-                failed.append(vol)
-                continue
+                if e.errno == errno.ENOENT:
+                    try_again = True
```
themylogin, line 437: Should we log it?

themylogin (approved):

src/middlewared/middlewared/plugins/failover_/event.py
yocalebo: Yeah, probably should.

PR #7939, opened 2021-12-06
NAS-113041 / 22.02 / Set ftp TimeoutNoTransfer

themylogin:


yocalebo (changes requested):
I don't like making `TimeoutIdle` = `TimeoutNoTransfer`. Especially for the original ticket creator that is using a `legacy backup application`. The man pages show that they're different options.  (They also don't do a good job of explaining the difference). Anyways, we already have a ton of ftp options in the webUI, so why not add another one?

anodos325:
Maybe we can just allow `options` to override our hard-coded values.

sonicaj (approved):

sonicaj (commented):
I agree with @yocalebo, i think it probably makes more sense to have another field ?

PR #7941, opened 2021-12-06
NAS-113424 / 22.02 / add failsafe middlewared.log

yocalebo:
The idea is to have a logging handler for `middlewared` service that does not get touched by `stop_logging` or `reconfigure_logging` etc. This can be valuable when system dataset migration/move fails, and logging is now "broken" and you can't track down which log file is being used by the currently running process.

### DIFF
--- a/src/freenas/usr/local/bin/ixdiagnose
+++ b/src/freenas/usr/local/bin/ixdiagnose
@@ -147,7 +147,7 @@ fi
 echo "** 96%: Compressing System Logs"
 tar -chf - -C /var --exclude=log/journal --exclude=log/faillog --exclude=log/lastlog --exclude=log/sysstat log | tar -C $dir -xf -
 if [ -f "/tmp/failsafe_middlewared.log" ]; then
-	cp "/tmp/failsafe_middlewared.log" "$dir"
+	cp "/tmp/failsafe_middlewared*" "$dir"
 fi
 
 echo "** 97%: Compressing Debug Files"
--- a/src/freenas/usr/local/bin/ixdiagnose
+++ b/src/freenas/usr/local/bin/ixdiagnose
@@ -146,6 +146,10 @@ fi
 
 echo "** 96%: Compressing System Logs"
 tar -chf - -C /var --exclude=log/journal --exclude=log/faillog --exclude=log/lastlog --exclude=log/sysstat log | tar -C $dir -xf -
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/logger.py
```diff
@@ -32,6 +32,7 @@
 logging.getLogger('kubernetes_asyncio.client.rest').setLevel(logging.WARN)
 logging.getLogger('kubernetes_asyncio.config.kube_config').setLevel(logging.WARN)
 
+FAILSAFE = '/tmp/failsafe_middlewared.log'
```
themylogin, line 35: Can we name it `FAILSAFE_LOGFILE` to be consistent?

themylogin (changes requested):
Why don't we only use it between `stop_logging` and `reconfigure_logging`?

yocalebo:
> Why don't we only use it between `stop_logging` and `reconfigure_logging`?

I don't understand the question.

themylogin:
@yocalebo why do we write to this file always if we only need it after `stop_logging` is called until `reconfigure_logging` succeeds?

yocalebo:
> @yocalebo why do we write to this file always if we only need it after `stop_logging` is called until `reconfigure_logging` succeeds?

Because if we only write to this after `stop_logging` then we have a log file that has out of context log messages. So we look at `tmp/failsafe_middlewared.log` for a few messages (if any) and then we have to go back and look at `/var/log/middlewared.log`.

src/freenas/usr/local/bin/ixdiagnose
```diff
@@ -146,6 +146,10 @@ fi
 
 echo "** 96%: Compressing System Logs"
 tar -chf - -C /var --exclude=log/journal --exclude=log/faillog --exclude=log/lastlog --exclude=log/sysstat log | tar -C $dir -xf -
+if [ -f "/tmp/failsafe_middlewared.log" ]; then
+	cp "/tmp/failsafe_middlewared.log" "$dir"
```
themylogin, line 150: If we have 5 old rotated files, might as well copy them here

themylogin (approved):

src/freenas/usr/local/bin/ixdiagnose
yocalebo: great point. fixed

PR #7945, opened 2021-12-07
NAS-113687 / 22.02 / Document Boot CD and Python packages

themylogin:


themylogin:
time 1:30

src/middlewared/middlewared/job.py
```diff
@@ -223,6 +223,9 @@ class Job:
     :ivar logs_fd: Unbuffered binary file descriptor for writing logs (if the job was defined with `@job(logs=True)`
     """
 
+    pipes: Pipes
```
yocalebo, line 226: Just curious, why is this change needed?

themylogin: Sphinx autodoc issues a warning when a documented property (`:ivar` above) is not explicitly declared. We now treat warnings as errors so...

yocalebo (approved):

PR #7952, opened 2021-12-08
NAS-113702 / 22.02 / improve pool.import_pool

yocalebo:
Preparing for the changes needed to support the zpool cachefile, I've decided to clean this method up a bit. Numerous issues found and fixed.

1. `passphrase` schema attribute is deprecated and not supported on SCALE (used for GELI pools)
2. because of this, `job` object can have `pipes` kwarg removed and the associated logic inside the method
3. we were calling `zfs.pool.find_import` and then calling `zfs.pool.import_pool`. The latter method calls `zfs.pool.find_import` so we were calling it twice for no real benefit.
4. we were calling `pool.query` twice when it can be limited to only being called once
5. we were calling `service.reload, disk` and `self.restart_services()` when the latter method also called `service.reload, disk`. I removed the calls to `service.reload, disk` and the call to `self.restart_services()` because both of those methods are a no-op on SCALE.
6. When recursively resetting the `mountpoint` property, `zfs.dataset.query` returned object has nested `children` keys that list the child datasets recursively for every top-level dataset. This is unneeded so I've introduced a new method `zfs.dataset.child_dataset_names` that returns only the top-level dataset's name given a path and nothing more.
7. clean up the code and remove CORE related logic

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -1433,6 +1433,7 @@ class PoolService(CRUDService):
                             await delegate.toggle(attachments, True)
             await self.middleware.call('keyvalue.delete', key)
 
+        asyncio.ensure_future(self.middleware.call('service.restart', 'collectd')
         await self.middleware.call_hook('pool.post_import', {'passphrase': data.get('passphrase'), **pool})
         await self.middleware.call('pool.dataset.sync_db_keys', pool['name'])
         self.middleware.send_event('pool.query', 'ADDED', id=pool_id, fields=pool)
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -1347,15 +1347,14 @@ class PoolService(CRUDService):
         Bool('enable_attachments'),
     ))
     @returns(Bool('successful_import'))
-    @job(lock='import_pool', pipes=['input'], check_pipes=False)
+    @job(lock='import_pool')
     async def import_pool(self, job, data):
         """
         Import a pool found with `pool.import_find`.
 
         If a `name` is specified the pool will be imported using that new name.
 
-        `passphrase` is required while importing an encrypted pool. In that case this method needs to
-        be called using /_upload/ endpoint with the encryption key.
+        `passphrase` DEPRECATED. GELI not supported on SCALE.
 
         If `enable_attachments` is set to true, attachments that were disabled during pool export will be
         re-enabled.
@@ -1377,91 +1376,55 @@ class PoolService(CRUDService):
                 }]
             }
         """
-
-        pool = None
-        for p in await self.middleware.call('zfs.pool.find_import'):
-            if p['guid'] == data['guid']:
-                pool = p
-                break
-        if pool is None:
-            raise CallError(f'Pool with guid "{data["guid"]}" not found', errno.ENOENT)
-
+        # import zpool
         try:
-            job.check_pipe("input")
-            key = job.pipes.input.r
-        except ValueError:
-            key = None
+            guid = data['guid']
+            opts = {'altroot': '/mnt', 'cachefile': ZPOOL_CACHE_FILE}
+            any_host = True
+            use_cachefile = None
+            new_name = data.get('name')
+            await self.middleware.call('zfs.pool.import_pool', guid, opts, any_host, use_cachefile, new_name)
+        except Exception as e:
+            if e.errno == errno.ENOENT:
+                raise CallError(f'Pool with guid "{guid}" not found', errno.ENOENT)
+            else:
+                raise CallError(f'Failed to import pool with guid "{guid}"')
 
-        passfile = None
-        if key and data.get('passphrase'):
-            encrypt = 2
-            passfile = tempfile.mktemp(dir='/tmp/')
-            with open(passfile, 'w') as f:
-                os.chmod(passfile, 0o600)
-                f.write(data['passphrase'])
-        elif key:
-            encrypt = 1
+        # get the zpool name
+        if not new_name:
+            pool_name = (await self.middleware.call('zfs.pool.query', [['guid', '=', guid]], {'get': True}))['name']
         else:
-            encrypt = 0
+            pool_name = new_name
 
-        pool_name = data.get('name') or pool['name']
-        pool_id = None
-        try:
-            pool_id = await self.middleware.call('datastore.insert', 'storage.volume', {
-                'vol_name': pool_name,
-                'vol_encrypt': encrypt,
-                'vol_guid': data['guid'],
-                'vol_encryptkey': str(uuid.uuid4()) if encrypt else '',
-            })
-            pool = await self.middleware.call('pool.query', [('id', '=', pool_id)], {'get': True})
-            if encrypt > 0:
-                if not os.path.exists(GELI_KEYPATH):
-                    os.mkdir(GELI_KEYPATH)
-                with open(pool['encryptkey_path'], 'wb') as f:
-                    f.write(key.read())
+        # reset top-level zpool `aclinherit` property
+        opts = {'properties': {'aclinherit': {'value': 'passthrough'}}}
+        await self.middleware.call('zfs.dataset.update', pool_name, opts)
 
-            await self.middleware.call('pool.scrub.create', {
-                'pool': pool_id,
-            })
+        # Recursively reset dataset mountpoints for the zpool.
+        recursive = True
+        for child in await self.middleware.call('zfs.dataset.child_dataset_names', pool_name):
+            if child == os.path.join(pool_name, 'ix-applications'):
+                # We exclude `ix-applications` dataset since resetting it will
+                # cause PVC's to not mount because "mountpoint=legacy" is expected.
+                continue
+            try:
+                # Reset all mountpoints
+                await self.middleware.call('zfs.dataset.inherit', child, 'mountpoint', recursive)
+            except Exception as e:
+                # Let's not make this fatal
+                self.logger.warning('Failed to inherit mountpoints recursively for %r dataset: %r', child, e)
 
-            await self.middleware.call('zfs.pool.import_pool', pool['guid'], {
-                'altroot': '/mnt',
-                'cachefile': ZPOOL_CACHE_FILE,
-            }, True, None, pool_name)
+        # update db
+        pool_id = await self.middleware.call('datastore.insert', 'storage.volume', {
+            'vol_name': pool_name,
+            'vol_encrypt': 0,  # TODO: remove (geli not supported)
+            'vol_guid': data['guid'],
+            'vol_encryptkey': '',  # TODO: remove (geli not supported)
+        })
+        await self.middleware.call('pool.scrub.create', {'pool': pool_id})
 
-            await self.middleware.call('zfs.dataset.update', pool_name, {
-                'properties': {
-                    'aclinherit': {'value': 'passthrough'},
-                },
-            })
-
-            # We would like to reset mountpoints of all the datasets under this pool to ensure that we don't
-            # have any misconfigured mountpoints but we would meanwhile like to ensure that we don't do this
-            # for datasets which are managed by us like ix-applications as that will cause PVC's to not mount
-            # because legacy mountpoint is expected
-            for child in (await self.middleware.call(
-                'zfs.dataset.query', [['id', '=', pool_name]], {
-                    'extra': {'retrieve_properties': False, 'flat': False},
-                    'get': True,
-                }
-            ))['children']:
-                if child['name'] == os.path.join(pool_name, 'ix-applications'):
-                    continue
-                try:
-                    # Reset all mountpoints
-                    await self.middleware.call('zfs.dataset.inherit', child['name'], 'mountpoint', True)
-                except Exception as e:
-                    # Let's not make this fatal
-                    self.logger.error(
-                        'Failed to inherit mountpoints recursively for %r dataset: %r', child['name'], e
-                    )
-        except Exception:
-            if pool_id:
-                await self.middleware.call('datastore.delete', 'storage.volume', pool_id)
-            if passfile:
-                os.unlink(passfile)
-            raise
-
+        # reenable/restart any services dependent on this zpool
+        pool = await self.middleware.call('pool.query', [('id', '=', pool_id)], {'get': True})
         key = f'pool:{pool["name"]}:enable_on_import'
         if await self.middleware.call('keyvalue.has_key', key):
             for name, ids in (await self.middleware.call('keyvalue.get', key)).items():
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1377,91 +1376,55 @@ async def import_pool(self, job, data):
...
+            if e.errno == errno.ENOENT:
+                raise CallError(f'Pool with guid "{guid}" not found', errno.ENOENT)
+            else:
+                raise CallError(f'Failed to import pool with guid "{guid}"')
 
-            # We would like to reset mountpoints of all the datasets under this pool to ensure that we don't
-            # have any misconfigured mountpoints but we would meanwhile like to ensure that we don't do this
-            # for datasets which are managed by us like ix-applications as that will cause PVC's to not mount
-            # because legacy mountpoint is expected
-            for child in (await self.middleware.call(
```
sonicaj, line 1442: This is not recursively iterating on all child datasets. It just iterates on first level children similar to the changes you introduced

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1377,91 +1376,55 @@ async def import_pool(self, job, data):
...
-                    'aclinherit': {'value': 'passthrough'},
-                },
-            })
+            guid = data['guid']
+            opts = {'altroot': '/mnt', 'cachefile': ZPOOL_CACHE_FILE}
+            any_host = True
+            use_cachefile = None
+            new_name = data.get('name')
+            await self.middleware.call('zfs.pool.import_pool', guid, opts, any_host, use_cachefile, new_name)
+        except Exception as e:
```
sonicaj, line 1387: I think we should have try/else clause for updating database when we have imported the pool. If any call fails for any reason like `zfs.dataset.update`, we would at least have record of the import in our database

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -547,6 +547,14 @@ def path_to_dataset(self, path):
 
         return ds_name
 
+    def child_dataset_names(self, path):
+        # return child datasets given a dataset `path`.
+        try:
+            with libzfs.ZFS() as zfs:
+                return [child.name for child in zfs.get_dataset_by_path(path).children]
+        except libzfs.ZFSException:
+            return []
```
sonicaj, line 556: Why are we not raising the error and giving the illusion that there are no child datasets ?

src/middlewared/middlewared/plugins/pool.py
yocalebo: No it's not. It's returning all child datasets recursively.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1377,91 +1376,55 @@ async def import_pool(self, job, data):
...
-            await self.middleware.call('zfs.dataset.update', pool_name, {
-                'properties': {
-                    'aclinherit': {'value': 'passthrough'},
-                },
-            })
+            guid = data['guid']
+            opts = {'altroot': '/mnt', 'cachefile': ZPOOL_CACHE_FILE}
+            any_host = True
+            use_cachefile = None
+            new_name = data.get('name')
```
sonicaj, line 1385: Btw, let's please add validation to ensure that the pool name is not already in use

yocalebo: That's already validated by the import process?

sonicaj: The error raised in that case would be `Failed to import pool with guid 1234` :)

src/middlewared/middlewared/plugins/pool.py
sonicaj: Can you please confirm if it is iterating over all children recursively ;) ?

src/middlewared/middlewared/plugins/pool.py
sonicaj: Probably better if we insert before importing and removing on exception from database in case pool import but a later step fails in the importing pool method

src/middlewared/middlewared/plugins/pool.py
yocalebo: I've already confirmed. It is

yocalebo: Let me clarify. The returned dict has `children` keys that have the datasets listed recursively.

yocalebo:
```
{'properties': {}, 'id': 'ha2/parent', 'type': 'FILESYSTEM', 'name': 'ha2/parent', 'pool': 'ha2', 'children': [{'properties': {}, 'id': 'ha2/parent/child', 'type': 'FILESYSTEM', 'name': 'ha2/parent/child', 'pool': 'ha2', 'children': [{'properties': {}, 'id': 'ha2/parent/child/grandchild', 'type': 'FILESYSTEM', 'name': 'ha2/parent/child/grandchild', 'pool': 'ha2', 'children': [{'properties': {}, 'id': 'ha2/parent/child/grandchild/grand-grand-child', 'type': 'FILESYSTEM', 'name': 'ha2/parent/child/grandchild/grand-grand-child', 'pool': 'ha2', 'children': [{'properties': {}, 'id': 'ha2/parent/child/grandchild/grand-grand-child/grand-grand-grand-child', 'type': 'FILESYSTEM', 'name': 'ha2/parent/child/grandchild/grand-grand-child/grand-grand-grand-child', 'pool': 'ha2', 'children': [{'properties': {}, 'id': 'ha2/parent/child/grandchild/grand-grand-child/grand-grand-grand-child/grand-grand-grand-grand-child', 'type': 'FILESYSTEM', 'name': 'ha2/parent/child/grandchild/grand-grand-child/grand-grand-grand-child/grand-grand-grand-grand-child', 'pool': 'ha2', 'children': []}]}]}]}]}]}

Comment (sonicaj) [reply src/middlewared/middlewared/plugins/pool.py:1442]:
```
❯ python3
Python 3.9.2 (default, Feb 28 2021, 17:03:44)
[GCC 10.2.1 20210110] on linux
>>> from middlewared.client import Client as cl
>>> d=c.call('zfs.dataset.query', [['id', '=', 'evo']], {"extra": {"retrieve_properties": False, "flat": False}})
>>> len(d)
1
>>> len(d[0]['children'])
35
>>> exit()
❯ zfs list -r evo | wc -l
864
❯ zfs list -d 1 -H evo | wc -l
36
```

Comment (sonicaj) [reply src/middlewared/middlewared/plugins/pool.py:1442]: >We were iterating every single dataset (recursively) for the imported pool and resetting the mountpoint recursively. This made no sense. All we need to do is iterate the top-level datasets of the zpool and reset the mountpoints ?passing the recursive flag. This, potentially, saves a ton of calls to zfs.dataset.inherit

Ah, i was confused because of this ☝🏻

File: src/middlewared/middlewared/plugins/pool.py (hunk 4 above)
Comment (yocalebo) [reply src/middlewared/middlewared/plugins/pool.py:1385]: Then the error in `py-libzfs` should be fixed.

Comment (sonicaj) [reply src/middlewared/middlewared/plugins/pool.py:1385]: Could you please clarify which one we are referring to ? Also we catch the exception and we do not propagate it appropriately even with better handling in the underlying zfs plugin or py-libzfs ?

Comment (yocalebo) [reply src/middlewared/middlewared/plugins/pool.py:1385]: Sure, you asked to check to make sure the pool doesn't already exist. (Meaning it's already imported)
```
>>> c.call('zfs.pool.import_pool', 'ha2')
  File "/usr/lib/python3/dist-packages/middlewared/client/client.py", line 465, in call
    raise ClientException(c.error, c.errno, c.trace, c.extra)
middlewared.client.client.ClientException: [ENOENT] Pool ha2 not found.
>>>
root[~]# zpool list
NAME        SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
boot-pool    31G  2.73G  28.3G        -         -     0%     8%  1.00x    ONLINE  -
ha2        7.50G  1.36M  7.50G        -     1.50G     0%     0%  1.00x    ONLINE  /mnt
tank       7.50G  1.33G  6.17G        -     1.50G     7%    17%  1.00x    ONLINE  /mnt
root@[~]# zpool import ha2
cannot import 'ha2': a pool with that name already exists
```

Why is `zfs.pool.import_pool` not raising `EEXIST`?

File: src/middlewared/middlewared/plugins/pool.py (hunk 1 above)
Comment (yocalebo) [reply src/middlewared/middlewared/plugins/pool.py:1442]: I'll update the verbiage in the comment to not cause confusion. Thanks!

File: src/middlewared/middlewared/plugins/zfs.py (hunk 3 above)
Comment (yocalebo) [reply src/middlewared/middlewared/plugins/zfs.py:556]: Good point, I'll remove the seatbelt.

File: src/middlewared/middlewared/plugins/pool.py (hunk 2 above)
Comment (yocalebo) [reply src/middlewared/middlewared/plugins/pool.py:1387]: What happens if the zpool was exported via the CLI? When I try to insert first, it will raise a unique constraint traceback. How do we handle that?

Comment (sonicaj) [reply src/middlewared/middlewared/plugins/pool.py:1387]: Is that a behavior we support ? I am sure there can be quite other cases in different parts which can arise to manual intervention ?

Comment (yocalebo) [reply src/middlewared/middlewared/plugins/pool.py:1387]: If we remove the db entry but the zpool import is successful and so we lave the zpool imported, what does that really do for us?

Comment (yocalebo) [reply src/middlewared/middlewared/plugins/pool.py]: My changes still produce the same outcome btw. If importing fails (or anything after it), we won't reach the part of inserting a db entry.

Comment (sonicaj) [reply src/middlewared/middlewared/plugins/pool.py]: I am not sure what aspect you are referring to ? It would probably be nice if you could clarify why we are changing existing behavior ?

File: src/middlewared/middlewared/plugins/zfs.py [hunk 5]
```diff
@@ -547,6 +552,11 @@ def path_to_dataset(self, path):
 
         return ds_name
 
+    def child_dataset_names(self, path):
+        # return child datasets given a dataset `path`.
+        with libzfs.ZFS() as zfs:
```
sonicaj, line 557: We should catch `ZFSException` and raise it appropriately as a `CallError` like we do for other usages in zfs plugin so that if it needs to be handled for any reason in a middleware plugin, it can be done in a normalized manner ensuring isolated usages of `py-libzfs`.

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -88,6 +88,11 @@ def query(self, filters, options):
                 pools = [i.__getstate__(**state_kwargs) for i in zfs.pools]
         return filter_list(pools, filters, options)
 
+    def query_imported_fast(self):
+        # the equivalent of running `zpool list -H -o guid,name` from cli
+        with libzfs.ZFS() as zfs:
```
sonicaj, line 93: Same as below

sonicaj (changes requested):

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1473,14 +1433,7 @@ async def import_pool(self, job, data):
                             await delegate.toggle(attachments, True)
             await self.middleware.call('keyvalue.delete', key)
 
-        await self.middleware.call('service.reload', 'disk')
-        asyncio.ensure_future(self.restart_services())
```
sonicaj, line 1477: Btw `collectd` is restarted when this is called, that won't happen now - is that okay ?

src/middlewared/middlewared/plugins/pool.py
yocalebo: I'm not changing the behavior? If any API calls fail before we get to the point of inserting the db entry, then the zpool will be left imported. Before my changes, this would also occur.

sonicaj (changes requested):
We are not importing the pool anymore `import_pool` ? 😅

src/middlewared/middlewared/plugins/pool.py
sonicaj: The db entry is made first and then the pool is imported and on failure the entry is removed. Is that still happening ?

yocalebo: The end-result is the same. If anything fails during import process, no db entry is made.

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -88,6 +88,14 @@ def query(self, filters, options):
                 pools = [i.__getstate__(**state_kwargs) for i in zfs.pools]
         return filter_list(pools, filters, options)
 
+    def query_imported_fast(self):
```
sonicaj, line 91: I think we should rename this method as it will get pools which are available right now in the system which includes any which have been imported or created. So `query_available_pools_guid` or something along these lines seems better to me.

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -88,6 +88,14 @@ def query(self, filters, options):
...
+        # the equivalent of running `zpool list -H -o guid,name` from cli
+        try:
+            with libzfs.ZFS() as zfs:
+                return {str(i.guid): i.name for i in zfs.pools}
+        except libzfs.ZFSException as e:
+            raise CallError(f'Failed listing imported pools with error: {e}')
```
sonicaj, line 97: Let's reword the error here please and remove the `imported` wording

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1377,91 +1376,61 @@ async def import_pool(self, job, data):
...
-                'properties': {
-                    'aclinherit': {'value': 'passthrough'},
-                },
-            })
+            opts = {'altroot': '/mnt', 'cachefile': ZPOOL_CACHE_FILE}
+            any_host = True
+            use_cachefile = None
+            await self.middleware.call('zfs.pool.import_pool', guid, opts, any_host, use_cachefile, new_name)
+        except Exception as e:
+            raise CallError(f'Failed importing pool with guid "{guid}" with error {e}')
```
sonicaj, line 1397: I think we should raise `CallError` if import pool fails in zfs plugin instead of raising `ZFSException` as that seems to be the only usage where this is not happening. In that case, we don't need to handle the exception at all and we would/should have correct codes set as well like if the pool was not available for import, a call error would already be raised with the correct error code.

src/middlewared/middlewared/plugins/zfs.py
yocalebo: No. Why would I remove that word? It's, literally, trying to list imported pools.

src/middlewared/middlewared/plugins/zfs.py
yocalebo: No, this is querying looks that have been imported.

sonicaj: So a created pool == imported pool ? I don't mind the current name but it seems off. Looking apart from the import pool perspective, if i was to just get pools with their guids, it seemed off to me to use `query_imported_fast` method name

yocalebo:
At this point, the review has become pedantic as we're talking about interpretation of what error messages should read and what names of methods should be. Frankly, I'm tired of the review and so I'm closing this and will move onto something else. This was only an improvement so not really needed.

src/middlewared/middlewared/plugins/zfs.py
yocalebo: I'm not sure what "created" means in this context. Sure, the pools that have been listed here are also "created" but again, not sure what that means.

This list pools that have been imported only. 
```
with libzfs.ZFS() as zfs:
    print({i.guid: i.name for i in zfs.pools})
```

If you want to look at "created" pools (but not imported)"
```
with libzfs.ZFS() as zfs:
    for i in zfs.find_import():
        print(i)
```

sonicaj (approved):

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1347,15 +1347,14 @@ async def import_find(self, job):
...
+    @job(lock='import_pool')
     async def import_pool(self, job, data):
         """
         Import a pool found with `pool.import_find`.
 
         If a `name` is specified the pool will be imported using that new name.
 
-        `passphrase` is required while importing an encrypted pool. In that case this method needs to
-        be called using /_upload/ endpoint with the encryption key.
+        `passphrase` DEPRECATED. GELI not supported on SCALE.
```
themylogin, line 1357: Why keep this property at all?

PR #7984, opened 2021-12-13
NAS-113797 / 22.02 / prevent duplicate vlan entries in db

yocalebo:
If a name is not provided when creating a vlan, then we use the `vlan_tag` attribute as the name of the interface. This causes multiple issues. Below is a scenario:

1. create `vlan` with no name and give it a tag of 100
2. create another `vlan` with no name and give it a tag of 100

This does 2 things:
1. overwrites the original vlan entry in `network_interfaces` table with whatever information was provided in step # 2 above. (This is undefined behavior because a `do_update` call should be made in this specific scenario)
2. a duplicate entry will be made to `network_vlan` table

To remedy both of these problems, I'm using the `interface.get_next` method which is what `br` and `bond` interfaces use to generate their names. There should be no functional change.

While I'm here, I've removed the unnecessary `interface.get_next_name` method since it provides no benefit. Instead use the `interface.get_next` method for all the virtual interface types.

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -1059,8 +1059,7 @@ class InterfaceService(CRUDService):
 
         interface_id = None
         if data['type'] == 'BRIDGE':
-            # For bridge we want to start with 2 because bridge0/bridge1 may have been used for VM.
-            name = data.get('name') or await self.middleware.call('interface.get_next_name', InterfaceType.BRIDGE)
+            name = data.get('name') or await self.middleware.call('interface.get_next', 'br')
             try:
                 async for i in self.__create_interface_datastore(data, {
                     'interface': name,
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -1118,7 +1116,7 @@ async def do_create(self, data):
                         )
                 raise
         elif data['type'] == 'VLAN':
-            name = data.get('name') or f'vlan{data["vlan_tag"]}'
+            name = data.get('name') or await self.middleware.call('interface.get_next', 'vlan')
```
themylogin, line 1119: What will happen if `data["name"]` is specified and points to already existing interface? Won't the same thing happen?

yocalebo: On line `1054` we call `await self._common_validation` which has this logic:
```
1194         if 'name' in data and data['name'] in ifaces:
1195             verrors.add(f'{schema_name}.name', 'Interface name is already in use.')
```

so it should never reach this point?

rick-mesta:
@themylogin: anything else holding this up ?

themylogin (approved):

PR #7987, opened 2021-12-13
NAS-112591 / 22.02 / Alert for large number of snapshots

themylogin:


themylogin:
time 3:30

themylogin:
@william-gr is this really necessary for 13.0?

william-gr:
@themylogin no, if its not trivial to backport we can skip

themylogin:
@william-gr it's trivial to backport (except for tests), I was just under impression that 13.0 is only for OS upgrade and bugfixes

william-gr:
@themylogin right, lets keep this only for SCALE then

yocalebo:
@themylogin Is this supposed to create an alert for every snapshot that is `> max`?

themylogin:
@yocalebo it will only create one alert for the first dataset it finds

yocalebo (changes requested):
Okay, this will cause some horrific performance problems.
1. `libzfs.datasets_serialized` is kind of ridiculous. I honestly have no idea why it's not just running `zfs_iter_snapshots_sorted` in upstream `libzfs` API. I'm 99.99% confident that the `zfs.snapshot.query [] {"select": "name"}` is causing the vast majority of the performance problems. I'm testing this theory on a system and will let you know.
2. I see a flaw with this approach. The `512` number is only relevant for datasets that have been shared via SMB and the shares are enabled and have `shadowcopy` option enabled. To make matters worse, `ZFS_PROP_WRITTEN > 0` is what the underlying smb program uses before it even lists a previous version in windows explorer. (It basically skips "empty" snapshots). There is no easy way to do that right now so even if we add the logic in here it's going to be bugged almost immediately.
3. I _really_ do not want to traverse all of the snapshots for all datasets on the system only to raise an alert for 1 of them.
4. I don't think traversing all datasets and their snapshots is the right approach because it's going to cause performance issues no matter how we slice it so I think a different approach needs to be taken.

I propose that we only raise an alert for a dataset in the following 2 scenarios:
1. if the snapshot task is to produce more than 10k snapshots, then we either 1. need to raise another webUI pop-up asking the end-user to confirm that this will cause severe performance penalties (kind of like how we do when they create a striped zpool) and/or 2. we need to generate an alert for this dataset based on the `forseen_count` method that you've added (or we add a method that does a "predictive" count based on the parameters given to us by the end-user).
2. if the snapshot task is to produce more than 512 snapshots and the dataset is shared via SMB and is enabled and has `shadowcopy` enabled, then we need to raise an alert with some specific verbiage about how windows explorer might not show all of the snapshots

yocalebo:
Okay I did some very rudimentary performance analysis on this. On a test VM with 8GB of RAM and 2 cores with `9574` empty snapshots, it's taking `zfs.snapshot.query [] {'select': ['name']}` ~1.45 seconds to complete (using 16% of CPU core). Since this alert will run every 60 seconds, it's using 2.5% of the available time before this will run again. So I'll do some basic math and make some assumptions. I'll say it takes ~1.5 seconds for 10k snapshots.

If there are 100k snapshots, then it will take 15 seconds just to list the snapshots which is 25% of the total time this alert will take before it spins up another child process to run it again. Hopefully you can see how quickly this will spiral out of control.

themylogin:
@yocalebo sorry, forgot to add this, this is only to be ran every 24 hours.

Your (1) proposal will only create an alert if periodic snapshot task retention is not working properly. If user wants a periodic snapshot task that retains million snapshots, its forseen_count will also be a million and no alert will be triggered, right?

This also does not account for snapshots created by replication pushed from another host, or pull replication, or cron scripts, or anything else. They all do lower the performance, right?

And what's this 10k number? Is it overall number, or per-pool, or per-dataset?

yocalebo:
> @yocalebo sorry, forgot to add this, this is only to be ran every 24 hours.
> 
> Your (1) proposal will only create an alert if periodic snapshot task retention is not working properly. If user wants a periodic snapshot task that retains million snapshots, its forseen_count will also be a million and no alert will be triggered, right?

If a periodic snapshot task will retain a million snapshots, then what does `forseen_count` calculate? The name of that method implies that it will "foresee" (predict) this. If it does predict this, then why would we not raise an alert saying something like "This will retain 1 million snapshots which will cause a severe performance issue".

> This also does not account for snapshots created by replication pushed from another host, or pull replication, or cron scripts, or anything else. They all do lower the performance, right?

Then we need to rethink this problem entirely and maybe scrap it. We have enterprise customers on CORE where a child process is eating GB's of memory just to list the snapshots which is causing horrendous performance problems that leak over into the zpool IO. So it's not just a `middlewared` service problem, it's an event that causes an outage that we're causing.

> And what's this 10k number? Is it overall number, or per-pool, or per-dataset?

10k is just an unspoken arbitrary number (from experience with enterprise customers) that is the upper threshold when significant performance problems are encountered. It took ~1.5 seconds to list them and eats 16% CPU on a zpool that had 0 IO going on and those snapshots were empty. It gets exponentially worse when those snapshots have changed data and IO is ongoing on the system.

william-gr:
I dont think it would become a problem if we are running this only once a day as it currently is. zfs list -t snapshot -s name -o name as this is supposedly being used does not need to retrieve any extra metadata and its quite fast. Doing this regularly once a day will potentially prevent the number of snapshots getting extremely large.

themylogin:
> why would we not raise an alert saying something like "This will retain 1 million snapshots which will cause a severe performance issue".

Because when such a periodic snapshot task was created, we warned user that they'll be having a large amount of snapshots. We emit warnings when a pool with stripe vdev is created, but we don't emit alerts if such a pool is present in the system.

> We have enterprise customers on CORE where a child process is eating GB's of memory just to list the snapshots

Is this because of `py-libzfs`? Does `zfs list -t snapshot -o name -s name` work normally? In that case we might fix this in a separate ticket.

sonicaj:
Good point @themylogin about doing a performance comparison with CLI. I have just carried a small test and results generally align with the performance we get with CLI. Running py-libzfs without middleware, we can even get slightly better results :)

```
❯ time zfs list -t snapshot -o name -s name | wc -l
11550
zfs list -t snapshot -o name -s name  0.05s user 7.95s system 99% cpu 8.069 total
wc -l  0.00s user 0.00s system 0% cpu 8.068 total
❯ time midclt call zfs.snapshot.query '[]' '{"select": ["name"]}' 1>/dev/null
midclt call zfs.snapshot.query '[]' '{"select": ["name"]}' > /dev/null  0.19s user 0.02s system 2% cpu 8.194 total
❯
❯ python3
Python 3.9.2 (default, Feb 28 2021, 17:03:44)
[GCC 10.2.1 20210110] on linux
>>> import libzfs
>>> import time
>>> start=time.time();z=libzfs.ZFS();snaps = z.snapshots_serialized(['name']);print(time.time() - start)
7.557496547698975
>>>
```

themylogin:
@sonicaj I think Caleb's point was also memory consumption

yocalebo:
> Good point @themylogin about doing a performance comparison with CLI. I have just carried a small test and results generally align with the performance we get with CLI. Running py-libzfs without middleware, we can even get slightly better results :)
> 
> ```
> ❯ time zfs list -t snapshot -o name -s name | wc -l
> 11550
> zfs list -t snapshot -o name -s name  0.05s user 7.95s system 99% cpu 8.069 total
> wc -l  0.00s user 0.00s system 0% cpu 8.068 total
> ❯ time midclt call zfs.snapshot.query '[]' '{"select": ["name"]}' 1>/dev/null
> midclt call zfs.snapshot.query '[]' '{"select": ["name"]}' > /dev/null  0.19s user 0.02s system 2% cpu 8.194 total
> ❯
> ❯ python3
> Python 3.9.2 (default, Feb 28 2021, 17:03:44)
> [GCC 10.2.1 20210110] on linux
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import libzfs
> >>> import time
> >>> start=time.time();z=libzfs.ZFS();snaps = z.snapshots_serialized(['name']);print(time.time() - start)
> 7.557496547698975
> >>>
> ```

Yes, and if you look you will see that the CLI is spinning a CPU core. I'm also interested in memory consumption in the worst case.

We're also not addressing these issues:
1. we're setting the `max` to 512 which is irrelevant for datasets that aren't shared via smb
2. even if the dataset is shared via SMB and has `shadowcopy` enabled, then we're not checking to see if those snapshots are "empty" (`ZFS_PROP_WRITTEN > 0`). SMB does this behind the scenes and can actually populate more than 512 snapshots since they're empty.
3. we're iterating all datasets and all their snapshots only to raise an alert for the first dataset that is `> max`.

sonicaj:
@yocalebo can you please confirm what numbers we are seeing with py-libzfs in comparison to zfs cli for memory consumption ?

themylogin:
@yocalebo do you see any way out besides not alerting for a large number of existing snapshots because listing snapshots is too expensive?

yocalebo:
> @yocalebo do you see any way out besides not alerting for a large number of existing snapshots because listing snapshots is too expensive?

Yeah, I think it's too expensive. The only idea that I have is to do some basic (assumptive) math on the front-end when the snapshot task is created to say something like "The task that you've configured will store 20k snapshots before they are pruned. This is not recommended and can cause severe performance penalties." or something like that. (Maybe we already have this, I can't remember)

william-gr:
I dont necessarily agree its that expensive if only done once a day, and using -o name -s name. If you run that on a system with a couple dozen of thousand of snapshots you dont notice such a load. And thats an extreme case, people should get to that.

themylogin:
@yocalebo so, did we see a case where running `zfs list -t snapshot -o name -s name` caused any issues?

themylogin:
@yocalebo ping :)

yocalebo (approved):

themylogin:
@yocalebo sorry, it’s a mess. Current state of this PR is not listing snapshots for the discussed alert. Do you approve only this or will you approve listing snapshots too?

yocalebo:
> @yocalebo sorry, it’s a mess. Current state of this PR is not listing snapshots for the discussed alert. Do you approve only this or will you approve listing snapshots too?

I have access to hardware where I can create a stupid amount of snapshots. Let me do some very basic testing to see what happens when we try to list the snapshots.

themylogin:
Merging after confirmation in #middleware channel

PR #7989, opened 2021-12-14
NAS-104989 / 22.02 / Return IP network if schema wants an IP network

themylogin:
time 1:00

### DIFF
--- a/src/middlewared/middlewared/pytest/unit/test_schema.py
+++ b/src/middlewared/middlewared/pytest/unit/test_schema.py
@@ -652,6 +652,45 @@ def test__schema_ipaddr_cidr(value, expected):
         assert ipaddrv(self, value) == expected
 
 
+@pytest.mark.parametrize("value,expected", [
+    ('192.168.0.0%enp0s3', ValidationErrors),
+    ('22::56%enp0s3', '22::56%enp0s3'),
+])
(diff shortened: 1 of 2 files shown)


yocalebo (changes requested):
The original ticket was for CORE. Have you tested that this even applies to SCALE?

themylogin:
@yocalebo this ticket fixes a high-level validation issue that will benefit every middleware subsystem. There are no NFS-related changes here.

yocalebo:
> @yocalebo this ticket fixes a high-level validation issue that will benefit every middleware subsystem. There are no NFS-related changes here.

The original ticket was for undefined behavior on freeBSD's `mountd` because of how we were generating the exports file. So this is indirectly changing NFS share creating/update behavior. I ask about SCALE because the webUI (on nightlies) correctly formats the payload on the front-end so it is working "by mistake". I agree with the changes, btw. I just wish there was a little more context because this change affects every single API endpoint that uses the `IPAddr` validator while the original ticket was specific to CORE's `mountd` and the exports file (which doesn't even exist on SCALE).

yocalebo (approved):

themylogin:
@yocalebo there are 3 consumers of `IPAddr(..., network=True)`, all of them actually need a network so converting from `IP` to `IP/32` won't break anything

PR #7991, opened 2021-12-14
NAS-111093 / 13.0 / Make reporting graphs consistent with top

themylogin:
Do not try to "stack" them, stacking is only supported in SCALE UI.
Here it just confuses users with incorrect numbers.


themylogin:
time 1:00

yocalebo (approved):

PR #7994, opened 2021-12-14
NAS-112977 / 13.0 / make middlewared usable on systems with 100's of disks

yocalebo:
Where to begin. This fixes many numerous problems and I've tried to break up this change into many commits. There were a few primary problems that I've found while implementing these changes.

1. we listen to `devd` for disk `DESTROY` events. On a system with 1.2k disks, when we created a zpool we ran `gpart destroy` on the disks. This caused a cascading number of problems that eventually lead to the `middlewared` process getting bogged down to the point it was unusable. For example:
--- when `gpart destroy` is run on a disk _multiple_ _duplicate_ events are generated via devd
--- we processed these events and ran `(disk.sync_all).wait()`
--- `disk.sync_all` is painful by itself but now we're running it consecutively 1000's of times during the zpool creation

2. when we format disks, we decided to throw all the fork+exec'ing (subprocess) into the main event loop. This overloaded the main event loop and it eventually fought over the GIL and cause, ultimately, massive lock contention. I've moved the formatting of the disks to a `ThreadPoolExecutor` and only do 16 disks in parallel. This change produced dramatically different results in terms of CPU usage and the time it took to complete the formatting.

The major new feature that I've added is a single thread that caches the disk information. This caches the formatted json output that we expect as well as the xml information. The `bsd.geom` plugin was implemented to do this but it was clearly abandoned once it was implemented because we were running `geom.scan()` at every opportunity we got. This is massively painful because the xml information for a system with 1.2k disks is kind of absurd (100k+ of lines). This thread will cache this information as needed which dramatically reduces the time it takes to do common tasks related to xml/disks. I've also moved it underneath the management of the `middlewared` service to make it "easier" to manage moving forward (instead of having to go to another python module in another repo just to see what it's doing).


yocalebo:
TODO: Still working on how best to handle devd events. Unfortunately, there is no easy way to differentiate a disk being physically removed/added as opposed to just having a partition destroyed/created etc.

src/middlewared/middlewared/plugins/disk_/info.py
```diff
@@ -43,16 +43,8 @@ async def get_partition_uuid_from_name(self, part_type_name):
 
     @private
     async def get_valid_zfs_partition_type_uuids(self):
-        # https://salsa.debian.org/debian/gdisk/blob/master/parttypes.cc for valid zfs types
-        # 516e7cba was being used by freebsd and 6a898cc3 is being used by linux
-        return [
-            '6a898cc3-1dd2-11b2-99a6-080020736631',
-            '516e7cba-6ecf-11d6-8ff8-00022d09712b',
-        ]
+        return ['516e7cba-6ecf-11d6-8ff8-00022d09712b']
```
william-gr, line 46: Is still a valid zfs partition, no? pool might have been created on Linux but used on BSD

yocalebo: True, I fixed.

src/middlewared/middlewared/plugins/disk_/disk_events.py
```diff
@@ -1,66 +1,48 @@
...
-
-
-if osc.IS_FREEBSD:
-    import sysctl
-    RE_ISDISK = re.compile(r'^(da|ada|vtbd|mfid|nvd|pmem)[0-9]+$')
+import sysctl
 
 
 async def added_disk(middleware, disk_name):
+    await middleware.call('geom.add_disk', disk_name)
```
sonicaj, line 7: As `add_disk` is in reality invalidating the cache, probably better we don't hide that and rename this to reflect what's really happening

src/middlewared/middlewared/plugins/disk_/disk_info.py
```diff
@@ -1,17 +1,14 @@
 import os
 import re
-from xml.etree import ElementTree as etree
 
 import bsd
-import sysctl
 from middlewared.service import Service
-from .disk_info_base import DiskInfoBase
```
sonicaj, line 8: We are making changes to not consume `DiskInfoBase`, why keep those files still ( `disk_info_base.py` ) ? Although probably better/nicer to make these cleanup changes in a different PR :)

src/middlewared/middlewared/plugins/disk_/identify.py
```diff
@@ -17,68 +19,65 @@ async def device_to_identifier(self, name, disks=None):
...
         value = search.group('value').replace("'", '%27')
 
         if tp == 'uuid':
-            _find = f'.//config[rawuuid = "{value}"]/../../name'
+            _find = f'.//config[rawuuid="{value}"]/../../name'
             if geom_xml:
                 search = geom_xml.find(_find)
             else:
-                search = geom.class_by_name('PART').xml.find(_find)
+                search = self.middleware.call_sync('geom.get_class_xml', 'PART').find(_find)
```
sonicaj, line 59: So we are certain we will have xml here but that doesn't hold at L23 ? Why the inconsistency ?

src/middlewared/middlewared/plugins/disk_/swap_mirror.py
```diff
@@ -0,0 +1,89 @@
...
+
+        cp = await run('gmirror', 'destroy', name, check=False, encoding='utf8')
+        if cp.returncode:
+            raise CallError(f'Failed to destroy mirror {mirror_name}: {cp.stderr}')
+
+    @private
+    @filterable
+    def get_swap_mirrors(self, filters, options):
+        xml = self.middleware.call_sync('geom.get_xml')
+        if not xml:
```
sonicaj, line 30: Same as above, when do we expect this to not get us anything ?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -405,39 +405,45 @@ async def upgrade(self, oid):
...
-
-            for key in x:
-                if key == 'type' and isinstance(x[key], str):
-                    x[key] = x[key].upper()
-                else:
-                    x[key] = self.transform_topology(x[key], dict(options, geom_scan=False))
-        elif isinstance(x, list):
-            for i, entry in enumerate(x):
-                x[i] = self.transform_topology(x[i], dict(options, geom_scan=False))
+        info = self.middleware.call_sync('disk.label_to_dev_and_disk')
```
sonicaj, line 413: Can you please clarify what we are accomplishing apart from removing the recursion ? My concern is not having changes introduced unless really required to stable branch :)

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -745,26 +736,14 @@ async def do_create(self, job, data):
                 'pool.dataset.insert_or_update_encrypted_record', encrypted_dataset_data
             )
 
-            if osc.IS_FREEBSD:
-                await self.middleware.call('pool.save_encrypteddisks', pool_id, formatted_disks, disks_cache)
```
sonicaj, line 749: Do we not need to do this anymore ?

sonicaj (changes requested):
I am curious if this works as desired due to device leftover service which hasn't been removed and should ideally result in middleware not starting i think :) 
Can you please confirm ?

I'll be having another pass at it as well for the pool plugin. Can you please also confirm if the changes introduced in the pool plugin apart from the formatting part are required/necessary ? I ask due to risk of regression in a stable branch :)

sonicaj:
Also in general i do think we should avoid cleaning large parts of linux code or re-writing service parts in the same PR where our usecase is different as i think it increases potential of missing a potential regression in a diff ( a good example is the swap service parts )

PR #7996, opened 2021-12-15
NAS-113810 / 22.02 / Fix KerberosPrincipal parameter in ganesha config

anodos325:


anodos325:
Closing this out because we switched to kernel NFS server.

PR #7997, opened 2021-12-15
NAS-113698 / 22.02 / Added some sleeps in ad_idmap test 8

ericbsd:
This is to avoid restarting systemd service more then 5 time per 10 second which is the default limit.


anodos325 (approved):

rick-mesta:
@q5sys : this is the last issue on the middleware team's radar for RC.2 -- would you kindly review 🙏🏼 ?

PR #8003, opened 2021-12-16
NAS-113693 / 13.0 / add MCA and APEI event detection

yocalebo:
Requested by support and PM departments.


src/freenas/usr/local/lib/middlewared_truenas/alert/source/mca_and_apei_events.py
```diff
@@ -0,0 +1,37 @@
...
+                'criticality': 'Potential loss of functionality',
+                'environment': 'Production',
+                'name': 'Automatic Alert',
+                'email': 'auto-support@ixsystems.com',
+                'phone': '-',
+            }
+            job = await self.middleware.call('support.new_ticket', info)
+            await job.wait()
+            if job.error:
+                self.logger.warning(f'Failed to generate proactive ticket for MCA/APEI event(s): {job.error!r}')
```
themylogin, line 37: Should we just fail here, creating an "alert checker running failed" alert to catch user's attention?

src/freenas/usr/local/lib/middlewared_truenas/alert/source/mca_and_apei_events.py
```diff
@@ -0,0 +1,37 @@
...
+
+from middlewared.alert.base import AlertSource
+from middlewared.alert.schedule import IntervalSchedule
+
+
+class MCAAlertSource(AlertSource):
+    products = ('ENTERPRISE',)
+    schedule = IntervalSchedule(timedelta(hours=24))
+
+    async def check(self):
```
themylogin, line 11: Is this desired to be posted for the same system every 24 hours?

Should we just return a normal alert and the just hide it from being displayed? (it'll be a new code)

yocalebo: > Is this desired to be posted for the same system every 24 hours?
Yes, it was requested by the support team.

> Should we just return a normal alert and the just hide it from being displayed? (it'll be a new code)
Support and PM team do not want to raise these types of alerts to the end-user since there are some of them that are not necessarily fatal.

src/freenas/usr/local/lib/middlewared_truenas/alert/source/mca_and_apei_events.py
yocalebo: > Should we just fail here, creating an "alert checker running failed" alert to catch user's attention?

No, we don't want to raise this to user's attention for the same reason I responded below.

src/freenas/usr/local/lib/middlewared_truenas/alert/source/mca_and_apei_events.py
themylogin: I think we should instead just start a coroutine loop with `sleep(86400)` like we do for usage stats.

Creating an alert has a side effect of this (never emitted) alert appearing in alert configuration dialogs.

yocalebo: @themylogin yep, agree 100%. I've completely forgot about @periodic decorator 😬 I'll fix

themylogin (approved):

rick-mesta (approved):

PR #8022, opened 2021-12-19
NAS-113848 / 22.02 / Add validation to ensure that iso file can be consumed by libvirt

sonicaj:


yocalebo (changes requested):
I know this is WIP but we probably want to think about this a little more before we implement your solution.

I'd much rather resolve `username` to a uid and then use `os.stat` module (so we don't have to subprocess) and check the `st_uid` attribute for the given path and then `st_mode`.

Also, this only checks posix mode bits and not NFSv4 style acls so maybe we need to name the method a little more to what it's checking.

yocalebo:
Spoke to Waqar in chat and this change is fine for what's being accomplished but a new ticket will be opened up to rewrite the `run_command_with_user_context` (or maybe write an entirely different method to do something similar) that doesn't involve `subprocess` (since this is just trying to access path). 😄

src/middlewared/middlewared/plugins/filesystem_/perm_check.py
```diff
@@ -0,0 +1,44 @@
...
+        Str('path', empty=False),
+        Dict(
+            'permissions',
+            Bool('read', default=None, null=True),
+            Bool('write', default=None, null=True),
+            Bool('execute', default=None, null=True),
+        )
+    )
+    @returns(Bool())
+    def check_access_as_user(self, username, path, perms):
```
themylogin, line 23: `can_access_as_user`

themylogin (changes requested):

src/middlewared/middlewared/plugins/vm/vm_devices.py
```diff
@@ -374,6 +375,36 @@ async def validate_device(self, device, old=None, vm_instance=None, update=True)
...
+                    'user.query', [['username', '=', LIBVIRT_USER]], {'get': True}
+                )
+                libvirt_group = await self.middleware.call('group.query', [['group', '=', LIBVIRT_USER]], {'get': True})
+                current_owner = os.stat(path)
+                is_valid = False
+                if current_owner.st_uid != libvirt_user['uid']:
+                    if await self.middleware.call('filesystem.can_access_as_user', LIBVIRT_USER, path, {'read': True}):
+                        is_valid = True
+                    else:
+                        os.chown(path, libvirt_user['uid'], libvirt_group['gid'])
```
themylogin, line 399: So using an ISO file from a data storage will result in file's owner change?

themylogin: Do we need to continue with this behavior? Seems a little bit erratic.

themylogin (approved):

PR #8023, opened 2021-12-19
NAS-113649 / 22.02 / Make sure walinuxagent only runs on azure

sonicaj:
This PR adds changes to ensure that walinuxagent only runs on azure and not just on hyper-v.


src/middlewared/middlewared/plugins/system_/system.py
```diff
@@ -17,3 +18,11 @@ async def vm(self):
             self.is_vm = p.stdout.strip() != "none"
 
         return self.is_vm
+
+    @private
+    async def running_in_azure(self):
```
themylogin, line 23: How about making this method `vm_provider` and returning `"AZURE"` or `None`?

themylogin (changes requested):

src/middlewared/middlewared/plugins/system_/system.py
```diff
@@ -0,0 +1,33 @@
...
+        if self.is_vm is None:
+            p = await run(["systemd-detect-virt"], check=False, encoding="utf-8", errors="ignore")
+            self.is_vm = p.stdout.strip() != "none"
+
+        return self.is_vm
+
+    @private
+    async def vm_provider(self):
+        if self.vm_hypervisor is None:
+            self.vm_hypervisor = VMProvider.NONE
```
themylogin, line 27: Why not normal `None`?

themylogin (approved):

src/middlewared/middlewared/plugins/system_/system.py
sonicaj: Just seemed neater to use the enum - no specific reason

PR #8027, opened 2021-12-21
Optionally expand get_user_obj output to include grouplist

anodos325:
There may be some situations where we want the full list
of groups of which the specified user is a member. Since
this potentially makes the call slightly more expensive,
make additional info optional and off by default

### DIFF
--- a/src/middlewared/middlewared/plugins/cache.py
+++ b/src/middlewared/middlewared/plugins/cache.py
@@ -342,7 +342,7 @@ class DSCache(Service):
         })
         return [x['val'] for x in entries]
 
-    def get_uncached_user(self, username=None, uid=None):
+    def get_uncached_user(self, username=None, uid=None, getgroups=False):
         """
         Returns dictionary containing pwd_struct data for
         the specified user or uid. Will raise an exception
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/cache.py
```diff
@@ -355,13 +356,15 @@ def get_uncached_user(self, username=None, uid=None):
...
+
         return {
             'pw_name': u.pw_name,
             'pw_uid': u.pw_uid,
             'pw_gid': u.pw_gid,
             'pw_gecos': u.pw_gecos,
             'pw_dir': u.pw_dir,
-            'pw_shell': u.pw_shell
+            'pw_shell': u.pw_shell,
+            'grouplist': os.getgrouplist(u.pw_name, u.pw_gid) if getgroups else []
```
sonicaj, line 367: I don't think we should be sending an empty list in the response as that potentially makes the consumer think that the list is in reality empty which is not true. Maybe not add the key at all or send `null` for distinction ?

anodos325: Sure. I can omit if it's not specified.

sonicaj (approved):

PR #8028, opened 2021-12-22
NAS-113937 / 22.02 / Improve executing commands with a user context implementation

sonicaj:
This PR adds changes to use python to change user context in a process and then having a reusable implementation which can be used in other places as required to execute certain functions as some user.

### DIFF
--- a/src/middlewared/middlewared/plugins/filesystem_/perm_check.py
+++ b/src/middlewared/middlewared/plugins/filesystem_/perm_check.py
@@ -11,10 +11,10 @@ from middlewared.service import CallError, Service
 from middlewared.utils.osc import set_user_context
 
 
-def check_perms(user: str, path: str, check_perms: dict, pipe: Connection) -> None:
+def check_access(user: str, path: str, check_perms: dict, pipe: Connection) -> None:
     set_user_context(user)
 
-    flag = False
+    flag = True
     for perm, check_flag in filter(
         lambda v: v[0] is not None, (
             (check_perms['read'], os.R_OK),
@@ -23,10 +23,9 @@ def check_perms(user: str, path: str, check_perms: dict, pipe: Connection) -> No
         )
     ):
         perm_check = os.access(path, check_flag)
-        flag |= (perm_check if perm else not perm_check)
+        flag &= (perm_check if perm else not perm_check)
 
     pipe.send(flag)
-    pipe.close()
 
 
 class FilesystemService(Service):
@@ -58,9 +57,13 @@ class FilesystemService(Service):
             raise CallError('At least one of read/write/execute flags must be set', errno.EINVAL)
 
         parent_con, child_con = Pipe()
-        proc = Process(target=check_perms, args=(username, path, perms, child_con), daemon=True)
-        proc.start()
-        can_access = parent_con.recv()
-        parent_con.close()
-        proc.join()
+        try:
+            proc = Process(target=check_access, args=(username, path, perms, child_con), daemon=True)
+            proc.start()
+            can_access = parent_con.recv()
+            proc.join()
+        finally:
+            child_con.close()
+            parent_con.close()
+
         return can_access
--- a/src/middlewared/middlewared/plugins/filesystem_/perm_check.py
+++ b/src/middlewared/middlewared/plugins/filesystem_/perm_check.py
@@ -2,18 +2,13 @@ import errno
 import os
 import pathlib
 
-from multiprocessing import Pipe, Process
-from multiprocessing.connection import Connection
-
 from middlewared.schema import accepts, Bool, Dict, returns, Str
 from middlewared.service import CallError, Service
 
-from middlewared.utils.osc import set_user_context
+from middlewared.utils.osc import run_with_user_context
 
 
-def check_access(user: str, path: str, check_perms: dict, pipe: Connection) -> None:
-    set_user_context(user)
-
+def check_access(path: str, check_perms: dict) -> bool:
     flag = True
     for perm, check_flag in filter(
         lambda v: v[0] is not None, (
@@ -25,7 +20,7 @@ def check_access(user: str, path: str, check_perms: dict, pipe: Connection) -> N
         perm_check = os.access(path, check_flag)
         flag &= (perm_check if perm else not perm_check)
 
-    pipe.send(flag)
+    return flag
 
 
 class FilesystemService(Service):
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/utils/osc/linux/user_context.py
```diff
@@ -1,25 +1,94 @@
...
+            (os.getgid, user_details.pw_gid),
+            (os.getegid, user_details.pw_gid),
+        )
+    ):
+        raise Exception(f'Unable to set user context to {user!r} user')
+
+    try:
+        os.chdir(user_details.pw_dir)
+    except Exception:
+        os.chdir('/')
```
anodos325, line 36: I don't believe cwd should be relevant for this, so perhaps chdir() into /var/empty?

src/middlewared/middlewared/plugins/filesystem_/perm_check.py
```diff
@@ -1,10 +1,32 @@
...
+
+    flag = False
+    for perm, check_flag in filter(
+        lambda v: v[0] is not None, (
+            (check_perms['read'], os.R_OK),
+            (check_perms['write'], os.W_OK),
+            (check_perms['execute'], os.X_OK),
+        )
+    ):
+        perm_check = os.access(path, check_flag)
```
yocalebo, line 25: Do we need to wrap this in a try: except so that on any crash this side of the pipe is closed and we don't leave a dangling FD? Or maybe move the closing of the pipe to the caller?

sonicaj: Definitely, thanks for pointing it out

src/middlewared/middlewared/plugins/filesystem_/perm_check.py
```diff
@@ -1,10 +1,32 @@
...
+from multiprocessing.connection import Connection
+
 from middlewared.schema import accepts, Bool, Dict, returns, Str
 from middlewared.service import CallError, Service
 
-from middlewared.utils.osc import run_command_with_user_context
+from middlewared.utils.osc import set_user_context
+
+
+def check_perms(user: str, path: str, check_perms: dict, pipe: Connection) -> None:
```
anodos325, line 14: Maybe check_access() rather than check_perms() since former is somewhat more like what we're doing.

themylogin:
I understand why this is necessary for `can_access_as_user` (although in both cases we spawn a subprocess and I doubt that python subprocess is lighter than sh), but how is new `run_command_with_user_context` better than just using `sudo`?

sonicaj:
@themylogin the suggestion by @yocalebo was that we should consider using python ( https://github.com/truenas/middleware/pull/8022#issuecomment-997934971 ) and as we have a generic method to set user context, the implementation for `run_command_with_user_context ` was updated as well to consume the same endpoint.
If you think it's better that we should continue using the old subprocess call, i can make the required change.

themylogin:
@sonicaj I am not in the position to force you make changes back and forth. @yocalebo what do you think about my previous comment?

yocalebo:
> @sonicaj I am not in the position to force you make changes back and forth. @yocalebo what do you think about my previous comment?

The only reason why I brought this up is that the original implementation was 3 fork + execs

1 for `subprocess.Popen`
1 for `sudo`
1 for `sh -c commandline`

themylogin:
@yocalebo how many `fork`/`exec` calls do we have now? And is all the pipe proxying-marshalling-unmarshalling overhead (which growth with the amount of output) worth fixed initial cost?

yocalebo:
The related ticket for checking if `libvirt` has access was what sparked this conversation. I will agree that creating a `Pipe` and a `Queue` object are not needed for simply checking access to the underlying filesystem. It seems this has morphed into something more complicated than what I had originally had in mind.

I just wanted something simple like `filesystem.check_access`. The idea that I had was something like this. (Not using a process but uses its own thread since we're `os.setresuid` and friends.

```
def child_process(*args, **kwargs):
    os.setresuid
    os.setreguid
    etc, etc
    
    result = False
    for perm perms:
        result &= os.access(blah, blah)
        
def check_access(*args, **kwargs):
    with concurrent.futures.ThreadPoolExecutor() as exc:
        return exc.submit(child_process).result()
```

sonicaj:
@yocalebo i suspect doing this at thread level won't work - @anodos325 and i were talking about it before. @anodos325 can you please correct me if i am wrong ?

```
   C library/kernel differences
       At the kernel level, user IDs and group IDs are a per-thread
       attribute.  However, POSIX requires that all threads in a process
       share the same credentials.  The NPTL threading implementation
       handles the POSIX requirements by providing wrapper functions for
       the various system calls that change process UIDs and GIDs.
       These wrapper functions (including those for setresuid() and
       setresgid()) employ a signal-based technique to ensure that when
       one thread changes credentials, all of the other threads in the
       process also change their credentials.  For details, see nptl(7).
```

sonicaj:
@yocalebo i suspect doing this at thread level won't work - @anodos325 and i were talking about it before. @anodos325 can you please correct me if i am wrong ?

```
   C library/kernel differences
       At the kernel level, user IDs and group IDs are a per-thread
       attribute.  However, POSIX requires that all threads in a process
       share the same credentials.  The NPTL threading implementation
       handles the POSIX requirements by providing wrapper functions for
       the various system calls that change process UIDs and GIDs.
       These wrapper functions (including those for setresuid() and
       setresgid()) employ a signal-based technique to ensure that when
       one thread changes credentials, all of the other threads in the
       process also change their credentials.  For details, see nptl(7).
```

yocalebo:
> @yocalebo i suspect doing this at thread level won't work - @anodos325 and i were talking about it before. @anodos325 can you please correct me if i am wrong ?
> 
> ```
>    C library/kernel differences
>        At the kernel level, user IDs and group IDs are a per-thread
>        attribute.  However, POSIX requires that all threads in a process
>        share the same credentials.  The NPTL threading implementation
>        handles the POSIX requirements by providing wrapper functions for
>        the various system calls that change process UIDs and GIDs.
>        These wrapper functions (including those for setresuid() and
>        setresgid()) employ a signal-based technique to ensure that when
>        one thread changes credentials, all of the other threads in the
>        process also change their credentials.  For details, see nptl(7).
> ```

Yeah, I convinced myself that using a thread would work and it will but it changes all threads euid/gid so I updated the comment to use a `ProcessPoolExecutor` instead 😄

sonicaj:
Well circling back to the problem at hand - where do we stand now ?

themylogin:
@themylogin I can't speak for "me", but my proposal is:
* Keep `run_command_with_user_context` as it was before
* Add `run_with_user_context method that will run a function in a multiprocessing pool as in @yocalebo's comment
* Make `can_access_as_user` use that method

sonicaj:
@themylogin we can't have a generic method to which we pass a callable which we can expect to have user context set already as callables can't be serialized.
So we can do (1) and change (2)/(3) a bit to not use pipe and use result ( although i do suspect it would be doing something similar under the hood ) ?

themylogin:
@sonicaj we can create a process pool of one process, first submit a function that does `os.setresuid` and second submit the callable, won't it work?

themylogin (approved):

PR #8029, opened 2021-12-22
NAS-113583 / 22.02 / Allow expanding pools in scale

sonicaj:
Earlier we had an upstream issue where zfs pools could not be expanded before when pools were created using partition uuids, however in latest scale the issue does not exist anymore. While being there, freebsd specific code has been removed from the module as well.


yocalebo (approved):
Nice!

PR #8030, opened 2021-12-22
NAS-113795 / 22.02 / refactor network_alias table

yocalebo:
This fixes many problems:
1. as far back as I can glean from git history we have been incorrectly writing to the `network_alias` table on an HA system. If 2+ ip addresses are added to an interface, then we create a _separate_ entry for each controller IP and the VIP in the `network_alias` table (so 3 total entries in `network_alias` for every IP added to an interface) instead of making it 1 singular table entry in the db. I've written a migration script to fix this.
2. we designed the `network_alias` table to store an IPv4 _and_ an IPV6 address on a singular table entry. This made the backend logic overly complex and prone to many failure scenarios. Furthermore, it has never worked because of the issue I describe in step # 1. In the migration script, I'm redoing the `network_alias` table so that we do not store a v4 and/or a v6 entry in the table. Instead, there is an `alias_version` column which will be `4` or `6` respectively to represent the IP version. I've further simplified the columns in the `network_alias` table to make logic vastly simpler.
3. it's possible to have a `int_ipv4address` _AND_ an `int_ipv6address` entry in the `network_interfaces` table for an interface. However, we weren't properly handling this logic in `interface/configure.py` so I've resolved that problem.
4. `network_alias` table on SCALE HA systems were just ignored completely in `interface/configure.py` so I've fixed this issue
5. because of all these changes, it fixes a `KeyError` crash on SCALE HA when adding a 2nd (or more) ip to an interface

Finally, I've added changes so that `convert_aliases_to_datastores` can be unit tested and written the associated unit tests.  I'd much rather not have this situation happen again 😄

### DIFF
--- a/src/middlewared/middlewared/alembic/versions/22.02/2021-12-20_20-37_fix_network_alias.py
+++ b/src/middlewared/middlewared/alembic/versions/22.02/2021-12-20_20-37_fix_network_alias.py
@@ -1,0 +1,128 @@
+"""fix network_alias table
+
+Revision ID: 6e41203881b2
+Revises: a3f3b07bb1aa
+Create Date: 2021-12-20 20:37:29.496586+00:00
+
+"""
+import sqlalchemy as sa
+from alembic import op
+from ipaddress import ip_interface
+
+
+# revision identifiers, used by Alembic.
+revision = '6e41203881b2'
+down_revision = 'a3f3b07bb1aa'
+branch_labels = None
+depends_on = None
+
+
+def pull_out_entries(rows):
+    a_addresses = set()
+    b_addresses = set()
+    vips = set()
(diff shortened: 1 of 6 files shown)


yocalebo:
retest this please

yocalebo:
scale ci builder has to rebuild all packages from scratch, so I'll wait to retest until that is done.

yocalebo:
retest this please

yocalebo:
The build failure is erroneous. It successfully built here: http://ci.tn.ixsystems.com/jenkins/job/Pull%20Requests/job/TrueNAS%20SCALE/job/PR%20-%20freenas/4768/

themylogin (approved):
Great job! Can't find fault with anything

PR #8031, opened 2021-12-22
NAS-113776 / 12.0 / Fix TLS certificate issue on Minio web console

Qubad786:
We need to set `MINIO_SERVER_URL` in order to fix TLS issue with minio web console.

ref: https://docs.min.io/minio/baremetal/reference/minio-server/minio-server.html#envvar.MINIO_SERVER_URL
> MINIO_SERVER_URL may be necessary if the MinIO Server TLS certificates do not contain any IP Subject Alternative Names (SAN). Specifically, the Console uses the MinIO Server IP address by default. If the Server TLS does not contain that IP address, then the Console cannot validate the TLS connection.

With self-signed cert configured:
Screen Shot 2021-12-22 at 8 49 11 PM

### DIFF
--- a/src/middlewared/middlewared/alembic/versions/12.0/2021-12-22_12-59_add_server_tls_uri_to_s3_config.py
+++ b/src/middlewared/middlewared/alembic/versions/12.0/2021-12-22_12-59_add_server_tls_uri_to_s3_config.py
@@ -1,0 +1,57 @@
+"""Add tls_server_uri to s3 config
+
+Revision ID: 9c11f6c6f152
+Revises: fee786dfe121
+Create Date: 2021-12-22 12:59:17.737066+00:00
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from OpenSSL import crypto
+
+# revision identifiers, used by Alembic.
+revision = '9c11f6c6f152'
+down_revision = 'fee786dfe121'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table('services_s3', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('s3_tls_server_uri', sa.String(length=128), nullable=True))
+
+    # Try to get tls_server_uri in following order:
+    # 1. SAN from certificate
+    # 2. Common name from certificate
+    # 3. Fallback to localhost
+    conn = op.get_bind()
+    if s3_conf := conn.execute("SELECT s3_certificate_id FROM services_s3 WHERE s3_certificate_id IS NOT NULL").fetchone():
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/alembic/versions/12.0/2021-12-22_12-59_add_server_tls_uri_to_s3_config.py
```diff
@@ -0,0 +1,55 @@
...
+                map(lambda i: cert.get_extension(i), range(cert.get_extension_count()))
+                if isinstance(cert, crypto.X509)
+                else cert.get_extensions()
+            )):
+                if 'subjectAltName' == ext.get_short_name().decode():
+                    cert_san = [s.strip() for s in ext.__str__().split(',') if s]
+
+            s3_tls_server_uri = 'localhost'
+            if cert_san:
+                s3_tls_server_uri = cert_san[0].split(':')[-1]
```
sonicaj, line 43: Let's strip this as well because iirc we can have values like `DNS: somedomain`

src/middlewared/middlewared/alembic/versions/12.0/2021-12-22_12-59_add_server_tls_uri_to_s3_config.py
```diff
@@ -0,0 +1,55 @@
...
+        batch_op.add_column(sa.Column('s3_tls_server_uri', sa.String(length=128), nullable=True))
+
+    # Try to get tls_server_uri in following order:
+    # 1. SAN from certificate
+    # 2. Common name from certificate
+    # 3. Fallback to localhost
+    conn = op.get_bind()
+    if s3_conf := conn.execute("SELECT s3_certificate_id FROM services_s3 WHERE s3_certificate_id IS NOT NULL").fetchone():
+        if cert_data := conn.execute("SELECT cert_certificate FROM system_certificate WHERE id = :cert_id", cert_id=s3_conf[0]).fetchone():
+            cert = crypto.load_certificate(crypto.FILETYPE_PEM, cert_data[0])
```
sonicaj, line 30: I think we should have the parsing from cert logic in try/except - however I am also concerned of us missing out a case where this can potentially break and us saving `localhost` to database. @themylogin can you please advise if we should have this in try/except with defaulting to `localhost` in case of exception ?

themylogin: We should necessarily try/except here

themylogin (changes requested):

themylogin (approved):

themylogin (approved):

PR #8034, opened 2021-12-22
NAS-113494 / 22.02 / speed up disk wiping

yocalebo:
On CORE `os.fsync` doesn't get propagated down to the disk by design so it's a NO-OP. However, on SCALE it seems to do that on some disks which makes wiping disks abysmally slow. This isn't needed in either scenario so remove it.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/wipe.py
+++ b/src/middlewared/middlewared/plugins/disk_/wipe.py
@@ -38,20 +38,17 @@ class DiskService(Service):
                 for i in range(_32):
                     # wipe first 32MB
                     os.write(f.fileno(), to_write)
-                    os.fsync(f.fileno())
 
                 # seek to 32MB before end of drive
                 os.lseek(f.fileno(), (size - (CHUNK * _32)), os.SEEK_SET)
                 for i in range(_32):
                     # wipe last 32MB
                     os.write(f.fileno(), to_write)
-                    os.fsync(f.fileno())
             else:
                 iterations = (size // CHUNK)
                 length = len(str(iterations))
                 for i in range(iterations):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/disk_/wipe.py
```diff
@@ -38,20 +38,17 @@ def _wipe(self, data):
                 for i in range(_32):
                     # wipe first 32MB
                     os.write(f.fileno(), to_write)
-                    os.fsync(f.fileno())
```
sonicaj, line 41: I have following queries:
1) Why did we used to do this before ( my concern is introducing a regression by not doing it anymore ) ?
2) On freebsd `fsync(2)` does seem to make sure that this is propagated to disk ( https://www.freebsd.org/cgi/man.cgi?query=fsync&sektion=2&format=html )
3) What performance increase are we looking at ?

yocalebo: 1. if you look at history of this file you'll see that I rewrote this method from scratch so we were never doing `os.fsync` before. It was something I added when I rewrote it.
2. `fsync` to a character device is not propagated to disk the way you think it is on freeBSD
3. potentially many hours

sonicaj: I am confused, `fsync` to a character device - which one is that in our context 😅  ?

That is indeed a lot of time, so we never needed to use `fsync` ?

yocalebo: Right, we never needed fsync calls

sonicaj (approved):

PR #8046, opened 2021-12-23
NAS-113982 / 13.0 / add geom plugin and cache geom/xml info

yocalebo:
Add a private `GeomCache` plugin that caches the formatted geom information for disks as well as cache the raw xml data.

NOTE: These are not plumbed in or used anywhere at the moment. Later PR's will use these.

### DIFF
--- a/src/middlewared/middlewared/plugins/geom.py
+++ b/src/middlewared/middlewared/plugins/geom.py
@@ -1,0 +1,130 @@
+import re
+import threading
+from xml.etree import ElementTree as etree
+
+import sysctl
+from middlewared.service import Service
+from bsd.disk import get_ident_with_name
+
+
+class GeomCache(Service):
+    DISKS = {}  # formatted cache for geom DISKS (parsed xml)
+    XML = None  # raw xml cache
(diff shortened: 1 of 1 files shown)


themylogin (changes requested):
No need for an extra thread, Check out how https://github.com/truenas/middleware/blob/master/src/middlewared/middlewared/plugins/disk_/smartctl.py#L14 works - basically, the same thing

yocalebo:
> No need for an extra thread, Check out how https://github.com/truenas/middleware/blob/master/src/middlewared/middlewared/plugins/disk_/smartctl.py#L14 works - basically, the same thing

What's wrong with another thread?

themylogin:
@yocalebo why increase thread count in already bloated middleware just to sleep 99.99% of the time if we can avoid that?

yocalebo:
> @yocalebo why increase thread count in already bloated middleware just to sleep 99.99% of the time if we can avoid that?

That's a valid point. I'll fix, after I get back from vacation🙂

themylogin (approved):

rick-mesta (approved):

PR #8051, opened 2021-12-24
NAS-113641 / 22.02 / Fix display device URI

Qubad786:


sonicaj (commented):
If we have multiple display devices and the first one had a specific bind ip whereas the latter didn't, this would result in the latter using wrong host address.

sonicaj (approved):

PR #8057, opened 2021-12-27
NAS-113492 / 13.0 / Provide feedback when connection error occurs

Qubad786:


sonicaj (commented):
Let's please handle this in iocage so the feedback is improved for CLI users as well.

PR #8064, opened 2021-12-28
NAS-113811 / 12.0 / Resolve double-cabled disks issue

themylogin:


themylogin:
time 08:00

themylogin:
time 00:45

yocalebo (commented):
I have a few questions/comments.

1. Why do we have 2 alembic migrations when it's altering the same table? Can it not be simplified to 1 migration script?
2. Why are we adding `disk_bus` as I don't see it being used anywhere other than just adding it as more information to be returned.
3. If we do need to add `disk_bus`, why do we default to `UNKNOWN` or `USB`? If this value is raised in a `verror` then it'll probably be worthwhile to return that value? Or am I missing something obvious?

themylogin:
> Why do we have 2 alembic migrations when it's altering the same table? Can it not be simplified to 1 migration script?

Ok, I'll squash them

> Why are we adding disk_bus as I don't see it being used anywhere other than just adding it as more information to be returned?

It will be used in the UI to distinguish an error/advice message between poorly-cabled SATA disks and USB disks reporting bogus serial numbers

> If we do need to add disk_bus, why do we default to UNKNOWN or USB?

Currently we only care about disk being USB or not. More buses can be added in the future, I didn't want to get involved with detecting all possible buses for 12.0.

> If this value is raised in a verror

It's not

yocalebo:
> Currently we only care about disk being USB or not. More buses can be added in the future, I didn't want to get involved with detecting all possible buses for 12.0.
> 

But this means we're going to be writing `UNKNOWN` to the database for all enterprise paying customers when we have an alert that specifically warns them when a USB disk is plugged into the system?

yocalebo (approved):

rick-mesta:
@themylogin: Is there a way we can minimize the scope of the changes to only address the _fix_ and leave any refactoring and/or test related changes for 13 ? We would like to introduce as minimal changes to U7.1 as possible, to make for a more stable release.

yocalebo:
> @themylogin: Is there a way we can minimize the scope of the changes to only address the _fix_ and leave any refactoring and/or test related changes for 13 ? We would like to introduce as minimal changes to U7.1 as possible, to make for a more stable release.

I agree, I kind of glanced over the fact that this was for the 12 repo 😬

themylogin:
backport

themylogin:
https://www.meme-arsenal.com/memes/02e005908f7acbe3e260cb1c7aab799c.jpg

PR #8067, opened 2021-12-28
NAS-113853 / 22.02 / Do not restart all the services when removing vdev from pool

themylogin:


src/middlewared/middlewared/plugins/disk_/wipe.py
```diff
@@ -54,7 +54,7 @@ def _wipe(self, data):
     @accepts(
         Str('dev'),
         Str('mode', enum=['QUICK', 'FULL', 'FULL_RANDOM'], required=True),
-        Bool('synccache', default=True),
```
sonicaj, line 57: We shouldn't change the name of the variable as existing users consuming this via API would be affected by this change

themylogin: These are positional arguments, they are never referenced by their names

src/middlewared/middlewared/plugins/s3.py
```diff
@@ -72,6 +73,7 @@ async def config_extend(self, s3):
         ('edit', {'name': 'secret_key', 'method': lambda x: setattr(
             x, 'validators', [Match(r'^\w+$', explanation='Should only contain alphanumeric characters')]
         )}),
+        ('rm', {'name': 'tls_server_uri'}),
```
sonicaj, line 76: I am not sure i follow this change, we should be able to change this and by doing this - we wouldn't be anymore ?

themylogin: It was not included in the entry schema so I decided it was not editable. Should it be?

sonicaj: Yes, it definitely is editable :)

sonicaj (approved):
Thanks!

yocalebo (commented):
Why change from os.SEEK_SET which is literally 0?

themylogin:
@yocalebo this is a position argument, I don't understand why was `os.SEEK_SET` used instead of a zero.

themylogin:
time 0:30

PR #8077, opened 2021-12-30
NAS-113892 / 22.02 / Improve feedback on manual SMART test for nvme disks

Qubad786:
```
❯ midclt call -job smart.test.manual_test '[{"identifier": "{serial}xxx", "type": "SHORT"}]'
[EINVAL] disks.0.identifier: {serial}xxx is an nvme disk. We do not support S.M.A.R.T test for nvme disks.
```

### DIFF
--- a/src/middlewared/middlewared/plugins/smart.py
+++ b/src/middlewared/middlewared/plugins/smart.py
@@ -377,17 +377,22 @@ class SMARTTestService(CRUDService):
                 'Please specify at least one disk.'
             )
         else:
-            disks_choices = await self.disk_choices(True)
+            supported_disks = await self.disk_choices(True)
             devices = await self.middleware.call('device.get_storage_devices_topology')
-
+            valid_disks = [
+                disk['identifier']
+                for disk in await self.middleware.call('disk.query', [
+                    ('identifier', 'in', [disk['identifier'] for disk in disks])
+                ], {'force_sql_filters': True})
+            ]
             for index, disk in enumerate(disks):
-                if current_disk := disks_choices.get(disk['identifier']):
+                if current_disk := supported_disks.get(disk['identifier']):
                     test_disks_list.append({
                         'disk': current_disk['name'],
                         **disk
                     })
                 else:
-                    if await self.middleware.call('disk.query', [('identifier', '=', disk['identifier'])]):
+                    if disk['identifier'] in valid_disks:
                         verrors.add(
                             f'disks.{index}.identifier',
                             f'{disk["identifier"]} does not support S.M.A.R.T test.'
--- a/src/middlewared/middlewared/plugins/smart.py
+++ b/src/middlewared/middlewared/plugins/smart.py
@@ -387,10 +387,16 @@ class SMARTTestService(CRUDService):
                         **disk
                     })
                 else:
-                    verrors.add(
-                        f'disks.{index}.identifier',
-                        f'{disk["identifier"]} is not valid. Please provide a valid disk identifier.'
-                    )
+                    if await self.middleware.call('disk.query', [('identifier', '=', disk['identifier'])]):
+                        verrors.add(
+                            f'disks.{index}.identifier',
+                            f'{disk["identifier"]} does not support S.M.A.R.T test.'
+                        )
+                    else:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/smart.py
```diff
@@ -387,10 +387,17 @@ async def manual_test(self, disks):
...
                 else:
-                    verrors.add(
-                        f'disks.{index}.identifier',
-                        f'{disk["identifier"]} is not valid. Please provide a valid disk identifier.'
-                    )
+                    disk_info = await self.middleware.call('disk.query', [('identifier', '=', disk['identifier'])], {'get': True})
+                    if disk_info['devname'].startswith('nvme'):
+                        verrors.add(
+                            f'disks.{index}.identifier',
+                            f'{disk["identifier"]} is an nvme disk. We do not support S.M.A.R.T test for nvme disks.'
```
anodos325, line 394: Perhaps avoid usage of "We" in validation message.

themylogin (changes requested):
Can this be consistent with `smart.test.disk_choices` please? I.e. there is only one method that determines whether a disk is suitable for a S.M.A.R.T. test or not (and why).

src/middlewared/middlewared/plugins/smart.py
Qubad786: would the following work?
```
S.M.A.R.T tests are not supported for nvme disks for now.
```

Qubad786: I see its an experimental feature, we cannot say that its supported on their side.. 🤔

src/middlewared/middlewared/plugins/smart.py
```diff
@@ -387,10 +387,19 @@ async def manual_test(self, disks):
...
-                        f'disks.{index}.identifier',
-                        f'{disk["identifier"]} is not valid. Please provide a valid disk identifier.'
-                    )
+                    if await self.middleware.call('disk.query', [
+                        ('identifier', '=', disk['identifier']),
+                        ('devname', '^', 'nv')
+                    ]):
+                        verrors.add(
+                            f'disks.{index}.identifier',
+                            f'{disk["identifier"]} is an nvme disk. S.M.A.R.T tests are not supported for nvme disks right now.'
```
anodos325, line 396: S.M.A.R.T tests are not currently supported for nvme disks.

Qubad786:
At this point, I think we should just improve the message as we could just say that disk identifier does not support S.M.A.R.T test..

yocalebo (approved):

src/middlewared/middlewared/plugins/smart.py
```diff
@@ -389,7 +389,7 @@ async def manual_test(self, disks):
                 else:
                     verrors.add(
                         f'disks.{index}.identifier',
-                        f'{disk["identifier"]} is not valid. Please provide a valid disk identifier.'
+                        f'{disk["identifier"]} does not support S.M.A.R.T test.'
```
sonicaj, line 392: In case an invalid identifier is supplied, we should raise an appropriate message that the supplied identifier is invalid. So let's please have this separately for the case where the disk identifier is valid but the disk does not support SMART test.

src/middlewared/middlewared/plugins/smart.py
```diff
@@ -387,10 +387,16 @@ async def manual_test(self, disks):
                         **disk
                     })
                 else:
-                    verrors.add(
-                        f'disks.{index}.identifier',
-                        f'{disk["identifier"]} is not valid. Please provide a valid disk identifier.'
-                    )
+                    if await self.middleware.call('disk.query', [('identifier', '=', disk['identifier'])]):
```
sonicaj, line 390: We should probably avoid calling `disk.query` for each disk and rather call this once above and save the result and consume them or maybe otherwise use `force_sql_filters` in query which would optimize this call.
```
❯ midclt call disk.query '[["identifier", "=", "{uuid}0f723d8d-29a9-4296-b215-a8723aac88c2"]]' '{"extra": {"force_sql_filters": true}}'
```

themylogin (approved):

PR #8082, opened 2022-01-03
NAS-114116 / 13.0 / Make minio console port configurable

Qubad786:
>MinIO by default selects a random port for the MinIO Console on each server startup. Browser clients accessing the MinIO Server are automatically redirected to the MinIO Console on its dynamically selected port. This behavior emulates the legacy web browser behavior while reducing the the risk of a port collision on systems which were running MinIO before the embedded Console update.
>
>You can select an explicit static port by passing the minio server --console-address commandline option when starting each MinIO Server in the deployment.

ref: https://docs.min.io/minio/baremetal/console/minio-console.html#static-vs-dynamic-port-assignment

### DIFF
--- a/src/middlewared/middlewared/alembic/versions/12.0/2022-01-03_11-46_add_s3_console_bindport.py
+++ b/src/middlewared/middlewared/alembic/versions/12.0/2022-01-03_11-46_add_s3_console_bindport.py
@@ -1,0 +1,26 @@
+"""
+Add s3 console bindport
+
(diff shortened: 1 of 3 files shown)


Qubad786:
That'd be backported to 12.0-stable and I'll be cherry-picking + fix into master as well.

themylogin (approved):

rick-mesta (approved):

PR #8083, opened 2022-01-03
NAS-114031 / 22.02 / Correctly use tag reference when updating chart release images

sonicaj:
When we have a chart release consuming an image like `nextcloud:23`, if we normalize the tag, we have something like `registry-1.docker.io/library/nextcloud:23` which is how it's resolved internally by docker but the actual tag reference is still `nextcloud:23`.

To clarify, if we try to update app's images, we end up pulling  `registry-1.docker.io/library/nextcloud:23`, however the app is consuming `nextcloud:23` which is still pointing to the old image.

This PR introduces changes to not use the parsed reference but rather consume the app specified tag. It also introduces changes to not try and pull images which have digest as the tag as that doesn't help/work.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/upgrade.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/upgrade.py
@@ -417,6 +417,10 @@ class ChartReleaseService(Service):
         )
         for reference in results['resources']['container_images']:
             parsed_reference = await self.middleware.call('container.image.normalize_reference', reference)
+            if parsed_reference['reference_is_digest']:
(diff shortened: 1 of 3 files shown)


Qubad786 (approved):
LGTM – That also solves: https://jira.ixsystems.com/browse/NAS-113972 – you can take it if haven't created the ticket already. 😉

Qubad786 (approved):
LGTM – That also solves: https://jira.ixsystems.com/browse/NAS-113972 – you can take it if haven't created the ticket already. 😉

shadowempire123:
also solves https://jira.ixsystems.com/browse/NAS-114300 :)

PR #8085, opened 2022-01-04
NAS-114025 / 22.02 / Make apt binaries non-executable

sonicaj:
This commit adds changes to make apt binaries non-executable to make it harder for people to footshoot themselves by installing packages or removing them.


themylogin (changes requested):
Won’t chmod -x /usr/bin/apt* work? Also let’s add a comment on why we are doing this. The fact that chmod -x removes execution bit is known already.

PR #8100, opened 2022-01-06
NAS-114177 / 12.0 / fix disk.sync with multipath disks

yocalebo:


themylogin (commented):
Should `sync_all` behavior also be changed?

What if `disk` currently has another `serial_lunid` and it is the database that has just not been updated yet?

yocalebo:
> Should `sync_all` behavior also be changed?

I'd rather not touch that method. On the system that reproduced the problem, when running `sync_all` it simply updated the names of the disks and kept all the other information so we didn't lose the zpool mapping information.

> What if `disk` currently has another `serial_lunid` and it is the database that has just not been updated yet?

Then the logic I added won't be executed because it'll be considered a new disk

themylogin:
> Then the logic I added won't be executed because it'll be considered a new disk

Why? The database entry for the specific `disk_identifier` (`{serial_lunid}xxx_yyy`) will be returned so we will enter `if ident and qs:` branch

rick-mesta:
@themylogin / @yocalebo: what's the status of this bad boy, gents. As noted in the team channel, `12.0-U8` should be blockers only, so please advise if this should go in 🙏🏼

yocalebo:
> > Then the logic I added won't be executed because it'll be considered a new disk
> 
> Why? The database entry for the specific `disk_identifier` (`{serial_lunid}xxx_yyy`) will be returned so we will enter `if ident and qs:` branch

My logic only gets executed if the database disk name doesn't match the newly found name and both of those disks exist. No logic should have changed other than when both disks exist then they are skipped. (i.e.  `/dev/da5` and `/dev/da17` exist and have same serial and lunid)

rick-mesta:
@themylogin: has @yocalebo addressed your issues ? This is one of three issues holding up U8

themylogin:
@yocalebo ok, let's consider an example. Here is a `storage.disk` table state

```
disk_identifier            | disk_name
{serial_lunid}XXX_000      | ada0
```

and here are currently present devices

```
name | serial | lunid
ada0 | YYY    | 111
ada1 | XXX    | 000
```

We call

```
call("disk.sync", "ada1")
```

Your code will not update table, because:
* `name = "ada1"`
* `ident = "{serial_lunid}XXX_000"`
* `disk = {"disk_identifier": "{serial_lunid}XXX_000", "disk_name": "ada0"}`
* `disk["disk_name"] != name`
* `self.both_disks_exist("ada0", "ada1") is True`

Is this the desired behavior?

yocalebo:
How does `ada0` end up in the system with a different lunid and serial but the database is outdated? Another event would had to have been processed before your hypothetical situation would occur right? If `ada0` has `XXX_000` but then changes to `YYY_111` a remove and then another add event would have to occur, no?

themylogin:
@yocalebo before your changes `disk.sync` was not relying on a specific database state to proceed successfully. Can we keep it this way?

yocalebo:
> @yocalebo before your changes `disk.sync` was not relying on a specific database state to proceed successfully. Can we keep it this way?

That's the entire problem. This is wiping out database state that is valid.

themylogin:
@yocalebo what do you mean by "this"?

yocalebo:
@themylogin I'm working with QE again on this problem. I've got logs from a live system and will review again. The culprit for causing all the problems is `enclosure.sync_disk` but it's because `disk.sync` is updating the database entry with a new `da` name. Therefore, the `self._get_slot_for_disk` method fails because the new name doesn't actually exist in the enclosure information because it's being hidden from `enclosure.query` output because we don't show duplicate enclosure ids (enclosures cabled for multipath)

yocalebo:
@themylogin I've made it so that we only `return` in `disk.sync` if `da` names do not match, both `da` devices exist and the serial_lunid match between the disk we received the event for and the disk that is returned in the db.

themylogin (approved):
That looks right, thank you!

rick-mesta (approved):

PR #8101, opened 2022-01-06
NAS-113727 / 13.0 / Fix a deadlock situation when creating many snapshots simultaneously

themylogin:
@rick-mesta this is another candidate that might be worth including in 12.0. It prevents WebUI from hanging when many snapshots are being created simultaneously using middleware (user had 10+ cloud sync tasks using "create snapshot" option firing up at the same time).

### DIFF
--- a/src/middlewared/middlewared/service.py
+++ b/src/middlewared/middlewared/service.py
@@ -518,6 +518,20 @@ class CRUDService(ServiceChangeMixin, Service):
         """
         return await self._get_instance(id)
 
+    @private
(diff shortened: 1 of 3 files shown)


themylogin:
time 7:00

rick-mesta (approved):
LGTM: Id like to hear from @yocalebo or @sonicaj just to make sure I'm not glossing over something important

yocalebo (approved):
Nice!

themylogin:
backport

PR #8102, opened 2022-01-06
NAS-114178 / 12.0 / fix enclosure plugin with multipath

yocalebo:
Prevent duplicate enclosures from being displayed so webUI doesn't break.


rick-mesta (approved):
I see `seen_enclosures` set being built, but not used. Where is it used ?

yocalebo:
> I see `seen_enclosures` set being built, but not used. Where is it used ?

On line 67

rick-mesta:
Heh... my point was that once you've created the set of unique enclosures, where is that used later on 😁 ? Or is that all you use it for, to filter duplicates ?

yocalebo:
> Heh... my point was that once you've created the set of unique enclosures, where is that used later on 😁 ? Or is that all you use it for, to filter duplicates ?

Yeah just to filter duplicates.

PR #8106, opened 2022-01-06
NAS-114187 / 22.02 / Add allow interfaces directive to avahi config

anodos325:
Deny interfaces will be interpreted before allow interfaces.
This will prevent avahi from registering veth and macvtap interfaces.


rick-mesta (approved):

yocalebo (changes requested):
Typo `middleare.call_sync`

I thought `macvtap` interfaces were in the `deny_interfaces` variable?

anodos325:
> Typo `middleare.call_sync`
> 
> I thought `macvtap` interfaces were in the `deny_interfaces` variable?

`interface.internal_interfaces` returns `macvtap`, but not `macvtap0`, `macvtap1`, etc. Hence, it's not particularly useful in this case for the deny list.

yocalebo (approved):

PR #8111, opened 2022-01-07
NAS-114005 / 22.02 / Add a flag to clarify if a pci device has reset mechanism defined

sonicaj:
This will be useful when user adds a PCI device for passthrough and UI can raise an alert/warning letting the user know that depending on the device vendor it's possible user might experience issues with the VM when it stops/starts as there is no reset mechanism defined by the PCI device.


src/middlewared/middlewared/plugins/vm/pci.py
```diff
@@ -108,9 +111,13 @@ async def passthrough_device(self, device):
 
         return {
             **node_info,
+            'device_path': data['device_path'],
             'drivers': drivers,
             'available': not error_str and all(d == 'vfio-pci' for d in drivers),
             'error': f'Following errors were found with the device:\n{error_str}' if error_str else None,
+            'reset_mechanism_defined': await self.middleware.run_in_thread(
```
themylogin, line 118: Why bother with threads when reading from sysfs?

themylogin: Also these new fields should be added to the `returns` schema

themylogin (changes requested):

themylogin (approved):

PR #8112, opened 2022-01-10
NAS-114223 / 22.02 / Do not take snapshot of docker dataset when making a backup of apps

sonicaj:
Docker dataset is re-initialized from scratch as containers are ephemeral in kubernetes during restore. Docker zfs driver creates at least 1 dataset per layer in container/image if not more where we end up with a huge number of datasets. This change makes sure we do not take a snapshot of the docker dataset as we end up with potentially thousands of snapshots over time.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/backup.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/backup.py
@@ -79,9 +79,18 @@ class KubernetesService(Service):
 
         job.set_progress(95, 'Taking snapshot of ix-applications')
 
-        self.middleware.call_sync(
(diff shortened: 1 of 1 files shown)


themylogin (changes requested):
This will be an inconsistent snapshot. Let me introduce an internal method to create recursive snapshots that exclude some of the datasets.

themylogin:
@sonicaj https://github.com/truenas/middleware/pull/8114

 `midclt call zettarepl.create_recursive_snapshot_with_exclude tank test-snapshot '["tank/.system"]'`

themylogin (approved):

Ornias1993 (approved):
Yeah, great cleanup!
Saves us a few thousand snapshots after a few backups :)

PR #8114, opened 2022-01-10
NAS-114228 / 22.02 / zettarepl.create_recursive_snapshot_with_exclude

themylogin:


sonicaj (commented):
How about having this under `zfs.snapshot` namespace ?

themylogin:
This imports `zettarepl` and I prefer to keep such methods in a special place

sonicaj:
@themylogin we can call zettarepl plugin internally from the `zfs.snapshot` namespace ?

sonicaj:
It's just that, from consumer perspective - it makes more sense to be consuming this from `zfs.snapshot` namespace

themylogin:
Done, but internally it'll still be better to use zettarepl method directly to avoid using process pool

themylogin:
time 0:30

sonicaj (approved):

PR #8115, opened 2022-01-10
NAS-114092 / 22.02 / use /proc for tracking processes using a path

yocalebo:
There should be no functional change. This does essentially the same as what lsof does but instead gets the information directly from `/proc`. This gets rid of subprocessing to `lsof` which is causing 100% cpu usage on certain systems. Also gets rid of a call to `psutil`.

### DIFF
--- a/tests/api2/test_pool_dataset_track_processes.py
+++ b/tests/api2/test_pool_dataset_track_processes.py
@@ -1,0 +1,25 @@
+import sys
+import os
+sys.path.append(os.getcwd())
+
+from functions import POST, make_ws_request
+from auto_config import pool_name, ip
+from middlewared.test.integration.utils import ssh
+
+TEST_DATASET = f'{pool_name}/testing_processes'
+
+
+def test_01_create_dataset():
+    results = POST('/pool/dataset', {'name': TEST_DATASET})
+    assert results.status_code == 200, results.text
+
+
+def test_02_open_path_and_check_proc():
+    path = f'/mnt/{TEST_DATASET}'
+    test_file = f'{path}/test_file'
+    open_pid = ssh(f'python -c "import time; f = open(\"{test_file}\", \"w+\"); time.sleep(10)" & echo $!')
+    assert open_pid.isdigit()
+
+    payload = {'msg': 'method', 'method': 'pool.dataset.processes_using_paths', 'params': list(path)}
+    res = make_ws_request(ip, payload)
+    assert int(res[0]['pid']) == int(open_pid), res
--- a/tests/api2/test_pool_dataset_track_processes.py
+++ b/tests/api2/test_pool_dataset_track_processes.py
@@ -1,25 +1,37 @@
 import sys
 import os
+import time
 sys.path.append(os.getcwd())
 
-from functions import POST, make_ws_request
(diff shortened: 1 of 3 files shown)


themylogin (changes requested):
Can we move this into a separate file? `pool.py` is already 4k lines long.

Also this method is a perfect candidate for an integration test.

yocalebo:
> Can we move this into a separate file? `pool.py` is already 4k lines long.

Sure

> Also this method is a perfect candidate for an integration test.

I'm not sure how an integration test would work for a method like this since it's tracking processes that have an open object underneath the path. Can you enlighten me on mock up behavior like that?

themylogin:
Create a test dataset, run `python -c 'import time; f = open("/mnt/tank/test/file", "w"); time.sleep(10)' & echo $!` and assert that PID being one and only one in `pool.dataset.processes`

tests/api2/test_pool_dataset_track_processes.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def test_02_open_path_and_check_proc():
+    test_file = f'/mnt/{TEST_DATASET}/test_file'
+    open_pid = ssh(f'python -c "import time; f = open(\"{test_file}\", \"w+\"); time.sleep(10)" & echo $!')
+    assert open_pid.isdigit()
+
+    payload = {'msg': 'method', 'method': 'pool.dataset.processes_using_paths', 'params': list(TEST_DATASET)}
+    res = make_ws_request(ip, payload)
+    assert int(res[0]['pid']) == int(open_pid)
```
themylogin, line 24: Great job, thank you for adding tests!

Can we please also:
* assert `name` and `cmdline` contents since the code under test is responsible for filling them
* assert that `len(res) == 1` to ensure that code under test does not add extra processes to the list
* kill that python process after we're done with the test
* destroy the test dataset after we're done with the test. The easiest way is to use `with dataset(...)` decorator like my other tests do and skip having fake tests like `test_01_create_dataset` completely. WIth that in mind we may stop adding numbers to the test names because they became independent units which execution order does not matter (like it should be ideally).

Also it might be nice to:
* sleep for some time between launching python process and asserting an open file because it might take quite a time for python interpreter to initialize if testing VM host is busy
* you might like how `from middlewared.test.integration.utils import call` works (just `call("method", arg1, arg2) `)

themylogin (changes requested):

tests/api2/test_pool_dataset_track_processes.py
```diff
@@ -0,0 +1,37 @@
...
+    with dataset(TEST_DATASET) as ds:
+        opened = False
+        try:
+            path = f'/mnt/{ds}'
+            test_file = f'{path}/test_file'
+            cmdline = f'python -c "import time; f = open(\"{test_file}\", \"w+\"); time.sleep(10)" & echo $!'
+            open_pid = ssh(cmdline)
+            assert open_pid.isdigit(), open_pid
+            opened = True
+
```
themylogin, line 24: You can start `try` block here and no `opened` variable would be needed

themylogin (approved):
Great test, thank you!

PR #8119, opened 2022-01-10
NAS-114231 / 13.0 / fix off-by-one in Mini XL enclosure mapping

yocalebo:


themylogin:
Is that it? IIRC the problem looked like "0-indexed" vs "1-indexed"

yocalebo:
> Is that it? IIRC the problem looked like "0-indexed" vs "1-indexed"

I missed my 2nd commit in local repo. The problem is that it started at `0` instead of starting at `1` for the mapping.

themylogin:
Cool! Another question that tortured me: why did it change between FreeBSD 12 and 13? And why only Mini XL is affected?

yocalebo:
> Cool! Another question that tortured me: why did it change between FreeBSD 12 and 13? And why only Mini XL is affected?

13 was rewritten so it would scale on systems with 100's of hard drives. When rewriting the plugin, I found that there was a disk element slot that was ignored but actually took up position `0` in the array. Instead of ignoring that element, I always start with an actual disk element and not the categorical identifier. For example:
```
/dev/ses0: Enclosure Status <OK>
Element 0x0: Array Device Slot, status: Unsupported (0x00 0x00 0x00 0x00), descriptor: 'Drive Slots'
```

As you can see that's an `Array Device Slot` but it's actually not. It's the "categorical identifier" for the drive slots. I ignore that element and start with an element that actually can hold a hard drive.

themylogin:
@yocalebo do we need to update all mappings in that case?

yocalebo:
> @yocalebo do we need to update all mappings in that case?

I already did when I committed the original PR a few months ago.

themylogin:
There is a number of mappings that have `MappingSlot(0, 0, False),` (e.g. MINI-X+), do they not need similar change?

yocalebo:
> There is a number of mappings that have `MappingSlot(0, 0, False),` (e.g. MINI-X+), do they not need similar change?

I'll work with wolf for the other `MINI` hardware mappings. This ticket will stay as is since it's fixing the specific problem for MINI XL's.

anodos325 (approved):

PR #8120, opened 2022-01-10
NAS-114239 / 12.0 / Add alert for CVE-2021-20316-related misconfiguration

anodos325:


themylogin (changes requested):
This alert will still be present for one day even after the configuration is changed to a secure one.

Ideally this should be a one-shot alert that will be checked on NFS and SMB configuration generation.

anodos325:
> This alert will still be present for one day even after the configuration is changed to a secure one.
> 
> Ideally this should be a one-shot alert that will be checked on NFS and SMB configuration generation.

 I'd rather not run it every time we generate those config files (thinking of service.query, etc).

themylogin:
@anodos325 what does `service.query` has to do with this?

anodos325:
Okay. Have consulted support for what they want, and it looks like preference toward oneshot.

rick-mesta (approved):

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -268,6 +286,30 @@ def key_cp(encoding):
             ] + initial
         }
 
+    @private
+    @periodic(interval=86400)
```
themylogin, line 290: What is the point of doing this every day?

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -268,6 +286,30 @@ def key_cp(encoding):
...
+    async def do_smb_alerts(self):
+        await self.unix_extensions_alert()
+        await self.middleware.call('sharing.smb.share_configuration_alert')
+
+    @private
+    async def unix_extensions_alert(self):
+        smb_running = await self.middleware.call('service.started', 'cifs')
+        if not smb_running:
+            del_job = await self.middleware.call('alert.oneshot_delete', 'UnsupportedUnixExtensions', None)
+            await del_job.wait()
```
themylogin, line 300: Why wait for this? It might involve sending alerts, which might involve networking, which might take a long time.

themylogin (changes requested):
What about when the system with invalid configuration first starts? Or when a service that has invalid configuration is first started? That's why I proposed to do this in configuration generation routine like other oneshot alerts go (e.g. nginx SSL certificates)

src/middlewared/middlewared/plugins/smb.py
anodos325: Good point. I didn't think that through.

src/middlewared/middlewared/plugins/smb.py
anodos325: Checking the running configuration in case users edit via CLI. Thought it's not that impactful to have a periodic check.

themylogin: @anodos325 what CLI?

rick-mesta:
@themylogin : are we good to go here ?

src/middlewared/middlewared/plugins/smb.py
anodos325: Manual edits of smb running configuration through text editor or `net conf` utility.

themylogin: Do we support this? Why?

themylogin (commented):
@anodos325 now I see how this works on boot (side effect of `@periodic`) but I still don't see who creates this alert if a disable service is started manually.

src/middlewared/middlewared/plugins/smb.py
themylogin: Do we support this? Why?

anodos325:
> @anodos325 now I see how this works on boot (side effect of `@periodic`) but I still don't see who creates this alert if a disable service is started manually.

My thought was that the periodic tests would pick up this case. I was disinclined to have checks on every NFS / SMB service start / stop.

rick-mesta:
Let's come to some agreement, @themylogin -- this continues to hold up u8

themylogin:
@anodos325 it will pick up a day later in the worst case

What is the issue with running checks in configuration generation routine like other oneshot alerts go? (e.g. nginx SSL certificates)

mustanggb:
This is fixed in Samba 4.15.0, so is it planned to remove this alert in TrueNAS 13.0?

PR #8121, opened 2022-01-10
NAS-112667 / 13.0 / Set default max_parallel_replication_tasks to 5

themylogin:
Inifinite default value seems to crash on FreeBSD (see
https://jira.ixsystems.com/browse/NAS-112667).

Also, there is a very few cases where running many replications in parallel
would benefit performance. If the network is bandwidth-limited for a single
connection, 5 parallel connections are usually enough to utilize its full capacity.


rick-mesta (approved):
After discussion with @themylogin, we believe the root cause of the issue might be in the threads library in FreeBSD. From the looks of it, and after consulting the support team, this seems to be an acceptable work around for the time being. So, we're agreeing to deliver the workaround, but keeping the issue alive for the OS team to investigate.

PR #8124, opened 2022-01-11
Fix KeyError for edge case where backup of domain secrets is incomplete

anodos325:
Setting stored_ts to None here will force a re-sync of secrets entry
to our db.


anodos325:
Will make separate 13.0 PR. 12.0-stable PR may or may not be needed (depending on what happens with train).

PR #8131, opened 2022-01-12
`ssh-rsa` was excluded from default `PubkeyAcceptedAlgorithms`. Isn't…

themylogin:
… it a little bit premature?

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/ssh/sshd_config
+++ b/src/middlewared/middlewared/etc_files/local/ssh/sshd_config
@@ -84,6 +84,7 @@ PasswordAuthentication yes
 GSSAPIAuthentication yes
 % endif
 PubkeyAuthentication yes
+PubkeyAcceptedAlgorithms +ssh-rsa
 ${ssh_config['options']}
 % if twofactor_enabled:
 # These are forced to be enabled with 2FA


anodos325:
Out of curiosity are we seeing issues among users because of this?

themylogin:
This issue only appeared in 13.0 nightlies due to the openssh server upgrade so we don't have any users yet. `ssh-rsa` keys are still pretty common.

anodos325:
>ssh-rsa keys are still pretty common.

https://www.openssh.com/txt/release-8.2

```
 * The RFC8332 RSA SHA-2 signature algorithms rsa-sha2-256/512. These
   algorithms have the advantage of using the same key type as
   "ssh-rsa" but use the safe SHA-2 hash algorithms. These have been
   supported since OpenSSH 7.2 and are already used by default if the
   client and server support them.
```
   
Don't the algorithms use the same keys?

anodos325:
If we haven't observed issues yet, then maybe we can let new more secure default stand and play it by ear during alpha / beta?

themylogin:
Sorry, I mixed this up. The only real issue is that TrueNAS 12 won't be able to replicate to/from 13 because 12 has an old version of python SSH client (upgrading port right now)

anodos325:
I don't think it's a bad idea to present user some way of allowing ssh-rsa (even if it's through an auxiliary parameter). We may also need docs to make a note for people replicating EOL Truenas to TrueNAS 13.

themylogin:
https://jira.ixsystems.com/browse/NAS-114283 ticket created

PR #8136, opened 2022-01-13
NAS-114306 / 22.02 / Prevent users from sharing ix-applications via SMB and NFS

anodos325:


anodos325:
Per side-channel discussions, the decision was made to mount the ix-applications dataset outside of /mnt.

Ornias1993:
> Per side-channel discussions, the decision was made to mount the ix-applications dataset outside of /mnt.

Nice solution, the ix-applications dataset is a whole can-of-worms to give the average user access to

PR #8137, opened 2022-01-13
NAS-114674 / 22.02.1 / 22.02 / allow variables for portal path

Ornias1993:
Currently portal path does not accept the use of variables.
This means the path cannot be formed dynamically by using kubernetes resources.

This PR also processes the path through the logic requried to use said variables, just like the other portal options

I'm not 100% sure this synax is correct, please verify!

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
@@ -63,7 +63,7 @@ class ChartReleaseService(Service):
         cleaned_portals = {}
         for portal_type, schema in portals.items():
             t_portals = []
-            path = schema.get('path') or '/'
+            path = tag_func(schema.get('path')) or '/'
             for protocol in filter(bool, map(tag_func, schema['protocols'])):
                 for host in filter(bool, map(tag_func, schema['host'])):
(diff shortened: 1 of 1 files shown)


Ornias1993:
@sonicaj this is your "department" I think...

src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
```diff
@@ -63,7 +63,7 @@ def tag_func(key):
         cleaned_portals = {}
         for portal_type, schema in portals.items():
             t_portals = []
-            path = schema.get('path') or '/'
+            path = tag_func(schema.get('path')) or '/'
```
sonicaj, line 66: It should be `tag_func(schema.get('path')  or '/')`

sonicaj (commented):
Thank you @Ornias1993 for your contribution!

src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
Ornias1993: > It should be `tag_func(schema.get('path') or '/')`

```suggestion
            path = tag_func(schema.get('path') or '/')
```

sonicaj (approved):
Thanks!

PR #8138, opened 2022-01-14
NAS-114327 / 22.02.1 / Give admins more insight into SMB connections to NAS

anodos325:
Modern SMB clients shouldn't be using NTLMv1 for authentication,
but I'm increasingly seeing this on the forums (and users not
being aware that their Windows computer's security settings
have somehow been downgraded). This has also happened for
customers as well. Alerting on NTLMv1 generally will help
possibly avoid a support call / debugging session.

At some point in time in the future, SMB1 will be deprecated
in Samba. We can start generating running logs of SMB1 clients
in notice alerts to nag users to start looking into upgrading
legacy clients.

### DIFF
--- a/src/middlewared/middlewared/alert/source/smb.py
+++ b/src/middlewared/middlewared/alert/source/smb.py
@@ -1,0 +1,70 @@
+import time
+from middlewared.alert.base import AlertClass, AlertCategory, Alert, AlertLevel, AlertSource
+from middlewared.alert.schedule import CrontabSchedule
+
+
+async def generate_alert_text(auth_log):
+    alert_text = [
+        f'{x["Authentication"]["clientAccount"]}: {x["Authentication"]["remoteAddress"]}' for x in auth_log
+    ]
+    return alert_text
+
+
+class SMBLegacyProtocolAlertClass(AlertClass):
+    category = AlertCategory.SHARING
+    level = AlertLevel.NOTICE
+    title = "SMB1 connections to TrueNAS server have been performed in last 24 hours"
+    text = "The following clients have established SMB1 sessions: %(err)s."
+
+
+class NTLMv1AuthenticationAlertClass(AlertClass):
+    category = AlertCategory.SHARING
+    level = AlertLevel.WARNING
+    title = "NTLMv1 authentication has been attempted in the last 24 hours"
+    text = "The following clients have attempted NTLMv1 authentication: %(err)s"
+
+
+class SMBLegacyProtocolAlertSource(AlertSource):
+    schedule = CrontabSchedule(hour=1)  # every 24 hours
+    run_on_backup_node = False
+
+    async def check(self):
+        if not await self.middleware.call('service.started', 'cifs'):
+            return
+
+        now = time.time()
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/alert/source/smb.py
```diff
@@ -0,0 +1,70 @@
...
+    async def check(self):
+        if not await self.middleware.call('service.started', 'cifs'):
+            return
+
+        smb_conf = await self.middleware.call('smb.config')
+        if smb_conf['ntlmv1_auth']:
+            return
+
+        now = time.time()
+        auth_log = await self.middleware.call('smb.status', 'AUTH_LOG', [
```
themylogin, line 61: Will these be unique clientAccount@remoteAddress pairs? Or 1000 attempts by the same client will result in a log text or 1000 pairs? (even if not, we still probably want to show only N recent records from different hosts)

rick-mesta:
@themylogin / @yocalebo: ping 🔔 ... can we get some traction on this PR, please ?

src/middlewared/middlewared/alert/source/smb.py
```diff
@@ -0,0 +1,70 @@
+import time
+from middlewared.alert.base import AlertClass, AlertCategory, Alert, AlertLevel, AlertSource
+from middlewared.alert.schedule import CrontabSchedule
+
+
+async def generate_alert_text(auth_log):
```
themylogin, line 6: This does not need to be `async`

themylogin (changes requested):

src/middlewared/middlewared/alert/source/smb.py
anodos325: Fixed

rick-mesta:
@themylogin : ping 🔔 - has Andrew addressed your concerns here  ?

src/middlewared/middlewared/plugins/smb_/status.py
```diff
@@ -61,7 +62,21 @@ async def status(self, info_level, filters, options, status_options):
...
 
             for e in logfile_entries.splitlines():
-                ret.append(json.loads(e.strip()))
+                # sample timestamp entry:
+                # "2021-12-31T05:42:33.954029-0800"
+                entry = json.loads(e.strip())
+                ts, extra = entry['timestamp'].split('.', 1)
+                # add timezone info
+                ts += extra[6:]
+                usec = extra[:6]
```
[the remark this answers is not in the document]
anodos325:
```
ts = talloc_asprintf(frame, "%s.%06ld %s", buffer, tv.tv_usec, tz);
```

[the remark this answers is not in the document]
anodos325: It appears to not be the case:
```
    "timestamp": "2022-02-17T00:41:54.998990-0800",
    "type": "Authentication",
```
But my brain has bit-rotted since I wrote this part of the alert, and so I'll review Samba code for how it's generating this timestamp and how I'm converting to timeval.

themylogin (approved):

PR #8011, opened 2021-12-16
NAS-113861 / 23.10 / SQLAlchemy 1.4 compatibility

themylogin:
These changes are incompatible with SQLAlchemy 1.3. TrueNAS SCALE uses SQLAlchemy 1.3, TrueNAS CORE 13 uses SQLAlchemy 1.4. This ticket should be merged once we update TrueNAS SCALE to the distribution that ships SQLAlchemy 1.4.

PR #8139, opened 2022-01-15
NAS-114333 / 13.0 / Expand `__[POSTCOMPILE_encrypted_provider_1` into valid SQL syntax

themylogin:
SQLAlchemy includes a variant on a bound parameter known as `BindParameter.expanding`,
which is a “late evaluated” parameter that is rendered in an intermediary state when
a SQL construct is compiled, which is then further processed at statement execution
time when the actual known values are passed. “Expanding” parameters are used for
`ColumnOperators.in_()` expressions by default so that the SQL string can be safely
cached independently of the actual lists of values being passed to a particular
invocation of `ColumnOperators.in_()`.

We need to render the `IN` clause with real bound parameter symbols as we store these
statements in HA database journal and execute them in a lower-level connection that
does not support these SQLAlchemy substitutions.

### DIFF
--- a/src/middlewared/middlewared/plugins/datastore/connection.py
+++ b/src/middlewared/middlewared/plugins/datastore/connection.py
@@ -54,7 +54,7 @@ class DatastoreService(Service):
         options.setdefault('ha_sync', True)
         options.setdefault('return_last_insert_rowid', False)
 
-        compiled = stmt.compile(self.engine)
+        compiled = stmt.compile(self.engine, compile_kwargs={"render_postcompile": True})
 
         sql = compiled.string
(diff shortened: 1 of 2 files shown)


themylogin:
This commit was cherry-picked to https://github.com/truenas/middleware/pull/8011 and will be merged when the time comes.

themylogin:
time 0:30

themylogin:
time 0:30

yocalebo (approved):
I can confirm that testing this on real hardware (where I saw the problem) it fixes the problem.

rick-mesta (approved):

PR #8145, opened 2022-01-18
Update avahi DEV_INFO service model payload

KhaosT:
macOS Big Sur and later dropped support for "RackMac" sidebar icon in Finder. This PR updated the model field to use the new Mac Pro rack instead.

Before:
Screen Shot 2022-01-17 at 10 47 30 PM

After:
Screen Shot 2022-01-17 at 10 46 25 PM


anodos325:
Thanks for the PR. I hadn't noticed the issue with the sidebar. I'll double-check on my server.

anodos325 (approved):

anodos325:
Yup. LGTM. I'll back port to 13-stable.

PR #8149, opened 2022-01-18
NAS-114250 / 13.0 / fix creating zpools on 13

yocalebo:
I _really_ tried to keep the changes here as minimal as possible. These changes are also part of the massive PR against 13 that I filed awhile ago but ended up closing in favor of breaking up the commits into smaller chunks.

Anyways, I've broken all of these commits up into small chunks so they can be followed easily enough without getting lost in all the changes.

Without these changes, on hardware that QE gave me access to, creating a zpool via the webUI on 13 never completes and the OS goes into a semi-catatonic state with `g_event` thread spinning a cpu core at 100% indefinitely. This occurs because `disk.format` calls `disk.gptid_from_part_type` for _EVERY_ disk that gets formatted. This ends up calling `kern.geom.confxml` which is expensive and painful. We do this just to get the `gptid` label that was put on the disk. To make matters worse, we did this in parallel up to a max of 16 times. On a system with 100's of hard drives, you can see how this will never scale or work properly.

My changes fix this design problem by simply formatting the disks in their entirety. Once the formatting has been done, I only call `kern.geom.confxml` once afterwards.

The unfortunate reality is that I had to touch a bunch of surrounding methods because they were all designed for this specific flow control. Also, we still support updating GELI encrypted zpools. I tested the GELI related methods and found quite a few other bugs so I went ahead and fixed those. The biggest one being that on `pool.export` `geli detach` and then `geli clear` raced with `disk.format` if you exported the zpool and chose the option to destroy the zpool. We don't need to run `disk.format` on disks that were geli encrypted because we run `geli clear` which wipes the metadata from the disks. `disk.format` does nothing in this scenario. I also found that when attaching/detaching/decrypting geli providers, we weren't passing the correct arguments to the CLI so I've normalized the device names in their respective methods.

Lastly, I separated the `disk.encrypt` from `pool_/format_disks.py` since it overly complicated `def format_disks` unnecessarily.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/attach_disk.py
+++ b/src/middlewared/middlewared/plugins/pool_/attach_disk.py
@@ -25,7 +25,7 @@ class PoolService(Service):
         is the STRIPED disk GUID which will be converted to mirror. If `target_vdev` is mirror, it will be converted
         into a n-way mirror.
         """
-        pool = await self.get_instance(oid)
+        pool = await self.middleware.call('pool.get_instance', oid)
         verrors = ValidationErrors()
         if not pool['is_decrypted']:
             verrors.add('oid', 'Pool must be unlocked for this action.')
@@ -63,26 +63,20 @@ class PoolService(Service):
         verrors.check()
 
         disks = {options['new_disk']: {'create_swap': topology_type == 'data'}}
-        await self.middleware.call(
-            'pool.format_disks', job, disks,
-            {'enc_keypath': pool['encryptkey_path'], 'passphrase': options.get('passphrase')}
-        )
-
-        # just formatted and encrypted the disk so have to invalidate
-        # the in-memory cache of disk info
+        await self.middleware.call('pool.format_disks', job, disks)
         await self.middleware.call('geom.cache.invalidate')
 
-        devname = await self.middleware.call(
-            'disk.gptid_from_part_type',
-            options['new_disk'],
-            await self.middleware.call('disk.get_zfs_part_type')
-        )
+        zfs_part = await self.middleware.call('disk.get_zfs_part_type')
+        new_devname = await self.middleware.call('disk.gptid_from_part_type', options['new_disk'], zfs_part)
         if pool['encrypt'] > 0:
-            devname = f'/dev/{devname}.eli'
+            new_devname = f'{new_devname}.eli'
+            enc_disks = [{'devname': new_devname}]
+            enc_options = {'enc_keypath': pool['encryptkey_path'], 'passphrase': options.get('passphrase')}
+            await self.middleware.call('pool.encrypt_disks', job, enc_disks, enc_options)
 
         guid = vdev['guid'] if vdev['type'] == 'DISK' else vdev['children'][0]['guid']
         extend_job = await self.middleware.call('zfs.pool.extend', pool['name'], None, [
-            {'target': guid, 'type': 'DISK', 'path': devname}
+            {'target': guid, 'type': 'DISK', 'path': f'/dev/{new_devname}'}
         ])
         try:
             await job.wrap(extend_job)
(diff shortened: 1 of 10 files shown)


rick-mesta:
@anodos325 / @sonicaj / @themylogin : I'm currently looking through these changes, but they are massive -- I'd appreciate a sanity check

yocalebo:
> @anodos325 / @sonicaj / @themylogin : I'm currently looking through these changes, but they are massive -- I'd appreciate a sanity check

The changes should be minimal if each commit is reviewed separately 😉

rick-mesta (approved):
LGTM - But, I'd still like to get a warm fuzzy 🤗 if we could get other members of the Middleware team to take a gander.

themylogin (approved):

src/middlewared/middlewared/plugins/pool_/attach_disk.py
```diff
@@ -44,52 +44,52 @@ async def attach(self, job, oid, options):
...
 
-        guid = vdev['guid'] if vdev['type'] == 'DISK' else vdev['children'][0]['guid']
-        disks = {options['new_disk']: {'create_swap': topology_type == 'data', 'vdev': []}}
-        enc_disks = await self.middleware.call(
-            'pool.format_disks', job, disks,
-            {'enc_keypath': pool['encryptkey_path'], 'passphrase': options.get('passphrase')}
-        )
+        disks = {options['new_disk']: {'create_swap': topology_type == 'data'}}
+        await self.middleware.call('pool.format_disks', job, disks)
+        await self.middleware.call('geom.cache.invalidate')
```
sonicaj, line 67: Do we perhaps want to invalidate cache at the end of `pool.format_disks` instead of making sure we do it each time after calling it ?

themylogin:
@yocalebo I replaced your last commit (`call`/`call_sync` mixup)

src/middlewared/middlewared/plugins/pool_/encrypt_disks.py
```diff
@@ -0,0 +1,61 @@
...
+
+        with ThreadPoolExecutor(max_workers=16) as exc:
+            # The called methods `subprocess` out many times for each disk
+            # so this is painful on systems with large amount of disks.
+            # This is, unfortunately, the best we can really do right now
+            # without implementing gpart code natively in python.
+            # (cython or ctypes wrapper maybe???)
+            futures = [
+                exc.submit(
+                    self.middleware.call_sync(
```
sonicaj, line 38: We should not be submitting to the executor this way - we should be passing it a callable not calling the callable and then passing it
```
def dummy():
    return None
with ThreadPoolExecutor(max_workers=1) as exc:
    futures = [exc.submit(dummy) for i in range(5)]
    results = []
    for idx, fut in enumerate(as_completed(futures), start=1):
        results.append(fut.result())
print(results)

```

src/middlewared/middlewared/plugins/pool_/encrypt_disks.py
```diff
@@ -0,0 +1,61 @@
...
+                        options['enc_keypath'], options.get('passphrase_path')
+                    )
+                )
+                for i in disks
+            ]
+            try:
+                for idx, fut in enumerate(as_completed(futures), start=1):
+                    try:
+                        fut.result()
+                    except TypeError:
```
sonicaj, line 49: This is not required once we do the above

src/middlewared/middlewared/plugins/pool_/format_disks.py
```diff
@@ -1,75 +1,35 @@
...
+        self.middleware.call_sync('disk.sed_unlock_all')  # unlock any SED drives
+        swapgb = self.middleware.call_sync('system.advanced.config')['swapondrive']
+        total_disks = len(disks)
+        with ThreadPoolExecutor(max_workers=16) as exc:
+            # The called methods `subprocess` out many times for each disk
+            # so this is painful on systems with large amount of disks.
+            # This is, unfortunately, the best we can really do right now
+            # without implementing gpart code natively in python.
+            # (cython or ctypes wrapper maybe???)
+            futures = [exc.submit(
```
sonicaj, line 25: Same as above

src/middlewared/middlewared/plugins/pool_/encrypt_disks.py
sonicaj: Also if i am not wrong, this means that we are currently serializing the calls and not actually doing it in parallel

src/middlewared/middlewared/plugins/pool_/attach_disk.py
yocalebo: great idea. I'll make that change

themylogin (approved):
Looks great now, thank you!

PR #8150, opened 2022-01-18
NAS-114332 / 13.0 / fix enclosure mapping on 13 for Mini XL+

yocalebo:
The `TRUENAS-MINI` line of products doesn't support additional enclosures. However, we still map the on-board enclosure device to make it more user friendly in the webUI. When this mapping occurs, the ses number was being dropped which is needed when sync'ing enclosure disks to zpools. This adds back the `number: 0` key which represents the ses number.


yocalebo:
Used wrong ticket number. Will close and open new PR with correct name.

PR #8155, opened 2022-01-20
NAS-114363 / 22.02 / Respect chart dev specified requirement for cert/ca variables

Qubad786:
At the moment, it is not possible for a user to always specify a certificate. This patch makes improvements to allow chart developers to make specifying a certificate/ca as a requirement.


sonicaj (approved):

Ornias1993:
Cool, love this <3

PR #8161, opened 2022-01-20
NAS-114163 / 22.02 / ensure unique mac addresses for bonds

yocalebo:
Different physical systems using the same hardware have shown that linux kernel's mechanism for generating a mac address for the `bond` interface is deterministic and causes collisions. This means systemA with bond0 will have the same mac address as systemB with bond0.

To fix this, I generate a LAA (locally administered address) which guarantees that we do not collide with the UAA (universally administered address). (i.e. we don't want to generate a mac address that has the same vendor bits as Intel/Microsoft/Juniper etc etc)

### DIFF
--- a/src/middlewared/middlewared/plugins/interface/lag.py
+++ b/src/middlewared/middlewared/plugins/interface/lag.py
@@ -89,6 +89,13 @@ class InterfaceService(Service):
             # be sure and bring the lagg back up after making changes
             iface.up()
 
+        # set mac address if needed
+        iface_link_addr = iface.link_address.address.address
+        db_link_addr = lagg['lagg_interface']['int_link_address']
+        if iface_link_addr and iface_link_addr != db_link_addr:
+            self.logger.info('Setting link address for %r to %s', name, db_link_addr)
+            iface.link_address = db_link_addr
+
         members_database = []
         members_configured = {p[0] for p in iface.ports}
         for member in members:
--- a/src/middlewared/middlewared/plugins/interface/netif_linux/interface.py
+++ b/src/middlewared/middlewared/plugins/interface/netif_linux/interface.py
@@ -90,6 +90,10 @@ class Interface(AddressMixin, BridgeMixin, LaggMixin, VlanMixin, VrrpMixin):
         except IndexError:
             return None
 
+    @link_address.setter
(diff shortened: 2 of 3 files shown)


src/middlewared/middlewared/plugins/interface/lag.py
```diff
@@ -89,6 +89,13 @@ def lag_setup(self, lagg, members, disable_capabilities, parent_interfaces, sync
             # be sure and bring the lagg back up after making changes
             iface.up()
 
+        # set mac address if needed
+        iface_link_addr = iface.link_address.address.address
```
themylogin, line 93: `iface.link_address` can be `None`, or are we sure that this will never happen?

src/middlewared/middlewared/plugins/interface/lag.py
```diff
@@ -89,6 +89,13 @@ def lag_setup(self, lagg, members, disable_capabilities, parent_interfaces, sync
...
+        db_link_addr = lagg['lagg_interface']['int_link_address']
+        if iface_link_addr and iface_link_addr != db_link_addr:
+            self.logger.info('Setting link address for %r to %s', name, db_link_addr)
+            iface.link_address = db_link_addr
```
themylogin, line 97: Can't `db_link_addr` be `None` too?

sonicaj:
@yocalebo can you please confirm if both systems have different `/etc/machine-id` file contents ?

sonicaj:
We had an issue similar to like this one before - https://github.com/truenas/scale-build/pull/154. Just got suspicious seeing old version of scale in affected versions in the ticket

yocalebo:
> We had an issue similar to like this one before - [truenas/scale-build#154](https://github.com/truenas/scale-build/pull/154). Just got suspicious seeing old version of scale in affected versions in the ticket

Sure I can check but I'm not sure how that applies to mac addresses? Are you saying the kernel generates it's mac address for `bond` interfaces using the `/etc/machine-id` file?

sonicaj:
@yocalebo yes, that is accurate :)

yocalebo:
> @yocalebo yes, that is accurate :)

Well that's great news. I don't need to add this hack then 😄

PR #8164, opened 2022-01-21
NAS-114012 / 22.02 / fix dns.sync when only DHCP is running

yocalebo:
If we don't have DNS servers configured explicitly by user, then check if DHCP has handed us one when calling `dns.sync`.

### DIFF
--- a/src/middlewared/middlewared/plugins/network_/dns.py
+++ b/src/middlewared/middlewared/plugins/network_/dns.py
@@ -1,8 +1,10 @@
 import contextlib
+import re
 
 from middlewared.service import Service, filterable, filterable_returns, private
 from middlewared.schema import Dict, IPAddr, ValidationErrors
 from middlewared.utils import filter_list
+from middlewared.plugins.interface.netif import netif
 
 
 class DNSService(Service):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/network_/dns.py
```diff
@@ -53,11 +55,41 @@ def sync(self):
...
+                ignore.extend(self.middleware.call_sync('failover.internal_interfaces'))
+                ignore = tuple(ignore)
+                interfaces = list(filter(lambda x: not x.startswith(ignore), netif.list_interfaces().keys()))
+
+            for iface in interfaces:
+                dhclient_running, dhclient_pid = self.middleware.call_sync('interface.dhclient_status', iface)
+                if dhclient_running:
+                    leases = self.middleware.call_sync('interface.dhclient_leases', iface)
+                    if reg := re.search(r'option domain-name-servers (.+)', leases or ''):
+                        # just get the first dns address we come across
```
themylogin, line 93: Why only the first?

yocalebo: Because the `leases` variable can contain duplicate entries and I don't want to pollute `resolv.conf`. I'll push a commit to fix that particular scenario in a moment.

themylogin (approved):

PR #8166, opened 2022-01-21
Update _device-info._tcp. model payload (#8145)

anodos325:
macOS Big Sur and later stopped support for "RackMac" side bar icon in Finder. Update this to use the new Mac Pro rack style.


anodos325:
This can be merged after BETA1 is released.

rick-mesta (approved):

PR #8167, opened 2022-01-21
NAS-113793 / 22.02 / remove int_pass column since it's not used

yocalebo:
On SCALE HA, we've moved to using VRRP. We no longer use the `int_pass` column so drop it.


yocalebo:
ping @truenas/middleware 😄

anodos325 (approved):

PR #8169, opened 2022-01-22
NAS-114431 / 22.02 / VM performance improvements

themylogin:


themylogin:
time 1:30

sonicaj (approved):

src/middlewared/middlewared/plugins/vm/events.py
```diff
@@ -24,8 +24,14 @@ def callback(conn, dom, event, detail, opaque):
             7: 'PMSUSPENDED'
             Above is event mapping for internal reference
             """
-            vms = {f'{d["id"]}_{d["name"]}': d for d in self.middleware.call_sync('vm.query')}
-            if dom.name() not in vms:
+            vm_id = dom.name().split('_')[0]
+            vm = None
+            if vm_id.isdigit():
+                if vms := self.middleware.call_sync('vm.query', [['id', '=', int(vm_id)]], {'force_sql_filters': True}):
```
sonicaj, line 30: Actually probably better to use `get_instance` instead as it by default sets that to true

themylogin: `get_instance` raises exception if the instance is not found. We'll probably need a method that returns `None`, but that's not the idea here.

sonicaj: True that

PR #8170, opened 2022-01-22
NAS-114435 / 22.02.1 / Fix `IoThreadPool`

themylogin:
This is to be merged for 22.MM, 22.02 performance fix will be proposed in a separate PR.

* By default `ThreadPoolExecutor._idle_semaphore` is set to 0, so `no_idle_threads` was
  always `True` and `IoThreadPoolExecutor` was never employed. Fix this in new `ThreadPool`
* Remove `ws_threadpool`, this is an artifact from the past. We already have a semaphore
  to prevent UI clients from calling too many methods simultaneously.
* Perform general cleanup.


themylogin:
time 4:00

yocalebo:
Really liking the general clean up but I'm not sure I understand how `IoThreadPoolExecutor` was never employed. On a nightly, running middlewared in TRACE mode, I can see many methods using `IoThreadPoolExecutor`
```
middlewared._call():1291 - Calling 'failover.licensed' in executor <middlewared.utils.io_thread_pool_executor.IoThreadPoolExecutor object at 0x7fbaeeafee20>
[2022/01/23 08:49:11] (TRACE) middlewared._call():1291 - Calling 'pool.pool_extend_context' in executor <middlewared.utils.io_thread_pool_executor.IoThreadPoolExecutor object at 0x7fbaeeafee20>
[2022/01/23 08:49:11] (TRACE) middlewared._call():1291 - Calling 'pool.pool_extend' in executor <middlewared.utils.io_thread_pool_executor.IoThreadPoolExecutor object at 0x7fbaeeafee20>
```

themylogin:
@yocalebo there are `Calling %r in single-use thread` logs at the very beginning of middleware startup. Then someone calls `self.middleware.run_in_thread`, which employs the executor directly, increases its semaphore value and then things start working like they should. When we were working on catalog fixes, `IoThreadPoolExecutor` sometimes was kept in that incorrect state for the entire catalog sync process.

yocalebo:
> @yocalebo there are `Calling %r in single-use thread` logs at the very beginning of middleware startup. Then someone calls `self.middleware.run_in_thread`, which employs the executor directly, increases its semaphore value and then things start working like they should. When we were working on catalog fixes, `IoThreadPoolExecutor` sometimes was kept in that incorrect state for the entire catalog sync process.

Interesting. That makes more sense.

yocalebo (approved):
These are great changes! I approve these changes, but I'd like to make 2 comments.

1. `start_daemon_thread` takes a kwarg of `daemon=True` which implies that it could be called with `daemon=False` which would make it a non-daemon thread but the name of the method implies that it's always a daemon thread. It'd be nice to probably either remove `daemon` kwarg or just ignore it inside the function.

2. This changes default behavior by spinning up single-use threads as daemon threads. I know the pros/cons either way but just mentioning it here for posterity sake.

yocalebo:
Oh 1 last little nit-pick. In the PR body you mention `ThreadPool` but you named it `ThreadExecutor` in the code. Personally, I like the `ThreadPool` name much better 😄

themylogin:
@yocalebo thank you! I'll fix all this once `master` becomes a branch for `22.MM`

yocalebo:
ping @themylogin we've branched so wondering if this can be merged as I'd like to get this in nightlies so it can be tested a little more thoroughly 😄

yocalebo (approved):

PR #8172, opened 2022-01-22
Performance improvements for crypto/vm plugins

sonicaj:

1. Separate certs/cas extend method. Motivation for this change is that the old implementation retrieved certs/cas recursively in the same method for each row in question and for cas it could be worse in cases when complete ca chain was being retrieved again and again for each row. It was doing a stat call creating ca/certs directory in extends method which was already being numerous times in a cert/ca query call which has also been fixed.
2. Do not make unnecessary `certificate.query` calls as they can be a relatively expensive operation if you have many certs and system is already under load.
3. Do not perform unnecessary i/o again and again each time a VM operation is performed and instead cache those usages.
4. Improve retrieving vm status and do not switch between asynchronous/synchronous methods for each row


src/middlewared/middlewared/plugins/vm/vms.py
```diff
@@ -74,13 +82,13 @@ async def bootloader_options(self):
         return BOOT_LOADER_OPTIONS
 
     @private
-    async def extend_vm(self, vm):
+    async def extend_vm(self, vm, context):
```
themylogin, line 85: Do you recall which was the performance impact of this change? If there was one, we have a bigger problem.

src/middlewared/middlewared/plugins/vm/lifecycle.py
```diff
@@ -49,16 +49,15 @@ async def check_setup_libvirt(self):
 
     @private
     def initialize_vms(self, timeout=30):
-        if self.middleware.call_sync('vm.query') and self.middleware.call_sync('vm.supports_virtualization'):
```
themylogin, line 52: You can run `vm.query` once here and then re-use the result.

src/middlewared/middlewared/plugins/vm/connection.py
```diff
@@ -33,7 +34,9 @@ def _is_kvm_supported(self):
         # give the impression that virtualization can be used. We have checks in place to check if system
         # supports virtualization but if we incorporate that check in all of the vm exposed methods which
         # consume libvirt, it would be an expensive call as we figure that out by making a subprocess call
-        return os.path.exists('/dev/kvm')
+        if self.KVM_SUPPORTED is None:
+            self.KVM_SUPPORTED = os.path.exists('/dev/kvm')
```
themylogin, line 38: Are `kvm-ok` exit code and existence of `/dev/kvm` somehow connected?

themylogin (changes requested):
Great optimization job was done yesterday!

However, I am worried about such a big change (even if it was just refactoring) so close to the release. Our current certificate tests are very basic (to say the least), can we consider adding just a few more? (now we have an ability to test websocket API rather than REST API; check out `test_s3.py` for examples)

src/middlewared/middlewared/alert/source/certificates.py
```diff
@@ -82,7 +82,7 @@ async def _get_service_certs(self):
                 'type': _type,
             },
             {
-                'id': (await self.middleware.call('system.general.config'))['ui_certificate']['id'],
+                'id': (await self.middleware.call('datastore.config', 'system.settings'))['stg_guicertificate']['id'],
```
william-gr, line 85: We had an agreement to not use datastore API directly, since it might have change behind its back and its a private API.

sonicaj: True that, i'll make the required changes. Thank you for pointing it out

william-gr (approved):
I didnt review in depth, but looks generally OK

themylogin (approved):

PR #8173, opened 2022-01-23
NAS-114393 / 22.02 / Performance improvements when retrieving catalogs

sonicaj:
After profiling/investigating middleware process, it was found that when reading large catalog like truecharts - we ran into GIL issues where the thread responsible for reading the large catalog was taking too much time of the main process which is also responsible for managing/delegating middleware calls which resulted in system lag as middleware became very slow in processing requests.
After discussing with @themylogin, we decided to read the catalog in a separate process and the results show incredible improvements to the middleware process responsiveness and reading the catalog information as well. Earlier it used to take 4-5 minutes to read the catalog and it takes approx 50 secs now and all the while middleware is fully responsive and no lag is experienced on that end.

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/features.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/features.py
@@ -19,11 +19,11 @@ SUPPORTED_FEATURES = {
 }
 
 
-class CatalogService(Service):
+def version_supported(version_details: dict) -> bool:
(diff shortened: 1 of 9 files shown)


Ornias1993 (approved):
I can confirm we ran into similair issues in our CI as well, split and/or parallel processing is indeed the solution in this case 👍

themylogin (approved):

yocalebo (approved):

PR #8174, opened 2022-01-23
NAS-114405 / 22.02 / Issues with k8s backup / restore

Qubad786:
Following issues were fixed in this PR:
- User specified secrets were not being restored
- Zvolume was not being created
- Redeploy now doesn't include schema validation now
- CRDs were not getting restored

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/redeploy.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/redeploy.py
@@ -1,5 +1,10 @@
+import errno
+import os
+
 from middlewared.schema import accepts, Str, Ref, returns
-from middlewared.service import job, Service
+from middlewared.service import CallError, job, Service
+
+from .utils import add_context_to_configuration, CONTEXT_KEY_NAME, get_action_context
 
 
 class ChartReleaseService(Service):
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/kubernetes_linux/backup.py
```diff
@@ -58,7 +58,8 @@ def backup_chart_releases(self, job, backup_name):
 
             secrets = self.middleware.call_sync(
                 'k8s.secret.query', [
-                    ['type', '=', 'helm.sh/release.v1'], ['metadata.namespace', '=', chart_release['namespace']]
+                    ['type', 'in', ['helm.sh/release.v1', 'Opaque']],
```
Ornias1993, line 61: Good catch there, this explains why the secrets where not being backed up correctly :)

src/middlewared/middlewared/plugins/kubernetes_linux/restore.py
```diff
@@ -221,7 +222,25 @@ def restore_backup(self, job, backup_name, options):
...
+                    crd_path = os.path.join(crds_path, crd_file)
+                    try:
+                        self.middleware.call_sync('k8s.cluster.apply_yaml_file', crd_path)
+                    except Exception as e:
+                        failed_crds.append(f'Unable to create CRD(s) at {crd_path!r}: {e}')
+
+            if failed_crds:
+                self.logger.error(
+                    'Failed to restore CRD(s) for %r chart release:\n%s', chart_release, '\n'.join(failed_crds)
+                )
```
Ornias1993, line 241: Looks like a good fix for the CRD issue, although in the future we might want to standardise something like this in a more DRY fasion for both Update and Restore

Ornias1993 (approved):
Looks pretty nice, ticks precisely all the issue I was having while testing restore. 👍

src/middlewared/middlewared/plugins/chart_releases_linux/redeploy.py
```diff
@@ -0,0 +1,45 @@
...
+    class Config:
+        namespace = 'chart.release'
+
+    @accepts(Str('release_name'))
+    @returns(Ref('chart_release_entry'))
+    @job(lock=lambda args: f'chart_release_redeploy_{args[0]}')
+    async def redeploy(self, job, release_name):
+        """
+        Redeploy will initiate a rollout of new pods according to upgrade strategy defined by the chart release
+        workloads. A good example for redeploying is updating kubernetes pods with an updated container image.
```
Ornias1993, line 21: slight nitpick: redeploy is technically not limited to pods, which might be confusing for people that do not know it calles `helm update` in the end.

```suggestion
        Redeploy will initiate a new rollout of the Helm chart according to upgrade strategy defined by the chart release
        workloads. A good example for redeploying is updating kubernetes pods with an updated container image.
```

Qubad786: Thanks for pointing, addressed.

sonicaj (approved):
LGTM

PR #8176, opened 2022-01-24
NAS-114501 / 22.02 / services/nfs - convert to kernel NFS server

anodos325:


src/middlewared/middlewared/plugins/etc.py
```diff
@@ -120,7 +120,10 @@ class EtcService(Service):
         'nfsd': [
             {'type': 'py', 'path': 'nfsd', 'platform': 'FreeBSD', 'checkpoint': 'pool_import'},
             {'type': 'mako', 'path': 'default/nfs-common', 'platform': 'Linux'},
-            {'type': 'mako', 'path': 'ganesha/ganesha.conf', 'platform': 'Linux', 'checkpoint': 'pool_import'},
+            {'type': 'mako', 'path': 'default/nfs-kernel-server', 'platform': 'Linux'},
```
yocalebo, line 123: nit-pick but can remove `platform` key since this branch only runs on SCALE.

src/middlewared/middlewared/plugins/service_/services/nfs.py
```diff
@@ -9,29 +10,4 @@ class NFSService(SimpleService):
 
     freebsd_rc = "nfsd"
```
yocalebo, line 11: nit can remove `freebsd_rc` variable

yocalebo (commented):
Haven't closely reviewed, just minor nits at first glance. I'll come back around to this once it's closer to getting merged.

rick-mesta (approved):

yocalebo (commented):
Remove ganesha as a dependency in debian control file?

anodos325:
Per recommendation of performance team, I will add migration to increase the number of servers to 16.

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -394,6 +394,30 @@ async def do_delete(self, id):
...
         await self._service_change("nfs", "reload")
 
+    @private
+    async def validate_paths(self, data, schema_name, verrors):
+        """
+        Allowing a list of `paths` for NFS shares is for API consistency between
+        Linux and FreeBSD. In Linux case, each path will be treated as a separate
+        export with identical configuration. Since the `locked` property applies
+        to the share as a whole, we enforce same validation on Linux as on FreeBSD
+        (restricting paths to the same filesystem).
```
themylogin, line 404: This seems fine for 22.02, but having this restriction only because our locked shares API does not fit with this particular feature of NFS seems excessive. Do we have plans/ticket to fix this in the future?

src/middlewared/middlewared/plugins/service_/services/nfs.py
```diff
@@ -7,31 +7,4 @@ class NFSService(SimpleService):
...
-        await self._freebsd_service("rpcbind", "stop", force=True)
-
-    async def _reload_freebsd(self):
-        await self.middleware.call("nfs.setup_v4")
-        await self._freebsd_service("mountd", "reload", force=True)
-
-    async def _stop_linux(self):
-        await self._systemd_unit("nfs-ganesha-lock", "stop")
-        await super()._stop_linux()
+    systemd_unit = "nfs-kernel-server"
```
themylogin, line 10: So we only need to restart `nfs-kernel-server` and the rest will be reconfigured automatically?

themylogin (approved):
LGTM!

src/middlewared/middlewared/plugins/nfs.py
anodos325: Well, I think right thing to do would be to have only one path per export created in GUI, and do a migration to convert from list to string (creating new shares as needed). Unfortunately, this would have the side-effect of changing the API for sharing.nfs.

src/middlewared/middlewared/plugins/service_/services/nfs.py
anodos325: The systemd unit should take care of everything. Once this gets merged, I'll probably die a death by a thousand papercuts as nothing ever works the way I expect.

anodos325:
Two remaining items:
1) fix API tests
2) verify that we close client state where appropriate when encrypting datasets.

src/middlewared/middlewared/plugins/nfs.py
themylogin: Do you want me to create a ticket?

anodos325: @themylogin sure, you can do that. Maybe reasonable to put in bluefin.

themylogin: @anodos325 on the other hand - can we just consider the entire share locked if at least one path is locked? This way only users who use dataset encryption will really be affected.

ericbsd (approved):
LGTM

yocalebo:
retest this please

PR #8177, opened 2022-01-24
NAS-114454 / 22.02 / Add unit/api tests for crypto plugin

sonicaj:


src/middlewared/middlewared/plugins/crypto_/__init__.py
```diff
@@ -15,7 +16,7 @@ async def setup(middleware):
         # create a self signed cert if it doesn't exist and set ui_certificate to it's value
         try:
             if not any('freenas_default' == c['name'] for c in certs):
-                cert, key = await middleware.call('cryptokey.generate_self_signed_certificate')
+                cert, key = generate_self_signed_certificate()
```
yocalebo, line 19: async method calling a sync method that calls other sync methods.

sonicaj: In 99.99% of cases it would only be called once during TrueNAS setup ( not middleware lifecycle ) and take less then a second to complete. Is it problematic ?

themylogin: Formally speaking, I think it performs I/O (consuming random seed) so it might block the loop? What's the need to move this outside plugins anyway?

src/middlewared/middlewared/plugins/crypto_/certificate_authorities.py
```diff
@@ -126,33 +127,16 @@ async def get_serial_for_certificate(self, ca_id):
...
-                    data['serial'] for data in
-                    await self.middleware.call(
-                        'datastore.query',
-                        'system.certificate',
-                        [('signedby', '=', ca_id)],
-                        {
-                            'prefix': self._config.datastore_prefix,
-                            'extend': self._config.datastore_extend
-                        }
+                    data['serial'] for data in await self.middleware.call(
```
themylogin, line 130: Why do we switch to `certificate.query` if we only need serials?

themylogin (approved):
Great job!!! I can't carefully review all your refactoring, but implementing tests certainly adds a lot of confidence that crypto plugin works all correctly now. Did you find any bugs on the way?

src/middlewared/middlewared/plugins/crypto_/certificate_authorities.py
sonicaj: We need to parse the certificate to get the serial and on that note we don't need all the other normalization, i'll quickly make changes to do just that

sonicaj:
I found a few bugs along the way which i fixed during the refactoring - adding tests certainly gives more confidence, although would still be nicer if we add more detailed tests but it should be good for now at least i think :)

themylogin (approved):

PR #8178, opened 2022-01-25
NAS-114306 / 22.02 / Do not allow ix-applications path to be shared

Qubad786:
Make sure that the path does not get validated if its path to k8s dataset or one of its children.

### DIFF
--- a/src/middlewared/middlewared/async_validators.py
+++ b/src/middlewared/middlewared/async_validators.py
@@ -4,7 +4,6 @@ import socket
 from pathlib import Path
 
 from middlewared.validators import IpAddress
-from middlewared.utils import osc
 
 
 async def check_path_resides_within_volume(verrors, middleware, name, path, gluster_bypass=False):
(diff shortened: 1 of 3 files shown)


yocalebo (approved):

anodos325:
I think there was some conversation about whether to approach it this way vs blocking particular services in the first go-around with this ticket. @sonicaj are we okay with this?

yocalebo:
If this dataset doesn't need to be touched by end user, then we should hide it from the webUI as well.

yocalebo (changes requested):
Wait, how expensive is calling `kubernetes.config`? `check_path_resides_within_volume` gets called in numerous places so we need to consider the performance impact.

Qubad786:
@yocalebo it's as expensive as doing a database call as we do not do anything else when retrieving configuration.. Is that a concern?

yocalebo:
> @yocalebo it's as expensive as doing a database call as we do not do anything else when retrieving configuration.. Is that a concern?

If the name `ix-applications` is a reserved dataset name, then we just need to assume that if anyone is trying to do something underneath that namespace, regardless of whether or not apps are being used, needs to be considered wrong.

Ornias1993:
> If the name `ix-applications` is a reserved dataset name, then we just need to assume that if anyone is trying to do something underneath that namespace, regardless of whether or not apps are being used, needs to be considered wrong.

When a pool is selected for Apps, it should already (re)move any existing ix-applications datasets.
The only way to disable the Apps system is by not setting or unsetting the pool for it.

So if a user has:
`POOL1/ix-applications` (selected for Apps)
and
`POOL2/ix-applications` (not selected for apps, for example a clone of POOL1)

and the user moves Apps to POOL2, it nukes/moves the data in `POOL2/ix-applications`

Hence, there should be no issues with user defined `POOL/ix-applications` on no-apps pools conflicting with Apps pools :)

src/middlewared/middlewared/async_validators.py
```diff
@@ -22,27 +21,32 @@ async def check_path_resides_within_volume(verrors, middleware, name, path, glus
...
+        # subtract 2 here to remove the '/' and 'mnt' parents
+        for i in range(0, len(rp.parents) - 2):
+            if rp.parents[i].is_mount() and rp.parents[i].name == ".glusterfs":
+                using_gluster_path = True
+                break
+
+    if using_gluster_path:
+        verrors.add(name, "A path being used by Gluster is not allowed")
+
+    k8s_ds = (await middleware.call('kubernetes.config'))['dataset']
```
Ornias1993, line 45: Could use a little note here as well, to sync with above note about cluster:
```suggestion

    # We must make sure that the internal Apps datasets are not shared, 
    # as this might lead to breaking changes to its, very specific, structure and settings.
    k8s_ds = (await middleware.call('kubernetes.config'))['dataset']
```

Ornias1993 (approved):
One nitpick on documentation... that's it.

yocalebo (approved):
Much better 👍

src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -142,6 +142,9 @@ def listdir(self, path, filters, options):
         if not path.is_dir():
             raise CallError(f'Path {path} is not a directory', errno.ENOTDIR)
 
+        if 'ix-applications' in path.parts:
+            raise CallError('Ix-applications is a system managed dataset and it\'s contents cannot be listed')
```
yocalebo, line 146:
```suggestion
            raise CallError('Ix-applications is a system managed dataset and its contents cannot be listed')
```
`it's` is a contraction for `it is` or 'it has' so remove the apostrophe.

PR #8195, opened 2022-02-01
NAS-114573 / 22.02 / fix process pool deadlock (work-around)

yocalebo:
This causes a process pool deadlock on a system with 20k snapshots because of an inherit design issue with our process pool. After discussion with @themylogin, we've determined to remove the call that is causing the failure condition and will design a proper process pool fix for a later scale release.

### DIFF
--- a/src/middlewared/middlewared/plugins/zfs.py
+++ b/src/middlewared/middlewared/plugins/zfs.py
@@ -1030,12 +1030,6 @@ class ZFSSnapshot(CRUDService):
         select = options.pop('select', None)
         result = filter_list(snapshots, filters, options)
 
-        if not select or 'retention' in select:
-            if isinstance(result, list):
-                result = self.middleware.call_sync('zettarepl.annotate_snapshots', result)
-            elif isinstance(result, dict):
-                result = self.middleware.call_sync('zettarepl.annotate_snapshots', [result])[0]
-
         if select:
             if isinstance(result, list):
                 result = [{k: v for k, v in item.items() if k in select} for item in result]


yocalebo:
retest this please

themylogin:
I was going to propose the same thing, but there are multiple important reasons against this:
* Some plugins have global state which now must be shared between processes
* What if this sync method calls an async method? We don't have an event loop in the worker process
* This all eventually leads us to running 6 full-featured middleware processes (1 original + 5 workers) with corresponding resource consume increase

I think we can just disable annotating snapshots for Angelfish (since UI team did not have the resources to add a column that consumes this data) and discuss fixing this for Bluefin.

yocalebo:
> I was going to propose the same thing, but there are multiple important reasons against this:
> 
> * Some plugins have global state which now must be shared between processes
> * What if this sync method calls an async method? We don't have an event loop in the worker process
> * This all eventually leads us to running 6 full-featured middleware processes (1 original + 5 workers) with corresponding resource consume increase
> 
> I think we can just disable annotating snapshots for Angelfish (since UI team did not have the resources to add a column that consumes this data) and discuss fixing this for Bluefin.

Thank you for your response. I'm glad you describe the same issues that I was seeing. I thought I was going insane. 😄

themylogin:
@rick-mesta is this the time we branch 22.02?

(TLDR: we need to remove some code for 22.02 but keep it in `master`)

rick-mesta:
I do think we need to branch for Bluefin, but we may need to do that right after we release `22.02` and leave `master` for Bluefin development. @william-gr will obviously make the final call.

william-gr:
Lets just keep one PR open for Bluefin for now, we dont have enough changes for Bluefin to grant a new branch at this time.

yocalebo:
@themylogin ping whenever you can. I removed the call to `zettarepl.annotate_snapshots` as a stop-gap for now 😄

themylogin (approved):

PR #8197, opened 2022-02-02
NAS-114581 / 22.02 / fix exorbitant middlewared service memory usage

yocalebo:
The fact that we even have this makes me squirm in my seat. The "debugability" that this provides doesn't outweigh the cons that it introduces in my opinion...but I digress. I've found 4 major problems.

1. because we put all websocket results in this queue, a reference is held which prevents the memory from being reclaimed from the parent middlewared process. The _ONLY_ way memory is "reclaimed" and therefore shrinks in size is when the deque is full and a new entry is added _AND_ the entry in the deque that is being replaced is smaller than the entry being inserted.....so this gets out of hand quite quickly
2. we were storing the non-serialized results so the memory usage was, in theory, considerably larger than what it should be
3. we set the deque size to 1000 entries which is kind of insane, this allows the potential of growing exponentially
4. we're performing a `copy.deepcopy` for every single websocket call just so we can format the `params` key before we store it in the `deque`. We can get rid of this by storing in the `deque` as the last step in the associated method.

This almost immediately starts eating memory on large systems (systems with 100's of hard drives) because every 5 minutes a websocket call is made to `disk.temperatures`. On a system with lots of hard drives, the result is quite large. While a single entry isn't that big a deal, 1000 of them is.

Furthermore, this is used by the webUI team and we capture this in the debug by running `core.get_websocket_messages` HOWEVER when you call that message via websocket it returns the entire contents of the `deque` and then turns around and stores those results in the `deque`.....

To fix, I've done somethings:

1. shrunk the deque to 50....no reason to have 1000
2. if the serialized string of the result is greater than 1MB in size, then instead of storing the result we overwrite the results with a string letting the end user know.
3. get rid of a `copy.deepcopy` by changing the order in which we append to the `deque` since we were only doing it so we don't mess up the structure of the original argument passed to us

With my fixes, this resolves massive memory usage on a system with 20k snapshots and calling `zfs.snapshot.query` multiple times. (Calling it 5 times grew the parent process to ~1.4GB of resident memory).

### DIFF
--- a/src/middlewared/middlewared/main.py
+++ b/src/middlewared/middlewared/main.py
@@ -106,12 +106,27 @@ class Application(object):
         self.__callbacks[name].append(method)
 
     def _send(self, data):
+        serialized = json.dumps(data)
+        if sys.getsizeof(serialized) > 1000000:
+            # no reason to store data in the deque that
+            # is larger than 1MB after being serialized.
+            # This gets _really_ painful on systems with
+            # many (100's) of snapshots because running a
+            # simple `zfs.snapshot.query` via midclt from
+            # the cli produces ridiculously large output.
+            # Caching that in the main middleware process
+            # is excessive and only hurts us. Instead we'll
+            # just store a string indicating it was excluded.
+            message = 'RESULT EXCLUDED BECAUSE THE SIZE IS > 1MB'
+        else:
+            message = serialized
+
         self.middleware.socket_messages_queue.append({
             'type': 'outgoing',
             'session_id': self.session_id,
-            'message': data,
+            'message': message,
         })
-        asyncio.run_coroutine_threadsafe(self.response.send_str(json.dumps(data)), loop=self.loop)
+        asyncio.run_coroutine_threadsafe(self.response.send_str(serialized), loop=self.loop)
 
     def _tb_error(self, exc_info):
         klass, exc, trace = exc_info
@@ -890,7 +905,7 @@ class Middleware(LoadPluginsMixin, RunInThreadMixin, ServiceCallMixin):
         self.__terminate_task = None
         self.jobs = JobsQueue(self)
         self.mocks = {}
-        self.socket_messages_queue = deque(maxlen=1000)
+        self.socket_messages_queue = deque(maxlen=50)
 
     def __init_services(self):
         from middlewared.service import CoreService
--- a/src/middlewared/middlewared/main.py
+++ b/src/middlewared/middlewared/main.py
@@ -107,7 +107,8 @@ class Application(object):
     def _send(self, data):
         serialized = json.dumps(data)
         asyncio.run_coroutine_threadsafe(self.response.send_str(serialized), loop=self.loop)
-        if sys.getsizeof(serialized) > 1000000:
+        _1MB = 1000000
+        if sys.getsizeof(serialized) > _1MB:
             # no reason to store data in the deque that
             # is larger than 1MB after being serialized.
             # This gets _really_ painful on systems with
@@ -116,8 +117,8 @@ class Application(object):
             # the cli produces ridiculously large output.
             # Caching that in the main middleware process
             # is excessive and only hurts us. Instead we'll
-            # just store a string indicating it was excluded.
-            message = 'RESULT EXCLUDED BECAUSE THE SIZE IS > 1MB'
+            # truncate to 1MB.
+            message = serialized[:_1MB]
         else:
             message = serialized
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/main.py
```diff
@@ -106,12 +105,27 @@ def register_callback(self, name, method):
...
+            # no reason to store data in the deque that
+            # is larger than 1MB after being serialized.
+            # This gets _really_ painful on systems with
+            # many (100's) of snapshots because running a
+            # simple `zfs.snapshot.query` via midclt from
+            # the cli produces ridiculously large output.
+            # Caching that in the main middleware process
+            # is excessive and only hurts us. Instead we'll
+            # just store a string indicating it was excluded.
+            message = 'RESULT EXCLUDED BECAUSE THE SIZE IS > 1MB'
```
themylogin, line 120: Can we store just first megabyte of the message and cut the rest in this case? `message = message[:1000000] + '...'`

yocalebo:
> > (systems with 100's of hard drives) because every 5 seconds a websocket call is made to disk.temperatures.
> 
> It is intended to run every 5 minutes, or did you mean 5 minutes? 😀

Oops, you're right, I've mistaken `reporting.cpu_temperatures` (that's occurring every 10 seconds) in place of `disk.termperatures` which is running every 5 mins. I'll update my verbiage

themylogin (approved):
Looks good to me, but let's have an extra pair of careful eyes

src/middlewared/middlewared/main.py
yocalebo: Great idea, I'll fix.

src/middlewared/middlewared/main.py
```diff
@@ -275,19 +290,6 @@ async def on_close(self, *args, **kwargs):
         self.middleware.unregister_wsclient(self)
 
     async def on_message(self, message):
-        if message.get('msg') == 'method' and message.get('method') and isinstance(message.get('params'), list):
```
sonicaj, line 278: I am torn over this, if UI sends a malformed response for some reason and we send an error ( which we log in the dequeue ) but we have not logged the incoming call to which it was replied, we potentially lose a debugging case ? Also in practice how many failed calls do really happen or in other words what improved metrics do we see wrt less resource consumption if we do not log the failure scenarios ?

sonicaj (commented):
LGTM generally, one minor nit.

Also @william-gr / @undsoft do you think length of 50 is reasonable for logged websocket incoming/outcoming responses combined ?

src/middlewared/middlewared/main.py
```diff
@@ -106,12 +105,28 @@ def register_callback(self, name, method):
         self.__callbacks[name].append(method)
 
     def _send(self, data):
+        serialized = json.dumps(data)
+        asyncio.run_coroutine_threadsafe(self.response.send_str(serialized), loop=self.loop)
+        _1MB = 1000000
+        if sys.getsizeof(serialized) > _1MB:
```
william-gr, line 111: why sys.getsizeof? Its a string, can use len()?

src/middlewared/middlewared/main.py
```diff
@@ -890,7 +880,7 @@ def __init__(
         self.__terminate_task = None
         self.jobs = JobsQueue(self)
         self.mocks = {}
-        self.socket_messages_queue = deque(maxlen=1000)
+        self.socket_messages_queue = deque(maxlen=50)
```
william-gr, line 883: How about 200? xD

src/middlewared/middlewared/main.py
william-gr: Also 1MB seems excessive, I would limit it to 10KiB or something? what else would use such a big message?

yocalebo:
```
>>> zz = 'qweoirupqweriopkjlasdf;klasdklfjpouiqweriuoqwiueorask;jldf;lkjasdfasdklff'
>>> len(zz)
73
>>> sys.getsizeof(zz)
122
```

src/middlewared/middlewared/main.py
yocalebo: I don't really see a reason to log the error information in the `deque` if that's what you're implying 😄

src/middlewared/middlewared/main.py
william-gr: Since its a string is just a few extra bytes for the object, negligible

src/middlewared/middlewared/main.py
undsoft: I've opened a random machine, started at Storage and went directly to Advanced -> Save Debug. There were 34 messages since clicking on Advanced menu item. Some of them were job related, but this is likely to happen on user's machine as well.
Since it's not unlikely for user to do something else before going to report an issue, I feel like 50 is going to be unusable and 200 is going to be barely usable.

william-gr: We are starting with 200. We can re-evaluate if you notice this is not enough after the release...

PR #8198, opened 2022-02-02
NAS-114541 / 22.02 / Fix a few edge cases with exports generation

anodos325:
* enclose path with quotes
* default to "sys" if nfs.sec returns an empty list


rick-mesta (approved):
🚢 it

PR #8204, opened 2022-02-04
NAS-114594 / 22.02.1 / Validate CN / SANs and tls_server_uri

Qubad786:
1. Update the migration and fallback to `localhost` if its `SANs` / `CN` does not contain anything except for a wildcard domain.
2. Add validation into the `s3.config` API so that wild card domain cannot be updated into `tls_server_uri`

### DIFF
--- a/src/middlewared/middlewared/alembic/versions/12.0/2021-12-22_12-59_add_server_tls_uri_to_s3_config.py
+++ b/src/middlewared/middlewared/alembic/versions/12.0/2021-12-22_12-59_add_server_tls_uri_to_s3_config.py
@@ -7,11 +7,27 @@ from alembic import op
 import sqlalchemy as sa
 from OpenSSL import crypto
 
+from middlewared.validators import Hostname
+
 # revision identifiers, used by Alembic.
 revision = '9c11f6c6f152'
 down_revision = 'fee786dfe121'
 branch_labels = None
 depends_on = None
+
+
+def is_valid_hostname(hostname: str):
+    """
+    Validates hostname and makes sure it
+    does not contain a valid card.
+    """
+    validate_hostname = Hostname()
+    try:
+        validate_hostname(hostname)
+    except ValueError:
+        return False
+    else:
+        return "*" not in hostname
 
 
 def upgrade():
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/alembic/versions/12.0/2021-12-22_12-59_add_server_tls_uri_to_s3_config.py
```diff
@@ -7,13 +7,29 @@
...
+    Validates hostname and makes sure it
+    does not contain a valid card.
+    """
+    validate_hostname = Hostname()
+    try:
+        validate_hostname(hostname)
+    except ValueError:
+        return False
+    else:
+        return "*" not in hostname
```
sonicaj, line 30: We should update the `Hostname` validator so that it correctly validates an actual hostname and have this there instead

src/middlewared/middlewared/alembic/versions/12.0/2021-12-22_12-59_add_server_tls_uri_to_s3_config.py
```diff
@@ -7,13 +7,29 @@
 import sqlalchemy as sa
 from OpenSSL import crypto
 
+from middlewared.validators import Hostname
```
sonicaj, line 10: We should avoid importing from middleware in migrations, `Hostname` should be regex iirc, we can just use it directly

Qubad786: May I know why should we not import a util from `middlewared` package? 🤔

sonicaj: http://docs.middleware.ixsystems.net/database/migrations.html#migrations-tips

src/middlewared/middlewared/alembic/versions/12.0/2021-12-22_12-59_add_server_tls_uri_to_s3_config.py
```diff
@@ -13,6 +15,17 @@
...
 depends_on = None
 
+# Pattern is taken from middlewared.validators.Hostname
+hostname_re = re.compile(r'^[a-z\.\-0-9]*[a-z0-9]$', flags=re.IGNORECASE)
+
+
+def is_valid_hostname(hostname: str):
+    """
+    Validates hostname and makes sure it
+    does not contain a valid card.
```
sonicaj, line 25: `wild` ;) ?

src/middlewared/middlewared/plugins/s3.py
```diff
@@ -41,7 +41,7 @@ class Config:
         Str('access_key', max_length=20, required=True),
         Str('secret_key', max_length=40, required=True),
         Bool('browser', required=True),
-        Str('tls_server_uri', null=True, required=True),
+        Str('tls_server_uri', validators=[Hostname()], null=True, required=True),
```
sonicaj, line 44: Does `Hostname` validator error out when `null` is provided ?

sonicaj (commented):
Just one more minor nit

sonicaj (approved):
Thanks!

PR #8207, opened 2022-02-04
NAS-114632 / 22.02 / Fix websocket regression

sonicaj:
This commit fixes an issue where we change message parameters which results in auth.login and other methods being called with the dumped version of args resulting in middleware client failing to authenticate/process other calls successfully.

### DIFF
--- a/src/middlewared/middlewared/main.py
+++ b/src/middlewared/middlewared/main.py
@@ -348,12 +348,16 @@ class Application(object):
             await self.unsubscribe(message['id'])
 
         if message.get('msg') == 'method' and message.get('method') and isinstance(message.get('params'), list):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/main.py
```diff
@@ -348,12 +349,17 @@ async def on_message(self, message):
             await self.unsubscribe(message['id'])
 
         if message.get('msg') == 'method' and message.get('method') and isinstance(message.get('params'), list):
-            message['params'] = self.middleware.dump_args(message.get('params', []), method_name=message['method'])
+            log_message = copy.deepcopy(message)
```
themylogin, line 352: Why `deepcopy` and not just `log_message = dict(message, params=...)`?

themylogin (approved):

PR #8209, opened 2022-02-04
NAS-114378 / 22.02 / Make sure locked datasets have immutable flag set on their mountpoint

sonicaj:
There were 2 cases where locked dataset(s) mountpoint did not have immutable flag set which posed a risk that a user/app or any other consumer might manipulate their mountpoint.

1. If a pool had unlocked datasets but the system was rebooted, the system would not in that case have set immutable flag on locked datasets and these mountpoints could be exploited.
2. If a pool is imported, locked dataset(s) mountpoint did not have immutable flag set and hence they were also exposed to the same problem.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -3608,6 +3608,11 @@ class PoolDatasetService(CRUDService):
                 if attachments:
                     await delegate.delete(attachments)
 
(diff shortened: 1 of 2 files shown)


themylogin (changes requested):
Can we add an integration test at least for dataset unlocking method? (I can't find tests for pool import; seems we need to write that one for 22.MM)

themylogin (approved):
Thank you!

PR #8213, opened 2022-02-04
NAS-114519 / 22.02 / Configure timeout on kubernetes_asyncio client instance

Qubad786:
Kubernetes service went unresponsive and `middleware` was stalled forever to hear back from `kubernetes` service since mw is making API calls to `kubernetes` service (its unknown so far why service went unresponsive for this instance for the reporter). However, `middleware` should timeout at some point if `kubernetes` goes unresponsive.

It looks like there wasn't any parameter that could be used to set a timeout in `kubernetes_asyncio` client and client also does not have any default timeout.

I have made those updates so that we can have a standard timeout for these API requests: https://github.com/truenas/kubernetes_asyncio/pull/20

Reporter has also reported that his `debug` generation was stalled on trying to dump kubernetes service details. I also updated kubernetes service checks in debug generation to fix this.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
@@ -1,3 +1,4 @@
+import asyncio
 import ipaddress
 import itertools
 
@@ -7,6 +8,7 @@ from middlewared.common.listen import ConfigServiceListenSingleDelegate
 from middlewared.schema import Bool, Dict, Int, IPAddr, Patch, returns, Str
 from middlewared.service import accepts, CallError, job, private, ConfigService, ValidationErrors
 
+from .k8s import api_client
 from .utils import applications_ds_name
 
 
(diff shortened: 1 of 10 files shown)


src/middlewared/middlewared/plugins/kubernetes_linux/update.py
```diff
@@ -314,12 +316,24 @@ async def route_interface_choices(self):
...
-            raise CallError('Please configure kubernetes pool.')
-        if not await self.middleware.call('service.started', 'kubernetes'):
-            raise CallError('Kubernetes service is not running.')
+            error = 'Please configure kubernetes pool.'
+        if not error and not await self.middleware.call('service.started', 'kubernetes'):
+            error = 'Kubernetes service is not running.'
+
+        if not error:
+            try:
+                async with api_client({'node': True}, {'request_timeout': 2}) as (api, context):
```
sonicaj, line 329: To confirm, have we tested if 2 seconds is reasonable ?

sonicaj (approved):

src/middlewared/middlewared/plugins/kubernetes_linux/update.py
Qubad786: its almost instant if service is up and serving since its local network – 2000ms should be fine :)

sonicaj: Great, thanks for confirming!

PR #8214, opened 2022-02-06
NAS-114013 / 22.02.1 / Explicitly restart/stop serial-getty service based on user specified settings

sonicaj:
This PR introduces changes to disable serial-getty services by default. Now moving on, we explicitly enable/disable the service on serial port which user has specified for serial console.


src/middlewared/middlewared/etc_files/ttys_config.py
```diff
@@ -1,19 +1,13 @@
-import os
-import signal
-import subprocess
+from middlewared.utils import run
```
themylogin, line 1: This does not generate any config files anymore, can we drop `ttys` pseudo-service and move it to a middleware plugin?

src/middlewared/middlewared/plugins/system.py
```diff
@@ -360,23 +360,48 @@ async def do_update(self, data):
...
+                    )
+                    if cp.returncode:
+                        self.logger.error('Failed to %r serialconsole: %r', action, cp.stderr.decode())
+            if (
                 original_data['serialspeed'] != config_data['serialspeed'] or
                 original_data['serialport'] != config_data['serialport']
             ):
-                generate_grub = True
+                if original_data['serialport'] != config_data['serialport']:
+                    commands = [
```
themylogin, line 386: Where is this variable used?

themylogin (changes requested):
Nice!

Can we request test VM to have two serial ports and then try testing changing this value?

src/middlewared/middlewared/plugins/system.py
sonicaj: Which one ?

themylogin: `commands`

themylogin (changes requested):

themylogin (approved):
Super cool, thank you!

PR #8239, opened 2022-02-10
NAS-114753 / 22.02.1 / Add tests for NFS share parameters

anodos325:
This commit adds tests that verify that NFS
share parameters may be successfully set through the
REST API.

### DIFF
--- a/tests/api2/test_300_nfs.py
+++ b/tests/api2/test_300_nfs.py
@@ -13,7 +13,7 @@ apifolder = os.getcwd()
 sys.path.append(apifolder)
 from functions import PUT, POST, GET, SSH_TEST, DELETE, wait_on_job
 from auto_config import pool_name, ha, hostname
-from auto_config import dev_test
+from auto_config import dev_test, password, user
 # comment pytestmark for development testing with --dev-test
 pytestmark = pytest.mark.skipif(dev_test, reason='Skip for testing')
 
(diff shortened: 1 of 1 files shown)


tests/api2/test_300_nfs.py
```diff
@@ -261,40 +291,230 @@ def test_29_removing_nfs_mountpoint(request):
...
+    	*(sec=sys,rw,subtree_check)
+    """
+    depends(request, ["pool_04", "ssh_password"], scope="session")
+
+    paths = [NFS_PATH, NFS_PATH]
+    results = PUT(f"/sharing/nfs/id/{nfsid}/", {'paths': paths})
+    assert results.status_code == 200, results.text
+
+    exports_paths = [x['path'] for x in parse_exports()]
+    diff = set(exports_paths) ^ set(paths)
```
themylogin, line 311: The same path is exported twice, isn't it? `set` will have only one item so I don't understand how will this test ensure that we have two exports?

anodos325: Dumb oversight on my part. :)
I've updated to create two subdirectories and then add them both as paths (which is more in line with how this feature is designed anyway).

anodos325:
Alright. I'll wait for it to finish separate run through ci (including new tests) to make sure it works in our CI infrastructure properly before merge.

themylogin (approved):

anodos325:
CI passed.

PR #8255, opened 2022-02-11
NAS-114480 / 13.0 / Update truecommand for TN 13

sonicaj:

1. Mark wireguard interface as internal when retrieving used ips in a system
2. Allow matching arguments of process when checking if service has started
3. Update freebsd proc name to reflect it's argument `wg-quick` as `wireguard-go` process is no longer there
4. When truecommand service is started, health check can potentially run up to 30 minutes later, we initiate the health check right after starting truecommand so middleware updates it's status wrt truecommand quickly
5. If API keys changes, we should be clearing out wireguard public/private keys

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -1941,7 +1941,7 @@ class InterfaceService(CRUDService):
 
         """
         list_of_ip = []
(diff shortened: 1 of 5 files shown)


sonicaj:
backport

yocalebo (approved):

PR #8257, opened 2022-02-11
NAS-114777 / 22.02.1 / fix pool_dataset_track_processes api test

yocalebo:
This is failing solely because other processes are showing up as having the path being opened like `systemd-udevd`. Instead of assuming this will be the only process opening the path, just iterate the results and look for the pid.


ericbsd (approved):
LGTM

PR #8264, opened 2022-02-11
NAS-114815 / 22.02.1 / Add regression tests for NFS server config

anodos325:
* verify that NFSv4 support is being enabled / disabled properly
* verify that udp support is being enabled / disabled properly
* verify that NFS-related port configuration options are set
* verify that number of servers is set correctly

### DIFF
--- a/tests/api2/test_300_nfs.py
+++ b/tests/api2/test_300_nfs.py
@@ -62,6 +62,23 @@ def parse_exports():
 
     return rv
 
+
+def parse_server_config(fname="nfs-kernel-server"):
+    results = SSH_TEST(f"cat /etc/default/{fname}", user, password, ip)
+    assert results['result'] is True, results['error']
+    conf = results['output'].splitlines()
+    rv = {}
+
+    for line in conf:
+        if not line or line.startswith("#"):
+            continue
+
+        k,v = line.split("=", 1)
+        rv.update({k: v})
+
+    return rv
+
+
 # Enable NFS server
 def test_01_creating_the_nfs_server():
     paylaod = {"servers": 10,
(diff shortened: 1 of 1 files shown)


anodos325:
Will hold off on merge until CI run through test branch completes successfully.

ericbsd (approved):
LGTM

tests/api2/test_300_nfs.py
```diff
@@ -206,10 +223,20 @@ def test_18_removing_nfs_mountpoint(request):
 
 
 def test_19_updating_the_nfs_service(request):
+    """
+    This test verifies that service can be updated in general,
+    and also that the 'servers' key can be altered.
+    Latter goal is achieved by readig the nfs config file
```
rick-mesta, line 229: typo: readig

tests/api2/test_300_nfs.py
```diff
@@ -561,6 +588,83 @@ def test_36_check_nfsdir_subtree_behavior(request):
...
+    assert "-N 4" in s["RPCNFSDOPTS"], str(s)
+
+    results = PUT("/nfs/", {"v4": True})
+    assert results.status_code == 200, results.text
+
+    s = parse_server_config()
+    assert "-N 4" not in s["RPCNFSDOPTS"], str(s)
+
+
+def test_39_check_nfs_service_udp_parameter(request):
```
rick-mesta, line 618: RFC5661 says:

"_Where an NFSv4.1 implementation supports operation over the IP network protocol, any transport used between NFS and IP MUST be among the IETF-approved congestion control transport protocols.  At the time this document was written, the only two transports that had the above attributes were TCP and the Stream Control Transmission Protocol (SCTP)._ "

Would it make more sense for us to test that if UDP is specified for an NFSv4.x share, we say something about the config not being valid ? Or some such ?

anodos325: That's a good idea. I think we need to convert `v4` to be a RO parameter (for compatibility) and basically allow users to select which protocols they want to support... i.e. allow disabling NFSv3 if they so desire.

PR #8266, opened 2022-02-11
NAS-114777 / 22.02.1 / no seriously, fix pool_dataset_track_processes api

yocalebo:
Quoting issue led to the fact that the child process silently exited because of syntax issue which lead to the fact that the child process was holding on to stdout/stderr which lead to the fact that the the wrong `params` was being passed to the websocket call which lead to the fact that the returned `cmdline` was incorrect which lead to the fact that all of my assertion failure messages were completely worthless. After a shameful amount of time spent tracking all these issues down, I can confirm with 100% certainty that this now successfully runs.....at least on the machine I was using. I dont computer good.


ericbsd (approved):
LGTM

PR #8270, opened 2022-02-11
Bring snapshots annotation back

themylogin:
This reverts commit 0be81b9e0a755ef49b4b3ef47a678c927079f8a9.


yocalebo:
This causes deadlock unless I'm missing a PR that addresses the process pool issue?

themylogin:
@yocalebo this is to fix the test and give UI team some API to work with. The deadlock issue will be handled separately https://jira.ixsystems.com/browse/NAS-114821

PR #8274, opened 2022-02-11
Expose ability to query active NFSv3 and NFSv4 clients

anodos325:


anodos325:
After a little more thinking, users may benefit from being able to query these APIs.

rick-mesta (approved):

PR #8280, opened 2022-02-14
NAS-114315 / 22.02.1 / Do not reset mountpoints while importing pools

Qubad786:
## Background
It was pretty unclear why we would do this. After discussing with Waqar I came to know that we had a faulty behavior merged once that mounted those kinds of paths and that code patch was written to undo those mounts – pretty untrivial that we didn't write some sort of migration for existing users and left this hacky patch around.

For this specific user, he had manually created /mnt/mnt dorectory that caused the actual pool's mountpoints to reset 🤦  

With this, the behavior where PVC's mountpoints were reset, is also fixed.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -1750,17 +1750,6 @@ class PoolService(CRUDService):
                     'Failed to set cache file for %s', pool['name'], exc_info=True,
                 )
 
-            try:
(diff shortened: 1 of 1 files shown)


yocalebo:
So the importing of the zpool was not finalizing because we were resetting the mountpoint??

rick-mesta (approved):

Qubad786:
@yocalebo that's a different issue

sonicaj (approved):

PR #8293, opened 2022-02-15
NAS-114833 / 22.02.1 / Remove disable offload capabilities

yocalebo:
This was added on CORE for an extremely rare situation where we have HA customers using jails/vms and an underlying limitation with the kernel. This doesn't apply to SCALE HA so remove it entirely.


amotin:
IIRC it was not (only) for HA but to avoid interface flap on first plugin bridge attach.  I guess Linux does not flap?

yocalebo:
> IIRC it was not (only) for HA but to avoid interface flap on first plugin bridge attach. I guess Linux does not flap?

It does not from my testing.

amotin:
> > IIRC it was not (only) for HA but to avoid interface flap on first plugin bridge attach. I guess Linux does not flap?
> 
> It does not from my testing.

Good. Just make sure it is tested on different NICs.

themylogin (approved):

PR #8294, opened 2022-02-15
NAS-114850 / 22.02.1 / Add empty `/etc/default/locale` file to be read by `/etc/pam.d/sshd`

themylogin:
We should not have any default locale as https://wiki.debian.org/Locale suggests
("If you have users who access the system through ssh, it is recommended that you
  choose None as your default locale")


themylogin:
time 0:15

sonicaj (approved):

PR #8295, opened 2022-02-16
NAS-114834 / 22.02.1 / Wait for k8s taints to be added

sonicaj:
This commit adds changes to wait for k8s taints to be added as we saw for a user that even after doing a k8s patch with api client, the operation was not idempotent and did not return control when the taints had been added but just the request had been registered with k8s api server. So with this change, we wait for the taints to be added before giving the consumer control back.


sonicaj:
@yocalebo good point. I'll add a timeout for good measure

yocalebo (approved):

PR #8296, opened 2022-02-16
NAS-114793 / 13.0 / fix iscsi portal IP choices when HA and ALUA disabled

yocalebo:
1. When this is an HA system and ALUA is disabled, we don't need to show the controllers IP addresses and only show the VIPs.
2. minor clean up and flake8 fixes.
3. removed the `::` option from `listen_ip_choices` because we don't even return IPv6 addresses in that method so there is no reason to show that option as it only causes confusion


Qubad786 (approved):
LGTM!

PR #8298, opened 2022-02-16
NAS-114821 / 22.02.1 / Abort erroneous websocket connections and handle connection aborts in websocket client

themylogin:


themylogin:
time 8:00

yocalebo (approved):
I can confirm this fixes the problem. Just 1 nit-pick though. If `reason=None` then the log message just shows `reason=b''`. Maybe make it default to `Not provided` or something?

yocalebo (changes requested):
Hold on, I'm seeing the child processes are hanging onto the memory and not releasing it.

yocalebo (approved):
Okay this fixes the immediate problem of causing a deadlock with our process pool so I'll approve it. I'll open up another ticket for the child processes not releasing memory.

PR #8299, opened 2022-02-16
NAS-114895 / 13.0 / Fix enclosure alert 13

yocalebo:
Enclosure plugin was rewritten in 13 to scale on systems with 100's (even 1k+ hdds) but this alert was missed.


rick-mesta:
@themylogin: Vlad, could we get your eyes 👀 on this, please ?

themylogin (approved):

PR #8300, opened 2022-02-16
NAS-114859 / 13.0 / Fix broken serial/console access in CORE

sonicaj:
This commit fixes an issue introduced recently which broke serial/console access.


yocalebo (commented):
There's really no reason to have that global variable and can just append the dict to the array without doing the copy

yocalebo (approved):

rick-mesta (approved):

PR #8302, opened 2022-02-17
NAS-114904 / 22.02.1 / Improve `zfs.snapshot.query` `{"count": True}` performance.

themylogin:
`usage.py` does `call('zfs.snapshot.query', [], {'count': True})` every 24 hours


themylogin:
time 0:30

yocalebo (approved):

PR #8307, opened 2022-02-17
NAS-114784 / 22.02 / Disable subtree checking if NFS export is a mountpoint (#8247)

anodos325:
This is optimization for case where we're exporting entire
ZFS dataset. If share is subdirectory, then presumption is
that we don't want to potentially expose paths outside of
it.


anodos325:
Waiting on feedback from perf team about whether this is needed in 22.02-release.

PR #8308, opened 2022-02-17
NAS-113826 / 22.02.1 / Allow enabling/disabling STP on bridge interfaces at creation

yocalebo:
This is very important for when we're connected to an actual enterprise(y) network that has BPDU guard on. When that feature is turned on and the port the TrueNAS is connected to exchanges BPDUs with the switch, then the (depending on how the switch feature is configured) switch can disable the switchport that the TrueNAS is connected to. Add the ability to toggle this setting at creation of the bridge defaulting to on.

### DIFF
--- a/src/middlewared/middlewared/alembic/versions/22.02/2022-02-17_14-24_add_stp_column.py
+++ b/src/middlewared/middlewared/alembic/versions/22.02/2022-02-17_14-24_add_stp_column.py
@@ -18,7 +18,7 @@ depends_on = None
 
 def upgrade():
     with op.batch_alter_table('network_bridge', schema=None) as batch_op:
-        batch_op.add_column(sa.Column('stp', sa.Boolean(), nullable=False))
+        batch_op.add_column(sa.Column('stp', sa.Boolean(), server_default='1', nullable=False))
 
 
 def downgrade():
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/alembic/versions/22.02/2022-02-17_14-24_add_stp_column.py
```diff
@@ -0,0 +1,26 @@
...
+# revision identifiers, used by Alembic.
+revision = '7a143979d99b'
+down_revision = 'cda35deeed4f'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table('network_bridge', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('stp', sa.Boolean(), nullable=False))
```
themylogin, line 21: Adding non-nullable column will fail if there already are rows in the table. Solutions are here http://docs.middleware.ixsystems.net/database/migrations.html#adding-a-new-column-to-an-existing-table

src/middlewared/middlewared/plugins/network.py
```diff
@@ -33,6 +33,7 @@ class NetworkBridgeModel(sa.Model):
     id = sa.Column(sa.Integer(), primary_key=True)
     members = sa.Column(sa.JSON(type=list), default=[])
     interface_id = sa.Column(sa.ForeignKey('network_interfaces.id', ondelete='CASCADE'))
+    stp = sa.Column(sa.Boolean(), default=True)
```
themylogin, line 36: We don't need defaults here, middleware plugin should insert a row with all the fields set

themylogin (changes requested):

themylogin (approved):

PR #8310, opened 2022-02-17
NAS-114913 / 22.12 / remove ladvd

yocalebo:
1. the service file was generating `s3` configuration (this is a typo obviously but has been there since 12.0 CORE...)
2. the configuration that is passed to us from the end-user is silently ignored (never ported from CORE)
3. the service doesn't have a config file so we have to do something non-trivial just to "configure" this service based on what the end-user provided
4. very small subset of SCALE users are using this (based on stats gathered from Jan 1, 2022 through today).


themylogin (approved):
I don't even know what that is 😳

PR #8312, opened 2022-02-17
Add private API to get / set NFSD threadpool mode

anodos325:


rick-mesta (approved):

yocalebo (commented):
Are these guaranteed to exist always?

anodos325:
> Are these guaranteed to exist always?

As long as user doesn't get overly creative with customizing the system.

PR #8315, opened 2022-02-18
NAS-114762 / 13.0 / Make WebUI Shells persistent

lordlimecat:
Currently, the WebUI shell ends when navigating the UI, getting logged out, or restarting the backend middlewared.

By declaring a WEB_CLIENT variable for web shells and detecting those variables in .zlogin, we can have the web shell survive a `service middlewared restart` as well as timeouts or navigation.

This is especially useful when troublehooting or running long tasks like rsync.

### DIFF
--- a/src/freenas/root/.zlogin
+++ b/src/freenas/root/.zlogin
@@ -1,3 +1,7 @@
+if (( ${+WEB_CLIENT} )); then
+	screen -xRR WebShell
+fi
+
 if [ -f /usr/local/sbin/hactl ]; then
 	/usr/local/sbin/hactl status -q
 fi
--- a/src/middlewared/middlewared/main.py
+++ b/src/middlewared/middlewared/main.py
@@ -591,6 +591,7 @@ class ShellWorkerThread(threading.Thread):
                 'HOME': '/root',
                 'LANG': 'en_US.UTF-8',
                 'PATH': '/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/sbin:/usr/local/bin:/root/bin',
+                'WEB_CLIENT': 'xterm',
             })
 
         # Terminal baudrate affects input queue size


themylogin:
@lordlimecat great idea, but I see two issues with the current implementation
* Now there can only be one Web Shell (two browser windows will use the same `screen`)
* If user starts a resource-consuming app on the web shell and then closes the browser window expecting that app to be terminated, it will continue running indefinitely
We might have to discuss this with our management, I will create a JIRA proposal shortly

lordlimecat:
Thanks-- those issues did occur to me. I also encountered some bad behavior with screen when using 'man find' and scrolling down the --path option; man seemed to seize up and I had to re-enter the session by navigating away. I'm going to poke around and see if I can nail down why it's happening.

I don't have a good solution for a resource-consuming process, this seems a natural and unavoidable corollary to allowing long-running tasks like an rsync. Personally, I would suggest that that risk-- a user essentially shooting themselves in the foot-- is less of a detriment than the risk of interrupting a potentially important CLI task (imagine interrupting a partition resize, for instance).

If the core idea here is desirable, I might suggest an option on the user page to enable a session manager (defaulted to on). That way, anyone whose workflow requires multiple web shells can revert to that. It might not hurt to also allow selection of which session manager (tmux vs shell), if tabs fit someone's workflow. If those options are tied to env variables, it is trivially to enable that functionality in .zlogin.

For the vast majority of users, I suspect that the experience in defaulting to screen sessions is a net plus as things will tend to work more intuitively. Linux's Cockpit project works generally along these lines and if I recall two web shells will simply show a view of the same console.

lordlimecat:
From the Jira thread, it looks like there is interest as long as some UI element makes it explicit that the shell session is persistent. 

I have some python experience but virtually no web UI / framework experience and do not know how to implement that, but if someone can expose that intent via environmental variable (e.g. WEB_SHELL_PERSIST={none | screen | tmux}) I can handle the rest.

william-gr:
There is no point in keeping this PR open so I will be closing this now until we can reach a consensus of implementation that will please most. Thank you for you contribution this far.

PR #8319, opened 2022-02-20
NAS-114929 / 22.02.1 / add python3-nvme as build depends to middleware

yocalebo:


src/middlewared/debian/control
```diff
@@ -68,6 +68,7 @@ Build-Depends: alembic,
                python3-setproctitle,
                python3-setuptools,
                python3-sgio,
+               python3-nvme,
```
sonicaj, line 71: Is it used in middlewared directly ? `py-fenced` debian package is broken right now as stand alone, so the best practice here is to add this there and the debian packaging system can automatically fix the dependencies

sonicaj (changes requested):

src/middlewared/debian/control
yocalebo: `python3-fenced` is already added as a depends here?

sonicaj: If you try installing the `python3-fenced` debian package, it will fail to work on a standard debian machine because it's `Depends` have not been correctly specified.

If it's dependencies are correctly specified, we don't need to add it's dependencies in `middlewared` as it's not using those directly. So this should be added in `py-fenced` instead.

yocalebo: We're using `python3-sgio` directly in middlewared and I expect to use `python3-nvme` too

sonicaj: In that case it should be added in build depends as well for middleware, but the `py-fenced` package should still be fixed so that these packages are addd in it's depends. Although then specifying sgio/nvme in middleware is redundant but i don't mind.

sonicaj: We should be adding this to `Depends` as well ( it's not required though when `py-fenced` is fixed ) and also fixing `py-fenced` debian package to properly fix it there as otherwise the same could be happening with the other packages we have i.e listing their dependencies in middleware whereas it should be listed in their respective packages

sonicaj: To clarify, when we fix `py-fenced` dependencies, this PR is no longer required as required dependencies will already be pulled automatically.

sonicaj:
This is no longer required.

PR #8323, opened 2022-02-21
NAS-114906 / 22.02.1 / Unregister failed calls in the middleware client to prevent memory leak

themylogin:
Additionally investigated memory leak in the main middleware process. tracemalloc led me to the bytearray objects of Websocket request payload being read. Even explicitly clearing those did not free up memory. However, the memory is freed within 5-60 seconds of the "payload too large" websocket error. It seems that those bytearray memory is not given back to the OS immediately and that's how python memory management works. Upon running stress tests I was not able to make middleware process eat up all the memory — it was always freed shortly.

### DIFF
--- a/src/middlewared/middlewared/client/client.py
+++ b/src/middlewared/middlewared/client/client.py
@@ -467,31 +467,33 @@ class Client(object):
 
         c = Call(method, params)
         self._register_call(c)
-        self._send({
-            'msg': 'method',
-            'method': c.method,
-            'id': c.id,
-            'params': c.params,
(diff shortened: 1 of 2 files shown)


themylogin:
time 8:00

themylogin:
time 8:00

yocalebo (approved):
Fun times!

themylogin:
time 0:10

themylogin:
time 0:05

themylogin:
time 0:05

PR #8327, opened 2022-02-21
NAS-114755 / 22.02.1 / Nas-114755

lordlimecat:
First stab at this.
This first pull request is mostly Todos, with some minor logic introduced in control.py. I am seeking feedback on my approach here.

The idea is to perform the following:

- When an interface is added to a bridge, remove its IP from the actual interface (but not from the database)
- When it is removed from the bridge, re-add its IP from the DB
- When an IP change is made to a bridge member, store that change in DB but do not commit it to the interface, and communicate (somehow) this fact (to UI? to logs)

That last case is to allow someone to stage an IP address to go live when the interface is removed from bridge. This can allow avoiding chicken-and-egg problems where bridge removal destroys connectivity. 

Please provide feedback, and if this looks like a good approach I will proceed.

### DIFF
--- a/src/middlewared/middlewared/plugins/interface/bridge.py
+++ b/src/middlewared/middlewared/plugins/interface/bridge.py
@@ -33,6 +33,7 @@ class InterfaceService(Service):
             # remove members from the bridge that aren't in the db
             self.logger.info('Removing member interface %r from %r', member, name)
             iface.delete_member(member)
+            ## TODO: Add IP address back to former bridge member based on db
 
         for member in db_members - os_members:
             # now add members that are written in db but do not exist in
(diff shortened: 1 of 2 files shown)


lordlimecat:
If this is a good approach, the next commits would be to implement the logic structure but not affect how things work. Implement logging, and the changes can be tested without risking regression.

yocalebo (commented):
@lordlimecat thanks for your contribution. It's bad timing but I'm getting ready to push a PR that dramatically changes pretty much all of this so this becomes obsolete almost immediately. Maybe opening a ticket first and describing what you're trying to accomplish would be best. If you can respond with the ticket number here, I'll be sure and pick it up and we discuss further there.

yocalebo:
Oh wait, this is for the existing ticket already. So this PR should be opened up against truenas/13.0-stable branch. The "master" branch is for SCALE only.

lordlimecat:
If that's the case it can be discarded-- and seems like a good justification for having opened the PR with todos. No harm no foul.

PR #8334, opened 2022-02-23
NAS-114967 / 13.0 / add send_event key to datastore plugin

yocalebo:
By default, when an update/insert/delete operation occurs we will
emit an event via our event plugin to be processed for the webui.
This is important, for example, when a new disk is inserted/removed.
In either of the above scenarios, the webUI will process this event
and update the front-end accordingly. It negates the front-end having
to poll the backend (which is expensive). However, on very large
systems (i.e. systems with 100+ disks) emitting an event can become absurdly
expensive. This reason why this becomes expensive is because for every
db operation, we run the plugins associated "query" method. So if we
update 1000 table entries, then we run "disk.query" 1000 times. In
real world testing, this has shown to take roughly 82 seconds to update
100 entries on the `storage_disk` table when there are 641 entries total.
The database was on a NVMe disk. The solution to this is adding the
`send_event` key. If this is set to False, then an event will not be
sent for the db operation. It is the callers responsibility to emit an event
after all the db operations are complete.

### DIFF
--- a/src/middlewared/middlewared/plugins/datastore/write.py
+++ b/src/middlewared/middlewared/plugins/datastore/write.py
@@ -22,7 +22,7 @@ update 1000 table entries, then we run "disk.query" 1000 times. In
 real world testing, this has shown to take roughly 82 seconds to update
 100 entries on the `storage_disk` table when there are 641 entries total.
 The database was on a NVMe disk. The solution to this is adding the
-`send_event` key. If this is set to False, then an event will not be
+`send_events` key. If this is set to False, then an event will not be
 sent for the db operation. It is the callers responsibility to emit an event
 after all the db operations are complete.
 """
@@ -40,7 +40,7 @@ class DatastoreService(Service, FilterMixin, SchemaMixin):
             'options',
             Bool('ha_sync', default=True),
             Str('prefix', default=''),
-            Bool('send_event', default=True),
+            Bool('send_events', default=True),
         ),
     )
     async def insert(self, name, data, options):
@@ -74,7 +74,7 @@ class DatastoreService(Service, FilterMixin, SchemaMixin):
 
         await self._handle_relationships(pk, relationships)
 
-        if options['send_event']:
+        if options['send_events']:
             await self.middleware.call('datastore.send_insert_events', name, insert)
 
         return pk
@@ -87,7 +87,7 @@ class DatastoreService(Service, FilterMixin, SchemaMixin):
             'options',
             Bool('ha_sync', default=True),
             Str('prefix', default=''),
-            Bool('send_event', default=True),
+            Bool('send_events', default=True),
         ),
     )
     async def update(self, name, id_or_filters, data, options):
@@ -124,7 +124,7 @@ class DatastoreService(Service, FilterMixin, SchemaMixin):
             if result.rowcount != 1:
                 raise RuntimeError('No rows were updated')
 
-            if options['send_event']:
+            if options['send_events']:
                 await self.middleware.call('datastore.send_update_events', name, id)
 
         await self._handle_relationships(id, relationships)
@@ -180,7 +180,7 @@ class DatastoreService(Service, FilterMixin, SchemaMixin):
             'options',
             Bool('ha_sync', default=True),
             Str('prefix', default=''),
-            Bool('send_event', default=True),
+            Bool('send_events', default=True),
         ),
     )
     async def delete(self, name, id_or_filters, options):
@@ -197,7 +197,7 @@ class DatastoreService(Service, FilterMixin, SchemaMixin):
             },
         )
 
-        if not isinstance(id_or_filters, list) and options['send_event']:
+        if not isinstance(id_or_filters, list) and options['send_events']:
             await self.middleware.call('datastore.send_delete_events', name, id_or_filters)
 
         return True
--- a/src/middlewared/middlewared/plugins/datastore/write.py
+++ b/src/middlewared/middlewared/plugins/datastore/write.py
@@ -8,6 +8,26 @@ from .filter import FilterMixin
 from .schema import SchemaMixin
 
 
+"""
+By default, when an update/insert/delete operation occurs we will
+emit an event via our event plugin to be processed for the webui.
+This is important, for example, when a new disk is inserted/removed.
+In either of the above scenarios, the webUI will process this event
+and update the front-end accordingly. It negates the front-end having
+to poll the backend (which is expensive). However, on very large
+systems (i.e. systems with 100+ disks) emitting an event can become absurdly
+expensive. This reason why this becomes expensive is because for every
+db operation, we run the plugins associated "query" method. So if we
+update 1000 table entries, then we run "disk.query" 1000 times. In
+real world testing, this has shown to take roughly 82 seconds to update
+100 entries on the `storage_disk` table when there are 641 entries total.
+The database was on a NVMe disk. The solution to this is adding the
+`batch_operation` key. If this is set to True, then an event will not be
+sent for the db operation. It is the callers responsibility to emit an event
+after all the db operations are complete.
+"""
+
+
 class DatastoreService(Service, FilterMixin, SchemaMixin):
 
     class Config:
@@ -20,6 +40,7 @@ class DatastoreService(Service, FilterMixin, SchemaMixin):
             'options',
             Bool('ha_sync', default=True),
             Str('prefix', default=''),
+            Bool('batch_operation', default=False),
         ),
     )
     async def insert(self, name, data, options):
@@ -53,7 +74,8 @@ class DatastoreService(Service, FilterMixin, SchemaMixin):
 
         await self._handle_relationships(pk, relationships)
 
-        await self.middleware.call('datastore.send_insert_events', name, insert)
+        if not options['batch_operation']:
+            await self.middleware.call('datastore.send_insert_events', name, insert)
 
         return pk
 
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -55,6 +55,7 @@
 ]
 ZPOOL_CACHE_FILE = '/data/zfs/zpool.cache'
 ZPOOL_KILLCACHE = '/data/zfs/killcache'
+TRUENAS_DB = '/data/freenas-v1.db'
```
themylogin, line 58: `from middlewared.utils.db import FREENAS_DATABASE`

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -778,12 +779,33 @@ async def do_create(self, job, data):
         asyncio.ensure_future(self.restart_services())
 
         pool = await self.get_instance(pool_id)
-        await self.middleware.call_hook('pool.post_create_or_update', pool=pool)
-        await self.middleware.call_hook(
-            'dataset.post_create', {'encrypted': bool(encryption_dict), **encrypted_dataset_data}
-        )
+        if not await self.middleware.call('failover.licensed'):
```
themylogin, line 782: Can it be better to deal with this in the specific hook itself?

themylogin (changes requested):

src/middlewared/middlewared/plugins/pool.py
yocalebo: I got this error and didn't want to mess with it.
```
  File "/usr/local/lib/python3.9/site-packages/middlewared/main.py", line 1074, in _call_hook_base                                                                                                                                                                                                                               yield hook, hook['method'](self, *args, **kwargs)                                                                                                                                                                                                                                                                        TypeError: hook_sync_geli() got an unexpected keyword argument 'ha_sync'
```

src/middlewared/middlewared/plugins/pool.py
yocalebo: That doesn't exist on 13. Furthermore, the product isn't called freenas anymore

src/middlewared/middlewared/plugins/pool.py
yocalebo: When trying to do these changes inside the hook itself.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -778,12 +779,33 @@ async def do_create(self, job, data):
...
+                'pool_create', pool, encryption_dict, encrypted_dataset_data
+            ))
+
         return pool
 
+    @private
+    async def sync_zfs_guid_and_sync_db_to_remote_node(self, schema, pool, enc_dict=False, enc_data={}):
+        await self.middleware.call('disk.sync_zfs_guid', pool, False)
+        try:
+            await self.middleware.call('failover.send_small_file', TRUENAS_DB)
```
william-gr, line 802: I am not sure you can do that... replacing the entire database underneath middlewared feet on passive node, a lot of hilarity can ensue...
I thought there was a method to re-send the whole database using sqlite?

src/middlewared/middlewared/plugins/datastore/write.py
```diff
@@ -20,6 +40,7 @@ class Config:
             'options',
             Bool('ha_sync', default=True),
             Str('prefix', default=''),
+            Bool('batch_operation', default=False),
```
themylogin, line 43: Can we give it a name that truly reflects its purpose? E.g. `Bool('send_events', default=True),`

src/middlewared/middlewared/plugins/disk_/zfs_guid.py
```diff
@@ -38,38 +33,32 @@ async def sync_zfs_guid(self, pool_id_or_pool):
...
+                    self.logger.debug("Removing disk %r zfs_guid as %r has it", disk["identifier"], devname)
+                    events.update({"id": disk["identifier"], "zfs_guid": None})
                     await self.middleware.call(
-                        "datastore.update", "storage.disk", disk["identifier"], {"zfs_guid": None}, {"prefix": "disk_"},
+                        "datastore.update", "storage.disk", disk["identifier"],
+                        {"zfs_guid": None}, {"prefix": "disk_", "batch_operation": True},
                     )
 
+        if events:
+            self.middleware.send_event('disk.query', 'CHANGED', fields=events)
```
themylogin, line 57: This will send only one event for the latest disk updated.

themylogin (changes requested):
Does this issue also affect `disk.sync_all`?

yocalebo:
> Does this issue also affect `disk.sync_all`?

Yes it definitely does and I originally was going to push a separate PR to fix sync_all but I went ahead and included it here in this PR.

src/middlewared/middlewared/plugins/disk_/zfs_guid.py
```diff
@@ -38,38 +33,32 @@ async def sync_zfs_guid(self, pool_id_or_pool):
...
+                    self.logger.debug("Removing disk %r zfs_guid as %r has it", disk["identifier"], devname)
+                    events.append({"id": disk["identifier"], "zfs_guid": None})
                     await self.middleware.call(
-                        "datastore.update", "storage.disk", disk["identifier"], {"zfs_guid": None}, {"prefix": "disk_"},
+                        "datastore.update", "storage.disk", disk["identifier"],
+                        {"zfs_guid": None}, {"prefix": "disk_", "send_event": False},
                     )
 
+        if events:
+            self.middleware.send_event('disk.query', 'CHANGED', fields=events)
```
themylogin, line 57: This does not match the signature of events that were sent by the original code (it was an event per single row, with `id` field set). Same for `disk.sync_all`.

themylogin (changes requested):

src/middlewared/middlewared/plugins/disk_/zfs_guid.py
yocalebo: I understand but I don't know of a better way of doing this. I figured the webUI is going to look for the difference. Maybe the `id` key can be `BATCH` or something and they can catch that.

themylogin: @yocalebo can you send multiple events in a `for` loop?

yocalebo: oh...you're right. I thought `middleware.send_event` was doing the `disk.query` and was trying to send in 1 call but that's not the case. I'll fix.

src/middlewared/middlewared/plugins/disk_/sync.py
```diff
@@ -208,18 +215,26 @@ async def sync_all(self, job):
...
         # make sure the database entries for enclosure slot information for each disk
         # matches with what is reported by the OS
         await self.middleware.call('enclosure.sync_disks')
 
-        if changed:
+        if changed or deleted:
             await self.middleware.call('disk.restart_services_after_sync')
+
+        for change in changed:
+            self.middleware.send_event('disk.query', 'CHANGED', id=change['id'], fields=change['disk'])
```
themylogin, line 234: Thank you, it's almost perfect now!

Unfortunately `change['disk']` now contains raw database fields: keys with `disk_` prefix, `disk.disk_extend` method not applied... API clients expect this to be the same as `disk.query` returns. I think the simplest way will be to re-query those objects using `call("disk.query", [["id", "in", [...]])`

themylogin (changes requested):

src/middlewared/middlewared/plugins/disk_/sync.py
```diff
@@ -176,20 +183,26 @@ async def sync_all(self, job):
...
-                )
-                if qs:
+                if qs is None:
+                    try:
+                        qs = await self.middleware.call('datastore.query', 'storage.disk', {'get': True})
+                    except MatchNotFound:
+                        # there isn't a single db entry for a disk??? not likely but better safe
+                        # than sorry
+                        qs = {}
+                if disk := qs.get(disk_identifier):
```
themylogin, line 204: `qs` will be an array, not a `dict`, won't it be?

themylogin (changes requested):

src/middlewared/middlewared/plugins/datastore/write.py
```diff
@@ -20,6 +40,7 @@ class Config:
             'options',
             Bool('ha_sync', default=True),
             Str('prefix', default=''),
+            Bool('send_event', default=True),
```
themylogin, line 43: These are `events`

src/middlewared/middlewared/plugins/disk_/sync.py
```diff
@@ -208,18 +216,27 @@ async def sync_all(self, job):
...
 
         # make sure the database entries for enclosure slot information for each disk
         # matches with what is reported by the OS
         await self.middleware.call('enclosure.sync_disks')
 
-        if changed:
+        if changed or deleted:
             await self.middleware.call('disk.restart_services_after_sync')
+            disks = await self.middleware.call('disk.query', [], {'prefix': 'disk_'})
+            for change, delete in zip_longest(changed, deleted, fillvalue=None):
```
themylogin, line 234: What does `zip_longest` do here?

themylogin (changes requested):

src/middlewared/middlewared/plugins/disk_/sync.py
yocalebo: Allows us to iterate over `change` and `delete` using  a single for loop and it also provides the `fillvalue` keyword so if the 2 variables are uneven then it won't break the loop early (like the standard `zip` does).

themylogin (approved):
Thank you!

rick-mesta (approved):

PR #8340, opened 2022-02-25
NAS-115009 / 22.12 / Enable Mixed port types in kubernetes services

Ornias1993:
This flag enables mixed port types for kubernetes services for bluefin.
As the kubernetes feature will move to Beta state come 1.24 and this feature flag is one of the most used and most usefull featureflags, as some Apps and loadbalancers will not work correctly without this.

It will also allow catalog maintainers significant cleanup of their GUI, as it will not require per-port-type service anymore.

**note:**
*Can also be backported to 22.02 if someone wants that in there...*

### DIFF
--- a/src/middlewared/middlewared/etc_files/rancher/k3s/flags.py
+++ b/src/middlewared/middlewared/etc_files/rancher/k3s/flags.py
@@ -28,6 +28,7 @@ def render(service, middleware):
         'audit-log-maxbackup=10',
         'audit-log-maxsize=100',
         'service-account-lookup=true',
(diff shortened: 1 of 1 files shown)


Ornias1993:
@sonicaj

sonicaj (approved):
Thanks!

Ornias1993:
@sonicaj As the 22.02.1 nightly also ship the newer kubernetes version, did you make sure to also backport this change?
It will significantly help cleanup and streamline Apps prior to Bluefin launch... (without affecting current users, as it does not apply unless actually used by app creator)

PR #8342, opened 2022-02-25
Fix logger call in exports etc file

anodos325:
This fixes a crash when exports are generated with
invalid / non-existing user for maproot or mapall


yocalebo (approved):

anodos325:
backport

anodos325:
backport

PR #8354, opened 2022-02-25
NAS-114837 / 22.12 / dont migrate vlan/br/bond interfaces on SCALE (they're already migrated)

yocalebo:


themylogin (approved):

src/middlewared/middlewared/plugins/interface/link_address.py
```diff
@@ -80,48 +78,54 @@ async def setup(middleware):
...
+        for db_interface in db_interfaces:
+            if db_interface["interface"].startswith(("vlan", "br", "bond")):
+                # vlan interfaces dont need to be migrated since they
+                # share the same name between CORE and SCALE
+                # "br" and "bond" interface names are already converted
+                # to scale
+                continue
+
+            if m := RE_FREEBSD_BRIDGE.match(db_interface["interface"]):
+                name = f"br{m.group(1)}"
```
sonicaj, line 90: Would this condition ever be true ?

themylogin: Why not? Didn't we have `bridgeN` interfaces in the database in CORE?

PR #8358, opened 2022-02-25
Removal of proto states

aidik:
Due to a complaint of Georgian users (https://jira.ixsystems.com/browse/DOCS-3260), I removed Abkhazia and South Ossetia. Those quasi republics are de iure integral parts of Georgia and de facto areas occupied by Russia. 

I believe that this is necessary because of the way how those areas are handled (https://github.com/truenas/middleware/blob/f037ba76f130a2edddf2b44a72c0226d5f500a56/src/middlewared/middlewared/plugins/system.py#L946-L948) makes us take a political statement that should not be part of this software. 

Also, I don't think that any other columns but 'Common Name' and 'ISO 3166-1 2 Letter Code' are used so there is no need to re-number the other rows in the csv. 

This issue is manifesting in both SCALE and CORE.


aidik:
Just a note. This seems to be the same in both SCALE and CORE.

kmoore134 (approved):

william-gr (approved):

PR #8360, opened 2022-02-26
NAS-115010 / 22.12 / Disable docker-compose cli usage

sonicaj:


debian/debian/postinst
```diff
@@ -56,6 +56,9 @@ usermod -a -G dialout nut
 # Usbhid-ups driver does not start and needs nut as the group for it's executable to correctly execute NAS-113642
 chown root:nut /lib/nut/usbhid-ups
 
+# We would like to disable docker-compose to prevent people from foot-shooting
+chmod -x /usr/bin/docker-compose
```
Ornias1993, line 60: There happen to be two binaries that both need to be disabled:
`/usr/bin/docker-compose`
and
`/bin/docker-compose`

Local testing shows that just disabling `/usr/bin/docker-compose` leads to it falling back to `/bin/docker-compose`

Ornias1993 (changes requested):

debian/debian/postinst
sonicaj:
```
❯ ls -l /bin/docker-compose
-rwxr-xr-x 1 root root 420 Nov 22  2019 /bin/docker-compose
❯ ls -l /usr/bin/docker-compose
-rwxr-xr-x 1 root root 420 Nov 22  2019 /usr/bin/docker-compose
❯
❯ chmod -x /usr/bin/docker-compose
❯ ls -l /usr/bin/docker-compose
-rw-r--r-- 1 root root 420 Nov 22  2019 /usr/bin/docker-compose
❯ ls -l /bin/docker-compose
-rw-r--r-- 1 root root 420 Nov 22  2019 /bin/docker-compose
```
I am not sure why it was not working for you - it works nicely for me ^^^

yocalebo (approved):

debian/debian/postinst
Ornias1993: Could you try executing the docker-compose command itself instead of LS?
That's what showed me it didn't work previously.

Ornias1993: (just double checking this is going to have the appropiate result though)

sonicaj:
```
❯ docker-compose
zsh: permission denied: docker-compose
```

Ornias1993: Coolio :)
No idea why my test failed then 👍

Ornias1993 (approved):

PR #8361, opened 2022-02-26
NAS-115036 / 22.12 / fix bridge migration on upgrade from core

yocalebo:


src/middlewared/middlewared/plugins/interface/link_address.py
```diff
@@ -79,11 +79,13 @@ async def setup(middleware):
...
-                # vlan interfaces dont need to be migrated since they
-                # share the same name between CORE and SCALE
-                # "br" and "bond" interface names are already converted
-                # to scale
+            if db_interface["interface"].startswith(("vlan", "bond")):
+                # "vlan" interfaces dont need to be migrated since they
+                # share the same name as CORE
+                # "bond" has already been migrated
+                continue
+            elif db_interface["interface"][:2] == "br" and db_interface["interface"].find("bridge") == -1:
```
sonicaj, line 87: Do we require another conditional clause with the same logic in it's block ? Better to add this as an OR condition above

sonicaj (approved):

src/middlewared/middlewared/plugins/interface/link_address.py
yocalebo: I fixed 👍

sonicaj: Thanks!

src/middlewared/middlewared/plugins/interface/link_address.py
```diff
@@ -79,11 +79,10 @@ async def setup(middleware):
 
         # Migrate BSD network interfaces to Linux
         for db_interface in db_interfaces:
-            if db_interface["interface"].startswith(("vlan", "br", "bond")):
-                # vlan interfaces dont need to be migrated since they
-                # share the same name between CORE and SCALE
-                # "br" and "bond" interface names are already converted
-                # to scale
+            iface = db_interface["interface"]
+            if iface.startswith(("vlan", "bond")) or iface[:2] == "br" and iface.find("bridge") == -1:
```
sonicaj, line 83: Let's please have the last 2 in brackets ( the bridge chekcs )

PR #8364, opened 2022-02-28
NAS-115059 / 13.0 / Optimize SMB passdb sync

anodos325:
Design has changed significantly since this code was originally
written. Passdb sync and updates now only happen when protected
behind a job and so this area can be significantly optimized.
(Removing extra datastore queries and subprocess calls).

### DIFF
--- a/src/middlewared/middlewared/plugins/smb_/passdb.py
+++ b/src/middlewared/middlewared/plugins/smb_/passdb.py
@@ -15,6 +15,20 @@ class SMBService(Service):
         service_verb = 'restart'
 
     @private
+    async def smbpasswd_dump(self):
+        out = {}
+        p = await run([SMBCmd.PDBEDIT.value, '-d', '0', '-Lw'], check=False)
+        if p.returncode != 0:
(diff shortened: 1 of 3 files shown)


anodos325:
There are some roughly equivalent changes to make in SCALE and so I've flagged for backport (but won't merge immediately).
I need to run this through CI before merging.

src/middlewared/middlewared/plugins/smb_/passdb.py
```diff
@@ -14,6 +14,20 @@ class Config:
         service = 'cifs'
         service_verb = 'restart'
 
+    @private
+    async def smbpasswd_dump(self):
+        out = {}
+        p = await run([SMBCmd.PDBEDIT.value, '-d', '0', '-Lw'], check=False)
+        if p.returncode != 0:
+            CallError(f'Failed to dump passdb file: {p.stderr.decode()}')
```
yocalebo, line 22: Missing `raise` here?

yocalebo (changes requested):

yocalebo (approved):

rick-mesta (approved):

PR #8387, opened 2022-03-01
NAS-115083 / 22.12 / Add link state of interfaces for reporting realtime event source

sonicaj:
This PR adds changes to add link state for reporting realtime event source as this will be useful for the UI to update the link state for interfaces based on recent changes which might have been applied by the user. Also this removes internal interfaces from the `reporting.realtime` output.

### DIFF
--- a/src/middlewared/middlewared/plugins/interface/netif_linux/netif.py
+++ b/src/middlewared/middlewared/plugins/interface/netif_linux/netif.py
@@ -32,8 +32,9 @@ def destroy_interface(name):
         run(["ip", "link", "set", name, "down"])
 
 
-def get_interface(name):
-    return list_interfaces()[name]
+def get_interface(name, safe_retrieval=False):
+    ifaces = list_interfaces()
+    return ifaces.get(name) if safe_retrieval else ifaces[name]
 
 
 def list_interfaces():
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/reporting/events.py
```diff
@@ -208,7 +209,18 @@ def run_sync(self):
             stats_time = time.time()
             for i in glob.glob('/sys/class/net/*/statistics'):
                 iface_name = i.replace('/sys/class/net/', '').split('/')[0]
-                data['interfaces'][iface_name]['speed'] = last_interface_speeds['speeds'].get(iface_name)
+                if iface_name.startswith(tuple(netif.INTERNAL_INTERFACES)):
```
themylogin, line 212: Can you please add this as a method? It seems that it can be used in other middleware components?

src/middlewared/middlewared/plugins/reporting/events.py
```diff
@@ -208,7 +209,18 @@ def run_sync(self):
...
+                    continue
+
+                try:
+                    iface_obj = netif.get_interface(iface_name)
```
themylogin, line 216: Can you please optimize this method's implementation while we're here?

themylogin (changes requested):
Great job, but since this is ran every 2 seconds, I'm going to be picky

src/middlewared/middlewared/plugins/interface/netif_linux/netif.py
```diff
@@ -32,8 +32,9 @@ def destroy_interface(name):
         run(["ip", "link", "set", name, "down"])
 
 
-def get_interface(name):
-    return list_interfaces()[name]
+def get_interface(name, safe_retrieval=False):
```
themylogin, line 35: Where is this argument used and what is it for?

themylogin (changes requested):

themylogin (approved):

PR #8388, opened 2022-03-01
NAS-115066 / 22.12 / Add truecommand connection information to debug

sonicaj:
Sample output in the debug
```
+--------------------------------------------------------------------------------+
+Truecommand connection - midclt call truecommand.connected | jq 'del(.truecommand_ip,.truecommand_url)' @1646156868 +
+--------------------------------------------------------------------------------+
{
  "connected": true,
  "status": "CONNECTED",
  "status_reason": "Truecommand service is connected."
}
```

### DIFF
--- a/src/freenas/usr/local/libexec/freenas-debug/system/system.sh
+++ b/src/freenas/usr/local/libexec/freenas-debug/system/system.sh
@@ -207,4 +207,8 @@ system_func()
 	section_header "Failed updates /data/update.failed"
 	sc /data/update.failed
 	section_footer
+
+	section_header "Truecommand connection"
+	midclt call truecommand.connected | jq 'del(.truecommand_ip,.truecommand_url)'
+	section_footer
 }	


yocalebo (approved):
Looks better if you remove `| jq 'del(.truecommand_ip,.truecommand_url)'` from the `section_header` imo.

sonicaj:
Sure, however that would imply that the output of that command would not contain ip/uri as in the header it wasn't mentioned we are running the jq bit which i think can be problematic in terms of causing confusion. Thoughts ?

yocalebo:
> Sure, however that would imply that the output of that command would not contain ip/uri as in the header it wasn't mentioned we are running the jq bit which i think can be problematic in terms of causing confusion. Thoughts ?

The `+---+` is only like 80 chars or something and the header you've got is significantly longer but that's purely cosmetic so up to you 😄 i don't mind either way

sonicaj:
@yocalebo i removed the command altogether :D The title is sufficient to show what is being shown :p

PR #8389, opened 2022-03-01
NAS-114954 / 22.12 / refactor network_interfaces table

yocalebo:
This does, essentially, the same thing that was done in https://github.com/truenas/middleware/pull/8030. This PR is for  the `network_interface` table instead. It had a bunch of redundant, unnecessary columns that made the backend code considerably more complex than what it needed to be. I've broken the commits up into small chunks so they're easy to follow and I've updated the unit tests as well.

### DIFF
--- a/src/middlewared/middlewared/plugins/interface/configure.py
+++ b/src/middlewared/middlewared/plugins/interface/configure.py
@@ -17,23 +17,14 @@ class InterfaceService(Service):
     @private
     def configure(self, data, aliases, options):
         options = options or {}
-
         name = data['int_interface']
         self.logger.info('Configuring interface %r', name)
-
         iface = netif.get_interface(name)
-
-        addrs_database = set()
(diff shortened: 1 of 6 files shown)


yocalebo:
retest this please

src/middlewared/middlewared/plugins/interface/configure.py
```diff
@@ -165,13 +114,15 @@ def configure(self, data, aliases, options):
             try:
```
themylogin, line 114: Our Linux netif implementation does not manage descriptions, we might just remove this (keep in API for user's notes)

themylogin (approved):

src/middlewared/middlewared/plugins/interface/configure.py
yocalebo: Good point. I'll make another ticket/PR.

PR #8390, opened 2022-03-01
NAS-115098 / 22.12 / Add ability to specify render_context for etc files

anodos325:
It's often the case that several members of the same etc files
group will perform identical queries. Add ability to specify
a general configure context (methods whose results should be
avaliable to all members of the etc_files group) to avoid
duplicate queries. This is helpful for cases where method
being called is expensive.

### DIFF
--- a/src/middlewared/middlewared/plugins/etc.py
+++ b/src/middlewared/middlewared/plugins/etc.py
@@ -30,7 +30,7 @@ class MakoRenderer(object):
     def __init__(self, service):
         self.service = service
 
-    async def render(self, path):
+    async def render(self, path, ctx):
         try:
             # Mako is not asyncio friendly so run it within a thread
             def do():
@@ -44,6 +44,7 @@ class MakoRenderer(object):
                     FileShouldNotExist=FileShouldNotExist,
                     IS_FREEBSD=osc.IS_FREEBSD,
                     IS_LINUX=osc.IS_LINUX,
+                    render_ctx=ctx
                 )
 
             return await self.service.middleware.run_in_thread(do)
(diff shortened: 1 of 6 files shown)


anodos325:
Had thought about this before lunch and thought this sort of capability might be useful to have. Although not strictly necessary for NFS case (these particular queries are pretty fast), there are other cases where it may be beneficial. NFS is here is example of idea.

src/middlewared/middlewared/plugins/etc.py
```diff
@@ -335,12 +339,25 @@ def __init__(self, *args, **kwargs):
             'py': PyRenderer(self),
         }
 
+    async def gather_context(self, methods):
```
themylogin, line 342: Let's call it consistently: it's either `context` or `ctx`, not both

src/middlewared/middlewared/plugins/etc.py
```diff
@@ -335,12 +339,25 @@ def __init__(self, *args, **kwargs):
...
+
+        return rv
+
     async def generate(self, name, checkpoint=None):
         group = self.GROUPS.get(name)
         if group is None:
             raise ValueError('{0} group not found'.format(name))
 
         async with self.LOCKS[name]:
+            if group[0]['type'] == 'ctx':
```
themylogin, line 355: Looks a little bit hacky. Can we just change definition style for groups that require a context? Something like
```
        'nfsd': {
            'ctx': {'methods': ['sharing.nfs.query', 'nfs.config']}},
            'entries': [
                {'type': 'mako', 'path': 'default/nfs-common'},
                {'type': 'mako', 'path': 'default/nfs-kernel-server'},
            ],
```

themylogin (changes requested):
Great idea!

src/middlewared/middlewared/plugins/etc.py
anodos325: Good idea.

themylogin (approved):
Super, thank you!

yocalebo (approved):

PR #8392, opened 2022-03-02
NAS-115094 / 22.02 / Allow Cluster Locations for CloudSync

kmoore134:
Setting up my first cluster here and found that the validation for CloudSync was preventing me from saving backup jobs pointed at the /cluster/ mount points. Added the override flag to the validation to let those pass. Testing here, appears to be working properly.


themylogin (approved):

anodos325:
We have handling for "CLUSTER:" prefix for paths in most of the filesystem plugins. I think we should be consistent and require "CLUSTER:/mydir" rather than directly exposing fuse mountpoint.

william-gr (changes requested):
Agreed, do you want us to take care of it, Kris?

kmoore134:
Yes please @william-gr that would be helpful!

kmoore134:
Oh, this will need to be fixed as well for it to work:

https://github.com/truenas/middleware/blob/master/src/middlewared/middlewared/plugins/cloud_sync.py#L145

PR #8394, opened 2022-03-02
Fix serial console setting not being properly reflected

Qubad786:
Improvements to console keymap:

- [x] Add keyboard layouts validation into general settings
- [x] Apply apply configured keymap to the console


src/middlewared/middlewared/plugins/service_/services/pseudo/misc.py
```diff
@@ -199,7 +199,7 @@ async def start(self):
 class SysconsService(SimpleService):
```
themylogin, line 199: Can we get rid of this pseudo-service and just do the job directly in the plugin?

themylogin (changes requested):

src/middlewared/middlewared/plugins/service_/services/pseudo/misc.py
Qubad786: Sure!

themylogin (approved):
Super cool, thank you!

PR #8413, opened 2022-03-03
NAS-115119 / 22.12 / Convert SSH etc group to use common render_ctx

anodos325:
This also adds more rigid control over permissions of config and
key files. We will also begin deleting unexpected contents from
/etc/ssh if they are detected.

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/ssh/config.py
+++ b/src/middlewared/middlewared/etc_files/local/ssh/config.py
@@ -1,36 +1,83 @@
 import base64
 import os
 import re
+import stat
+import shutil
 
-from middlewared.utils import osc
+from contextlib import suppress
 
-if osc.IS_FREEBSD:
-    SSH_CONFIG_PATH = '/usr/local/etc/ssh'
-else:
-    SSH_CONFIG_PATH = '/etc/ssh'
+SSH_CONFIG_PATH = '/etc/ssh'
+
+SSH_KEYS = [
+    'ssh_host_key', 'ssh_host_key.pub', 'ssh_host_dsa_key', 'ssh_host_dsa_key.pub', 'ssh_host_dsa_key-cert.pub',
+    'ssh_host_ecdsa_key', 'ssh_host_ecdsa_key.pub', 'ssh_host_ecdsa_key-cert.pub', 'ssh_host_rsa_key',
+    'ssh_host_rsa_key.pub', 'ssh_host_rsa_key-cert.pub', 'ssh_host_ed25519_key', 'ssh_host_ed25519_key.pub',
+    'ssh_host_ed25519_key-cert.pub'
+]
+
+DEFAULT_FILES = ['moduli', 'ssh_config', 'ssh_config.d', 'sshd_config', 'sshd_config.d']
 
 
-def generate_ssh_config(middleware):
-    ssh_config = middleware.call_sync('ssh.config')
-    for k in [
-        'ssh_host_key', 'ssh_host_key.pub', 'ssh_host_dsa_key', 'ssh_host_dsa_key.pub', 'ssh_host_dsa_key-cert.pub',
-        'ssh_host_ecdsa_key', 'ssh_host_ecdsa_key.pub', 'ssh_host_ecdsa_key-cert.pub', 'ssh_host_rsa_key',
-        'ssh_host_rsa_key.pub', 'ssh_host_rsa_key-cert.pub', 'ssh_host_ed25519_key', 'ssh_host_ed25519_key.pub',
-        'ssh_host_ed25519_key-cert.pub'
-    ]:
+def generate_ssh_config(middleware, ssh_config, dirfd):
+    for k in SSH_KEYS:
+        must_remove = True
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/etc_files/local/ssh/config.py
```diff
@@ -1,36 +1,84 @@
...
+    for k in SSH_KEYS:
+        must_remove = True
         s_key = re.sub(r'([.-])', '_', k).replace('ssh_', '', 1)
         if ssh_config[s_key]:
             decoded_key = base64.b64decode(ssh_config[s_key])
             if decoded_key:
-                with open(os.path.join(SSH_CONFIG_PATH, k), 'wb') as f:
+                mode = 0o644 if k.endswith('.pub') else 0o600
+
+                fd = os.open(k, os.O_WRONLY | os.O_CREAT, mode=mode, dir_fd=dirfd)
```
yocalebo, line 30: `fd` is being leaked here?

yocalebo (changes requested):

src/middlewared/middlewared/etc_files/local/ssh/config.py
anodos325: No. open() builtin closes the fd passed into it when exiting the context manager.

yocalebo: That's nice. I didn't realize python implemented the `_at` syscalls 👍

src/middlewared/middlewared/etc_files/local/ssh/config.py
```diff
@@ -1,36 +1,84 @@
...
+                        middleware.logger.debug(
+                            "%s/%s: file has unexpected permissions [%s]. "
+                            "Changing to new value [%s].",
+                            SSH_CONFIG_PATH, k, stat.S_IMODE(st.st_mode), mode
+                        )
+                        os.fchmod(f.fileno(), mode)
+
+                    if st.st_uid != 0 or st.st_gid != 0:
+                        middleware.logger.debug(
+                            "%s/%s: unexpected owner [%d:%d]. "
```
yocalebo, line 52: Just nit-picking but I'd say something like `unexpected user or group ownership`

src/middlewared/middlewared/etc_files/local/ssh/config.py
```diff
@@ -1,36 +1,84 @@
...
+    expected_files = SSH_KEYS + DEFAULT_FILES
+    for f in os.listdir(dirfd):
+        if f not in expected_files:
+            st = os.stat(f, dir_fd=dirfd)
+            if stat.S_ISDIR(st.st_mode):
+                middleware.logger.debug("%s: removing unexpected directory.",
+                                        os.path.join(SSH_CONFIG_PATH, f))
+                shutil.rmtree(os.path.join(SSH_CONFIG_PATH, f))
+            else:
+                middleware.logger.debug("%s: removing unexpected directory.",
```
yocalebo, line 74: `: removing unexpected file`?

src/middlewared/middlewared/etc_files/local/ssh/sshd_config.mako
```diff
@@ -1,18 +1,27 @@
...
-		os.makedirs('/root/.ssh')
+	ssh_config = render_ctx['ssh.config']
+
+	os.makedirs('/root/.ssh', mode=0o700, exist_ok=True)
+	for p in ['/root/.ssh', '/root']:
+		st = os.stat(p)
+		if stat.S_IMODE(st.st_mode) != 0o700:
+			middleware.logger.debug("%s: adjusting permissions to 0o700", p)
+			os.chmod(p, 0o700)
+		if st.st_uid != 0 or st.st_gid != 0:
```
yocalebo, line 15: Are you meaning to log when changing ownership since you're logging when changing mode?

src/middlewared/middlewared/etc_files/local/ssh/config.py
```diff
@@ -1,36 +1,84 @@
...
+            with suppress(FileNotFoundError):
+                os.remove(os.path.join(SSH_CONFIG_PATH, k))
 
-    for f in [
-        'ssh_host_key', 'ssh_host_dsa_key', 'ssh_host_ecdsa_key', 'ssh_host_rsa_key', 'ssh_host_ed25519_key'
-    ]:
-        if os.path.exists(os.path.join(SSH_CONFIG_PATH, f)):
-            os.chmod(os.path.join(SSH_CONFIG_PATH, f), 0o600)
+    expected_files = SSH_KEYS + DEFAULT_FILES
+    for f in os.listdir(dirfd):
```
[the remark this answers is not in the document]
yocalebo: You'll want to test the code that I posted because `dirfd` is being passed to a context manager and I'm not sure if it closes it once it leaves scope so you might get `Bad File Descriptor` in the `def render` method. Not sure but just pointing it out before I forget.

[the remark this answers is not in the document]
anodos325: The context manager in this case doesn't close `dirfd`.

PR #8423, opened 2022-03-03
NAS-115126 / 22.12 / Avoid pwd and grp lookups in etc file generation

anodos325:
Where possible we should avoid using pwd and grp modules to look
up users / groups. Users can't change ids of builtins and so
we don't need to look this up every time we generate an etc file.
This is to prevent a significantly misbehaving directory service
(which can cause NSS lookups to hang longer than middleware timeout)
from adversely affecting etc file generation.

Hence we do the following here:
* allow "owner" and "group" in etc group config to be an int
  representing uid or gid
* ensure that we're running all permissions-related items in a
  separate thread by shifting this logic to a separate function

### DIFF
--- a/src/middlewared/middlewared/plugins/etc.py
+++ b/src/middlewared/middlewared/plugins/etc.py
@@ -12,6 +12,19 @@ import imp
 import os
 import pwd
 import stat
+import enum
+
+
+class EtcUSR(enum.IntEnum):
+    ROOT = 0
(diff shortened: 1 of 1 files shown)


yocalebo (changes requested):
Instead of hard-coding random integers, can we maybe add an int enum?

yocalebo (approved):

PR #8427, opened 2022-03-03
NAS-115001 / 22.12 / Correctly validate disk device path

Qubad786:


themylogin (changes requested):
This looks like a specific virtualization provider implementation detail. Why can't we just add this `+` when building the XML for libvirtd?

sonicaj (approved):

sonicaj:
@Qubad786 you have flake8 errors

src/middlewared/middlewared/plugins/vm/vm_devices.py
```diff
@@ -48,6 +49,21 @@ class Config:
...
         cli_namespace = 'service.vm.device'
 
+    @accepts()
+    @returns(Dict(additional_attrs=True, example={'vms/test 1': '/dev/zvol/vms/test+1'}))
+    async def disk_choices(self):
+        """
+        Returns disk choices for device type "DISK".
+        """
+        return {
+            vol['id']: os.path.join('/dev/zvol', vol['id'].replace(' ', '+'))
```
themylogin, line 59: Shouldn't keys be the encoded value (with `+`) and values be human-readable value? Also let's use `zvol_name_to_path` here.

themylogin (changes requested):

PR #8429, opened 2022-03-03
NAS-115049 / 22.12 / Remove typo from nfs-common config generation

anodos325:
The "-N 2" option was copied over from the nfs-kernel-server mako
file where it disabled the NFSv2 protocol. In this case, '-N'
option actually prevents rpc.statd from starting.


anodos325:
This PR also removes unused nfsd.py file (FreeBSD-specific).

yocalebo (approved):

rick-mesta (approved):

PR #8432, opened 2022-03-04
NAS-114980 / 22.12 / switch to using mpm_event from mpm_prefork

anodos325:
Convert to using the default mpm for webdav. mpm_prefork was
crashing on a NULL dereference and is most likely not a great
choice for our purposes..

Since this change means that apache server is no longer setuid,
ensure that apache is running as webdav rather than www-data
so that users don't lose access to files written by webdav.


themylogin (approved):

anodos325:
backport

PR #8440, opened 2022-03-04
NAS-115135 / 22.12 / Improve webdav validation and disable deprecated TLS versions

anodos325:
Remove support for TLSv1, TLSv1.1 and add support for TLSv1.3
Add webdav.cert_choices method

### DIFF
--- a/src/middlewared/middlewared/plugins/webdav.py
+++ b/src/middlewared/middlewared/plugins/webdav.py
@@ -2,8 +2,9 @@ import asyncio
 import os
 
 from middlewared.common.attachment import LockableFSAttachmentDelegate
-from middlewared.schema import accepts, Bool, Dict, Int, Patch, Str, ValidationErrors
+from middlewared.schema import accepts, returns, Bool, Dict, Int, Patch, Str, ValidationErrors
 from middlewared.service import SharingService, SystemServiceService, private
+from middlewared.plugins.etc import EtcUSR, EtcGRP
 import middlewared.sqlalchemy as sa
 
 
@@ -87,8 +88,8 @@ class WebDAVSharingService(SharingService):
         if data['perm']:
             await self.middleware.call('filesystem.chown', {
                 'path': data['path'],
-                'uid': (await self.middleware.call('dscache.get_uncached_user', 'webdav'))['pw_uid'],
-                'gid': (await self.middleware.call('dscache.get_uncached_group', 'webdav'))['gr_gid'],
+                'uid': EtcUSR.WEBDAV.value,
+                'gid': EtcGRP.WEBDAV.value,
                 'options': {'recursive': True}
             })
 
@@ -127,8 +128,8 @@ class WebDAVSharingService(SharingService):
         if not old['perm'] and new['perm']:
             await self.middleware.call('filesystem.chown', {
                 'path': new['path'],
-                'uid': (await self.middleware.call('dscache.get_uncached_user', 'webdav'))['pw_uid'],
-                'gid': (await self.middleware.call('dscache.get_uncached_group', 'webdav'))['gr_gid'],
+                'uid': EtcUSR.WEBDAV.value,
+                'gid': EtcGRP.WEBDAV.value,
                 'options': {'recursive': True}
             })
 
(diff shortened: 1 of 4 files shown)


yocalebo (approved):

src/middlewared/middlewared/plugins/webdav.py
```diff
@@ -208,6 +209,14 @@ async def do_update(self, data):
 
         return await self.config()
 
+    @private
+    async def cert_choices(self):
+        webui_cert = await self.middleware.call('system.general.config')
+
+        filter = [["id", "!=", webui_cert['id']]] if webui_cert else []
```
rick-mesta, line 216: so, in case of where webdav is using the same certs as the webserver, you still pass an empty `filter` list over to `certificate.query`, right ? no ill side effects from that ?

sonicaj: We are querying the config and getting it's id and not getting the certificate's id

sonicaj (changes requested):

src/middlewared/middlewared/plugins/webdav.py
```diff
@@ -244,13 +253,19 @@ async def validate(self, data, schema_name):
                     'WebDAV SSL protocol specified without choosing a certificate'
                 )
             else:
+                cert_choices = await self.cert_choices()
+                cert_in_choices = any(filter(lambda x: x['id'] == cert_ssl, cert_choices))
+                if not cert_in_choices:
+                    verrors.add(
+                        f'{schema_name}.certssl',
+                        f'Certificate is not permitted. Valid choices are {[x["name"] for x in cert_choices]}'
```
sonicaj, line 261: It's better to do `", ".join(x["name"] for x in cert_choices)`

sonicaj: Btw is this even required ? The method below handles the validation already ?

sonicaj: I am not sure why using the same cert as the UI one is a problem 🤔 ?

src/middlewared/middlewared/plugins/webdav.py
anodos325: Yeah, whoops about the wrong id. :D

src/middlewared/middlewared/plugins/webdav.py
anodos325: It's a matter of web browser behavior. Firefox would not allow me to access our webui after accessing the webdav server first.

anodos325: Hmm... and now I'm not seeing the SSL error anymore. I can remove that part of validation then.

sonicaj: Can you share the error you got when this happened please ? I think it shouldn't be an issue

sonicaj (commented):
Let's please still have the certificate choices method still so that UI can consume that to show cert choices ( not sure if it's using certificate.query directly and probably not filtering out CSRs )

src/middlewared/middlewared/plugins/webdav.py
anodos325: SEC_ERROR_REUSED_ISSUER_AND_SERIAL

src/middlewared/middlewared/plugins/webdav.py
```diff
@@ -226,6 +227,15 @@ async def upper(self, data):
 
         return data
 
+    @private
```
sonicaj, line 230: This would be public/user facing ;)

src/middlewared/middlewared/plugins/webdav.py
```diff
@@ -226,6 +227,15 @@ async def upper(self, data):
 
         return data
 
+    @accepts()
```
sonicaj, line 230: We are missing `returns` decorator and a doc string as this is a public method 😅

src/middlewared/middlewared/plugins/webdav.py
anodos325: Okay. That was just an issue with a janky default cert on my server.

src/middlewared/middlewared/plugins/webdav.py
anodos325: Okay.

src/middlewared/middlewared/plugins/webdav.py
anodos325: Meh

sonicaj:
flake8 issues with missing import bwt :p

PR #8443, opened 2022-03-04
NAS-115001 / 22.02.1 / Correct vm disks path on SCALE

Qubad786:
Volumes are mounted as `/dev/zvol/vms/test 1` in freebsd but as `/dev/zvol/vms/test+1` in linux.

This PR adds:
1. An API to provide list of disks with correct zvol paths so UI can consume it as is unlike today (i.e. UI pre-fixes the `/dev/zvol` into volume name).
2. Add a migration for correcting existing vm disk devices for users who would be upgrading from CORE

PS: I have confirmed this behavior with @freqlabs on freebsd vs linux

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/vm_devices.py
+++ b/src/middlewared/middlewared/plugins/vm/vm_devices.py
@@ -49,7 +49,7 @@ class VMDeviceService(CRUDService):
         cli_namespace = 'service.vm.device'
 
     @accepts()
-    @returns(Dict())
+    @returns(Dict(additional_attrs=True, example={'vms/test 1': '/dev/zvol/vms/test+1'}))
     async def disk_choices(self):
         """
         Returns disk choices for device type "DISK".
@@ -57,7 +57,9 @@ class VMDeviceService(CRUDService):
         return {
             vol['id']: os.path.join('/dev/zvol', vol['id'].replace(' ', '+'))
             for vol in await self.middleware.call(
-                'pool.dataset.query', [['type', '=', 'VOLUME']], {'extra': {'properties': []}}
+                'pool.dataset.query', [['type', '=', 'VOLUME'], ['locked', '=', False]], {
+                    'extra': {'properties': ['encryption', 'keystatus']}
+                }
             )
         }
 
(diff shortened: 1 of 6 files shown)


freqlabs:
For context, this commit explains the udev rule: https://github.com/openzfs/zfs/commit/5f91bd3dea49a529e87e0aa39595f074fd09736a

src/middlewared/middlewared/plugins/vm/vm_devices.py
```diff
@@ -48,6 +48,19 @@ class Config:
         datastore_extend = 'vm.device.extend_device'
         cli_namespace = 'service.vm.device'
 
+    @accepts()
+    @returns(Dict())
```
sonicaj, line 52: Let's please add an `example` dict here

src/middlewared/middlewared/plugins/vm/vm_devices.py
```diff
@@ -48,6 +48,19 @@ class Config:
...
+    async def disk_choices(self):
+        """
+        Returns disk choices for device type "DISK".
+        """
+        return {
+            vol['id']: os.path.join('/dev/zvol', vol['id'].replace(' ', '+'))
+            for vol in await self.middleware.call(
+                'pool.dataset.query', [['type', '=', 'VOLUME']], {'extra': {'properties': []}}
```
sonicaj, line 60: Let's please also check if the volume is locked or not as we shouldn't be showing locked volume as a valid choice

themylogin (changes requested):
This looks like a specific virtualization provider implementation detail. Why can't we just add this `+` when building the XML for libvirtd?

Qubad786:
@themylogin its possible but in my opinion, we should make existing paths normalized in database and consume them dorectly instead of looking to add + everytime we get it from db.

Qubad786:
Additionally, WebUI currently make this path on its side with space, I've delegated that to the middleware with choices API so it stays consistent.

themylogin:
@Qubad786 do you mean that `/dev/zvol/vms/test+1` is a normalized path, but `/dev/zvol/vms/test 1` is not?

Qubad786:
`/dev/zvol/vms/test 1` is not correct path as zfs is mounting as `/dev/zvol/vms/test+1` in linux

Qubad786:
Ah, thanks for pulling out, I'll address it!

themylogin:
@Qubad786 please take note of new utils implemented here https://github.com/truenas/middleware/pull/8452

themylogin (approved):

rick-mesta (approved):

PR #8445, opened 2022-03-04
NAS-115130 / 22.12 / More logging to debug rare `disk.sync_zfs_guid` issue

themylogin:
Sometimes disks are not mapped with their vdev guids. This happens randomly and rarely
(usually users are able to resolve this issue just by rebooting the system). I failed to
repeat this. Let's have better debugging for the future reports of this issue.

I decided to drop those awk scripts that provide "normalized" data, raw zpool output is
perfectly readable and it's much reliable to have unprocessed data in the debug.


themylogin:
time 3:00

anodos325 (approved):

PR #8451, opened 2022-03-07
NAS-114213 / 22.12 / Delete certificates that no longer exist

Qubad786:
Remove certificate files from the filesystem once they're removed from database.

### DIFF
--- a/src/middlewared/middlewared/etc_files/generate_ssl_certs.py
+++ b/src/middlewared/middlewared/etc_files/generate_ssl_certs.py
@@ -1,22 +1,28 @@
+import itertools
 import os
 import shutil
 import subprocess
 
-from middlewared.service import CallError
+from middlewared.main import Middleware
+from middlewared.service import CallError, Service
 
 
-def write_certificates(certs, cacerts):
+def write_certificates(certs: list, cacerts: list) -> set:
+    expected_files = set()
     for cert in certs:
         if cert['chain_list']:
+            expected_files.add(cert['certificate_path'])
             with open(cert['certificate_path'], 'w') as f:
                 f.write('\n'.join(cert['chain_list']))
 
         if cert['privatekey']:
+            expected_files.add(cert['privatekey_path'])
             with open(cert['privatekey_path'], 'w') as f:
                 os.fchmod(f.fileno(), 0o400)
                 f.write(cert['privatekey'])
 
         if cert['type'] & 0x20 and cert['CSR']:
+            expected_files.add(cert['csr_path'])
             with open(cert['csr_path'], 'w') as f:
                 f.write(cert['CSR'])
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/etc_files/generate_ssl_certs.py
```diff
@@ -1,23 +1,27 @@
...
 import subprocess
 
-from middlewared.service import CallError
-from middlewared.utils import osc
+from middlewared.main import Middleware
+from middlewared.service import CallError, Service
 
 
-def write_certificates(certs, cacerts):
+def write_certificates(certs: list, cacerts: list, expected_files: set) -> None:
```
themylogin, line 10: You can return a set of written files instead of requiring extra arguments

src/middlewared/middlewared/etc_files/generate_ssl_certs.py
```diff
@@ -68,18 +67,26 @@ def write_crls(cas, middleware):
...
+    write_crls(cas, middleware, expected_files)
 
-    write_crls(cas, middleware)
+    # We would like to remove certificates which have been deleted
+    found_files = set(itertools.chain(
+        map(lambda f: '/etc/certificates/' + f, os.listdir('/etc/certificates')),
+        map(lambda f: '/etc/certificates/CA/' + f, os.listdir('/etc/certificates/CA'))
+    ))
+    for to_remove in found_files - expected_files:
+        os.unlink(to_remove)
```
themylogin, line 92: How does this handle non-empty directories?

themylogin (changes requested):
That's good. However, can we please add some integration tests?

tests/api2/test_certs.py
```diff
@@ -264,3 +264,40 @@ def test_revoking_ca():
             assert set(c['certificate'] for c in root_ca['revoked_certs']) == check_set, root_ca
         finally:
             call('certificate.delete', cert['id'], job=True)
+
+
+def test_certs_existence_on_filesystem():
```
themylogin, line 269: What does this test actually test?

tests/api2/test_certs.py
```diff
@@ -264,3 +264,40 @@ def test_revoking_ca():
...
+                        expected_files.add(cert['privatekey_path'])
+                    if cert['cert_type_CSR']:
+                        expected_files.add(cert['csr_path'])
+                    if any(cert[k] for k in ('CA_type_existing', 'CA_type_internal', 'CA_type_intermediate')):
+                        expected_files.add(cert['crl_path'])
+                found_files = {
+                    f['path']
+                    for p in ('/etc/certificates', '/etc/certificates/CA') for f in call('filesystem.listdir', p)
+                }
+                assert found_files == expected_files, f'Expected {expected_files!r} but found {found_files!r}'
```
themylogin, line 301: pytest assertion error will dump the set difference, no need to add custom message

themylogin (changes requested):

tests/api2/test_certs.py
```diff
@@ -264,3 +264,40 @@ def test_revoking_ca():
...
+
+
+def test_deleted_certs_dont_exist_on_filesystem():
+    with intermediate_certificate_authority('root_ca', 'intermediate_ca') as (root_ca, intermediate_ca):
+        with certificate_signing_request('csr_test') as csr:
+            with intermediate_certificate_authority('root_ca2', 'intermediate_ca2') as (root_ca2, intermediate_ca2):
+                # no-op
+                pass
+            with certificate_signing_request('csr_test2') as csr2:
+                pass
```
themylogin, line 276: Sorry, I still don't really understand what we are testing here. Do we create `'root_ca2'`, `'intermediate_ca2'`, `'csr_test2'`, immediately delete them and assert that corresponding files do not exist? What does the code below test then?

Qubad786: Yes for the first one @themylogin. 
Regarding the code below, I am trying to just mimic the behavior where we have few certificates in both db and disk just an extra check to verify that we've them in place as required.

themylogin: Please move the code below to a separate test then. One test should test one thing (otherwise if, for example, the first part fails, we'll never know the result of the second one).

themylogin (approved):
Thank you!

PR #8454, opened 2022-03-07
NAS-114409 / 22.12 / Allow deleting unused images on app deletion

Qubad786:
# Background

An app's docker image is not removed when it's deleted through `helm uninstall` cmd. This PR:

1. Adds an option to delete its images while app's uninstallation
2. Adds APIs: Get chart releases that are consuming image (to be deleted), Find and delete images.
3. Handle tags vs digests during the picking of image to be removed
4. Add tests

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/images.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/images.py
@@ -1,0 +1,51 @@
+import errno
+
+from middlewared.plugins.docker_linux.utils import normalize_reference, get_chart_releases_consuming_image
+from middlewared.schema import Dict, Str, returns
+from middlewared.service import accepts, CallError, private, Service
+
+
+class ChartReleaseService(Service):
+
+    class Config:
+        namespace = 'chart.release'
+
+    @accepts(Str('chart_release_name'))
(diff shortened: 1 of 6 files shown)


src/middlewared/middlewared/plugins/chart_releases_linux/images.py
```diff
@@ -0,0 +1,55 @@
...
+                await self.middleware.call('container.image.delete', image['id'])
+            except Exception as e:
+                failed_to_delete[', '.join(image['complete_tags'])] = str(e)
+        return failed_to_delete
+
+    @private
+    async def get_to_delete_unused_app_images(self, chart_release):
+        chart_release = await self.middleware.call(
+            'chart.release.get_instance', chart_release, {'extra': {'retrieve_resources': True}}
+        ) if isinstance(chart_release, str) else chart_release
```
sonicaj, line 45: Let's be consistent please, i suppose this conditional is for debug ?

sonicaj (commented):
LGTM, one minor nit

sonicaj (commented):
@Qubad786 there are flake8 errors. Please fix which are applicable and while we are at it, can you please suppress `I101` error ? It seems too vague to have enforced. Thanks!

sonicaj (approved):

PR #8467, opened 2022-03-08
NAS-115201 / 22.12 / Proxy VM Display devices via nginx

sonicaj:
This PR adds changes to reverse proxy vm display devices via nginx as we were seeing users who had enabled https redirect which resulted in HSTS header being set and the vm display device URI opening up with https but our implementation of websockify not supporting it.
After discussing with William/Vlad, we decided it was best to proxy vm display devices via nginx and reduce use cases where we have portals opening up from our UI but not supporting HTTPS.


src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
```diff
@@ -81,20 +81,8 @@
 #
 #    FreeNAS nginx configuration file
```
yocalebo, line 82: Might as well remove mention of `FreeNAS`

src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
```diff
@@ -187,6 +175,19 @@ http {
             rewrite ^.* $scheme://$http_host/ui/ redirect;
         }
 
+% for dev in middleware.call_sync('vm.get_running_display_devices'):
+        location ~ ${dev["path"]}(.*)$ {
```
themylogin, line 179: Is using regex matching absolutely necessary here? Can it be just normal location matching + a `rewrite` directive?

src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
```diff
@@ -187,6 +175,19 @@ http {
...
+            proxy_pass ${dev["redirect_uri"]}/$1?$args;
+            proxy_http_version 1.1;
+            proxy_set_header X-Real-Remote-Addr $remote_addr;
```
themylogin, line 182: Are all these headers required? My biggest doubt is `Host`.

william-gr (commented):
Doesnt this mean you have to reload nginx every time you start/stop a vm?

src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
```diff
@@ -187,6 +176,19 @@ http {
             rewrite ^.* $scheme://$http_host/ui/ redirect;
         }
 
+% if middleware.call_sync('vm.generate_nginx_route'):
```
william-gr, line 179: Why have this conditional? Could have this always enabled and then not need to reload nginx?
I dont think there is any harm

src/middlewared/middlewared/plugins/vm/lifecycle.py
```diff
@@ -64,6 +64,8 @@ def initialize_vms(self, timeout=30):
                     self.middleware.logger.error(
                         'Unable to setup %r VM object: %s', vm_data['name'], str(e), exc_info=True
                     )
+            self.middleware.call_sync('service.reload', 'http')
```
themylogin, line 67: What's the point of adding haproxy if we keep reloading nginx?

sonicaj: We were only doing it once ideally in the middleware lifecycle. I can remove this and always have the vm endpoint in nginx.

themylogin (approved):
That's great, thank you!

PR #8468, opened 2022-03-08
NAS-115171 / 22.12 / Correctly specify type of raw/disk device in libvirt xml

sonicaj:


src/middlewared/middlewared/plugins/vm/devices/storage_devices.py
```diff
@@ -9,6 +9,8 @@
 
 class StorageDevice(Device):
 
+    TYPE = NotImplementedError
```
themylogin, line 12: Just `NotImplemented`

themylogin (approved):

PR #8470, opened 2022-03-08
document return types for gluster plugins

yocalebo:


rick-mesta (approved):

src/middlewared/middlewared/plugins/gluster_linux/peer.py
```diff
@@ -21,6 +22,15 @@ class Config:
         cli_namespace = 'service.gluster.peer'
 
     @filterable
+    @filterable_returns(Dict(
```
sonicaj, line 25: Why are we duplicating this and not using `ENTRY` ?

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -70,6 +94,7 @@ async def common_validation(self, data, schema_name):
         Int('redundancy'),
         Bool('force'),
     ))
+    @returns(List('volumes', items=[Dict('volume', additional_attrs=True)]))
```
sonicaj, line 97: Also i assume the volume is same as above - just specifying ENTRY is enough and all of these usages are covered automatically

sonicaj:
Btw did we test these changes ? We have middleware not starting

PR #8480, opened 2022-03-08
NAS-115211 / 22.12 / Remove legacy LDAP truenas_cacerts.pem file

anodos325:
For historical reasons we had single cacert file with manually
concatenated cacerts. Convert LDAP service to use newer
general-purpose one. Also use existing fd for fchmod call.


anodos325:
Need to test / verify that openldap is still happy with this cert file.

sonicaj (approved):

PR #8486, opened 2022-03-09
use clusterjob.submit in gluster.volume.start/stop

yocalebo:


src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -154,10 +154,9 @@ async def start(self, data):
         options = {'args': (name,), 'kwargs': data}
         result = await self.middleware.call('gluster.method.run', volume.start, options)
 
-        # this will send a request to all peers
-        # in the TSP to FUSE mount this volume locally
-        data = {'event': 'VOLUME_START', 'name': name, 'forward': True}
-        await self.middleware.call('gluster.localevents.send', data)
+        # this will send a request to all peers in the TSP to FUSE mount this volume locally
+        mnt_job = await self.middleware.call('clusterjob.submit', 'gluster.fuse.mount', {'name': name})
+        await mnt_job.wait()
```
sonicaj, line 159: Do we want to check if the job failed ?

sonicaj (commented):
start/stop methods themselves are not jobs but are waiting for jobs inside - could we face timeout in these situations ?

src/middlewared/middlewared/plugins/gluster_linux/volume.py
yocalebo: No longer doing this

src/middlewared/middlewared/plugins/gluster_linux/volume.py
```diff
@@ -192,15 +192,10 @@ async def stop(self, data):
...
 
-        # this will send a request to all peers
-        # in the TSP to unmount the FUSE mountpoint
-        data = {'event': 'VOLUME_STOP', 'name': name, 'forward': True}
-        await self.middleware.call('gluster.localevents.send', data)
-
-        return result
+        # this will send a request to all peers in the TSP to unmount the FUSE mountpoint
+        await self.middleware.call('gluster.localevents.send', {'event': 'VOLUME_STOP', 'name': name, 'forward': True})
+        return await self.middleware.call('gluster.method.run', volume.stop, {'args': (name,), 'kwargs': data})
```
sonicaj, line 198: What does `gluster.method.run` return ? Right now `returns` decorator states that nothing would be returned from this endpoint

yocalebo: It returns nothing

sonicaj (approved):

PR #8496, opened 2022-03-09
NAS-115101 / 22.12 / Enable nscd for hosts

anodos325:
There are some applications with borderline pathlogocial behavior
regarding lookups of hostnames. Enable nscd by default for
hosts, but leave disabled for passwd and group so that its caching
doesn't interfere with winbindd (the duplicate layers of caching
result in undefined behavior).

Actual nscd config is heavily based on default debian nscd config
with the following changes:
- disabling undesired components
- enabling paranoia mode (restarting the nscd service every 60 min

### DIFF
--- a/src/middlewared/middlewared/etc_files/nscd.conf.mako
+++ b/src/middlewared/middlewared/etc_files/nscd.conf.mako
@@ -1,0 +1,101 @@
+#
+# /etc/nscd.conf
+#
+# An example Name Service Cache config file.  This file is needed by nscd.
+#
+# WARNING: Running nscd with a secondary caching service like sssd may lead to
+#          unexpected behaviour, especially with how long entries are cached.
+#
+# Legal entries are:
+#
+#	logfile			<file>
+#	debug-level		<level>
+#	threads			<initial #threads to use>
+#	max-threads		<maximum #threads to use>
+#	server-user             <user to run server as instead of root>
+#		server-user is ignored if nscd is started with -S parameters
+#       stat-user               <user who is allowed to request statistics>
+#	reload-count		unlimited|<number>
+#	paranoia		<yes|no>
+#	restart-interval	<time in seconds>
+#
+#       enable-cache		<service> <yes|no>
+#	positive-time-to-live	<service> <time in seconds>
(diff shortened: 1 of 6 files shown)


src/middlewared/middlewared/plugins/service_/services/pseudo/misc.py
```diff
@@ -247,12 +247,7 @@ class DSCacheService(PseudoServiceBase):
     name = "dscache"
 
     async def start(self):
-        ldap_enabled = (await self.middleware.call('ldap.config'))['enable']
-        if ldap_enabled:
-            await systemd_unit("nscd", "restart")
-        else:
-            await systemd_unit("nscd", "stop")
-
+        await systemd_unit("nscd", "start")
```
themylogin, line 250: Shouldn'it it be just always started now?

src/middlewared/middlewared/etc_files/nscd.conf.mako
```diff
@@ -0,0 +1,101 @@
...
+#	auto-propagate		<service> <yes|no>
+#
+# Currently supported cache names (services): passwd, group, hosts, services
+#
+<%
+    import os
+    from contextlib import suppress
+
+    with suppress(FileExistsError):
+        os.mkdir("/var/nscd", mode=0o755)
```
themylogin, line 42: `exist_ok=True`

themylogin: And what is this directory anyway?

anodos325: I thought `exist_ok=True` is keyword argument for `os.makedirs` rather than `os.mkdir`. Path was wrong, should have been /var/run/nscd (it's where nscd domain socket is located). nscd service will fail to start if for some reason path doesn't exist.

src/middlewared/middlewared/plugins/service_/services/pseudo/misc.py
anodos325: Good point.

yocalebo (approved):

PR #8497, opened 2022-03-10
NAS-115216 / 22.12 / Allow normalizing jobs results

sonicaj:
This PR adds changes to normalize jobs results to prevent leaking sensitive information like dataset keys in the debug.

### DIFF
--- a/src/freenas/usr/local/libexec/freenas-debug/system/system.sh
+++ b/src/freenas/usr/local/libexec/freenas-debug/system/system.sh
@@ -148,7 +148,7 @@ system_func()
 	section_footer
 
 	section_header "Middleware Jobs - 'midclt call core.get_jobs'"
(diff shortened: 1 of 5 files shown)


src/freenas/usr/local/libexec/freenas-debug/system/system.sh
```diff
@@ -148,7 +148,7 @@ system_func()
 	section_footer
 
 	section_header "Middleware Jobs - 'midclt call core.get_jobs'"
-	midclt call core.get_jobs '[["state", "!=", "SUCCESS"]]' | jq .
+	midclt call core.get_jobs '[["state", "!=", "SUCCESS"]]' '{"extra": {"normalize_result": true}}' | jq .
```
themylogin, line 151: What does `normalize` here mean? How is `"********"` instead of an e.g. password "normal"?

Can we use a clearer name please?

themylogin (approved):

PR #8499, opened 2022-03-10
NAS-115198 / 13.0 / Reuse fd in etc file generation and run unlink in thread (#8466)

anodos325:
This contains a few improvements to permissions management in
etc plugin:

- write_if_changed is updated to allow passing an fd rather than
  only a path name
- permissions check and file contents check are now run from same
  thread.
- permissions are set on file before writing.
- set default permissions on config files to 0644 unless dev
  specifies something different.


anodos325:
FreeBSD doesn't have same /proc behavior, but does have O_EMPTY_PATH flag that allows basically re-opening an fd with different flags.

rick-mesta (approved):

PR #8503, opened 2022-03-10
NAS-115232 / 22.12 / Improve net global

yocalebo:
Many problems fixed:

1. `/etc/hostname` generation should only contain the hostname (not the dns domain name according to man pages)
2. it was taking 7 seconds to change any single item in the `network.globalconfiguration` plugin
3. we were restarting the various service announcement daemons twice (or more) without actually checking what changed
4. we were restarting `collectd` twice or more
5. if you simply click "save" in the webUI (without changing anything) it will still run through this logic which is pointless

With these changes.
1. `/etc/hostname` is properly generated (without the dns domain name)
2. in my testing, changes now take 0.5 to 3 seconds total (instead of 7+)
3. only toggle the service announcement daemons once (if appropriate)
4. only restart collectd once (if appropriate)
5. if nothing changed, then return early

### DIFF
--- a/src/middlewared/middlewared/etc_files/hostname.py
+++ b/src/middlewared/middlewared/etc_files/hostname.py
@@ -4,12 +4,8 @@ from middlewared.service import CallError
 
 
 def render(service, middleware):
-    config = middleware.call_sync("network.configuration.config")
-    hostname = config['hostname_local']
-    if config['domain']:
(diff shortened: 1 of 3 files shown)


anodos325 (approved):

yocalebo:
backport

PR #8505, opened 2022-03-10
NAS-115235 / 22.12 / Restrict job log dir to root

anodos325:
Since cron jobs can contain arbitrary output (user scripts), protect job log dir from being world-readable.

### DIFF
--- a/src/middlewared/middlewared/job.py
+++ b/src/middlewared/middlewared/job.py
@@ -577,6 +577,7 @@ class Job:
         if self.logs_path is not None:
             fd = self.logs_fd
             os.makedirs(LOGS_DIR, exist_ok=True)
+            os.chmod(LOGS_DIR, 0o700)
             self.logs_fd = open(self.logs_path, 'ab', buffering=0)
             if fd is not None:
                 fd.close()


src/middlewared/middlewared/job.py
```diff
@@ -577,6 +577,7 @@ def start_logging(self):
         if self.logs_path is not None:
             fd = self.logs_fd
             os.makedirs(LOGS_DIR, exist_ok=True)
+            os.chmod(LOGS_DIR, 0o700)
```
themylogin, line 580: `os.makedirs` has `mode` argument.

themylogin (commented):
I don't really understand the relation between those two. Who are we protecting logs from?

src/middlewared/middlewared/job.py
anodos325: Yeah, but that only sets mode when dir is freshly created. It doesn't reset mode.

themylogin: @anodos325 what is the case when it's not freshly created? It is wiped on middleware startup.

william-gr (approved):

PR #8508, opened 2022-03-10
NAS-115237 / 22.12 / Add better handling for NFS hosts that don't resolve

anodos325:
Raise a ValidationError if a hostname doesn't resolve during
share creation or update.

Omit export line when generating exports and generate an alert
if hostname cannot be resolved.

### DIFF
--- a/src/middlewared/middlewared/alert/source/nfs_host.py
+++ b/src/middlewared/middlewared/alert/source/nfs_host.py
@@ -1,0 +1,11 @@
+from middlewared.alert.base import AlertCategory, AlertClass, AlertLevel, SimpleOneShotAlertClass
+
+
+class NFSHostnameLookupFailAlertClass(AlertClass, SimpleOneShotAlertClass):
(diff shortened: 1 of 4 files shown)


yocalebo (approved):

themylogin (approved):

themylogin (commented):
IIRC `etc.generate` is called early, before the network is set up, so we won't be able to resolve anything at this point and an alert will be created. Should we add `'checkpoint': 'interface_sync'` to `exports`?

anodos325:
> IIRC `etc.generate` is called early, before the network is set up, so we won't be able to resolve anything at this point and an alert will be created. Should we add `'checkpoint': 'interface_sync'` to `exports`?

D'oh. Good catch. Yeah, I think we probably need to do that, but I will have to verify that it doesn't break nfsd startup on boot (nfsd will refuse to start if exports is missing). I'm off tomorrow and so won't get around to investigating / testing until Monday.

PR #8514, opened 2022-03-11
Retrieve complete list of valid timezones

Qubad786:
### Background
In TrueNAS SCALE, some timezone-related files are symlinked, so these are missed by `find` command when used without follow argument.

### Solution
Updating `find` command arguments with follow.


themylogin (changes requested):
Why do we spawn a subprocess instead of using `os.walk`?

While we're here, let's add a test please.

themylogin (approved):

PR #8534, opened 2022-03-14
NAS-115268 / 22.12 / Fix id used for user-related config file regression tests

anodos325:
The wrong user db id was being passed to the test.


ericbsd (approved):
LGTM

PR #8535, opened 2022-03-14
NAS-115185 / 22.12 / Get correct stats path for ups service when configured in slave mode

sonicaj:
For a user we are seeing where when ups is configured in slave mode, we don't have his stats living at '/var/db/collectd/rrd/remotehost_ip_here` and instead we have a directory named after master's driver.


themylogin (approved):
What if we change UPS driver

themylogin (approved):
Even better, thank you!

PR #8554, opened 2022-03-15
NAS-115295 / 22.12 / add system.product_type API test

yocalebo:
`system.is_freenas` is deprecated and will eventually be removed. Furthermore, `system.is_freenas` is kind of tacky since we changed the name quite some time ago.

Check `system.product_type` (which is the replacement for `system.is_freenas`) instead.


ericbsd (approved):
LGTM

rick-mesta (approved):

PR #8563, opened 2022-03-16
NAS-115310 / 22.12 / simplify and fix test_003_network.py

yocalebo:
Many fixes, the primary one being we actually test the results on an HA system instead of only on a non-HA system.


ericbsd (approved):
LGTM

yocalebo:
backport

PR #8573, opened 2022-03-17
NAS-115348 / 22.12 / improve loggin in device_info_linux.py

yocalebo:
Some debugs I've come across are filled with errors that don't help understand where the error is occurring.
```
[2022/02/24 14:56:15] (DEBUG) middlewared.get_disks():50 - Failed to retrieve disk details for sda : unsupported operand type(s) for *: 'int' and 'NoneType'
```

3 fixes while I'm here:
1. `self.middleware.logger` to `self.logger`
2. use `exc_info=True`
3. flake8 and minor clean-up

### DIFF
--- a/src/middlewared/middlewared/plugins/device_/device_info_linux.py
+++ b/src/middlewared/middlewared/plugins/device_/device_info_linux.py
@@ -31,12 +31,12 @@ class DeviceService(Service, DeviceInfoBase):
         disks = {}
         disks_data = self.retrieve_disks_data()
 
-        for block_device in pyudev.Context().list_devices(subsystem='block', DEVTYPE='disk'):
-            if block_device.sys_name.startswith(('sr', 'md', 'dm-', 'loop', 'zd')):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -32,24 +31,22 @@ def get_disks(self):
...
 
             try:
-                disks[block_device.sys_name] = self.get_disk_details(block_device, self.disk_default.copy(), disks_data)
-            except Exception as e:
-                self.middleware.logger.debug(
-                    'Failed to retrieve disk details for %s : %s', block_device.sys_name, str(e)
-                )
+                disks[dev.sys_name] = self.get_disk_details(dev, self.disk_default.copy(), disks_data)
+            except Exception:
+                self.logger.debug('Failed to retrieve disk details for %s : %s', dev.sys_name, exc_info=True)
```
sonicaj, line 49: If we have already logged it once, we shouldn't log it again. There are lots of debugs where we have repeated entires of the same log again and again.

PR #8578, opened 2022-03-18
clean up device_ plugins

yocalebo:
Also, copy and pasted the `serial_port_choices` method in `osc.linux.system` to `middlewared.utils.serial_devices` so we can remove the `osc` plugin in a few more files.


src/middlewared/middlewared/plugins/device_/device_info.py
```diff
@@ -6,27 +6,46 @@
...
+        'model': None,
+        'type': 'UNKNOWN',
+        'bus': 'UNKNOWN',
+        'serial': '',
+        'size': None,
+        'serial_lunid': None,
+        'blocks': None,
+    }
 
+    @private
```
themylogin, line 44: Should we just make the whole service private?

themylogin (approved):
Cool!

IMO `serial` is a more pythonic module name then `serial_devices`

PR #8579, opened 2022-03-18
NAS-115355 / 22.12 / fix test_400_service.py

yocalebo:
1. The test results printed the service id (an integer) instead of the service name.  Now it prints service name
(i.e. `api2/test_400_service.py::test_02_get_service_info_for[4] PASSED` to `api2/test_400_service.py::test_01_enable_service[cifs] PASSED`)
2. It chose (randomly) 4 services to actually start which doesn't make much sense so I removed that because
`cifs` is handled in the `test_420_smb.py` script
`iscsitarget` is handled in the `test_260_iscsi.py` script
`snmp` is handled in the `test_440_snmp.py` script
`rsync` is handled in the `test_380_rsync.py` script
3. dramatically simplify the file and remove unnecessary code
4. rename this to `test_400_enable_disable_services.py` since that's what it's actually doing

### DIFF
--- a/tests/api2/test_400_enable_disable_services.py
+++ b/tests/api2/test_400_enable_disable_services.py
@@ -17,31 +17,21 @@ service_names = list(all_services.keys())
 
 @pytest.mark.dependency(name='ENABLE')
 @pytest.mark.parametrize('svc', service_names)
-def test_01_enable_service(svc):
+def test_01_enable_and_verify_service(svc):
     results = PUT(f'/service/id/{all_services[svc]["id"]}/', {'enable': True})
     assert results.status_code == 200, results.text
 
-
-@pytest.mark.dependency(name='VERIFY_ENABLE')
-@pytest.mark.parametrize('svc', service_names)
-def test_02_verify_service_was_enabled(svc, request):
-    depends(request, ['ENABLE'])
     results = GET(f'/service/id/{all_services[svc]["id"]}/')
     assert results.status_code == 200, results.text
     assert results.json()['enable'], results.text
 
 
-@pytest.mark.dependency(name='DISABLE')
 @pytest.mark.parametrize('svc', service_names)
-def test_03_disable_service(svc, request):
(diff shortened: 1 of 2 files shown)


themylogin:
So this module does not test services being able to start anymore?

yocalebo:
> So this module does not test services being able to start anymore?

No it was randomly choosing 4 services that really didn't test anything. The other integration tests for those services start/stop the service appropriately.

themylogin:
> > So this module does not test services being able to start anymore?
> 
> No it was randomly choosing 4 services that really didn't test anything. The other integration tests for those services start/stop the service appropriately.

I don't see any randomness in the original code, can you please elaborate?

yocalebo:
Why did we randomly choose `cifs, iscsitarget, rsync, snmp`. What does it accomplish by starting those here?

themylogin:
@yocalebo so by "randomness" you mean our wonderful QE team approach to developing tests, I get it.

Do we already have other integration tests that verify that each of these services is able to start in default configuration? If not, I would like to ask you to add them so we don't lose at least those.

Also it would be better to group `test_01_enable_service` and `test_02_verify_service_was_enabled` into one test case, same for disabing. What sense does it have to split this into two cases?

yocalebo:
> @yocalebo so by "randomness" you mean our wonderful QE team approach to developing tests, I get it.
> 
> Do we already have other integration tests that verify that each of these services is able to start in default configuration? If not, I would like to ask you to add them so we don't lose at least those.
> 
> Also it would be better to group `test_01_enable_service` and `test_02_verify_service_was_enabled` into one test case, same for disabing. What sense does it have to split this into two cases?

Okay sounds good, I did verify that those services are tested elsewhere.
`cifs` is handled in the various `test_420_smb.py` script
`iscsitarget` is handled in the `test_260_iscsi.py` script
`snmp` is handled in the `test_440_snmp.py` script
`rsync` is handled in the `test_380_rsync.py` script

themylogin (approved):

PR #8580, opened 2022-03-18
NAS-115356 / 22.12 / Disable root squashing in export config on demand

anodos325:
If user has specified that they want to set anonuid and anongid to id 0, then disable root squashing explicitly.


rick-mesta (approved):

phpfs:
@anodos325 Thank you very much, this addresses our issue perfectly!

One small question - would it make sense to move the definition of the `map_ids` dictionary into the `generate_options()` function _(into line 33)_ so it is reset for every host? Or this is already handled somehow? :)

PR #8590, opened 2022-03-19
NAS-114311 / 22.12 / Add APIs to prune unused containers/images

Qubad786:
This PR adds an API that prunes:
1. Stopped containers if `remove_stopped_containers` is set
2. Dangling Images
3. Images that are not used by any container if `remove_unused_images` is set


sonicaj (commented):
Let's please add some integration tests for this.

sonicaj (approved):
Thanks!

PR #8593, opened 2022-03-19
NAS-115222 / 22.12 / Explicitly ask for web port for Display devices

sonicaj:
Right now we have some magic to determine display device client/web port via server port. This PR adds changes to be explicit about it and let the user configure a specific port if he/she desires.

### DIFF
--- a/src/middlewared/middlewared/alembic/versions/22.12/2022-03-19_20-20_display_devices.py
+++ b/src/middlewared/middlewared/alembic/versions/22.12/2022-03-19_20-20_display_devices.py
@@ -1,0 +1,47 @@
+"""
+Normalize VM Display Device Port(s)
+
+Revision ID: d46d345ef8a8
+Revises: 0b02e21a1a10
+Create Date: 2022-03-19 20:20:11.823446+00:00
+
+"""
+import json
+
+from alembic import op
+
+
+revision = 'd46d345ef8a8'
+down_revision = '0b02e21a1a10'
+branch_labels = None
+depends_on = None
+
+
+def get_web_port(port):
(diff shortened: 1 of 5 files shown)


yocalebo (commented):
Are these ports used on the host system? If they are, are we actually seeing if these ports are already in use or does this get handled somewhere and I'm just not seeing it?

sonicaj:
@yocalebo we are ensuring that these ports are not being used by any other VM Display devices ?
If you are referring to scanning all the system ports and ensuring none of them are in use by apps or any other system service, then i believe that's a different scope for a different ticket ;) ( as we don't do that anywhere right now )

yocalebo:
> @yocalebo we are ensuring that these ports are not being used by any other VM Display devices ?
> 
> If you are referring to scanning all the system ports and ensuring none of them are in use by apps or any other system service, then i believe that's a different scope for a different ticket ;) ( as we don't do that anywhere right now )

No, what I'm saying is to try a temporary bind on the port that was given to us and check for errno.EADDRINUSE or use the socket.connect_ex which returns a bool instead of raising an exception

sonicaj:
@yocalebo how does that fix the case when some VM is not running atm but has same port as this new display device ?

Again, i think a better approach is to fix this at framework level - maybe writing a delegate for it 🤔

sonicaj:
Also the same problem lies with the other services/apps we have ( apps in themselves do have checks to not re-use the same port ). So imho we should look at fixing this from that perspective and not just having a specific check for VMs.

yocalebo (approved):
Not sure if you updated this migration but master has had another  migration merged. If you have, ignore me 😄

PR #8595, opened 2022-03-20
NAS-115329 / 13.0 / rate limit and add settle time to disk events

yocalebo:
Now that we are caching disk information on 13 (to make middleware usable on very large systems (100's of HDDs)), it has been found that when an expansion shelf is attached/detached we receive a "burst" of events all within <= 1 second but we're missing some of these events so the in-memory cache isn't properly invalidated/updated.

This PR adds the following logic.
1. schedule a task to be run within (`SETTLE_TIME`) seconds
2. if another event is received within `SETTLE_TIME`, then we cancel the current task and schedule another task to run within `SETTLE_TIME`.
3. if we keep receiving events while we already have a scheduled task, then we cancel the scheduled task and schedule a new one to run at `MAX_WAIT_TIME` seconds, then we will log a warning and then reset the deque object.
4. we now monitor for `ses` events (attaching/detaching) of expansion shelves and act accordingly

The simple ideas that this tries to enforce:
1. give middleware service `SETTLE_TIME` (or `MAX_WAIT_TIME`) seconds to ensure we don't have a burst of events
2. ensure middleware doesn't miss any disk events so the cache remains valid
3. rate limit the amount of devd events we act upon

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/disk_events.py
+++ b/src/middlewared/middlewared/plugins/disk_/disk_events.py
@@ -2,16 +2,20 @@ import asyncio
 
 DISK = ('da', 'ada', 'vtbd', 'mfid', 'nvd', 'pmem')
 SHELF = ('ses',)
+TYPES = ('CREATE', 'DESTROY')
 PREVIOUS = {'method': '', 'task': None}
 MAX_WAIT_TIME = 60
-SETTLE_TIME = 2
+SETTLE_TIME = 5
 LAST_EVENT_TIME = None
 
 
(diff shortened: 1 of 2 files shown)


yocalebo:
This has been tested programmatically as well as on a real system with > 800 hard drives.

rick-mesta:
@anodos325: Could I get your 👀 on this bad boy, please ? Even though @yocalebo has tested this on real HW for a couple of days, you may catch something obvious 🙏🏼

anodos325 (approved):

PR #8596, opened 2022-03-20
NAS-114497 / 22.12 / Allow using min/max txg for querying snapshots

sonicaj:


themylogin:
Can we use `txg` to paginate a huge list of snapshots?

themylogin (approved):

themylogin (changes requested):
Can we have an integration test please?

sonicaj:
Definitely

yocalebo (approved):

PR #8597, opened 2022-03-21
NFS server: change mapping of gids fully optional

phpfs:
Hi there!

We found that with TrueNAS Scale 22.02 RELEASE, the `--manage-gids` option is now always added to `rpc.mountd` options even though it was not enabled in web UI. Based upon the code, I think that this currently tied to enabling NFSv4. From our testing, it is totally fine to run NFSv4 without managed groups hence this PR provides this functionality again :)

Best, phpfs

P.S.: Open to discuss different solutions here! (@anodos325)


anodos325:
Yeah, that seems reasonable.

anodos325 (approved):

phpfs:
Could you please backport this change? We are very much dependent upon it. @anodos325

PR #8607, opened 2022-03-21
NAS-115383 / 22.12 / Prevent contents of /etc/exports.d from altering NFS config

anodos325:
knfsd / exportfs will attempt to load any exports file added
to /etc/exports.d. This can lead to filesystems being
unexpectedly exported and a broken NFS configuration.

In this PR, we add an additional check when generating
exports where we remove any unexpected file from
/etc/exports.d and then make the directory immutable.

If for some reason, a file cannot be removed, then we
intentionally generate an invalid exports config with
a comment stating that the user needs to clean up whatever
he or she did there.

### DIFF
--- a/src/middlewared/middlewared/etc_files/exports.mako
+++ b/src/middlewared/middlewared/etc_files/exports.mako
@@ -1,6 +1,7 @@
 <%
     import ipaddress
     import socket
+    import os
     from pathlib import Path
 
     map_ids = {
@@ -95,11 +96,70 @@
 
         return hostname
 
+    def disable_sharenfs():
+        datasets = []
+        with open('/etc/exports.d/zfs.exports', 'r') as f:
+           for line in f:
+               if not line.strip() or line.startswith('#'):
+                   continue
+
+               ds_name = middleware.call_sync(
+                   'zfs.dataset.path_to_dataset',
+                   line.rsplit(" ", 1)[0]
+               )
+               if ds_name is None:
+                   middleware.logger.warning("%s: dataset lookup failed", line)
+                   continue
+
+               datasets.append(ds_name)
+
+        for ds in datasets:
+            try:
+                middleware.call_sync(
+                    'zfs.dataset.update',
+                    ds,
+                    {'properties': {'sharenfs': {'value': 'off'}}}
+                )
+            except Exception:
+                middleware.logger.warning("%s: failed to disable sharenfs", ds, exc_info=True)
+
+        return
+
+    def disable_exportsd():
+        immutable_disabled = False
+        for file in os.listdir('/etc/exports.d'):
+            if not immutable_disabled:
+                middleware.call_sync('filesystem.set_immutable', False, '/etc/exports.d')
+                immutable_disabled = True
+
+            if file == 'zfs.exports':
+                middleware.logger.warning("Disabling sharenfs ZFS property on datasets")
+                disable_sharenfs()
+            else:
+                middleware.logger.warning("%s: unexpected file found in exports.d", file)
+
+            try:
+                os.remove(os.path.join('/etc/exports.d', file))
+            except Exception:
+                middleware.logger.warning(
+                    "%s: failed to remove unexpected file in exports.d",
+                    file, exc_info=True
+                )
+                return False
+
+        if not immutable_disabled and middleware.call_sync('filesystem.is_immutable', '/etc/exports.d'):
+            return True
+
+        middleware.call_sync('filesystem.set_immutable', True, '/etc/exports.d')
+        return True
+
     entries = []
     gaierrors = []
     config = render_ctx["nfs.config"]
     shares = render_ctx["sharing.nfs.query"]
-    if not shares:
+    poison_exports = not disable_exportsd()
+
+    if not poison_exports and not shares:
         raise FileShouldNotExist()
 
     has_nfs_principal = middleware.call_sync('kerberos.keytab.has_nfs_principal')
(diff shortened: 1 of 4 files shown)


anodos325:
When I can't think of what to name something, I just keep adding words. More concise or useful property name would be appreciated.

anodos325:
Flagging Vlad for review so that he can tell me to write a unit test (among other things) :)

src/middlewared/middlewared/etc_files/exports.mako
```diff
@@ -95,6 +96,15 @@
 
         return hostname
 
+    def disable_exportsd():
+        if middleware.call_sync('filesystem.is_immutable', '/etc/exports.d'):
```
themylogin, line 100: What if user removes immutable flag, puts his stuff there and sets it back? :grinning:

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1346,6 +1347,56 @@ async def import_find(self, job):
...
+            properties = {
+                'aclinherit': 'PASSTHROUGH',
+            }
+        else:
+            properties = {
+                'aclinherit': 'DISCARD',
+                'aclmode': 'DISCARD'
+            }
+
+        properties['user_propeties'] = ds['user_proeprties']
```
themylogin, line 1374: `proeprties`

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1346,6 +1347,56 @@ async def import_find(self, job):
...
+
+        recursive = True
+        for child in (await self.middleware.call('pool.dataset.query', [('id', '=', ds['id'])], {'get': True}))['children']:
```
themylogin, line 1377: We used `zfs.dataset.child_dataset_names`, which has a better performance, hasn't it? @sonicaj Why do we need to retrieve all datasets metadata if we only need their names?

Also what about root dataset `sharenfs`, `sharesmb` and `mountpoint` properties?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -1406,23 +1457,15 @@ async def import_pool(self, job, data):
...
-                # We exclude `ix-applications` dataset since resetting it will
-                # cause PVC's to not mount because "mountpoint=legacy" is expected.
-                continue
-            try:
-                # Reset all mountpoints
-                await self.middleware.call('zfs.dataset.inherit', child, 'mountpoint', recursive)
-            except Exception as e:
-                # Let's not make this fatal
-                self.logger.warning('Failed to inherit mountpoints recursively for %r dataset: %r', child, e)
+        if ds_defaults_version != TRUENAS_DEFAULTS_VERSION:
```
themylogin, line 1467: This check will prevent us from resetting bad properties set via CLI after the pool first import.

themylogin (dismissed):

themylogin:
@anodos325 and yes, now you can use this https://github.com/truenas/middleware/pull/8608 to add an integration test (create pool, screw everything up, export it, import it and check that bad properties were removed)

src/middlewared/middlewared/etc_files/exports.mako
anodos325: Indeed. I was not thinking creatively enough.

src/middlewared/middlewared/plugins/pool.py
anodos325: I thought primary concern was users importing foreign pools?

themylogin: Foot-shooters are also a big concern, aren't they?

src/middlewared/middlewared/plugins/pool.py
anodos325: I backed out the pool changes in favor of skipping zfs_sharenfs() in py-libzfs during pool import there.

I'll see about maybe being more intelligent with how we interpret failures to inherit mountpoint property (i.e. disable sharenfs / sharesmb in this case (if they're set).

src/middlewared/middlewared/etc_files/exports.mako
```diff
@@ -105,6 +164,15 @@
     has_nfs_principal = middleware.call_sync('kerberos.keytab.has_nfs_principal')
     global_sec = middleware.call_sync("nfs.sec", config, has_nfs_principal) or ["sys"]
 
+    try:
+        disable_exportsd()
+
+    except FileShouldNotExist:
```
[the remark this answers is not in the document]
anodos325: Heh. Okay.

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -980,7 +980,15 @@ def inherit(self, name, prop, recursive=False):
                     raise CallError(f'Property {prop!r} not found.', errno.ENOENT)
                 zprop.inherit(recursive=recursive)
         except libzfs.ZFSException as e:
-            raise CallError(str(e))
+            if prop != 'mountpoint':
+                raise CallError(str(e))
+
+            err = e.code.name
+            if err not in ("SHARENFSFAILED", "SHARESMBFAILED"):
```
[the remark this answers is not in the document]
anodos325: An example of why it would fail is if for some reason immutable flag is set on /etc/exports.d.

anodos325:
Flagged Eric for review of test.

ericbsd (approved):
Test LGTM

src/middlewared/middlewared/plugins/zfs.py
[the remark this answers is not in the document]
anodos325: Added comment

yocalebo (approved):

PR #8613, opened 2022-03-22
NAS-115412 / 22.12 / improve disk.get_dev_size

yocalebo:
Been meaning to push this PR for a while but this simplifies and speeds up (in the case a disk partition is passed to us) the `disk.get_dev_size` method. The `pyudev.Devices` object that we create gives us all the information we need without having to call another method in another module that goes out to disk and reads a file.

### DIFF
--- a/tests/api2/test_disk_get_dev_size.py
+++ b/tests/api2/test_disk_get_dev_size.py
@@ -1,0 +1,11 @@
+import pytest
+
+from middlewared.test.integration.utils import call
+
+DISKS = list(call('device.get_disks').keys())
+TYPES = (type(None), int)
+
+
+@pytest.mark.parametrize('disk', DISKS)
+def test_get_dev_size_for(disk):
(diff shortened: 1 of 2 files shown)


themylogin (approved):
It would be great to have an integration test!

yocalebo:
> It would be great to have an integration test!

done 👍

tests/api2/test_disk_get_dev_size.py
```diff
@@ -0,0 +1,11 @@
...
+
+from middlewared.test.integration.utils import call
+
+DISKS = list(call('device.get_disks').keys())
+TYPES = (type(None), int)
+
+
+@pytest.mark.parametrize('disk', DISKS)
+def test_get_dev_size_for(disk):
+    assert isinstance(call('disk.get_dev_size', disk), TYPES)
```
themylogin, line 11: What does this test? :(

We can make a call for sda and sda3 (or whatever ZFS partition is on boot pool device) and compare results with e.g. fdisk output

PR #8618, opened 2022-03-23
net.inet.tcp.recvbuf_inc was removed in 13

yocalebo:
This was removed in 13 and so, therefore, not needed.

### DIFF
--- a/src/autotune/files/autotune.py
+++ b/src/autotune/files/autotune.py
@@ -100,7 +100,6 @@ DEF_KNOBS = {
         'net.inet.tcp.sendbuf_max',
         'net.inet.tcp.recvbuf_max',
         'net.inet.tcp.sendbuf_inc',
-        'net.inet.tcp.recvbuf_inc',
         'vfs.zfs.vdev.async_read_max_active',
         'vfs.zfs.vdev.sync_read_max_active',
         'vfs.zfs.vdev.async_write_max_active',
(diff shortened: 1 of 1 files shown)


amotin:
I suppose people already having the tunable will keep it even after this, still getting the noise on boot.

anodos325 (approved):

yocalebo:
@themylogin asking for your specific review because I routinely break migration code and I'd rather not do that if possible 😄

src/middlewared/middlewared/alembic/versions/13.0/2022-03-23_11-40_remove_recvbuf_inc_tunable.py
```diff
@@ -0,0 +1,17 @@
...
+Revision ID: 88bfe11b5be5
+Revises: 99aef90c4cd6
+Create Date: 2022-03-23 11:40:32.149486+00:00
+
+"""
+from alembic import op
+
+
+revision = '88bfe11b5be5'
+down_revision = '99aef90c4cd6'
```
themylogin, line 12: Does it work without other variables and `downgrade` defined?

themylogin (commented):
You'll have to backport this migration to 22.02 and `master` even if it does not have sense

themylogin:
@yocalebo so `branch_labels` and `depends_on` are optional?

themylogin (approved):

themylogin:
Also it would be easier to backport only to 22.02.1, generate a merge migration there, and then backport the backport to 22.12, otherwise you'll have to copy this merge migration manually.

PR #8624, opened 2022-03-23
NAS-115438 / 22.12 / Convert nfs_paths to nfs_path and remove alldirs

anodos325:
nfs_paths as a list is a FreeBSD-ism that doesn't really fit well
in with how exports are generated. This PR migrates cases where
users have multiple paths for a single export into separate exports
with a single path each. Tests are accordingly changed for new
NFS share structure.

This benefits us in two ways:
1) it gives some flexibility for implementing NFS aliases
2) it simplifies logic for determining whether an export is locked


anodos325:
This is currently targeting 22.12 as opposed to 22.02 because it changes NFS sharing API in a significant way. I'm open to backporting to 22.02 if we feel like that's a better approach.

src/middlewared/middlewared/alembic/versions/22.12/2022-03-23_16-23_fix-nfs-paths.py
```diff
@@ -0,0 +1,47 @@
...
+    with op.batch_alter_table('sharing_nfs_share', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('nfs_path', sa.TEXT(), nullable=True))
+
+    conn = op.get_bind()
+    nfs_shares = [dict(row) for row in conn.execute("SELECT * FROM sharing_nfs_share").fetchall()]
+    for entry in nfs_shares:
+        # use existing entry for first path and add new entries for subsequent paths
+        _id = entry.pop("id")
+        paths = eval(entry['nfs_paths'])
+        first_path = paths.pop(0)
```
sonicaj, line 30: Just for good measure, do you think we should maybe handle this safely if for some reason the list is empty and migration fails ?

src/middlewared/middlewared/alembic/versions/22.12/2022-03-23_16-23_fix-nfs-paths.py
```diff
@@ -0,0 +1,47 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('sharing_nfs_share', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('nfs_path', sa.TEXT(), nullable=True))
+
+    conn = op.get_bind()
+    nfs_shares = [dict(row) for row in conn.execute("SELECT * FROM sharing_nfs_share").fetchall()]
+    for entry in nfs_shares:
+        # use existing entry for first path and add new entries for subsequent paths
+        _id = entry.pop("id")
+        paths = eval(entry['nfs_paths'])
```
sonicaj, line 29: Why not use `json.loads` here ?

src/middlewared/middlewared/alembic/versions/22.12/2022-03-23_16-23_fix-nfs-paths.py
```diff
@@ -0,0 +1,47 @@
...
+        first_path = paths.pop(0)
+        conn.execute(f'UPDATE sharing_nfs_share SET nfs_path = "{first_path}" WHERE id = {_id}')
+        if not paths:
+            continue
+
+        for path in paths:
```
sonicaj, line 35: Probably do this safely as well ensuring `path` is not empty ?

sonicaj (commented):
We have other usages of `path_field` in the NFS class which are not being handled as it still expects it to be a list rather then a string. Let's please cover those usages as well

src/middlewared/middlewared/alembic/versions/22.12/2022-03-23_16-23_fix-nfs-paths.py
```diff
@@ -0,0 +1,56 @@
...
+        # use existing entry for first path and add new entries for subsequent paths
+        _id = entry.pop("id")
+        try:
+            paths = json.loads(entry['nfs_paths'])
+        except Exception:
+            paths = ["/var/empty"]
+
+        try:
+            first_path = paths.pop(0)
+        except KeyError:
```
sonicaj, line 37: It would be `IndexError`

anodos325: Thanks. I was finishing up a few more changes before moving to my test server to validate migration :)

sonicaj (approved):
Thanks!

PR #8629, opened 2022-03-24
NFS: Fix map_id for multiple shares

phpfs:
Hi @anodos325 ,

I noticed that our latest changes to NFS export generation did not consider multiple shares with different configurations. Is this correct? If so, this PR supplies a possible solution.

Best, phpfs

### DIFF
--- a/src/middlewared/middlewared/etc_files/exports.mako
+++ b/src/middlewared/middlewared/etc_files/exports.mako
@@ -32,7 +32,6 @@
             'mapall_group': -1,
         }
 
-        all_squash = False
         if share["security"]:
             sec = f'sec={":".join(share["security"])}'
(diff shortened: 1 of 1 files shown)


anodos325:
https://github.com/truenas/middleware/pull/8630

anodos325:
Regression test: https://github.com/truenas/middleware/pull/8631
22.02 PR: https://github.com/truenas/middleware/pull/8633
Fix for regression caused by this merge: https://github.com/truenas/middleware/pull/8632 :)

phpfs:
@anodos325 I think there is one small issue remaining: when we set "no_root_squash", we should **not** set "anonuid" and "anongid" - or is there a reason for that?!

PR #8630, opened 2022-03-24
NFS: Fix map_id for multiple shares (#8629)

anodos325:
Ensure that root_squash is properly evaluated on a per-share basis when generating exports file.


phpfs:
by the way - when is 22.02.01 going to be released? :)

PR #8631, opened 2022-03-24
NAS-115448 / 22.12 / Expand maproot tests

anodos325:
When maproot_user and maproot_group are set to 'root', then we
also add configuration option 'no_root_squash' to the exports
line. This test ensures that this parameter is being properly
added and that it does not get added to other shares.


yocalebo (approved):

ericbsd (approved):
LGTM

PR #8635, opened 2022-03-24
NAS-115450 / 22.12 / Remove anonuid / anongid if user sets uid 0 / gid 0 as maproot

anodos325:
Linux anonuid / anongid don't 100% match up with maproot behavior
on FreeBSD. The anonuid and anongid also are what is used in case
of INVALID_UID or INVALID_GID for fsuid/fsgid in fs/nfsd/auth.c

This means that we don't want this parameter populated when user
wants to simply not squash root.

### DIFF
--- a/src/middlewared/middlewared/etc_files/exports.mako
+++ b/src/middlewared/middlewared/etc_files/exports.mako
@@ -60,6 +60,10 @@
 
         try:
             maproot = do_map(share, "maproot", map_ids)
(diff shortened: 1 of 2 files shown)


anodos325:
In practice this shouldn't happen because I believe ZFS will assert on uid / gid == -1

phpfs (approved):
Looks good to me!

PR #8639, opened 2022-03-24
NAS-115456 / 22.12 / Fix logic for ms-account validation

anodos325:
We shouldn't  allow this to be set on builtin accounts.


yocalebo (approved):

denysbutenko:
@anodos325, should we make changes in WebUI?

anodos325:
@denysbutenko, yeah, let's do that too.

PR #8642, opened 2022-03-25
NAS-115469 / 22.12 / Optimize acl_is_trivial

anodos325:
This method is called when we call filesystem.listdir and
filesystem.stat. Unfortunately because of how it was implemented
for cases with relatively largish numbers of files, it would
become quite slow. filesystem.listdir took ~30 seconds for a
directory with 1000 files. Now that the ZFS ACL_IS_TRIVIAL
flag is included in the ACL flags, we can directly check this
via a getxattr call. With this new optimization, filesystem.listdir
on the same directory now takes less than 1 second.


anodos325:
depends on https://github.com/truenas/zfs/pull/52

anodos325:
In future when I have time I'll alter the xattr handler for the NFSv4 xattr to only return it in listxattr results if the ACL is non-trivial. This will allow us to determine whether it's trivial via simple os.listxattr() call.

sonicaj (approved):

anodos325:
Decided to go route with xattr handler changes for listdir(), PR changed accordingly.

PR #8649, opened 2022-03-28
NAS-115485 / 22.12 / Bug fix for normalizing SAN

sonicaj:
This commit fixes an issue where if we have a CSR, it can already potentially have normalized SAN. In that case, let's not normalize it again as that will result in an inconsistent SAN value.


themylogin (changes requested):
LGTM, but add a unit test please

themylogin (approved):
Thank you!

PR #8661, opened 2022-03-29
update paths to path

q5sys:
Due to change here: https://github.com/truenas/middleware/commit/558ea76887b99134f16b11914671b31e3847def4

The test needs to be updated as well, since we've got nightly API test failures: https://ci.tn.ixsystems.net/jenkins/job/TrueNAS%20SCALE%20-%20Unstable/job/API%20Tests%20-%20TrueNAS%20SCALE%20(Full%20-%20Nightly%20ISO)/857/testReport/junit/api2/test_032_ad_kerberos/test_30_check_nfs_exports_sec/


yocalebo (changes requested):
The fix is in PR https://github.com/truenas/middleware/pull/8660. This still leaves the variable as a list object which won't work.

PR #8663, opened 2022-03-29
NAS-115521 / 22.12 / return stp info in interface.query

yocalebo:
webUI needs this info to be returned

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -269,16 +269,12 @@ class InterfaceService(CRUDService):
                 })
 
         if itype == InterfaceType.BRIDGE:
-            bridge = self.middleware.call_sync(
-                'datastore.query',
-                'network.bridge',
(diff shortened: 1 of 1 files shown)


Qubad786 (approved):

src/middlewared/middlewared/plugins/network.py
```diff
@@ -269,16 +269,12 @@ def iface_extend(self, iface_state, configs, ha_hardware, fake=False):
...
-                'datastore.query',
-                'network.bridge',
-                [('interface', '=', config['id'])],
-            )
-            if bridge:
-                bridge = bridge[0]
-                iface.update({'bridge_members': bridge['members']})
+            filters = [('interface', '=', config['id'])]
+            if br := self.middleware.call_sync('datastore.query', 'network.bridge', filters):
+                iface.update({'bridge_members': br[0]['members'], 'stp': br[0]['stp']})
```
sonicaj, line 274: Has this been added already to `returns` decorator ?

PR #8675, opened 2022-03-30
NAS-115542 / 22.12 / fix uuidd service

yocalebo:
2 issues.

1. `uuidd` user/group don't exist so override the systemd unit file with `nobody/nogroup` since we don't need or use this service. If a user manually starts this service, it will at least start with the overridden user/group information.
2. mask `uuidd.service` and `uuidd.socket` because, again, we don't need or use it

### DIFF
--- a/debian/debian/postinst
+++ b/debian/debian/postinst
@@ -71,8 +71,10 @@ done
 systemctl daemon-reload
 systemctl enable zfs-zed
 
-# We need to mask libvirtd related sockets and other services so that they don't start automatically
-systemctl mask exim4-base.service exim4.service libvirtd.socket libvirtd-ro.socket libvirtd-admin.socket libvirtd-tls.socket libvirtd-tcp.socket
+# We need to mask certain services so that they don't start automatically
+systemctl mask libvirtd.socket libvirtd-ro.socket libvirtd-admin.socket libvirtd-tls.socket libvirtd-tcp.socket
+systemctl mask exim4-base.service exim4.service
+systemctl mask uuidd.service uuidd.socket
 
 systemctl set-default truenas.target
 
--- a/src/freenas/etc/systemd/system/uuidd.service.d/override.conf
+++ b/src/freenas/etc/systemd/system/uuidd.service.d/override.conf
@@ -1,4 +1,0 @@
-# we dont create a uuidd user/group so override with nobody/nogroup
-[Service]
-User=nobody
-Group=nogroup
--- a/src/freenas/etc/systemd/system/uuidd.service.d/override.conf
+++ b/src/freenas/etc/systemd/system/uuidd.service.d/override.conf
@@ -1,0 +1,4 @@
+# we dont create a uuidd user/group so override with nobody/nogroup
+[Service]
+User=nobody
+Group=nogroup


sonicaj (commented):
If we are not going to use / consume it, why maintain an override file for it ?

yocalebo:
> If we are not going to use / consume it, why maintain an override file for it ?

So that the service is at least functional. If a user manually starts the service, it will at least start with the override.

sonicaj:
Why do we want a user to start a service manually when we are explicitly disabling it ?

yocalebo:
> Why do we want a user to start a service manually when we are explicitly disabling it ?

We don't want the user to start it manually. This just allows it to be started if they like since they can do whatever they want. This mostly hides the failing console messages when the system boots.

sonicaj:
But if it is not starting and we don't it started, would those console messages appear still ?

sonicaj:
If a user wants to go to the extent where they are enabling services explicitly and changing system state - i believe they are already way beyond what we would actually like them to do and they can configure the user themselves at that point.

yocalebo:
> But if it is not starting and we don't want it started, would those console messages appear still ?

No, with my changes the messages won't appear.

sonicaj:
Right, so the override file is not required ?

yocalebo:
I don't see why it's an issue to have an override file.

sonicaj:
I don't see why it is required :) ?
We don't want the service to be started/configured and explicitly disable it in this regard - why should we maintain an override file for it ?

yocalebo:
> I don't see why it is required :) ? We don't want the service to be started/configured and explicitly disable it in this regard - why should we maintain an override file for it ?

We don't care if the service is started or enabled but this hides the failure messages on the console.

sonicaj:
Ah, so you are saying just explicitly disabling it is not enough and we need an override file for it to ensure the messages don't appear on the console ?

yocalebo:
> Ah, so you are saying just explicitly disabling it is not enough and we need an override file for it to ensure the messages don't appear on the console ?

No, I must not be explaining well enough and I'm tired of going back and forth over something that is trivial. So I've removed it.

sonicaj (approved):
Thanks for clarifying! To the best of my understanding ( if i am following correctly :p ), we won't be seeing the erroneous messages in the console now with the current diff

wolfspyre:
The issue is that as it was without this, that when a user queries the system's state to assess what services are in a failed state, uuidd is listed as failed. 

The uninitiated do not know if this is by design, laziness, or indicative of something broken.

If you're taking the stance that this is an appliance, and the lower layers are essentially off limits to users, then the desired state should reflect everything working as expected.

IOW: the default state of services... whether they are in use, or not, shouldn't be in error. 

The uuidd service was disabled... ish... but in a way that caused a failure state.

Caleb's fix leaves the world in a state of homeostasis. actively disabling the service, and making sure that IF someone gets a wild hair and starts it, it will actually start cleanly. Whether that is done with an override file, or by re-enabling the uuidd user/group are ... 6 of one, half a dozen of the other... either solve the problem of not generating red herrings when trying to diagnose something wonky.


IOW: 

The purpose of the change is essentially the difference between

-  'I dont use hammers, so I dont care if there is a hammer lying on the ground there'

versus:

- ' we don't use hammers, but I'mma pick this hammer thats lying on the floor up and put it away so someone doesn't stub their toe'

PR #8679, opened 2022-03-30
NAS-115561 / 22.12 / Provide `sensors -j` output in debug

themylogin:
I have ask for this too often before telling another AMD user that we can't do anything
with the fact that their CPU provides three temperatures with obscure labels, none of
which is the actual temperature of any point of their CPU.


themylogin:
time 0:05

yocalebo (approved):

PR #8689, opened 2022-03-31
NAS-115580 / 13.0 / Fix AD join failure messages

anodos325:
This fixes a couple of issues that crept in while backporting
middleware fixes to work with Samba 4.15.


anodos325:
Only impacts 13.

rick-mesta (approved):

PR #8692, opened 2022-03-31
NAS-115582 / 22.12 / add network_/common.py plugin

yocalebo:
This fixes 3 issues:

1. we were raising `CallError` for all interface CRUD validations instead of `ValidationErrors`
2. in `do_delete` we cached the database entries in memory before doing the appropriate validation, so it was possible to commit and checkin the changes even though an error was raised
3. move the `__check*` methods to their own module to clean up `network.py`


yocalebo:
retest this please

PR #8703, opened 2022-04-01
NAS-115595 / remove interfaces parameter from Auxiliary parameters Blacklist

intelicore:
Bug: https://jira.ixsystems.com/browse/NAS-115595


anodos325:
We need to maintain this blacklist to prevent inconsistent state in clustered environments.

PR #8710, opened 2022-04-03
NAS-112058 / 22.12 / Allow intel GPU's to be used by multiple Apps

Ornias1993:
Reopen of #7431 as @sonicaj didn't create the promissed PR there and users are complaining about lack of progress on the iX Discord.
---
As discussed and not reported on Jira (afaik) here:
https://www.truenas.com/community/threads/intel-gpu-plugin.94955/

This allows the Intel GPU's using the Intel GPU plugin, to be used by multiple Apps at once.
As it currently throws errors when used by more than one App at the same time.

It's currently set to 5, which is more than the total amount of available GPU-supporting Apps in both Official and TrueCharts combined.


sonicaj (approved):

sonicaj:
backport

PR #8715, opened 2022-04-04
NAS-115037 / 22.12 / Add changes to comply with openssl security level 2

sonicaj:
This PR adds changes to ensure we comply with openssl security level 2 which enforces certain restrictions on what kind of certs can be used.

It also adds a migration which will ensure that if a user has a malformed cert configured before these changes went in or is coming from CORE for example, we validate his cert and generate a self signed cert if his selected cert is invalid. Motivation behind this is that without this change nginx would fail to start complaining about openssl errors.

### DIFF
--- a/src/middlewared/middlewared/migration/0006_system_cert.py
+++ b/src/middlewared/middlewared/migration/0006_system_cert.py
@@ -1,0 +1,8 @@
+async def migrate(middleware):
+    # This is required because earlier middleware was running considering openssl security level to be 1
+    # but with moving to debian, the default now is 2 which enforces security standards. It means that potentially
+    # user might have configured cert for system which might not comply with security level 2 and UI becoming
+    # inaccessible in this regard because of that
+    system_cert = (await middleware.call('system.general.config'))['ui_certificate']
+    if await middleware.call('certificate.cert_services_validation', system_cert['id'], 'certificate', False):
+        await middleware.call('certificate.setup_self_signed_cert_for_ui')
--- a/src/middlewared/middlewared/migration/0006_system_cert.py
+++ b/src/middlewared/middlewared/migration/0006_system_cert.py
@@ -8,3 +8,4 @@ async def migrate(middleware):
         'certificate.cert_services_validation', system_cert['id'], 'certificate', False
     ):
         await middleware.call('certificate.setup_self_signed_cert_for_ui')
+        await middleware.call('service.restart', 'http')
--- a/src/middlewared/middlewared/migration/0006_system_cert.py
+++ b/src/middlewared/middlewared/migration/0006_system_cert.py
@@ -4,5 +4,7 @@ async def migrate(middleware):
     # user might have configured cert for system which might not comply with security level 2 and UI becoming
     # inaccessible in this regard because of that
     system_cert = (await middleware.call('system.general.config'))['ui_certificate']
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/migration/0006_system_cert.py
```diff
@@ -0,0 +1,8 @@
+async def migrate(middleware):
+    # This is required because earlier middleware was running considering openssl security level to be 1
+    # but with moving to debian, the default now is 2 which enforces security standards. It means that potentially
+    # user might have configured cert for system which might not comply with security level 2 and UI becoming
+    # inaccessible in this regard because of that
+    system_cert = (await middleware.call('system.general.config'))['ui_certificate']
```
themylogin, line 6: Do we always have this certificate at this point? I recall seeing systems without it.

src/middlewared/middlewared/plugins/crypto_/certificates.py
```diff
@@ -103,10 +104,18 @@ async def cert_services_validation(self, id, schema_name, raise_verrors=True):
...
-                    f'{cert["name"]}\'s private key size is less then 1024 bits'
+                    f'{cert["name"]}\'s private key size is less then {valid_key_size[cert["key_type"]]} bits'
+                )
+
+            if cert['until'] and datetime.datetime.strptime(
+                cert['until'], '%a %b  %d %H:%M:%S %Y'
+            ) < datetime.datetime.now():
+                verrors.add(
+                    schema_name,
+                    f'{cert["name"]!r} has expired ( it was valid until {cert["until"]} )'
```
themylogin, line 118: Why the whitespace after bracket? `expired ( it was valid`

src/middlewared/middlewared/plugins/crypto_/default_cert.py
```diff
@@ -0,0 +1,52 @@
...
+        cert, key = await self.middleware.call('cryptokey.generate_self_signed_certificate')
+
+        cert_dict = {
+            'certificate': cert,
+            'privatekey': key,
+            'name': cert_name,
+            'type': CERT_TYPE_EXISTING,
+        }
+
+        # We use datastore.insert to directly insert in db as jobs cannot be waited for at this point
```
themylogin, line 46: > jobs cannot be waited for at this point

Why?

sonicaj: This is an old comment and it was during the setup function of crypto plugin. Haven't verified if this holds right now 🤔

src/middlewared/middlewared/migration/0006_system_cert.py
sonicaj: We would have it as the crypto setup function ensures that. However i can still do this safely.

src/middlewared/middlewared/plugins/crypto_/default_cert.py
themylogin: Nothing changed since then. If you are going to check this again and it really fails, can you ping me please?

sonicaj: Definitely

sonicaj: We can definitely wait for jobs here if we want to - verified :)

themylogin (approved):
Thank you!

PR #8718, opened 2022-04-04
NAS-115646 / 22.12 / Optimize and fix tunable plugin for SCALE

yocalebo:
Many issues found:

1. tunable plugin was still referencing `loader` and `rc` as a tunable type which is CORE specific
2. There were unused methods in `tunable.py`
3. The CRUD logic didn't follow the "standard" pattern that we use today
4. We were running `subprocess` for every tunable that the end-user set
5. API tests weren't actually testing to see if the tunables being set were reflected in kernel
6. We're not using the convenient `systemd-sysctl` unit file

All issues are fixed but the primary one is that we no longer `subprocess` out for each tunable the end-user has set. Instead, we read what is in the database, write a singular file `/etc/sysctl.d/tunable.conf` with all the tunable entries that the end-user has specified and then run `service.restart sysctl` which reads the file from that directory and will set the parameters as expected. (No more `subprocess`)


yocalebo:
Builds successfully and new API tests pass.

yocalebo:
@ericbsd adding you to review the new API test module that I wrote. I have tested it on a live machine and it works.

ericbsd (commented):
tests/api2/test_520_tunable.py LGTM

src/middlewared/middlewared/etc_files/sysctl.py
```diff
@@ -0,0 +1,26 @@
...
+        elif tun['orig_value']:
+            # tunable is disabled and we have an original value before it was created so update
+            to_write[tun['var']] = tun['orig_value']
+
+    if not to_write:
+        # no tunables or all disabled so make sure the config file is removed
+        CONF_FILE.unlink(missing_ok=True)
+    else:
+        with CONF_FILE.open('w') as f:
+            f.writelines('\n'.join(f'{tunable}={value}' for tunable, value in to_write.items()))
```
sonicaj, line 22: Can values have spaces in them or quotes and in that case does the current configuration handles it appropriately ?

src/middlewared/middlewared/plugins/service_/services/sysctl.py
```diff
@@ -0,0 +1,8 @@
+from .base import SimpleService
+from middlewared.utils import run
```
sonicaj, line 2: Where is this used ?

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -25,212 +26,110 @@ class TunableService(CRUDService):
...
         """
         return {k: k for k in TUNABLE_TYPES}
 
     @accepts(Dict(
         'tunable_create',
-        Str('var', validators=[Match(r'^[\w\.\-]+$')], required=True),
+        Str('var', required=True),
         Str('value', required=True),
-        Str('type', enum=TUNABLE_TYPES, required=True),
+        Str('type', enum=TUNABLE_TYPES, default='SYSCTL', required=True),
```
sonicaj, line 68: Is this required anymore now that there's only one type of tunable ?

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -25,212 +26,110 @@ class TunableService(CRUDService):
...
     ENTRY = Patch(
         'tunable_create', 'tunable_entry',
         ('add', Int('id')),
     )
 
     @private
-    async def default_sysctl_config(self):
-        return self.__default_sysctl
+    def get_system_tunables(self):
+        if not TunableService.SYSTEM_TUNABLES:
```
sonicaj, line 40: You could also use functools.cache and not maintain a separate variable for future

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -25,212 +26,110 @@ class TunableService(CRUDService):
     class Config:
         datastore = 'system.tunable'
         datastore_prefix = 'tun_'
-        datastore_extend = 'tunable.upper'
         cli_namespace = 'system.tunable'
 
-    def __init__(self, *args, **kwargs):
-        super(TunableService, self).__init__(*args, **kwargs)
-        self.__default_sysctl = {}
+    SYSTEM_TUNABLES = deque()
```
sonicaj, line 31: Out of curiosity, any specific reason for deque ?

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -25,212 +26,110 @@ class TunableService(CRUDService):
...
+            tunables = subprocess.run(['sysctl', '-aN'], stdout=subprocess.PIPE)
+            for tunable in filter(lambda x: x, tunables.stdout.decode().split('\n')):
+                TunableService.SYSTEM_TUNABLES.append(tunable)
+        return TunableService.SYSTEM_TUNABLES
 
     @private
-    async def get_default_value(self, oid):
-        return self.__default_sysctl[oid]
+    def get_original_value(self, var):
+        with open(f'/proc/sys/{var.replace(".", "/")}') as f:
```
sonicaj, line 48: Do we have a guarantee that this file will always exist ? And if not, can we add some exception handling please ?

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -25,212 +26,110 @@ class TunableService(CRUDService):
...
-            'datastore.delete',
-            self._config.datastore,
-            id
-        )
-
-        await self.middleware.call('service.reload', tunable['type'].lower())
-
-        return response
+        _id = await self.get_instance(_id)
+        if _id['orig_value']:
```
sonicaj, line 130: I am torn on what to do here, we have explicit check in creation that if we are not able to determine original value - we are not going to allow that tunable to be created which implies that we should already have this value. Having this check means that we just have it for already configured tunables. I am not sure what might be the best approach here, would be really nice if we could somehow normalise the database to avoid this maybe by running another migration and determining the original value but haven't seen if the changed values will be applied already when we run the migration

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -25,212 +26,110 @@ class TunableService(CRUDService):
...
+            return old
 
-        await self.clean(new, 'tunable_update', old=old)
-        await self.validate(new, 'tunable_update')
+        if (comment := data.get('comment', '').strip()) and comment != new['comment']:
+            new['comment'] = comment
 
-        await self.lower(new)
+        if not new['enabled'] and new['orig_value']:
+            # before we delete from db, let's set the tunable back to it's original value
```
sonicaj, line 116: What are we deleting from the database ?

src/middlewared/middlewared/plugins/tunables.py
yocalebo: I routinely forget about that decorator 😄

src/middlewared/middlewared/plugins/tunables.py
yocalebo: The reason why I didn't is that this is a private method and if you look where it's being called from then we make sure that the tunable actually exist in `SYSTEM_TUNABLES`. This means the file will exist any other exception and it should crash here.

The only exception to this is if someone upgrades from core to scale and they have a bunch of sysctls set then if they try and `do_update` those it will crash. However, that will happen with or without these changes. I will be pushing another PR to try and fix the core to scale upgrade situation but wanted to get this done first.

src/middlewared/middlewared/plugins/tunables.py
yocalebo: Yeah, I thought about this too. I'm not positive that when `middlewared` service starts, that the system sysctl values are at their original value by the time that service starts. In theory they should be because I could change it so that `systemd-tunable` service runs after `middlewared`. Not sure, I'll have to test

src/middlewared/middlewared/plugins/service_/services/sysctl.py
yocalebo: Yeah, not sure. I'll fix

src/middlewared/middlewared/plugins/tunables.py
yocalebo: Memory efficiency compared to a list as well as faster `.append` attribute

src/middlewared/middlewared/plugins/tunables.py
yocalebo: Nope, definitely not

src/middlewared/middlewared/plugins/tunables.py
yocalebo: copy and paste error 😄

src/middlewared/middlewared/etc_files/sysctl.py
yocalebo: It is impossible to validate every single sysctl that the linux kernel provides and even more so to validate every combination that the user provides. I'm just passing whatever the user gives to us as is. The onus is on them to make sure it's the correct syntax.

src/middlewared/middlewared/plugins/tunables.py
yocalebo: After testing when `ix-update.service` runs, the kernel hasn't applied any user configured sysctl tunables so I can normalize the database on upgrade 😄

src/middlewared/middlewared/alembic/versions/22.12/2022-04-05_00-25_add_tun_orig_value.py
```diff
@@ -0,0 +1,36 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table('system_tunable', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('tun_orig_value', sa.String(length=512), server_default='', nullable=False))
+
+    for key in ('loader', 'rc'):  # these are keys specific to the CORE platform and don't apply on SCALE
+        op.execute(f'DELETE FROM system_tunable WHERE tun_type = "{key}" COLLATE NOCASE')
```
sonicaj, line 24: We could do this in one SQL statement as well :p

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -25,212 +26,107 @@ class TunableService(CRUDService):
...
+    def get_or_set(self, var, value=None):
+        with open(f'/proc/sys/{var.replace(".", "/")}', 'r' if not value else 'w') as f:
+            return f.read().strip() if not value else f.write(value)
 
     @accepts()
-    @returns(Dict(
-        'tunable_type_choices',
-        *[Str(k, enum=[k]) for k in TUNABLE_TYPES],
-    ))
+    @returns(Dict('tunable_type_choices', *[Str(k, enum=[k]) for k in TUNABLE_TYPES]))
```
sonicaj, line 52: We can remove this now as there is only 1 type

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -25,212 +26,107 @@ class TunableService(CRUDService):
...
         """
         return {k: k for k in TUNABLE_TYPES}
 
     @accepts(Dict(
         'tunable_create',
-        Str('var', validators=[Match(r'^[\w\.\-]+$')], required=True),
+        Str('var', required=True),
         Str('value', required=True),
-        Str('type', enum=TUNABLE_TYPES, required=True),
+        Str('type', default='SYSCTL'),
```
sonicaj, line 63: Type is not required anymore, let's please drop this column as well

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -25,212 +26,107 @@ class TunableService(CRUDService):
...
+    async def do_update(self, _id, data):
         """
         Update Tunable of `id`.
         """
-        old = await self.get_instance(id)
+        old = await self.get_instance(_id)
 
         new = old.copy()
         new.update(data)
+        if old == new:
```
sonicaj, line 106: We are note validating on update if the newer tunable is valid or not ?
Also what happens when both value/var are changed ?

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -25,212 +26,107 @@ class TunableService(CRUDService):
...
-        )
-
-        await self.middleware.call('service.reload', tunable['type'].lower())
-
-        return response
-
-    @private
-    async def lower(self, data):
-        data['type'] = data['type'].lower()
+        _id = await self.get_instance(_id)
```
sonicaj, line 126: Maybe not use `_id` here as it's a complete dictionary and just causes confusion in the usage 😅

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -25,212 +26,106 @@ class TunableService(CRUDService):
...
+            # nothing updated so return early
+            return old
 
-        await self.clean(new, 'tunable_update', old=old)
-        await self.validate(new, 'tunable_update')
+        if (comment := data.get('comment', '').strip()) and comment != new['comment']:
+            new['comment'] = comment
 
-        await self.lower(new)
+        await self.middleware.run_in_thread(self.get_or_set, new['var'], new['orig_value'])
```
sonicaj, line 113: Let's look at following failure cases:
1. `new['var']` value is `blahblah`
2. `old['var']` was valid and `new['var']` is valid but has changed, we don't set the old variable's default value and neither do we get the newer value's default
3. We are now unconditionally setting default value which doesn't make sense ( doing it when it was disabled does imho or variable has changed )

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -25,212 +26,106 @@ class TunableService(CRUDService):
...
+        old = await self.get_instance(_id)
 
         new = old.copy()
         new.update(data)
+        if old == new:
+            # nothing updated so return early
+            return old
 
-        await self.clean(new, 'tunable_update', old=old)
-        await self.validate(new, 'tunable_update')
```
sonicaj, line 107: We should also validate that if newer variable changed, we don't already have it in the database like we are doing in the creation

src/middlewared/middlewared/plugins/tunables.py
yocalebo: It doesn't make sense to update a tunable that exists in the database with a completely new variable for a different tunable so I think I'm going to remove that possibility. You either create a new entry or update an existing one with a different `value` or `comment` or `enable`.

sonicaj: That works with me ;)

sonicaj (approved):

PR #8721, opened 2022-04-05
NAS-115642 / 22.12 / generate dual-nvdimm config on gen3 mseries hardware

yocalebo:
By default, we statically load the ntb kernel modules (`modules.d/truenas.conf`) and (statically) load the associated kernel options in `modprobe.d/truenas.conf`. These options are for single nvdimm devices (x and gen1/2 m series), however our gen3+ m-series utilize dual-nvdimms which requires different kernel module options.

This makes it so that we add a new file `modprobe.d/truenas-dual-nvdimm.conf` and because the length of the file name is longer than `modprobe.d/truenas.conf`, the `systemd-modules-load.service` will overwrite any duplicate kernel module options from the file with the longer name.

This also makes it so that `ix-etc.service` loads before the `systemd-modules-load.service`.

### DIFF
--- a/src/middlewared/middlewared/etc_files/dual-nvdimm.mako
+++ b/src/middlewared/middlewared/etc_files/dual-nvdimm.mako
@@ -1,4 +1,6 @@
 <%
+    from packaging import version
+
     info = middleware.call_sync('system.dmidecode_info')
     vers = info['system-version']
     prod = info['system-product-name']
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/etc_files/dual-nvdimm.mako
```diff
@@ -0,0 +1,38 @@
...
+	raise FileShouldNotExist()
+
+    try:
+        curr_vers = version.parse(vers)
+        min_vers = version.Version('3.0')
+    except Exception as e:
+        middleware.logger.error('Failed determining hardware version: %r', e)
+        raise FileShouldNotExist()
+
+    if curr_vers.major == min_vers:
```
themylogin, line 22: When should this condition be true?
```
>>> version.parse('3.0').major == version.Version('3.0')
False
```

yocalebo: oops, fixed

PR #8722, opened 2022-04-05
NAS-115439 / 22.12 / Making sure helm apps are redeployed when some cert is renewed

Qubad786:
## Problem
White upgrading certs from middleware few scenarios might result in inconsistency like:
1. K8s not running
2. App rollback resulting in consuming an old cert
3. K8s complete cluster restore resulting in apps potentially using old cert(s)

### Desired behavior
Once certs are upgraded, apps consuming those certs should be updated regardless of any scenario(s).

### Solution
Redeploy apps which are consuming an outdated cert on the problematic scenarios.


Ornias1993 (approved):
It's a bit tricking, but i like this idea a lot :)

sonicaj (approved):

PR #8723, opened 2022-04-05
NAS-115671 / 22.12 / Properly disable haproxy service by default

sonicaj:


yocalebo:
I'm seeing that the asterisk doesn't work. NFS related services are failing too and I was messing around and got it working by typing the exact service name. I haven't pushed a commit or investigated further but more just FYI. You might want to make sure this works the way you think it does.

sonicaj:
@yocalebo i have confirmed it does
```
❯ systemctl status haproxy.service
● haproxy.service - HAProxy Load Balancer
     Loaded: loaded (/lib/systemd/system/haproxy.service; disabled; vendor preset: disabled)
     Active: inactive (dead)
       Docs: man:haproxy(1)
             file:/usr/share/doc/haproxy/configuration.txt.gz
❯
```

Also some services are really annoying and for those we need to explicitly mask/disable them in post install

yocalebo (approved):

PR #8735, opened 2022-04-06
Fix `system.info` schema

themylogin:


yocalebo (commented):
This should be set to a bool when it gets to "Error Correction Type".

themylogin:
@yocalebo if `Error Correction Type` is not present in the dmidecode output (e.g. if I run it on VirtualBox) then this variable stays an empty string which contradicts schema declaration and fails in debug mode.

yocalebo:
> @yocalebo if `Error Correction Type` is not present in the dmidecode output (e.g. if I run it on VirtualBox) then this variable stays an empty string which contradicts schema declaration and fails in debug mode.

Then add an "or not sect" on the last elif line and then please update the unit test with the text output from your virtual box machine so we can catch it in future 😄

themylogin:
@yocalebo we already have such entities among test cases. I updated unit test.

yocalebo (approved):

themylogin:
backport

PR #8739, opened 2022-04-06
NAS-115639 / 22.12 / properly display zfs dataset recordsize choices

yocalebo:
We need to base the maximum dataset recordsize off of `zfs_max_recordsize` module option. Furthermore, ZFS supports up to 16MB recordsize. All you have to do is `echo '16777216' > /sys/module/zfs/parameters/zfs_max_recordsize`.

This changes it so that we always determine the maximum size based off of that module option.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -3093,9 +3093,7 @@ class PoolDatasetService(CRUDService):
         Inheritable(Str('deduplication', enum=['ON', 'VERIFY', 'OFF'])),
         Inheritable(Str('checksum', enum=ZFS_CHECKSUM_CHOICES)),
         Inheritable(Str('readonly', enum=['ON', 'OFF'])),
-        Inheritable(Str('recordsize', enum=[
-            '512', '1K', '2K', '4K', '8K', '16K', '32K', '64K', '128K', '256K', '512K', '1024K',
-        ]), has_default=False),
+        Inheritable(Str('recordsize'), has_default=False),
         Inheritable(Str('casesensitivity', enum=['SENSITIVE', 'INSENSITIVE', 'MIXED']), has_default=False),
         Inheritable(Str('aclmode', enum=['PASSTHROUGH', 'RESTRICTED', 'DISCARD']), has_default=False),
         Inheritable(Str('acltype', enum=['OFF', 'NOACL', 'NFSV4', 'NFS4ACL', 'POSIX', 'POSIXACL']), has_default=False),
@@ -3518,6 +3516,9 @@ class PoolDatasetService(CRUDService):
                     f'{schema}.special_small_block_size',
                     'This field can be "INHERIT", 0 or multiple of 512, up to 1048576'
                 )
+            if rs := data.get('recordsize'):
+                if rs not in await self.middleware.call('pool.dataset.recordsize_choices'):
+                    verrors.add(f'{schema}.recordsize', '{rs!r} is an invalid recordsize.')
         elif data['type'] == 'VOLUME':
             if mode == 'CREATE' and 'volsize' not in data:
                 verrors.add(f'{schema}.volsize', 'This field is required for VOLUME')
--- a/src/middlewared/middlewared/plugins/pool_/dataset_recordsize.py
+++ b/src/middlewared/middlewared/plugins/pool_/dataset_recordsize.py
@@ -1,0 +1,36 @@
+from middlewared.schema import accepts, returns, List, Str
+from middlewared.service import Service
+
+
+class PoolDatasetService(Service):
+
+    class Config:
+        namespace = 'pool.dataset'
+
+    # https://openzfs.github.io/openzfs-docs/Performance%20and%20Tuning/Module%20Parameters.html#zfs-max-recordsize
+    # Maximum supported (at time of writing) is 16MB.
+    MAPPING = {
+        1 << 9: '512B',
+        1 << 10: '1K',
+        1 << 11: '2K',
+        1 << 12: '4K',
+        1 << 13: '8K',
+        1 << 14: '16K',
+        1 << 15: '32K',
+        1 << 16: '64K',
+        1 << 17: '128K',
+        1 << 18: '256K',
+        1 << 19: '512K',
+        1 << 20: '1M',
+        1 << 21: '2M',
+        1 << 22: '4M',
+        1 << 23: '8M',
+        1 << 24: '16M',
+    }
+
+    @accepts()
+    @returns(List(items=[Str('recordsize_value')]))
+    def recordsize_choices(self):
+        with open('/sys/module/zfs/parameters/zfs_max_recordsize') as f:
+            val = int(f.read().strip())
+            return [v for k, v in self.MAPPING.items() if k <= val]
--- a/src/middlewared/middlewared/pytest/unit/plugins/pool/test_recordsize_choices.py
+++ b/src/middlewared/middlewared/pytest/unit/plugins/pool/test_recordsize_choices.py
@@ -1,0 +1,10 @@
+import io
+from unittest.mock import patch
+
+from middlewared.plugins.pool_.dataset_recordsize import PoolDatasetService
+
+
+def test_recordsize_choices():
+    with patch("middlewared.plugins.pool_.dataset_recordsize.open") as mock:
+        mock.return_value = io.StringIO("32768\n")
+        assert PoolDatasetService(None).recordsize_choices() == ["512B", "1K", "2K", "4K", "8K", "16K", "32K"]


themylogin:
If this value can be changed on runtime, shouldn't we query it on each dataset operation and validate record size dynamically?

wolfspyre:
that might be beneficial, but i’d imagine it’s an infrequent enough adjustment that it’s reasonable to assume it won’t change more than a few times over a hosts’ lifetime.

I’d think it reasonable to assess at startup of middlewared and adopt a ‘if you change this tunable, kick middlewared to pick up the change’ mindset?

it’s also worth remembering that this could also impact replication (if the receiving host has a recordsize smaller than the sending dataset, things can get wonky)

[= The contents of this message have been written, read, processed, erased, sorted, sniffed, compressed, rewritten, misspelled, overcompensated, lost, found, and most importantly delivered entirely with recycled electrons =]

yocalebo:
> If this value can be changed on runtime, shouldn't we query it on each dataset operation and validate record size dynamically?

We're reading the module parameter on every update and create request or am I missing something?

src/middlewared/middlewared/plugins/pool_/dataset_recordsize.py
```diff
@@ -0,0 +1,31 @@
...
+        for raw_val, human_readable in MAPPING.items():
+            if int(raw_val) <= val:
+                choices.append(human_readable)
+        return choices
+
+
+# https://openzfs.github.io/openzfs-docs/Performance%20and%20Tuning/Module%20Parameters.html#zfs-max-recordsize
+# Maximum supported (at time of writing) is 16MB.
+MAPPING = {
+    f'{2 << 8}': '512B',
```
amotin, line 14: I am just curios why 2 << 8, not 1 << 9?

themylogin: > We're reading the module parameter on every update and create request or am I missing something?

Current code reads this parameter only once (on middleware startup)

yocalebo: Just because of "symmetry" I guess. Left side of << are 2 which looks nice. I can change to 1<< 9 🙂

yocalebo: > > We're reading the module parameter on every update and create request or am I missing something?
> 
> 
> 
> Current PR code reads this parameter only once (on middleware startup)

Then I have misunderstood how the schema variables are instantiated on each incoming request. My changes aren't correct so I'll make changes to bring them inside the method and check them there.

Ornias1993 (approved):
Very neat solution that was long due!

Testing setup also looks nice!

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3518,6 +3516,13 @@ async def __common_validation(self, verrors, schema, data, mode, parent=None, cu
                     f'{schema}.special_small_block_size',
                     'This field can be "INHERIT", 0 or multiple of 512, up to 1048576'
                 )
+            if rs := data.get('recordsize'):
+                choices = await self.middleware.call('pool.dataset.recordsize_choices')
+                if rs not in choices:
+                    verrors.add(
+                        f'{schema}.recordsize',
+                        '{rs!r} is an invalid recordsize. Valid choices are {", ".join(choices)}'
```
themylogin, line 3524: We never do this elsewhere `Valid choices are {", ".join(choices)}` and this list can be pretty big

src/middlewared/middlewared/plugins/pool_/dataset_recordsize.py
```diff
@@ -0,0 +1,38 @@
...
+class PoolDatasetService(Service):
+
+    class Config:
+        namespace = 'pool.dataset'
+
+    # https://openzfs.github.io/openzfs-docs/Performance%20and%20Tuning/Module%20Parameters.html#zfs-max-recordsize
+    # Maximum supported (at time of writing) is 16MB.
+    MAPPING = {
+        f'{1 << 9}': '512B',
+        f'{2 << 9}': '1K',
```
themylogin, line 13: Why `2 << 9` and not `1 << 10`?
Why the mapping keys are strings?

src/middlewared/middlewared/pytest/unit/plugins/pool/test_dataset_recordsize_choices.py
```diff
@@ -0,0 +1,66 @@
...
+        EXPECTED_RESULT[raw_val] = set((
+            '512B', '1K', '2K', '4K', '8K', '16K', '32K', '64K', '128K', '256K', '512K',
+            '1M', '2M', '4M', '8M', human_readable
+        ))
+
+
+@pytest.mark.parametrize('recordsize', list(MAPPING.keys()))
+def test_generate_choices_for_recordsize(recordsize):
+    # set the max recordsize first
+    with open('/sys/module/zfs/parameters/zfs_max_recordsize', 'w') as f:
```
themylogin, line 63: An unit test should never manipulate the OS 😃

You can mock `middlewared.plugins.pool_.dataset_recordsize.open` to return a `StringIO` that will contain the string you need.

src/middlewared/middlewared/pytest/unit/plugins/pool/test_dataset_recordsize_choices.py
```diff
@@ -0,0 +1,66 @@
...
+        EXPECTED_RESULT[raw_val] = set(('512B', '1K', '2K', '4K', '8K', human_readable))
+    elif human_readable == '32K':
+        EXPECTED_RESULT[raw_val] = set(('512B', '1K', '2K', '4K', '8K', '16K', human_readable))
+    elif human_readable == '64K':
+        EXPECTED_RESULT[raw_val] = set(('512B', '1K', '2K', '4K', '8K', '16K', '32K', human_readable))
+    elif human_readable == '128K':
+        EXPECTED_RESULT[raw_val] = set(('512B', '1K', '2K', '4K', '8K', '16K', '32K', '64K', human_readable))
+    elif human_readable == '256K':
+        EXPECTED_RESULT[raw_val] = set(('512B', '1K', '2K', '4K', '8K', '16K', '32K', '64K', '128K', human_readable))
+    elif human_readable == '512K':
```
themylogin, line 30: This is excessive. A function that maps `/sys/module/zfs/parameters/zfs_max_recordsize` contents to the list of the possible values would be smaller than this implementation + test 😃 One middle value (e.g. `512K` should be sufficient), it will make the entire function 100% covered.

themylogin (changes requested):
Thank you for adding tests!

src/middlewared/middlewared/plugins/pool.py
yocalebo: Okay, but why does that matter? How do you propose we give the consumer the valid available choices?

themylogin: Via `pool.dataset.recordsize_choices` which should be a documented public method

yocalebo: So making them read the API documentation when they given an invalid recordsize in the webUI is what you're suggesting?

src/middlewared/middlewared/pytest/unit/plugins/pool/test_dataset_recordsize_choices.py
yocalebo: Worrying about how "small" something is in code doesn't really make any logical sense. It accomplishes what it sets out to do just fine. Whether or not it uses 100 lines or 10 lines is irrelevant imo 😄

src/middlewared/middlewared/plugins/pool_/dataset_recordsize.py
yocalebo: Because I originally had it `2 << 8` and was asked the inverse of what you asked. I'm not changing it again. I will be using `2 <<`

src/middlewared/middlewared/plugins/pool.py
yocalebo: Furthermore, if you read the code you'll see that we're doing _exactly_ what I'm doing just a few lines above
```
3515                 verrors.add(
3516                     f'{schema}.special_small_block_size',
3517                     'This field can be "INHERIT", 0 or multiple of 512, up to 1048576'
3518                 )
```

yocalebo:
@themylogin Okay, I've rewritten this to try and paint the bike-shed everyone's favorite color. 😄

tests/api2/test_dataset_recordsize_choices.py
```diff
@@ -0,0 +1,14 @@
...
+import os
+sys.path.append(os.getcwd())
+
+from middlewared.test.integration.utils import call
+
+EXPECTED = set(['512B', '1K', '2K', '4K', '8K', '16K', '32K', '64K', '128K', '256K', '512K', '1M'])
+
+
+def test_pool_dataset_recordsize_choices():
+    with open('/sys/module/zfs/parameters/zfs_max_recordsize', 'w') as f:
```
themylogin, line 11: Now almost LGTM, but please fix this https://github.com/truenas/middleware/pull/8739#discussion_r845582528

themylogin (changes requested):

tests/api2/test_dataset_recordsize_choices.py
yocalebo: This isn't a unit test....

themylogin: Same issue: it will affect other tests executed in the same environment.

Either convert it back to unit test and mock `open` method, or move retrieving `zfs_max_recordsize` to a plugin method and mock it using our integration test mocking tools.

yocalebo: It's using ssh() to make the change

themylogin: So that machine's zfs module parameters would be changed, which will affect other tests executed on the same machine.

yocalebo: I'm tired of wasting my time so I've removed the tests.

themylogin: Should I add the test then?

yocalebo: sure

themylogin (approved):

PR #8740, opened 2022-04-07
Alert if the total number of snapshots is too large

themylogin:


yocalebo (approved):

src/middlewared/middlewared/plugins/snapshot.py
```diff
@@ -377,6 +377,17 @@ def max_count(self):
...
+    @accepts()
+    @returns(Int())
+    def max_total_count(self):
+        """
+        Returns a maximum amount of snapshots (total) the system can sustain.
+        """
+
+        # Having too many snapshots results in various performance complications (mainly, when listing them).
+        # This is a random round number that is large enough and does not cause issues in most use cases.
+        return 10000
```
rick-mesta, line 389: should this be a hard coded value like this ?

themylogin: @rick-mesta what are other options?

rick-mesta: A tuneable that doesn't require us to rebuild the software, perhaps ?

I'm just thinking that what works for one customer might not necessarily work for another. So this might be a way to provide relief in case 10k is a low ceiling for some. I don't know how these types of tuneables are available in TrueNAS.

themylogin: @william-gr what's our current position wrt tuneables like @rick-mesta proposes?

yocalebo:
If we make this a tunable, then why add it in the first place? :) If someone doesn't like the alert, they'll just tune it and ignore it.

rick-mesta:
Ugh... I didn't mean to start a bike-shed argument. If we think 10k is an appropriate hi-watermark, then let's just move on.

yocalebo:
> Ugh... I didn't mean to start a bike-shed argument. If we think 10k is an appropriate hi-watermark, then let's just move on.

But bike-shedding is fun! The 10k was the number that we've seen in the support field where when going above this, things start to get weird.

PR #8743, opened 2022-04-07
NAS-115628 / 22.12 / `iscsi.extent.disk_choices`: only query snapshots for zvols

themylogin:


themylogin:
time 05:00

src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -854,11 +854,13 @@ async def disk_choices(self):
             if key not in used_zvols:
                 diskchoices[key] = f'{zvol_name} ({zvol_size})'
 
-        zfs_snaps = await self.middleware.call('zfs.snapshot.query', [], {'select': ['name']})
+        zfs_snaps = await self.middleware.call(
+            'zfs.snapshot.query',
+            [['dataset', 'in', [zvol['name'] for zvol in zvols]]],
```
yocalebo, line 859: Why use `dataset` as the filter name when we're querying zvols? It makes more sense to me to have it named `zvol`.

themylogin: Because `dataset` is the key where `'zfs.snapshot.query'` puts dataset name, no matter if the snapshot is for filesystem of for a zvol.

yocalebo: Well sure but what I was suggesting was changing it so that `plugins/zfs.py` accepts `[pool, dataset, zvol]` and then just fills out `kwargs['datasets']` as necessary so we don't have to change `py-libzfs`

themylogin: Why? Our filters work the same way all over the middleware (`[field_name, "=", value]`), why change this specifically fro zfs snapshots?

yocalebo: How has what I suggested any different than doing `zfs.snapshot.query [['pool', '=', ['blah']]`?

themylogin: @yocalebo because we have `"pool"` field in `zfs.snapshot.query` output

yocalebo: sigh

PR #8760, opened 2022-04-11
NAS-115223 / 22.12 / `"silent": false` on service startup to fetch and raise an error if service startup failed

themylogin:


themylogin:
time 10:00

yocalebo (approved):

PR #8763, opened 2022-04-11
NAS-115025 / 22.12 / UPS Powerdown fixes

sonicaj:

1. Do not shutdown UPS if `powerdown` flag is unset. We were changing the filename earlier to `nokillpower` but that has no effect as `upsmon -K` just checks what `POWERDOWNFLAG` we have specified and if the flag file is there it shuts down.
2. Correct `nut` group in etc plugin so that ups configuration files have correct permissions.
3. Make sure we remove powerdown flag on each ups start.


src/middlewared/middlewared/etc_files/local/nut/ups_perms.py
```diff
@@ -20,6 +20,9 @@ def ups_config_perms(middleware):
         if ups_config['mode'].lower() != 'master':
             os.remove(file)
 
+    if os.path.exists(UPS_POWERDOWN_FLAG_FILE):
```
yocalebo, line 23: Why not pathlib.Path(UPS_POWERDOWN_FLAG_FILE).unlink(missing_ok=True)?

src/middlewared/middlewared/plugins/etc.py
```diff
@@ -29,7 +29,7 @@ class EtcGRP(enum.IntEnum):
     ROOT = 0
     SHADOW = 42
     NSLCD = 115
-    NUT = 126
+    NUT = 128
```
themylogin, line 32: Can this be read from `assets/account/builtin/linux/etc/group` instead of hardcode?

src/middlewared/middlewared/etc_files/local/nut/ups_perms.py
sonicaj: No reason, this does the job as well :D

src/middlewared/middlewared/plugins/ups.py
```diff
@@ -83,6 +84,10 @@ class Config:
         service_verb = 'restart'
         cli_namespace = 'service.ups'
 
+    @private
+    async def get_powerdown_flag_name(self):
```
themylogin, line 88: Why do we need this as a method?

themylogin (changes requested):

src/middlewared/middlewared/etc_files/local/nut/ups_perms.py
themylogin: This is prone to race condition (which will never occur, but still)

src/middlewared/middlewared/plugins/webdav.py
```diff
@@ -88,8 +87,8 @@ async def do_create(self, data):
         if data['perm']:
             await self.middleware.call('filesystem.chown', {
                 'path': data['path'],
-                'uid': EtcUSR.WEBDAV.value,
-                'gid': EtcGRP.WEBDAV.value,
+                'uid': await self.middleware.call('webdav.get_webdav_user_id'),
```
themylogin, line 90: Can we use `etc` method here?

src/middlewared/middlewared/plugins/etc.py
```diff
@@ -510,6 +483,16 @@ async def generate_checkpoint(self, checkpoint):
     async def get_checkpoints(self):
         return self.checkpoints
 
+    async def get_user_id(self, username):
+        if not self.SYS_USERS:
```
themylogin, line 487: I have a few proposals:
1. Move these methods to `user`/`group` plugins since they can be used by other plugins (e.g. webdav)
2. Name methods like `get_internal_user_id` to avoid confusion (someone might try to use them to retrieve normal user ids)
3. Warm up this cache on `user`/`group` plugin initialization because it's gonna be used anyway

src/middlewared/middlewared/plugins/etc.py
```diff
@@ -391,27 +368,23 @@ async def gather_ctx(self, methods):
...
-            gid = grp.getgrnam(entry["group"]).gr_gid
-
+        uid = self.middleware.call_sync('etc.get_user_id', user_name)
+        gid = self.middleware.call_sync('etc.get_group_id', group_name)
         st = os.fstat(fd)
         uid_to_set = -1
         gid_to_set = -1
 
-        if uid != -1 and st.st_uid != uid:
+        if uid is not None and st.st_uid != uid:
```
themylogin, line 384: How can it be `None`? Don't internal users always exist?

themylogin (changes requested):

src/middlewared/middlewared/plugins/etc.py
yocalebo: `get_service_uid` / `get_service_gid` maybe?

src/middlewared/middlewared/plugins/etc.py
```diff
@@ -91,6 +72,9 @@ async def render(self, path, ctx):
 
 class EtcService(Service):
 
+    SYS_GROUPS = {}
```
yocalebo, line 75: `SYS_GROUPS/USERS` variables aren't need here since they're being populated in the `internal_cache.py` file right?

sonicaj: Ah shoot, yup :p

src/middlewared/middlewared/etc_files/local/apache24/Includes/webdav.conf.mako
```diff
@@ -4,7 +4,11 @@
     import stat
 
     from contextlib import suppress
-    from middlewared.plugins.etc import EtcUSR, EtcGRP
+
+    from middlewared.plugins.webdav import WEBDAV_USER
+
+    webdav_uid = middleware.call_sync('user.get_internal_user_id', WEBDAV_USER)
```
themylogin, line 10: Sorry, I mean `builtin` instead of `internal` (that's the term we already use)

themylogin (approved):
Thank you!

PR #8767, opened 2022-04-12
DOCS-3526 / 22.02.1 / Add a sleep of 5 second after garbing the device in iscsi test 35

ericbsd:
That make sure that the device is ready to be formatted and mounted

It is a back-port of 13.0-stable test

### DIFF
--- a/tests/api2/test_260_iscsi.py
+++ b/tests/api2/test_260_iscsi.py
@@ -163,7 +163,8 @@ def test_10_Waiting_for_iscsi_connection_before_grabbing_device_name(request):
             file_device_name = iscsictl_list[3]
             assert True
             break
-        sleep(3)
+        sleep(1)
+    sleep(5)
 
 
 @bsd_host_cfg
@@ -369,6 +370,7 @@ def test_35_waiting_for_iscsi_connection_before_grabbing_device_name(request):
             assert True
             break
         sleep(1)
+    sleep(5)
 
 
 @bsd_host_cfg
@@ -469,6 +471,7 @@ def test_48_waiting_for_iscsi_connection_before_grabbing_device_name(request):
             assert True
             break
         sleep(1)
+    sleep(5)
 
 
 @bsd_host_cfg


Mrt134 (approved):

yocalebo:
This seems wrong. You're already sitting in a `while True:` loop and sleeping 1 second there and you're sleeping there forever until the iscsi service is connected? Why add another `sleep(5)` outside of that loop?

ericbsd:
> This seems wrong. You're already sitting in a `while True:` loop and sleeping 1 second there and you're sleeping there forever until the iscsi service is connected? Why add another `sleep(5)` outside of that loop?

@yocalebo sometimes the device is not ready for the format and mount. It occurs when tnbuilds03 is under load I had to do the same thing on 13.0.

There are 30 VMs running on  tnbuilds03 plus API and some UI tests so that 5 seconds of sleep prevent this https://jira.ixsystems.com/browse/DOCS-3526.

yocalebo:
> > This seems wrong. You're already sitting in a `while True:` loop and sleeping 1 second there and you're sleeping there forever until the iscsi service is connected? Why add another `sleep(5)` outside of that loop?
> 
> @yocalebo sometimes the device is not ready for the format and mount. It occurs when tnbuilds03 is under load I had to do the same thing on 13.0.
> 
> There are 30 VMs running on tnbuilds03 plus API and some UI tests so that 5 seconds of sleep prevent this https://jira.ixsystems.com/browse/DOCS-3526.

That shouldn't have anything to do with whether or not you can connect to an iscsi port on a VM. I suspect that when the `{ip}:3260` shows up in the `iscsictl -L` output, the service isn't responding immediately when the next api test runs. You could get rid of the potentially infinite loop and just use base python to try to open a connection to the remote port?
    
    

> > This seems wrong. You're already sitting in a `while True:` loop and sleeping 1 second there and you're sleeping there forever until the iscsi service is connected? Why add another `sleep(5)` outside of that loop?
> 
> @yocalebo sometimes the device is not ready for the format and mount. It occurs when tnbuilds03 is under load I had to do the same thing on 13.0.
> 
> There are 30 VMs running on tnbuilds03 plus API and some UI tests so that 5 seconds of sleep prevent this https://jira.ixsystems.com/browse/DOCS-3526.

Hmm, the DOCS ticket you link doesn't show why the `newfs`  command is actually failing? Also, why is this PR here on `stable/angelfish`? Commands like `newfs` and `iscsictl` don't exist on scale?

ericbsd:
> That shouldn't have anything to do with whether or not you can connect to an iscsi port on a VM. I suspect that when the `{ip}:3260` shows up in the `iscsictl -L` output, the service isn't responding immediately when the next api test runs. You could get rid of the potentially infinite loop and just use base python to try to open a connection to the remote port?

There is a timeout of 15 seconds with `@pytest.mark.timeout(15)` for that test so the loop is not a problem. 

What module would work on tnbuilds03? 

> Hmm, the DOCS ticket you link doesn't show why the `newfs` command is actually failing? Also, why is this PR here on `stable/angelfish`? Commands like `newfs` and `iscsictl` don't exist on scale?

Those tests do not run inside of SCALE they run on tnbuilds03 which uses `newfs` and `iscsictl`.

yocalebo:
I see, thanks for clarifying. So, any function decorated with `@bsd_host_cfg` will only get run if it's on the freebsd host. Anyways, why not just use base python to try and open a remote socket to the `{ip}:3260`? Also are we capturing the returncode or error message for why `newfs` failing?

ericbsd:
When I converted the test from old shell script I did keep it close to what it was at the time. I didn't looked at any other solutions.


As for the newfs I could add the stderr output, but when using pytest with -vs i was seeing that /dev/da* was not found. Adding a sleep of 3 did fixed the issue I did change it for 5 just to be on the safe side.

All that I could probably wait for /dev/da* to exist.

PR #8771, opened 2022-04-12
DOCS-3530 / 22.12 / Fixed api2.test_system_dataset.test_system_dataset_migrate error

ericbsd:
/var/log/middlewared.log is not on the host of the test.


themylogin (approved):
Always forget about this, thank you!

PR #8772, opened 2022-04-12
DOCS-3530 / 22.02.01 / Fixed api2.test_system_dataset.test_system_dataset_migrate error

ericbsd:
/var/log/middlewared.log is not on the host of the test.


ericbsd:
backported from #8771

Mrt134 (approved):

PR #8778, opened 2022-04-13
NAS-115771 / 13.0 / Add oneshot alert for deprecated services that are running

anodos325:
Currently this only contains AFP in 13, but having general-purpose
alert may be useful so that users can plan migration off things
that we will remove in the future.


themylogin (commented):
Does it work if AFP service is started by FreeBSD rc on boot? IIRC `service.start` is not called there.

yocalebo (approved):

yocalebo:
Merge after 13.0-RELEASE is shipped.

themylogin (approved):

PR #8781, opened 2022-04-13
NAS-114942 / 22.12 / S3: Use credentials region if regions are supported but task region is empty

themylogin:


themylogin:
time 5:00

src/middlewared/middlewared/test/integration/utils/mock_binary.py
```diff
@@ -0,0 +1,57 @@
...
+import textwrap
+
+from .call import call
+from .ssh import ssh
+
+
+class BinaryMock:
+    def _load(self):
+        try:
+            return json.loads(ssh("cat /tmp/mocked_binary_launch", check=False).strip())
```
yocalebo, line 14: Should we statically define the `/tmp/mocked_binary_launch` path since you're using it in multiple places?

yocalebo (approved):

PR #8784, opened 2022-04-14
NAS-115783 / 22.12 / Generated dhclient.conf files are broken

dberlin:
The bug contains more info, but the generated dhclient.conf files contain errors, and it's only by luck that dhclient ignores the errors rather than failing.
This patch fixes the missing semicolon for the send line, and the classless option that doesn't exist in dhclient.

At a glance, the supersede statement may be wrong as well, but i'm not familiar enough with dhclient.conf errors.


dberlin:
https://github.com/truenas/middleware/pull/8785 fixes this in a better way, so closing this

PR #8800, opened 2022-04-19
NAS-115848 / 23.10 / FreeIPA groups do not work

dberlin:
FreeIPA does not have a uniqueMember attribute by default.
(in fact, you have to generate them if you want to integrate vsphere and FreeIPA, for example)

As a result, the default configuration of "map group member uniqueMember" causes users to lose all their groups :)
There is no need to do this mapping, as the default group mapping is correct.

I have given an example before/after for a user on the Jira bugreport

### DIFF
--- a/src/middlewared/middlewared/plugins/ldap.py
+++ b/src/middlewared/middlewared/plugins/ldap.py
@@ -876,7 +876,6 @@ class LDAPService(TDBWrapConfigService):
 
             default_naming_context = rootdse[0]['data']['defaultnamingcontext'][0]
             aux_params = [
-                'map group member uniqueMember',
                 f'base passwd cn=users,cn=accounts,{default_naming_context}',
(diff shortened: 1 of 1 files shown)


anodos325:
IIRC when I was originally testing this on a fresh FreeIPA deployment UniqueMember was present on groups and without filter, group entries would appear twice in `getent group` output.

That said, this auto-configuration only happens once and is a convenience feature. Admin can easily change / comment out if needed.

dberlin:
Maybe you are thinking of another directory system?
UniqueMember hasn't been present in freeipa for at least half a decade  (IE since FreeIPA 3.x) .
I can't speak to before then.  Happy to ask their mailing list if it's helpful.

You have to add generation of uniqueMember as a compat feature to make vsphere work, as an example:
https://www.freeipa.org/page/HowTo/vsphere5_integration
(You can see this applies to freeipa 3.x+)

PR #8801, opened 2022-04-19
NAS-115849 / 13.0 / Hook improvements

yocalebo:
Optimizing middlewared service on a system with ~1.2k hard drives has exposed some fundamental design flaws. 1 of those is the fact of how we're using hooks. The overly simplified problem is that we have hooks being called which is creating async tasks and each task (potentially) could be calling the same method. In some scenarios we're actually calling methods 2, 3 and sometimes 4+ times. This is painful normally but excruciating on larger systems. To make matters worse, we've designed hooks to be called based off of OS devd events. This means we could get a bunch of devd events (triggered by an API endpoint) that call specific methods and then in the API endpoint that triggered the devd events we go ahead and explicitly call a hook that does the same thing as what our devd hooks do.....

This is the best idea that I could come up with and that is to add the ability to "block" (disable) hooks temporarily. The idea is that in some API endpoints you would disable conflicting hooks, run the necessary logic, then re-enable the hooks that you disabled. Testing this on a very large system has shown that adding a single vdev to a > 200 drive zpool cuts the time down from like ~12mins to ~10 seconds.

Fixes implemented:
1. add the ability to "block" (disable) hooks temporarily
2. fix a `RuntimeWarning` that showed up when I was QA'ing these changes on a live system

### DIFF
--- a/src/middlewared/middlewared/main.py
+++ b/src/middlewared/middlewared/main.py
@@ -1073,7 +1073,12 @@ class Middleware(LoadPluginsMixin, RunInThreadMixin, ServiceCallMixin):
         return self.__hooks
 
     def update_hook(self, hook_name, block):
-        ignore = ('core.on_connect', 'ha_permission')  # if you block these, you're going to have a bad time
+        # if you block these, you're going to have a bad time
+        ignore = (
+            'core.on_connect',
+            'ha_permission',
+            'datastore.post_execute_write',  # for HA systems
+        )
 
         if hook_name == 'all':
             for rname, rhooks in filter(lambda x: x[0] not in ignore, self.__hooks.copy().items()):
--- a/src/middlewared/middlewared/main.py
+++ b/src/middlewared/middlewared/main.py
@@ -1066,7 +1066,22 @@ class Middleware(LoadPluginsMixin, RunInThreadMixin, ServiceCallMixin):
             'method': method,
             'inline': inline,
             'sync': sync,
+            'block': False,
         })
+
+    def get_hooks(self):
+        return self.__hooks
+
+    def update_hook(self, hook_name, block):
+        ignore = ('core.on_connect', 'ha_permission')  # if you block these, you're going to have a bad time
+
+        if hook_name == 'all':
+            for rname, rhooks in filter(lambda x: x[0] not in ignore, self.__hooks.copy().items()):
+                for idx, _ in enumerate(rhooks):
+                    self.__hooks[rname][idx]['block'] = block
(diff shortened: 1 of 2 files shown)


themylogin:
This is a nice improvement, following your work I was going to propose a similar (and, IMO, safer) model.

I wanted to add `batch_hook_calls` context manager to use it like this:

```
with batch_hook_calls("pool.post_create_or_update"):
    for disk in ["sda", "sdb", "sdc", "sdd"]:
        for pool in ["tank", "test"]:
            self.middleware.call_hook("pool.post_create_or_update", pool)
```

which will group equal hook calls during its lifetime and execute them once upon exit, so when we know a batch operation will occur, we can run post-processing hooks once. This will work basically the same way as your implementation: temporarily block hooks, then unblock them. Please let me know what do you think of this idea.

Even if we decide not do implement this hook grouping, a context manager will still be useful for disabling hooks.

I also propose to make hooks explicitly blockable/groupable and non-blockable/non-groupable by default to avoid accidental mistakes.

yocalebo:
> which will group equal hook calls during its lifetime and execute them once upon exit, so when we know a batch operation will occur, we can run post-processing hooks once. This will work basically the same way as your implementation: temporarily block hooks, then unblock them. Please let me know what do you think of this idea.
> 
I think that's a great idea.

> Even if we decide not do implement this hook grouping, a context manager will still be useful for disabling hooks.
> 
I agree with this too, a context manager would definitely be useful

> I also propose to make hooks explicitly blockable/groupable and non-blockable/non-groupable by default to avoid accidental mistakes.
>
I broke middleware in an extravagant way testing these changes so I agree with this as well 😄 


Do you want to merge this one, and then you can open a separate PR to make changes to my code as well as add the context/groupable changes in another PR?

themylogin:
@yocalebo do you have another changeset that uses this new hook block method already?

yocalebo:
> @yocalebo do you have another changeset that uses this new hook block method already?

I've got a local change that uses these changes but haven't pushed it yet.

themylogin:
@yocalebo my changes will break your changes, can you push them too so I can adapt them and test them?

themylogin:
I'll also do my changes against `master` branch to be able to add tests

yocalebo:
> I'll also do my changes against `master` branch to be able to add tests

Push your PR and I'll close this one. I'm still working on a few other PRs at the moment so no big deal.

yocalebo:
Will open fresh PR (if needed) after @themylogin pushes his changes.

PR #8816, opened 2022-04-22
NAS-115840 / 22.12 / Blacklist irdma driver.

amotin:
This driver tends to allocate 160-210MB of RAM per Intel NIC port,
that overflows CMA region we need for ntb_transport and can't just
increase to few GB considering small systems.  Wasting GBs of RAM
for smaller systems with good NICs is also not good.  Since we
currently do not use any RDMA services, blocking this driver seems
to be the obvious choice.

Ticket:	NAS-115840


yocalebo:
backport

PR #8817, opened 2022-04-22
NAS-115904 / 22.12 / Update system keytab using net ads keytab create rather than trying to change trust pw

dberlin:


dberlin:
Attempting to use net ads changetrustpw to force a keytab update consistently fails on all of my machines.
The documented way to force a keytab update is to use net ads keytab create, which consistently works on all of my machines :)

anodos325:
I'll take a look at this maybe this next week.

anodos325:
Looks like CLI options have expanded since I last looked at this. keytab create and forcing password change are both wrong. There's net ads method to add SPN to system keytab. Will look into doing that as well. IIRC there is also work that needs to be done on winbindd to have the scheduled machine account password changes update system keytab and so may need to roll that into this.

dberlin:
I must have missed that method (i really wish net  sorted the command names by name)
Oh i see, i think you mean net ads add_update_ads.

Yeah, that would work and eliminate the need to update the keytab this way at all.
If you are going to poke at it, i can drop the pull request.
If not, I also am happy to revamp how the NFS spns are added to use net ads add_update_ads.
Let me know.

anodos325:
We don't use SSSD (which is what Fedora uses for realmd / adcli stuff). I can look into it, but it will take me a week or so to circle around to this. Too many irons in the fire.

dberlin:
No worries. I can try to look into it and see what i find.   It can certainly be worked around manually by editing the DNS names after the fact and regen'ing the keytab, so it's not horrible, just curious to see if it's possible to make it happen at join time. 

I know y'all don't use adcli/realmd/etc, which i assume is because you have to run the samba server anyway for shares, so it doesn't make sense to introduce more dependencies/ways of doing it if you can avoid it. I have to imagine trying to keep adcli/realmd/samba in sync about the state of the world would ... be a mess.  SSSD itself is not too bad, but yeah.

anodos325:
> No worries. I can try to look into it and see what i find. It can certainly be worked around manually by editing the DNS names after the fact and regen'ing the keytab, so it's not horrible, just curious to see if it's possible to make it happen at join time.
> 
> I know y'all don't use adcli/realmd/etc, which i assume is because you have to run the samba server anyway for shares, so it doesn't make sense to introduce more dependencies/ways of doing it if you can avoid it. I have to imagine trying to keep adcli/realmd/samba in sync about the state of the world would ... be a mess. SSSD itself is not too bad, but yeah.

It's generally problematic because a commonly used idmapping scheme does not exist in winbindd. You really want your IDs on clients and servers to match.

anodos325:
Ended up going this route. https://github.com/truenas/middleware/pull/9058

PR #8821, opened 2022-04-23
Use `secrets` instead of `random`

oittaa:
Since Python 3.6 `secrets` has been the recommended module for creation of random values that might be security critical. https://docs.python.org/3/library/secrets.html

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -14,7 +14,7 @@ import errno
 import glob
 import hashlib
 import os
-import random
+import secrets
 import shlex
 import shutil
 import string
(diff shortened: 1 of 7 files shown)


oittaa:
After you've tested this, I would highly recommend backporting this since those `random` generated auth tokens aren't that difficult to guess if you know an approximate time of creation.

yocalebo:
> After you've tested this, I would highly recommend backporting this since those `random` generated auth tokens aren't that difficult to guess if you know an approximate time of creation.

@oittaa thanks again for pointing out the obvious and even offering a patch. If you don't mind, I'm going to open up a ticket for this problem and I'd like to push a PR that does essentially what you've done, however, I'd like to make a few general use functions that can be imported and used everywhere instead of doing `import secrets` in each module. We actually already have a general use function in `middlewared.utils.generate` but it's also using `random` module. 😄

oittaa:
Sounds good to me!

yocalebo:
@oittaa thanks again for the contribution. I'm going to close this in favor of https://github.com/truenas/middleware/pull/8829

PR #8822, opened 2022-04-23
NAS-115915 / 22.12 / Reduce the vulnerability to timing attacks

oittaa:
It is recommended to use the [compare_digest()](https://docs.python.org/3/library/hmac.html#hmac.compare_digest) function instead of the == operator to reduce the vulnerability to timing attacks.

https://docs.python.org/3/library/crypt.html#examples


yocalebo:
Thanks again for the patch! The documentation/examples that you linked make it very clear 😄

yocalebo (approved):

PR #8823, opened 2022-04-23
NAS-115913 / 22.12 / Secure temporary dir with `generate_ssh_key_pair`

oittaa:
https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

Btw. Is there a reason for not using some more modern key type like `ed25519` instead of `rsa`?

### DIFF
--- a/src/middlewared/middlewared/plugins/keychain.py
+++ b/src/middlewared/middlewared/plugins/keychain.py
@@ -2,7 +2,6 @@ import base64
 import enum
 import errno
 import os
-import random
 import re
 import string
 import subprocess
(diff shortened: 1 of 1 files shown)


yocalebo:
Thanks for the code clean-up!

> Btw. Is there a reason for not using some more modern key type like `ed25519` instead of `rsa`?
There probably isn't any technical reason why we're not using a more modern key, but since you bring it up can you please open a ticket at jira.ixsystems.com so we can keep it on our radar?

yocalebo (approved):

oittaa:
> There probably isn't any technical reason why we're not using a more modern key, but since you bring it up can you please open a ticket at jira.ixsystems.com so we can keep it on our radar?

Done: https://jira.ixsystems.com/browse/NAS-115914

PR #8828, opened 2022-04-23
Password storage

oittaa:
A single round of SHA-512 is not an acceptable way to store passwords in 2022. Python has supported bcrypt in `crypt` calls since version 3.7, so it can be used with minimal changes.

https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html
https://docs.python.org/3/library/crypt.html#crypt.METHOD_BLOWFISH

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/apache24/webdav_config.py
+++ b/src/middlewared/middlewared/etc_files/local/apache24/webdav_config.py
@@ -1,22 +1,17 @@
 import re
 import os
-import secrets
 import hashlib
-import crypt
 
 from contextlib import suppress
 from string import digits, ascii_uppercase, ascii_lowercase
 
+from middlewared.plugins.account import crypted_password
 from middlewared.plugins.webdav import WEBDAV_USER
 
 
 def generate_webdav_auth(middleware, render_ctx, dirfd):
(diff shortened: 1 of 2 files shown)


oittaa:
Another option might be to use `Argon2id`. I don't know if you're using the native Debian/FreeBSD login tools somewhere or if everything is done through these Python interfaces. If everything is done through Python, then you could use Argon2id from `passlib`, which is already included. 

```python
from passlib.hash import argon2
argon2.hash("password")
```
At least on my computer it seems to use reasonable defaults.
```
'$argon2id$v=19$m=65536,t=3,p=4$zjkH4BzDuHfOec9Zy5mzlg$k9JR6r0rCwksYBdeZ9ntsgnvXA+oaAFNziP4BKcX57U'
```

https://passlib.readthedocs.io/en/stable/lib/passlib.hash.argon2.html

But at the moment `bcrypt` appears to be good enough in foreseeable future, and it might be easier to just use it since basically every operating system supports it natively.

yocalebo:
Ticket for this is here: https://jira.ixsystems.com/browse/NAS-115917. Will close this ticket for now. Thanks again for the PR's and suggestions!

PR #8833, opened 2022-04-25
NAS-115931 / 13.0 / use defaultdict in geom.cache plugin

yocalebo:
Memory usage in python is typically a non-issue....until it is 😄 On a large system with ~1240 hard drives, the difference between using the standard `dict()` compared to `defaultdict()` is measurable. There is no functional change which is just icing on the cake.

### DIFF
--- a/src/middlewared/middlewared/plugins/geom.py
+++ b/src/middlewared/middlewared/plugins/geom.py
@@ -2,6 +2,7 @@ import re
 import threading
 from xml.etree import ElementTree as etree
 from itertools import zip_longest
+from collections import defaultdict
 
 import sysctl
 from middlewared.service import Service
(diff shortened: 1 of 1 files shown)


themylogin:
Why does `defaultdict` used in exactly the same way consumes less memory than standard dict? Can this be a bug in python worth reporting.

yocalebo:
> Why does `defaultdict` used in exactly the same way consumes less memory than standard dict? Can this be a bug in python worth reporting.

I'm not sure honestly. With 100% reproducibility, however, using a `defaultdict()` uses 13MB less of resident memory compared to a `dict()`. I haven't dug into the internals on why this is occurring.

william-gr (approved):

PR #8842, opened 2022-04-27
NAS-114952 / 13.0 / Properly initialize service announcements on boot (#7429)

anodos325:
Wait until system ready before starting nmbd, wsdd, avahi.

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -156,6 +156,23 @@ class NetworkConfigurationService(ConfigService):
 
         return verrors
 
+    @private
+    async def toggle_announcement(self, data):
+        for srv, enabled in data.items():
+            service_name = ANNOUNCE_SRV[srv]
+            started = await self.middleware.call('service.started', service_name)
+            verb = None
+
+            if enabled:
+                verb = 'restart' if started else 'start'
+            else:
+                verb = 'stop' if started else None
+
+            if not verb:
+                continue
+
+            await self.middleware.call(f'service.{verb}', service_name)
+
     @accepts(
         Dict(
             'global_configuration_update',
@@ -209,6 +226,9 @@ class NetworkConfigurationService(ConfigService):
         new_config = config.copy()
         is_ha = True
         sync_group_mappings = False
+        srv = config['service_announcement'] | data.get('service_announcement', {})
+        new_config.update(data)
+        new_config['service_announcement'] = srv
 
         if not (
                 not await self.middleware.call('system.is_freenas') and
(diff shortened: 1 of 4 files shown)


rick-mesta (approved):
LGTM, but I'd appreciate an extra pair of eyes from @themylogin and/or @sonicaj

src/middlewared/middlewared/plugins/network.py
```diff
@@ -209,6 +226,9 @@ async def do_update(self, data):
         new_config = config.copy()
         is_ha = True
         sync_group_mappings = False
+        srv = config['service_announcement'] | data.get('service_announcement', {})
```
sonicaj, line 229: How would i got about removing a service announcement which i had enabled before ?

anodos325:
```
>>> d1 = {"netbios": True, "mdns": False}
>>> d2 = {"netbios": False, "mdns": False}
>>> d1 | d2
{'netbios': False, 'mdns': False}
```
Maybe not understanding question

themylogin: @anodos325 how does this behave when `service_announcement` is declared at L185 like
```
            Dict(
                'service_announcement',
                Bool('netbios', default=False),
                Bool('mdns', default=True),
                Bool('wsd', default=True),
            ),
```
?

anodos325: Well, goal there was to make sure that dict was normalized before being converted to JSON and written to DB. I'll switch normalization to inside update() method so that we go <defaults> | <what's on disk> | <payload>.

src/middlewared/middlewared/plugins/network.py
```diff
@@ -210,6 +227,14 @@ async def do_update(self, data):
         is_ha = True
         sync_group_mappings = False
 
+        # Apply on-disk settings and user provided ones to our
+        # defaults to ensure that the JSON stored in db
```
sonicaj, line 231: If there is potential that the db isn't normalized, we should be handling that via a migration ?

src/middlewared/middlewared/plugins/network.py
sonicaj: Thank you for clarifying @anodos325 :P I was looking at it the wrong way :D

src/middlewared/middlewared/plugins/network.py
anodos325: Well, the JSON is eventually written as a string in the db which means that it's not quite as robust as if things were properly typed. So me being a bit paranoid. We can change this in bluefin, but I'd rather not in 13. (This is backport from SCALE BTW).

william-gr (approved):
We are behind schedule here, and I was told this code is already in SCALE and working.

We can make the improvements in SCALE only unless anyone seems some major issue here (DM me on Slack please)

PR #8849, opened 2022-04-27
NAS-115541 / 22.12 / Correctly handle network interface `description` and make sure it is …

themylogin:
…never null


themylogin:
@yocalebo good enough candidate for a 22.02.1 backport? Fixes CLI crash when trying to set interface description for an interface that is not in the database yet.

themylogin:
time 1:30

rick-mesta (approved):

PR #8863, opened 2022-04-29
NAS-115976 / 22.12 / Add tests for renames over SMB protocol

anodos325:
We have had regressions here on a few occassions. Let's
add some CI tests so that it doesn't get missed again.


rick-mesta (approved):

ericbsd (approved):
LGTM

PR #8867, opened 2022-05-02
NAS-115927 / 22.12 / fix regression in interface.update

yocalebo:
When the `network_alias` and `network_interface` tables were fixed because of other unrelated bugs, I introduced a regression where on `interface.update` the 1st IP address of an interface cannot be removed.

The fix is trivial but all it does is make sure `convert_aliases_to_datastore` always returns a dict with default keys/values.


yocalebo:
retest this please

sonicaj (approved):

PR #8878, opened 2022-05-05
NAS-116049 / 13.0 / add release-date to dmi cache

yocalebo:
Adding this information to what we cache because we're `subprocess`'ing (unnecessarily) in an alert to get this information.

### DIFF
--- a/src/middlewared/middlewared/plugins/system_/dmi.py
+++ b/src/middlewared/middlewared/plugins/system_/dmi.py
@@ -1,4 +1,5 @@
 import subprocess
+from datetime import datetime
 
 from middlewared.service import private, Service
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/system_/dmi.py
```diff
@@ -46,7 +49,9 @@ def _parse_dmi(self, lines):
                 continue
 
             sect, val = [i.strip() for i in line.split(':', 1)]
-            if sect == 'Manufacturer':
+            if sect == 'Release Date':
+                SystemService.CACHE['release-date'] = val
```
themylogin, line 53: Can we parse it into `date` instance here to avoid doing that in every consumer?

themylogin (changes requested):

src/middlewared/middlewared/plugins/system_/dmi.py
```diff
@@ -7,6 +8,8 @@ class SystemService(Service):
     # DMI information is mostly static so cache it
     HAS_CACHE = False
     CACHE = {
+        'bios-release-date-raw': '',
+        'bios-release-date-formatted': '',
```
themylogin, line 12: It was `formatted` (in an unknown format) before you parsed it. I would say now it is `parsed` (as ZFS properties), isn't it?
And will we ever need this date as raw?

themylogin (approved):
Now that we don't have `raw` version, do we need `parsed` suffix at all?

PR #8879, opened 2022-05-05
NAS-116053 / 13.0 / cache truenas.get_chassis_hardware

yocalebo:
This method is called by `system.is_enterprise` which, surprisingly, is called by many other methods. This is information based off dmidecode information (which is burned into BIOS) so cache it for a micro-optimization.

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/truenas.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/truenas.py
@@ -5,6 +5,7 @@ import os
 
 from middlewared.schema import accepts, Bool, Dict, Str
 from middlewared.service import job, private, Service
+from middlewared.utils.functools import cache
 import middlewared.sqlalchemy as sa
 
 EULA_FILE = '/usr/local/share/truenas/eula.html'
@@ -39,6 +40,7 @@ class TruenasCustomerInformationModel(sa.Model):
 class TrueNASService(Service):
 
     @accepts()
+    @cache
     async def get_chassis_hardware(self):
         """
         Returns what type of hardware this is, detected from dmidecode.


src/freenas/usr/local/lib/middlewared_truenas/plugins/truenas.py
```diff
@@ -80,35 +82,36 @@ async def get_chassis_hardware(self):
         Really NFI about hardware at this point.  TrueNAS on a Dell?
         TRUENAS-UNKNOWN
         """
+        if TrueNASService.CHASSIS is not None:
```
themylogin, line 85: Can we just use `@functools.cache` instead of this?

themylogin (changes requested):

src/freenas/usr/local/lib/middlewared_truenas/plugins/truenas.py
yocalebo: It doesn't work with our `@accepts()` decorator no matter how you stack them.

yocalebo: It doesn't work with our `@accepts()` decorator no matter how you stack them.

themylogin: @yocalebo can you try using https://github.com/truenas/middleware/pull/8883 please?

themylogin:
```
    @accepts()
    @returns(Str('system_chassis_hardware'))
    @cli_private
    @cache
    async def get_chassis_hardware(self):
```

yocalebo: nice

themylogin (approved):
Thank you!

PR #8881, opened 2022-05-05
NAS-116063 / 13.0 / optimize system_is_enterprise_ix_hardware calls in smartctl

yocalebo:
We're calling `system.is_enterprise_ix_hardware` for every disk. This isn't really that big of a deal until you're on a system with 1.2k+ disks. Call this method once in the caller and pass to the callees appropriately.


themylogin (changes requested):
Let's group `middleware, devices, enterprise_hardware` into one `SmartContext` namedtuple to avoid changing argument list every time we need a new thing cached

themylogin (approved):

PR #8899, opened 2022-05-09
NAS-116118 / 22.12 / include pool name and GUID in lsblk output

anodos325:
Include UUID which is the zpool GUID and label which is zpool
name.

### DIFF
--- a/src/freenas/usr/local/libexec/freenas-debug/zfs/zfs.sh
+++ b/src/freenas/usr/local/libexec/freenas-debug/zfs/zfs.sh
@@ -141,8 +141,8 @@ zfs_func()
 	done
 	section_footer
 
-	section_header "lsblk -o name,partuuid -l"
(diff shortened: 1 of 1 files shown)


sonicaj (approved):
Maybe add `| jq` as well to pretty print the json ?

anodos325:
I'll adjust the ZFS lsblk output actually. The main thing I was concerned about getting zpool names / guid added to list. I'll modify PR accordingly.

anodos325:
> Maybe add `| jq` as well to pretty print the json ?

lsblk already somewhat formats the JSON. I think it's more readable in standard way of printing.

yocalebo:
exclude zvols? `-e 230`?

anodos325:
I switched back to list view because it's simpler to read. Maybe later I'll write a function to parse the JSON with jq and have additional output grouping disks by pool.

PR #8900, opened 2022-05-09
NAS-115885 / 22.12 / Allow to restrict UI/API access to certain IPs/networks

themylogin:
time 07:30


themylogin:
> I don't think this will work with DHCP.

why?

themylogin:
> Are we validating against ui_address_choices for these? I think we probably need to protect users from configurations that may result in them locking themselves out of their NAS.

How can we validate this? If I run on 0.0.0.0 and allow access only from 192.168.0.3, is this a valid configuration?

themylogin:
I agree we might need a method to rollback this value (also other Web UI related values) after a certain amount of time has passed, similar to what we did for network interfaces configuration, can I address this in a new separate PR?

themylogin:
@yocalebo these are client addresses, not listen addresses, listen addresses are left intact by this PR.

yocalebo:
> @yocalebo these are client addresses, not listen addresses, listen addresses are left intact by this PR.

derp, thanks 😄

anodos325:
> > Are we validating against ui_address_choices for these? I think we probably need to protect users from configurations that may result in them locking themselves out of their NAS.
> 
> How can we validate this? If I run on 0.0.0.0 and allow access only from 192.168.0.3, is this a valid configuration?

I obviously was having a bad case of the Mondays. Yes, this makes sense. We'll need to make sure there's a rollback period like with network configuration changes.

anodos325:
Do you have a migration where you add this new parameter?

anodos325:
Docstring for new key should state that restarting http service is required in order for setting to take effect.

themylogin:
@anodos325 this is not a new behavior, other UI fields behave the same way. I'll write a better documentation in a PR that allow automatic rollback of UI settings that will take in account the possibility of such a rollback.

yocalebo (approved):

PR #8904, opened 2022-05-09
NAS-116090 / 13.0 / fix off-by-one for enclosure mgmt for MINIs

yocalebo:
We did this for the other hardware platforms but seems the minis were left out.

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
@@ -25,8 +25,8 @@ MAPPINGS = [
             MappingSlot(0, 2, False),
             MappingSlot(0, 3, False),
             MappingSlot(0, 4, False),
-            MappingSlot(0, 1, False),
-            MappingSlot(0, 2, False),
+            MappingSlot(1, 1, False),
+            MappingSlot(1, 2, False),
         ]),
     ]),
     ProductMapping(re.compile(r"(TRUE|FREE)NAS-MINI-3.0-X$"), [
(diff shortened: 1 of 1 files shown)


src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
```diff
@@ -11,55 +11,55 @@
...
     ProductMapping(re.compile(r"(TRUE|FREE)NAS-MINI-3.0-E\+$"), [
         VersionMapping(re.compile(".*"), [
-            MappingSlot(0, 0, False),
             MappingSlot(0, 1, False),
             MappingSlot(0, 2, False),
             MappingSlot(0, 3, False),
-            MappingSlot(0, 0, False),
+            MappingSlot(0, 4, False),
             MappingSlot(0, 1, False),
+            MappingSlot(0, 2, False),
```
rick-mesta, line 29: Doesn't this just repeat line 25 ? it seems like line 28 also repeats line 24 🤔

themylogin (approved):

themylogin:
Is backporting required?

src/freenas/usr/local/lib/middlewared_truenas/plugins/enclosure_/map.py
yocalebo: It does, good catch

yocalebo: Actually, this makes it so that TrueNAS MINI E enclosure mgmt probably hasn't worked the way it's supposed to have 🙂

rick-mesta: `for i in ${UNTIL_CORRECT}; do
    repeat code` 😜

rick-mesta (approved):

PR #8913, opened 2022-05-10
NAS-116136 / 22.12 / optimize failover.disabled_reasons on HA

yocalebo:
We call `failover.call_remote system.ready` with a timeout of 5 seconds so that we don't hang the `failover.status` calls. However, we're continually calling `failover.call_remote` no matter the return of the prior API call. This changes it so that if we timeout checking `system.ready` then we short-circuit.


yocalebo:
nope

PR #8934, opened 2022-05-11
NAS-115771 / 22.12 / Add oneshot alert for deprecated services that are running

anodos325:
Currently this only contains AFP in 13, but having general-purpose
alert may be useful so that users can plan migration off things
that we will remove in the future.

Make backend changes to flag service as "deprecated"

Deprecated services generate alert when they are running that
they will soon be removed. This allows us to nag users to
make plans to migrate away from the service.


anodos325:
This is backport of feature from 13. I think it would be useful to have as we may need in future to deprecate some features or services.

yocalebo (approved):

PR #8941, opened 2022-05-12
NAS-116154 / 22.12 / Improve error message when quota cannot be set on user/group

themylogin:


themylogin:
time 0:45

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -4040,14 +4040,12 @@ async def set_quota(self, ds, data):
...
                 else:
-                    id_type = ('user', 'uid') if 'user' in quota_type else ('group', 'gid')
                     xid = int(q["id"])
 
                 if xid == 0:
                     verrors.add(
                         f'quotas.{i}.id',
-                        f'{quota_type}: Setting {id_type[0]} quota on {id_type[1]} [{xid}] '
-                        'is not permitted.'
+                        f'Setting {quota_type} quota on {id_type[0]}id [{xid}] is not permitted.'
```
anodos325, line 4048: id_type is now a string so we probably don't want first char here. Also space missing between `{id_type}` and "id".

themylogin: We want `uid` for `user` and `gid` for `group`, that's what `{id_type[0]}id` is

anodos325: Ah, right. got it.

anodos325 (approved):

PR #8942, opened 2022-05-12
NAS-116174 / 22.12 / Show most important alerts first

themylogin:
When vmware snapshot task fails, an alert is issued. User ignored this alert and was unaware that vmware snapshots were not taken due to network misconfiguration. I had a hard time noting this alert too because it was Nth after N "enclosure is HEALTHY" alerts and was not visible in a small browser window. We might want to sort alerts by their level first.

### DIFF
--- a/src/middlewared/middlewared/plugins/alert.py
+++ b/src/middlewared/middlewared/plugins/alert.py
@@ -347,10 +347,18 @@ class AlertService(Service):
         """
 
         as_ = AlertSerializer(self.middleware)
+        classes = (await self.middleware.call("alertclasses.config"))["classes"]
 
         return [
             await as_.serialize(alert)
-            for alert in sorted(self.alerts, key=lambda alert: (alert.klass.title, alert.datetime))
+            for alert in sorted(
+                self.alerts,
+                key=lambda alert: (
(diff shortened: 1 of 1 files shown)


themylogin:
time 0:20

src/middlewared/middlewared/plugins/alert.py
```diff
@@ -347,10 +347,18 @@ async def list(self):
...
         as_ = AlertSerializer(self.middleware)
+        classes = (await self.middleware.call("alertclasses.config"))["classes"]
 
         return [
             await as_.serialize(alert)
-            for alert in sorted(self.alerts, key=lambda alert: (alert.klass.title, alert.datetime))
+            for alert in sorted(
+                self.alerts,
+                key=lambda alert: (
+                    -get_alert_level(alert, classes).value,
```
sonicaj, line 357: typo `-` ?

themylogin: No, I want the highest level first, so CRITICAL (7) should come before INFO (1)

sonicaj (approved):
Thank you for clarifying!

PR #8948, opened 2022-05-13
NAS-115802 / 22.12 / Make sure system reflects newer cert before redeploying cert attachments

sonicaj:
This commit fixes an issue where we re-deployed cert attachments but did not update the certs on the filesystem which resulted in issues like nginx still using the old cert.


yocalebo (approved):
Is this a problem in 13?

sonicaj:
@yocalebo we don't have certificate attachment delegates in 13 which would be quite an invasive change for a stable branch. So until we would like to add the complete feature, nope - this is not to be backported yet to 13

PR #8957, opened 2022-05-16
NAS-116233 / 22.12 / Fix alternate code path for SMB connection rename in tests

anodos325:
On servers running API test that lack the rename() python smbclient binding we need to wrap around
the `smbclient` cli utility. I forgot to keep a copy of the credentials around when establishing the
SMB connection to use for the CLI utility.


yocalebo (approved):

ericbsd (approved):
LGTM

PR #8972, opened 2022-05-17
NAS-115860 / 22.12 / fix importing zpools on SCALE

yocalebo:
Discovered and witnessed on an internal M50 (no expansion shelf) as well as an R50b with an ES60 expansion shelf (in the field). The simplified version is that `/dev/disk/by-partuuid` symlinks aren't being created by the time the raw devices are being populated inside `/dev/`.

Since we're trying to import the zpools specifying `/dev/disk/by-partuuid` AND `/dev/` it means zpool is importing the disks via gptid but if it can't find one, it's choosing a random raw device. The device letters for raw devices aren't guaranteed between reboots and often change so when the zpool is imported, certain devices are "missing" and other drives are being put in their place.

This is painful because the zpool is now imported and in an unhealthy state. This adds a simple helper script that gets called as a prerequisite to the `ExecStart` entries in the `ix-zfs.service` file. Testing this on the M50, fixed the problem and the zpool imported with gptid's and produced a healthy pool.

NOTE: This is still "flawed" and not something that I want to do but our hand is forced. The "proper" solution is for openzfs to actually integrate with systemd and use the proper mechanism(s) to automagically handle this.

The `ix-zfs.service` has a time limit of 15mins while we allow udev "events" to be received for a maximum of 10mins. This worked on the R50b and the M50 so I'm leaving those time limits for now.

BONUS: adding `After=systemd-udevd-settle.service` is what I initially thought would be the "solution" but that emits a DeprecationWarning from systemd. Furthermore, reading the documentation uses big scary verbiage about how this is absolutely the wrong approach, so I've refrained from doing that.

### DIFF
--- a/debian/debian/ix-zfs.service
+++ b/debian/debian/ix-zfs.service
@@ -8,6 +8,7 @@ After=middlewared.service
 [Service]
 Type=oneshot
 RemainAfterExit=yes
+ExecStartPre=/usr/bin/python3 /usr/lib/python3/dist-packages/middlewared/scripts/ix_zfs_service_helper.py
 ExecStart=-midclt call disk.sed_unlock_all
 ExecStart=midclt call -job --job-print description pool.import_on_boot
 StandardOutput=null
--- a/debian/debian/ix-zfs.service
+++ b/debian/debian/ix-zfs.service
@@ -8,7 +8,7 @@ After=middlewared.service
 [Service]
 Type=oneshot
 RemainAfterExit=yes
-ExecStartPre=/usr/bin/python3 /usr/lib/python3/dist-packages/middlewared/scripts/ix_zfs_service_helper.py
+ExecStartPre=/usr/bin/zpoolhelper
 ExecStart=-midclt call disk.sed_unlock_all
 ExecStart=midclt call -job --job-print description pool.import_on_boot
 StandardOutput=null
--- a/src/middlewared/middlewared/scripts/zpoolhelper.py
+++ b/src/middlewared/middlewared/scripts/zpoolhelper.py
@@ -1,3 +1,4 @@
+#!/usr/bin/python3
 from collections import deque
 from time import time, sleep
 from sys import exit
--- a/src/middlewared/middlewared/scripts/zpoolhelper.py
+++ b/src/middlewared/middlewared/scripts/zpoolhelper.py
@@ -5,7 +5,7 @@ from sys import exit
 
 from pyudev import Context, Monitor, MonitorObserver
 
-dq = deque()
+dq = deque(maxlen=2)
 
 
 def callback(dev):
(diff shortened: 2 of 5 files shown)


debian/debian/ix-zfs.service
```diff
@@ -8,6 +8,7 @@ After=middlewared.service
 [Service]
 Type=oneshot
 RemainAfterExit=yes
+ExecStartPre=/usr/bin/python3 /usr/lib/python3/dist-packages/middlewared/scripts/ix_zfs_service_helper.py
```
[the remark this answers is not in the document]
yocalebo: It's not in `/usr/bin/` unless I put it as an entry point in `middlewared/setup.py`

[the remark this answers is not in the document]
yocalebo: I guess I can add it there to further complicate/hide/abstract away inner workings 😄

src/middlewared/middlewared/scripts/zpoolhelper.py
```diff
@@ -0,0 +1,54 @@
...
+
+    max_time_to_wait = 600  # total time to wait in seconds
+    interval = 5  # seconds to sleep before checking for new event
+    last_event = {}
+    while max_time_to_wait > 0:
+        max_time_to_wait -= interval
+        sleep(interval)
+
+        try:
+            event = dq[-1]
```
sonicaj, line 36: I am not sure what the size can be here but is there a good reason to use `deque` or `list` here as we are not really interested in keeping all the items ?

yocalebo: `deque` is vastly more performant, vastly more memory efficient and the `.append` method is thread-safe 😄. The good news is that once `ix-zfs.service` exits memory is freed since this is a forked child process done by systemd.

sonicaj: Nope, i meant that do we even need to use `list` or `deque` ?

sonicaj: I think let's set `maxlen` for it to 5 or 10 ?

sonicaj: Actually woudl require testing as iirc it would be blocking and won't require further additions

yocalebo: I'd rather not set a length on the deque. The memory is freed once the script exits. There is no reason to artificially limit the size since it doesn't gain us anything.

sonicaj (approved):
If you think it's good, i don't mind. I was just not seeing a reason to let a list/deque grow to huge lengths when we are only interested in last 2 events maximum

yocalebo:
> If you think it's good, i don't mind. I was just not seeing a reason to let a list/deque grow to huge lengths when we are only interested in last 2 events maximum

That's a valid point, let me make sure `.append(` doesn't block once it's full

yocalebo:
retest this please

themylogin:
@yocalebo @sonicaj why did we remove `flake8-import-order`?

themylogin:
Also why is this a script outside of middleware?

PR #8981, opened 2022-05-18
NAS-116275 / 13.0 / Remove osc plugin from boot_/format.py

yocalebo:


anodos325 (approved):

rkojedzinszky:
Is the invalidate call at the right position just after wiping disks? Shouldn't it be placed after gpart operations?

yocalebo:
> Is the invalidate call at the right position just after wiping disks? Shouldn't it be placed after gpart operations?

Sheesh, you're right. I'll fix 👍

yocalebo:
Should be fixed.

PR #8982, opened 2022-05-18
fix R50 and R50b enclosure mapping on SCALE

yocalebo:
1. This adds the `R50b` device to our mapping code so it gets mapped the same as our `R50` since the SAS expanders are wired the same. The only exception to this is that R50b has 2 rear nvme instead of 3 like on the R50
2. Use `pyudev.Context` to list enclosures since they're enumerated in proper order so that the static mapping works (this is working by chance so I'll fix this particular problem in a separate PR since it effects every platform that gets mapped)
3. remove `ses_enclosure_freebsd` module for obvious reasons
4. R50 and R50b don't have a `/dev/ses2` so comment those out since rear bays for the nvme drives will have to be fixed entirely on SCALE

Note: There will be more PRs to fix the various other issues related to enclosure mapping on SCALE. This is just a preliminary get things into a semi-broken state instead of a fully broken state 😄


themylogin (approved):

themylogin:
We can also start renaming `ses_enclosure_linux.py` to just `ses_enclosure.py`

PR #8984, opened 2022-05-18
NAS-116283 / 22.12 / fix pool.is_upgraded_by_name

yocalebo:
`py-libzfs` gives us the pool feature flags so there is no need to `subprocess` out (potentially twice) when checking to see if a pool's feature flags have been upgraded.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -13,7 +13,6 @@ import os
 import re
 import secrets
 import shutil
-import subprocess
 import uuid
 
 from collections import defaultdict
@@ -32,7 +31,7 @@ from middlewared.service import (
 )
 from middlewared.service_exception import InstanceNotFound, ValidationError
 import middlewared.sqlalchemy as sa
-from middlewared.utils import Popen, filter_list, run
+from middlewared.utils import filter_list, run
 from middlewared.utils.asyncio_ import asyncio_map
 from middlewared.utils.path import is_child
 from middlewared.utils.size import MB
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -99,6 +99,14 @@ def query_imported_fast(self):
         except libzfs.ZFSException as e:
             raise CallError(f'Failed listing imported pools with error: {e}')
 
+    def is_upgraded(self, pool_name):
+        is_upgraded = False
```
themylogin, line 103: We should probably raise `ENOENT` `CallError` here if the pool name was invalid and only keep `return False if the pool name is invalid` behavior for the public method.

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -99,6 +99,14 @@ def query_imported_fast(self):
...
+        with libzfs.ZFS() as zfs:
+            for pool in filter(lambda x: x.name == pool_name, zfs.pools):
+                is_upgraded = not any((i.state.name not in ('ACTIVE', 'ENABLED') for i in pool.features))
```
themylogin, line 106: Double negatives are harder to understand. Isn't this `all((i.state.name in ('ACTIVE', 'ENABLED') for i in pool.features)))`

Also is `i.state` an enum? If so, it's better to compare with enum values than with their string representations.

themylogin (changes requested):

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -99,6 +99,13 @@ def query_imported_fast(self):
         except libzfs.ZFSException as e:
             raise CallError(f'Failed listing imported pools with error: {e}')
 
+    def is_upgraded(self, pool_name):
+        with libzfs.ZFS() as zfs:
+            for pool in filter(lambda x: x.name == pool_name, zfs.pools):
+                return all((i.state.value in (1, 2) for i in pool.features))
```
themylogin, line 105: Why? I meant `i.state in (libzfs.FeatureState.ENABLED, libzfs.FeatureState.ACTIVE)`, will it work?

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -432,33 +431,7 @@ async def is_upgraded(self, oid):
...
-                for line in [i for i in data.split('\n') if i.startswith('feature') and '\t' in i]:
-                    prop, value = line.split('\t', 1)
-                    if value not in ('active', 'enabled'):
-                        return False
-                return True
-            else:
-                return False
-        else:
-            return False
+        return await self.middleware.call('zfs.pool.is_upgraded', name)
```
themylogin, line 434: This method should either keep returning `False` for unavailable pools or all its invocations should handle `CallError` (e.g. here https://github.com/truenas/middleware/blob/a26e1bc815b21cf85f5c8cfa71611ae809a5ead1/src/middlewared/middlewared/plugins/pool.py#L558 it might be called for a non-existing pool that is only present in the configuration database).

themylogin (changes requested):

PR #8989, opened 2022-05-18
NAS-112344 / 13.0 / Do not overreact to `.fail` TLD

themylogin:
SCALE does not have this regex so no backport is needed


themylogin:
time 0:15

yocalebo (approved):

PR #8992, opened 2022-05-19
NAS-116163 / 22.12 / ensure unique /etc/machine-id on upgrade from CORE

yocalebo:
`/etc/machine-id` is used by kernel to generate globally unique mac addresses along with many other things. We're ensuring that this is unique on fresh installation or upgrade between SCALE releases but we're not ensuring this on upgrade from CORE to SCALE so all CORE machines being upgraded to SCALE have the same `/etc/machine-id` which means, for example, virtual interfaces will share the same mac addresses. This should fix that by removing the associated files and regenerating it at upgrade time.

### DIFF
--- a/src/middlewared/middlewared/plugins/update_/freebsd_to_scale_linux.py
+++ b/src/middlewared/middlewared/plugins/update_/freebsd_to_scale_linux.py
@@ -1,6 +1,5 @@
-import contextlib
 import logging
-import os
+from pathlib import Path
 
 from middlewared.service import job, private, Service
 from middlewared.utils import run
@@ -12,9 +11,11 @@ class UpdateService(Service):
 
     @private
     def remove_files(self):
-        with contextlib.suppress(FileNotFoundError):
-            for i in ("/data/freebsd-to-scale-update", "/var/lib/dbus/machine-id", "/etc/machine-id"):
-                os.unlink(i)
+        for i in ("/data/freebsd-to-scale-update", "/var/lib/dbus/machine-id", "/etc/machine-id"):
+            try:
+                Path(i).unlink(missing_ok=True)
+            except Exception:
+                logger.error('Failed removing %r', i, exc_info=True)
 
     @private
     @job()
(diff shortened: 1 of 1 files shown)


anodos325 (approved):

src/middlewared/middlewared/plugins/update_/freebsd_to_scale_linux.py
```diff
@@ -9,14 +9,20 @@
 
 
 class UpdateService(Service):
+
+    @private
+    def remove_files(self):
+        with contextlib.suppress(FileNotFoundError):
```
themylogin, line 15: Context manager needs to be applied to each remove operation independently; otherwise, non-existence of one file will prevent the removal of subsequent ones. Also, IIRC, in the pathlib library, there is a flag to remove a file a be ok if it does not exist.

src/middlewared/middlewared/plugins/update_/freebsd_to_scale_linux.py
```diff
@@ -9,14 +9,20 @@
...
     @job()
     async def freebsd_to_scale(self, job):
-        logger.info("Updating FreeBSD installation to SCALE")
-
-        with contextlib.suppress(FileNotFoundError):
-            os.unlink("/data/freebsd-to-scale-update")
+        logger.info("Updating CORE installation to SCALE")
 
+        await self.middleware.run_in_thread(self.remove_files)
+        await run(["systemd-machine-id-setup"], check=False)
```
themylogin, line 25: Will it be a no-op if machine-id exists already (if we upgraded using ISO instller which can set up machine id)

themylogin (changes requested):

src/middlewared/middlewared/plugins/update_/freebsd_to_scale_linux.py
yocalebo: Yes, `systemd-machine-id-setup` doesn't change an existing `/etc/machine-id` so it will be a no-op.

themylogin (approved):
LGTM, thank you!

PR #8995, opened 2022-05-19
Fix certificate paths of k3s

Qubad786:
## Problem

We have a user who exported his pool via UI and imported the pool via command line changing the name in the process and then exporting it back and importing it via UI.
What happened during this is that the cert locations referenced to old pool name path whereas that did not hold anymore as the name had changed.

## Solution

Make sure each time a pool is configured to be used for Apps, we fix it's certificate paths as required. This looks after the case as well where user might have replicated the ix-applications dataset on his own.


Ornias1993 (approved):
I can verify I had this problem once as well, the fix is pretty KISS. :)

sonicaj (approved):

PR #9004, opened 2022-05-20
Look up libvirt-quemu ids through NSS

anodos325:
Look up the actual on-system ids for this user before chown and
use proper filesystem API for the chown.


anodos325:
This PR is now obsolete due to other work in relevant plugins.

PR #9005, opened 2022-05-20
NAS-116326 / 22.12 / Significantly increase maxPods per node

Ornias1993:
Current singleNode setup combined with the current build-in loadbalancer that adds 1 extra(!) pod per loadbalancer, users run into issues where the max amount of pods is limited to 110 pods.

This increases said limit to 250 pods, which is slightly more than double, but not so-large that it would cause significant issues.
In the future, if need be, this could easily be adapted to grab a database value instead, but that does not seem to be really needed currently.

Also minor formatting change:
It puts the major k3s settings (kubelet, api-server and kubecontroller), right beneat eachother in the yaml file. For clean reading.

Note:
Also needs to be backported to 22.02.2.

### DIFF
--- a/src/middlewared/middlewared/etc_files/rancher/k3s/flags.py
+++ b/src/middlewared/middlewared/etc_files/rancher/k3s/flags.py
@@ -30,6 +30,9 @@ def render(service, middleware):
         'service-account-lookup=true',
         'feature-gates=MixedProtocolLBService=true',
     ]
+    kubelet_args = [
+        'max-pods=250',
+    ]
     os.makedirs('/etc/rancher/k3s', exist_ok=True)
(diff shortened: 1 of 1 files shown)


Ornias1993:
@sonicaj and @Qubad786 Issues reported on the forums here:
https://www.truenas.com/community/threads/add-a-limit-on-the-number-of-pods.101306/

Simple fix.

Qubad786 (approved):

Qubad786:
LGMT @Ornias1993!

Ornias1993:
@Qubad786 As this is a pretty minor fix for a known issue, mind adding the backport tag to 22.02.2?

yocalebo:
backport

PR #9018, opened 2022-05-23
NAS-116289 / 22.12 / Generate new uuid when cloning a VM

sonicaj:

1. Generate different uuid when cloning a vm as vms cannot share the same uuid
2. Allow force deleting a VM even if libvirt complains undefining a domain


Qubad786 (approved):

sonicaj:
backport

PR #9019, opened 2022-05-23
NAS-116234 / 22.12 / Properly raise exception if specified user does not exist

sonicaj:
This commit adds changes to make sure that we raise an exception if specified user does not exist instead of abruptly terminating the process pool.

### DIFF
--- a/src/middlewared/middlewared/utils/osc/linux/user_context.py
+++ b/src/middlewared/middlewared/utils/osc/linux/user_context.py
@@ -3,7 +3,6 @@ import concurrent.futures
 import functools
 import logging
 import os
-import pwd
 import subprocess
 
 from typing import Any, Callable, Optional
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/utils/osc/linux/user_context.py
```diff
@@ -41,6 +44,10 @@ def set_user_context(user: str) -> None:
 
 
 def run_with_user_context(func: Callable, user: str, func_args: Optional[list] = None) -> Any:
+    try:
+        pwd.getpwnam(user)
```
anodos325, line 48: We have method that wraps around this. `user.get_user_obj` We should use this. Maybe pass the resulting dict into set_user_context.

anodos325:
```
    @accepts(Dict(
        'get_user_obj',
        Str('username', default=None),
        Int('uid', default=None),
        Bool('get_groups', default=False)
    ))
```

May want to use `get_groups` option as well so that we can setresuid() into correct list of users / groups.

anodos325 (approved):

PR #9023, opened 2022-05-24
NAS-116373 / 22.12 / include file type in filesystem.stat output

anodos325:
This is a convenience feature to show whether the target is
directory, symlink, etc with appropriate call to lstat() /
stat() depending on circumstances. Also include in stat() output
whether the path is a mountpoint.


src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -242,6 +244,7 @@ def stat_entry(entry):
 
         return filter_list(rv, filters=filters or [], options=options or {})
 
+    """
```
themylogin, line 247: Did you intent to remove it?

themylogin (commented):
Would you like to add some integration tests?

src/middlewared/middlewared/plugins/filesystem.py
anodos325: No :) Conference brain.

themylogin (approved):

PR #9032, opened 2022-05-26
NAS-116426 / 22.12 / add db storageclass name [WIP] [DNM]

Ornias1993:
WORK IN PROGRES, DO NOT MERGE


Ornias1993:
Never mind, found another solution

PR #9033, opened 2022-05-26
NAS-116433 / 22.02.2 / device.get_disks() optimizations

yocalebo:
This optimizes retrieving disk information on SCALE. On a system with 439 disks, `device.get_disks` was taking `0.867` seconds. After my changes it takes `0.439` seconds which is ~66% decrease in time it takes to get disk information.

My changes do roughly the following:
1. remove a call to `lsblk -OJdb` when querying disks. `lsblk` gets the _same_ disk information that `pyudev` gets
2. There was a `self.logger` crash in `get_disks()` method
3. Stop doing `self.disk_default.copy()` for each disk. Every little thing adds up when we're dealing with a system with many hundreds (or more) disks.
4. Remove the extra `subprocess` calls to `sg_vpd`. If udev doesn't have the information, `sg_vpd` isn't going to get it either since the kernel asks the disks for the same information using the same requests....
5. we had overly complicated logic to calculate the `number` key which isn't necessary when we can get a unique number from udev (device_number)

### DIFF
--- a/src/middlewared/middlewared/plugins/device_/device_info_linux.py
+++ b/src/middlewared/middlewared/plugins/device_/device_info_linux.py
@@ -33,14 +33,15 @@ class DeviceService(Service, DeviceInfoBase):
                 continue
 
             try:
-                disks[dev.sys_name] = self.get_disk_details(dev, is_nvme=dev.sys_name.startswith('nvme'))
+                disks[dev.sys_name] = self.get_disk_details(dev)
             except Exception:
                 self.logger.debug('Failed to retrieve disk details for %s', dev.sys_name, exc_info=True)
 
         return disks
 
     @private
-    def get_disk_details(self, dev, is_nvme=False):
+    def get_disk_details(self, dev):
+        is_nvme = dev.sys_name.startswith('nvme')
         size = mediasize = self.safe_retrieval(dev.attributes, 'size', None, asint=True)
         ident = serial = self.safe_retrieval(dev.properties, 'ID_SERIAL_SHORT' if is_nvme else 'ID_SCSI_SERIAL', '')
         model = descr = self.safe_retrieval(dev.properties, 'ID_MODEL', None)
--- a/src/middlewared/middlewared/plugins/device_/device_info_linux.py
+++ b/src/middlewared/middlewared/plugins/device_/device_info_linux.py
@@ -14,9 +14,10 @@ from middlewared.utils.gpu import get_gpus
 from middlewared.utils import osc
 
 RE_DISK_SERIAL = re.compile(r'Unit serial number:\s*(.*)')
-RE_NVME_PRIVATE_NAMESPACE = re.compile(r'nvme[0-9]+c')
 RE_SERIAL = re.compile(r'state.*=\s*(\w*).*io (.*)-(\w*)\n.*', re.S | re.A)
 RE_UART_TYPE = re.compile(r'is a\s*(\w+)')
+RE_NVME_PRIV = re.compile(r'nvme[0-9]+c')
+IGNORE = ('sr', 'md', 'dm-', 'loop', 'zd')
 
 
 class DeviceService(Service, DeviceInfoBase):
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -28,28 +29,79 @@ def get_serials(self):
...
-            device_type = os.path.join('/sys/block', dev.sys_name, 'device/type')
-            if os.path.exists(device_type):
-                with open(device_type, 'r') as f:
-                    if f.read().strip() != '0':
-                        continue
-            # nvme drives won't have this
 
             try:
-                disks[dev.sys_name] = self.get_disk_details(dev, self.disk_default.copy(), disks_data)
+                disks[dev.sys_name] = self.get_disk_details(dev, is_nvme=dev.sys_name.startswith('nvme'))
```
themylogin, line 40: Can we calculate `is_nvme` inside `get_disk_details`?

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -14,9 +14,10 @@
 from middlewared.utils import osc
 
 RE_DISK_SERIAL = re.compile(r'Unit serial number:\s*(.*)')
-RE_NVME_PRIVATE_NAMESPACE = re.compile(r'nvme[0-9]+c')
 RE_SERIAL = re.compile(r'state.*=\s*(\w*).*io (.*)-(\w*)\n.*', re.S | re.A)
 RE_UART_TYPE = re.compile(r'is a\s*(\w+)')
+RE_NVME_PRIV = re.compile(r'nvme[0-9]+c')
+IGNORE = ('sr', 'md', 'dm-', 'loop', 'zd')
```
themylogin, line 20: We have the exact same set in `disk_events.py` and almost the same set in `enclosure.py`, can we use a common variable?

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -28,28 +29,79 @@ def get_serials(self):
...
+            'blocks': None,
+            'serial_lunid': None,
+            'rotationrate': None,
+            'stripesize': None,  # remove this? (not used)
+        }
+
+        if disk['size'] and disk['sectorsize']:
+            disk['blocks'] = int(disk['size'] / disk['sectorsize'])
+
+        if (info := self.safe_retrieval(dev.attributes, 'queue/rotational', None)) and info == '1':
```
themylogin, line 78:
```suggestion
        if self.safe_retrieval(dev.attributes, 'queue/rotational', None) == '1':
```

src/middlewared/middlewared/plugins/device_/device_info_linux.py
```diff
@@ -28,28 +29,79 @@ def get_serials(self):
...
+            if type(value) == bytes:
+                value = value.strip().decode()
+            else:
+                value = value.strip()
+            return value if not asint else int(value)
+
+        return default
+
     @private
     def retrieve_disks_data(self):
```
themylogin, line 106: This method does not seem to be used anymore, do we keep it?

themylogin (changes requested):

src/middlewared/middlewared/plugins/device_/device_info_linux.py
yocalebo: Why?

themylogin: To simplify function call to just `self.get_disk_details(dev)`. It has no sense to make caller calculate `is_nvme`. Imagine this
```
self.get_disk_details(dev, is_nvme=dev.sys_name.startswith('nvme'),
                      size=self.safe_retrieval(dev.attributes, 'size', None, asint=True))
```
Pretty weird, right?

themylogin (changes requested):

src/middlewared/middlewared/plugins/device_/device_info_linux.py
yocalebo: No, not to me 🤷‍♂️

themylogin (approved):
LGTM, thank you!

PR #9034, opened 2022-05-27
NAS-116385 / 22.12 / Hyper-V enlightenments

heppu:
Add support for Hyper-V enlightenments which are important part of running Windows VMs on. I tried to follow TrueNAS philosophy of keeping it simple and hence the single flag to enable most commonly supported flags. Other option could be adding a option per each Hyper-V feature but that would quickly clutter the UI.

This is my first PR for TrueNAS so please instruct me if something is missing/incorrect.

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/supervisor/supervisor_base.py
+++ b/src/middlewared/middlewared/plugins/vm/supervisor/supervisor_base.py
@@ -240,7 +240,7 @@ class VMSupervisorBase(LibvirtConnectionMixin):
             # Add features
             self.get_features_xml(),
             # Clock offset
-            create_element('clock', offset='localtime' if self.vm_data['time'] == 'LOCAL' else 'utc'),
+            self.get_clock_xml(),
             # Devices
             self.devices_xml(),
             # Command line args
@@ -278,10 +278,15 @@ class VMSupervisorBase(LibvirtConnectionMixin):
             features.append(
                 create_element('kvm', attribute_dict={'children': [create_element('hidden', state='on')]})
             )
+
+        if self.vm_data['hyperv_enlightenments']:
+            features.append(self.get_hyperv_xml())
+
         return create_element(
             'features', attribute_dict={
                 'children': [
                     create_element('acpi'),
+                    create_element('apic'),
                     create_element('msrs', unknown='ignore'),
                 ] + features,
             }
@@ -296,6 +301,36 @@ class VMSupervisorBase(LibvirtConnectionMixin):
                 create_element('currentMemory', unit='M', attribute_dict={'text': str(self.vm_data['min_memory'])})
             )
         return memory_xml
+
+    def get_clock_xml(self):
+        timers = []
+        if self.vm_data['hyperv_enlightenments']:
+            timers = [create_element('timer', name='hypervclock', present='yes')]
+
+        return create_element(
+            'clock', attribute_dict={'children': timers},
+            offset='localtime' if self.vm_data['time'] == 'LOCAL' else 'utc'
+        )
+
+    # Documentation for each enlightenment can be found from:
+    # https://github.com/qemu/qemu/blob/master/docs/system/i386/hyperv.rst
+    def get_hyperv_xml(self):
+        return create_element(
+            'hyperv', attribute_dict={
+                'children': [
+                    create_element('relaxed', state='on'),
+                    create_element('vapic', state='on'),
+                    create_element('spinlocks', state='on', retries='8191'),
+                    create_element('reset', state='on'),
+                    create_element('frequencies', state='on'),
+                    # All enlightenments under vpindex depend on it.
+                    create_element('vpindex', state='on'),
+                    create_element('synic', state='on'),
+                    create_element('ipi', state='on'),
+                    create_element('tlbflush', state='on'),
+                ],
+            }
+        )
 
     def cpu_xml(self):
         features = []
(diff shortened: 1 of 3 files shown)


heppu:
UI support for this feature: https://github.com/truenas/webui/pull/6722

src/middlewared/middlewared/plugins/vm/supervisor/supervisor_base.py
```diff
@@ -274,14 +274,51 @@ def get_features_xml(self):
...
             )
+
+        if self.vm_data['hyperv_enlightenments']: 
+            features.append(self.get_hyperv_xml())
+        
         return create_element(
             'features', attribute_dict={
                 'children': [
                     create_element('acpi'),
+                    create_element('apic'),
```
themylogin, line 285: @sonicaj what do you think, can adding this break anything for the existing users?

src/middlewared/middlewared/plugins/vm/vms.py
```diff
@@ -141,6 +143,9 @@ async def do_create(self, data):
 
         `hide_from_msr` is a boolean which when set will hide the KVM hypervisor from standard MSR based discovery and
         is useful to enable when doing GPU passthrough.
+
+        `hyperv_enlightenments` can be used to enable subset of predefined Hyper-V enlightenments for Windows guests. These
```
themylogin, line 147: Can we perhaps make it `Str('hyperv_enlightenments', null=True, enum=[None, 'WINDOWS']),` to allow more enlightenments presets for different types of OSes? @sonicaj

src/middlewared/middlewared/plugins/vm/supervisor/supervisor_base.py
heppu: I'm by no means expert on this topic but how I implemented this feature is opt-in so by default it doesn't have any effect on existing users. Also it allows users to try if their virtualization stack support these features by simply checking the hyper-v checkbox in UI. If their machine deosn't support the features then the UI will prompt error message like "Unable start VM foobar, CPU doesn't support featuate XYZ", or something similar (I verfied this on my SCALE instance by enabling some new flags not included in this review). After that user can simply uncheck the hyper-v checkbox and machine will start like it used to.

I'm open to adjusting the default set on enlightenments but this seemed like a reasonable default set which I have experimented with. Also it's important to remember that this doesn't take away anything from existing user, only enables better Windows support for those who are able to use it.

themylogin: `<apic/>` will be added to `<features>` for all VMs, does not matter if enlightenments are enabled or not, that's why I am asking.

heppu: Oh sorry my bad. Based on how the libvirt's documentation is worded I would say it shouldn't:

> APIC allows the use of programmable IRQ management.

So it should only enable things, not force.
https://libvirt.org/formatdomain.html#hypervisor-features

heppu:
Done

themylogin (changes requested):

src/middlewared/middlewared/alembic/versions/22.12/2022-05-26_18-50_hyperv_enlightenments.py
```diff
@@ -0,0 +1,24 @@
+"""
+Hyper-V enlightenments
+
+Revision ID: ac02454965d1
+Revises: 59fbc9897ec3
```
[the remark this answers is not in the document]
heppu: Done. There seems to be also new merge file in migrations so I don't know if this will now break the migration flow or does it just work.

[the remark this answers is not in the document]
heppu: Where does this `bca4e75ed8fc` come from?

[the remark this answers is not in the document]
themylogin: https://github.com/truenas/middleware/blob/master/src/middlewared/middlewared/alembic/versions/22.12/2022-05-31_12-02_merge.py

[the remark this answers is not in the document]
heppu: Oh there was yet another new merge. Done.

heppu:
Updated migration revision and tested locally.

src/middlewared/middlewared/plugins/vm/supervisor/supervisor_base.py
sonicaj: We should be good i believe

src/middlewared/middlewared/plugins/vm/vms.py
sonicaj: I think this is a good start, we do have some other hypervisor features which other OSes can benefit from but that would be separate from hyperv changes.

sonicaj (approved):
Thank you @heppu for your contribution

sonicaj:
ok to test

sonicaj:
retest this please

sonicaj (commented):
@heppu can you please update your migrations so that down revision reflects the latest version available in master ? Thanks!

heppu:
@sonicaj  Done.

sonicaj:
@heppu you are using wrong ticket number btw. Can you share the ticket number you created for this ticket please otherwise we can generate a new ticket ? Thanks!

themylogin (approved):

PR #9035, opened 2022-05-27
NAS-116203 / 22.02.2 / Improve zpool capacity alert, make it proactive, allow disabling proactive support for specific alert classes

themylogin:


themylogin:
@william-gr please evaluate if we want this to be backported to 13.0

themylogin:
time 4:00

yocalebo (approved):

PR #9037, opened 2022-05-27
NAS-116454 / 22.12 / Add mount flags to filesystem.statfs() output

anodos325:
ZFS exposes many dataset properties via string at end of mountinfo.
The info is available already in data we gather for fstype return. Include it
as well. Some of info zfs provides here isn't available via simple statfs(2) call
on Linux.


themylogin (approved):
LGTM, but, as always, it would be great to see some integration tests to at least understand what flags are being returned.

PR #9040, opened 2022-05-27
NAS-117011 / 23.10 / Check that writes to ADS work on DOS RO file

anodos325:
It is expected to still be able to write to an alternate
data stream on a file that has the DOS readonly bit set.


pkgdemon:
retest this please

anodos325:
Tested using win32api and determined that you cannot use an open file handle to open a new data stream on a file without re-evaluating the RO bit.

PR #9045, opened 2022-05-28
NAS-115711 / Support CPU and NUMA pinning

heppu:
This PR adds features which are extremely important for predictable VM performance and cache locality. 

To achive these improvements 3 options were added in vm model:
1. string to define cpuset
2. string to define nodeset
3. boolean to performe 1:1 CPU pinning
3.1 only allowed when `n. vcpus == n. cpus in cpuset`
3.2 pinning happens by following order defined in cpuset 

With these 3 simple options we can cover most of the cases where CPU pinning is usually applied without exposing libvirt's complex API to cpu pinning.

I have been running Windows 10 VM with GPU passtrough on threadripper pro machince with this and https://github.com/truenas/middleware/pull/9034 applied and these are the changes I have been able to notice:

- No more stutter in UI
- No more "stuck keypresses" when key was pressed only once
- No more lag in mouse
- No more sudden FPS drops while gaming


src/middlewared/middlewared/plugins/vm/supervisor/supervisor_base.py
```diff
@@ -248,6 +247,11 @@ def get_domain_children(self):
             *self.commandline_xml(),
         ]
 
+        if self.vm_data.get('pin_vcpus') and self.vm_data.get('cpuset'):
```
themylogin, line 250: Why `.get`? (same question in other places too)

themylogin (commented):
Looks great to me!

Can you please move numeric set parser and validator to `src/middlewared/middlewared/plugins/vm` (it is not used anywhere else) and add unit tests for `parse_numeric_set`?

src/middlewared/middlewared/alembic/versions/22.12/2022-05-27_12-30_cpu_pinning.py
```diff
@@ -0,0 +1,28 @@
+"""CPU Pinning
+
+Revision ID: 0267cef97ec3
+Revises: 59fbc9897ec3
+Create Date: 2022-05-27 12:30:03.593598+00:00
+
+"""
+from email.policy import default
```
themylogin, line 8: What is this for?

heppu: I forgot to remove it. I created this migration manually by copying existing one since I wasn't able to get alembic create one for me. Could that possibly cause some problems? Also on that note is there some dev vm/container which has a proper python/tooling setup for this project which can be used?

src/middlewared/middlewared/plugins/vm/supervisor/supervisor_base.py
heppu: It's long time since I did some python development but since those are defined as null=true I was thinking that could accessing those possibly lead into some KeyErrors if `.get` is not used?

src/middlewared/middlewared/alembic/versions/22.12/2022-05-27_12-30_cpu_pinning.py
themylogin: Your migration is correct.

We don't have any special dev images, you can just install the necessary tools using `apt` (since TrueNAS is just a Debian Bullseye). To update the code we usually just rsync `/usr/lib/python3/dist-packages/middlewared` or mount it from developer machine with NFS. I can build a custom ISO image from this PR if you like, but if you were able to fully test it by updating usual Nightly build python files and restarting middleware, that should not be necessary. Thank you for your contribution!

heppu: Thank you but no need for that iso like you said. I was using the Makefile to sync code into my Truenas machine and it worked just fine. It was just the alembic command that didn't work when I tried to create new migration file like instructed in the docs. Maybe there's something wrong in my local python setup but if the manually created file is fine then I'm good. =)

heppu:
I have been able to achieve the most stable result for Windows VM by I combining the hyper-v PR https://github.com/truenas/middleware/pull/9045 so if some has time to take a look at it that would be awesome.

themylogin:
@heppu everything almost good, there are two leftover `.get` in https://github.com/truenas/middleware/pull/9045/files#diff-0ba7a01db9e48a9c800d93ff85fa3b6ce5f891f7cf9ede343c8f960033fb4bb6R222 and a number of style guide errors here https://github.com/truenas/middleware/runs/6656795585?check_suite_focus=true

heppu:
Should be now fixed.

heppu:
Here's the related UI PR: https://github.com/truenas/webui/pull/6733

themylogin:
@heppu a few flake8 errors left
```
+src/middlewared/middlewared/plugins/vm/supervisor/supervisor_base.py:315:1: W293 blank line contains whitespace
+src/middlewared/middlewared/plugins/vm/supervisor/supervisor_base.py:317:1: W293 blank line contains whitespace
+src/middlewared/middlewared/plugins/vm/supervisor/supervisor_base.py:324:1: W293 blank line contains whitespace
+src/middlewared/middlewared/plugins/vm/numeric_set.py:5:1: E302 expected 2 blank lines, found 1
```

heppu:
Fixed

themylogin (approved):
Approved, thank you!

themylogin:
Jira URL: https://jira.ixsystems.com/browse/NAS-116479

PR #9051, opened 2022-05-30
NAS-116470 / 22.12 / Ensure disk_choices methods don't show in-use zvols

anodos325:
Add function to lookup available zvols via /sys/block and /dev/zvol.
This can be significantly faster than using py-libzfs. Plumb new
function into iscsi.extent.disk_choices and vm.devices.disk_choices
so that they are aware of each other, and in-use iscsi extents
don't get presented as options for VM disks and vice-versa.

### DIFF
--- a/src/middlewared/middlewared/plugins/iscsi.py
+++ b/src/middlewared/middlewared/plugins/iscsi.py
@@ -8,6 +8,7 @@ from middlewared.service import CallError, CRUDService, private, ServiceChangeMi
 import middlewared.sqlalchemy as sa
 from middlewared.utils import run
 from middlewared.utils.path import is_child
+from middlewared.utils.size import format_size
 from middlewared.validators import Range
 
 import bidict
@@ -833,34 +834,20 @@ class iSCSITargetExtentService(SharingService):
         """
         diskchoices = {}
 
-        zvol_query_filters = [('type', '=', 'DISK')]
-
-        used_zvols = [
-            i['path'] for i in (await self.query(zvol_query_filters))
-        ]
-
         zvols = await self.middleware.call(
-            'pool.dataset.query', [
-                ('type', '=', 'VOLUME'),
-                ('locked', '=', False)
-            ]
+            'zfs.dataset.unlocked_zvols_fast',
+            [['attachment', '=', None]], {},
+            ['SIZE', 'RO', 'ATTACHMENT']
         )
-        zvol_list = []
-        for zvol in zvols:
-            zvol_name = zvol['name']
-            zvol_size = zvol['volsize']['value']
-            zvol_list.append(zvol_name)
-            key = os.path.relpath(zvol_name_to_path(zvol_name), '/dev')
-            if key not in used_zvols:
-                diskchoices[key] = f'{zvol_name} ({zvol_size})'
 
-        zfs_snaps = await self.middleware.call(
-            'zfs.snapshot.query',
-            [['dataset', 'in', [zvol['name'] for zvol in zvols]]],
-            {'select': ['name']},
-        )
-        for snap in zfs_snaps:
-            diskchoices[os.path.relpath(zvol_name_to_path(snap['name']), '/dev')] = f'{snap["name"]} [ro]'
+        for zvol in zvols:
+            key = os.path.relpath(zvol['path'], '/dev')
+            if zvol['ro']:
+                description = f'{zvol["name"]} [ro]'
+            else:
+                description = f'{zvol["name"]} ({format_size(zvol["size"])})'
+
+            diskchoices[key] = description
 
         return diskchoices
 
--- a/src/middlewared/middlewared/plugins/vm/vm_devices.py
+++ b/src/middlewared/middlewared/plugins/vm/vm_devices.py
@@ -4,7 +4,7 @@ import re
 
 import middlewared.sqlalchemy as sa
 
-from middlewared.plugins.zfs_.utils import zvol_name_to_path, zvol_path_to_name
+from middlewared.plugins.zfs_.utils import zvol_path_to_name
 from middlewared.schema import accepts, Bool, Dict, Error, Int, Patch, returns, Str
 from middlewared.service import CallError, CRUDService, private, ValidationErrors
 from middlewared.utils import run
@@ -55,14 +55,17 @@ class VMDeviceService(CRUDService):
         """
         Returns disk choices for device type "DISK".
         """
-        return {
-            zvol_name_to_path(vol['id']): vol['id']
-            for vol in await self.middleware.call(
-                'pool.dataset.query', [['type', '=', 'VOLUME'], ['locked', '=', False]], {
-                    'extra': {'properties': ['encryption', 'keystatus']}
-                }
-            )
-        }
+        out = {}
+        zvols = await self.middleware.call(
+            'zfs.dataset.unlocked_zvols_fast',
+            [["OR", [["attachment", "=", None], ["attachment.method", "=", "vm.devices.query"]]]],
+            {}, ['ATTACHMENT']
+        )
+
+        for zvol in zvols:
+            out[zvol['path']] = zvol['name']
+
+        return out
 
     @private
     async def create_resource(self, device, old=None):
--- a/src/middlewared/middlewared/plugins/zfs.py
+++ b/src/middlewared/middlewared/plugins/zfs.py
@@ -7,13 +7,15 @@ from copy import deepcopy
 
 import libzfs
 
-from middlewared.schema import accepts, Any, Bool, Dict, Int, List, Str
+from middlewared.plugins.zfs_.utils import zvol_path_to_name, unlocked_zvols_fast
+from middlewared.schema import accepts, Any, Bool, Dict, Int, List, Ref, Str
 from middlewared.service import (
     CallError, CRUDService, ValidationErrors, filterable, job, private,
 )
 from middlewared.utils import filter_list, filter_getattrs
 from middlewared.utils.path import is_child
 from middlewared.validators import Match, ReplicationSnapshotNamingSchema
+
 
 SEARCH_PATHS = ['/dev/disk/by-partuuid', '/dev']
 
@@ -578,6 +580,42 @@ class ZFSDatasetService(CRUDService):
             }
             for dataset in self.query()
         ]
+
+    @accepts(
+        Ref('query-filters'),
+        Ref('query-options'),
+        List(
+            'additional_information',
+            items=[Str('desideratum', enum=['SIZE', 'RO', 'DEVID', 'ATTACHMENT'])]
+        )
+    )
+    def unlocked_zvols_fast(self, filters, options, additional_information):
+        """
+        Fast check for zvol information. Supports `additional_information` to
+        expand output on an as-needed basis. Adding additional_information to
+        the output may impact performance of 'fast' method.
+        """
+        def get_attachments():
+            extents = self.middleware.call_sync('iscsi.extent.query', [('type', '=', 'DISK')])
+            iscsi_zvols = {
+                i['path'][len('zvol/'):]: i for i in extents
+            }
+
+            vm_devices = self.middleware.call_sync('vm.device.query', [['dtype', '=', 'DISK']])
+            vm_zvols = {
+                zvol_path_to_name(i['attributes']['path']): i for i in vm_devices
+            }
+            return {
+                'iscsi.extent.query': iscsi_zvols,
+                'vm.devices.query': vm_zvols
+            }
+
+        data = {}
+        if 'ATTACHMENT' in additional_information:
+            data['attachments'] = get_attachments()
+
+        zvol_list = list(unlocked_zvols_fast(additional_information, data).values())
+        return filter_list(zvol_list, filters, options)
 
     def common_load_dataset_checks(self, ds):
         self.common_encryption_checks(ds)
--- a/src/middlewared/middlewared/plugins/zfs_/utils.py
+++ b/src/middlewared/middlewared/plugins/zfs_/utils.py
@@ -8,6 +8,12 @@ logger = logging.getLogger(__name__)
 __all__ = ["zvol_name_to_path", "zvol_path_to_name"]
 
 
+class ZFSCTL(enum.IntEnum):
+    # from include/os/linux/zfs/sys/zfs_ctldir.h in ZFS repo
+    INO_ROOT = 0x0000FFFFFFFFFFFF
+    INO_SNAPDIR = 0x0000FFFFFFFFFFFD
+
+
 def zvol_name_to_path(name):
     return os.path.join("/dev/zvol", name.replace(" ", "+"))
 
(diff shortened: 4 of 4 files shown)


src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -833,34 +833,22 @@ async def disk_choices(self):
...
-            key = os.path.relpath(zvol_name_to_path(zvol_name), '/dev')
-            if key not in used_zvols:
-                diskchoices[key] = f'{zvol_name} ({zvol_size})'
-
-        zfs_snaps = await self.middleware.call(
-            'zfs.snapshot.query',
-            [['dataset', 'in', [zvol['name'] for zvol in zvols]]],
-            {'select': ['name']},
+            'zfs.dataset.available_zvols_fast',
+            {"additional_information": ["SIZE", "RO", "ATTACHMENT"]}
```
sonicaj, line 838: Just a nit, but can we use single quotes keeping consistent from above ;) ?

sonicaj: Just to be sure, this is not going to get locked zvols ?

src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -833,34 +833,22 @@ async def disk_choices(self):
...
-        zvol_list = []
-        for zvol in zvols:
-            zvol_name = zvol['name']
-            zvol_size = zvol['volsize']['value']
-            zvol_list.append(zvol_name)
-            key = os.path.relpath(zvol_name_to_path(zvol_name), '/dev')
-            if key not in used_zvols:
-                diskchoices[key] = f'{zvol_name} ({zvol_size})'
-
-        zfs_snaps = await self.middleware.call(
```
sonicaj, line 857: We are not getting this anymore i believe ?

src/middlewared/middlewared/plugins/vm/vm_devices.py
```diff
@@ -55,14 +55,19 @@ async def disk_choices(self):
...
-            for vol in await self.middleware.call(
-                'pool.dataset.query', [['type', '=', 'VOLUME'], ['locked', '=', False]], {
-                    'extra': {'properties': ['encryption', 'keystatus']}
-                }
-            )
-        }
+        out = {}
+        zvols = await self.middleware.call(
+            'zfs.dataset.available_zvols_fast',
+            {"additional_information": ["ATTACHMENT"]}
```
sonicaj, line 61: quotes usage

src/middlewared/middlewared/plugins/vm/vm_devices.py
```diff
@@ -55,14 +55,19 @@ async def disk_choices(self):
...
+        )
+
+        for zvol in zvols.values():
+            if zvol['attachment']:
```
sonicaj, line 65: It should be fine to use the same volume in multiple vms as long as it's not happening together, so i think we should not have this here.

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -586,6 +588,116 @@ def query_for_quota_alert(self):
             for dataset in self.query()
         ]
 
+    def available_zvols_fast(self, options=None):
```
sonicaj, line 591: How about perhaps adding an `accepts` decorator which can do some schema validation to ensure devs don't make mistakes when calling this internally ?

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -586,6 +588,116 @@ def query_for_quota_alert(self):
...
+                """
+                Normalize all places where we store zvol info so that
+                indexed by dataset name for quick lookup
+                """
+                extents = self.middleware.call_sync('iscsi.extent.query', [('type', '=', 'DISK')])
+                iscsi_zvols = {
+                    i['path'][iscsi_offset:]: i for i in extents
+                }
+
+                vm_devices = self.middleware.call_sync('vm.device.query', [["dtype", "=", "disk"]])
```
sonicaj, line 657: `disk` should be `DISK`

sonicaj: Also let's please fix the quotes usage

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -586,6 +588,116 @@ def query_for_quota_alert(self):
...
+                if not files:
+                    continue
+
+                for file in files:
+                    path = root + "/" + file
+                    zvol_name = path[offset:]
+                    dev_name = os.readlink(path).split('/')[-1]
+
+                    out.update({
+                        zvol_name: {
```
sonicaj, line 676: Btw just to be sure, how are we handling the spaces in the zvol name here ? Does it show them or does it use the escaped version ?

src/middlewared/middlewared/plugins/iscsi.py
anodos325: I tried it and locked volumes don't get presented over /dev/zvol, IMHO it would be a pretty significant bug if they were.

src/middlewared/middlewared/plugins/vm/vm_devices.py
anodos325: You mean it's safe as long as multiple VMs referencing it aren't running at the same time? Do we have validation to prevent this?

src/middlewared/middlewared/plugins/zfs.py
themylogin, line 591: If you add `@filterable` and return a list instead of a dict, you'll be able to call it like `call("zfs.pool.available_zvols_fast", [["attachment","=",None]])`

Also why `available_zvols_fast`? It seems that it can query both available and unavailable zvols ("available" in terms on being or not being used by ISCSI/VM, or is there a different meaning here?). Should we name it `query_zvols_fast`?

Return schema can be made compatible with `zfs.pool.query` to allow using it as a drop-in replacement.

src/middlewared/middlewared/plugins/vm/vm_devices.py
sonicaj: Yup we should iirc, and if it's not happening - i can take care of that bit ;)

src/middlewared/middlewared/plugins/iscsi.py
```diff
@@ -833,34 +833,22 @@ async def disk_choices(self):
...
-        zvol_query_filters = [('type', '=', 'DISK')]
-
-        used_zvols = [
-            i['path'] for i in (await self.query(zvol_query_filters))
-        ]
-
         zvols = await self.middleware.call(
-            'pool.dataset.query', [
-                ('type', '=', 'VOLUME'),
-                ('locked', '=', False)
```
themylogin, line 845: How does this new approach handle locked datasets?

themylogin (commented):
Great job! I have a few suggestions:

src/middlewared/middlewared/plugins/vm/vm_devices.py
yocalebo: Why would we allow configuration of the same zvol to 2 different VMs but disallow the starting of both VMs at the same time?

sonicaj: So that the same disk can be used by both VMs ( not concurrently ) ?

yocalebo: That's my point....why allow the config if the VMs can't be powered on concurrently? That seems like an edge-case. Even more so, the practicality of powering on a VM, writing data to the zvol, powering it down, then powering on the other one seems....strange.

sonicaj: I don't understand ? So you are saying if i want to read data from a single disk on both VMs ( not concurrently ), i should powerdown the first one, remove the device from the first one and then attach the disk to the new VM and power it up and then repeat the process if i want to access vm 1 again ?

yocalebo: I'm saying that allowing a zvol to be configured to 2 different VMs makes very little sense and I don't see any practical use-case for it.

sonicaj: I don't see why we should explicitly disable that workflow also given it's already working ? So we are also changing behaviour in this case for no good reason to the best of my understanding ?

yocalebo: Because it's setting the user up for data corruption. It doesn't matter if the workflow works right now or not. Giving the user a loaded gun without any safeguards is the incorrect way to design something.

sonicaj: I don't follow, if the zvol is not concurrently being used by VMs, how is that causing data corruption ?

yocalebo: How are you ensuring the data being written to the zvol is flushed to disk? How are you ensuring the zvol that is moved the other VM isn't formatted and overwritten instead of being mounted?

sonicaj: For the second question, if a user is doing this he's managing the formatting part as he can't access it otherwise.

For the first question, how are your changes going to prevent the issue described ? User can still add the device to the second VM ( btw VMs are not running at this point, same applies to starting the other VM having same zvol ), how do we know the data has been flushed at that point to the disk ?

If you still think we should change the existing workflow, sure :)

yocalebo: When a VM is powered down, the OS is responsible for flushing data to the disk before power off so it's "guaranteed" without any user intervention. So I'm coming full circle again to my original point. What is the practical use-case when VMs have to be powered down before making any changes?

1. Create VM1
2. Create Zvol1
3. Attach Zvol1 to VM1
4. VM1 makes changes to Zvol1
5. VM1 is powered down
6. Zvol1 is attached to VM2
7. User has to go into guest OS and mount the zvol inside the VM
8. Changes are made
9. VM2 is powered down
10. Zvol1 is changed to VM1
11. Etc etc

Is that really practical?

sonicaj: So if on VM shutdown it's ensured data is flushed, we won't run into the case of data corruption as the second VM won't use the zvol unless the first one has been shutdown ? We should be good ?

yocalebo: Yes, if the user can't change a zvol on a VM while it's powered on, then we don't have to worry about data being flushed.

src/middlewared/middlewared/plugins/zfs.py
anodos325: available as in being available via /sys/block. I.e. unlocked on an imported pool.

src/middlewared/middlewared/plugins/iscsi.py
anodos325:
```
root@truenas[~]# midclt call zfs.dataset.available_zvols_fast | jq
{
  "dozer/TESTVOL@now": {
    "name": "dozer/TESTVOL@now",
    "dev": "zd32"
  },
  "dozer/TESTVOL": {
    "name": "dozer/TESTVOL",
    "dev": "zd0"
  },
  "dozer/smb-vss/sub1/zvol02": {
    "name": "dozer/smb-vss/sub1/zvol02",
    "dev": "zd16"
  }
```

They are returned _if_ the ZFS snapdev is visible. I'm not sure how this was supposed to work with a hidden snapdev on Linux (probably doesn't since it's not available through /sys/block).

src/middlewared/middlewared/plugins/iscsi.py
anodos325: Locked datasets aren't available through /dev/zvol (handled by ZFS).

src/middlewared/middlewared/plugins/zfs.py
themylogin: Can `unlocked` be a better term? `query_unlocked_zvols_fast`

anodos325: Changed to "unlocked".

yocalebo: "available" makes more sense to me since zvols can be unmounted

src/middlewared/middlewared/plugins/zfs.py
anodos325: Improved output to clearly differentiate between `name` and `path`:

```
root@truenas[~]# midclt call zfs.dataset.unlocked_zvols_fast | jq
{
  "tank space/zvol with lots of spaces and trailing space ": {
    "path": "/dev/zvol/tank+space/zvol+with+lots+of+spaces+and+trailing+space+",
    "name": "tank space/zvol with lots of spaces and trailing space ",
    "dev": "zd48"
  },
```

VM disk choices:
```
root@truenas[~]# midclt call vm.device.disk_choices | jq
{
  "/dev/zvol/tank+space/zvol+with+lots+of+spaces+and+trailing+space+": "tank space/zvol with lots of spaces and trailing space ",
```

ISCSI disk choices:
```
root@truenas[~]# midclt call iscsi.extent.disk_choices | jq
{
  "zvol/tank+space/zvol+with+lots+of+spaces+and+trailing+space+": "tank space/zvol with lots of spaces and trailing space  (1.02 MiB)",
  ```

File: src/middlewared/middlewared/plugins/iscsi.py (hunk 2 above)
Comment (anodos325) [reply src/middlewared/middlewared/plugins/iscsi.py:857]: Created NAS-116482 for someone else to investigate interaction between snapdev property and exposing snapshots of zvols.

File: src/middlewared/middlewared/plugins/zfs.py (hunk 7 above)
Comment (anodos325) [reply src/middlewared/middlewared/plugins/zfs.py:679]: Confirmed out-of-band that above is correct.

File: src/middlewared/middlewared/plugins/iscsi.py [hunk 9]
```diff
@@ -833,34 +834,22 @@ async def disk_choices(self):
...
-            [['dataset', 'in', [zvol['name'] for zvol in zvols]]],
-            {'select': ['name']},
+            'zfs.dataset.unlocked_zvols_fast',
+            {'additional_information': ['SIZE', 'RO', 'ATTACHMENT']}
         )
-        for snap in zfs_snaps:
-            diskchoices[os.path.relpath(zvol_name_to_path(snap['name']), '/dev')] = f'{snap["name"]} [ro]'
+
+        for zvol in zvols.values():
+            if zvol['attachment'] is not None:
```
themylogin, line 843: You could take advantage of a filterable interface here

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -586,6 +588,123 @@ def query_for_quota_alert(self):
...
+                    if do_get_ro is True:
+                        out[zvol_name]['ro'] = get_ro(dev_name)
+
+                    if do_get_attachment:
+                        out[zvol_name]['attachment'] = get_attachment(zvol_name, attachment_data)
+
+            return out
+
+        opts = options or {}
+        info_level = opts.get('additional_information', [])
```
themylogin, line 693: Having an `@accepts` decorator with `Dict` schema will reduce the number of these `or {}` and `.get(..., [])` and prevent typos in key names or enum values

yocalebo (approved):

yocalebo:
retest this please

src/middlewared/middlewared/plugins/zfs.py
anodos325: Switched to list. Put another arg in accepts decorator for the filterable version.

src/middlewared/middlewared/plugins/iscsi.py
anodos325: Same filtering goes on in either one function or another. I was returning dict so that I can get O(1) lookups for info based on dataset names. This is more important for me in future function to do same for datasets. ZFS objset kstats don't differentiate between whether something is a zvol or filesystem.

themylogin (approved):

PR #9053, opened 2022-05-31
NAS-116477 / 22.12 / Add tests for validating file permissions check works as intended

sonicaj:


anodos325:
What about testing combination of:
/mnt/dozer/testdir/file

where testdir is owned by root:wheel with perms of 0700, but file is 0777. Permissions check should fail due to lack of execute on parent dir.

anodos325 (approved):

PR #9058, opened 2022-05-31
NAS-116486 / 22.12 / Simplify process of adding NFS SPNs

anodos325:
Upstream tooling for this is better now. add_update_ads updates
local keytab and also updates compuer object in AD.


anodos325:
Can be safely backported to 13 because we use samba 4.15 there.

yocalebo (approved):

PR #9062, opened 2022-05-31
NAS-116495 / 22.12 / convert most of sysdataset plugin to sync

anodos325:
There were quite a few IO calls being made outside of threads.

### DIFF
--- a/src/middlewared/middlewared/plugins/sysdataset.py
+++ b/src/middlewared/middlewared/plugins/sysdataset.py
@@ -732,9 +732,10 @@ async def setup(middleware):
             if os.path.exists('/var/cache/nscd'):
                 shutil.rmtree('/var/cache/nscd')
 
-            os.makedirs('/tmp/cache/nscd', exist_ok=True)
+            os.makedirs('/var/run/nscd/cache', exist_ok=True)
+
         if not os.path.islink('/var/cache/nscd'):
-            os.symlink('/tmp/cache/nscd', '/var/cache/nscd')
+            os.symlink('/var/run/nscd/cache', '/var/cache/nscd')
 
     middleware.register_hook('pool.post_create', pool_post_create)
     # Reconfigure system dataset first thing after we import a pool.
--- a/src/middlewared/middlewared/plugins/sysdataset.py
+++ b/src/middlewared/middlewared/plugins/sysdataset.py
@@ -2,7 +2,7 @@ from middlewared.schema import accepts, Bool, Dict, Int, returns, Str
 from middlewared.service import CallError, ConfigService, ValidationErrors, job, private
 from middlewared.service_exception import InstanceNotFound
 import middlewared.sqlalchemy as sa
-from middlewared.utils import Popen, run
+from middlewared.utils import filter_list, MIDDLEWARE_RUN_DIR
 from middlewared.utils.size import format_size
 from middlewared.plugins.boot import BOOT_POOL_NAME_VALID
 
@@ -12,14 +12,14 @@ except ImportError:
     CTDBConfig = None
 
 import asyncio
-from contextlib import asynccontextmanager
+from contextlib import contextmanager, suppress
 import errno
 import json
 import os
 from pathlib import Path
-import psutil
 import shutil
 import subprocess
+import threading
 import uuid
 
 SYSDATASET_PATH = '/var/db/system'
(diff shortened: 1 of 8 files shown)


anodos325:
Noticed while looking at some async path validator code.

themylogin (changes requested):
Sorry, but it looks so messy now... I don't see why these methods should be asynchronous in the first place, can we just make them entirely synchronous?

anodos325:
> Sorry, but it looks so messy now... I don't see why these methods should be asynchronous in the first place, can we just make them entirely synchronous?

Yeah, that's fair. I'll make them synchronous.

anodos325:
> @anodos325 can you rebase this please before we review?

Hmm... we should have coordinated sysdataset changes while this was pending.

anodos325:
I will re-run CI to make sure that this isn't introducing new regressions.

anodos325:
@themylogin ready for review

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -244,129 +259,136 @@ async def destination_pool_error(self, new_pool):
...
+
+        if mounted_pool and mounted_pool.split('/')[0] != config['pool']:
             self.logger.debug('Abandoning dataset on %r in favor of %r', mounted_pool, config['pool'])
-            async with self._release_system_dataset():
-                await self.__umount(mounted_pool, config['uuid'])
-                await self.__setup_datasets(config['pool'], config['uuid'])
-                mounted = await self.__mount(config['pool'], config['uuid'])
+            with self._release_system_dataset():
+                self.__umount(mounted_pool, config['uuid'])
+                self.middleware.call_sync('systemdataset.setup_datasets', config['pool'], config['uuid'])
```
themylogin, line 311: Missing underscore in called method name.
Does this pass tests? If yes, we need the one that fails here.

themylogin: It's better not to have any underscore at all. `@private` and that's it.

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -244,129 +259,136 @@ async def destination_pool_error(self, new_pool):
...
                 [
+                    ['fs_type', '=', 'zfs'],
                     ['OR', [
-                        # Pools without encryption
-                        ['encrypted', '=', False],
-                        # Encrypted pools that are not locked
-                        ['locked', '=', False],
-                        # Passphrase-encrypted pools (system dataset will be unencrypted in this case)
-                        ['key_format.value', '=', 'PASSPHRASE'],
+                        ['mount_source', 'rnin', '/'],
```
themylogin, line 382: Previous code did not include key-encrypted pools, this code does.

themylogin (changes requested):
Great job!

src/middlewared/middlewared/plugins/sysdataset.py
anodos325: Previous did include mounted key-encrypted pools. I think in above case the only case that is now being excluded is unmounted passphrase-encrypted pools. I'm not sure if that's an intentional design feature, is it?

themylogin: We should be able to use a passphrase-encrypted pool for system dataset even if it is not unlocked yet.

anodos325: Okay. Fixing.

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -244,110 +316,142 @@ async def destination_pool_error(self, new_pool):
...
-            asyncio.ensure_future(self.middleware.call('service.restart', 'rrdcached'))
-            asyncio.ensure_future(self.middleware.call('service.restart', 'syslogd'))
+        return self.middleware.call_sync('systemdataset.config')
 
-            # The following should be backgrounded since they may be quite
-            # long-running.
-            await self.middleware.call('smb.configure', False)
+    @private
+    def query_pool_for_system_dataset(self, exclude_pool):
+        for p in self.middleware.call_sync('pool.query'):
```
yocalebo, line 437: zfs.pool.query_imported_fast instead of pool.query?

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -480,76 +607,93 @@ def __get_datasets(self, pool, uuid):
...
 
-        await self.__setup_datasets(_to, config['uuid'])
+        self.middleware.call_sync('systemdataset.setup_datasets', _to, config['uuid'])
 
         if _from:
-            path = '/tmp/system.new'
-            if not os.path.exists('/tmp/system.new'):
-                os.mkdir('/tmp/system.new')
+            path = f'{MIDDLEWARE_RUN_DIR}/system.new'
+            if not os.path.exists(f'{MIDDLEWARE_RUN_DIR}/system.new'):
```
yocalebo, line 623: os.mkdir(exist_ok=True)?

anodos325: That's for os.makedirs() rather than os.mkdir(). I wrapped in contextlib.suppress though.

src/middlewared/middlewared/plugins/sysdataset.py
anodos325: Sure. I also now have set `retrieve_children: False` which should also help speed things up.

yocalebo:
No unexpected regressions with integration tests on non-HA or HA systems. HA failover is still functional with these changes. I'm merging.

PR #9070, opened 2022-06-01
NAS-116481 / 22.12 / Validate zfs resource names

Qubad786:
## Context

Right now when we create a zfs resource i.e pool/dataset/snapshot, we do not validate it's name and it errors out when we create the specific resource.

## Solution

Use py-libzfs for getting the validation which zfs performs to determine if the name in question is valid or not.

### DIFF
--- a/src/middlewared/middlewared/plugins/zfs.py
+++ b/src/middlewared/middlewared/plugins/zfs.py
@@ -8,6 +8,7 @@ from copy import deepcopy
 import libzfs
 
 from middlewared.plugins.zfs_.utils import zvol_path_to_name, unlocked_zvols_fast
+from middlewared.plugins.zfs_.validation_utils import validate_snapshot_name
 from middlewared.schema import accepts, Any, Bool, Dict, Int, List, Ref, Str
 from middlewared.service import (
     CallError, CRUDService, ValidationErrors, filterable, job, private,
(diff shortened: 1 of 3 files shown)


pkgdemon:
retest this please

anodos325:
> @anodos325 the methods are not using libzfs handle ?

Indeed they're not. Sorry, I was looking at `zfs_validate_name()` vs `zfs_name_valid()`.

anodos325:
If we use `zfs_validate_name()` and `zpool_name_valid()` libzfs will print the exact reason for failure, but with the caveat I mentioned above about thread-safety.

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -1163,6 +1165,9 @@ def do_create(self, data):
                 if data[k]:
                     verrors.add(f'snapshot_create.{k}', 'This option is not supported when excluding datasets')
 
+        if name and not validate_snapshot_name(f'{dataset}@{name}'):
+            verrors.add('snapshot_create.name', 'Please specify a valid snapshot name')
```
themylogin, line 1169: Same note for error messages as before

themylogin (changes requested):

themylogin (approved):

PR #9072, opened 2022-06-01
NAS-116517 / 22.12 / fix blocking event loop in auth.py

yocalebo:
`psutil.net_connections` performs file IO which can block the main event loop.

### DIFF
--- a/src/middlewared/middlewared/plugins/auth.py
+++ b/src/middlewared/middlewared/plugins/auth.py
@@ -591,7 +591,7 @@ class TwoFactorAuthService(ConfigService):
         return pyotp.random_base32()
 
 
-def check_perms(app):
+def authenticate_connection(app):
     remote_addr, remote_port = get_remote_addr_port(app)
     if not (remote_addr.startswith('127.') or remote_addr == '::1'):
         return
@@ -609,8 +609,7 @@ def check_perms(app):
 
 async def check_permission(middleware, app):
     """
-    Authenticates connections coming from loopback and from
-    root user.
+    Authenticates connections coming from loopback and from root user.
     """
     sock = app.request.transport.get_extra_info('socket')
     if sock.family == socket.AF_UNIX:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/auth.py
```diff
@@ -591,17 +591,7 @@ def generate_base32_secret(self):
...
-    root user.
-    """
-    sock = app.request.transport.get_extra_info('socket')
-    if sock.family == socket.AF_UNIX:
-        # Unix socket is only allowed for root
-        AuthService.session_manager.login(app, UnixSocketSessionManagerCredentials())
-        return
-
+def authenticate_connection(app):
     remote_addr, remote_port = get_remote_addr_port(app)
```
themylogin, line 595: If we want to eliminate all nanodelays and achieve super turbo boost performance, then we don't need to perform this in a thread, only the `get_peer_process` bit.

themylogin (commented):
Can IO with sysfs block? For how long?

src/middlewared/middlewared/plugins/auth.py
yocalebo: It's not about "achieve super turbo boost performance", It's about doing what's right. Doing file IO in the main event loop is incorrect but you already know this, however, so I appreciate constructive reviews instead of reviews filled with sarcasm.

Also, if you look at the code, you'll see that we call `get_remote_addr_port` which can call `get_peer_process` in the `else` branch. Then we call `get_peer_process` again potentially.

yocalebo:
> Can IO with sysfs block? For how long?

If you looked at the ticket, you'll see the traceback being reported. So it's being blocked for at least >= 2 seconds

src/middlewared/middlewared/plugins/auth.py
rick-mesta: I think the whole point of doing things in a thread is precisely because we don't want to block for I/O in the main event loop.

rick-mesta (approved):

PR #9100, opened 2022-06-02
NAS-116631 / 22.02.2 / `wait_to_hang_and_dump_core` script

themylogin:


themylogin:
time 0:30

src/middlewared/middlewared/scripts/wait_to_hang_and_dump_core.py
```diff
@@ -0,0 +1,57 @@
...
+        ).group(1),
+    )
+    logger.info("middlewared PID: %d", middlewared_pid)
+
+    subprocess.run(
+        ["gdb", "-p", str(middlewared_pid), "-batch", "-ex", "generate-core-file"],
+        check=True,
+    )
+
+    core_file = sorted(pathlib.Path.cwd().glob("core.*"), key=os.path.getmtime, reverse=True)[0]
```
yocalebo, line 49: If another program core dumps after this is finished and before we `gzip` it then we'll get the wrong core file right? The naming of the core file should have some identifiable information in it so we get the middlewared core dump.

yocalebo (approved):

PR #9101, opened 2022-06-02
NAS-116257 / 13.0 / Fix counting all snapshots on the system

themylogin:
No SCALE backport is needed


themylogin:
time 0:20

anodos325:
On FreeBSD you can count snapshots via stat(2) on ZFS snapdirs. st_nlink.

PR #9071, opened 2022-06-01
NAS-116484 / 22.02.2 / optimize disk.sync_all on SCALE

yocalebo:
Same principles that were done in 13 here https://github.com/truenas/middleware/pull/8937 are applied here to SCALE.

Differences in time that I've measured:

Inserting 1255 _new_ disk entries into db OLD took 1:54.17
Inserting 1255 _new_ disk entries into db NEW takes 1:00.24 **(~62% faster)**

Updating 1255 _existing_ disk entries into db OLD took 20:04.96
Updating 1255 _existing_ disk entries into db NEW takes 1:20.05 **(~175% faster)**

We're still missing some optimizations on SCALE (like `enclosure.sync_disk`) but this is a good first step.

PR #9103, opened 2022-06-02
NAS-116544 / 22.12 / limit ix-syncdisks.service to 5mins

yocalebo:
When https://github.com/truenas/middleware/pull/9071 is merged, ix-syncdisks.service shouldn't take longer than ~240 seconds in the worst case scenario on the largest system we sell. Without this timeout, the service can block boot-up indefinitely.

After testing `disk.sync_all` on a system with ~1250 disks being partitioned and a part of the zpool, it took ~2:31 seconds to complete so 5mins is a good safe spot.

### DIFF
--- a/debian/debian/ix-syncdisks.service
+++ b/debian/debian/ix-syncdisks.service
@@ -10,6 +10,7 @@ Type=oneshot
 RemainAfterExit=yes
 ExecStart=midclt call --job true --job-print description disk.sync_all
 StandardOutput=null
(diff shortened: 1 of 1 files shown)


yocalebo:
backport

yocalebo:
backport

yocalebo:
ping @truenas/middleware updated comment after testing on real hardware. This should be good to go.

anodos325 (approved):

PR #9105, opened 2022-06-02
NAS-116545 / 22.12 / Add key to SMB share presets to indicate cluster-safety

anodos325:
Adds clarification for TC team which of presets should be available
for cluster shares.


anodos325:
> Is it used somehow?

In TrueCommand. sharing.smb.presets

themylogin (approved):
It would be nice if `presets` had `@returns` decorator like our other public methods do.

anodos325:
> It would be nice if `presets` had `@returns` decorator like our other public methods do.

Yeah, I'll do that next time I have free time to do such things.

PR #9111, opened 2022-06-02
NAS-115435 / 22.12 / Fix GSSAPI config in LDAP config

anodos325:
Ensure that we have proper realm name populated in LDAP config
files. In typical case we have a kerberos keytab and ticket
available, but if ticket has expired this may cause failure
to bind to the server.

### DIFF
--- a/src/freenas/usr/local/libexec/freenas-debug/ldap/ldap.sh
+++ b/src/freenas/usr/local/libexec/freenas-debug/ldap/ldap.sh
@@ -1,0 +1,121 @@
+#!/bin/sh
+#+
+# Copyright 2011 iXsystems, Inc.
+# All rights reserved
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted providing that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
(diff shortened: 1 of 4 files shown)


src/freenas/usr/local/libexec/freenas-debug/ldap/ldap.sh
```diff
@@ -0,0 +1,121 @@
...
+	#
+	section_header "LDAP Settings"
+	echo ${CONF} | jq
+	section_footer
+
+	#
+	#	Dump nsswitch.conf
+	#
+	section_header "${PATH_NS_CONF}"
+	sc "${PATH_NS_CONF}"
```
yocalebo, line 62: `${PATH_NS_CONF}` and `${PATH_KRB5_CONFIG}` seem to be a CORE variable?

yocalebo (changes requested):

src/freenas/usr/local/libexec/freenas-debug/ldap/ldap.sh
yocalebo: I'd probably just do `cat /etc/nslcd.conf` and `cat /etc/openldap/ldap.conf` to remove complexity.

anodos325: They're set in AD script.

yocalebo: Nope, I'm wrong

yocalebo (approved):

PR #9114, opened 2022-06-03
NAS-116521 / 22.12 / Add support for VM memory ballooning

Qubad786:
## Problem

User complained that guest ballooning was not working for him on windows even after he had installed necessary extensions.

## Solution

Ballooning also needs to be enabled from host side for it to properly work. For linux guests we do not require any extensions to be installed inside the guest and we can see memory being given back to the host as guest uses less resources.
We needed to specify minimum value which libvirt should provide to the guest always, if not provided it defaults to the memory defined which is the maximum memory which the guest can consume.


src/middlewared/middlewared/plugins/vm/supervisor/supervisor_base.py
```diff
@@ -246,6 +246,12 @@ def get_domain_children(self):
             # Command line args
             *self.commandline_xml(),
         ]
+        # Memory Ballooning - this will be memory which will always be allocated to the VM
+        # If not specified, this defaults to `memory`
+        if self.vm_data['min_memory']:
```
themylogin, line 251: Group this together with `memory` element into a separate method like `cpu_xml` please

themylogin (approved):

anodos325:
Jira ticket number here is wrong.

Qubad786:
@anodos325 I'll update the right ticket..Mixed ticket numbers

m9x3mos:
I see this was pulled in but I am not seeing any documentation on configuration for this new feature and I am not seeing it in the GUI.

Qubad786:
@m9x3mos created a ticket so this is available to be configured in the UI https://ixsystems.atlassian.net/browse/NAS-120289

PR #9132, opened 2022-06-03
Expose ZFS dosmodes in filesystem plugin

anodos325:
This is primarily to help with CI for SMB server, but may
be generally useful when we in future want to have a
proper file browser.


anodos325:
WIP currently. Need to add path validation and tests. For testing will plumb through SMB tests initially.

anodos325:
Flagging for some initial lookover.

src/middlewared/middlewared/plugins/filesystem_/dosmode.py
```diff
@@ -0,0 +1,65 @@
...
+    flags_in = 0
+    for flag, enabled in dosflags.items():
+        if not enabled:
+            continue
+        flags_in |= DOSFlag[flag.upper()]
+
+    fd = os.open(path, os.O_RDONLY)
+    try:
+        current_flags = get_dosflags_impl(path, fd) & ~DOSFlag.ALL
+        set_dosflags_impl(fd, path, flags_in | current_flags)
```
themylogin, line 57: How do you reset a flag in this case? If some flag is already set in `current_flags`, how will `|` operator manage to unset it?

themylogin (commented):
Where are those stored? xattrs?

anodos325:
> Where are those stored? xattrs?

They are stored in ZFS. Not userspace xattrs. ZFS has internal dosmode implementation. For instance, XAT_READONLY makes the file RO (enforced in zfs_zaccess()). XAT_ARCHIVE bit gets toggled if file is changed. 

This allows backup application to remove archive when it has been backed up, and then any FS change afterwards will reset it allowing a future run to quickly check this value to determine whether the file has been changed and requires fresh backup. (Same behavior as Windows).

FreeBSD exposes these through stat(2) and allows setting though chflags(2). Linux has no equivalent so we had to expose through custom ioctls.

src/middlewared/middlewared/plugins/filesystem_/dosmode.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def set_dosflags(path: str, dosflags: dict) -> None:
+    o_path_fd = os.open(path, os.O_PATH)
+    isdir = stat.S_ISDIR(os.fstat(o_path_fd).st_mode)
+    fd = pathref_reopen(o_path_fd, os.O_DIRECTORY if isdir else os.O_RDWR, close_fd=True)
+
+    try:
+        current = get_dosflags_impl(path, fd)
+        to_set = DOSFlag.flags_to_dict(current) | dosflags
```
themylogin, line 72: I still don't understand how does flag unsetting work? Or is it not supported? I also don't see testing flag unset in your test

themylogin:
> > Where are those stored? xattrs?
> 
> They are stored in ZFS. Not userspace xattrs. ZFS has internal dosmode implementation. For instance, XAT_READONLY makes the file RO (enforced in zfs_zaccess()). XAT_ARCHIVE bit gets toggled if file is changed.
> 
> This allows backup application to remove archive when it has been backed up, and then any FS change afterwards will reset it allowing a future run to quickly check this value to determine whether the file has been changed and requires fresh backup. (Same behavior as Windows).
> 
> FreeBSD exposes these through stat(2) and allows setting though chflags(2). Linux has no equivalent so we had to expose through custom ioctls.

Is there any sense in backing those up into cloud?

anodos325:
> > > Where are those stored? xattrs?
> > 
> > 
> > They are stored in ZFS. Not userspace xattrs. ZFS has internal dosmode implementation. For instance, XAT_READONLY makes the file RO (enforced in zfs_zaccess()). XAT_ARCHIVE bit gets toggled if file is changed.
> > This allows backup application to remove archive when it has been backed up, and then any FS change afterwards will reset it allowing a future run to quickly check this value to determine whether the file has been changed and requires fresh backup. (Same behavior as Windows).
> > FreeBSD exposes these through stat(2) and allows setting though chflags(2). Linux has no equivalent so we had to expose through custom ioctls.
> 
> Is there any sense in backing those up into cloud?

Yes, there is. I wish the OS team had done this via an xattr handler rather than ioctl. :/ If you feel that it's needed I can make a PR to add an xattr handler for them (so that you can read / set via `system.dosmode` xattr for example).

src/middlewared/middlewared/plugins/filesystem_/dosmode.py
anodos325: https://peps.python.org/pep-0584/
I believe that's the right pep
```
>>> dosmode = {"readonly": True, "system": False}
>>> dosmode | {"readonly": False}
{'readonly': False, 'system': False}
```

themylogin: Sorry, I thought those were still bit masks :(

anodos325:
> Do you propose to duplicate them to xattrs? I see no need for this, just want to prepare for resuming xattr cloud sync backup work

You can actually create synthetic xattrs in an xattr handler. Basically on xattr read, convert the ZFS znode pflags into something in the provided buffer, and on write convert it into the corresponding ZFS change. This is basically what I do with system.nfs4_acl_xdr xattr. It doesn't _actually_ exist (we're converting to/from internal ZFS ACL).

PR #9135, opened 2022-06-05
NAS-106532 / 22.02.2 / M50/M60 PLX Enclosures on SCALE

themylogin:
I waited 1.5 years for this


themylogin:
time 4:00

src/middlewared/middlewared/plugins/enclosure_/m50_plx.py
```diff
@@ -1,58 +1,48 @@
...
-                slot = int(m.group(1))
-            except IndexError:
+        addresses_to_slots = {
+            (slot / "address").read_text().strip(): slot.name
+            for slot in Path("/sys/bus/pci/slots").iterdir()
+        }
+
+        slot_to_nvme = {}
+        context = pyudev.Context()
+        for nvme_device in context.list_devices(subsystem="nvme"):
```
yocalebo, line 25: This for loop can be simplified quite a bit.
```
for nvme in filter(lambda x: x.parent.parent.get('PCI_SUBSYS_ID') == '10B5:8717', ctx.list_devices(subsystem='nvme')):
    if (slot := addrs_to_slots.get(nvme.parent.split('.')[0])) is not None:
        slot_to_nvme[int(slot.split('-', 1)[1])] = nvme.sys_name
```

However, if you choose to keep the logic in this loop then you're going to want to get the `DEVPATH` attribute via the `nvme_device.properties['DEVPATH']` because accessing it via `nvme_device` has been deprecated in py-udev and will be removed in py-udev version 1.0.0

yocalebo (changes requested):
woot woot, this is great! Just a couple of notes.

yocalebo (commented):
For some reason my other comments were lost so adding them now.

```
controller_path = Path(nvme_device["DEVPATH"]).parent.parent
port = pyudev.Devices.from_path(context, str(controller_path.parent))
if port.get("PCI_SUBSYS_ID") != "10B5:8717":
    continue
```
is the same as `nvme_device.parent.parent.get('PCI_SUBSYS_ID')` so you shouldn't need to create `pathlib.Path` objects.

yocalebo (commented):
Finally, since this is for M50 and M60 we should rename the module to `mseries_plx.py` (or `mseries_nvme.py` like I did for the `r50_nvme.py` module)

Also, we're returning the string `m50_plx_enclosure` on an M60 device so you can see what I did in the `r50_nvme.py` file since we have an r50 and an r50b system.

yocalebo:
Okay, this is really the final comment 😄 When I wrote the r50 nvme drive mappings, I was doing it based off pci addresses but those can change depending on what is plugged into the system. Instead, I'm using acpi handles which are more persistent.

I think we need to do the same here because we're basing this detection solely on the pci subsystem vendor ID which means if we start selling a different nvme drive this detection no longer works.

The acpi handle on the m50/m60 is `b'\\_SB_.PC03.BR3A'` so we could do something like so:
```
RE_NVME_PRIV = re.compile(r'nvme[0-9]+c')

for i in filter(lambda x: (x.attributes.get('path') or '') == b'\\_SB_.PC03.BR3A', ctx.list_devices(subsystem='acpi'):
    try:
        phys_node = Devices.from_path(ctx, i.sys_name + '/physical_node')
    except pyudev.DeviceNotFoundByNameError:
        self.logger.error('Failed to find pci slot information for rear nvme drives', exc_info=True)
    else:
        for idx, child in enumerate(phys_node.children, start=1):
            if child.sys_name.startswith('nvme') and not RE_NVME_PRIV.match(child.sys_name):
                slot_to_nvme[idx] = child.sys_name

themylogin:
@yocalebo I addressed your review, but kept my logic for PCI slots number, because the code you proposed will report sequential slot numbers even if we have NVMes inserted only to, e.g., slots 1 and 4

yocalebo (approved):
Fantastic!

PR #9138, opened 2022-06-06
Add cloudflare DDNS provider

jlsjonas:
Based on https://github.com/troglobit/inadyn/blob/master/examples/cloudflare-ipv4-only.conf

Note: Supported on inadyn 2.9.x (9 dev 2021), couldn't see the version used anywhere in the repo, please correct if I missed it!

### DIFF
--- a/src/middlewared/middlewared/plugins/dyndns.py
+++ b/src/middlewared/middlewared/plugins/dyndns.py
@@ -57,6 +57,7 @@ class DynDNSService(SystemServiceService):
         """
         return {
             'default@changeip.com': 'changeip.com',
+            'example.org': 'cloudflare.com',
             'default@cloudxns.net': 'cloudxns.net',
             'default@ddnss.de': 'ddnss.de',
             'default@dhis.org': 'dhis.org',


Ornias1993 (commented):
This also needs docs and/or UI changes, because, for example, the "username", "password" fields do not *actually* refer to what they're named after.

Most likely this is also why cloudflare is not listed.

rick-mesta:
https://jira.ixsystems.com/browse/NAS-116604

jlsjonas:
I agree that they don't map exactly to what they're referred to by Cloudflare, sadly enough (afaik?) I can't provide a PR for the UI 😬 
This could however be documented on https://www.truenas.com/docs/core/coretutorials/services/configuredynamicdns/ (I can provide a PR tomorrow if this could speed things up)

Thanks for considering!

k-obrien:
> I agree that they don't map exactly to what they're referred to by Cloudflare, sadly enough (afaik?) I can't provide a PR for the UI 😬

I think the UI you're referring to is [here](https://github.com/truenas/webui/tree/master/src/app/pages/services/components/service-dynamic-dns). You should be able to raise a PR. I'm keen to see this merged.

src/middlewared/middlewared/plugins/dyndns.py
```diff
@@ -57,6 +57,7 @@ async def provider_choices(self):
         """
         return {
             'default@changeip.com': 'changeip.com',
+            'example.org': 'cloudflare.com',
```
themylogin, line 60: It will render `provider example.org`, won't it?

themylogin:
Even unstable debian has inadyn 2.8.1 (this needs 2.9.1), do we want to bother with building custom inadyn package to get this feature early? @william-gr

william-gr:
@jlsjonas I am sorry but we cannot take this in this state nor move forward given the package is not available to us yet.

We advise you to use a docker container or an App for this purpose.

PR #9140, opened 2022-06-06
Properly wait for libvirt guests to terminate

Qubad786:
## Problem

When TrueNAS is shutdown, VM guests were not gracefully shutting down and instead were forcefully terminated even though libvirt service is running at that point ( when middleware is waiting for vms to gracefully shutdown ).

## Solution

We have `libvirt-guests` service which can be used to gracefully shutdown the VMs.


rick-mesta:
@themylogin / @sonicaj / @yocalebo / @anodos325: Team, could we please get some feedback to Rehan on this ? This still needs to be backported to `22.02.2` 🙏🏼

sonicaj (approved):

PR #9183, opened 2022-06-09
NAS-116659 / 22.02.2 / Fix iSCSI disk tests

anodos325:
Output format strings are slightly different from what they
originally were. Additionaly zvol snapdev needs to be visible
in order for snapshots of zvols to be available as choices


ericbsd (approved):
LGTM

yocalebo (approved):

PR #9192, opened 2022-06-09
NAS-116212 / 22.02.3 / Do not allow to SMART tests to be scheduled at the same time

themylogin:
@william-gr please evaluate the true target version of these changes


themylogin:
time 14:30

william-gr:
@themylogin definitely not 13...
Lets do it for 22.02.3 if you think its safe, otherwise Bluefin only

themylogin:
backport

PR #9199, opened 2022-06-10
Revert "Do not show SCALE trains in 12.0-U2"

themylogin:
This reverts commit 8e741801255ad2938096d2c72d9d0352e61d999d.

### DIFF
--- a/src/middlewared/middlewared/plugins/update_/trains_freebsd.py
+++ b/src/middlewared/middlewared/plugins/update_/trains_freebsd.py
@@ -116,8 +116,8 @@ class UpdateService(Service):
                 'sequence': train.LastSequence(),
             }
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/update_/trains_freebsd.py
```diff
@@ -116,8 +116,8 @@ def get_trains_data(self):
                 'sequence': train.LastSequence(),
             }
 
-        # scale_trains = self.middleware.call_sync('update.get_scale_trains_data')
-        # trains.update(**scale_trains['trains'])
+        scale_trains = self.middleware.call_sync('update.get_scale_trains_data')
```
william-gr, line 119: To confirm, is this only returning trans for CORE and not Enterprise?

PR #9200, opened 2022-06-10
NAS-116683 / 22.12 / Add new helper function to convert mountinfo to dict

anodos325:
We were using libzfs to convert paths to dataset name. Internally
libzfs does this thorugh stat() on the path and looking up
mount entries. We can instead do this ourselves without having
to use a libzfs handle.

To facilitate this add a new general-purpose method to return
a dict containing mount information indexed by devid. The
reason to index this way is because it allows efficient lookup
of an entry based on stat / lstat output. Refactor
filesystem.statvfs to use this as well.

### DIFF
--- a/src/middlewared/middlewared/plugins/zfs.py
+++ b/src/middlewared/middlewared/plugins/zfs.py
@@ -15,6 +15,7 @@ from middlewared.service import (
 )
 from middlewared.utils import filter_list, filter_getattrs
 from middlewared.utils.path import is_child
+from middlewared.utils.osc import getmntinfo
 from middlewared.validators import Match, ReplicationSnapshotNamingSchema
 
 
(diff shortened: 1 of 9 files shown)


anodos325:
Will convert filesystem.stat() plugin to use statx in future PR. statx gives some nice benefits like returning btime.

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -630,12 +632,14 @@ def common_encryption_checks(self, ds):
...
+        st = os.lstat(path)
+        if stat.S_ISLNK(st.st_mode):
+            raise CallError(f'{path}: path is a symbolic link')
+
+        mntinfo = getmntinfo(st.st_dev)[st.st_dev]
+        ds_name = mntinfo['mount_source']
+        if mntinfo['fs_type'] != 'ZFS':
+            raise CallError(f'{path}: not a ZFS filesystem')
 
         if ds_name is not None:
```
sonicaj, line 644: Is this possible ?

sonicaj (commented):
LGTM, can we add some integration tests and unit tests as well please ? Thanks!

anodos325:
> LGTM, can we add some integration tests and unit tests as well please ? Thanks!

We have coverage via filesystem.statfs tests (which wrap around this). But I can add more explicit tests for path_to_dataset.

anodos325:
f_fsid is unsigned long and so just presenting as string here.

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -627,19 +628,27 @@ def common_encryption_checks(self, ds):
         if not ds.encrypted:
             raise CallError(f'{id} is not encrypted')
 
-    def path_to_dataset(self, path):
+    def path_to_dataset(self, path, raise_=False):
```
sonicaj, line 631: We are not using `raise_` anywhere, let's please remove it and it's usages

src/middlewared/middlewared/pytest/unit/utils/test_mountinfo.py
```diff
@@ -0,0 +1,89 @@
...
+    assert mntent['device_id'] == {'major': 0, 'minor': 77, 'dev_t': 77}
+    assert mntent['root'] == '/'
+    assert mntent['mountpoint'] == '/mnt/tank space /Dataset With a space'
+    assert mntent['mount_opts'] == ['RW', 'NOATIME']
+    assert mntent['fs_type'] == 'zfs'
+    assert mntent['mount_source'] == 'tank space /Dataset With a space'
+    assert mntent['super_opts'] == ['RW', 'XATTR', 'POSIXACL']
+
+
+def test__getmntinfo():
```
sonicaj, line 86: Would be nice if we tested the actual output of the contents as well to ensure that they actually match to what we are expecting and not just that we were able to parse the fake file.

sonicaj (approved):

PR #9203, opened 2022-06-12
NAS-116278 / 22.12 / Allow users to specify authorized networks for iscsi targets

sonicaj:
This PR adds changes allowing users to specify authorized networks which will be authorized to access a certain target.
Earlier we used to have this in initiator level where with initiator + auth networks, we could manage this with targets but scst does not offer us that and we can only allow/deny connections at target level.

### DIFF
--- a/src/middlewared/middlewared/plugins/iscsi_/targets.py
+++ b/src/middlewared/middlewared/plugins/iscsi_/targets.py
@@ -19,7 +19,7 @@ class iSCSITargetModel(sa.Model):
     iscsi_target_name = sa.Column(sa.String(120), unique=True)
     iscsi_target_alias = sa.Column(sa.String(120), nullable=True, unique=True)
     iscsi_target_mode = sa.Column(sa.String(20), default='iscsi')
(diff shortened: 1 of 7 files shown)


src/middlewared/middlewared/plugins/iscsi_/targets.py
```diff
@@ -19,6 +19,7 @@ class iSCSITargetModel(sa.Model):
     iscsi_target_name = sa.Column(sa.String(120), unique=True)
     iscsi_target_alias = sa.Column(sa.String(120), nullable=True, unique=True)
     iscsi_target_mode = sa.Column(sa.String(20), default='iscsi')
+    iscsi_target_auth_networks = sa.Column(sa.JSON(), default=[])
```
themylogin, line 22: `default` is not needed here. Also `default=[]` is dangerous for the same reason as in function args list.

themylogin (approved):
Actually, you can add integration tests for this: insert various targets and assert allow/deny files contents

PR #9204, opened 2022-06-12
NAS-116559 / 22.12 / Add help text for application names regex

Qubad786:
## Problem

Right now invalid application name fails with reporting valid regex to the user however that is not helpful.

## Solution

Add an explanation as to what is allowed and not allowed so user has a better idea of what he can use for app name.


sonicaj (approved):

Qubad786:
backport

PR #9206, opened 2022-06-13
NAS-116701 / Fix VM Raw file support

dberlin:
RAW file support is broken because it creates a domain XML element with the
wrong attribute (dev rather than file). This fixes it to create the correct
XML element.


sonicaj (approved):
Thank you @dberlin for your contribution!

sonicaj:
backport

PR #9208, opened 2022-06-13
NAS-116704 / 22.12 / Fix debug size

themylogin:
1) Truncate uncompressed log files by bytes count, not by lines count.
2) Increase bytes count limit (middleware logs are often truncated to a too short size)
3) Also truncate compressed log files

Some guy had a 380M debug, most of it were OOM-killer stack traces happening every 5 seconds or so, gzipped into kern.log.


themylogin:
time 1:00

sonicaj (approved):

PR #9212, opened 2022-06-14
Add (cli-only) support for switching io type for VM storage devices

dberlin:
QEMU supports a number of io types for storage devices some of which can have very significant affect on performance. 
For example, io_uring is *much* faster in practice and overhead for my windows guests, both on files and zvols (particularly on zvols).  My guests get io_uring gets 2x the throughput on 4k random reads regardless of queue depth or thread settings.

This commit simply enables one to set the io type, it does not attempt to change the default.

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/devices/storage_devices.py
+++ b/src/middlewared/middlewared/plugins/vm/devices/storage_devices.py
@@ -86,4 +86,5 @@ class DISK(StorageDevice):
     )
 
     def create_source_element(self):
+        iotype = self.data['attributes']['iotype']
         return create_element('source', dev=self.data['attributes']['path'], io=iotype.lower())
--- a/src/middlewared/middlewared/plugins/vm/devices/storage_devices.py
+++ b/src/middlewared/middlewared/plugins/vm/devices/storage_devices.py
@@ -22,11 +22,12 @@ class StorageDevice(Device):
         virtio = self.data['attributes']['type'] == 'VIRTIO'
         logical_sectorsize = self.data['attributes']['logical_sectorsize']
         physical_sectorsize = self.data['attributes']['physical_sectorsize']
+        iotype = self.data['attributes']['iotype']
 
         return create_element(
             'disk', type=self.TYPE, device='disk', attribute_dict={
                 'children': [
-                    create_element('driver', name='qemu', type='raw', cache='none'),
+                    create_element('driver', name='qemu', type='raw', cache='none', io=iotype),
                     self.create_source_element(),
                     create_element(
                         'target', bus='sata' if not virtio else 'virtio',
@@ -61,6 +62,7 @@ class RAW(StorageDevice):
         Int('size', default=None, null=True),
         Int('logical_sectorsize', enum=[None, 512, 4096], default=None, null=True),
         Int('physical_sectorsize', enum=[None, 512, 4096], default=None, null=True),
+        Str('iotype', enum=['native', 'threads', 'io_uring'], default='native'),
     )
 
     def create_source_element(self):
@@ -80,6 +82,7 @@ class DISK(StorageDevice):
         Int('zvol_volsize'),
         Int('logical_sectorsize', enum=[None, 512, 4096], default=None, null=True),
         Int('physical_sectorsize', enum=[None, 512, 4096], default=None, null=True),
+        Str('iotype', enum=['native', 'threads', 'io_uring'], default='native'),
     )
 
     def create_source_element(self):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/vm/devices/storage_devices.py
```diff
@@ -61,6 +62,7 @@ class RAW(StorageDevice):
         Int('size', default=None, null=True),
         Int('logical_sectorsize', enum=[None, 512, 4096], default=None, null=True),
         Int('physical_sectorsize', enum=[None, 512, 4096], default=None, null=True),
+        Str('iotype', enum=['native', 'threads', 'io_uring'], default='native'),
```
themylogin, line 65: Our enums are uppercase

src/middlewared/middlewared/plugins/vm/devices/storage_devices.py
```diff
@@ -22,11 +22,12 @@ def xml_linux(self, *args, **kwargs):
         virtio = self.data['attributes']['type'] == 'VIRTIO'
         logical_sectorsize = self.data['attributes']['logical_sectorsize']
         physical_sectorsize = self.data['attributes']['physical_sectorsize']
+        iotype = self.data['attributes']['iotype']
```
themylogin, line 25: How do we handle already existing devices that don't have this key set? We'll need a database migration.

themylogin (changes requested):

src/middlewared/middlewared/plugins/vm/devices/storage_devices.py
dberlin: Good point.
libvirt has a default it uses if the attribute  doesn't exist in the XML, so it's easy enough *there* to only use it if it exists.
Is there a way to have it be optional that wouldn't require a database migration?   I'm not familiar enough with the details of how the schemas work to know exactly how it deals with this case.
IE if i add null to the enum as a valid option, make it the default option, and set required=false, will it still require a migration?

Or would you prefer i just add a database migration?

src/middlewared/middlewared/plugins/vm/devices/storage_devices.py
dberlin: I'll fix. Is is okay to just .lower() it for the XML (that is why they were lower case here - these are the strings that libvirt recognizes), or would you prefer I assume there is no straight mapping from the XML to the enum and add a mapping as well?

themylogin: Just `.lower()` will be enough, we do it in a lot of similar places

src/middlewared/middlewared/plugins/vm/devices/storage_devices.py
themylogin: We would really prefer to have all the values to be always present in the configuration dictionaries (`default=` will automatically add it if it is not specified), so yes, a database migration will be the best solution here.

src/middlewared/middlewared/plugins/vm/devices/storage_devices.py
dberlin: SGTM

dberlin:
So this is ready for another look.   Enums were uppercased. etc.
I went and wrote a database migration, and then realized it doesn't need one.

The iotype option is not a separate column in the schema of device, but instead, it was added to the attribute dict - the device schema just has JSON attributes stored as a text column named "attributes".  As a result, there is no database change from  adding attributes.

IE it looks like this:
```

CREATE TABLE IF NOT EXISTS "vm_device" (
        id INTEGER NOT NULL,
        dtype VARCHAR(50) NOT NULL,
        attributes TEXT NOT NULL,
        vm_id INTEGER NOT NULL,
        "order" INTEGER,
        PRIMARY KEY (id),
        FOREIGN KEY(vm_id) REFERENCES vm_vm (id)
);

INSERT INTO vm_device VALUES(16,'DISK','{"path": "/dev/zvol/main-storage/vms/truenasbuild-y4cbr8", "type": "VIRTIO", "physical_sectorsize": null, "logical_sectorsize": null, "iotype": "NATIVE"}',3,1001);
```

Because there is a default value, it will DTRT regardless of whether the attribute is present in the attribute dict or not.

(As an aside, midcli can't render the attributes field because it is stored this way, i'm gonna go fix that)

yocalebo:
@dberlin thanks a lot for your contribution. Since this hasn't been updated in a while, I've opened up a jira ticket here: https://ixsystems.atlassian.net/browse/NAS-117383 to investigate adding this feature.

PR #9213, opened 2022-06-14
NAS-116717 / 22.12 / Fix `disk.query` event being sent with `disk_` key prefixes

themylogin:


themylogin:
time 1:00

anodos325:
I vaguely recall that there was a reason Caleb was using the datastore plugin.

themylogin:
time 1:00

sonicaj (approved):

yocalebo (changes requested):
disk.query is slow on large systems which is why I didn't use it. A KeyError happened on 13 but was happening because when we formatted the disks, a disk event would call this method which would race before geom formatting would finish. Also, this method runs without KeyErrors 100% of the time when called in isolation.

themylogin:
@yocalebo was that KeyError fixed somewhere?

yocalebo:
> @yocalebo was that KeyError fixed somewhere?

Yeah, I was able to reproduce and confirm it was fixed when you and I added the event blocking context manager.

PR #9219, opened 2022-06-14
NAS-116723 / 22.12 / Disk/partition sizes are always expressed in blocks of 512 bytes

themylogin:
4k disks on r50b-2.dc1 were reporting 8x larger size.


themylogin:
time 0:30

rick-mesta (approved):

PR #9226, opened 2022-06-15
NAS-116643 / 13.0 / Fix `{uuid}` device identifiers

themylogin:


themylogin:
time 12:00

rick-mesta:
@sonicaj / @anodos325 / @Qubad786: would one of you fellas take a look at this PR, please ?

william-gr (approved):
I don't think many people know this code other than Caleb and me (or the remaining of it).
I dont know the performance implications here of using PART vs DISK and iterating providers but the change looks reasonable to me.

Need to make sure Caleb reviews this when he is back.

william-gr:
@yocalebo just tagging you for review (after merge) for when you get back next week.

themylogin:
@william-gr this code is only executed if a disk drive does not report its serial number — weird USB dongles, weird hypervisors... I can’t imagine such a system with 100 or 1000 drives where performance implications might occur.

william-gr:
Makes sense

PR #9233, opened 2022-06-16
NAS-116725 / 22.02.3 / Show only block NVMe devices in R50 enclosures

themylogin:


themylogin:
time 0:30

william-gr:
Can you educate me what a nvme device that is not a block looks like?

themylogin:
@william-gr `/dev/nvme1` is a device that has `subsystem == "nvme"`. `/dev/nvme1n1` is a device that has `subsystem == "block"`. `disk.query` returns `nvme1n1` devices, not `nvme1`.

william-gr (approved):
Makes sense, and that will make it match the disk to show in enclosure screen from UI

PR #9243, opened 2022-06-17
NAS-116767 / 22.12 / Add test coverage for SMB ACL "map modify" behavior

anodos325:
We had regression in related to this and therefore
need explicit testing. In this case POSIX write bit
for special ACEs should be treated as a generic write
bit in the Windows SD rather than simply granting
WRITE_DATA.

### DIFF
--- a/tests/api2/test_427_smb_acl.py
+++ b/tests/api2/test_427_smb_acl.py
@@ -14,6 +14,7 @@ from auto_config import (
 )
 from pytest_dependency import depends
 from time import sleep
+from utils import create_dataset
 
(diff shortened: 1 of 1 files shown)


tests/api2/test_427_smb_acl.py
```diff
@@ -14,6 +14,7 @@
 )
 from pytest_dependency import depends
 from time import sleep
+from utils import create_dataset
```
ericbsd, line 17: Where is this from?

ericbsd (approved):
LGTM

PR #9249, opened 2022-06-18
NAS-116332 / 22.12 / Don't allow putting files on locked paths

Qubad786:
## Problem

`filesystem.put` allows uploading files to locked paths.

## Solution

Validate that middleware does not allow uploading files to locked paths.

### DIFF
--- a/src/middlewared/middlewared/plugins/filesystem.py
+++ b/src/middlewared/middlewared/plugins/filesystem.py
@@ -460,8 +460,11 @@ class FilesystemService(Service):
         else:
             openmode = 'wb+'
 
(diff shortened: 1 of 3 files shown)


anodos325:
Aren't we making the path in question immutable when it's locked? Perhaps just check for failing with a permission error?

anodos325:
LGTM, but let's get Eric's feedback since you're making a change to one of the API test functions.

ericbsd (approved):

ericbsd:
LGTM

anodos325 (approved):

PR #9251, opened 2022-06-21
NAS-116808 / 22.12 / Improve IPMI password validation

anodos325:
This addresses a couple of issues with IPMI password validation

1) The webui lists password complexity requirements that are
   not validated within middleware before sending payload
   through ipmitool.

2) Password length of 20 characters is enforced in middleware,
   but this value exceeds that supported by some ipmi implementations.

The first item is problematic because the error messages from ipmitool
for passwords that lack sufficient complexity are less than helpful
("Request data field length limit exceeded").

The second item is also incredibly problematic because on some
devices a password of more than 16 characters will be silently
truncated to 16 characters.

This PR shifts validation to the middleware schema by introducing
two changes:

a) Password() schema type. This currently only hardcodes 'private'
key, but the guiding principal is that we can start shifting
places where we are handling passwords to have a common schema
so that it is simpler to identify and audit where middleware is
handling passwords.

b) PasswordComplexity validator. This checks for presence of required
categories of characters in the password. It includes optional argument
to specify the count of these categories that must be present in the
password.

### DIFF
--- a/src/middlewared/middlewared/validators.py
+++ b/src/middlewared/middlewared/validators.py
@@ -3,6 +3,7 @@ import ipaddress
 import re
 from urllib.parse import urlparse
 import uuid
+from string import digits, ascii_uppercase, ascii_lowercase, punctuation
 
 from zettarepl.snapshot.name import validate_snapshot_naming_schema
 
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/validators.py
```diff
@@ -209,6 +210,58 @@ def __call__(self, value):
             raise ValueError(f'Invalid UUID: {e}')
 
 
+class PasswordComplexity:
+    def __init__(self, required_types, required_cnt=None):
+        self.required_types = required_types
```
sonicaj, line 215: If an invalid type is specified, what happens ?

anodos325: Hmm... Linux doesn't have EDOOFUS 🤔

PR #9252, opened 2022-06-21
NAS-116810 / 22.12 / Place mako module directory under /run

anodos325:
We don't need to use /tmp for this purpose. Shifting to dedicated
/run directory will help avoid having other processes / manual user
intervention affect our mako directory.


themylogin (approved):
Does this solve any real usecase issue? If yes, what and how? I saw mysterious mako errors about compiled template file not being present in /tmp but I was unable to understand how was this possible.

yocalebo:
retest this please

PR #9268, opened 2022-06-22
chelsio_adapter_config: Match card more precisely

freqlabs:
The previous code in 12 matched card=0x00001425, which corresponds to
subvendor=0x1425 subdevice=0x0000 in 13.  Replicate the old behavior
exactly rather than matching any card with the Chelsio vendor ID.


rick-mesta (approved):
Oooo.... aaah... 👍🏼

amotin (approved):
This should be better. I still thinking though whether space before "subvendor" would have sense, but now it is abstract cosmetics.

PR #9270, opened 2022-06-22
NAS-116842 / 13.0 / fix regression syncing disks in 13

yocalebo:
Commit 6caafedae03ca3d62da7eb0f5ede76ca8b11f0f9 made it so that `disk.sync_all` was completely broken because it no longer worked with anything other than disks having partition labels written to them. This means the first disk that was manually wiped, caused all disks in the db to be expired and therefore not displayed in `disk.query`.

Fix 3 problems:
1. revert the problem commit
2. had an improper boolean logic check (`if this and that` changed to `if this and not that`)
3. remove the walrus operator `:=` and check for explicit `None` so `uuid` information is returned appropriately.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/sync.py
+++ b/src/middlewared/middlewared/plugins/disk_/sync.py
@@ -128,7 +128,7 @@ class DiskService(Service, ServiceChangeMixin):
         _value = search.group('value').replace('\'', '%27')  # escape single quotes to html entity
         if _type == 'uuid':
             found = next(geom_xml.iterfind(f'.//config[rawuuid="{_value}"]/../../name'), None)
-            if found is not None and found.text.startswith('label'):
+            if found is not None and not found.text.startswith('label'):
                 return found.text
         elif _type == 'label':
             found = next(geom_xml.iterfind(f'.//provider[name="{_value}"]/../name'), None)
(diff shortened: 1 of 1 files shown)


themylogin:
The original code did not work well with disks that do not report serial numbers, did you test this case?

themylogin:
@yocalebo I spent a lot of time fixing this and it was not the only issue. Do you want me to test this PR with disks that do not report serials or is it already tested?

yocalebo:
> @yocalebo I spent a lot of time fixing this and it was not the only issue. Do you want me to test this PR with disks that do not report serials or is it already tested?

Sure you can test but be sure and test on physical hardware because the previous commit causes all disks to be removed from `disk.query`. Also the commit that caused the regression is for someone using truenas inside vmware which isn't really a supported configuration. As a matter of fact, most of the time where disks don't report serial numbers is when truenas is virtualized. So keep that in mind because I don't want to waste too much time on this "problem" since it effects only a small subset of users.

william-gr:
Its not such a small subset, its pretty big actually, so I would prefer we don't regress on that either :P

william-gr:
To put into perspective, we have 37.000+ VMWare Virtual disks reported in our usage stats.

PR #9272, opened 2022-06-22
NAS-116859 / 22.12 / fix snapshot and repl attachment delegates

themylogin:
This is necessary for the new UI design.

Fix `PeriodicSnapshotTaskFSAttachmentDelegate` and `ReplicationFSAttachmentDelegate`

Add integration tests


themylogin:
time 07:00

src/middlewared/middlewared/plugins/pool_/snapshot_count.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class PoolDatasetService(Service):
+
+    class Config:
+        namespace = "pool.dataset"
+
+    @accepts(Str("dataset"))
+    @returns(Int())
+    def snapshot_count(self, dataset):
```
anodos325, line 14: This will only get snapshots if the dataset is mounted. Is that an issue?

src/middlewared/middlewared/plugins/pool_/snapshot_count.py
```diff
@@ -0,0 +1,20 @@
...
+        mountpoint = Path("/mnt") / dataset
```
anodos325, line 15: We might be able to look up through getmntent in middlewared/utils. c.f. zfs.py

anodos325: The .zfs/snapshot directory has a hardcoded inode number. We can double-check that as well.

src/middlewared/middlewared/plugins/pool_/snapshot_count.py
```diff
@@ -0,0 +1,23 @@
...
+        zfs_dir = mountpoint / ".zfs/snapshot"
+        if zfs_dir.is_dir():
+            stat = zfs_dir.stat()
+            if stat.st_ino == ZFSCTL.INO_SNAPDIR:
+                return stat.st_nlink - 2
```
themylogin, line 21: @anodos325 IIRC this is how it should work?

themylogin:
time 1:00

undsoft:
@themylogin 
Thanks for the new endpoint. What's the future plan for this? Are we intending on adding more fields in the future or is it going to be data-protection specific?
I would like to avoid having a slow large endpoint. We are trying to build UI in way that would show as much information right away and load less relevant information incrementally.

themylogin:
@undsoft yes, that'll be data protection specific. Having per-card endpoints seems to be a reasonable compromise between a number of parallel calls made and progressive UI load.

anodos325:
Generally, it would be preferable to look up dataset names via `filesystem.mount_info` and then get path from `mountpoint` there. In cases where dataset may not be mounted and we want a synthetic path, then we should as fallback check with zfs.dataset.query before prepending `/mnt`.

anodos325:
Would we be able to skip the attachment checks if dataset is unmounted (i.e. no filesystem.mount_info results)?

src/middlewared/middlewared/plugins/pool_/snapshot_count.py
```diff
@@ -0,0 +1,23 @@
...
+
+class PoolDatasetService(Service):
+
+    class Config:
+        namespace = "pool.dataset"
+
+    @accepts(Str("dataset"))
+    @returns(Int())
+    def snapshot_count(self, dataset):
+        if mountpoint := self.middleware.call_sync("pool.dataset.mountpoint", dataset, False):
```
william-gr, line 16: docstring

src/middlewared/middlewared/plugins/pool_/dataset_summary.py
```diff
@@ -0,0 +1,31 @@
...
+    @returns(Dict(
+        Int("cloud_sync_task_count"),
+        Int("replication_task_count"),
+        Int("rsync_task_count"),
+        Int("snapshot_count"),
+        Int("snapshot_task_count"),
+    ))
+    async def summary(self, dataset):
+        path = f"/mnt/{dataset}"
+
```
anodos325, line 20: You need to convert dataset to mountpoint here. If not mounted, then check for dataset existence before creating synthetic path. Otherwise, you'll give bogus results for non-existent datasets.

anodos325:
> > Would we be able to skip the attachment checks if dataset is unmounted (i.e. no filesystem.mount_info results)?
> 
> Why? Tasks still exist in this case. Snapshot and replication tasks even perform just fine.

Sorry, just saw paths and thought immediately "we shouldn't do that if it's not mounted" :) It is a little odd to think of an unmounted dataset as being attached to a path.

src/middlewared/middlewared/plugins/pool_/snapshot_count.py
```diff
@@ -0,0 +1,23 @@
...
+            zfs_dir = Path(mountpoint) / ".zfs/snapshot"
+            if zfs_dir.is_dir():
+                stat = zfs_dir.stat()
+                if stat.st_ino == ZFSCTL.INO_SNAPDIR:
+                    return stat.st_nlink - 2
+
+        return self.middleware.call_sync("zfs.snapshot.query", [["dataset", "=", dataset]], {"count": True})
```
sonicaj, line 23: We should specify `name` here as well to optimize this call

sonicaj: Actually `count` would do that already, nvm :P

src/middlewared/middlewared/plugins/pool_/snapshot_count.py
anodos325: Yeah. That should be fine.

themylogin:
time 0:01

themylogin:
time 0:01

themylogin:
time 0:01

themylogin:
time 0:01

sonicaj (commented):
LGTM, minor nit

themylogin:
@anodos325 ping, are there any pending changes left?

yocalebo (changes requested):
@themylogin it seems you and I are implementing similar functionality but doing it in a different way. I'm trying to add "summarized" information for each dataset that have a share attached to it and I'm doing it a completely different way. You can see here: https://github.com/truenas/middleware/pull/9377.

I'm adding `mount_info` key to all `sharing.*.query` calls because we're already determining if the dataset is locked and it's just a matter of getting the same mount information that you're getting from `filesystem.mount_info`.

Either way, maybe you and I can go to DM and see if we can figure how best (and most efficient way) to get this done.

yocalebo (approved):

anodos325 (approved):

PR #9283, opened 2022-06-23
NAS-116869 / 22.12 / simplify and speed up interface.query (round 1)

yocalebo:
This uses `IPRoute` which is very fast and useful for 1-off "scripts" to attain information at the time of the instantiation of the `IPRoute` class.

This also uses `NDB` which synchronizes state with the kernel when a change is made. For example, changing mtu. This also simplifies the logic because we were reading sysfs values to get information when all of it is provided in the `IPRoute` class.

Before my changes, `interface.query` was taking ~0.232 seconds on my VM with 6 NICs and after it takes ~0.157 seconds which is ~33% faster.

There should be no functional change.

### DIFF
--- a/src/middlewared/middlewared/plugins/interface/netif_linux/interface.py
+++ b/src/middlewared/middlewared/plugins/interface/netif_linux/interface.py
@@ -1,10 +1,10 @@
-from pyroute2 import NDB, IPRoute
+from pyroute2 import NDB
 
-from .address import AddressFamily, AddressMixin
+from .address import AddressMixin
 from .bridge import BridgeMixin
-from .bits import InterfaceFlags, InterfaceV6Flags, InterfaceLinkState
+from .bits import InterfaceFlags, InterfaceV6Flags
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/interface/netif_linux/interface.py
```diff
@@ -41,55 +47,45 @@ def description(self, value):
...
-        if up:
-            self.down()
-            self.up()
+    def mtu(self, value):
+        with NDB(log='off') as ndb:
+            with ndb.interfaces[self.orig_name] as dev:
+                dev['mtu'] = value
+                if dev['state'] == 'up':
+                    dev['state'] = 'down'
+                    dev['state'] = 'up'
```
rick-mesta: is line 59 superfluous ? I mean, you got here **because** it was already `up`, no ?

rick-mesta (approved):

PR #9294, opened 2022-06-23
NAS-116883 / 13.0 / fix drive identify on x and m series

yocalebo:
This was raising a `KeyError` because `enc.status()['elements']['Array Device Slot']` doesn't have a key of `Array Device Slot` since the data returned from that function is the raw, unformatted data. Just need to use the formatted data from `enclosure.query` which we're storing in `info` variable.


yocalebo:
ping @truenas/middleware I've confirmed with QE that this fixes x and m series drive identification.

william-gr (approved):

PR #9296, opened 2022-06-24
NAS-116895 / 22.12 / Do not show ix-applications children on any pool

sonicaj:
This commit adds changes to not add ix-applications children on any pool it might be living on regardless of the pool being configured for apps or not.


anodos325 (approved):

Ornias1993 (approved):
nice cleanup! :)

PR #9298, opened 2022-06-24
NAS-116870 / 22.12 / Normalise volblocksize attribute for volumes

Qubad786:
## Problem

It was requested that we had a discrepancy between `recordsize` and `volblocksize` as the former allows specifying `512B` while the latter did not.

## Solution

Allow specifying `512` and `512B` for both `recordsize` and `volblocksize` attrs.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -3096,7 +3096,7 @@ class PoolDatasetService(CRUDService):
         Str('type', enum=['FILESYSTEM', 'VOLUME'], default='FILESYSTEM'),
         Int('volsize'),  # IN BYTES
         Str('volblocksize', enum=[
-            '512', '1K', '2K', '4K', '8K', '16K', '32K', '64K', '128K',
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3583,10 +3583,11 @@ async def __common_validation(self, verrors, schema, data, mode, parent=None, cu
 
                 if 'volblocksize' in data:
 
-                    if data['volblocksize'].isdigit():
-                        block_size = int(data['volblocksize'])
+                    blocksize = f'{data["volblocksize"]}B' if data['volblocksize'].isdigit() else data['volblocksize']
```
yocalebo, line 3586: Does it make more sense to do something like this? It removes a few if/else branches and an `int` cast. What do you think?
```suggestion
                    if data['volblocksize'][:3] == '512':
                        block_size = 512
                    else:
                        block_size = int(blocksize[:-1]) * 1024
```

yocalebo:
retest this please

PR #9311, opened 2022-06-27
NAS-116907 / 22.12 / Fix off-by-one and statically map Well-Known SIDs

anodos325:
A recent fix for case where we were erroneously purging winbindd
idmap tdb entries introduced an off-by-one error that was caught
by our CI. This PR fixes this issue and also introduces a static
mapping for S-1-3-4 (Owner Rights). The static mapping is to
ensure that in at least normal case of non-clustered server where
users may wish to utilize this special ACE to manage permissions
granted by virtue of being a file's owner we will make sure that
this ACE is consistently evaluated on replication targets.


anodos325:
Custom CI job passed. This should be clear.

PR #9318, opened 2022-06-29
Improve logging for when we disable a directory service

anodos325:
During the "started" methods for directory services, which are
used in health checks, we re-run validation on the user's
directory service configuration. If it fails validation, then
we automatically disable the directory service.

This PR more explicitly handles ValidationErrors during
directoryservices.get_state. An errno of EINVAL in CallError
means that we have failed validation and have disabled the service,
and in turn need to reflect this state in the directoryservices
cache.


william-gr:
time 0:01

william-gr:
time 0:01

william-gr:
time 0:01

yocalebo (approved):

PR #9319, opened 2022-06-29
prevent boot delays on large disk systems

yocalebo:
2 problems discovered.

1. `lvm2-monitor.service` is miserably slow and scans every disk so disable it entirely because we don't use `lvm` anywhere in our product
2. `initramfs` will try to resume from `/dev/dm*` devices by default which can arbitrarily slow down boot process, but we don't support resume from suspend so disable that functionality as well.

### DIFF
--- a/debian/debian/postinst
+++ b/debian/debian/postinst
@@ -71,6 +71,10 @@ systemctl mask libvirtd.socket libvirtd-ro.socket libvirtd-admin.socket libvirtd
 systemctl mask exim4-base.service exim4.service
 systemctl mask uuidd.service uuidd.socket
 
+# We don't use LVM and this service can add significan boot delays
+# on large disk systems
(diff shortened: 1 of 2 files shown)


debian/debian/postinst
```diff
@@ -71,6 +71,10 @@ systemctl mask libvirtd.socket libvirtd-ro.socket libvirtd-admin.socket libvirtd
 systemctl mask exim4-base.service exim4.service
 systemctl mask uuidd.service uuidd.socket
 
+# We don't use LVM and this service can add significan boot delays
```
freqlabs, line 74:
```suggestion
# We don't use LVM and this service can add significant boot delays
```

freqlabs (approved):

sonicaj (approved):

PR #9323, opened 2022-06-29
Switch to using ctdb client python bindings

anodos325:
Using the python bindings helps us avoid a significant amount of
subprocess calls during normal clustered operations. This PR
also adds returns decorators for ctdb.general methods.


anodos325:
Currently leak-checking python bindings before merge into samba repo.

yocalebo (approved):

PR #9337, opened 2022-07-01
Update actions

wrt54g:
I updated 2 github actions:
actions/checkout v2 to v3
actions/setup-python v2 to v4


william-gr:
Thank you, we will update ourselves when we see the need.

PR #9342, opened 2022-07-01
NAS-116953 / 22.02.3 / fix reporting.realtime memory leak

yocalebo:
Fix many problems.

1. The code that did `glob.glob` and multiple `with open` for each interface slowly leaked memory. I have no idea why but more importantly is that we can fix this by simply removing that code and use tools that already do the hard work for us.
2. the `speed` of each interface was wrong
3. we were only getting the link state of each interface every 5 minutes but we continued to get the traffic stats on the `interval`?
4. the `received_bytes_rate` and `send_bytes_rate` were calculated incorrectly

My changes move the interface related statistics to its' own module and class. I use a couple different well-known 3rd party python modules that gather these stats using `ioctl` or native C bindings which are vastly faster than what we were doing. More importantly, however, no more memory leak and makes the webUI dashboard more responsive since CPU usage goes way down using these methods.

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/ifstat.py
+++ b/src/middlewared/middlewared/plugins/reporting/ifstat.py
@@ -69,6 +69,8 @@ class IfStats(object):
                 new_data[nic]['sent_bytes'] = tx_bytes - self.prev_data.get(nic, {}).get('sent_bytes', 0)
                 new_data[nic]['received_bytes_rate'] = new_data[nic]['received_bytes'] / self.interval
                 new_data[nic]['sent_bytes_rate'] = new_data[nic]['sent_bytes'] / self.interval
+                curr_data[nic]['received_bytes_rate'] = rx_bytes / self.interval
+                curr_data[nic]['sent_bytes_rate'] = tx_bytes /self.interval
             else:
                 # nic could have been up and is now down so no reason to do calculation
                 # just fill with zeros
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/reporting/ifstat.py
```diff
@@ -0,0 +1,89 @@
...
+        self.interval = interval
+        self.prev_data = prev_data
+        self.ignore = ignore_ifaces
+        self.eth = Ethtool()
+
+    def __enter__(self):
+        return self.read()
+
+    def __exit__(self, _type, _value, _tb):
+        self.eth.close()
```
themylogin, line 18: I am not sure what `open`/`close` does, but can it have a little performance impact to keep this open during the whole `reporting.realtime` lifetime?

themylogin (approved):

src/middlewared/middlewared/plugins/reporting/ifstat.py
yocalebo: Nothing too measurable from what I could see. `Ethtool()` is opening a standard socket so it's closing it on exit.

yocalebo: However, I am going to investigate that a little more to see if I can make the class instantiation persistent in a later PR.

PR #9343, opened 2022-07-01
NAS-117014 / 22.12 / Replace subprocess calls for clustered tdb backend

anodos325:
Since we now have a python ctdb client, remove subprocess calls
when handling cluster-backed plugins.

### DIFF
--- a/src/middlewared/middlewared/plugins/tdb/base.py
+++ b/src/middlewared/middlewared/plugins/tdb/base.py
@@ -9,9 +9,9 @@ from .wrapper import TDBPath
 import os
 import json
 import errno
+import ctdb
 
 from contextlib import closing
-from subprocess import run
 from base64 import b64encode, b64decode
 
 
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/tdb/base.py
```diff
@@ -84,10 +68,10 @@ def get_connection(self, name, options):
 
         if options['cluster']:
             dbid = self._ctdb_get_dbid(name, options)
-            handle = self._get_handle(name, dbid, options)
-            self.handles[name].update({'handle': handle})
+            handle = self._get_handle(name, dbid, options, self.logger)
+            #self.handles[name].update({'handle': handle})
```
yocalebo, line 72: Still needed or can be removed?

src/middlewared/middlewared/plugins/tdb/base.py
```diff
@@ -183,6 +167,7 @@ def append_entries(tdb_key, tdb_data, state):
         handle = self.get_connection(data['name'], data['tdb-options'])
 
         with closing(handle) as tdb_handle:
+            #self.middleware.run_in_thread(self._traverse, tdb_handle, append_entries, state)
```
yocalebo, line 170: Same here as well

anodos325:
Yeah, forgot those comments. Getting ready to validate writes, but I think this is mostly correct.

yocalebo (approved):

PR #9345, opened 2022-07-03
Allow users to control validation for host paths wrt apps

sonicaj:


anodos325:
Will the value for this be exposed in debug output? May be useful for troubleshooting issues where things go wrong. Do we also want to disallow this on Enterprise (since this won't be a supported configuration IIRC)?

sonicaj:
Yes it will be exposed in the debug under `Kubernetes` section. About the enterprise bit, i have another ticket to have a separate train for enterprise customers - will be handling this there which will be soonish.

anodos325 (approved):

PR #9346, opened 2022-07-03
NAS-115275 / 22.12 / Allow users to specify their own webui portals for apps

sonicaj:
This commit adds changes to allow users to specify ui portals for apps deployed via 'launch docker image' button.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
@@ -32,7 +32,7 @@ class ChartReleaseService(Service):
             self.PORTAL_CACHE.pop(release_name, None)
 
     @private
-    def retrieve_portals_for_chart_release(self, release_data, node_ip=None):
+    def retrieve_portals_for_chart_release(self, release_data, node_ip):
         with PORTAL_LOCK:
             if release_data['name'] not in self.PORTAL_CACHE:
                 self.PORTAL_CACHE[release_data['name']] = self.retrieve_portals_for_chart_release_impl(
@@ -41,21 +41,21 @@ class ChartReleaseService(Service):
             return self.PORTAL_CACHE[release_data['name']]
 
     @private
-    def retrieve_portals_for_chart_release_impl(self, release_data, node_ip=None):
+    def retrieve_portals_for_chart_release_impl(self, release_data, node_ip):
         questions_yaml_path = os.path.join(
             release_data['path'], 'charts', release_data['chart_metadata']['version'], 'questions.yaml'
         )
         if not os.path.exists(questions_yaml_path):
             return {}
 
+        if release_data['chart_metadata']['name'] == 'ix-chart':
+            return self.get_ix_chart_portal(release_data, node_ip)
+
         with open(questions_yaml_path, 'r') as f:
             portals = yaml.safe_load(f.read()).get('portals') or {}
 
         if not portals:
             return portals
-
-        if not node_ip:
-            node_ip = self.middleware.call_sync('kubernetes.node_ip')
 
         def tag_func(key):
             return self.parse_tag(release_data, key, node_ip)
@@ -72,6 +72,16 @@ class ChartReleaseService(Service):
             cleaned_portals[portal_type] = t_portals
 
         return cleaned_portals
+
+    @private
+    def get_ix_chart_portal(self, release_data, node_ip):
+        portal_config = release_data['config'].get('portalDetails')
+        if not portal_config or not release_data['config'].get('enableUIPortal'):
+            return {}
+        host = node_ip if portal_config['useNodeIP'] else portal_config['host']
+        return {
+            portal_config['portalName']: [f'{portal_config["protocol"]}://{host}:{portal_config["port"]}']
+        }
 
     @private
     def parse_tag(self, release_data, tag, node_ip):


src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
```diff
@@ -41,22 +41,22 @@ def retrieve_portals_for_chart_release(self, release_data, node_ip=None):
...
-    def retrieve_portals_for_chart_release_impl(self, release_data, node_ip=None):
+    def retrieve_portals_for_chart_release_impl(self, release_data, node_ip):
         questions_yaml_path = os.path.join(
             release_data['path'], 'charts', release_data['chart_metadata']['version'], 'questions.yaml'
         )
         if not os.path.exists(questions_yaml_path):
             return {}
 
+        if release_data['chart_metadata']['name'] == 'ix-chart':
+            return self.get_ix_chart_portal(release_data, node_ip)
```
Ornias1993, line 52: There is no reason to limit this to ix-chart here

src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
```diff
@@ -73,6 +73,16 @@ def tag_func(key):
...
 
+    @private
+    def get_ix_chart_portal(self, release_data, node_ip):
+        portal_config = release_data['config'].get('portalDetails')
+        if not portal_config or not release_data['config'].get('enableUIPortal'):
+            return {}
+        host = node_ip if portal_config['useNodeIP'] else portal_config['host']
+        return {
+            portal_config['portalName']: [f'{portal_config["protocol"]}://{host}:{portal_config["port"]}']
+        }
```
Ornias1993, line 84: Preferably this gets writhen with a nested dicts to follow the standing practice of allowing multiple portals to be defined.

Ornias1993 (changes requested):

src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
sonicaj: I don't follow, can you clarify ?

src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
sonicaj: `ix-chart` is a simple one docker image deployment, i don't see any reason for making this confusing to an average user to allow multiple portals. Do you ?

Ornias1993: I'm not saying you should expose multiple portals, I'm saying to at least allow for it in the backend so other catalogs can actually use it. Considering support for multiple portals already exists.

Also:
This fix would allow us to significantly simplify we expose our own portals (not user defined) as well, as with this we can just deploy the portals from values.yaml directly.

src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
Ornias1993: There is no reason this should be made "iX-Official" (hence ix-chart) only.

src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
sonicaj: I don't follow, other catalogs/apps can already specify as many portals as they want ? This is just specific to official catalogs `ix-chart` app ?

src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
sonicaj: Why ? Other apps can already define as many portals as they want ?

Ornias1993: No they cannot, not dynamically.

We had to implement weird hacks using configmaps to implement basically this feature and even so we cannot dynamically enable or disable them from values.yaml.

src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
Ornias1993: No we cannot define portals from values.yaml.
At least not enable or disable them dynamically.

We have had to implement very hacky configmap code to even support dynamic portals at all.
---
Also:
Having "iX Only" features added is a bad practice to begin with, I don't understand what the issue is... When A downstream tells you "please just make this function accessable for everyone not just yourself as we want to use it as well", what is the issue against doing so?

sonicaj: @Ornias1993 if a chart dev is writing an application, can you please tell me how the portal is going to be dynamic ? i.e i write an application for nginx for example, i know it will always run this `X` path because i wrote the app, how is it really dynamic ?
About the user feedback, users can specify protocol/port already, so how does this help ?

src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
sonicaj: Are you talking about users configuring port or protocol or is it something else because i still don't follow you ? And if it's port/protocol, we still go with what we already have even in the official apps because i'm not seeing any motivation to change that workflow unless we have a good use case to do so ?

src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
sonicaj:
```
Also:
Having "iX Only" features added is a bad practice to begin with, I don't understand what the issue is... When A downstream tells you "please just make this function accessable for everyone not just yourself as we want to use it as well", what is the issue against doing so?
```
Problem is i am not even sure how/why the downstream is requesting this or what the use case is even in general ? These changes are just for apps which are deployed by user themselves via `ix-chart`. It does not make sense for any other application.
So it would be nice if you can clarify what you think we should have and why ?

Ornias1993: > @Ornias1993 if a chart dev is writing an application, can you please tell me how the portal is going to be dynamic ? i.e i write an application for nginx for example, i know it will always run this X path because i wrote the app, how is it really dynamic ?
About the user feedback, users can specify protocol/port already, so how does this help ?

It's called "different points of entry", remember we support native kubernetes ingress.
You don't know which entrypoint a user is going to configure and currently we have to hack at it to apply this because questions.yaml doesn't accept templating and this would solve that issue instantly.

Byond that: We also have had to deploy apps that have per-install requirements for custom URL entrypoints with tokens. 

> Problem is i am not even sure how/why the downstream is requesting this or what the use case is even in general ?

The problem is the need to ask, because it's kinda obvious that entrypoints are not always the static.

> These changes are just for apps which are deployed by user themselves via ix-chart. It does not make sense for any other application.

As I always like to repeat: Assumptions are the mother of all fuckups. You just "assume" it doesn't make sense, based on the limited featureset of the iX Helm-Charts/Apps. Which is an inherently flawed way of dealing with these things.

But even assuming it only makes sense with "user customisable deployments", you guys already know for ages we have TrueCharts Custom App as an iX-Chart alternative.

So this whole "explain... explain... explain..." looks somewhat as a charade to sugarcoat a wish to have ix-only "specialsauces" added.


---

It's kinda annoying to have to spend time to explain quite standard kubernetes functionality/requirements/behavior/design-choices.

src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
Ornias1993: Making this iX-Only, us simply not needed in any way, shape or form. There is no technical reason to do so. This seems highly suspicious to be frank, as it simply does not make any sense to have a simple values.yaml parser locked behind an "ix-chart" if statement.

It's on you to explain why this so harshly needs to be ix-locked, as you are proposing to ix-lock this feature.

---

As I already explained: we really need this to be able to simplify our portal generator. I'm not going to write a long tedious writedown of our codebase, as it's opensource and we both have beter things to do.

Questions.yaml is not accessable using helm templating, values.yaml is.
We write our charts using a lot more helm templating functionality than every would be possible using the questions.yaml system/file.

Hence being able to set things in values.yaml allows us a LOT more flexibility, as it gives us access to those variables using the native helm templating engine directly.

src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
Ornias1993: TLDR:
Our (and most other!) Helm-Charts are a lot more complex than anything in the iX-Catalog. such as:
- Supporting multiple loadbalancers, nodeport and ingress (some of which at the same time!)
- Having Apps with very specific per-install dependend url requirements

sonicaj: Thank you for clarifying, but this requires more thought wrt implementation as it should be much more generic then what is being offered here. Can you please make a ticket at jira how you propose the implementation should look like and we can take it from there ? Thanks!

Ornias1993: From the values.yaml side of things, it would just mean to not have that freaking ix-chart-only flag.
Really, I need to explain how values.yaml doesn't discriminate between ix-chart and every other helm-chart in existence?

If you can parse those in ix-chart, you can in every other chart.

PR #9349, opened 2022-07-04
NAS-117049 / 22.12 / improve getting zfs arcstats

yocalebo:
Improve getting zfs arcstats in the `reporting.realtime` subscription.

1. Don't `f.readlines()` since that loads the entire contents of the file into a list in memory. We can iterate over the file object which is quite a bit faster and more memory efficient.
2. Keep track of the 4 stats that we're pulling out of the file and then break the for-loop early as to not iterate over the entire contents of the file unnecessarily.
3. move this logic to it's own module like we are doing with `DiskStats` and `IfStats`

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/arcstats.py
+++ b/src/middlewared/middlewared/plugins/reporting/arcstats.py
@@ -1,0 +1,44 @@
+class ZfsArcStats(object):
+
+    def __init__(self, *args, **kwargs):
+        self.args = args
+        self.kwargs = kwargs
+
+    def read():
+        hits = misses = total = 0
+        fhits = fmisses = fmax = fsize = False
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/reporting/arcstats.py
```diff
@@ -0,0 +1,44 @@
+class ZfsArcStats(object):
```
sonicaj, line 1: Do we see any use case for this being a class ?

yocalebo: Not really but it follows the same design as the `DiskStats` and `IfStats` classes to enforce some form of uniformity in this file 😄

sonicaj (approved):
LGTM, however it would be nice if we had a unit test for the parsing

PR #9354, opened 2022-07-05
NAS-117061 / 22.02.3 / fix small memory leak in netif.list_interfaces()

yocalebo:
Fix small (200 - 900 bytes each time this is called) memory leak in `netif.list_interfaces()`. I'm already doing this on master but there were other improvements that were included there so this is the compromise for stable/angelfish branch.


anodos325 (approved):

yocalebo:
backport

PR #9360, opened 2022-07-06
NAS-117072 / 22.12 / Include nftables ruleset in network debug

ghost:
iptables is deprecated in Debian in favor of nftables.  The iptables
commands are configured to use iptables-nft, which wraps nftables in an
iptables command interface for compat.  We should prefer to use nftables
where we have the option.

With this in mind, it is useful to also know the full nftables ruleset
when debugging network issues, in particular with regard to kube-router.


freqlabs:
time 0:10

PR #9371, opened 2022-07-08
NAS-116829 / 22.12 / Use standard `Message-ID` generation method.

themylogin:
Also do not use any host but `localhost`, because iCloud Mail does not like it.


themylogin:
time 4:00

yocalebo (approved):

PR #9376, opened 2022-07-11
Add comment field to ACL templates

anodos325:
Per request of TrueCommand team add comment describing ACL
template. This is also probably useful for end-users who
add their own templates. In case of AD-joined servers we
lookup and add Domain Users and Domain Admins where
appropriate. Since the exact name of these groups can
vary based on AD localization, we have to look them up
via Windows SID.


anodos325:
AD groups need to be looked up because TC will present users with templates and not allow fine-tuning (so we need to be more intelligent on backend).

yocalebo (approved):

PR #9377, opened 2022-07-11
NAS-117137 / 22.12 / add mount_info to sharing plugins

yocalebo:
This addes `mount_info` key to all sharing plugins. This will be used for the new storage page redesign.

### DIFF
--- a/src/middlewared/middlewared/plugins/nfs.py
+++ b/src/middlewared/middlewared/plugins/nfs.py
@@ -536,6 +536,7 @@ class SharingNFSService(SharingService):
         data["hosts"] = " ".join(data["hosts"])
         data["security"] = [s.lower() for s in data["security"]]
         data.pop(self.locked_field, None)
+        data.pop(self.mount_info_field, None)
         return data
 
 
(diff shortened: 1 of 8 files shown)


yocalebo:
retest this please

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -524,12 +525,25 @@ def validate_hosts_and_networks(self, other_shares, data, schema_name, verrors,
                 )
 
     @private
-    async def extend(self, data):
+    async def extend(self, data, ctx):
         data["networks"] = data.pop("network").split()
         data["hosts"] = data["hosts"].split()
         data["security"] = [s.upper() for s in data["security"]]
+        data["mount_info"] = {}
+        for row in filter(lambda x: x["id"] == data["id"], ctx["rows"]):
```
sonicaj, line 533: I am not sure I follow the use of context here ?
This loop will only run once, and you don't even use `row` here ?

yocalebo: I was trying to find if a share path is locked before trying to get the mount info but I learned that the extend_context method doesn't have access to this information because of how the class inheritance is setup for `LockableFSAttachmentDelegate`. I've added WIP label to this and will come back to it later.

PR #9380, opened 2022-07-11
NAS-114156 / 22.12 / add endpoint specifying if default route will be forcefully removed

yocalebo:
Trying to improve configuring the network on a TrueNAS and this endpoint will be used by the webUI to prompt user about the default gateway being removed. Currently unused.


yocalebo:
retest this please

yocalebo:
I'm going to move this to PR: https://github.com/truenas/middleware/pull/9389/files

PR #9383, opened 2022-07-12
Allow querying supports_smart for disk

themylogin:


themylogin:
time 2:00

yocalebo (changes requested):
We need something better than this. WebUI team is going to call this because they're asking for it for the storage page redesign.

themylogin:
@yocalebo they’ll only call this for a single instance, I don’t see a big issue here. You don’t want to cache this value in disk.sync_all, right?

yocalebo:
@themylogin `subprocess` for multi-hundred+ disk system will negate all the other performance optimizations that we've put into the smartctl related plugins. Also, it'll take forever for the page to load initially so the end-user will close the tab and try to load it again which becomes a vicious cycle. Or am I misunderstanding how this data is going to be loaded on the front-end?

Either way, on enterprise systems we need to always return `True` since we already make that assumption in another part of the smart plugin.

themylogin:
> Or am I misunderstanding how this data is going to be loaded on the front-end?

They'll only specify this option for loading a single disk page, S.M.A.R.T. availability will be queried for a single disk only.

william-gr:
Doesn't matter what the UI will do, right? The API will allow you to return it for every disk unless I am misreading the code?

themylogin:
@william-gr I can block this option if the query result count is > 1, will that be good?

william-gr:
I would imagine so

src/middlewared/middlewared/plugins/disk.py
```diff
@@ -155,6 +167,10 @@ async def disk_extend_context(self, rows, extra):
         if context['passwords']:
             context['disks_keys'] = await self.middleware.call('kmip.retrieve_sed_disks_keys')
 
+        if context['supports_smart']:
+            if len(rows) > 1:
+                raise CallError('`supports_smart` cannot be queried if disk count is greater than 1')
```
anodos325, line 172: Does it make more sense to maybe raise an exception in query method if `supports_smart` is specified without the `get` option?

Don't forget to update docstring :)

anodos325: Idea being that caller is explicitly requesting that they only want one result.

themylogin: @anodos325 but it's perfectly fine to query by `disk_identifier` and expect an empty list as one of the possible return values.

src/middlewared/middlewared/plugins/disk.py
```diff
@@ -134,6 +138,13 @@ async def disk_extend(self, disk, context):
         else:
             disk.pop('passwd')
             disk.pop('kmip_uid')
+
+        disk['supports_smart'] = None
+        if context['supports_smart']:
+            disk['supports_smart'] = disk['name'].startswith('nvme') or bool(RE_SMART_AVAILABLE.search(
```
yocalebo, line 144: Please make this `True` always on enterprise hardware like we do in the `SMARTCTX` namedtuple.

src/middlewared/middlewared/plugins/disk.py
anodos325: Okay. docstring still needs to be updated to reflect new behavior (raises exception if more than one result).

anodos325 (approved):

yocalebo (approved):

PR #9389, opened 2022-07-12
NAS-117159 / 22.12 / save/rollback default gateway on interface changes

yocalebo:
These 2 new methods will be used by front-end team so that a new modal can be presented to end-user allowing them to save a default gateway. This prevents the predominant scenario of ripping out the default gateway without any warning when a user unchecks the dhcp box and configures a static IP and the default route was installed via DHCP.

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -365,11 +365,17 @@ class InterfaceService(CRUDService):
         """
         # FIXME: What about IPv6??
         ifaces = self.middleware.call_sync('datastore.query', 'network.interfaces')
-        rt = netif.RoutingTable()
-        will_be_removed1 = not ifaces and rt.default_route_ipv4
+        rtgw = netif.RoutingTable().default_route_ipv4
+        if rtgw is not None:
+            rtgw = rtgw.gateway.exploded
 
+        # no interfaces are configured in db but a default route exist in kernel
+        will_be_removed1 = not ifaces and rtgw
+
+        # we have a default route in kernel and we have a route specified in the db
+        # and they do not match
         dbgw = self.middleware.call_sync('network.configuration.config')['ipv4gateway']
-        will_be_removed2 = rt.default_route_ipv4 and (dbgw != rt.default_route_ipv4)
+        will_be_removed2 = rtgw and (dbgw != rtgw)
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -352,6 +353,102 @@ def iface_extend(self, iface_state, configs, ha_hardware, fake=False):
...
+        is seen most often by end-users/support team.
+        """
+        if not self._original_datastores:
+            raise CallError('There are no pending interface changes.')
+
+        gw = ip_address(gw)
+        defgw = {'gc_ipv4gateway': gw.exploded}
+        for iface in await self.middleware.call('datastore.query', 'network.interfaces'):
+            if gw in ip_interface(f'{iface["int_address"]}/{iface["int_netmask"]}').network:
+                await self.middleware.call('datastore.update', 'network.globalconfiguration', 1, defgw)
```
themylogin, line 442: Are we sure this row always exists and has `id = 1`? I have never made a good assumption about our database that wasn't soon broken by a ticket from a user that uses FreeNAS since 9.1.

themylogin (changes requested):

src/middlewared/middlewared/plugins/network.py
```diff
@@ -352,6 +353,102 @@ def iface_extend(self, iface_state, configs, ha_hardware, fake=False):
...
+        configured gateway doesn't match the one currently installed in kernel.
+        """
+        # FIXME: What about IPv6??
+        ifaces = self.middleware.call_sync('datastore.query', 'network.interfaces')
+        rtgw = netif.RoutingTable().default_route_ipv4
+        if rtgw is not None:
+            rtgw = rtgw.gateway.exploded
+
+        # no interfaces are configured in db but a default route exist in kernel
+        will_be_removed1 = not ifaces and rtgw
```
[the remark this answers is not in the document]
yocalebo: thanks!

src/middlewared/middlewared/plugins/network.py
yocalebo: I'm willing to make that assumption since no row in that table will cause other things to break before we get here and if `id != 1` then I'd like to know how they got into that state 😄

themylogin (approved):

PR #9405, opened 2022-07-14
NAS-117077 / 13.0 / cam and ses drivers are racy and fix disk.sync* race

yocalebo:
2 problems found:

1st problem exposed a race condition between `disk.sync` and `disk.sync_all`

1. physically add 3 drives to head-unit
2. `disk.sync_all` kicks off for the first 2 drives
3. the 3rd disk triggers `disk.sync`
4. `disk.sync` finishes before `disk.sync_all` adding the 3rd disk to the db
5. `disk.sync_all` finishes and removes the 3rd disk from the db because it wasn't present when it queried the kernel for disk information

To fix this, make `disk.sync` a job (like `disk.sync_all`) and make it share the same job lock. Successfully reproduced and verified these changes fix the issue.

2nd problem exposed the fact that CAM and SES drivers on freeBSD are racy in certain edge-cases. When 3 ssds were moved from head-unit to jbod, the 3rd disk would not be updated with enclosure slot information. After investigating I found that there was around 60 seconds between when CAM driver probed the disk and when SES driver actually showed the disk. To work-around this, there is a `GEOM::physpath` event that gets sent to devd that we should listen to. This ensures we update the db cache when a disk is moved.

### DIFF
--- a/src/middlewared/middlewared/plugins/device_/devd_events.py
+++ b/src/middlewared/middlewared/plugins/device_/devd_events.py
@@ -82,9 +82,15 @@ async def devd_listen(middleware):
         if parsed['system'] in ('CAM', 'ACPI'):
             continue
 
-        await middleware.call_hook(
-            f'devd.{parsed["system"]}'.lower(), data=parsed,
-        )
+        if parsed['type'] == 'GEOM::physpath' and parsed.get('devname'):
+            # treat GEOM::physpath as DEVFS (even though it's geom)
+            # to fix a rare race condition between CAM and SES drivers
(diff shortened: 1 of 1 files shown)


rick-mesta (approved):

themylogin:
LGTM, but the easiest thing to do is to rename `sync` to `sync_impl` and make it a job, then add a normal `sync` method that will start this job and wait for it to end, that way `sync` will remain a non-job and API will remain intact.

yocalebo:
> on-job and API will remain intact.

That's a good idea. I'll try that and test

PR #9408, opened 2022-07-14
NAS-116508 / 22.12 / Add tests for running as some user

sonicaj:
This commit adds some tests to see if our implementation of running as some user works as intended. The `cronjob` plugin uses the internal implementation we have for running as some X user, so we add some cronjobs to be run as specific users to see if we get desired results.

### DIFF
--- a/tests/api2/test_run_as_user_impl.py
+++ b/tests/api2/test_run_as_user_impl.py
@@ -13,8 +13,10 @@ def create_cron_job(owner, ownerGroup, user):
     test_folder = '/tmp/test'
     ssh(f'mkdir {test_folder}')
     ssh(f'chown -R {owner}:{ownerGroup} {test_folder}')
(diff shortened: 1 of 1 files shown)


tests/api2/test_run_as_user_impl.py
```diff
@@ -0,0 +1,42 @@
...
+def test_01_running_as_valid_user():
+    with create_cron_job(owner='apps', ownerGroup='apps', user='apps') as cron_job:
+        with run_cron_job(cron_job['id']) as job_detail:
+            assert job_detail['results']['error'] is None
+
+
+def test_02_running_as_invalid_user():
+    with create_cron_job(owner='root', ownerGroup='root', user='apps') as cron_job:
+        with run_cron_job(cron_job['id']) as job_detail:
+            assert job_detail['results']['error'] is not None, job_detail
```
themylogin, line 42: Great tests, but it'll be good to make some assertions about the error text here too

themylogin (approved):

PR #9409, opened 2022-07-15
NAS-117045 / 22.12 / Resolve minimum memory key error

Qubad786:
## Problem

`min_memory` is an optional attribute which might or might not be specified and if not specified we fail with a `KeyError`.

## Solution

Safely check existence of `min_memory` attribute in the payload before performing checks on it.

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/vms.py
+++ b/src/middlewared/middlewared/plugins/vm/vms.py
@@ -113,7 +113,7 @@ class VMService(CRUDService, VMSupervisorMixin):
         Str('nodeset', default=None, null=True, validators=[NumericSet()]),
         Bool('pin_vcpus', default=False),
         Int('memory', required=True, validators=[Range(min=20)]),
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/vm/vms.py
```diff
@@ -175,7 +175,7 @@ async def __common_validation(self, verrors, schema_name, data, old=None):
         if not data.get('uuid'):
             data['uuid'] = str(uuid.uuid4())
 
-        if data['min_memory'] and data['min_memory'] > data['memory']:
+        if data.get('min_memory') is not None and data['min_memory'] > data['memory']:
```
themylogin, line 178: Why don't we just make it `default=None`? You can even add `strict=True` to the `Dict` declaration and middleware will crash on startup if any of the dict members is not required and does not have any default value. It's better to have all default values explicitly specified in one place.

themylogin (approved):

PR #9418, opened 2022-07-18
NAS-117231 / 22.12 / Add CTDB event integration

anodos325:
CTDB can be configured to run event scripts at various times /
situations. This commit adds some initial plumbing to pass
these through to middlewared alerts and also subscribers to
ctdb events. The primary ones we are concerned with are:

INIT - add checks that middlewared is healthy and that the
ctdb shared volume is healthy before starting the ctdb process.
If this fails then ctdb will not start.

STARTUP - there are some services that we want to enable / disable
clusterwide. Add support for a special configuration file on
the ctdb_shared_volume that contains services that should always
be running when cluster is healthy. Services here will be restarted
when CTDB finishes starting.

MONITOR - CTDB monitors the services (same as STARTUP) at
configurable intervals and starts the service if it is stopped.
This also checks the status of the CTDB shared volume. Failure
of CTDB shared volume health check will cause CTDB on the node to
become UNHEALTHY. Failure of monitored service in this script
will not degrade CTDB status, but will generate alert / event.
Alert / event will only be triggered on state change based on
sentinel file.

STARTRECOVERY / RECOVERED - This sends events when recovery
operations are started and completed. Should be very infrequent
per documentation. Recovery master only.

IPREALLOCATED - triggered if public IP address changes between
nodes. This will generate an event from recovery master containing
new public IP info.

The actual event script called by CTDB will be located in
Samba repo, but use the CtdbEvent class defined here.

Two new services are added in this PR:
`ctdb.event` - processes events sent from the CTDB daemon, creating
or removing alerts as needed, and sending alerts to TrueCommand.

`ctdb.service` - used for getting basic info about ctdb-managed
services. When ctdb manages services, each node will write basic
status information during the monitoring interval to the ctdb shared
volume.

This is currently not utilized in this PR, but in future can be used to
give cluster-wide state for services through the normal services API.

ctdb.service.set is used to enable ctdb monitoring / management
of supported services. Currently this is limited to the SMB service.

### DIFF
--- a/src/middlewared/middlewared/plugins/cluster_linux/ctdb_event.py
+++ b/src/middlewared/middlewared/plugins/cluster_linux/ctdb_event.py
@@ -1,0 +1,118 @@
+import enum
+
+from middlewared.service import job, Service
+
+
+EVENT_FN_MAP = {
+    'INIT': 'event_init',
+    'SETUP': 'event_not_implemented',
+    'STARTUP': 'event_not_implemented',
+    'SHUTDOWN': 'event_not_implemented',
+    'MONITOR': 'event_monitor',
+    'STARTRECOVERY': 'event_recovery',
+    'RECOVERED': 'event_recovery',
+    'TAKEIP': 'event_not_implemented',
+    'RELEASEIP': 'event_not_implemented',
+    'IPREALLOCATED': 'event_ip_reallocated',
+}
+
+
+class CtdbEventType(enum.Enum):
+    INIT = enum.auto()
+    SETUP = enum.auto()
+    STARTUP = enum.auto()
+    SHUTDOWN = enum.auto()
+    MONITOR = enum.auto()
+    STARTRECOVERY = enum.auto()
+    RECOVERED = enum.auto()
+    TAKEIP = enum.auto()
+    RELEASEIP = enum.auto()
+    IPREALLOCATED = enum.auto()
+
+    def get_fn(self):
+        return EVENT_FN_MAP[self.name]
+
+
+class CtdbEventService(Service):
+
+    class Config:
+        namespace = 'ctdb.event'
+        private = True
+
+    def event_ip_reallocated(self, data):
+        """
+        This is notification only. Not an error. Indicates
+        that our public IPs have shifted between nodes.
+
+        We only send out event info from recovery master.
+        """
+        if not self.middleware.call_sync('ctdb.general.is_rec_master'):
+            return
+
+        public_ips = self.middleware.call_sync('ctdb.general.ips')
+        self.middleware.send_event(
+            'ctdb.status', 'CHANGED', fields={'event': data['event'], 'data': public_ips}
+        )
+
+    def event_init(self, data):
+        """
+        This event gets triggered when CTDB is starting.
+        Expected failure mode is when ctdb_shared_volume isn't properly mounted.
+        """
+        ev = data.pop('event')
+        if data['status'] == 'SUCCESS':
+            self.middleware.call_sync('alert.oneshot_delete', 'CtdbInitFail', None)
+        else:
+            self.middleware.call_sync(
+                'alert.oneshot_create',
+                'CtdbInitFail',
+                {'errmsg': data['reason']}
+            )
+
+        self.middleware.send_event(
+            'ctdb.status', 'CHANGED', fields={'event': ev, 'data': data}
+        )
+
+    def event_monitor(self, data):
+        """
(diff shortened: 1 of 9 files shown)


anodos325:
NOTE: event scripts must be explicitly enabled in order for CTDB to run them.

anodos325:
Samba repo wrapper script: https://github.com/truenas/samba/pull/135/files

src/middlewared/middlewared/plugins/cluster_linux/ctdb_event.py
```diff
@@ -0,0 +1,118 @@
...
+
+        ev = data.pop('event')
+        self.middleware.send_event(
+            'ctdb.status', 'CHANGED', fields={'event': ev, 'data': data}
+        )
+
+    def event_not_implemented(self, arg_unused):
+        raise NotImplementedError()
+
+    @job(lock="event_results", transient=True)
```
yocalebo, line 111: Have we considered how we're going to handle the situation when we receive a bunch of events back-to-back? This will queue up jobs indefinitely so wondering if we need to put a `lock_queue_size` on this?

src/middlewared/middlewared/scripts/ctdb_events.py
```diff
@@ -0,0 +1,276 @@
...
+
+        The other is a node-specific service status file
+        /cluster/ctdb_shared_vol/.clustered_services.{pnn}
+
+        This file is updated with current run state of service
+        on every monitoring cycle. This allows any node
+        on the server to see a cached copy of run state
+        of all nodes (based on last monitoring interval).
+        """
+        self.pnn = ctdb.Client().pnn
```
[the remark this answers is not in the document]
yocalebo: Heh, nvm `Client()` is our middlewared client, I thought it was ctdb client, ignore me.

src/middlewared/middlewared/scripts/ctdb_events.py
```diff
@@ -0,0 +1,276 @@
...
+                else:
+                    if srvinfo['enable']:
+                        self.client.call('service.update', srvinfo['id'], {'enable': False})
+
+                    if srvinfo['state'] == 'RUNNING':
+                        self.client.call('service.stop', srv['name'])
+
+                self.node_status[srv['name']] = final_state
+
+        if last_failed and payload['status'] == 'SUCCESS':
```
yocalebo, line 198: Do we care if the sentinel exists if the `payload['status']` is success? Would it not be better to ju;st check for `SUCCESS` and remove the sentinel always suppressing filenotfounderror?

src/middlewared/middlewared/scripts/ctdb_events.py
```diff
@@ -0,0 +1,276 @@
...
+            os.remove(CTDB_MONITOR_FAILED_SENTINEL)
+            self.client.call('ctdb.event.process', payload)
+
+        elif payload['status'] == 'FAILURE' and not last_failed:
```
yocalebo, line 202: Same here as above. If it's failure, just always create the sentinel overwriting it if it already exists?

src/middlewared/middlewared/plugins/cluster_linux/ctdb_event.py
```diff
@@ -0,0 +1,118 @@
...
+            self.middleware.call_sync('alert.oneshot_delete', 'CtdbClusteredService', None)
+
+        else:
+            self.middleware.call_sync(
+                'alert.oneshot_create',
+                'CtdbClusteredService',
+                {'errmsg': data['reason']}
+            )
+
+            self.middleware.send_event(
```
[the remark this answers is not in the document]
anodos325: Yeah, don't know what crazy pills I was taking when I wrote that.

src/middlewared/middlewared/scripts/ctdb_events.py
anodos325: There was another indentation error. Idea is to use the sentinel file to determine whether our service state has changed. Although when I was originally writing this I didn't keep track of run state of services in the gluster volume so maybe I can use that instead of sentinel.

anodos325: Service state change is deciding factor for whether we passing monitoring results to middleware.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_event.py
anodos325: Sure. I put in three (totally not arbitrarily chosen).

yocalebo (approved):

PR #9432, opened 2022-07-19
Enforce password complexity rules for user accounts

anodos325:
Require minimum of 8 character passwords with at least 2
different character types.


yocalebo:
Will revisit soon.

PR #9434, opened 2022-07-20
Fix middleware import in middleware tests utils

sonicaj:
This commit adds changes to not import anything from middleware package as only middleware.test is installed on runners.


ericbsd (approved):
LGTM

PR #9440, opened 2022-07-20
NAS-117265 / 22.12 / SCALE Bluefin incremental api2/test_iscsi_auth_network.py -- Attribut…

rick-mesta:
…eError

### DIFF
--- a/tests/api2/test_iscsi_auth_network.py
+++ b/tests/api2/test_iscsi_auth_network.py
@@ -97,7 +97,7 @@ def configure_iscsi_service():
             call('service.stop', 'iscsitarget')
 
 
-@functools.lru_cache
+@functools.lru_cache(maxsize=None)
 def get_nas_ip_subnet():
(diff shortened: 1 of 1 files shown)


tests/api2/test_iscsi_auth_network.py
```diff
@@ -97,7 +97,7 @@ def configure_iscsi_service():
             call('service.stop', 'iscsitarget')
 
 
-@functools.cache
+@functools.lru_cache
```
yocalebo, line 100:
```suggestion
@functools.lru_cache(maxsize=None)
```

This does what `functools.cache` does in python 3.9

yocalebo (changes requested):

tests/api2/test_iscsi_auth_network.py
rick-mesta: aha... fixing

PR #9445, opened 2022-07-20
NAS-116295 / 22.12 / Improve VM memory reporting and ZFS ARC management wrt VMs

sonicaj:

1. Fixes a bug on how we deduced available memory for vms. We were taking into account the complete memory of VMs which user had requested disregarding the fact that when we get `free` physical memory, existing used memory of VMs is already accounted there. So we just need to get the memory which the VM `can` consume and is not consuming right now.
2. When starting a VM, `overcommit` to the user actually means forcing the start of the VM even if we don't have enough memory available. In reality what happened is that when it was set, we manipulated the ARC to shrink arc max but the same behaviour was not done when we were calculating `available/free` memory for VMs and we always accounted for ARC. So now if `overcommit` is specified, only then do we consider shrinkable ARC as free memory - otherwise we don't.
3. ARC max was basically a tunable which we were changing on each VM start. Now when middleware started, it cached the value of ARC max and considered that to be the default which we would consider in our calculations when giving back memory on stopping VMs. However this had a fatal flaw, what if middleware was restarted ? It meant that middleware would never know the actual default/initial value of arc max. In Linux, this is basically `total physical memory / 2` and we add that change here.
4. Add an endpoint which would show VM memory related statistics if that VM was to be started. Motivation behind this endpoint is to gather information and show to the user preferably in the UI with a pop-up or something similar when user wants to start a VM that this much memory is being requested and this much is available and if you `overcommit`, we would be doing this with ZFS ARC and this would be the new ZFS ARC value etc.
5. Do not unconditionally always change ZFS ARC if `overcommit` is set. It should only be done selectively based on actual need of the VM.

### DIFF
--- a/src/middlewared/middlewared/plugins/sysctl/sysctl_info.py
+++ b/src/middlewared/middlewared/plugins/sysctl/sysctl_info.py
@@ -22,7 +22,7 @@ class SysctlService(Service):
     def get_default_arc_max(self):
         # Default value of arc max on systems
         # https://github.com/truenas/zfs/blob/f38f3a4a3ee430a15fa221ba9fc2b6b8fd17c040/module/os/linux/zfs/arc_os.c#L85
-        return int(psutil.virtual_memory().total / 2)
+        return psutil.virtual_memory().total // 2
 
     def get_arc_max(self):
         return self.get_arcstats()['c_max']
--- a/src/middlewared/middlewared/plugins/sysctl/sysctl_info.py
+++ b/src/middlewared/middlewared/plugins/sysctl/sysctl_info.py
@@ -1,4 +1,5 @@
 import os
+import psutil
 
 from middlewared.service import CallError, Service
 from middlewared.utils import run
@@ -17,6 +18,11 @@ class SysctlService(Service):
         if cp.returncode:
             raise CallError(f'Unable to retrieve value of "{sysctl_name}" sysctl : {cp.stderr.decode()}')
         return cp.stdout.decode().split('=')[-1].strip()
+
+    def get_default_arc_max(self):
+        # Default value of arc max on systems
+        # https://github.com/truenas/zfs/blob/f38f3a4a3ee430a15fa221ba9fc2b6b8fd17c040/module/os/linux/zfs/arc_os.c#L85
+        return int(psutil.virtual_memory().total / 2)
 
     def get_arc_max(self):
         return self.get_arcstats()['c_max']
--- a/src/middlewared/middlewared/plugins/vm/memory.py
+++ b/src/middlewared/middlewared/plugins/vm/memory.py
@@ -19,7 +19,7 @@ class VMService(Service, VMSupervisorMixin):
             raise CallError(f'Cannot guarantee memory for guest {vm["name"]}', errno.ENOMEM)
 
         if memory_details['current_arc_max'] != memory_details['arc_max_after_shrink']:
-            self.middleware.logger.debug(
+            self.logger.debug(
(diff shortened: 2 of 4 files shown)


src/middlewared/middlewared/plugins/vm/vm_memory_info.py
```diff
@@ -65,21 +70,89 @@ async def get_available_memory(self, overcommit):
...
+        """
+        vm = await self.middleware.call('vm.get_instance', vm_id)
+        if vm['status']['state'] == 'RUNNING':
+            # TODO: Let's add this later as we have a use case in the UI - could be useful to
+            #  show separate info of each VM in the UI moving on
+            raise CallError(f'Unable to retrieve {vm["name"]!r} VM information as it is already running.')
+
+        arc_max = await self.middleware.call('sysctl.get_arc_max')
+        arc_min = await self.middleware.call('sysctl.get_arc_min')
+        shrinkable_arc_max = 0 if arc_max <= arc_min else arc_max - arc_min
```
freqlabs, line 128:
```suggestion
        shrinkable_arc_max = max(0, arc_max - arc_min)
```

src/middlewared/middlewared/plugins/vm/vm_memory_info.py
```diff
@@ -65,21 +70,89 @@ async def get_available_memory(self, overcommit):
...
+        vm_requested_memory = vm_min_memory or vm_max_memory
+
+        overcommit_required = vm_requested_memory > available_memory
+        arc_to_shrink = 0
+        if overcommit_required:
+            more_required = vm_requested_memory - available_memory
+            if shrinkable_arc_max < more_required:
+                arc_to_shrink = shrinkable_arc_max
+            else:
+                arc_to_shrink = shrinkable_arc_max - more_required
```
freqlabs, line 143:
```suggestion
            arc_to_shrink = min(shrinkable_arc_max, more_required)
```

src/middlewared/middlewared/plugins/sysctl/sysctl_info.py
```diff
@@ -18,6 +19,11 @@ async def get_value(self, sysctl_name):
             raise CallError(f'Unable to retrieve value of "{sysctl_name}" sysctl : {cp.stderr.decode()}')
         return cp.stdout.decode().split('=')[-1].strip()
 
+    def get_default_arc_max(self):
+        # Default value of arc max on systems
+        # https://github.com/truenas/zfs/blob/f38f3a4a3ee430a15fa221ba9fc2b6b8fd17c040/module/os/linux/zfs/arc_os.c#L85
+        return int(psutil.virtual_memory().total / 2)
```
freqlabs, line 25: For paranoia about conversions between float and int 😅 
```suggestion
        return psutil.virtual_memory().total // 2
```

freqlabs (approved):

src/middlewared/middlewared/plugins/vm/memory.py
```diff
@@ -8,44 +8,41 @@
...
+        if not overcommit:
+            raise CallError(f'Cannot guarantee memory for guest {vm["name"]}', errno.ENOMEM)
 
-        if arc_max > arc_min:
-            new_arc_max = max(arc_min, arc_max - memory_bytes)
-            self.middleware.logger.debug('Setting ARC from %s to %s', arc_max, new_arc_max)
-            await self.middleware.call('sysctl.set_arc_max', new_arc_max)
-        return True
+        if memory_details['current_arc_max'] != memory_details['arc_max_after_shrink']:
+            self.middleware.logger.debug(
```
yocalebo, line 22: self.logger?

yocalebo (approved):

PR #9447, opened 2022-07-21
NAS-116411 / 22.12 / add `pool.dataset.details`

yocalebo:
This will be used by front-end team for storage page redesign. This gathers the information they're after in 1 parent for loop trying to be as optimized as possible.

I've tested this on real hardware with > 1k+ datasets, ~15K snapshots, 90 smb shares, 20 iscsi shares, 36 nfs shares Along with 1 rsync, snapshot, replication and cloudsync task respectively.

When I call this method, it takes ~2.2 seconds which is fast enough since it's been tested against absolute worst case scenario.

On a side-note, I'm querying with `flat: False` and then collapsing the children datasets into a list so a weakref is made so that the original structure of the returned data can be kept in tact because the webUI uses that to expand/collapse the dataset tree. (i.e. show only parent datasets and allow user to drill-down into the child datasets if any exist)

NOTE: I am not determining the zvol snapshot count at this time because it makes this considerably slower. Makes it so slow that this endpoint isn't usable. This will become moot once an upstream openzfs pr is merged which adds the snapshot count as a zfs property. (It has been approved, just not merged)

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/dataset_details.py
+++ b/src/middlewared/middlewared/plugins/pool_/dataset_details.py
@@ -49,7 +49,7 @@ class PoolDatasetService(Service):
             i['smb_shares'] = self.get_smb_shares(i, info['smb'])
             i['iscsi_shares'] = self.get_iscsi_shares(i, info['iscsi'])
             i['vms'] = self.get_vms(i, info['vm'])
-            i['apps'] = self.get_apps(i,  info['app'])
+            i['apps'] = self.get_apps(i, info['app'])
             i['replication_tasks_count'] = self.get_repl_tasks_count(i, info['repl'])
             i['snapshot_tasks_count'] = self.get_snapshot_tasks_count(i, info['snap'])
             i['cloudsync_tasks_count'] = self.get_cloudsync_tasks_count(i, info['cloud'])
--- a/src/middlewared/middlewared/plugins/pool_/dataset_details.py
+++ b/src/middlewared/middlewared/plugins/pool_/dataset_details.py
@@ -34,33 +34,26 @@ class PoolDatasetService(Service):
                 ]
             }
         }
-
-        mntinfo = getmntinfo()
-        nfsshares = self.middleware.call_sync('sharing.nfs.query')
-        smbshares = self.middleware.call_sync('sharing.smb.query')
-        iscsishares = self.build_iscsi_share_info()
-        repltasks = self.middleware.call_sync('datastore.query', 'storage.replication', [], {'prefix': 'repl_'})
-        snaptasks = self.middleware.call_sync('datastore.query', 'storage.task', [], {'prefix': 'task_'})
-        cldtasks = self.middleware.call_sync('datastore.query', 'tasks.cloudsync')
-        rsynctasks = self.middleware.call_sync('rsynctask.query')
         collapsed = []
-        info = self.middleware.call_sync('pool.dataset.query', filters, options)
-        for dataset in info:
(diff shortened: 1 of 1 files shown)


anodos325 (approved):

src/middlewared/middlewared/plugins/pool_/dataset_details.py
```diff
@@ -0,0 +1,221 @@
...
+
+            try:
+                st = os.stat(f'{ds["mountpoint"]}/.zfs/snapshot')
+            except FileNotFoundError:
+                # means that the zfs snapshot dir doesn't exist which
+                # can only happen if the dataset is encrypted and "locked"
+                # (unmounted) or the dataset is unmounted
+                locked = ds['encrypted']
+            else:
+                if st.st_ino == ZFSCTL.INO_SNAPDIR:
```
themylogin, line 106: `pool.dataset.snapshot_count` uses the same logic, let's have it in only one place please.

src/middlewared/middlewared/plugins/pool_/dataset_details.py
```diff
@@ -0,0 +1,221 @@
...
+    def get_nfs_shares(self, ds, nfsshares, mntinfo):
+        nfs_shares = []
+        found = False
+        for nfsshare in nfsshares:
+            if nfsshare['path'] == ds['mountpoint']:
+                # the share path is the actual dataset
+                found = True
+            else:
+                try:
+                    devid = os.stat(nfsshare['path']).st_dev
```
themylogin, line 129: You do `os.stat` for every NFS share path N times (where N = len(collapsed), can it make sense to cache it? Same for other share types.

src/middlewared/middlewared/plugins/pool_/dataset_details.py
```diff
@@ -0,0 +1,221 @@
...
+        for repl in filter(lambda x: x['direction'] == 'PUSH', repltasks):
+            # we only care about replication tasks that are configured to push
+            for src_ds in filter(lambda x: x == ds['id'], repl['source_datasets']):
+                count += 1
+
+        return count
+
+    @private
+    def get_cloudsync_tasks_count(self, ds, cldtasks):
+        return len([i for i in cldtasks if i['path'] == ds['mountpoint']])
```
themylogin, line 213: These seem to be handled differently than shares (we want exact path match), is that intentional?

themylogin (changes requested):

src/middlewared/middlewared/plugins/pool_/dataset_details.py
yocalebo: @themylogin I didn't use `snapshot_count` because it will run `zfs.snapshot.query` if it can't find the `/.zfs/snapshot` directory which makes this entire method unusable. That's why I said I'm not querying zvols right now.

src/middlewared/middlewared/plugins/pool_/dataset_details.py
yocalebo: Good point. I'll cache devid for the path and only stat if it doesn't exist.

src/middlewared/middlewared/plugins/pool_/dataset_details.py
themylogin: We can extract 
```
                if stat.st_ino == ZFSCTL.INO_SNAPDIR:
                    return stat.st_nlink - 2
```
into a separate method like `snapshot_count_for_stat`

themylogin (approved):

PR #9452, opened 2022-07-21
NAS-117245 / 22.12 / Improve `get_remote_addr_port` performance

themylogin:
`get_peer_process` calls `psutil.net_connections` which runs `readlink` on every file opened
in the system. This is slow when many files are open. To determine whether the other side
of the connection is nginx (which we do most of the time) we only need to check for connections
accepted by nginx's workers which is a much easier job.

### DIFF
--- a/src/middlewared/middlewared/utils/nginx.py
+++ b/src/middlewared/middlewared/utils/nginx.py
@@ -21,21 +21,22 @@ def get_remote_addr_port(request):
         except (KeyError, ValueError):
             pass
         else:
-            if process := get_peer_process(remote_addr, remote_port):
-                if process.name() == "nginx":
-                    try:
-                        with open("/var/run/nginx.pid") as f:
-                            nginx_pid = int(f.read().strip())
-                    except Exception:
-                        pass
-                    else:
-                        try:
-                            ppid = process.ppid()
-                        except psutil.ProcessNotFound:
-                            pass
-                        else:
-                            if ppid == nginx_pid:
-                                remote_addr = x_real_remote_addr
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/utils/nginx.py
```diff
@@ -21,21 +21,22 @@ def get_remote_addr_port(request):
...
+                    nginx_pid = int(f.read().strip())
+            except Exception:
+                pass
+            else:
+                try:
+                    process = psutil.Process(nginx_pid)
+                except psutil.ProcessNotFound:
+                    pass
+                else:
+                    if process.name() == "nginx":
```
yocalebo, line 35: wonder if using the context manager oneshot() would provide any speed boost
```
with process.oneshot():
    if process.name() == "nginx":
         blah
```

themylogin: @yocalebo `children` is not listed as a method that will benefit from `oneshot` https://psutil.readthedocs.io/en/latest/#psutil.Process.oneshot

yocalebo: Right but name() is.

yocalebo (approved):

src/middlewared/middlewared/utils/nginx.py
themylogin: What do you propose to group `name()` with to benefit from `oneshot()`?

yocalebo: Just making an observation. Not asking for anything to be changed.

PR #9458, opened 2022-07-22
NAS-117293 / 22.12 / Deprecate legacy behavior to allow empty homes path

anodos325:
In the days of TrueNAS 9 we allowed a customer to set an empty
string as the path for the SMB homes share. It was limited to
only LDAP users with samba schema. This has special behavior in
samba in that it will look up the path based on user home directory
and share that path out. Since these paths can be anywhere on the
server the result is potentially undefined behavior. Our product
has become more complex since those days with dataset-level
encryption and other features. Determining whether the share path
is locked also becomes impossible for fairly obvious reasons.
Hence, the undocumented customer-specific feature is being
removed.


themylogin (approved):

anodos325:
Waiting for CI run from custom branch in jenkins before merge.

PR #9463, opened 2022-07-22
NAS-117299 / 22.12 / Add internal endpoint to retrieve host paths being consumed by apps

sonicaj:
This PR adds an optimized / selective version of `chart.release.query` which only retrieves host paths which apps might be consuming.


yocalebo:
Tested on live hardware.

yocalebo (approved):

PR #9467, opened 2022-07-22
NAS-117306 / 22.12 / Fix ctdb jobs on pnn 0

anodos325:


yocalebo (approved):

anodos325:
This in specific affects time_info request for each node. Payload in this case is the pnn of the node requesting the info (which can be zero).

PR #9471, opened 2022-07-22
NAS-117392 / 22.12 / Add clustered time health check

anodos325:
We now have API endpoint in middleware that gets the time from
all cluster nodes. Use this as a periodic cluster health check
running on the node with the recovery master role to generate
an alert that time has drifted on one or more nodes to an
unacceptable level.

### DIFF
--- a/src/middlewared/middlewared/alert/source/ctdb.py
+++ b/src/middlewared/middlewared/alert/source/ctdb.py
@@ -1,4 +1,12 @@
-from middlewared.alert.base import AlertCategory, AlertClass, AlertLevel, SimpleOneShotAlertClass
+import errno
+
+from datetime import timedelta
+from middlewared.alert.base import Alert, AlertCategory, AlertClass, AlertLevel, AlertSource, SimpleOneShotAlertClass
+from middlewared.alert.schedule import IntervalSchedule
+from middlewared.service_exception import CallError
+
+ALLOWED_OFFSET_CLOCK_REALTIME = 120
+ALLOWED_OFFSET_NTP = 300000
 
 
 class CtdbInitFailAlertClass(AlertClass, SimpleOneShotAlertClass):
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/alert/source/ctdb.py
```diff
@@ -1,4 +1,9 @@
-from middlewared.alert.base import AlertCategory, AlertClass, AlertLevel, SimpleOneShotAlertClass
+from datetime import timedelta
+from middlewared.alert.base import Alert, AlertCategory, AlertClass, AlertLevel, AlertSource, SimpleOneShotAlertClass
+from middlewared.alert.schedule import IntervalSchedule
+
+ALLOWED_OFFSET_CLOCK_REALTIME = 120
```
yocalebo, line 5: Just curious but do you think we should store these 2 global options somewhere in the cluster plugin? Idk, can't really decide. I'll leave it up to you 😄

src/middlewared/middlewared/alert/source/ctdb.py
```diff
@@ -19,3 +24,67 @@ class CtdbClusteredServiceAlertClass(AlertClass, SimpleOneShotAlertClass):
...
+
+            return Alert(
+                ClusteredClockOffsetAlertSource,
+                {'verrs': errmsg},
+                key=None
+            )
+
+        worst_offset = max(rv['result'], key=get_offset)
+        if abs(worst_offset['ntp_peer']['offset']) > ALLOWED_OFFSET_NTP:
+            errmsg = f'NTP offset of node {worst_offset["pnn"]} exceeds 5 minutes.'
```
yocalebo, line 85: Also, do we have more than the `pnn`? I'm just wondering if we have ip address information so we could give a more helpful alert to end-user. i.e. `NTP offset of node {worst_offsent["pnn"]!r} with ip: {blah!r}....`

src/middlewared/middlewared/alert/source/ctdb.py
```diff
@@ -19,3 +27,102 @@ class CtdbClusteredServiceAlertClass(AlertClass, SimpleOneShotAlertClass):
...
+    run_on_backup_node = False
+
+    async def check(self):
+        def get_clock_realtime(entry):
+            return entry['clock_realtime']
+
+        def get_offset(entry):
+            return abs(entry['ntp_peer']['offset'])
+
+        if await self.middleware.call('smb.get_smb_ha_mode') != 'CLUSTERED':
```
yocalebo, line 50: Use the new `cluster.utils.is_clustered` api endpoint instead? Not sure, up to you.

yocalebo (approved):

PR #9478, opened 2022-07-26
NAS-116482 / 22.12 / Properly support snapshots based extents

Qubad786:
## Problem

ZFS snapshots based extents were not properly supported in `iscsi.extent` service which meant that users could specify snapshots as extents but have them not readyonly which in SCST's case resulted in scst failing to start.
Users were also not able to specify `snapdev` property for Volumes.

## Solution

Properly validate snapshot based extents and allow users to specify `snapdev` property.

### DIFF
--- a/src/middlewared/middlewared/plugins/iscsi_/extents.py
+++ b/src/middlewared/middlewared/plugins/iscsi_/extents.py
@@ -84,7 +84,7 @@ class iSCSITargetExtentService(SharingService):
         Bool('xen'),
         Str('rpm', enum=['UNKNOWN', 'SSD', '5400', '7200', '10000', '15000'],
             default='SSD'),
-        Bool('ro'),
+        Bool('ro', default=False),
         Bool('enabled', default=True),
         register=True
     ))
@@ -286,6 +286,9 @@ class iSCSITargetExtentService(SharingService):
             zvol_name = zvol_path_to_name(device)
             if not os.path.exists(device):
                 verrors.add(f'{schema_name}.disk', f'Device {device!r} for volume {zvol_name!r} does not exist')
+
+            if '@' in zvol_name and not data['ro']:
+                verrors.add(f'{schema_name}.ro', 'Must be set when disk is a ZFS Snapshot')
 
         elif extent_type == 'FILE':
             if not path:
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -1985,6 +1985,7 @@ def get_props_of_interest_mapping():
         ('special_small_blocks', 'special_small_block_size', None),
         ('pbkdf2iters', None, None),
         ('creation', None, None),
+        ('snapdev', None, None),
     ]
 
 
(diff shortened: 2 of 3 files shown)


src/middlewared/middlewared/plugins/iscsi_/extents.py
```diff
@@ -366,24 +366,16 @@ async def disk_choices(self):
         Return a dict of available zvols that can be used
         when creating an extent.
         """
-        diskchoices = {}
-
         zvols = await self.middleware.call(
             'zfs.dataset.unlocked_zvols_fast',
             [['attachment', '=', None]], {},
```
anodos325, line 371: Add additional filter here `['ro', '=', False]` and skip the other refactoring.

src/middlewared/middlewared/plugins/vm/vm_devices.py
```diff
@@ -59,11 +59,11 @@ async def disk_choices(self):
         out = {}
         zvols = await self.middleware.call(
             'zfs.dataset.unlocked_zvols_fast',
-            [["OR", [["attachment", "=", None], ["attachment.method", "=", "vm.devices.query"]]]],
-            {}, ['ATTACHMENT']
+            [['OR', [['attachment', '=', None], ['attachment.method', '=', 'vm.devices.query']]]],
+            {}, ['ATTACHMENT', 'RO']
```
anodos325, line 63: expand filter for `['ro', '=', False]`

anodos325:
Since this is fixing some basic functionality / easing people migrating from Core, I think it should go into angelfish. 
@yocalebo thoughts?

This means we're probably doing the migration out of order, right? Need write migration for angelfish, then after merging that branch, cherry-pick for PR against master and create merge migration IIRC. Though I do this infrequently and tend to forget precise steps. @themylogin can give accurate guidance.

yocalebo:
Yeah, this should go into 22.0.2.4 to ease migration from core which means the alembic migration will need to be done in stable and then backported to master.

yocalebo (changes requested):
Let's target this for 22.02.4 so we can ease migration from CORE to SCALE. Because of this, you'll need to make these changes in stable/angelfish repo first and the backport to master handling the alembic files accordingly. @themylogin can help in that respect if you have any questions.

themylogin:
@yocalebo I don't see any alembic migrations here, what do we want to do with alembic?

anodos325:
> @yocalebo I don't see any alembic migrations here, what do we want to do with alembic?

The PR changed as investigation into the underlying issue proceeded. No more migration now.

src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3099,6 +3100,7 @@ def transform(dataset):
         Bool('force_size'),
         Inheritable(Str('comments')),
         Inheritable(Str('sync', enum=['STANDARD', 'ALWAYS', 'DISABLED'])),
+        Inheritable(Str('snapdev', enum=['HIDDEN', 'VISIBLE']), has_default=False),
```
anodos325, line 3103: I wonder if we should default to `VISIBLE` on new zvol creation. If we do this though we may want to try creating 10K snapshots on a zvol and see whether anything breaks.

themylogin (approved):

yocalebo (approved):

PR #9491, opened 2022-07-26
NAS-117367 / 22.12 / Ensure that required paths are auto-created for clustered pwenc

anodos325:
Remove unnecessary check for `ctdb.general.healthy` from the
clustered generate_secret method. The secret_opener method
for the clustered plugin already checks that the fuse volume
is properly mounted. Only fuse/glusterfs access is required at this
point.

This also adds a missing os.open() call for case where we just
made a non-existent directory in pathref_open() method in
middlewared utils.


anodos325:
Only affects master.

yocalebo (approved):

PR #9492, opened 2022-07-27
NAS-117376 / 22.12 / Remove port from portal configuration in issci

sonicaj:
In SCALE, scst does not support specifying ports on portal level unlike freebsd. Scst only allows port to be configured for `iscsi-scst` daemon and then this port configuration is applied to all portals defined.

This PR introduces a migration moving port configuration to global iscsi configuration and updating relevant usages.

### DIFF
--- a/src/middlewared/middlewared/plugins/iscsi_/portal.py
+++ b/src/middlewared/middlewared/plugins/iscsi_/portal.py
@@ -4,7 +4,6 @@ import middlewared.sqlalchemy as sa
 
 from middlewared.schema import accepts, Dict, Int, IPAddr, List, Patch, Str
 from middlewared.service import CRUDService, private, ValidationErrors
-from middlewared.validators import Range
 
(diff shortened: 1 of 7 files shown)


src/middlewared/middlewared/plugins/iscsi_/portal.py
```diff
@@ -173,7 +172,7 @@ async def do_create(self, data):
 
     async def __save_listen(self, pk, new, old=None):
         """
-        Update database with a set new listen IP:PORT tuples.
+        Update database with a set new listen IP tuples.
```
yocalebo, line 175:
```suggestion
        Update database with new listen IPs.
```

yocalebo (approved):
This looks good to me; however, I'd like @themylogin to look it over since he's more knowledgeable with respect to alembic migrations.

themylogin (approved):

PR #9496, opened 2022-07-27
NAS-117378 / 22.12 / Bump up timeout values for permissions tests on cluster

anodos325:
Changing permissions on glusterfs fuse mounts performs somewhat
worse than I originally expected.


yocalebo (changes requested):
Can we make this a global variable to this file so we can change it in 1 place?

yocalebo (approved):

PR #9499, opened 2022-07-27
NAS-117074 / 22.12 / Fix kubernetes token issue

Qubad786:
## Problem

Recently we have seen some users who had malformed k3s server token when their server was shutdown unexpectedly i.e power loss.
I have not been able to locate exact reason why it got malformed in such cases but this results in kubernetes not working when the server is rebooted.

## Solution

The fix is quite straight forward, after investigation I found that the token is generally used to join multiple nodes and is used in encrypting bootstrap data of k3s which is used for joining nodes which does not apply to us.
If k3s had generated this token once, and the file is removed, k3s generates it again correctly which workarounds the problem of the token being malformed and the file being empty in some cases.

### DIFF
--- a/src/middlewared/middlewared/plugins/service_/services/kubernetes.py
+++ b/src/middlewared/middlewared/plugins/service_/services/kubernetes.py
@@ -1,10 +1,15 @@
-import os
 import asyncio
+import contextlib
+import os
+import re
 
 from middlewared.service import CallError
 from middlewared.utils import run
 
 from .base import SimpleService
+
+
+RE_TOKEN_FORMAT = re.compile(r'^K10.+:.+:.*')
 
 
 class KubernetesService(SimpleService):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/service_/services/kubernetes.py
```diff
@@ -58,6 +62,23 @@ async def before_start(self):
...
+        # is empty or malformed, in this case we should remove the token as it's generated automatically by k3s again
+        # and it's not useful to us as well per se as it's used for multi-node cluster )
+
+        await self.middleware.run_in_thread(
+            self.check_server_token, (await self.middleware.call('kubernetes.config'))['dataset']
+        )
+
+    def check_server_token(self, k8s_dataset):
+        token_path = os.path.join('/mnt', k8s_dataset, 'k3s/server/token')
+        if os.path.exists(token_path):
```
yocalebo, line 75: Checking a file for its' existence before opening it is prone to a race condition. I'd do something like this
```
token = ''
with contextlib.suppress(FileNotFoundError):
    with open(token_path) as f:
        token = f.read().strip()

if not RE_TOKEN_FORMAT.findall(token):
    os.unlink(token_path)
```

yocalebo (changes requested):

yocalebo (approved):
great!

PR #9511, opened 2022-07-28
NAS-117284 / 22.12 / Do not suggest to update to an older SCALE version

themylogin:
This only happens if current version is newer than the one on the update server (i.e. if nightlies are not being published), unlikely to happen on stable branch, hence no backport.


themylogin:
time 0:20

yocalebo (approved):

PR #9522, opened 2022-07-28
NAS-117371 / 22.12 / Make `type` field in pool topology and boot.get_state similar

rick-mesta:
Tested on local vm to produce the same results as `pool.query()`

### DIFF
--- a/src/middlewared/middlewared/plugins/boot.py
+++ b/src/middlewared/middlewared/plugins/boot.py
@@ -1,6 +1,6 @@
 import os
 
-from middlewared.schema import accepts, Bool, Datetime, Dict, Float, Int, List, Str, returns, Ref
+from middlewared.schema import accepts, Bool, Dict, Int, List, Str, returns
 from middlewared.service import CallError, Service, job, private
 from middlewared.utils import run
 from middlewared.validators import Range
--- a/src/middlewared/middlewared/plugins/boot.py
+++ b/src/middlewared/middlewared/plugins/boot.py
@@ -25,7 +25,6 @@ class BootService(Service):
         return BOOT_POOL_NAME
 
     @accepts()
-    @returns(Ref('pool_entry'))     # missing Patch
     async def get_state(self):
         """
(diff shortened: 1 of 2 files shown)


yocalebo (changes requested):
Looks good, however we now have to update the `@returns` decorator. Since it's a little extreme to list every single returned object (including all of its nested properties) we can do something like we did in `pool.dataset.details`

rick-mesta:
Kewl... I'm on it 👍🏼

src/middlewared/middlewared/plugins/boot.py
```diff
@@ -111,7 +111,26 @@ async def get_state(self):
...
+
+        # WebUI expects this to return the same data as `pool.pool_extend`
+        return {
+            'name': BOOT_POOL_NAME,
+            'status': info['status'],
+            'scan': info['scan'],
+            'topology': await self.middleware.call('pool.transform_topology', info['groups']),
+            'healthy': info['healthy'],
+            'warning': info['warning'],
+            'status_detail': info['status_detail'],
```
themylogin, line 124: It's great to have uniform data structures, but it would be even better if we could extract this into a separate method that'll be used both by `pool_extend` and `get_state`; otherwise, we are repeating ourselves here.

themylogin (changes requested):

src/middlewared/middlewared/plugins/boot.py
themylogin: We'll also be able to use `@returns(Ref('pool_entry'))` here (or `Patch` because we'll have some fields removed, e.g. `id`)

rick-mesta: Ok... I'll clean this up. Thx @themylogin

rick-mesta:
@yocalebo & @themylogin: I _**think**_ I got all your concerns addressed. Please let me know if this bad boy needs additional ❤️

themylogin (approved):

yocalebo (approved):

PR #9523, opened 2022-07-28
NAS-117424 / 22.12 / freenas-debug: Restore ZFS kstat capture 

freqlabs:
This was previously implemented for FreeBSD but never for Linux.

While here, some quick cleanup:
- Simplify iterating through pools. 
- Use meaningful variable names.


freqlabs:
time 1:00

amotin (approved):

PR #9525, opened 2022-07-28
NAS-117437 / 22.12 / Remove microsoft account option

anodos325:
This feature was initially added in FreeNAS 9 for convenience
of home users when Windows 10 was introduced. Windows 10 home
editions by default encouraged users to create microsoft accounts
and use them for authentication rather than using conventional
local user authentication. This had practical impact that
microsoft's SMB client would first attempt to authenticate to
new SMB servers using the microsoft account credentials
(email address and NT hash of password). The Microsoft Account
feature created a mapping between the user's email address
string and the local (FreeNAS / TrueNAS) user account name.
This mostly worked with some significant caveats in case of
active directory environments.

Unfortunately, times have changed and with Windows 11 Microsoft
now defaults to requiring hello signin when using microsoft
accounts for authentication (including over SMB). This means
that the "Microsoft  Account" feature is now broken by default
in Windows 11 and requires degrading security settings in
order to properly function. Since this is an unacceptable
requirement to users, the best course of action is to
simply remove the feature.


anodos325:
I left key in there so that we don't immediately break UI. Will remove key when ui team has updated.

anodos325:
Making a CI run also to make sure we don't break anything on that end.

themylogin (approved):

PR #9531, opened 2022-07-29
NAS-117436 / 13.0 / stop running file IO in main event loop

yocalebo:
Working on another ticket I saw that this specific alert was blocking the main event loop. `has_ipmi` calls `os.path.exists` 3 times which calls `os.stat` 3 times which can block event loop (especially prevalent  on slow boot disks).

### DIFF
--- a/src/middlewared/middlewared/alert/source/ipmi_sel.py
+++ b/src/middlewared/middlewared/alert/source/ipmi_sel.py
@@ -145,7 +145,7 @@ class IPMISELAlertSource(AlertSource):
     )
 
     async def check(self):
-        if not has_ipmi():
+        if not await self.middleware.run_in_thread(has_ipmi):
             return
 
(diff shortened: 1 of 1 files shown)


anodos325 (approved):

themylogin (approved):
Cool! But does FreeBSD issue IO request to `/` when accessing nonexistent file at `/dev`? 😮

yocalebo:
> Cool! But does FreeBSD issue IO request to `/` when accessing nonexistent file at `/dev`? 😮

I'm not sure but I'm not fixing this for non-existent devices. I'm fixing it for when the device(s) exist 😄

themylogin:
@yocalebo ah, then that's probably not a file IO, but some bus IO to the IPMI device (which is even worse)?

yocalebo:
> @yocalebo ah, then that's probably not a file IO, but some bus IO to the IPMI device (which is even worse)?

Right exactly. ipmi devices are notoriously (and often miserably) slow.

PR #9535, opened 2022-07-29
NAS-117419 / 22.12 / Better error message for `paramiko.ssh_exception.BadHostKeyException`

themylogin:


themylogin:
time 1:00

yocalebo (approved):

src/middlewared/middlewared/plugins/zettarepl.py
```diff
@@ -787,6 +787,15 @@ def _hold_task_reason(self, pools, dataset):
     async def _handle_ssh_exceptions(self):
         try:
             yield
+        except paramiko.ssh_exception.BadHostKeyException as e:
+            fingerprint = ":".join([hex(c)[2:] for c in e.key.get_fingerprint()])
+            raise CallError(
+                "Remote host identification has changed. Someone could be eavesdropping on you right now (man-in-the-"
+                "middle attack)! It is also possible that a host key has just been changed. The fingerprint for the "
```
anodos325, line 794: I think "Someone could be eavesdropping on you right now (man-in-the-middle attack)!" is unnecessarily emotive.

anodos325: I know it's standard from library.

themylogin: We should still a note about possible MITM, maybe not a such emotive one, otherwise it will look stupid ("go somewhere else, push a button, we can't do it for you for no reason") what do you think?

anodos325: Absolutely, it's worth explaining clearly what the problem is and possible implications of it. Maybe throw the error message at the docs team and see what they come up with.

PR #9546, opened 2022-07-31
NAS-117383 / 22.12 / Allow defining iotype for VM storage devices

Qubad786:
## Context

It was requested to allow defining `iotype` for storage devices. We default to `threads` now with this change as [Qemu documentation](https://www.qemu.org/docs/master/system/qemu-manpage.html) specifies.


themylogin (changes requested):
Should we add `iotype_choices` method for the UI team to use?

themylogin (approved):

PR #9548, opened 2022-08-01
NAS-117467 / 22.12 / Reuse tdb / ctdb handles

anodos325:
During testing there were intermittent failures when stressing
the clustercache interface (ctdb client op would fail with
deadlock avoided). This commit does two primary things:

1) Reuses tdb and ctdb handles (rather than opening / closing
   them on ops).

2) Initializes per-database reentrant lock to prevent multiple
   transactions from being initialized simultaneously.


yocalebo (approved):

william-gr:
backport

PR #9555, opened 2022-08-01
NAS-116380 / 22.12 / ensure unique ns/gws in network.general.summary

yocalebo:
Ensu;re duplicate default gateways and/or nameservers don't show up in the webUI.


yocalebo:
> Do we want to preserve the order of nameservers?

It's not necessary but probably the best thing to do here so made changes to preserve order.

themylogin (approved):
LGTM, thank you!

PR #9559, opened 2022-08-01
NAS-117479 / 22.12 / Skipping RAW device to be created

Qubad786:
### Issue
While cloning VMs, if there is a `RAW` device, it throws error as the path needs to be created manually, so the exception makes remaining devices to be skipped from creation.

### Fix
Skipping `RAW` type devices on cloning VMs.

### Future reference
Exceptions on cloning VMs should be managed appropriately, with proper rollback strategy.

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/clone.py
+++ b/src/middlewared/middlewared/plugins/vm/clone.py
@@ -112,8 +112,8 @@ class VMService(Service):
                         await self.__clone_zvol(vm['name'], zvol, created_snaps, created_clones)
                     )
                 if item['dtype'] == 'RAW':
-                    item['attributes']['path'] = ''
-                    self.logger.warn('For RAW disk you need copy it manually inside your NAS.')
+                    self.logger.warning('RAW disks must be copied manually. Skipping...')
+                    continue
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/vm/clone.py
```diff
@@ -112,8 +112,8 @@ async def clone(self, id, name):
                         await self.__clone_zvol(vm['name'], zvol, created_snaps, created_clones)
                     )
                 if item['dtype'] == 'RAW':
-                    item['attributes']['path'] = ''
-                    self.logger.warn('For RAW disk you need copy it manually inside your NAS.')
+                    self.logger.warn('For RAW disk you need copy it manually inside your NAS. Skipping...')
```
yocalebo, line 115: `logger.warn` has been deprecated. Can you use `self.logger.warning`

yocalebo, line 115:
```suggestion
                    self.logger.warning('RAW disks must be copied manually. Skipping...')
```

yocalebo (changes requested):

yocalebo (approved):
Thanks!

PR #9560, opened 2022-08-02
NAS-117302 / 22.12 / Integrate `disk_resize` script (overprovision)

yocalebo:
`disk_resize` was recently made to work on SCALE so this PR does the following:

1. create new `disk.resize` endpoint that takes a list of disks to be overprovisioned (or unoverprovisioned)
2. protect it with a lock so we don't clobber another instance of `disk_resize` for a particular disk
3. call it on pool.create

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/resize.py
+++ b/src/middlewared/middlewared/plugins/disk_/resize.py
@@ -7,6 +7,10 @@ from middlewared.service import Service, CallError, private, accepts, returns, j
 from middlewared.schema import Dict, Str, Int, List, Bool
 
 
+class UnexpectedFailure(Exception):
+    pass
+
+
 class DiskService(Service):
 
     @private
@@ -17,10 +21,15 @@ class DiskService(Service):
             err += f' SIZE: {disk["size"]} gigabytes'
             cmd.append(f'{disk["size"]}G')
 
-        cp = await run(cmd, stderr=subprocess.STDOUT, encoding='utf-8')
-        if cp.returncode != 0:
-            err += f' ERROR: {cp.stdout}'
-            raise OSError(cp.returncode, os.strerror(cp.returncode), err)
+        try:
+            cp = await run(cmd, stderr=subprocess.STDOUT, encoding='utf-8')
+        except Exception as e:
+            err += f' ERROR: {str(e)}'
+            raise UnexpectedFailure(err)
+        else:
+            if cp.returncode != 0:
+                err += f' ERROR: {cp.stdout}'
+                raise OSError(cp.returncode, os.strerror(cp.returncode), err)
 
     @accepts(
         List('disks', required=True, items=[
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/disk_/resize.py
```diff
@@ -0,0 +1,73 @@
...
+
+        if failures:
+            err = f'Failure resizing: {", ".join(failures)}'
+            if raise_error:
+                raise CallError(err)
+            else:
+                self.logger.error(err)
+        elif sync:
+            if len(data) > 1:
+                await (await self.middleware.call('disk.sync_all')).wait()
```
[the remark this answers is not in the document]
yocalebo: I believe so because `disk.sync` enumerates all disks on the system every time it's called 😄

src/middlewared/middlewared/plugins/disk_/resize.py
```diff
@@ -0,0 +1,88 @@
...
+        NOTE:
+            if `size` is given, the disk with `name` will be resized
+                to `size` (overprovision).
+            if `size` is not given, the disk with `name` will be resized
+                to it's original size (unoverprovision).
+        """
+        # since it's a list of disks, we could have received duplicate
+        # key,value pairs so ensure we don't try to run `disk_resize` on
+        # the same disk by removing duplicates
+        disks = [i for idx, i in enumerate(data) if i not in data[idx + 1]]
```
themylogin, line 66: I am not sure how does this work
```
>>> data = [{"name":"sda","size":16},{"name":"sda","size":16}]
>>> [i for idx, i in enumerate(data) if i not in data[idx + 1]]
  File "<stdin>", line 1, in <listcomp>
TypeError: unhashable type: 'dict'
```

yocalebo: Yeah, not really sure what I was doing there. Too many irons in the fire, I guess 😄 thanks for spotting it

src/middlewared/middlewared/plugins/disk_/resize.py
```diff
@@ -0,0 +1,94 @@
...
+
+        NOTE:
+            if `size` is given, the disk with `name` will be resized
+                to `size` (overprovision).
+            if `size` is not given, the disk with `name` will be resized
+                to it's original size (unoverprovision).
+        """
+        verrors = ValidationErrors()
+        disks = []
+        for disk in data:
```
themylogin, line 65: I'll keep bothering you... If someone requests `[{name: sda, size: 16}, {name: sda, size: 17}]` this check won't work.

themylogin (changes requested):

src/middlewared/middlewared/plugins/disk_/resize.py
```diff
@@ -0,0 +1,94 @@
...
+            if disk['name'] in disks:
+                verrors.add('disk.resize', 'Disk {disk["name"]!r} specified more than once.')
+            else:
+                disks.append(disk['name'])
+
+        verrors.check()
```
themylogin, line 71: I think you need `disks = data` here because now `disks` is an array of disk names, not an array of disks and their sizes like functions called below expect.

themylogin (changes requested):

src/middlewared/middlewared/plugins/disk_/resize.py
yocalebo: I'm not paying attention. I didn't mean to reuse `disks` variable 🤦

src/middlewared/middlewared/plugins/disk_/resize.py
```diff
@@ -0,0 +1,94 @@
...
+
+        exceptions = await asyncio.gather(*[self.resize_impl(disk) for disk in data], return_exceptions=True)
+        failures = []
+        success = []
+        for disk, exc in zip(disks, exceptions):
```
themylogin, line 76: Here too 😊

themylogin (changes requested):

src/middlewared/middlewared/plugins/disk_/resize.py
yocalebo: At this rate, I might as well just take rest of day off

themylogin (approved):
LGTM!

PR #9561, opened 2022-08-02
NAS-115620 / 22.12 / use nftables for SCALE HA fw rules

yocalebo:
`iptables` has been deprecated and will eventually be removed from upstream. Use the new `nft` (nftables) program and simplify this plugin.


yocalebo:
@sonicaj asking for your specific review since I don't know how this interacts with the various rules that we add for k8s.

sonicaj (approved):

PR #9564, opened 2022-08-03
NAS-117485 / 22.12 / Fix rare `OrderedDict mutated during iteration`

themylogin:
The issue occurs if a new job is started when we are iterating existing jobs.


themylogin:
time 0:30

sonicaj (approved):

PR #9574, opened 2022-08-03
NAS-117501 / 22.12 / disk_resize: Don't trigger udev events for NVMe

freqlabs:
Work around a bug in the Linux kernel for multipath NVMe devices.

The kernel uses GENHD_FL_HIDDEN to hide controllers from userspace, but
it's not hidden enough.  Udev sees the device, but it never handles
uevents, so it never settles when triggered.

Skip the triggering, we shouldn't need it.

### DIFF
--- a/src/freenas/usr/local/sbin/disk_resize
+++ b/src/freenas/usr/local/sbin/disk_resize
@@ -228,8 +228,7 @@ nvme)
 		exit 1
 	fi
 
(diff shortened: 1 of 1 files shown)


freqlabs:
time 2:00

yocalebo (approved):
Build failure is unrelated to these changes.

PR #9582, opened 2022-08-05
NAS-117524 / 22.12 / huge optimization to query_for_quota_alert

yocalebo:
On a system with ~1200 datasets, this method was taking ~10 seconds to complete and spiking a CPU core the entire time. After my changes, it takes ~0.9 seconds.

This is a ~91% decrease in time it takes to run this method. (Which, btw, is running every 5minutes in our alert system)

There is no change in functionality, I'm simply using some `extra` flags to `zfs.dataset.query` so we don't enumerate every property of every dataset and only the ones we want.


yocalebo:
retest this please

themylogin (approved):

PR #9590, opened 2022-08-05
NAS-117307 / 22.12 / Update ixvolumes during pool migration

Qubad786:
### Problem
After migrating ix-applications to a new pool with `kubernetes.update`, ix-volumes continue using the path which points to old pool unless they are edited/modified which is a serious concern.

### Solution
Update ix-volumes in helm configuration when ix-applications is migrated to use the newer pool.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/redeploy.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/redeploy.py
@@ -2,7 +2,7 @@ import errno
 import os
 
 from middlewared.schema import accepts, Str, Ref, returns
-from middlewared.service import CallError, job, Service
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/chart_releases_linux/redeploy.py
```diff
@@ -35,6 +40,13 @@ async def redeploy(self, job, release_name):
                 'isUpdate': True,
             }
         }, self.middleware)
+        if update_pool:
+            for index in range(len(config.get('ixVolumes', []))):
```
yocalebo, line 44: You can use enumerate here instead of range and len calls.

yocalebo (approved):

PR #9591, opened 2022-08-08
NAS-117527 / 13.0 / mute APEI/MCA corrected alerts

yocalebo:


themylogin (approved):

src/middlewared/middlewared/plugins/hardware_/mca_and_apei_events.py
```diff
@@ -3,8 +3,8 @@
 
 from middlewared.service import Service
 
-MCA = re.compile(r'.*MCA:.*(COR|UNCOR).*')
-APEI = re.compile(r'.*APEI.*(Recoverable|Fatal|Corrected|Informational).*')
+MCA = re.compile(r'.*MCA:.*(UNCOR).*')
+APEI = re.compile(r'.*APEI.*(Fatal).*')
```
amotin, line 7: I think you should keep "Recoverable" here still.  Unlike "Fatal" those errors do not lead to immediate panic, but they are still uncorrected, for example uncorrected memory error found by scrub -- not deadly right now, but still bad.

PR #9599, opened 2022-08-08
NAS-117557 / 22.12 / Improvements to ctdb.public.ips APIs

anodos325:
This introduces an optimization for case where changes should
happen on this node. `pnn` of -1 in update method means
"this node", absence of `pnn` in create method also means
"this node".

This PR forces ctdbd to delete the public IP address when disabling
to force migration away from node in question before disabling.

In addition to the two changes, a new method "ctdb.public.ips.delete"
is introduced which completely removes the public IPs line after
migrating away from the node in question.

### DIFF
--- a/src/middlewared/middlewared/plugins/cluster_linux/ctdb_ips.py
+++ b/src/middlewared/middlewared/plugins/cluster_linux/ctdb_ips.py
@@ -59,6 +59,10 @@ class CtdbIpService(Service):
                         f'{schema_name}.{data["interface"]}',
                         f'"{data["interface"]}" not found on this system.',
                     )
+
+        elif schema_name == 'public_delete':
+            return
+
         else:
             address = data.get('address') or data['public_ip']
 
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/cluster_linux/ctdb_ips.py
```diff
@@ -33,12 +33,16 @@ async def common_validation(self, data, schema_name, verrors):
...
         verrors.check()
 
+        if schema_name == 'public_delete':
+            return
+
         if schema_name in ('private_create', 'public_create'):
             existing_public_ips = {}
 
             if schema_name == 'public_create':
+                pnn = data.get('pnn') or await self.middleware.call('ctdb.general.pnn')
```
yocalebo, line 43: `pnn = data.get('pnn') or await self.middleware.call('ctdb.general.pnn')` can be called once above the `if schema_name in ('private_create', 'public_create'):` line I think

src/middlewared/middlewared/plugins/cluster_linux/ctdb_public.py
```diff
@@ -170,13 +171,30 @@ async def do_create(self, job, data):
...
         await self.middleware.call('ctdb.public.ips.reload')
 
         return await self.middleware.call('ctdb.public.ips.query', [('id', '=', data['pnn'])])
 
+    @accepts(Dict(
+        'delip_payload',
+        Int('pnn', required=False),
+        IPAddr('public_ip', required=True)
+    ))
+    def delete_ip(self, data):
```
yocalebo, line 188: We should probably protect this with a lock that is shared between `do_create` and `do_update` since it's a public endpoint, right? Or should this be private?

anodos325: Ah, forgot to make it private. The accepts() decorator is there for normalization.

yocalebo (approved):

PR #9601, opened 2022-08-08
NAS-117562 / 22.12 / Add more dataset details

yocalebo:


themylogin (approved):
Great job!
Some integration tests would be even more great. Especially with regards to `atime` and `casesensitive` keys.

PR #9602, opened 2022-08-08
iscsi_/extents.py blocks main event loop in many places

rick-mesta:


src/middlewared/middlewared/plugins/iscsi_/extents.py
```diff
@@ -302,14 +302,14 @@ async def clean_type_and_path(self, data, schema_name, verrors):
                         'You need to specify a filepath not a directory'
                     )
 
-            await check_path_resides_within_volume(
+            check_path_resides_within_volume(
```
themylogin, line 305: This is an `async def` function, it needs to be awaited

src/middlewared/middlewared/plugins/iscsi_/extents.py
```diff
@@ -401,7 +401,7 @@ async def save(self, data, schema_name, verrors):
 
             # create the extent
             if not os.path.exists(path):
-                await run(['truncate', '-s', str(data['filesize']), path])
+                run(['truncate', '-s', str(data['filesize']), path])
```
themylogin, line 404: This is an `async def` function, it needs to be awaited

src/middlewared/middlewared/plugins/iscsi_/extents.py
```diff
@@ -198,8 +198,8 @@ async def do_delete(self, id, remove, force):
             await self._service_change('iscsitarget', 'reload')
 
     @private
-    async def validate(self, data):
-        data['serial'] = await self.extent_serial(data['serial'])
+    def validate(self, data):
+        data['serial'] = self.extent_serial(data['serial'])
```
themylogin, line 202: `extend_serial` is an `async def` function, it needs to be awaited

src/middlewared/middlewared/plugins/iscsi_/extents.py
```diff
@@ -198,8 +198,8 @@ async def do_delete(self, id, remove, force):
             await self._service_change('iscsitarget', 'reload')
 
     @private
-    async def validate(self, data):
-        data['serial'] = await self.extent_serial(data['serial'])
+    def validate(self, data):
```
themylogin, line 201: If you change the function definition from `async` to non-async, you need to also change all its invocations from `await` to conventional calls

src/middlewared/middlewared/plugins/iscsi_/extents.py
```diff
@@ -265,7 +265,7 @@ async def clean_name(self, data, schema_name, verrors, old=None):
                 verrors.add(f'{schema_name}.name', 'Extent name must be unique')
 
     @private
-    async def clean_type_and_path(self, data, schema_name, verrors):
+    def clean_type_and_path(self, data, schema_name, verrors):
```
themylogin, line 268: Same not about definition and invocations here

themylogin (changes requested):

rick-mesta:
Thx for your feedback, @themylogin ... I'm trying to understand the underpinnings of the `async` / `await` semantics and will fix these soon.

rick-mesta:
https://ixsystems.atlassian.net/browse/NAS-117395

src/middlewared/middlewared/plugins/iscsi_/extents.py
```diff
@@ -302,14 +306,12 @@ async def clean_type_and_path(self, data, schema_name, verrors):
                         'You need to specify a filepath not a directory'
                     )
 
-            await check_path_resides_within_volume(
-                verrors, self.middleware, f'{schema_name}.path', path
-            )
+            self.validate_path_resides_in_volume(verrors, f'{schema_name}.path', path)
```
yocalebo, line 309: A synchronous function cannot call a coroutine so this will fail.

yocalebo (changes requested):
`async def validate` calls a function that calls `secrets.token_hex` which can block. There is also a function `extent_naa` that calls `hashlib.sha` which blocks.

src/middlewared/middlewared/plugins/iscsi_/extents.py
rick-mesta: I _think_ I've addressed all your concerns, @yocalebo ... but, please let me know if there are other gremlins 😛

src/middlewared/middlewared/plugins/iscsi_/extents.py
```diff
@@ -401,7 +408,7 @@ async def save(self, data, schema_name, verrors):
 
             # create the extent
             if not os.path.exists(path):
-                await run(['truncate', '-s', str(data['filesize']), path])
+                subprocess.run(['truncate', '-s', str(data['filesize']), path])
```
anodos325, line 412: Looking at implementation of truncate(1) in coreutils, I wonder if it would be better to implement in pure python, e.g. os.open(); os.ftruncate().

yocalebo (approved):

src/middlewared/middlewared/plugins/iscsi_/extents.py
```diff
@@ -105,11 +105,11 @@ async def do_create(self, data):
         `ro` when set to true prevents the initiator from writing to this LUN.
         """
         verrors = ValidationErrors()
-        await self.validate(data)
+        await self.middleware.run_in_thread(self.validate, data)
```
themylogin, line 108: This is perfectly working code, but here is how we do it everywhere else in order to be agnostic of whether called function is synchronous or asynchronous:
```suggestion
        await self.middleware.call('iscsi.extent.validate', data)
```

themylogin (changes requested):
Great job! There is one thing left (I left one suggestion, but it's applicable to a number of cases)

src/middlewared/middlewared/plugins/iscsi_/extents.py
rick-mesta: Thx @anodos325 -- I'll be more than happy to do that, but probably under the aegis of a separate PR.

src/middlewared/middlewared/plugins/iscsi_/extents.py
rick-mesta: Thx @themylogin -- is your point more about performing a `call` than a `run_in_thread` ?

themylogin: @rick-mesta yes

themylogin (approved):
LGTM, thank you!

rick-mesta:
Thx @themylogin 👍🏼

rick-mesta:
backport

rick-mesta:
backport

PR #9610, opened 2022-08-10
NAS-117602 / 22.12 / Privilege Management API

themylogin:


src/middlewared/middlewared/plugins/account_/privilege.py
```diff
@@ -0,0 +1,161 @@
...
+        List("local_groups", items=[Int("local_group")]),
+        List("ds_groups", items=[Int("ds_group")]),
+        List("allowlist", items=[Ref("allowlist_item")]),
+        Bool("web_shell", required=True),
+    )
+
+    @private
+    async def item_extend_context(self, rows, extra):
+        return {
+            "local_groups": {
```
themylogin, line 50: @anodos325 we probably want the same but for DS groups since the UI might want to display DS group names instead of DS group IDs. Will that have a good performance?

Ui might also probably want DS group names listing/autocompletion, will we be able to provide that?

themylogin:
time 10:00

themylogin:
time 10:00

themylogin:
@yocalebo @anodos325 ping

src/middlewared/middlewared/plugins/account_/privilege.py
```diff
@@ -0,0 +1,161 @@
...
+                g["id"]: {"id": g["id"], "name": g["group"]}
+                for g in await self.middleware.call("datastore.query", "account.bsdgroups", [], {"prefix": "bsdgrp_"})
+            },
```
anodos325, line 53: Why aren't you using group.query? You can get all local and DS groups at once there.

```
midclt call group.query '[]' '{"extra": {"additional_information": ["DS"]}}'
```
The `local` key for each entry differentiates between local and directory services groups:

```
  {
    "id": 90101103,
    "gid": 100001103,
    "name": "BILLY\\dnsupdateproxy",
    "group": "BILLY\\dnsupdateproxy",
    "builtin": false,
    "sudo": false,
    "sudo_nopasswd": false,
    "sudo_commands": [],
    "users": [],
    "local": false,
    "id_type_both": true,
    "nt_name": null,
    "sid": null
  }
  ```

Comment (anodos325) [reply src/middlewared/middlewared/plugins/account_/privilege.py:53]: There is a chance that a particular GID may not be resolvable through the query results (because we're hitting a cache for directory services case). In this situation you need to perform a second query for the group specifically by gid `[["gid", '=", gid]]`. 

There is special behavior in this case which will go directly do the getgrgid request, convert into our normal group dict, insert entry in cache, and return entry.

**Comment by anodos325** [comment]
Don't change the schema for `ldap` and `activedirectory`. These are clustered, but your other configuration options are node specific. Maybe make it a system setting.

File: src/middlewared/middlewared/plugins/account_/privilege.py [hunk 3]
```diff
@@ -0,0 +1,161 @@
...
+        response = await self.middleware.call(
+            "datastore.delete",
+            self._config.datastore,
+            id
+        )
+
+        return response
+
+    async def _validate(self, schema_name, data, id=None):
+        verrors = ValidationErrors()
```
anodos325, line 156: Are we doing any validation on the user-provided `resources`?

src/middlewared/middlewared/alembic/versions/22.12/2022-08-09_13-37_privilege.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('account_privilege',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('builtin_name', sa.String(length=200), nullable=True),
+    sa.Column('name', sa.String(length=200), nullable=False),
+    sa.Column('local_groups', sa.TEXT(), nullable=False),
+    sa.Column('ds_groups', sa.TEXT(), nullable=False),
```
anodos325, line 26: Does it make sense to have two different lists here since from OS standpoint they're just GIDs?

src/middlewared/middlewared/plugins/account_/privilege.py
```diff
@@ -0,0 +1,161 @@
...
+        List("allowlist", items=[Ref("allowlist_item")]),
+        Bool("web_shell", required=True),
+    )
+
+    @private
+    async def item_extend_context(self, rows, extra):
+        return {
+            "local_groups": {
+                g["id"]: {"id": g["id"], "name": g["group"]}
+                for g in await self.middleware.call("datastore.query", "account.bsdgroups", [], {"prefix": "bsdgrp_"})
```
anodos325, line 52: If you make a single list you can include the `local` key from `group.query` to differentiate between the two.

src/middlewared/middlewared/alembic/versions/22.12/2022-08-09_13-37_privilege.py
themylogin: Yes, I want a local fallback in cache DS is not reachable, that's why local groups must be explicitly specified for users (in most cases, one user) who is a local administrator.

src/middlewared/middlewared/plugins/account_/privilege.py
```diff
@@ -0,0 +1,161 @@
...
+        ("rm", {"name": "builtin_name"}),
+    ))
+    async def do_create(self, data):
+        """
+        Creates a privilege.
+
+        `local_groups` is a list of local user account groups that gain this privilege.
+
+        `ds_groups` is list of Directory Service group IDs that will gain this privilege.
+        """
```
anodos325, line 75: Can't users specify `allowlist` `name` and `web shell`? Maybe expand docstring.

src/middlewared/middlewared/plugins/account_/privilege.py
themylogin: As for API key, currently no. No values there could harm us in any way.

src/middlewared/middlewared/plugins/account_/privilege.py
themylogin: @anodos325 is `midclt call group.query '[]' '{"extra": {"additional_information": ["DS"]}}'` fast enough, does it always come from the cache?

src/middlewared/middlewared/plugins/account_/privilege.py
```diff
@@ -0,0 +1,161 @@
...
+                for g in await self.middleware.call("datastore.query", "account.bsdgroups", [], {"prefix": "bsdgrp_"})
+            },
+        }
+
+    @private
+    async def item_extend(self, item, context):
+        item["local_groups"] = [context["local_groups"][group_id]
+                                for group_id in item["local_groups"]
+                                if group_id in context["local_groups"]]
+        return item
```
anodos325, line 61: How are you handling case where group is added to list, then user deletes the local group?

src/middlewared/middlewared/plugins/account_/privilege.py
anodos325: Yes, it's always from cache. The only time it doesn't hit cache is if you have a filter like above.

src/middlewared/middlewared/plugins/account_/privilege.py
anodos325: But how do we tell user they've fat-fingered something?

themylogin: Nohow yet

src/middlewared/middlewared/plugins/account_/privilege.py
themylogin: Fixed.

There are no tests for user delete and no handling at all for user update. I am not sure how to test this, will think again on the second stage. not that important right now (I want to ship the API for the UI team first).

src/middlewared/middlewared/plugins/account_/privilege.py
themylogin: Done

themylogin:
time 6:00

themylogin:
@anodos325 fixed

anodos325:
```
root@truenas[/mnt/dozer/middleware/src/middlewared]# midclt call privilege.update 1 '{"ds_groups": [100000513]}'
[EINVAL] privilege_update.local_groups.0: This Directory Service group does not exist
```
Should this work?

```
root@truenas[/mnt/dozer/middleware/src/middlewared]# getent group 100000513
BILLY\domain admins:x:100000513:
```

```
root@truenas[/mnt/dozer/middleware/src/middlewared]# midclt call group.query '[["gid", "=", 100000513]]' '{"extra": {"additional_information": ["DS"
]}}'
[{"id": 100100513, "gid": 100000513, "name": "BILLY\\domain admins", "group": "BILLY\\domain admins", "builtin": false, "sudo": false, "sudo_nopasswd": false, "sudo_commands": [], "users": [], "local": false, "id_type_both": true, "nt_name": null, "sid": null}]
```

src/middlewared/middlewared/plugins/account_/privilege.py
```diff
@@ -0,0 +1,236 @@
...
+
+        await self._ensure_unique(verrors, schema_name, "name", data["name"], id)
+
+        groups = await self._groups()
+        for i, local_group_id in enumerate(data["local_groups"]):
+            if not self._local_groups(groups, [local_group_id]):
+                verrors.add(f"{schema_name}.local_groups.{i}", "This local group does not exist")
+        for i, ds_group_id in enumerate(data["ds_groups"]):
+            if not await self._ds_groups(groups, [ds_group_id]):
+                verrors.add(f"{schema_name}.local_groups.{i}", "This Directory Service group does not exist")
```
anodos325, line 169: `   verrors.add(f"{schema_name}.local_groups.{i}", "This Directory Service group does not exist")`
Shouldn't that be `f"{schema_name}.ds_groups.{i}"`?

themylogin:
@anodos325 fixed schema name and the actual bug in the DS groups validation

anodos325:
If I set an AD group in ds_groups and then delete the group from AD, it no longer appears in `privilege.query` results and so Administrator is unaware that the entry exists. This can lead to a security incident if GID is re-used for a different group. Contents of the list should always be visible in query results IMHO

themylogin:
@anodos325 fixed

anodos325:
We should have check during group creation for whether gid for new group is being used in any access rules. If it is, then I think we should raise a ValidationError (otherwise there can be unexpected privilege escalation on NAS). This can happen in two cases:

1. User deletes group and then reuses gid.
2. User creates a second group with the option to allow duplicate gids.

anodos325:
Since the situation with local users is fairly trivial to reproduce, we should probably include in our tests.
1. Create group, add to local groups, delete group, query privileges and verify that orphaned gid is still present
2. after (1) try to create new group reusing GID, verify fails with EINVAL.

anodos325:
When gid doesn't resolve, we return incomplete group entries.
```
      {
        "gid": 420
      },
```

I'm not sure whether it would make more sense to at a minimum include `group: None` for consistency for consumers.

src/middlewared/middlewared/plugins/account_/privilege.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        Creates a privilege.
+
+        `name` is a name for privilege.
+
+        `local_groups` is a list of local user account groups that gain this privilege.
+
+        `ds_groups` is list of Directory Service group IDs that will gain this privilege.
+
+        `allowlist` is a list of API endpoints allowed for this privilege. Empty list means all endpoints are allowed.
```
anodos325, line 75: Even if on backend we treat empty list as ALL, I don't think we should expose this to consumers. It seems like a loaded foot-gun. Maybe we should require something to be set, with special value `[ALL]` to indicate that all endpoints should be allowed.

yocalebo: I was getting ready to review with similar comment. I'd rather be explicit and have a designator like `[ALL]` to indicate all endpoints should be allowed.

anodos325:
This currently causes multiple CI failures. C.F. custom run from this branch. https://ci.tn.ixsystems.net/jenkins/job/TrueNAS%20SCALE%20-%20Unstable/job/API%20Tests%20-%20TrueNAS%20SCALE%20with%20middleware%20reinstalled%20from%20a%20branch%20Incremental/34/

anodos325:
Primary issue may be that it breaks user deletion:
```
root@truenas[~]# midclt call user.delete 67
[EINVAL] exclude_user_ids: Not a list
```

anodos325:
```
root@truenas[~]# midclt call privilege.update 1 '{"local_groups": [545, 1000]}'
'NoneType' object is not subscriptable
  File "/usr/local/lib/python3.9/dist-packages/middlewared/main.py", line 180, in call_method
    result = await self.middleware._call(message['method'], serviceobj, methodobj, params, app=self)
  File "/usr/local/lib/python3.9/dist-packages/middlewared/main.py", line 1284, in _call
  File "/usr/local/lib/python3.9/dist-packages/middlewared/service.py", line 931, in update
    rv = await self.middleware._call(
  File "/usr/local/lib/python3.9/dist-packages/middlewared/main.py", line 1284, in _call
  File "/usr/local/lib/python3.9/dist-packages/middlewared/schema.py", line 1152, in nf
    res = await f(*args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/middlewared/schema.py", line 1284, in nf
    return await func(*args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/middlewared/plugins/account_/privilege.py", line 129, in do_update
    await self._validate("privilege_update", new, id)
  File "/usr/local/lib/python3.9/dist-packages/middlewared/plugins/account_/privilege.py", line 169, in _validate
    if not await self._ds_groups(groups, [ds_group_id], include_nonexistent=False):
  File "/usr/local/lib/python3.9/dist-packages/middlewared/plugins/account_/privilege.py", line 221, in _ds_groups
    if group["local"]:
TypeError: 'NoneType' object is not subscriptable
```

Current privileges are:
```
root@truenas[~]# midclt call privilege.query | jq
[
  {
    "id": 1,
    "builtin_name": "LOCAL_ADMINISTRATOR",
    "name": "Local Administrator",
    "local_groups": [
      {
        "id": 91,
        "gid": 545,
        "group": "builtin_users",
        "builtin": true,
        "sudo": false,
        "sudo_nopasswd": false,
        "sudo_commands": [],
        "smb": true,
        "name": "builtin_users",
        "users": [
          66
        ],
        "local": true,
        "id_type_both": false,
        "nt_name": null,
        "sid": null
      },
      {
        "id": 98,
        "gid": 1000,
        "group": "smbuser",
        "builtin": false,
        "sudo": false,
        "sudo_nopasswd": false,
        "sudo_commands": [],
        "smb": false,
        "name": "smbuser",
        "users": [
          66
        ],
        "local": true,
        "id_type_both": false,
        "nt_name": null,
        "sid": null
      },
      {
        "gid": 9999,
        "group": null
      }
    ],
    "ds_groups": [
      {
        "gid": 420,
        "group": null
      },
      {
        "gid": 100000513,
        "group": null
      }
    ],
    "allowlist": [],
    "web_shell": true
  }
]
```

If I try to remove my directory services groups:
```
root@truenas[~]# midclt call privilege.update 1 '{"ds_groups": []}'            
[EINVAL] privilege_update.local_groups.2: This local group does not exist
```
I'm not sure that an invalid local group should prevent removal of directory services groups.

themylogin:
@anodos325 in `_ds_groups` I call
```
                    group = await self.middleware.call(
                        "group.query",
                        [["gid", "=", gid]],
                        {
                            "extra": {"additional_information": ["DS"]},
                            "get": True,
                        },
                    )
```
It should raise `MatchNotFound` if the group is not found because we specify `{"get": True}` (all our other methods behave this way). Looks like it just returns `None` or am I missing something?

> I'm not sure that an invalid local group should prevent removal of directory services groups.

You are still trying to save an invalid configuration. Most of our other methods behave the same: they don't allow saving invalid configuration even if you are just updating a small unrelated field. It's easier to program this way and I am not sure there'll be a good benefit of allowing to save an invalid configuration. Do you think I should fix this?

I ran the tests again https://ci.tn.ixsystems.net/jenkins/job/TrueNAS%20SCALE%20-%20Unstable/job/API%20Tests%20-%20TrueNAS%20SCALE%20with%20middleware%20reinstalled%20from%20a%20branch%20Incremental/35/#showFailuresLink the failures are the same as if we run from another branch, is any of them related to our changes?

anodos325:
>It should raise MatchNotFound if the group is not found because we specify {"get": True} (all our other methods behave this way). Looks like it just returns None or am I missing something?

Query raises MatchNotFound so the problem isn't there. I'll look more closely at where it's failing.

anodos325:
It's breaking on a call where the `include_nonexistent` argument is False. I think your `continue` is over-indented.

anodos325:
>  is any of them related to our changes?

```
Error Message
AssertionError: assert 1000 == (1001 + 1)
 +  where 1000 = call('group.get_next_gid')
Stacktrace
def test_group_next_gid():
        next_gid = call("group.get_next_gid")
        with mock("privilege.used_local_gids", f"""
            def mock(self):
                return {{{next_gid}: None}}
        """):
>           assert call("group.get_next_gid") == next_gid + 1
E           AssertionError: assert 1000 == (1001 + 1)
E            +  where 1000 = call('group.get_next_gid')

api2/test_account_privilege.py:100: AssertionError
```

I believe this is a test you're adding in this PR.

themylogin:
@anodos325 fixed both

src/middlewared/middlewared/alembic/versions/22.12/2022-08-17_13-37_privilege.py
```diff
@@ -0,0 +1,66 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('account_privilege',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('builtin_name', sa.String(length=200), nullable=True),
+    sa.Column('name', sa.String(length=200), nullable=False),
+    sa.Column('local_groups', sa.TEXT(), nullable=False),
+    sa.Column('ds_groups', sa.TEXT(), nullable=False),
+    sa.Column('allowlist', sa.TEXT(), nullable=False),
+    sa.Column('web_shell', sa.Boolean(), nullable=False),
```
anodos325, line 28: Do you need to initialize the allow list for the default administrators group to `*` ?

src/middlewared/middlewared/plugins/account_/privilege.py
anodos325: Need to update the docstring for `allowlist`

src/middlewared/middlewared/plugins/account_/privilege.py
```diff
@@ -0,0 +1,260 @@
...
+        "privilege_create",
+        ("rm", {"name": "builtin_name"}),
+    ))
+    async def do_create(self, data):
+        """
+        Creates a privilege.
+
+        `name` is a name for privilege.
+
+        `local_groups` is a list of local user account groups that gain this privilege.
```
anodos325, line 71: Should probably specify that these are GIDS (so that people don't think it's the DB IDs.

src/middlewared/middlewared/plugins/account_/privilege.py
```diff
@@ -0,0 +1,260 @@
...
+    @accepts(Patch(
+        "privilege_entry",
+        "privilege_create",
+        ("rm", {"name": "builtin_name"}),
+    ))
+    async def do_create(self, data):
+        """
+        Creates a privilege.
+
+        `name` is a name for privilege.
```
anodos325, line 69: May need to specify that names must be unique.

anodos325:
Once docstrings are updated, I don't see any other issues in testing. I'll defer to Caleb for approval of PR.

PR #9611, opened 2022-08-10
NAS-116747 / 22.12 / Provide string values for JSON fields that may overflow 2**53 on PiB-scale systems

themylogin:
Some JSON parsers (e.g. the one that is used in the `jq` utility for example) don't parse large values correctly, as a result we have this:
```
      "snapshot_limit": {
        "parsed": 18446744073709552000,
        "rawvalue": "18446744073709551615",
        "value": "none",
        "source": "DEFAULT",
        "source_info": null
      },
```
(the middleware JSON output was correct, but after going through `jq` the `parsed` and `rawvalue` become different).

Clients that know that their JSON parsing is poor can restore the correct ZFS property value using `rawvalue` which is always a string. For other JSON dictionaries that might contain integer values that are too large (this might happen when we start selling petabyte-scale systems) we'll provide `f'{key}_str` values like many other REST APIs do (e.g. Twitter tweet IDs, etc). These are all such values that I was able to find.

### DIFF
--- a/src/middlewared/middlewared/plugins/filesystem.py
+++ b/src/middlewared/middlewared/plugins/filesystem.py
@@ -483,12 +483,18 @@ class FilesystemService(Service):
         Int('total_blocks', required=True),
         Int('free_blocks', required=True),
         Int('avail_blocks', required=True),
+        Str('total_blocks_str', required=True),
+        Str('free_blocks_str', required=True),
+        Str('avail_blocks_str', required=True),
         Int('files', required=True),
         Int('free_files', required=True),
         Int('name_max', required=True),
         Int('total_bytes', required=True),
         Int('free_bytes', required=True),
         Int('avail_bytes', required=True),
+        Str('total_bytes_str', required=True),
+        Str('free_bytes_str', required=True),
+        Str('avail_bytes_str', required=True),
     ))
     def statfs(self, path):
         """
@@ -535,7 +541,7 @@ class FilesystemService(Service):
                 continue
             flags.append(flag)
 
-        return {
+        result = {
             'flags': flags,
             'fstype': mntinfo['fs_type'].lower(),
             'source': mntinfo['mount_source'],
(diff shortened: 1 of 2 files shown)


themylogin:
time 4:00

sonicaj (approved):

src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -552,6 +558,9 @@ def statfs(self, path):
             'free_bytes': st.f_bfree * st.f_frsize,
             'avail_bytes': st.f_bavail * st.f_frsize,
         }
+        for k in ['total_blocks', 'free_blocks', 'avail_blocks', 'total_bytes', 'free_bytes', 'avail_bytes']:
```
yocalebo, line 561: Can we add a comment about why we're doing this? 53 bit integer limitation for JSON isn't very well-known.

themylogin: @yocalebo I was thinking about doing that too, but I am not sure where. The same comment everywhere we do that? No. Can we just add this to some doc section?

yocalebo: Meh, I think the git commit explanation is good enough for any developer that comes along afterwards and wants to know why these fields exist.

yocalebo (approved):
Looks good, however, please wait for @anodos325 review too just to be sure.

src/middlewared/middlewared/plugins/filesystem.py
themylogin: git commit message will be lost with a refactoring commit, I'll add this to the docs

anodos325:
Are there any tests that need to be updated?

themylogin:
@anodos325 I'll ran the tests once they fix CI. I have not found any that will fail due to extra keys.

themylogin:
@anodos325 tests are fine, no regressions

themylogin:
@anodos325 are we good to merge this?

anodos325 (approved):

PR #9612, opened 2022-08-10
NAS-117599 / 22.12 / Remove failed pre-install job hooks before deleting chart release

sonicaj:
This PR introduces changes to remove any pods which might have been deployed by a pre-install helm hook which failed as this results in chart release failing to delete.


yocalebo (approved):
This looks fine to me, however, I'd like another set of eyes from @Qubad786 since they're much more familiar in this area.

Qubad786 (approved):
LGTM!

PR #9614, opened 2022-08-11
NAS-117601 / 22.12 / optmize system.general.get_ui_urls

yocalebo:
`psutil.net_connections` is too expensive and doesn't scale well on a busy system.

### DIFF
--- a/src/middlewared/middlewared/plugins/system_general/ui.py
+++ b/src/middlewared/middlewared/plugins/system_general/ui.py
@@ -144,7 +144,7 @@ class SystemGeneralService(Service):
                 urls.add(http_url)
                 if https_url:
                     urls.add(https_url)
-            elif i['address'] in config['ui_address']:
+            elif i['address'] in config['ui_address'] or i['address'] in config['ui_v6address']:
                 urls.add(http_url)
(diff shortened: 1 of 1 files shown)


themylogin (commented):
Is it possible to avoid using `net_connections` at all? When I was working on https://github.com/truenas/middleware/pull/9452 I discovered that it does `readlink` on every file opened in the system.

yocalebo:
> Is it possible to avoid using `net_connections` at all? When I was working on #9452 I discovered that it does `readlink` on every file opened in the system.

You read my mind. I actually added this comment in the jira ticket 😄 

```
I don’t like that this method uses psutils.net_connections every time it’s called as that doesn’t scale very well on a busy system. I think we should probably revisit this method in the future and simply use whatever is set in system.general db table instead of trying to parse what the nginx server is listening on.
```

themylogin:
@yocalebo we easily can do it now. My PR has the code which determines nginx main and worker processes, we'll only have to walk through their connections (I think it's good that we show IPs where nginx is actually listening at). You can assign this to me if you want.

themylogin (approved):

PR #9615, opened 2022-08-11
NAS-117617 / 22.12 / Allow setting snapdev field for filesystems

sonicaj:
This field is allowed to be set on filesystems by zfs and i believe we should do the same.


themylogin (approved):
How does it work? It's a no-op for filesystem, but child zvols inherit this value?

yocalebo:
While it's allowed, it does nothing for the filesystem.

sonicaj:
@themylogin yup, children inherit the property.

PR #9620, opened 2022-08-11
NAS-114198 / 22.12 / Fix sysctl tunables and introduce udev rules tunables

themylogin:


themylogin:
time 7:00

yocalebo:
What are you fixing about sysctl tunables?

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -58,79 +73,101 @@ async def tunable_type_choices(self):
...
-        data['orig_value'] = await self.middleware.call('tunable.get_or_set', data['var'])
-
-        if (comment := data.get('comment', '').strip()):
-            data['comment'] = comment
+        data['orig_value'] = ''
+        if data['type'] == 'SYSCTL':
+            data['orig_value'] = await self.middleware.call('tunable.get_sysctl', data['var'])
 
-        _id = await self.middleware.call(
+        id = await self.middleware.call(
```
yocalebo, line 108: Why change variable name that overwrites a builtin?

themylogin: That's what we do everywhere

themylogin:
> What are you fixing about sysctl tunables?

* Writes database-stored `orig_value` to the sysctl file for the disabled sysctls (but those can change between OS releases)
* Uses pseudo-service, I am trying to get rid of the services that are not actually services
* Small things (like incorrect field name in error or generally non-conventional code)

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -28,25 +28,40 @@ class Config:
...
+        self.set_sysctl(tunable['var'], tunable['orig_value'])
 
     @private
-    def get_or_set(self, var, value=None):
-        with open(f'/proc/sys/{var.replace(".", "/")}', 'r' if not value else 'w') as f:
-            return f.read().strip() if not value else f.write(value)
+    async def handle_tunable_change(self, tunable):
+        if tunable['type'] == 'UDEV':
+            await self.middleware.call('etc.generate', 'udev')
+            await run(['udevadm', 'control', '--reload-rules'])
```
yocalebo, line 64: Does this do anything?

themylogin: What exactly do you suspect of not doing anything?

yocalebo: What do you expect it to do?

themylogin: Reload the rules explicitly. Do you also want `udevadm trigger`?

yocalebo: I actually don't even see --reload-rules in Debian man pages. Also, if I'm understanding this it reloads all rules. "Signal systemd-udevd to reload the rules files and other databases like the kernel module index. Reloading rules and databases does not apply any changes to already existing devices; the new configuration will only be applied to new events."

themylogin: @yocalebo `--reload-rules` is an alias for `-R` https://github.com/systemd/systemd/blob/main/src/udev/udevadm-control.c#L66 I don't know why is it in all examples and in my fingers memory, but I can change it to `-R` if you want.

themylogin: I don't think it would be good to run `trigger` on every tunable change. Users can do that themselves if they want.

yocalebo: I don't believe udevadm control reload does anything but man pages are horrible in that regard and I'm too lazy to go read source right now.

yocalebo (approved):
Looks good. I'd like if we changed it to udevadm control -R just to make it consistent with Debian man pages.

yocalebo (changes requested):
This breaks SCALE HA systems because we ship with a `61-pmem.rules` file. Please don't merge this until you've tested this on real HA hardware.

themylogin:
@yocalebo real HA hardware in this case = anything with a pmem device?

yocalebo:
> @yocalebo real HA hardware in this case = anything with a pmem device?

yes please. Anything with a pmem device

yocalebo (approved):

PR #9621, opened 2022-08-11
NAS-117622 / 22.12 / Add SMB client failover test for cluster

anodos325:
This adds two new tests for cluster:
1) connect to SMB share on server, write to file. Reboot server.
   Re-establish session. Read contents of file.

   TODO: add python bindings for durable handles to that we
   can verify durable reconnects across node failover work correctly.

2) Verify that cluster becomes healthy again after reboot in (1)


yocalebo (approved):
Fantastic test

PR #9630, opened 2022-08-12
NAS-117416 / 22.12 / add alert for zpools that have USB disks

sonicaj:
This PR adds changes to add an alert for zpools that have USB disks. We saw a user configure apps to point to a zpool using a USB disk that caused dataset creation ( via zfs cli ) to take ~20 seconds and the same dataset on deletion took 1 minute. This caused a cascading set of failures because the zpool was so slow.

### DIFF
--- a/src/middlewared/middlewared/alert/source/pools.py
+++ b/src/middlewared/middlewared/alert/source/pools.py
@@ -6,7 +6,7 @@ class PoolUSBDisksAlertClass(AlertClass):
     category = AlertCategory.STORAGE
     level = AlertLevel.WARNING
     title = 'Pool consuming USB disks'
-    text = '%(pool)r is consuming %(disks)r USB disk(s) which is not advised behavior. ' \
+    text = '%(pool)r is consuming %(disks)r USB disk(s) which is not advised. ' \
            'Please replace the disk(s) immediately.'
 
 
--- a/src/middlewared/middlewared/alert/source/pools.py
+++ b/src/middlewared/middlewared/alert/source/pools.py
@@ -6,8 +6,7 @@ class PoolUSBDisksAlertClass(AlertClass):
     category = AlertCategory.STORAGE
     level = AlertLevel.WARNING
     title = 'Pool consuming USB disks'
-    text = '%(pool)r is consuming %(disks)r USB disk(s) which is not advised. ' \
-           'Please replace the disk(s) immediately.'
+    text = '%(pool)r is consuming USB devices %(disks)r which is not recommended.'
 
 
 class PoolDisksChecksAlertSource(AlertSource):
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/kubernetes_linux/update.py
```diff
@@ -142,8 +142,15 @@ async def validate_data(self, data, schema, old_data):
...
-        if data['pool'] and not await self.middleware.call('pool.query', [['name', '=', data['pool']]]):
-            verrors.add(f'{schema}.pool', 'Please provide a valid pool configured in the system.')
+        if data['pool']:
+            pool_details = await self.middleware.call('pool.query', [['name', '=', data['pool']]])
+            if not pool_details:
+                verrors.add(f'{schema}.pool', 'Please provide a valid pool configured in the system.')
+            elif await self.middleware.call('pool.get_usb_disks', pool_details[0]['id']):
+                verrors.add(
+                    f'{schema}.pool',
+                    f'Applications cannot be configured on {data["pool"]!r} pool as it consumes USB based disks.'
```
yocalebo, line 152: Since this returns an array of disks, probably would be nice to put them in the validation error so end user can see them.

src/middlewared/middlewared/alert/source/pools.py
```diff
@@ -0,0 +1,28 @@
+from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, Alert, AlertSource
+from middlewared.alert.schedule import CrontabSchedule
+
+
+class PoolUSBDisksAlertClass(AlertClass):
+    category = AlertCategory.STORAGE
+    level = AlertLevel.WARNING
+    title = 'Pool consuming USB disks'
+    text = '%(pool)r is consuming %(disks)r USB disk(s) which is not advised. ' \
```
yocalebo, line 9:
```suggestion
    text = '%(pool)r is consuming USB devices %(disks)r which is a non-supported configuration.' \
```

src/middlewared/middlewared/alert/source/pools.py
```diff
@@ -0,0 +1,28 @@
...
+           'Please replace the disk(s) immediately.'
```
yocalebo, line 10: Remove `Please replace the disk(s) immediately.` line

src/middlewared/middlewared/alert/source/pools.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class PoolDisksChecksAlertSource(AlertSource):
+    schedule = CrontabSchedule(hour=0)  # every 24 hours
+    run_on_backup_node = False
+
+    async def check(self):
+        alerts = []
+
+        for pool in await self.middleware.call('pool.query'):
```
yocalebo, line 20: Use `zfs.pool.query_imported_fast`

yocalebo (changes requested):

yocalebo (approved):

PR #9636, opened 2022-08-15
NAS-114305 / 22.12 / add cancel_rollback to interface API

yocalebo:
Requested by front-end team

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -563,19 +563,35 @@ class InterfaceService(CRUDService):
 
         await self.sync()
 
+    @private
+    async def checkin_impl(self, clear_cache=True):
+        if self._rollback_timer:
+            self._rollback_timer.cancel()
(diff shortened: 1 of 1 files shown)


themylogin (changes requested):
This code is no different from `checkin`, what is its purpose?

yocalebo:
> This code is no different from `checkin`, what is its purpose?

There is a difference, it doesn't remove the `_original_datastores` variable.

themylogin (changes requested):
@yocalebo so the idea is to stop timer but allow later manual rollback?

In this case, can we please re-use this code which is already repeated in this file twice?

yocalebo:
> @yocalebo so the idea is to stop timer but allow later manual rollback?
> 
> In this case, can we please re-use this code which is already repeated in this file twice?

The ticket that this is linked to has all those details.

themylogin:
Thank you.

Code reuse proposal still stands

themylogin (approved):

PR #9638, opened 2022-08-15
NAS-117398 / 22.12 / Storj iX cloud sync service

themylogin:


themylogin:
time 3:00

src/middlewared/middlewared/rclone/remote/storjix.py
```diff
@@ -0,0 +1,55 @@
...
+    rclone_type = "s3"
+
+    credentials_schema = [
+        Str("access_key_id", title="Access Key ID", required=True),
+        Str("secret_access_key", title="Secret Access Key", required=True),
+    ]
+
+    task_schema = []
+
+    async def list_buckets(self, credentials):
```
yocalebo, line 30: Your outer and inner methods are named the same. Please change name of inner method to something else or move it into its own method.

yocalebo (changes requested):

src/middlewared/middlewared/rclone/remote/storjix.py
themylogin: Will do that, but just curious why can't they be names the same?

yocalebo: Just seems strange to name it the same. When I see methods that are named the same, I'm thinking of method overwriting when inheriting from a parent or super class.

william-gr:
Was this tested using an account with the proper attribution? Just to make sure. Wasn't sure if this was already deployed on Storj side

themylogin:
@william-gr yes, I signed up for storj with US location and tested this

yocalebo (approved):
Thanks, lgtm. However, I'm curious if there is a way to get with QE/Storj so we can get some credentials from them for integration tests?

william-gr (approved):
Terrific, looks good to me (pending caleb's review)

Will this need require any UI change you think?

themylogin:
@william-gr here is UI ticket https://ixsystems.atlassian.net/browse/NAS-117676

@yocalebo I'll create a QE ticket

themylogin:
QE ticket https://ixsystems.atlassian.net/browse/DOCS-4119

themylogin:
backport

PR #9639, opened 2022-08-15
NAS-117684 / 22.12 / Separately test basic NFS ops for version 3 and version 4

anodos325:
This simplifies / reduces test count and also adds coverage
for basic ops over V4 protocol.


ericbsd (commented):
Is there any change that we need to do to the test runner?

anodos325:
> Is there any change that we need to do to the test runner?

No. This particular protocol API (SSH_NFS) is already in use in our tests. I ran it through a test run already.

ericbsd (approved):
LGTM

PR #9646, opened 2022-08-16
NAS-115869 / 22.12 / fix ntp on scale when dhcp is involved

yocalebo:
`ntp-systemd-wrapper` is set in the `ExecStart` clause of the `ntp.service` file. The one provided by Debian will check if a ntp dhcp file is present (which is created by default if DHCP hands out a NTP address) and will use that, ignoring the `/etc/ntp.conf` that we have generated.

This overwrites the upstream provided wrapper script with our own ignoring NTP peers provided via DHCP.


yocalebo:
retest this please

anodos325 (approved):

PR #9652, opened 2022-08-16
NAS-112047 / 22.12 / Allow to quiesce vms when taking zfs snapshots

sonicaj:


src/middlewared/middlewared/plugins/vm/vm_lifecycle.py
```diff
@@ -106,6 +109,46 @@ def restart(self, job, id):
 
         self.start(id, {'overcommit': True})
 
+    @item_method
+    @accepts(Int('id'))
+    @returns()
+    def suspend(self, id):
```
themylogin, line 115: Are "suspend" and "pause" the same thing?

src/middlewared/middlewared/plugins/zettarepl.py
```diff
@@ -213,9 +218,13 @@ def _observer(self, message):
...
                     context = self.vmware_contexts.pop(task_id, None)
-                    if context:
+                    vm_context = self.vm_contexts.pop(task_id, None)
+                    if context or vm_context:
                         with Client() as c:
-                            c.call("vmware.periodic_snapshot_task_end", context, job=True)
+                            if context:
+                                c.call("vmware.periodic_snapshot_task_end", context, job=True)
+                            if vm_context:
+                                c.call("vm.resume_suspended_vms", list(vm_context))
```
themylogin, line 227: Are we sure this call will never exceed the timeout of 60 seconds?

themylogin (dismissed):

src/middlewared/middlewared/plugins/vm/vm_lifecycle.py
sonicaj: Yes, they are the same thing

src/middlewared/middlewared/plugins/vm/attachments.py
```diff
@@ -1,22 +1,87 @@
...
+        datasets = {
+            d['id']: d for d in await self.middleware.call(
+                'pool.dataset.query', [['id', '^', f'{dataset}/']], {'extra': {'properties': []}}
+            )
+        }
+        to_ignore_vms = await self.get_vms_to_ignore_for_querying_attachments(True, [['pause_on_snapshot', '=', False]])
+        for device in await self.middleware.call(
+            'vm.device.query', [
+                ['dtype', 'in', ('DISK', 'RAW', 'CDROM')],
+                ['vm', 'nin', to_ignore_vms],
```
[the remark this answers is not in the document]
sonicaj: Yes `in/nin` work the same way - so in this case it would check vm key in the provided dict's keys.

src/middlewared/middlewared/plugins/zettarepl.py
sonicaj: The operation is quite quick, I would say if we have around 20-30 vms being suspended, it should be fine. Should we perhaps change these methods to a job ?

src/middlewared/middlewared/plugins/vm/vms.py
```diff
@@ -112,6 +114,7 @@ async def extend_vm(self, vm, context):
         Str('cpuset', default=None, null=True, validators=[NumericSet()]),
         Str('nodeset', default=None, null=True, validators=[NumericSet()]),
         Bool('pin_vcpus', default=False),
+        Bool('pause_on_snapshot', default=False),
```
yocalebo, line 117: You use `suspend` for all internal function names and also use `suspend_vms` as the schema arg in the zfs plugin but this schema arg uses `pause`. Can we change this to `suspend_on_snapshot`?

src/middlewared/middlewared/plugins/vm/vms.py
```diff
@@ -404,8 +408,7 @@ def status_impl(self, vm):
                 # Whatever happens, query shouldn't fail
                 return self._status(vm['name'])
             except Exception:
-                self.middleware.logger.debug(
-                    'Failed to retrieve VM status for %r', vm['name'], exc_info=True)
+                self.middleware.logger.debug('Failed to retrieve VM status for %r', vm['name'], exc_info=True)
```
yocalebo, line 411: `self.logger.debug`?

yocalebo (approved):

PR #9655, opened 2022-08-17
NAS-117186 / 22.12 / Update proftpd TLSProtocol to the new sane default

themylogin:


themylogin:
time 0:20

anodos325:
This seems like it's enabling TLSv1, TLSv1.1, and TLSv1.2. Is that intended? I think WebDav for instance we only allow TLSv1.2 and TLSv1.3.

themylogin:
@anodos325 fixed

anodos325 (approved):

PR #9657, opened 2022-08-17
NAS-117114 / 22.12 / Fix api docs not respecting port properly when redirecting

Qubad786:
## Context

Api docs were getting redirected in case of missing forward slash, but was missing port in case user had it changed. Changing the `$host` changing it to `$host_port` makes it accommodate ports in case of ports in url.


undsoft:
Looks like somebody used wrong ticket number.

Qubad786:
Ah yes @undsoft..fixing it

PR #9660, opened 2022-08-17
NAS-117663 / 13.0 / Revert "Load vendor's Realtek NIC driver. (#9075)"

amotin:
Appears the vendor driver can not handle physically non-contiguous
mbufs, that cause data corruptions with our iSCSI target.  Other
protocols should work fine, but we don't want to risk user data.

This reverts commit 2a15f78a84cb7c19c9be9457a97fa4b9fbaa2fc6.


rick-mesta (approved):

freqlabs (approved):
The lesser of two buggy drivers

PR #9664, opened 2022-08-18
NAS-117614 / 22.12 / Isolate middlewared run files in a dedicated directory

Qubad786:
## Context

It was requested that we isolate middleware related files under a dedicated directory under `/var/run`. The same is advised by [FHS standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html) when a program is to use multiple sentinel files.

### DIFF
--- a/src/middlewared/middlewared/client/client.py
+++ b/src/middlewared/middlewared/client/client.py
@@ -1,6 +1,6 @@
 from . import ejson as json
 from .protocol import DDPProtocol
-from .utils import ProgressBar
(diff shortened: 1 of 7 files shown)


themylogin (approved):

anodos325:
Did we make a successful run through API tests? (Just to make sure we exclude possibility that we have paths hardcoded in some of our other tools).

yocalebo:
No obvious regression from these changes according to API tests.

PR #9665, opened 2022-08-18
NAS-117618 / 22.12 / lessen the usage of pickle module

Qubad786:
## Context

It was requested that we avoid using `pickle` in `collectd_alert.py` and shift from using `/tmp` to dedicated middleware `/var/run/middleware` directory.

### DIFF
--- a/src/freenas/usr/local/libexec/collectd_alert.py
+++ b/src/freenas/usr/local/libexec/collectd_alert.py
@@ -25,14 +25,17 @@
 # SUCH DAMAGE.
 #
 
-import pickle as pickle
+import contextlib
+import json
 import os
 import re
 import sys
 
 from lockfile import LockFile, LockTimeout
+from middlewared.utils import MIDDLEWARE_RUN_DIR
 
-COLLECTD_FILE = '/tmp/.collectdalert'
+
+COLLECTD_FILE = os.path.join(MIDDLEWARE_RUN_DIR, '.collectdalert')
 
 
 def main():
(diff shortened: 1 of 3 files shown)


src/freenas/usr/local/libexec/collectd_alert.py
```diff
@@ -25,14 +25,14 @@
 # SUCH DAMAGE.
 #
 
-import pickle as pickle
 import os
 import re
 import sys
+import yaml
```
themylogin, line 31: Why yaml?

themylogin (changes requested):

yocalebo (changes requested):
Seems this is being merged into NAS-117614 instead of `master`?

src/freenas/usr/local/libexec/collectd_alert.py
```diff
@@ -46,9 +46,9 @@ def main():
 
     data = {}
     if os.path.exists(COLLECTD_FILE):
```
yocalebo, line 48: Let's go ahead and fix the 2 issues here.

1. this is a `flake8` error because it's an empty `except:` statement
2. let's remove the check for `os.path.exists` before opening this file for reading since it's prone to a race condition. Instead we should use `contextlib.suppress(FileNotFoundError)` context manager and then try to open for reading.

yocalebo (changes requested):

yocalebo (approved):

anodos325:
Did you investigate other usage of `pickle` e.g. mail plugin?

yocalebo (changes requested):
Okay looking at this a little closer I'd like to fix a few things.

1. `MIDDLEWARE_RUN_DIR` is defined in 2 different files which isn't ideal
2. We should probably define an `enum.Enum` class that houses the various configuration items (i.e. `/var/run/middleware` , `/middlewared.sock`, `middlewared-internal.sock` etc etc.) in a single place so we don't run into this problem in the future
3. We should review all usages of the `pickle` module while we're here since, as Andrew pointed out, the `MailQueue` class is using it as well.

yocalebo (approved):

yocalebo (approved):

PR #9667, opened 2022-08-18
NAS-117717 / 22.12 / hide default root user/group quotas

yocalebo:
No reason to expose this to end-user since it cannot be altered.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool.py
+++ b/src/middlewared/middlewared/plugins/pool.py
@@ -3984,9 +3984,10 @@ class PoolDatasetService(CRUDService):
         in an on-disk quota of 1 KiB.
         """
         dataset = (await self.get_instance(ds))['name']
-        quota_list = await self.middleware.call(
-            'zfs.dataset.get_quota', dataset, quota_type.lower()
-        )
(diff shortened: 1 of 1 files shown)


undsoft:
I don't think this is what we've discussed during the meeting though (about maintaining 1:1 relationship between UI and what's happening in ZFS).
I don't mind hiding it, but it would still be good to know:
1. Why is it there in the first place.
2. Why UI has Show All toggle.

yocalebo:
> I don't think this is what we've discussed during the meeting though (about maintaining 1:1 relationship between UI and what's happening in ZFS). I don't mind hiding it, but it would still be good to know:
> 
> 1. Why is it there in the first place.
> 2. Why UI has Show All toggle.

1. By default root has a quota on a path if there are files in that path. If another user/group quota is added to the path, then another entry will be returned.
2. Show All toggle can probably be gleaned from looking at the past ticket: https://ixsystems.atlassian.net/browse/NAS-104196. 

The webUI can pass query filters to this endpoint for `quota == 0` and `obj_quota == 0` which will hide disabled quotas. We need to show root because it's possible to make root have a quota on a given path.

PR #9678, opened 2022-08-22
NAS-117801 / 22.12 / Add some explicit tests for firstboot

anodos325:
Verify that our expected system dataset structure is created
and mounted with expected options during firstboot. Verify that
we don't have any unexpected services running before we start
tests.


yocalebo (approved):

ericbsd (approved):
LGTM

PR #9692, opened 2022-08-23
NAS-117725 / 22.12 / Fix ctdb.shared.volume.teardown and add integration tests

yocalebo:


cluster-tests/tests/cleanup/test_cluster_cleanup.py
```diff
@@ -0,0 +1,115 @@
...
+@pytest.mark.dependency(name='VERIFY_FUSE_UMOUNTED')
+def test_verify_all_gvols_are_fuse_umounted(ip, request):
+    depends(request, ['STOP_GVOLS'])
+
+    ans = make_request('get', '/gluster/volume/list')
+    assert ans.status_code == 200, ans.text
+
+    # we will try to check if each FUSE mountpoint is umounted `retries` times waiting at
+    # least `sleeptime` second between each attempt
+    retries = 10
```
anodos325, line 49: Should we maybe use the FUSE_OP_TIMEOUT from config? That way it can be tweaked if needed via environmental variable?

cluster-tests/tests/cleanup/test_cluster_cleanup.py
```diff
@@ -0,0 +1,115 @@
...
+            # give each node a little time to actually umount the fuse volume before we claim failure
+            rv = make_request('post', f'http://{ip}/api/v2.0/gluster/fuse/is_mounted', data={'name': i})
+            assert rv.status_code == 200
+            mounted = rv.json()
+            if mounted:
+                sleep(sleeptime)
+            elif not mounted:
+                break
+            elif retry == retries:
+                total = retries * sleeptime
```
anodos325, line 62: These conditionals need some love.

cluster-tests/tests/cleanup/test_cluster_cleanup.py
```diff
@@ -0,0 +1,115 @@
...
+                if 'Another transaction is in progress' in stop.text:
+                    # another command is running in the cluster so this isn't fatal but expected
+                    # so we'll sleep for `sleeptime` seconds to backoff and let cluster settle
+                    sleep(sleeptime)
+                else:
+                    assert False, f'Failed to stop {gvol!r}: {stop.text}'
+            elif stop.status_code == 200:
+                break
+            elif retry == retries:
+                assert False, f'Retried {retries} times to stop {gvol!r} but failed.'
```
anodos325, line 36: We could probably just:
`assert retry != retries, f'Retried {retries} times to stop {gvol!r} but failed.'`

yocalebo: Yep, you're right.

anodos325 (approved):

PR #9693, opened 2022-08-23
NAS-117624 / 22.12 / Fix vm devices choices

Qubad786:
## Context

`zfs.dataset.unlocked_zvols_fast` requires `RO` additional attribute to retrieve zvol read only status as we are using a filter to explicitly retrieve the results.


yocalebo (approved):

yocalebo (changes requested):
Wait, do we have integration tests for this? If not let's add one. We don't need to test actually setting up a VM, just need to test that this method returns zvols after we create one.

yocalebo (approved):
Awesome, thanks!

PR #9695, opened 2022-08-23
NAS-117800 / 22.02.4 / mask off unused systemd unit files

yocalebo:
This matches what we're already doing in master repo.


rick-mesta (approved):

doctore74:
thx

PR #9701, opened 2022-08-24
NAS-117802 / 22.12 / Update usage endpoint to use truenas domain

sonicaj:


yocalebo (changes requested):
Let's make this a global variable at top of module please. `USAGE_URL = https://...` or something

yocalebo (approved):

PR #9711, opened 2022-08-25
NAS-117848 / 22.12 / fix ipmi plugin and fix openipmi service

yocalebo:
Fix many problems

1. `ipmi` plugin subprocesses too many times to put in the main event loop so it's better to convert the plugin to be non-async
2. we were subprocess'ing 16 times every time middlewared started which is brutal, don't do that
3. we were calling blocking IO in the main event loop
4. `openipmi.service` loads the necessary kernel modules and creates the `/dev/ipmi0` device if a BMC is detected on the system. However, if a BMC doesn't exist the service always reported as failure to systemctl which is expected. Instead, we use dmidecode type 38 from SMBIOS spec to determine if IPMI device exists before we try to start `openipmi.service`
5. dmi functional tests were failing because we refactored the system plugin. The test wasn't updated so importing `system_.dmi` failed
6. add new functional tests for type 38 in `plugins/test_dmi.py`
7. `ipmi.query_sel` was broken and crashed in `TypeError` so fix that and also convert it to a `@job` because it's painfully slow on real hardware with a half-way decent IPMI implementation.

### DIFF
--- a/debian/debian/postinst
+++ b/debian/debian/postinst
@@ -70,6 +70,7 @@ systemctl enable zfs-zed
 systemctl mask libvirtd.socket libvirtd-ro.socket libvirtd-admin.socket libvirtd-tls.socket libvirtd-tcp.socket
 systemctl mask exim4-base.service exim4.service
 systemctl mask uuidd.service uuidd.socket
+systemctl mask openipmi.service
 
 # We don't use LVM and this service can add significant boot delays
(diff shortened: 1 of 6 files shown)


yocalebo:
Tested on real hardware and VMs as well as CI run. No immediate regressions found.

anodos325 (approved):

PR #9712, opened 2022-08-25
NAS-117865 / 22.12 / Fix timeout during idmap updates

anodos325:
Remove unnecessary winbindd restart. Wipe the winbindd cache
via python bindings rather than removing it.


yocalebo (approved):

yocalebo:
Approved with the stipulation that the flake8 issues are fixed 😄

PR #9718, opened 2022-08-26
NAS-117885 / 22.12 / Shift winbindd_cache.tdb path in middleware

anodos325:
The winbindd_cache.tdb file is stored in Samba's statedir to
facilitate offline logons. Since we don't enable or support
pam offline logons, we can safely move the cache away from
the system dataset to tmpfs. This avoids having potential
for winbindd to block system dataset moves.


yocalebo (approved):

yocalebo:
retest this please

PR #9728, opened 2022-08-29
NAS-117874 / 22.12 / handle disks formatted with Data Integrity Feature (dif)

yocalebo:
1. add new `dif` key to disks details dictionary
2. in `disk.format` raise a `CallError` if disk is formatted with DIF because we do not support it
3. add an alert that warns the user that their drives are formatted with DIF

Very simple scenario (that we've already seen) is that freeBSD ignores this feature altogether in kernel, however, linux kernel does not and will honor it. This means if a zpool is created on freeBSD with disks that are formatted with DIF and then moved to linux and the HBA honors DIF, the zpool will not import. Only way around that scenario is to disable this feature in the HBA driver so the zpool can be imported. However, the proper fix in that scenario is to replace each drive in the zpool 1-by-1 reformatting with DIF disabled. Removing DIF on a drive can take quite a long time because it has to fully erase the entire media.

Having an extra 8 bytes for checksum data (stored at the end of each 512byte sector) is completely unnecessary using ZFS since it does its own checksum calculation (by default) for each block written. DIF is redundant and, more importantly, a waste of bus bandwidth.

### DIFF
--- a/src/middlewared/middlewared/alert/source/dif_formatted_disks.py
+++ b/src/middlewared/middlewared/alert/source/dif_formatted_disks.py
@@ -8,7 +8,7 @@ class DifFormattedAlertClass(AlertClass, OneShotAlertClass):
     text = 'Disk(s): %s are formatted with Data Integrity Feature (DIF) which is unsupported.'
 
     async def create(self, disks):
-        return Alert(DifFormattedAlertClass, ', '.join(disks))
+        return Alert(DifFormattedAlertClass, ', '.join(disks), key=None)
 
     async def delete(self, alerts, query):
         return []
--- a/src/middlewared/middlewared/alert/source/dif_formatted_disks.py
+++ b/src/middlewared/middlewared/alert/source/dif_formatted_disks.py
@@ -1,0 +1,22 @@
+from middlewared.alert.base import Alert, AlertSource, AlertClass, AlertCategory, AlertLevel
+from middlewared.alert.schedule import CrontabSchedule
+
+
+class DifFormattedAlertClass(AlertClass):
+    category = AlertCategory.HARDWARE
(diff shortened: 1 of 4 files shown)


yocalebo:
retest this please

themylogin (commented):
I was wondering if we could just create one-shot alert on `disk.sync`/`disk.sync_all`, that way users will be notified immediately after system boot or disk insertion. This is not the gauge that might change during runtime, isn't it?

yocalebo:
> I was wondering if we could just create one-shot alert on `disk.sync`/`disk.sync_all`, that way users will be notified immediately after system boot or disk insertion. This is not the gauge that might change during runtime, isn't it?

good idea

src/middlewared/middlewared/alert/source/dif_formatted_disks.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class DifFormattedAlertClass(AlertClass, OneShotAlertClass):
+    category = AlertCategory.HARDWARE
+    level = AlertLevel.CRITICAL
+    title = 'Disk(s) Are Formatted With Data Integrity Feature (DIF).'
+    text = 'Disk(s): %s are formatted with Data Integrity Feature (DIF) which is unsupported.'
+
+    async def create(self, disks):
+        return Alert(DifFormattedAlertClass, ', '.join(disks))
```
themylogin, line 11: Can you please add `key=None`? Otherwise it will create a new alert each time an erroneous disks set is changed.

themylogin (approved):

src/middlewared/middlewared/alert/source/dif_formatted_disks.py
yocalebo: done

PR #9735, opened 2022-08-29
NAS-117927 / 22.12 / Remove dead smartctl code and fix functional tests

yocalebo:
The code that was removed is freeBSD specific and would have never worked on SCALE the way it was currently written. Furthermore, these hba drivers are pretty legacy while most don't even have in-kernel drivers. They have to be installed manually....which is also something we don't support.

Fixes employed in this PR:
1. remove dead code that was freeBSD specific
2. simplify `get_smartctl_args` now that most of it has been removed
3. fix the functional tests


themylogin (changes requested):
We don't even need `devices` in `context` right now

yocalebo:
> We don't even need `devices` in `context` right now

We actually do. I can remove a call to `await run(` because `devices` has a key that will be `USB` if it's a USB device.

themylogin (commented):
Cool! But doesn't `test_get_disk__unknown_usb_bridge` need updating too then?

yocalebo:
> Cool! But doesn't `test_get_disk__unknown_usb_bridge` need updating too then?

Yep sorry, I was too slow. I've updated it now.

themylogin (changes requested):
Sorry for nitpicking, but second question: do we even need `get_storage_devices_topology` now? I see there is only one remaining call, it can be perfectly substituted by the same `get_disks` method too

yocalebo:
> Sorry for nitpicking, but second question: do we even need `get_storage_devices_topology` now? I see there is only one remaining call, it can be perfectly substituted by the same `get_disks` method too

It's no problem and you're right. I just missed it. I'll fix and remove `get_storage_devices_topology' altogether

themylogin (approved):
Thank you!

PR #9746, opened 2022-08-30
NAS-117736 / 22.12 / Handle edge case for docker created time

sonicaj:
We are seeing an edge case where docker image might not have a valid creation timestamp which results in conversion to date time failing ( i was not able to create such an image though for reference ).
e.g
```
truecharts/tailscale                               latest            d689f53265f0   N/A             49.9MB
```

### DIFF
--- a/src/middlewared/middlewared/plugins/docker_linux/images.py
+++ b/src/middlewared/middlewared/plugins/docker_linux/images.py
@@ -1,4 +1,5 @@
 import aiodocker
+import contextlib
 import errno
 import itertools
 import os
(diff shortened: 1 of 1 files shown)


Ornias1993:
> We are seeing an edge case where docker image might not have a valid creation timestamp which results in conversion to date time failing ( i was not able to create such an image though for reference ). e.g
> 
> ```
> truecharts/tailscale                               latest            d689f53265f0   N/A             49.9MB
> ```

For future reference:
We've no idea why our container workflow pushed this specific container with an invalid data/time. However, if we randomly got this, we indeed all need to assume this can happen to any random container :)

Ornias1993 (approved):

yocalebo (approved):

PR #9747, opened 2022-08-30
NAS-117911 / 22.12 / Ensure SMB share ACL is preserved during various share ops

anodos325:
Background: SMB share ACLs are stored in Samba's share_info.tdb
and are keyed by the prefix 'SECDESC/' followed by the
share name. The primary CLI tool for interacting with the
share ACL is `sharesec` unfortunately because of sharesec's
reliance on libraries within samba that check whether the share
exists, some operations are not possible (like copying to a
non-existent share). Additionally, net_conf (the primary tool
for manipulating share acl entries) deletes share ACL entries
when a share is deleted. libsmbconf has no concept of a
"disabled" share this is implemented in middleware by adding
and removing the share as it is enabled and disabled.
Unfortunately, this results in the share ACL being deleted
and only restored during periodic task to back up the
share ACLs to our db.

Add two new helper functions to smb.sharesec:

1) smb.sharesec.toggle_share: backs up or restores a share's ACL
within share_info.tdb. We do this by creating an entry where
the share_name begins with a `#` character, and copying
ACL contents there. Since this is an invalid character
for a SHARE we don't have to worry about collisions here.

2) smb.sharesec.dup_share_acl: duplicates an existing share
ACL to a new entry. This allows us shift the share ACL
as shares are renamed.

Due to aforementioned limitations of the sharesec command, we
use python tdb bindings to accomplish these tasks. Since
share_info.tdb is on the system dataset, we now explicitly
close any open tdbs during systemdataset move.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -1182,6 +1182,11 @@ class SharingSMBService(SharingService):
                 # Forcibly closes any existing SMB sessions.
                 await self.close_share(oldname)
                 try:
+                    await self.middleware.call('smb.sharesec.dup_share_acl', oldname, newname)
+                except MatchNotFound:
+                    pass
+
+                try:
                     await self.middleware.call('sharing.smb.reg_delshare', oldname)
                 except MatchNotFound:
                     pass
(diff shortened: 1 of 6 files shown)


anodos325:
Still need to have add tests, but is ready for review otherwise.

yocalebo (approved):

yocalebo:
retest this please

PR #9755, opened 2022-08-31
NAS-117901 / 22.12 / optimize pools.py alert

yocalebo:
Move the `PoolUpgraded` and `PoolUSBDisks` alerts to be triggered on zfs based events to prevent polling.


src/middlewared/middlewared/alert/source/pools.py
```diff
@@ -9,22 +9,41 @@ class PoolUSBDisksAlertClass(AlertClass):
...
     schedule = CrontabSchedule(hour=0)  # every 24 hours
     run_on_backup_node = False
 
     async def check(self):
         alerts = []
+        usb_disks = {}
+        not_upgraded_zpools = []
+
+        for guid, info in (await self.middleware.call('zfs.pool.query_imported_fast')).items():
+            if info['state'] != 'ONLINE':
```
william-gr, line 33: Does this mean we wont get this for degraded pools? If so, why

themylogin (commented):
Same suggestion... can/should we do these checks on pool import or config change? IIRC we have some mechanism that will group 1000 equal consecutive pool config changes emitted by ZFS into one event so performance won't suffer.

yocalebo:
retest this please

src/middlewared/middlewared/alert/source/pools.py
```diff
@@ -1,30 +1,52 @@
...
+
+    async def is_upgraded(self, pool_name):
+        try:
+            return await self.middleware.call('zfs.pool.is_upgraded', pool_name)
+        except Exception:
+            return False
+
+    async def create(self, args):
+        pool = args['pool_name']
+        if await self.is_upgraded(pool):
```
themylogin, line 48: Isn't a pool "upgraded" if it has all the feature flags enabled? (and we generally want all pools to be upgraded and we create alerts if they are not upgraded)?

themylogin (approved):
Nice!

src/middlewared/middlewared/alert/source/pools.py
yocalebo: oops! Fixed :)

themylogin (approved):

PR #9760, opened 2022-09-01
NAS-117991 / 22.12 / Use zfs.pool.query_imported_fast in failover.status

yocalebo:
Use `zfs.pool.query_imported_fast` which is a vastly more performant way of getting basic zpool information compared to `pool.query` which is exactly what we want in methods like `failover.status` and `failover.status.get_local`


yocalebo:
retest this please

themylogin (approved):

PR #9761, opened 2022-09-02
Update GRUB console setup

usaleem-ix:
This commit fixes multiple issues we had found with GRUB console setup.

- Using console and serial as terminal_output for EFI boot mode causes
doubled output. In EFI boot mode, console outputs to all serial ports
by default, since both console and serial are writing to serial port.
terminal_output now uses 'gfxterm'.

- Changing the default serial port for serial console is possible via
WebUI. This updates the command line parameters for Linux console, but
serial port for GRUB console is not updated. GRUB console still uses
the default serial port. Adding '--unit' parameter to
GRUB_SERIAL_COMMAND with updated port number fixes this issue.

- Using GRUB_TERMINAL_INPUT and GRUB_TERMINAL_OUTPUT in place of
GRUB_TERMINAL allows to dictate input and output differently, which is
not possible with GRUB_TERMINAL variable. 'gfxterm' is not a valid
input.

- With serial console enabled, some of the output messages do not get
printed on serial console. The last console parameter dictates this,
which should be serial, if enabled.


themylogin (approved):

yocalebo:
Have we tested these changes on an x or m-series platform?

usaleem-ix:
> Have we tested these changes on an x or m-series platform?

No we have not tested these changes on x or m-series platform. Although, I have tested this on VMs with serial ports only (no gfx), with gfx and serial ports, and with gfx only options.

I think we should test this on x or m-series platform. I have built an image with these changes and the ones in https://github.com/truenas/grub2/pull/7. It is present [here](https://ci.tn.ixsystems.net/jenkins/job/TrueNAS%20SCALE%20-%20Unstable/job/Build%20-%20TrueNAS%20SCALE%20(Custom)/45/artifact/artifacts/).

src/freenas/usr/local/bin/truenas-grub.py
```diff
@@ -19,18 +19,22 @@ def get_serial_ports():
     # 22.02-RC.2 use real hostid. We need to be able to boot both of these configurations.
     config = [
         'GRUB_DISTRIBUTOR="TrueNAS Scale"',
+        'GRUB_TIMEOUT=10',
```
amotin, line 22: Is this related, or it is a leak?

src/freenas/usr/local/bin/truenas-grub.py
```diff
@@ -19,18 +19,22 @@ def get_serial_ports():
...
         'GRUB_CMDLINE_LINUX_DEFAULT="libata.allow_tpm=1 amd_iommu=on iommu=pt '
         'kvm_amd.npt=1 kvm_amd.avic=1 intel_iommu=on zfsforce=1'
         f'{f" {kernel_extra_options}" if kernel_extra_options else ""}"',
     ]
 
-    terminal = ["console"]
+    terminal_output = ["gfxterm"]
```
amotin, line 28: Do we need to use "gfxterm" if we are not using "serial"?  Does it give any benefits?  I would guess "console" to be more universal.  I guess it may react on serial console redirection setting in BIOS, at least I would check it.

src/freenas/usr/local/bin/truenas-grub.py
usaleem-ix: This is related. In the installer issue we realized that 5 seconds is sometimes not enough to get into serial console.

src/freenas/usr/local/bin/truenas-grub.py
usaleem-ix: Using 'gfxterm' instead of 'console' helps in two different problems. 'gfxterm' only outputs to graphics console and it would fail if there is not graphics option. Using 'gfxterm' and serial in this way would help in handling both EFI and legacy BIOS systems independent of serial redirection setting in BIOS.

amotin: > 'gfxterm' only outputs to graphics console and it would fail if there is not graphics option.

Right. Is it good in some way to fail there and get no console?

> Using 'gfxterm' and serial in this way would help in handling both EFI and legacy BIOS systems independent of serial redirection setting in BIOS.

Again I am not sure why cooperation with BIOS settings is bad?  If I have a system with console redirection in BIOS, I would prefer to have the redirection in GRUB menu also, sure without double output.  Speaking about EFI and BIOS it would be good to have the logic in comments for both cases if they behave differently.

usaleem-ix: > Right. Is it good in some way to fail there and get no console?

We should not fail and have no console in any case. This was the reason to patch the GRUB config generation. With serial console enabled, if 'gfxterm' fails, we fallback to serial console only. Without serial console, if 'gfxterm' fails we do set terminal_output to console in that case. And this happens only if serial is disabled.

> Again I am not sure why cooperation with BIOS settings is bad?

We basically cannot get rid of doubled output while using console and working with serial console options. We cannot know if serial redirection is enabled in BIOS. Moreover, this makes the serial console selection from WebUI inconsistent. Console writes to all serial ports in EFI while user could have selected a specific port from WebUI and complain that all serial ports have console.

usaleem-ix:
So far, this solution had precedence over serial redirection. This means that if serial redirection is enabled in BIOS user should be able to control all serial output including GRUB from WebUI through serial console option.

We have decide to update this as follows, so that console redirection works as well:

- If serial console is disabled, use 'console' as both `teminal_input` and `terminal_output` for both BIOS and EFI.
- If serial console is enabled and it is EFI boot mode, we should use 'gfxterm' combined with 'serial' as we have done here.
- If serial console is enabled with BIOS mode, we should use 'console' and 'serial' as both `terminal_input` and `terminal_output`.

usaleem-ix:
Sounds good. We can set 'console' in terminal output in that case for UEFI if it is 'disabled', otherwise if it is set to some serial port, we use 'gfxterm'.

amotin:
I actually meant "gfxterm" in case of "Disabled" , and "console" when "Configure Serial Console" is unchecked.

usaleem-ix:
Got it. Thank you.

usaleem-ix:
Updated to use console for:

- BIOS mode with serial console enabled
- BIOS mode with serial console disabled
- EFI mode with serial console disabled

For EFI mode with serial console enabled, we use 'gfxterm' instead of 'console' to fix double output.

src/freenas/usr/local/bin/truenas-grub.py
```diff
@@ -19,18 +20,24 @@ def get_serial_ports():
...
-        terminal.append("serial")
         port = get_serial_ports().get(advanced['serialport'], advanced['serialport'])
-        cmdline.append(f"console={port},{advanced['serialspeed']} console=tty1")
+        port_nr = port.replace('ttyS', '')
+        config.append(f'GRUB_SERIAL_COMMAND="serial --unit={port_nr} --speed={advanced["serialspeed"]} --word=8 --parity=no --stop=1"')
+        terminal_output.append("serial")
+        terminal_input.append("serial")
+        cmdline.append(f"console=tty1 console={port},{advanced['serialspeed']}")
+        if os.path.exists("/sys/firmware/efi"):
+            terminal_output[0] = "gfxterm"
```
amotin, line 40: It is probably fine, but it looks dirty to me to use both append() and direct indexes together.  I would probably move this "if" block before the terminal_output.append("serial") and do terminal_output = ["gfxterm"] instead.

amotin (approved):

src/freenas/usr/local/bin/truenas-grub.py
usaleem-ix: Thank you. I have updated it as you suggested.

usaleem-ix:
I am building an image with updated changes to test on x or m series hardware.

src/freenas/usr/local/bin/truenas-grub.py
freqlabs: In this case it would be giving us time to pick a boot environment. But really it just needs to be enough time to press an arrow key to stop the timeout.

We do provide a way to configure which BE to boot next through the web UI and cli, so there isn't much of a reason to go into this menu unless your BE is broken so badly that it doesn't boot. Hopefully that would be pretty rare.

These are potentially valuable seconds of downtime in the installed system. On the other hand, it's even worse if you miss the boot menu and need to go through another reboot cycle to try again. Ten seconds does feel pretty high, but not unreasonably so.

freqlabs (approved):
Assuming testing goes well, this looks like it should be OK.

usaleem-ix:
I validated these changes and the ones in https://github.com/truenas/grub2/pull/7 on X10 hardware with BIOS and EFI boot modes. These changes are working as expected. IPMI console is also operational to present GRUB boot menu.

We can go ahead and merge the PRs.

yocalebo:
Awesome, thanks @usaleem-ix !

PR #9766, opened 2022-09-02
NAS-118015 / 22.12 / prevent blocking event loop when checking updates

yocalebo:
Does 4 things:

1. no reason for `trains_linux.py` so combine it to `trains.py`
2. stop blocking the event loop by adding `get_manifest_file` method and calling it appropriately
3. simplify the code
4. move from `update.freenas.org` to `update.ixsystems.com` since the former is deprecated.


william-gr:
While you're there, should we change the update server from update.freenas.org? IIRC IT said its "deprecated"?

src/middlewared/middlewared/plugins/update_/trains.py
```diff
@@ -1,54 +1,62 @@
...
+
     @private
-    async def get_scale_trains_data(self):
-        async with async_timeout.timeout(INTERNET_TIMEOUT):
-            async with aiohttp.ClientSession(
-                raise_for_status=True, trust_env=True,
-            ) as session:
-                trains = await (await session.get(f"{scale_update_server()}/trains.json")).json()
+    def get_manifest_file(self):
+        with open(SCALE_MANIFEST_FILE) as f:
```
themylogin, line 16: I think you can read it on module load too (using `pathlib`), it does not change during system runtime

themylogin (approved):

PR #9776, opened 2022-09-06
Do not allow unsetting host path validation for enterprise users

Qubad786:
## Context

It was requested that we should not allow unsetting host path validation in Applications for ENTERPRISE users.


sonicaj (approved):

sonicaj (commented):
@Qubad786 we should be doing `failover.hardware != 'MANUAL'` instead and only disable this for HA compatible hardware as @yocalebo reminded me that `SCALE_ENTERPRISE` also includes truenas mini for which we do want the ability to allow users to disable validation. Thanks!

PR #9781, opened 2022-09-07
NAS-118041 / 22.12 / Do not backup catalogs dataset on kubernetes backup

Qubad786:
## Context

We should not be backing up contents of catalogs dataset on kubernetes backup and also making sure we give it a fresh start on `kubernetes.restore` so as to sync current state from user specified selection of catalogs.


Ornias1993 (approved):
As discussed with @sonicaj before, greatly appreciated :)

sonicaj (approved):

PR #9799, opened 2022-09-08
NAS-118092 / 22.02.4 / Properly terminate Web Shell and it's children

themylogin:
Sometimes when `/usr/bin/login` (which is a master pid for our Web Shell) is terminated, shell (and its children) stays alive. Let's terminate the entire process tree.

(cherry picked from commit 2a0d8af8bc99d3516d9c06e1a4404ed8de6cd033)

### DIFF
--- a/src/middlewared/middlewared/main.py
+++ b/src/middlewared/middlewared/main.py
@@ -60,6 +60,7 @@ import urllib.parse
 import uuid
 import tracemalloc
 
+import psutil
 from systemd.daemon import notify as systemd_notify
 
 from . import logger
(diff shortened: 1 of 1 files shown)


yocalebo (commented):
These are blocking calls

yocalebo (changes requested):
Woops, forgot to request changes when I added my previous comment.

yocalebo (approved):

src/middlewared/middlewared/main.py
```diff
@@ -794,29 +795,25 @@ async def run(self, ws, request, conndata):
         return ws
 
     async def worker_kill(self, t_worker):
-        # If connection has been closed lets make sure shell is killed
-        if t_worker.shell_pid:
+        def worker_kill_impl():
+            # If connection has been closed lets make sure shell is killed
+            if t_worker.shell_pid:
```
yocalebo, line 800: Actually, I can't remember details but does the inner method have access to the outer methods args without specifing `nonlocal t_worker` inside `worker_kill_impl`?

themylogin: @yocalebo yes, the inner method has access to all the variables from the outer scope, but if you assign to the outer scope variable `t_worker = t_worker + 1` it will become a local variable, so they added `nonlocal t_worker` to prevent this from happening

themylogin:
time 00:30

PR #9803, opened 2022-09-08
NAS-118093 / 22.02.4 / Dont block event look in check_permission hook

yocalebo:
This is called in the main event loop no matter what so for stable/angelfish I've added a `check_permission_impl` which is called in a thread from `async def check_permission`. A better fix will go into master.


themylogin (approved):

yocalebo:
backport

PR #9834, opened 2022-09-12
NAS-117952 / 22.12 / Add initial containers in pod logs choices

Qubad786:
## Problem

Init containers were not being accounted for when we were retrieving containers in a pod.

## Solution

Init containers need to be separately referenced in order to retrieve them properly.


Ornias1993 (commented):
Clean solution, looks good :)

sonicaj (approved):

PR #9842, opened 2022-09-13
NAS-118152 / 22.12 / set CPU governor to performance mode always

yocalebo:


themylogin (approved):

SuperQ:
Hmm, that Jira issue isn't public. Can you post a "why" this is being done?

I personally would like to keep my system in the default powersave mode, which better adjusts system performance based on the intel_pstate driver.

kmoore134:
In our testing it made a significant difference in overall system performance and throughput. We may allow it to be tuned back to other profiles later though, for the few who do not care about performance.

SuperQ:
"Don't care about performance" is a bit projecting assumptions about what people want.

kmoore134:
Ha, let me clarify then. Those who do not require max performance 100% of the time.

PR #9846, opened 2022-09-14
NAS-118168 / 22.12 / Do not require `remotehost` when rsync task is configured using an SS…

themylogin:
…H connection from the keychain


themylogin:
time 1:00

yocalebo (approved):

PR #9857, opened 2022-09-15
NAS-118197 / 22.12 / Fix k3s logs/exec issue

Qubad786:
## Problem

With recent upgrade of kubernetes, k3s logs could not be viewed and neither would `exec` work for pods with kubernetes complaining about SSL certificate not containing explicit node IP.

## Solution

K3s generates SSL certificates itself required for functioning of kubernetes but it only by default adds the node IP to the certificate. For users who have selected wildcard ( `0.0.0.0` ) in the settings, the SSL certificate will have a SAN entry for `0.0.0.0` only but the actual node IP would be some IP on one of the interfaces of the system..
This results in failure to see logs or exec in pods. Proposed workflow is to provide all available ips to k3s so it adds explicit SAN entries for each IP available and then for cases like where we are using wildcard IP, the SSL certificate of kubernetes would still contain the actual node IP and users can access logs or exec pods.

### DIFF
--- a/src/middlewared/middlewared/etc_files/rancher/k3s/flags.py
+++ b/src/middlewared/middlewared/etc_files/rancher/k3s/flags.py
@@ -41,6 +41,9 @@ def render(service, middleware):
             'cluster-dns': config['cluster_dns_ip'],
             'data-dir': os.path.join('/mnt', config['dataset'], 'k3s'),
             'node-ip': config['node_ip'],
+            'node-external-ip': ','.join([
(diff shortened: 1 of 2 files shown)


sonicaj (commented):
Good catch, however let's please add integration tests for this too so we would have been able to catch this earlier in that case. Thanks!

yocalebo (approved):

sonicaj (approved):
Thanks!

PR #9858, opened 2022-09-15
AUTO-235 / 22.12 / Changed device test to read SERIAL dynamically instead of statically

ericbsd:


yocalebo (changes requested):
You can do "assert any((i["drivername"]== "uart" for i in results.json))

q5sys (approved):

ericbsd:
> You can do "assert any((i["drivername"]== "uart" for i in results.json))

I will go with that.

yocalebo (approved):
Cool!

PR #9863, opened 2022-09-16
NAS-118123 / 22.12 / Flush alerts to the database immediately after deleting oneshot alerts

themylogin:


themylogin:
time 3:00

yocalebo (changes requested):
This means on every `alert.oneshot_delete, klass` we're going to empty the database and regenerate everything.

Can we not pass the `klass` to `flush_alerts` and only remove entries from the db that match that `klass` or something similar?

themylogin:
@yocalebo oneshot alert deletions, on average, do not happen more often than flush_alerts (which is performed every hour), hence, I don't see any reason that we need a more efficient `flush_alerts` implementation because of this change. However, I agree that we can have more efficient `flush_alerts` implementation in general, but can we do this in a separate ticket?

yocalebo:
Sure sounds good!

yocalebo (approved):

PR #9864, opened 2022-09-16
Add better net stats to collectd

yocalebo:
I'm investigating why `collectd` and `rrdcached` are using almost constant CPU time on a system with a large amount of datasets. As I was investigating, I noticed that the built-in `interface` plugin was lack-luster and furthermore we had not updated the collectd configuration so it was ignoring devices that could never exist on SCALE but also monitoring devices that we wanted to be ignored.

To remedy all of these problems, I've done the following:
1. remove the built-in `interface` module and replaced it with a python plugin called `ifstat`
2. this new plugin gives us quite a bit more data points to be mapped to each interface
3. the `ifstat` plugin now only monitors interfaces that we expect

3 new graphs are added
1. rx/tx packets
2. rx/tx errors
3. rx/tx dropped

1 graph updated
1. rx/tx bits per second

### DIFF
--- a/src/freenas/usr/local/lib/collectd_pyplugins/ifstat.py
+++ b/src/freenas/usr/local/lib/collectd_pyplugins/ifstat.py
@@ -1,0 +1,63 @@
+import traceback
+
+import collectd
+from pyroute2 import IPRoute
+
+from middlewared.client import Client
+
+READ_INTERVAL = 1.0
+
+
+class IfStat():
+    ignore = []
+
+    def config(self, config):
+        pass
+
+    def init(self):
+        collectd.info('Initializing "ifstat" python plugin')
+        try:
+            with Client() as c:
+                self.ignore = c.call('interface.internal_interfaces')
+        except Exception:
+            collectd.error(traceback.format_exc())
+
+    def valid_iface(self, dev):
+        ifname = dev.get_attr('IFLA_IFNAME')
+        if ifname is not None and ifname not in self.ignore:
+            return ifname
+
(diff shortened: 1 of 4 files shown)


themylogin:
Do we agree to drop all the existing stat in favor of the new method?

themylogin:
@yocalebo I understand, but this implementation writes to the other set of RRD filed, doesn't it? Meaning that updated systems will have empty network graphs on their first boot.

src/freenas/usr/local/lib/collectd_pyplugins/ifstat.py
```diff
@@ -0,0 +1,63 @@
+import traceback
+
+import collectd
+from pyroute2 import IPRoute
+
+from middlewared.client import Client
+
+READ_INTERVAL = 1.0
```
themylogin, line 8: Isn't every second too often? `interface` plugin files had a standard step of 10 seconds.

themylogin:
Is it backwards-compatible after your changes?
Also what was so lackluster about the built-in plugin? Is python implementation somehow less resource-consuming?

yocalebo:
> Is it backwards-compatible after your changes? Also what was so lackluster about the built-in plugin? Is python implementation somehow less resource-consuming?

It's backward compatible so webUI doesn't have to make any changes. We have no control over what the built-in plugin does. Furthermore, it only exports 4 categories of data. Mine exports 24 different stats per interface.

yocalebo:
meh, this isn't worth it

PR #9865, opened 2022-09-16
NAS-112452 / 22.12 / Hold/release snapshots

themylogin:


themylogin:
time 3:00

yocalebo (approved):
Looks good to me but let's for another set of eyes from @anodos325

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -1431,3 +1434,32 @@ def rollback_impl(self, args, id):
             )
         except subprocess.CalledProcessError as e:
             raise CallError(f'Failed to rollback snapshot: {e.stderr.strip()}')
+
+    @accepts(Str('id'))
+    def hold(self, id):
```
anodos325, line 1439: Should we expose recursive option for `zfs_hold()` and `zfs_release()`?

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -1431,3 +1434,44 @@ def rollback_impl(self, args, id):
...
+            'options',
+            Bool('recursive', default=False),
+        ),
+    )
+    def hold(self, id, options):
+        """
+        Holds snapshot `id`.
+
+        `truenas` tag will be added to the snapshot's tag namespace.
+        """
```
anodos325, line 1450: Document options?

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -1431,3 +1434,44 @@ def rollback_impl(self, args, id):
...
+        try:
+            with libzfs.ZFS() as zfs:
+                snapshot = zfs.get_snapshot(id)
+                snapshot.hold('truenas', options['recursive'])
+        except libzfs.ZFSException as err:
```
anodos325, line 1455: Since we're hardcoding tags, do we need special behavior for failure case where hold is already set?

themylogin: A special case that does what, no-op?

tests/api2/test_zfs_snapshot_hold.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def test_normal_snapshot():
+    with dataset("test") as ds:
+        with snapshot(ds, "test") as id:
+            assert call("zfs.snapshot.get_instance", id, {"extra": {"holds": True}})["holds"] == {}
+
+
+def test_held_snapshot():
+    with dataset("test") as ds:
```
anodos325, line 14: Maybe more descriptive dataset name for test. Less likely for collision if someone else uses `test` and doesn't properly clean up.

anodos325: Do we need test for recursive ops as well?

src/middlewared/middlewared/plugins/zfs.py
anodos325: It was merely a question. I was looking through in what situations it fails in `dsl_dataset_user_hold_check()`. If you don't think anything is required that's fine.

anodos325 (approved):

src/middlewared/middlewared/plugins/zfs.py
```diff
@@ -1431,3 +1434,49 @@ def rollback_impl(self, args, id):
...
+            raise CallError(f'Failed to hold snapshot: {err}')
+
+    @accepts(
+        Str('id'),
+        Dict(
+            'options',
+            Bool('recursive', default=False),
+        ),
+    )
+    def release(self, id, options):
```
sonicaj, line 1467: This is a public method, let's please add returns decorator as well

anodos325: Isn't ZFS plugin private?

tests/api2/test_zfs_snapshot_hold.py
themylogin: > if someone else uses test and doesn't properly clean up.

then they will be exposed and they will have to clean up properly

PR #9867, opened 2022-09-16
NAS-118211 / 22.12 / Fix inability to connect to iX servers

themylogin:


yocalebo (changes requested):
DM'ed but adding a comment so this doesn't get accidentally merged because this PR shouldn't be needed since I fixed the SSL cert issue a couple weeks ago.

PR #9870, opened 2022-09-17
NAS-117722 / 22.12 / Allow using legacy boot pool partition schema

themylogin:


themylogin:
time 5:00

src/middlewared/middlewared/plugins/boot_/boot_loader.py
```diff
@@ -10,9 +10,15 @@ class BootService(Service):
 
     @private
     async def install_loader(self, dev):
-        partition = await self.middleware.call('disk.get_partition_for_disk', dev, 2)
-
         await run('grub-install', '--target=i386-pc', f'/dev/{dev}')
+
+        if (
+            (await self.middleware.call('disk.list_partitions', dev))[1]['partition_type'] !=
```
yocalebo, line 16: Is it guaranteed that `disk.list_partitions, dev` will return an array with 2 items, minimally?

Also, why not do something like
```
parts = await self.middleware.call('disk.list_partitions', dev)
efi_uuid = await self.middleware.call('disk.get_efi_part_type')

if not parts:
    raise an error here?
elif any((i['partition_type'] == efi_uuid for i in parts)):
    return
```

yocalebo (changes requested):

src/middlewared/middlewared/plugins/boot_/boot_loader.py
themylogin: Yes, this is guaranteed, the disk is already partitioned at the moment, or something is horribly wrong

PR #9882, opened 2022-09-21
NAS-118271 / 22.12 / fix failover.become_passive

yocalebo:
`failover.become_passive` calls `service.restart keepalived` which breaks the other side of the FIFO connection that `middlewared` is blocking on waiting to receive failover events. When this happens, the other node receives a MASTER event (because `keepalived` on the other node migrated the VIP to it) and an event is generated. At the same time, however, the current MASTER node misses the failover event to become BACKUP because the other side of the FIFO connection was torn down.

So the other node receives the MASTER event, tries to start `fenced` but it fails to start because it detects the SCSI PR keys are actively changing on the disks which means the other node still has the zpool. So the failover process fails by design.

The _ONLY_ safe way to "become passive" is to violently reboot the current system. This ensures it gets out of the way _ENTIRELY_ so we avoid the potential situation of the zpool being imported on both nodes at the same time (which ultimately means data corruption).

### DIFF
--- a/src/middlewared/middlewared/plugins/failover.py
+++ b/src/middlewared/middlewared/plugins/failover.py
@@ -287,22 +287,41 @@ class FailoverService(ConfigService):
         return await self.middleware.call('system.general.get_ui_urls')
 
     @accepts()
-    @returns(Bool())
-    async def become_passive(self):
+    @returns()
+    def become_passive(self):
         """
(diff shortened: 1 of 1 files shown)


rick-mesta (approved):
LGTM, but as you're aware, I'm not the SME here 😇

anodos325 (approved):

PR #9885, opened 2022-09-21
NAS-118198 / 22.12 / Storj performance hints

themylogin:


themylogin:
time 0:10

yocalebo (changes requested):
Can we have a docstring stating something like "Storj recommended these settings" or something similar so we know why we're doing this.

yocalebo (approved):

yocalebo:
backport

PR #9886, opened 2022-09-21
NAS-111509 / 22.12 / Persist cloud sync and rsync task state into the database and across …

themylogin:
…HA nodes

PR #9892, opened 2022-09-22
VMware: retry snapshots deletion

themylogin:


themylogin:
time 14:00

yocalebo:
I'm just curious but how are testing this? Does this work on all versions of vsphere?

themylogin:
@yocalebo manually every time we change VMware API interaction (we have an internal vSphere installation for this). Here we don't make any changes, but I'll still test it with a built image.

yocalebo (approved):

themylogin:
We'll merge this after https://github.com/truenas/middleware/pull/9886

PR #9895, opened 2022-09-22
Added code to look 5 times to disks(pool["topology"]) in pool_replace_disk

ericbsd:
When the VM is slow, it fails randomly, so this should take care of that issue.


q5sys (approved):
LGTM

PR #9902, opened 2022-09-23
Revert "set CPU governor to performance mode always"

yocalebo:
This reverts commit 058034a092b8d1d5df55a49c2f0e65dba763e218.

Performance team sees performance improvements almost everywhere....except systems running a different brand of CPU 🙄 We'll work with them to characterize certain aspects and turn this back on in a controlled manner.


anodos325 (approved):

Ornias1993:
On another note:
There are a lot of users that want to run lower-power systems.
With the move to Linux with SCALE, low-power systems gained a huge bump in feasability.

Hardcoding things like this, interferes with people wanting to lower power consumption.
Suggestion: Make it at least toggleale somewhat at next attempt.

PR #9907, opened 2022-09-25
License Apps and VMs for enterprise

sonicaj:


yocalebo (approved):

yocalebo (changes requested):
Actually, this should be fairly easy to add a test or two to our integration test suite. We setup HA VMs so you should be able to check. Or if you prefer unit tests, that's fine too. You can choose

yocalebo (approved):
awesome, thanks!

PR #9909, opened 2022-09-25
Add an endpoint to determine if gluster cluster path is valid

sonicaj:
This commit adds a private endpoint which will be consumed by docker to validate if a cluster specified path is valid or not. Idea is that users can specify any cluster/host path and if it turns out to be a cluster path, docker will call middleware to validate the path itself and if it turns out to be valid, will let the app deploy or otherwise will fail it at it's CSI checks like we do for other cases currently.

It would have been nice however to have an endpoint to expose gluster volumes somehow ( the issue is not that we can't query gluster volumes right now ) but rather how user usually selects paths in path fields all across TN UI. The UI should probably be adaptable enough to intelligently handle `/cluster` cases.


src/middlewared/middlewared/plugins/chart_releases_linux/storage.py
```diff
@@ -0,0 +1,25 @@
...
+class ChartReleaseService(Service):
+
+    class Config:
+        namespace = 'chart.release'
+
+    @private
+    async def validate_cluster_path(self, path):
+        # Will return None if no error is found, otherwise an error string will be returned
+        if not path.startswith(f'{FuseConfig.FUSE_PATH_BASE.value}/'):
+            return f'Specified path does not start with valid path base {FuseConfig.FUSE_PATH_BASE.value!r}'
```
yocalebo, line 14:
```suggestion
            return f'Path must start with {FuseConfig.FUSE_PATH_BASE.value!r}'
```

yocalebo (changes requested):
Let's be less "wordy" when it doesn't start with `FUSE_PATH_BASE`.

yocalebo (approved):

PR #9916, opened 2022-09-27
NAS-118351 / 22.12 / Disable middleware debug mode being the default

themylogin:
No one seems to be interested to keep our return value schemas up-to-date, so middleware started in debug mode is unusable. Let's start it normally by default, we'll fix this in the future.


yocalebo (approved):

themylogin:
time 0:10

PR #9919, opened 2022-09-27
Allow adding command line args for vms

Qubad786:
## Problem

There are some features which are not exposed by libvirt but are present in QEMU and hinder user functionality. For example a user has a Tesla M40 GPU and it requires specific QEMU args which are not exposed in libvirt and hence he can't do GPU passthrough.

## Solution

Allow users to add command line arguments for VMs to handle such edge cases.


src/middlewared/middlewared/alembic/versions/22.12/2022-09-21_14-35_add_command_line_args.py
```diff
@@ -0,0 +1,31 @@
...
+revision = '7a74a8933a30'
+down_revision = 'aef1177c39c5'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('vm_vm', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('command_line_args', sa.TEXT(), nullable=False))
```
themylogin, line 22: Adding NOT NULL column without a server_default will fail if there already are rows in the table

src/middlewared/middlewared/plugins/vm/vms.py
```diff
@@ -51,6 +51,7 @@ class VMModel(sa.Model):
     arch_type = sa.Column(sa.String(255), default=None, nullable=True)
     machine_type = sa.Column(sa.String(255), default=None, nullable=True)
     uuid = sa.Column(sa.String(255))
+    command_line_args = sa.Column(sa.JSON(), default=[], nullable=False)
```
themylogin, line 54: Why is it JSON, aren't arguments just a string? You can use `shlex.split` to split them into separate arguments then.

themylogin (changes requested):

rick-mesta:
@themylogin : has @Qubad786 addressed all of your concerns ?

themylogin (approved):
Thank you, LGTM!

anotb:
This change seems to have got lost? I do not see it in the most recent Scale builds.

anotb:
Reopened in JIRA https://ixsystems.atlassian.net/browse/NAS-119992

PR #9920, opened 2022-09-27
Set aclinherit and aclmode to sane values on acltype change

anodos325:
There are limited combinations of these parameters that are valid. We should try to auto-configure something sane when users are switching acltype.


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -3509,6 +3509,22 @@ async def do_update(self, id, data):
         if data.get('user_properties_update'):
             props.update(await self.get_create_update_user_props(data['user_properties_update'], True))
 
+        if 'acltype' in props:
```
yocalebo, line 3512: Just a nit but instead of calling `props['acltype'].get('value')` in 3 separate branches you could do

`if 'acltype' in props and (acltype_value := props['acltype'].get('value')):
    if acltype_value == 'nfsv4':
        ....

yocalebo (approved):

PR #9932, opened 2022-10-03
NAS-117316 / 22.12 / Add ability to validate ports in middleware services

sonicaj:


src/middlewared/middlewared/plugins/nfs_/port_attachments.py
```diff
@@ -0,0 +1,13 @@
+from middlewared.common.ports import ServicePortDelegate
+
+
+class NFSServicePortDelegate(ServicePortDelegate):
+
+    name = 'nfs'
+    namespace = 'nfs'
+    port_fields = ['mountd_port', 'rpcstatd_port', 'rpclockd_port']
+    title = 'NFS Service'
```
anodos325, line 9: What about port 2049?

src/middlewared/middlewared/plugins/ports/ports.py
```diff
@@ -0,0 +1,38 @@
+import copy
+
+from middlewared.service import Service, ValidationErrors
+
+
+class PortService(Service):
+
+    DELEGATES = {}
+    SYSTEM_USED_PORTS = [
+        {'title': 'System', 'ports': [6000], 'namespace': 'system'},
```
anodos325, line 10: We should probably disallow 67, 123, 3702, and 5353.

src/middlewared/middlewared/plugins/ports/ports.py
```diff
@@ -0,0 +1,38 @@
...
+    ]
+
+    class Config:
+        private = True
+
+    async def register_attachment_delegate(self, delegate):
+        self.DELEGATES[delegate.namespace] = delegate
+
+    async def get_in_use(self):
+        ports = copy.deepcopy(self.SYSTEM_USED_PORTS)
```
yocalebo, line 20: Why `deepcopy`? Just iterate inside this function and create a new list and then simply `ports += self.SYSTEM_USED_PORTS`

yocalebo (changes requested):

src/middlewared/middlewared/plugins/ports/ports.py
```diff
@@ -0,0 +1,36 @@
...
+    ]
+
+    class Config:
+        private = True
+
+    async def register_attachment_delegate(self, delegate):
+        self.DELEGATES[delegate.namespace] = delegate
+
+    async def get_in_use(self):
+        ports = self.SYSTEM_USED_PORTS.copy()
```
yocalebo, line 18: There is no need for a shallow copy either. Here's what I was thinking about doing

```
ports = []
for delegate in self.DELEGATES.values():
    if used_ports := await delegate.get_ports():
        ports.append({
            'namespace': delegate.namespace,
            'title': delegate.title,
            'ports': used_ports,
        })

ports += self.SYSTEM_USED_PORTS
return ports

**Review by yocalebo — CHANGES_REQUESTED** [review]

File: src/middlewared/middlewared/plugins/vpn.py [hunk 5]
```diff
@@ -506,7 +510,7 @@ class OpenVPNClientModel(sa.Model):
     __tablename__ = 'services_openvpnclient'
 
     id = sa.Column(sa.Integer(), primary_key=True)
-    port = sa.Column(sa.Integer(), default=1194)
+    port = sa.Column(sa.Integer(), default=1195)
```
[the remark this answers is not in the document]
sonicaj: Motivation behind is to not have 2 services use the same port by default.

yocalebo (approved):
I see no immediate problems, but I'd like 1 other middleware team member to review it before we merge.

src/middlewared/middlewared/plugins/ports/ports.py
```diff
@@ -0,0 +1,36 @@
...
+    DELEGATES = {}
+    SYSTEM_USED_PORTS = [
+        {'title': 'System', 'ports': [67, 123, 3702, 5353, 6000], 'namespace': 'system'},
+    ]
+
+    class Config:
+        private = True
+
+    async def register_attachment_delegate(self, delegate):
+        self.DELEGATES[delegate.namespace] = delegate
```
themylogin, line 15: Should we raise an error if a delegate is already registered (key exists)?

src/middlewared/middlewared/plugins/chart_releases_linux/attachments.py
```diff
@@ -0,0 +1,20 @@
...
+
+    name = 'applications'
+    namespace = 'chart.release'
+    title = 'Applications'
+
+    async def get_ports(self):
+        ports = []
+        for chart_release in await self.middleware.call('chart.release.query'):
+            for port in chart_release['used_ports']:
+                ports.append(port['port'])
```
themylogin, line 14: What if we change return schema to `{8080: "Application 1", 8081: "Application 2"}`? That way we'll be able to report more specific error message (`This port is used by Application 1` instead of `This port is used by Applications`).

themylogin (approved):

PR #9935, opened 2022-10-03
NAS-118429 / 22.12 / properly wait on job in dir services

yocalebo:
`wait` is a coroutine so can't be called from a synchronous method. Use `wait_sync` instead.


anodos325 (approved):

yocalebo:
backport

PR #9937, opened 2022-10-04
NAS-118444 / 22.12 / add MISMATCH_VERSIONS to failover.disabled.reasons

yocalebo:
We already raise an alert in the webUI but we don't in `failover.disabled.reasons` which doesn't make any logical sense.


anodos325 (approved):

yocalebo:
Waiting on associated webUI pr here: https://github.com/truenas/webui/pull/7165

PR #9954, opened 2022-10-07
NAS-118328 / 22.12 / Do not encrypt ix-applications dataset 

sonicaj:
This PR adds changes to not allow encrypting ix-applications dataset even if parent is encrypted and also not allow migrating ix-applications dataset to another pool if source is encrypted as this results in various edge cases which after discussing with Caleb we have decided not to handle.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
@@ -122,26 +122,28 @@ class KubernetesService(ConfigService):
                         f'possible as {applications_ds_name(data["pool"])} already exists.'
                     )
 
-                if not await self.middleware.call(
+                ix_apps_ds = await self.middleware.call(
                     'zfs.dataset.query', [['id', '=', applications_ds_name(old_data['pool'])]], {
-                        'extra': {'retrieve_children': False, 'retrieve_properties': False}
+                        'extra': {'retrieve_children': False, 'retrieve_properties': True}
                     }
-                ):
+                )
+                if not ix_apps_ds:
                     # Edge case but handled just to be sure
                     verrors.add(
                         f'{schema}.migrate_applications',
                         f'{applications_ds_name(old_data["pool"])!r} does not exist, migration not possible.'
                     )
-                if not verrors and (
-                    await self.middleware.call(
+                if not verrors:
+                    source_root_ds = await self.middleware.call(
                         'pool.dataset.get_instance', old_data['pool'], {'extra': {'retrieve_children': False}}
                     )
-                )['encrypted']:
-                    verrors.add(
-                        f'{schema}.migrate_applications',
-                        f'Source {old_data["pool"]!r} is encrypted and it is not supported '
-                        'migrating encrypted applications data'
-                    )
+                    ix_apps_ds = ix_apps_ds[0]
+                    if source_root_ds['encrypted'] and ix_apps_ds['encrypted']:
+                        verrors.add(
+                            f'{schema}.migrate_applications',
+                            f'Source {applications_ds_name(old_data["pool"])!r} is encrypted and it is not supported '
+                            'migrating encrypted applications data'
+                        )
 
         network_cidrs = set([
             ipaddress.ip_network(f'{ip_config["address"]}/{ip_config["netmask"]}', False)
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
@@ -132,6 +132,16 @@ class KubernetesService(ConfigService):
                         f'{schema}.migrate_applications',
                         f'{applications_ds_name(old_data["pool"])!r} does not exist, migration not possible.'
                     )
+                if not verrors and (
+                    await self.middleware.call(
+                        'pool.dataset.get_instance', old_data['pool'], {'extra': {'retrieve_children': False}}
+                    )
+                )['encrypted']:
+                    verrors.add(
+                        f'{schema}.migrate_applications',
+                        f'Source {old_data["pool"]!r} is encrypted and it is not supported '
+                        'migrating encrypted applications data'
+                    )
 
         network_cidrs = set([
             ipaddress.ip_network(f'{ip_config["address"]}/{ip_config["netmask"]}', False)
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
@@ -272,7 +272,9 @@ class KubernetesService(Service):
     async def create_update_k8s_datasets(self, k8s_ds):
         create_props_default = self.k8s_props_default()
         for dataset_name in await self.kubernetes_datasets(k8s_ds):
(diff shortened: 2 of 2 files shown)


exander77:
@sonicaj That's not really solution I envisioned when I reported this issue. :D

How do I move my encrypted ix-application dataset then?

Ornias1993:
I can confirm that replication of ix-applications between encrypted datasets (more specifically: with an encrypted source), is wonky at times. This has been the case since Alpha.

However, I think we need to take current users into account. "cannot move apps, sorry" is not really an acceptable solution for current users.

exander77:
I am not sure if this is related, but after reboot today, I have some issue on the dataset I tried migrated `ix-applications` to.

image

https://www.truenas.com/community/threads/some-datasets-locked-for-no-apparent-reason.104495/

Some of my child datasets on that device are locked, but I can't unlock them.

Ornias1993:
> I am not sure if this is related, but after reboot today, I have some issue on the dataset I tried migrated `ix-applications` to.
> 
> image
> 
> https://www.truenas.com/community/threads/some-datasets-locked-for-no-apparent-reason.104495/
> 
> Some of my child datasets on that device are locked, but I can't unlock them.

Try relocking the parent and unlocking it again.

to be fair: it's issues like this i've seen more with bad/wrong replication options when (trying to) replicate ix-applications.

exander77:
> > I am not sure if this is related, but after reboot today, I have some issue on the dataset I tried migrated `ix-applications` to.
> > image
> > https://www.truenas.com/community/threads/some-datasets-locked-for-no-apparent-reason.104495/
> > Some of my child datasets on that device are locked, but I can't unlock them.
> 
> Try relocking the parent and unlocking it again.
> 
> to be fair: it's issues like this i've seen more with bad/wrong replication options when (trying to) replicate ix-applications.

Weird thing is that, I haven't touched that dataset at all. I only worked with the ix-applications one. Could the parent somehow be corrupted after this? I tried rebooting already. And when I create a new dataset, it is affected the same way. It may be completely unreated issue.

Btw, can anybody suggest how to unencrypted my ix-application dataset? I can move it around etc.

Ornias1993:
> > > I am not sure if this is related, but after reboot today, I have some issue on the dataset I tried migrated `ix-applications` to.
> > > image
> > > https://www.truenas.com/community/threads/some-datasets-locked-for-no-apparent-reason.104495/
> > > Some of my child datasets on that device are locked, but I can't unlock them.
> > 
> > 
> > Try relocking the parent and unlocking it again.
> > to be fair: it's issues like this i've seen more with bad/wrong replication options when (trying to) replicate ix-applications.
> 
> Weird thing is that, I haven't touched that dataset at all. I only worked with the ix-applications one. Could the parent somehow be corrupted after this? I tried rebooting already. And when I create a new dataset, it is affected the same way. It may be completely unreated issue.
> 
> Btw, can anybody suggest how to unencrypted my ix-application dataset? I can move it around etc.

To be clear:
This github issue is created because iX-Systems @sonicaj  has confirmed you should not be trying to move the ix-applications dataset when either the source or target pool/dataset is encrypted.

So it isn't weird, it's confirmed to be "not working as expected".
Please do not hijack this PR by trying to ask for support, as PR's are not intended as support chats.

src/middlewared/middlewared/plugins/kubernetes_linux/update.py
```diff
@@ -132,6 +132,16 @@ async def validate_data(self, data, schema, old_data):
...
                     )
+                if not verrors and (
+                    await self.middleware.call(
+                        'pool.dataset.get_instance', old_data['pool'], {'extra': {'retrieve_children': False}}
+                    )
+                )['encrypted']:
+                    verrors.add(
+                        f'{schema}.migrate_applications',
+                        f'Source {old_data["pool"]!r} is encrypted and it is not supported '
+                        'migrating encrypted applications data'
```
Ornias1993, line 143: Instead we can try to replicate it to the new dataset without replicating any of the dataset attributes. In my experience this scrubbed away all potential issues with encryption.

However, afterwards we do need to execute a flow to 
- have the PVC's and some other datasets set to `legacy` mountpath.
- Scrub unwanted parent dataset options comming from the target parent dataset

Ornias1993:
> I am aware, but my issue is most likely direct result of trying to move my `ix-applications`.
> 
> I repaired the "locked" dataset with unloading the key from the root and then unlocking it from my backup key file, then `zfs mount`. It solved it persistently and after reboot the dataset is again mounted as it should be. And it is an important info for everybody that may have a similar issue.
> 
> But I still have issue with my main datastore:
> 
> ```
> root@silverrock[/mnt/secure2]# ls -la
> total 35
> drwxr-xr-x 5 root root      5 Sep 26 16:01 .
> drwxr-xr-x 6 root root      6 Oct 10 06:04 ..
> drwxr-xr-x 6 root root      6 Oct 10 13:47 charts
> drwxr-xr-x 2 root root      2 Sep 26 16:01 ix-applications
> drwxrwx--- 8 root www-data 13 Oct 10 01:57 nextcloud2
> root@silverrock[/mnt/secure2]# rm -rf ix-applications 
> rm: cannot remove 'ix-applications': Operation not permitted
> root@silverrock[/mnt/secure2]# touch test
> touch: setting times of 'test': No such file or directory
> root@silverrock[/mnt/secure2]# 
> ```
> 
> I cannot write any files to the root of my datastore and I have issues creating new datastores: image

This PR is not for fixing your issue, it intends to block new users from trying what you did.
Even when merged, this is not going to solve your issue.

However your frequent comment do push my code related comment into being less vissible, which interferes with development.
Please move your help calls to the forums or your Jira ticket.

sonicaj:
To clarify, we will be adding some validation right now to not allow migrating encrypted ix-applications but in another PR conditionally allow migrating encrypted ix-applications dataset.

Ornias1993:
> To clarify, we will be adding some validation right now to not allow migrating encrypted ix-applications but in another PR conditionally allow migrating encrypted ix-applications dataset.

Awesome, that solves both cases!

exander77:
> > To clarify, we will be adding some validation right now to not allow migrating encrypted ix-applications but in another PR conditionally allow migrating encrypted ix-applications dataset.
> 
> Awesome, that solves both cases!

Yes, I wholeheartedly support any improvement in this regard. Btw my suggestion would be to create `ix-applications` dataset unencrypted (unless requested) even on encrypted pools, it would save me a lot of headache.

exander77:
I have created related issue in openzfs as this is most likely cause of my broken dataset:
https://github.com/openzfs/zfs/issues/14011

Ornias1993:
> I have created related issue in openzfs as this is most likely cause of my broken dataset: [openzfs/zfs#14011](https://github.com/openzfs/zfs/issues/14011)

Wrong place, It needs to be reported on the iX Systems Jira.

Ornias1993:
> Yes, I wholeheartedly support any improvement in this regard. Btw my suggestion would be to create `ix-applications` dataset unencrypted (unless requested) even on encrypted pools, it would save me a lot of headache.

This PR already creates it unencrypted.
Please, if you've no interest on commenting on the code ignore PR's (because if you read it, you would've known this was already part of this PR)

yocalebo (approved):

PR #9957, opened 2022-10-10
NAS-118480 / 22.12 / Configure kuberouter logs destination in sylog-ng cfg file

Qubad786:
## Context

We want `kube-router`to log errors to a separate file so we update syslog-ng configuration to filter out `kube-router` program logs to separate files to not spam other files.


sonicaj (commented):
This LGTM, however let's please add changes to `src/freenas/etc/logrotate.d/syslog-ng-truenas` as well to not keep older logs of kube-router and keep a limit on it's size as well like we do for k3s/containerd. Thanks!

sonicaj:
retest this please

PR #9959, opened 2022-10-11
NAS-118493 / 22.12 / Add integration test for testing port delegates

sonicaj:


tests/api2/test_port_delegates.py
```diff
@@ -0,0 +1,59 @@
...
+
+def test_port_delegate_validation_with_invalid_ports():
+    in_use_ports = []
+    for entry in call('port.get_in_use'):
+        in_use_ports.extend(filter(lambda i: i > 1024, entry['ports']))
+
+    for config_method, method, keys, payload in PAYLOAD:
+        validation_error = None
+        for key in keys:
+            payload[key] = in_use_ports[random.randint(0, len(in_use_ports) - 1)]
```
themylogin, line 28: Why do you use `random`? It's a recipe to non-repeatable failures or false negatives

themylogin (changes requested):

tests/api2/test_port_delegates.py
sonicaj: We should be choosing any port which is already being consumed so i wanted to randomly select any element in the list. Should we remove it instead ? ( btw wrt this usage i think we should be safe )

tests/api2/test_port_delegates.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def test_port_delegate_validation_with_invalid_ports():
+    in_use_ports = []
+    for entry in call('port.get_in_use'):
+        in_use_ports.extend(filter(lambda i: i > 1024, entry['ports']))
+
+    assert in_use_ports != [], 'No in use ports retrieved'
+
+    for config_method, method, keys, payload in PAYLOAD:
```
themylogin, line 28: It's better to use `pytest.mark.parametrize` and try each payload in a separate test. If you use `for` loop, one payload failure will result to subsequent not being tested.

themylogin (changes requested):

tests/api2/test_port_delegates.py
```diff
@@ -0,0 +1,68 @@
...
+
+    validation_error = None
+    for index, key in enumerate(keys):
+        payload[key] = in_use_ports[index] if len(in_use_ports) > index else in_use_ports[0]
+    try:
+        call(method, payload)
+    except ValidationErrors as ve:
+        validation_error = ve
+
+    assert validation_error is not None, 'Port validation exception expected'
```
themylogin, line 38: Sorry, one more thing. To ensure that an exception is raised use this:
```
    with pytest.raises(ValidationErrors) as ve:
        call("privilege.update", 1, {"allowlist": [{"method": "CALL", "resource": "system.info"}]})

    assert ve.value.errors[0].attribute == "privilege_update.allowlist"
```

themylogin (approved):
LGTM, thank you very much!

PR #9963, opened 2022-10-11
NAS-118513 / 22.12 / Do not configure swap or create swap devices on HA hardware

sonicaj:


yocalebo (changes requested):
Looks simple enough, however, can we add an integration test? Since our integration tests run against a HA system, we should be able to ensure that no swap devices exist after creating a zpool.

yocalebo (approved):

PR #9965, opened 2022-10-12
NAS-118291 / 22.12 / Replication: Only `Same as Source` and `None` retention policies can …

themylogin:
…be used with Naming regex


yocalebo (approved):

themylogin:
time 0:30

PR #9968, opened 2022-10-12
NAS-117976 / 13.0 / Fix NIS for HA

anodos325:
NIS start / stop wasn't being propagated to remote node.

### DIFF
--- a/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
+++ b/src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
@@ -616,6 +616,10 @@ class FailoverService(Service):
                 # Now we restart the appropriate services to ensure it's using correct certs.
                 self.run_call('service.restart', 'http')
 
+                if self.run_call('nis.config')['enable']:
+                    if not self.run_call('nis.started'):
+                        self.run_call('service.restart', 'nis', {'ha_propagate': False})
+
                 # restart the critical services first
                 # each service is restarted concurrently and given a timeout value of 15
                 # seconds to restart. This is done to prevent the possibility of a service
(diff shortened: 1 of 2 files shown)


rick-mesta (commented):
LGTM

yocalebo (changes requested):
DM'ed anodos325 for investigating doing this another way so we follow the same "paradigm" that we follow on HA systems.

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
```diff
@@ -850,6 +854,10 @@ def carp_backup(self, fobj, ifname, vhid, event, user_override):
 
                 self.run_call('etc.generate', 'cron')
 
+                if self.run_call('nis.config')['enable']:
+                    if not self.run_call('nis.started'):
+                        self.run_call('service.restart', 'nis')
```
yocalebo, line 859: 1. do we really need to restart nis when becoming the standby node?
2. if you insist, then we need to add `{'ha_propagate': false}` to this or it will send this request to the active controller

yocalebo (changes requested):

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
```diff
@@ -616,6 +616,10 @@ def carp_master(self, fobj, ifname, vhid, event, user_override, forcetakeover):
                 # Now we restart the appropriate services to ensure it's using correct certs.
                 self.run_call('service.restart', 'http')
 
+                if self.run_call('nis.config')['enable']:
+                    if not self.run_call('nis.started'):
+                        self.run_call('service.restart', 'nis')
```
yocalebo, line 621: Same as well up here, add `{'ha_propagate': False}`

yocalebo (approved):
Great, thanks!

PR #9969, opened 2022-10-12
NAS-118563 / 13.0 / fix race with db insert/service.restart on HA

yocalebo:


src/middlewared/middlewared/plugins/network.py
```diff
@@ -173,6 +173,60 @@ async def toggle_announcement(self, data):
...
             await self.middleware.call(f'service.{verb}', service_name)
 
+    @private
+    async def propagate_changes_to_remote(self, remote_actions, local_config, licensed):
+        """
+        On the internal HA VM used for integration tests, we're seeing that propagating db changes
+        that have been written to the active node are taking up to 4 seconds to propagate to the
+        standby node and be flushed to disk. This means we'll restart the dns "service" on the remote
+        node before the db changes have been flushed which has the perception that the db changes aren't
+        actually getting propagating...but they are we just have a race...because of an unfortunate design.
```
rick-mesta: `s/propagating/propagated` ?

rick-mesta (approved):

src/middlewared/middlewared/plugins/network.py
yocalebo: heh, I actually fixed that in my 2nd commit so I'm assuming looking at cached version 😄

rick-mesta: Hah... my bad 🤪

anodos325 (approved):

PR #9973, opened 2022-10-12
NAS-118567 / 22.12 / Allow migrating applications when encrypted

sonicaj:
This PR adds changes to allow migration of encrypted ix-applications dataset. However it enforces that destination pool's root dataset must be KEY encrypted and when the migration completes, ix-applications dataset would be inheriting encryption properties of destination's root dataset.


yocalebo (approved):
Please open another ticket so we can add an integration test for this. Doesn't necessarily need to test migrating encrypted ix-applications but we should test migrating in general. If we already are testing this functionality, then ignore me!

Qubad786 (approved):

exander77:
Does this work? Tried to migrate my `ix-applications` to a new pool, it just hangs. Should I turn k3s off first?

sonicaj:
@exander77 can you please create an issue at our jira and attach a debug of your system ?

In certain versions issue existed with migrating encrypted applications datasets where the migration was broken itself but that should not be the case now. Thanks!

PR #9976, opened 2022-10-12
NAS-118572 / 13.0 / Fix NIS user/group cache

anodos325:
The format for directory services caches changed a while back, but NIS plugin cache was not updated concurrently (possible backport from SCALE where NIS doesn't exist).


rick-mesta (approved):
🚢 it !

PR #9977, opened 2022-10-13
NAS-117406 / 22.12 / Do not rename network interfaces based on their stored MAC address on…

themylogin:
… HA hardware


themylogin:
time 0:15

PR #9978, opened 2022-10-13
NAS-114520 / 22.12 / Do not use `serial_lunid`, `serial` and `uuid` for disk identification if they are not unique

themylogin:


themylogin:
time 1:00

yocalebo (changes requested):
This adds 3 iterations to `sys_disks.values` and also adds a 4th iteration where we iterate every partition on every device in `sys_disks.values`.

Then we add another nested iteration of every partition for every device while the outside loop is iterating all partitions for the found device.

Surely, there has to be a more performant way of doing this

yocalebo:
Actually, this ticket was opened by team member that was explicitly creating disks in their VM with duplicate serials. Considering we've not had this problem for many years on real hardware, I don't want to fix this. If someone has disks with duplicate serials or duplicate uuids, there isn't much we can fix. There are more significant problems with that system.

yocalebo:
Please see comment I made in ticket.

PR #9980, opened 2022-10-13
NAS-118421 / 22.12 / Add validation to ensure server/client certs are specified for openvpn services

sonicaj:
This PR adds changes to force user to specify server/client certs for openvpn services. What happens is that before starting openvpn client/server service, we do check if these are present or not and if not we do not render the configuration file for the service which results in the service not starting, however this does not tell the user what he did wrong.

Why we did not validate this before was to allow users to unset certs in case they wanted to delete them and now for that case we have introduced a new variable which will handle this.

There is also another change, we allow users now to not specify client certificate as with some additional parameters which openvpn supports, it is possible to start openvpn with user credentials which some popular VPN providers like proton VPN provide.

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/openvpn/client/openvpn_client.conf.mako
+++ b/src/middlewared/middlewared/etc_files/local/openvpn/client/openvpn_client.conf.mako
@@ -6,7 +6,10 @@
 
 	config = middleware.call_sync('openvpn.client.config')
 	root_ca = middleware.call_sync('certificateauthority.query', [['id', '=', config['root_ca']]], {'get': True})
-	client_cert = middleware.call_sync('certificate.query', [['id', '=', config['client_certificate']]], {'get': True})
+	if config['client_certificate']:
+		client_cert = middleware.call_sync('certificate.query', [['id', '=', config['client_certificate']]], {'get': True})
+	else:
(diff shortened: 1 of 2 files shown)


sonicaj:
There is an actual case where client certificate is not required - going to investigate that some more before moving on with these changes

src/middlewared/middlewared/etc_files/local/openvpn/client/openvpn_client.conf.mako
```diff
@@ -6,7 +6,10 @@
...
 	config = middleware.call_sync('openvpn.client.config')
 	root_ca = middleware.call_sync('certificateauthority.query', [['id', '=', config['root_ca']]], {'get': True})
-	client_cert = middleware.call_sync('certificate.query', [['id', '=', config['client_certificate']]], {'get': True})
+	if config['client_certificate']:
+		client_cert = middleware.call_sync('certificate.query', [['id', '=', config['client_certificate']]], {'get': True})
+	else:
+		client_cert = None
 %>\
 client
 % if IS_LINUX:
```
yocalebo, line 15: Since you're here, why not remove the IS_LINUX/IS_FREEBSD checks?

yocalebo (approved):

PR #9981, opened 2022-10-13
NAS-118068 / 13.0 / add R50BM enclosure mgmt logic to CORE

yocalebo:
Tested on internal R50BM. To note, this platform uses same plx bridge (different OEM) as the m-series devices.


rick-mesta (approved):
LGTM

PR #9991, opened 2022-10-14
NAS-118583 / 22.12 / Add API for setting clock

anodos325:
This API endpoint is to be used to set the RTC clock to a client-provided unix timestamp. ntpd will be temporarily stopped while making clock adjustment.

### DIFF
--- a/tests/api2/test_470_system.py
+++ b/tests/api2/test_470_system.py
@@ -9,8 +9,8 @@ import sys
 import os
 apifolder = os.getcwd()
 sys.path.append(apifolder)
-from functions import GET  # , POST
-from auto_config import dev_test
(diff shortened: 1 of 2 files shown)


yocalebo (commented):
Integration test? 😄

tests/api2/test_470_system.py
```diff
@@ -43,3 +43,33 @@ def test_04_check_system_product_type():
 def test_05_check_system_debug():
     results = GET("/system/debug/")
     assert results.status_code == 200, results.text
+
+
+def test_06_cehck_system_set_time():
```
yocalebo, line 48: typo

yocalebo (approved):

PR #9995, opened 2022-10-17
Fixed cannot import name 'another_pool' with api2.test_swap_creation_on_ha

ericbsd:


sonicaj:
@ericbsd https://github.com/truenas/middleware/pull/9993

PR #9998, opened 2022-10-18
NAS-118301 / 22.12 / Add support for having kubernetes infrastructure migrations

sonicaj:


src/middlewared/middlewared/plugins/kubernetes_linux/infra_migrations.py
```diff
@@ -0,0 +1,83 @@
...
+        return os.path.join(
+            '/mnt', self.middleware.call_sync('kubernetes.config')['dataset'], self.MIGRATIONS_FILE_NAME
+        )
+
+    def applied(self):
+        try:
+            with open(self.migration_file_path, 'r') as f:
+                return json.loads(f.read())
+        except FileNotFoundError:
+            self.logger.error('%r migration file not found, creating one', self.migration_file_path)
```
themylogin, line 35: Isn't it normal that the file does not exist?

src/middlewared/middlewared/plugins/kubernetes_linux/infra_migrations.py
```diff
@@ -0,0 +1,83 @@
...
+        except json.JSONDecodeError:
+            self.logger.error('Malformed %r migration file found, re-creating', self.migration_file_path)
```
themylogin, line 37: The file might contain a valid JSON e.g. `{}` but it does not have the format you expect and it will cause unexpected error later.

themylogin (approved):

src/middlewared/middlewared/plugins/kubernetes_linux/infra_migrations.py
sonicaj: It won't for users who are going to migrate to the newer version of scale

src/middlewared/middlewared/plugins/kubernetes_linux/infra_migrations.py
sonicaj: I'll add some json validation on that end, thank you for pointing it out

src/middlewared/middlewared/plugins/kubernetes_linux/infra_migrations.py
themylogin: But it's normal for those who already were using apps on previous versions of SCALE

sonicaj: Nope it won't as migrations for k8s did not exist at that point. Ah i see, you are pointing out that we should not log this as an error in that case ?

That's tricky as it's possible say users upgraded to newer scale version which has migrations but then they select an apps pool which was used on previous version of scale, now that won't contain the migration file. Should we maybe log this as `warning` instead ?

themylogin: > users upgraded to newer scale version which has migrations but then they select an apps pool which was used on previous version of scale

That's not even a warning, that's a normal situation which will happen to everyone upgrading to 22.12. I don't think we should log everything at all; we'll see `info` messages executing every existing migration below.

PR #10009, opened 2022-10-19
Make sure we don't start TC on standby node

sonicaj:


rick-mesta:
Also, please remember that the Release Engineering team is still working to test `13.0-U3`, so unless this **_has_** to go into `U3`, please hold off.

sonicaj:
@rick-mesta just confirmed with William that yes we would definitely like this in U3. I have hot patched the test systems and we will be monitoring them for a couple of hours or so to see if i missed some case potentially.

rick-mesta:
carry on 👍🏼

PR #10012, opened 2022-10-24
Do not error out if route already exists in kube-router table

Qubad786:
## Problem

There can be cases where CNI already has added relevant routes to `kube-router` table and trying to add the route again will error out stopping the bootstrapping process of kubernetes cluster resulting in apps failing to deploy.

## Solution

Make adding routes to kube-router table non-fatal if they already exist.


yocalebo (approved):

src/middlewared/middlewared/plugins/kubernetes_linux/cni.py
```diff
@@ -113,10 +114,14 @@ def add_user_route_to_kube_router_table(self):
...
-                factory(0), factory(0), ipaddress.ip_address(config[f'{k}_gateway']), config[f'{k}_interface'],
-                table_id=kube_router_table.table_id,
-            ))
+            try:
+                rt.add(netif.Route(
+                    factory(0), factory(0), ipaddress.ip_address(config[f'{k}_gateway']), config[f'{k}_interface'],
+                    table_id=kube_router_table.table_id,
+                ))
+            except pyroute2.netlink.exceptions.NetlinkError as e:
+                if e.code != 17:
```
yocalebo, line 123:
```suggestion
                # ignore error if the route already exists in the routing table.
                if e.code != 17:
```

yocalebo (changes requested):
Add a small comment to this exception so future developers know what's going on easily.

yocalebo (approved):
thanks!

PR #10014, opened 2022-10-24
NAS-118702 / 22.12 / Map fake NVMe enclosures to actual block devices so that `disk.query`…

themylogin:
… can match enclosures

### DIFF
--- a/src/middlewared/middlewared/plugins/enclosure_/nvme.py
+++ b/src/middlewared/middlewared/plugins/enclosure_/nvme.py
@@ -7,7 +7,6 @@ from middlewared.service import Service, private
 
 
 class EnclosureService(Service):
-    RE_NVME = re.compile(r"^nvme[0-9]+$")
     RE_SLOT = re.compile(r"^0-([0-9]+)$")
 
     @private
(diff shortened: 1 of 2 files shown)


themylogin:
time 6:30

src/middlewared/middlewared/plugins/enclosure_/nvme.py
```diff
@@ -75,10 +74,10 @@ def map_plx_nvme_impl(self, prod):
                 pass
             else:
                 for child in physical_node.children:
-                    if not self.RE_NVME.fullmatch(child.sys_name):
+                    if child.properties.get("SUBSYSTEM") != "block":
                         continue
 
-                    if (slot := addresses_to_slots.get(child.parent.sys_name.split(".")[0])) is None:
+                    if (slot := addresses_to_slots.get(child.parent.parent.sys_name.split(".")[0])) is None:
```
yocalebo, line 80: Should we do a `hasattr(` check to make sure these attributes exist? Otherwise, it'll crash with `AttributeError`?

yocalebo (approved):

PR #10016, opened 2022-10-24
NAS-118706 / 22.12 / Do not try to update encryption property on ix-applications dataset

Qubad786:
## Problem

If user was using an encrypted pool for ix-applications and switches to another pool and tries to switch back to his old pool it would fail with following:
```
Error: [EFAULT] Failed to update properties: ZFSException("cannot set property for 'enc-pool/ix-applications': 'encryption' is readonly")
```

## Solution

Do not try to update encryption properties of k8s datasets as encryption is a create time attribute


yocalebo (approved):

heppu:
My apps stopped working with same error after I updated to latest nightly. I have password encrypted app pool. Should this fix the issue?

Qubad786:
@heppu it should

PR #10019, opened 2022-10-25
NAS-118620 / 22.12 / Do not crash `disk.temperature_agg` if `rrdtool` exit code is non-zero

themylogin:
(it might happen if the output contains `nan`)


themylogin:
time 0:40

yocalebo (approved):

PR #10021, opened 2022-10-25
NAS-118681 / 13.0 / Restart LDAP service if enabled on becoming master

anodos325:
This also removes logic to skip user / group cache refill on service restart.


yocalebo (changes requested):
we should move this after restarting critical services. Same with NIS

yocalebo (approved):

PR #10039, opened 2022-10-28
NAS-118779 / 22.12 / Fixes for starting apps/vms on failover

sonicaj:


src/middlewared/middlewared/plugins/failover_/event.py
```diff
@@ -530,6 +525,19 @@ def vrrp_master(self, job, fobj, ifname, event):
 
         return self.FAILOVER_RESULT
 
+    async def start_apps_vms(self):
+        if await self.middleware.call('vm.license_active'):
+            # start any VMs (this will log errors if the vm(s) fail to start)
+            # Initialize VMs first to make sure system has relevant
+            # objects for each VM initialized
+            await self.middleware.call('vm.initialize_vms')
```
yocalebo, line 533: How long does this block?

sonicaj: It's fairly quick, but i made some changes still to not wait for it.

yocalebo (approved):

PR #10041, opened 2022-10-28
NAS-118781 / 22.12 / Create recovery lock if it doesn't exist

anodos325:
The glusterfs ctdb mutex helper script was failing on fresh install because the reclock file didn't exist. Create it on lookup failure with ENOENT.

### DIFF
--- a/src/freenas/usr/local/sbin/ctdb_glfs_lock
+++ b/src/freenas/usr/local/sbin/ctdb_glfs_lock
@@ -40,7 +40,18 @@ class GlusterConn:
         return
 
     def take_reclock(self):
-        lock_file = self.glusterhdl.get_root_handle().lookup(self.reclock_path)
+        try:
+            lock_file = self.glusterhdl.get_root_handle().lookup(self.reclock_path)
+        except pyglfs.GLFSError as e:
+            if e.errno != errno.ENOENT:
+                raise
+
+            lock_file = self.glusterhdl.get_root_handle().create(
+                self.reclock_path,
+                os.O_CREAT | os.O_RDWR,
+                mode = 0o600,
+            )
+
         self.reclock_fd = lock_file.open(os.O_RDWR)
         lock_info = self.reclock_fd.posix_lock(
             fcntl.F_SETLK,


yocalebo (approved):

src/freenas/usr/local/sbin/ctdb_glfs_lock
```diff
@@ -40,7 +40,18 @@ class GlusterConn:
...
 
     def take_reclock(self):
-        lock_file = self.glusterhdl.get_root_handle().lookup(self.reclock_path)
+        try:
+            lock_file = self.glusterhdl.get_root_handle().lookup(self.reclock_path)
+        except pyglfs.GLFSError as e:
+            if e.errno != errno.ENOENT:
+                raise
+
+            lock_file = self.glusterhdl.get_root_handle().create(
```
yocalebo, line 49: Does `.create()` return the same as `.lookup()`?

anodos325: Yup. Both return a new glusterfs object handle.

```
PyDoc_STRVAR(py_glfs_obj_create__doc__,
"create(path, flags, stat=False, symlink_follow=False, mode=0o644)\n"
"--\n\n"
"Create new GLFS object (file) by path.\n\n"
"Parameters\n"
"----------\n"
"path : str\n"
"    Path of object relative to this handle.\n"
"flags : int\n"
"    open(2) flags to use to open the handle.\n"
"stat : bool, optional, default=True\n"
"    Retrieve stat information for object while performing create.\n"
"symlink_follow: bool, optional, default=True\n"
"    Follow symlinks while performing create.\n\n"
"mode : bool, optional, default=0o644\n"
"    Permissions to set on newly created directory.\n\n"
"Returns\n"
"-------\n"
"pyglfs.ObjectHandle\n"
"    New GLFS handle\n"
);
```

vs

```
PyDoc_STRVAR(py_glfs_obj_lookup__doc__,
"lookup(path, stat=False, symlink_follow=False)\n"
"--\n\n"
"Lookup existing GLFS object by path.\n\n"
"Parameters\n"
"----------\n"
"path : str\n"
"    Path of object relative to this handle.\n"
"stat : bool, optional, default=True\n"
"    Retrieve stat information for object while performing lookup.\n"
"symlink_follow: bool, optional, default=True\n"
"    Follow symlinks while performing lookup.\n\n"
"Returns\n"
"-------\n"
"pyglfs.ObjectHandle\n"
"    New GLFS handle\n"
);
```

PR #10043, opened 2022-10-28
NAS-118738 / 22.12 / Handle upstream k3s buggy klipper case

Qubad786:
## Context

Whenever a service of type `LoadBalancer` is created - k3s (when using builtin klipper load balancer) creates a daemonset against it in `kube-system` namespace and whenever the service is deleted (when scaling down app or deleting it) it is responsible for deleting the daemonset so relevant deployed pods by it can be deleted it as well.

## Problem

For some users upgrading from 22.02.4,k3s was not properly deleting older deployed daemonsets even on app deletion (which deletes the service as well) - this meant that old daemonsets with pods were still running for apps which were not present anymore or had been scaled down.

## Solution

We address the problem in the following steps:

1. Delete all `LoadBalancer` type services when bootstrapping cluster
2. Delete any rogue klipper daemonsets left
3. Finally re-deploy apps which weren't stopped so helm can re-create `LoadBalancer` services in question

This happens before apps run so it's safe to pursue and works around upstream k3s issue.

To clarify this bug in k3s is not present anymore in latest 1.25 version available in Bluefin.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
@@ -133,6 +133,38 @@ class KubernetesService(Service):
         except Exception as e:
             raise CallError(f'Failed to configure PV/PVCs support: {e}')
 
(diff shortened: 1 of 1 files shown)


sonicaj (approved):
LGTM!

PR #10046, opened 2022-10-31
NAS-118814 / 22.12 / Fix disk.sync_all when booting

yocalebo:
During boot, the standby node is trying to sync the database to the active which is not what we want. Make sure the failover status is what we want before trying to sync the entire db to the other side.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/sync.py
+++ b/src/middlewared/middlewared/plugins/disk_/sync.py
@@ -119,8 +119,10 @@ class DiskService(Service, ServiceChangeMixin):
         """
         # Skip sync disks on standby node
         licensed = self.middleware.call_sync('failover.licensed')
-        if licensed and self.middleware.call_sync('failover.status') == 'BACKUP':
-            return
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/disk_/sync.py
```diff
@@ -249,12 +251,12 @@ def sync_all(self, job):
             for delete in deleted:
                 self.middleware.send_event('disk.query', 'CHANGED', id=delete, cleared=True)
 
-        if licensed:
+        if licensed and status == 'MASTER':
```
sonicaj, line 254: Will this ever be `BACKUP` here ?

yocalebo: Huh? If it's `BACKUP` we return early up above?

sonicaj: Exactly, just curious about the changes as to if standby will never reach here what's the motivation behind this conditional ?

PR #10050, opened 2022-11-01
NAS-118571 / 22.12 / Fix used ports output when load balancer services are used

Qubad786:
## Context

When `LoadBalancer` type services are used with apps,system was not accounting for those services when determining `used_ports` of an application.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
@@ -150,12 +150,17 @@ class ChartReleaseService(CRUDService):
             resources_filters = [['metadata.namespace', '^', CHART_NAMESPACE_PREFIX]]
 
         ports_used = collections.defaultdict(list)
-        for node_port_svc in await self.middleware.call(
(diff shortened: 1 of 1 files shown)


yocalebo (approved):

Ornias1993 (commented):
There is one slight mistake here:
This should *only* be done when the default LoadBalancer "Integrated Loadbalancer"-checkbox, is enabled.

Non-default loadbalancers, such as metallb, might (and often do) use seperate IP's for seperate services.

For example:
With MetalLB, one can have 3 webservers using port 80, each with a different IP that the host has no idea about.

This version of the PR, would lead to issues in those cases... as the validator, that also uses this code portion, thinks the port is "already in use"

yocalebo (changes requested):
@Ornias1993 thanks for the input. @Qubad786 can you investigate this scenario?

Ornias1993:
> @Ornias1993 thanks for the input. @Qubad786 can you investigate this scenario?

No problem :)

Ornias1993 (commented):
Yeah that should work nicely.
In the future, clustering etc, we *might* want to revisit, but for now this is more than fine :)

yocalebo (approved):

PR #10054, opened 2022-11-02
NAS-118851 / 23.10 / API authentication using Directory Services

themylogin:


src/middlewared/Makefile
```diff
@@ -23,6 +23,7 @@ install_client:
 
 install_test:
 	bash install-dev-tools
+	apt install -y python3-pampy
```
themylogin, line 26: This is temporary to be able to run CI tests from this branch

src/middlewared/middlewared/plugins/auth_/authenticate.py
```diff
@@ -13,7 +15,7 @@ class Config:
     async def authenticate(self, username, password):
         if '@' in username:
             if (await self.middleware.call('datastore.config', 'system.settings'))['stg_ds_auth']:
-                return await self.ds_authenticate(username, password)
+                return await self.ds_authenticate(username.split('@')[0], password)
```
anodos325, line 18: Why are you removing the domain info? That tends to break PAM authentication as `pam_winbind` as it's part of the username.
```
>>> pam.authenticate("joiner", <password>)
False
>>> pam.authenticate("joiner@BILLY.GOAT", <password>)
True
```

src/middlewared/middlewared/plugins/auth_/authenticate.py
```diff
@@ -58,7 +60,14 @@ async def authenticate_local_user(self, user_id, username):
...
     async def ds_authenticate(self, username, password):
-        return None  # FIXME
+        if await self.middleware.call('auth.libpam_authenticate', username, password):
+            user = await self.middleware.call('user.get_user_obj', {'username': username, 'get_groups': True})
+            return await self.common_authenticate(username, 'ds_groups', set(user['grouplist']))
+
+    @private
+    def libpam_authenticate(self, username, password):
+        p = pam.pam()
+        return p.authenticate(username, password)
```
anodos325, line 70: You should  create a `middlewared` pam file that gets passed here as an argument. This allows us to directly control whether pam_winbind, pam_ldap, and pam_kerberos are used for authentication.

src/middlewared/middlewared/plugins/auth_/authenticate.py
```diff
@@ -13,7 +15,7 @@ class Config:
     async def authenticate(self, username, password):
         if '@' in username:
```
anodos325, line 16: Since we have PAM library, you can just use that for all authentication other than root. You just need to have a special pam config file that you regenerate when directory services settings are changed (for middleware auth).

anodos325: Using PAM file will also allow us to avoid a datastore.query in potentially hot code path.

anodos325: Sample pam file for etc_files if you're not familiar (note didn't test this particular one):

```
#
# /etc/pam.d/middleware - authentication settings for middlewared service
#

<%namespace name="pam" file="pam.inc.mako" />\
<%
        dsp = None
        if middleware.call_sync('datastore.config', 'system.settings')['stg_ds_auth']:
            dsp = pam.getDirectoryServicePam(middleware=middleware, render_ctx=render_ctx)
%>\

% if dsp:
${dsp.pam_auth()}
% endif

auth    [success=1 default=ignore]      pam_unix.so obscure sha512
# here's the fallback if no module succeeds
auth    requisite                       pam_deny.so
# prime the stack with a positive return value if there isn't one already;
# this avoids us returning an error just because nothing sets a success code
# since the modules above will each just jump around
auth    required                        pam_permit.so
# and here are more per-package modules (the "Additional" block)
```

themylogin: In some cases we need to allow root user login even if it has no password set in `/etc/shadow`, is there a method to do this with pam?

anodos325: `root` would be the one exception to this. I can't think of a clever way to avoid this. Maybe if pam.authenticate() fails, we can have special error handling for root that will fallback to comparing the password hash with what's in our db. 

You'll probably want to specify `nodelay` for pam_unix.so (assuming we don't want 2+ second delays on failure path).

anodos325: In theory we could maintain a berkley db with local overrides and place pam_userdb.so ahead of pam_unix.so.

src/middlewared/middlewared/plugins/auth_/authenticate.py
```diff
@@ -11,63 +13,37 @@ class Config:
...
-        return await self.common_authenticate(username, 'local_groups', gids)
+    def libpam_authenticate(self, username, password):
+        p = pam.pam()
+        return p.authenticate(username, password, service='middleware')
 
     @private
-    async def ds_authenticate(self, username, password):
-        return None  # FIXME
+    async def authenticate_user(self, query, local):
+        user = await self.middleware.call('user.get_user_obj', {**query, 'get_groups': True})
```
anodos325, line 43: Do we need to handle `KeyError` if for some reason user doesn't exist?

src/middlewared/middlewared/plugins/system_general/update.py
```diff
@@ -279,6 +279,9 @@ async def do_update(self, data):
         if config['crash_reporting'] != new_config['crash_reporting']:
             await self.middleware.call('system.general.set_crash_reporting')
 
+        if config['ds_auth'] != new_config['ds_auth']:
+            await self.middleware.call('etc.generate', 'pam')
```
anodos325, line 283: Should this be `pam_middleware`

anodos325: Or did you intend to move the middleware pam into the generic pam etc group?

themylogin:
time 2:00

themylogin:
time 4:00

anodos325:
Don't forget to squash this. LGTM, we can deal with any issues that come up in future commits. Don't want to let this bitrot again. :)

anodos325 (approved):

PR #10077, opened 2022-11-09
 NAS-118941 / 22.12 / fix pool.dataset.internal_dataset_filters

yocalebo:
This logic has been broken for a long time. We shouldn't include these datasets for any zpools since they are reserved for internal use only.


rick-mesta (approved):

yocalebo:
Waiting on HA API tests to run to ensure no regressions since this touches _everything_.

rick-mesta:
👍🏼

PR #10084, opened 2022-11-11
Fix race condition for display devices

Qubad786:
## Context

We use `haproxy` for reverse proxy to the VNC/SPICE clients for active vms and while starting/stopping a VM we reload `haproxy` to account for active display devices.

## Problem

We saw a user for whom sometimes randomly no endpoint for a RUNNING vm was created - each time a VM is started we reload haproxy which then was retrieving ACTIVE vms and generating an endpoint for them. However it seems that the VM status does not report ACTIVE yet and we generate haproxy configuration before the status changes which results in the endpoint not being present and display URI not working for the display device in question.

## Solution

After much deliberation i am not seeing any need for maintaining stopped/started vms separately in haproxy similar to how we are doing this in nginx where we don't care if there are any VMs at all and generate an endpoint in nginx unconditionally.
Changes have been made to only reload `haproxy` when starting VMs as that's the case when device changes will come in effect where it might have been added/removed or modified.


yocalebo (approved):

yocalebo:
approved so long as flake8 issue is resolved 👍

PR #10088, opened 2022-11-11
NAS-118973 / 22.12 / Split up API test protocols into separate files

anodos325:
Protocol tests have become more complex since they were first implemented. This splits NFS and SMB into separate files.


yocalebo (approved):

ericbsd (approved):
LGTM

PR #10089, opened 2022-11-12
NAS-119810 / 23.10 / Change IPMI sensor matching for generating alerts

rapperskull:
When using 'elist', as opposed to 'list', ipmitool generates a more detailed description of the sensor, but omits the number from the name. This means that on a system where all the sensors are described, virtually no alert is ever generated. This commit fixes the problem.


yocalebo:
@rapperskull thanks for the contribution! I can confirm that passing `elist` does remove the hex value after the sensor name. Since this branch is pretty out of date, I've opened up a fresh PR giving you credit here: https://github.com/truenas/middleware/pull/10399

PR #10090, opened 2022-11-14
NAS-119814 / 23.10 / Add ("Power Supply", "Fully Redundant") to IPMI_EVENTS_BLACKLIST

rapperskull:
I don't know where "Redundancy State" sensor came from (maybe OEM-specific?), but it's not a standard sensor type. Dell uses a "Power Supply" sensor to indicate the redundancy state.


yocalebo (approved):

yocalebo:
retest this please

PR #10094, opened 2022-11-15
NAS-119017 / 22.12 / Add glusterfs filesystem plugin

anodos325:
Use libgfapi python bindings to perform filesystem ops on glusterfs filesystem. This is less error-prone than relying on a fuse mount and gives us somewhat greater control and performance when doing recursive operations as well as a framework for testing the python bindings in our cluster API tests.


anodos325:
needs docs

yocalebo (approved):
This looks good and docstrings are fantastic but I'll feel way more comfortable once we get tests in place 😄

yocalebo:
Oops, fix the flake8 issue before merging please.

yocalebo:
retest this please

PR #10096, opened 2022-11-16
Redeploy apps after the taints have been removed

Qubad786:
## Problem

Apps are being redeployed while `NoExecute` taint is active which results in a case where if any app might have helm `pre-upgrade` hooks they won't be likely executed or at least be subject to a race condition where they might not run as the taint is removed after redeploying the apps.

## Solution

Let's redeploy the apps after the taint has been removed to avoid the problem where the hooks might be subject to the above race condition.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
@@ -154,6 +154,12 @@ class KubernetesService(Service):
             except CallError:
                 self.logger.error('Failed to remove %r daemonset', daemonset['metadata']['name'], exc_info=True)
 
+        await self.middleware.call(
+            'k8s.node.remove_taints', [
+                k['key'] for k in (node_config['spec']['taints'] or []) if k['key'] in ('ix-svc-start', 'ix-svc-stop')
+            ]
+        )
+
         # Let helm re-create load balancer services for scaled up apps
         chart_releases = await self.middleware.call('chart.release.query', [['status', 'in', ('ACTIVE', 'DEPLOYING')]])
         bulk_job = await self.middleware.call(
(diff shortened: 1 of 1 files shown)


Ornias1993 (commented):
Nice and clean solution @Qubad786 
Good start of my morning.

LGTM.

stavros-k (approved):

src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
```diff
@@ -154,6 +154,12 @@ async def post_start_internal(self):
             except CallError:
                 self.logger.error('Failed to remove %r daemonset', daemonset['metadata']['name'], exc_info=True)
 
+        await self.middleware.call(
+            'k8s.node.remove_taints', [
+                k['key'] for k in (node_config['spec']['taints'] or []) if k['key'] in ('ix-svc-start', 'ix-svc-stop')
+            ]
+        )
+
         # Let helm re-create load balancer services for scaled up apps
```
Ornias1993, line 163: We do need to wait here, for the other,, build-in, taints to be removed as well.

sonicaj (approved):

src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
sonicaj: Good point @Ornias1993, going to make a ticket for this as it is essentially separate from this and we should likely be not doing a bunch of other things too if built in taints are still active.

stavros-k: This worked fine for me btw. Tested live on my machine.

But yea build in taints need to be handled. For now it just left me couple of pods in "error" state, But new ones already have been spun up and working.

Ornias1993: @sonicaj Agreed, thats also what @stavros-k and I came up with and why we approved it either way :)

PR #10098, opened 2022-11-16
NAS-118819 / 22.12 / Handle edge case of malformed helm secrets

Qubad786:
## Problem

For a user we saw that his helm secret had become malformed due to reasons unknown which resulted in us failing to list any apps as when we queried secrets and tried to parse them to get their details that failed resulting in us failing to list down any apps.

## Solution

I explored helm behaviour in this case and noticed that `helm history` ignored the revision secret which was affected and showed down other valid revisions which were still available without having their secrets malformed.
Keeping in line with helm behaviour changes have been made to ignore decode exceptions in this case and just log that this happened once.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/secrets_management.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/secrets_management.py
@@ -1,3 +1,4 @@
+import binascii
 import gzip
 import json
 import os
@@ -13,6 +14,8 @@ from .utils import CHART_NAMESPACE_PREFIX, get_namespace
 
 
 class ChartReleaseService(Service):
+
+    LOGGED_SECRET = set()
 
     class Config:
         namespace = 'chart.release'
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/chart_releases_linux/secrets_management.py
```diff
@@ -14,6 +15,8 @@
 
 class ChartReleaseService(Service):
 
+    LOGGED_SECRET = set()
```
yocalebo, line 18: How large can this grow? I don't like idea of storing secret keys in memory for life of process while it's running.

Qubad786: For most users it does not happen and for the user in question it was 2 secrets..
I don't expect it to be a widespread problem

sonicaj (approved):
Interesting case :P

src/middlewared/middlewared/plugins/chart_releases_linux/secrets_management.py
yocalebo: Okay, I trust you 😄

yocalebo (approved):

PR #10099, opened 2022-11-17
NAS-119029 / 22.12 / Do not start S.M.A.R.T. if there are no devices to monitor

themylogin:


themylogin:
time 0:30

yocalebo (commented):
Will this still allow the service to run or will it prevent the service from starting if there are 0 devices to monitor?

yocalebo:
Great, thanks

yocalebo (approved):

PR #10100, opened 2022-11-17
NAS-118897 / 22.12 / Allow a token created by token to use the web shell

themylogin:


themylogin:
time 2:00

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -66,6 +66,16 @@ def is_valid(self):
     def notify_used(self):
         self.last_used_at = time.monotonic()
 
+    def root_credentials(self):
+        credentials = self.parent_credentials
+        while True:
+            if isinstance(credentials, TokenSessionManagerCredentials):
```
yocalebo, line 72: Does it make sense to just do something like
```
while True:
    if isinstance(credentials, TokenSessionManagerCredentials):
        credentials = credentials.token.parent_credentials
    else:
        return credentials

yocalebo (approved):
Great job!

PR #10102, opened 2022-11-17
NAS-119044 / 22.12 / Add endpoint to check service configuration prior to start

anodos325:
Various services have conditions that must be satisfied prior to starting the service. Since webui is reporting on start failures now, also allow presenting clear reasons why the service won't be started if it's horribly misconfigured or state is broken.

### DIFF
--- a/src/middlewared/middlewared/plugins/service.py
+++ b/src/middlewared/middlewared/plugins/service.py
@@ -136,6 +136,16 @@ class ServiceService(CRUDService):
 
         await self.middleware.call('service.generate_etc', service_object)
 
+        try:
+            await service_object.check_configuration()
+        except CallError:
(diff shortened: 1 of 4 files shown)


anodos325:
> The issue I see with this is that systemd will unconditionally start these services on boot. Should we add systemd files override that will perform these preliminary checks?

At least for knfsd and smbd the checks are performed prior to service start via smbd. This is more about addressing user confusion about why services fail to start via `service.start` / webui. Systemd journal lines are good for cases where something totally unexpected happens, but are somewhat less than clear for less technically adept end-users.

william-gr:
@themylogin ping

william-gr (approved):

themylogin (approved):

PR #10104, opened 2022-11-18
NAS-119026 / 22.12 / Properly wait for built in taints to be removed

Qubad786:
## Problem

Before redeploying apps we were not ensuring that no built-in taints are present and there was an infinite waiting cycle on pods to shift to running state before kube-router rules were configured.

## Solution

Wait for built in taints to be removed and properly implement wait for other different contexts in post start lifecycle of k8s service to confirm we don't have infinite waiting cycle.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
@@ -164,6 +164,23 @@ class KubernetesService(Service):
             ]
         )
 
+        # Wait for taints to be removed - this includes the built-in taints which might be present
+        # at this point i.e node.kubernetes.io/not-ready
+        taint_timeout = 600
+        sleep_time = 5
+        taints = None
+        while taint_timeout > 0:
+            taints = (await self.middleware.call('k8s.node.config'))['spec'].get('taints')
+            if not taints:
+                break
+
+            await asyncio.sleep(sleep_time)
+            taint_timeout -= sleep_time
+        else:
+            raise CallError(
+                f'Timed out waiting for {", ".join([taint["key"] for taint in taints])!r} taints to be removed'
+            )
+
         # Let helm re-create load balancer services for scaled up apps
         chart_releases = await self.middleware.call('chart.release.query', [['status', 'in', ('ACTIVE', 'DEPLOYING')]])
         bulk_job = await self.middleware.call(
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
```diff
@@ -160,6 +160,22 @@ async def post_start_internal(self):
...
+        # Wait for taints to be removed - this includes the built-in taints which might be present
+        # at this point i.e node.kubernetes.io/not-ready
+        taint_timeout = 600
+        taints = None
+        while taint_timeout > 0:
+            taints = (await self.middleware.call('k8s.node.config'))['spec'].get('taints')
+            if not taints:
+                break
+
+            await asyncio.sleep(5)
```
yocalebo, line 172: We should create another variable for sleep time `sleep_time` or something equivalent. Also, should we wait 5 seconds each iteration?

yocalebo (changes requested):

src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
Qubad786: We can change it to something else but 5 seconds looks like a decent delay?
In most cases we should not be hitting these sleeps at all..

Ornias1993:
@Qubad786 A that off topic, but I just noticed that you might not want to execute the LB redeploy, if the internal loadbalancer isn't even enabled at all. (line 180 in this PR )

Qubad786:
@Ornias1993 motivation for that change was to clean up a bug in k3s where it was not removing old daemonsets of apps which had been deleted..
That bug cannot be reproduced now, we would be likely removing these changes for the next major release as these are not required explicitly because k3s is capable of cleaning up service load balancers daemonsets and pods but at some point it was not doing that properly.

Ornias1993:
> @Ornias1993 motivation for that change was to clean up a bug in k3s where it was not removing old daemonsets of apps which had been deleted.. That bug cannot be reproduced now, we would be likely removing these changes for the next major release as these are not required explicitly because k3s is capable of cleaning up service load balancers daemonsets and pods but at some point it was not doing that properly.

To be clear:
It was the SVLB daemonsets this was about.
Which are only represent when the build-in loadbalancer is used ;-)

sonicaj:
> > @Ornias1993 motivation for that change was to clean up a bug in k3s where it was not removing old daemonsets of apps which had been deleted.. That bug cannot be reproduced now, we would be likely removing these changes for the next major release as these are not required explicitly because k3s is capable of cleaning up service load balancers daemonsets and pods but at some point it was not doing that properly.
> 
> To be clear: It was the SVLB daemonsets this was about. Which are only represent when the build-in loadbalancer is used ;-)
> 
> But kinda offtopic, so i'll leave it at that :)

No worries - good observation btw ;) 
Btw @Qubad786 still has a point as users could have enabled/disabled it at some point so we don't know for sure :P

sonicaj (approved):

PR #10106, opened 2022-11-18
NAS-118922 / 22.12 / fix and improve pool.replace

yocalebo:
1. don't call blocking I/O in main event loop
2. remove unused imports
3. logical refactor to make it easier to read

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/replace_disk.py
+++ b/src/middlewared/middlewared/plugins/pool_/replace_disk.py
@@ -80,12 +80,14 @@ class PoolService(Service):
         await self.middleware.call('pool.format_disks', job, format_opts)
 
         job.set_progress(30, 'Replacing disk')
+        new_devname = vdev[0].replace('/dev/', '')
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/pool_/replace_disk.py
```diff
@@ -45,78 +44,57 @@ async def replace(self, job, oid, options):
...
-                self.logger.warn('Failed to detach device', exc_info=True)
+                self.logger.warning('Failed to detach device with label %r', options['label'], exc_info=True)
         finally:
             # Needs to happen even if replace failed to put back disk that had been
             # removed from swap prior to replacement
             asyncio.ensure_future(self.middleware.call('disk.swaps_configure'))
 
-        if old_disk:
-            await self.middleware.call('disk.copy_settings', old_disk, disk)
+        if options['preserve_settings']:
```
themylogin, line 90: As we are changing logic here, should we add a test in order to ensure this is still working as expected?

themylogin (approved):

PR #10107, opened 2022-11-18
Refactor truecommand wireguard interface name

sonicaj:
This PR introduces changes to rename truecommand wireguard interface name to`ix-truecommand-wg` as by default it was consuming `wg0` right now but now with official wireguard application on the way, this will conflict as the app can only consume `wg0`.


yocalebo (changes requested):
We need to add this to `INTERNAL_INTERFACES` in `netif_linux/utils.py`

src/middlewared/middlewared/plugins/interface/netif_linux/utils.py
```diff
@@ -8,7 +8,8 @@
 
 
 INTERNAL_INTERFACES = [
-    "wg", "lo", "tun", "tap", "docker", "veth", "kube-bridge", "kube-dummy-if", "vnet", "openvpn", "macvtap",
+    "wg", "lo", "tun", "tap", "docker", "veth", "kube-bridge", "kube-dummy-if", "vnet",
```
Ornias1993, line 11: I think we want to remove "wg" on here. As i'm not sure users can actually use the wg interface from from app for host-services at the moment.

yocalebo (approved):

PR #10109, opened 2022-11-18
Allow to log in to the Web UI as root even with password disabled whe…

themylogin:
…n no other password-enabled local administrator is present

### DIFF
--- a/src/middlewared/middlewared/etc_files/web_ui_root_login_alert.py
+++ b/src/middlewared/middlewared/etc_files/web_ui_root_login_alert.py
@@ -1,0 +1,7 @@
+
+
+async def render(service, middleware, render_ctx):
+    if await middleware.call('privilege.always_has_root_password_enabled'):
+        await middleware.call('alert.oneshot_create', 'WebUiRootLogin', None)
+    else:
+        await middleware.call('alert.oneshot_delete', 'WebUiRootLogin', None)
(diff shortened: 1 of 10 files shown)


themylogin:
time 8:00

anodos325:
render context for etc plugin for groups already has user.query and group.query results. We should re-use those in                 `{'type': 'py', 'path': 'web_ui_root_login_alert'}` script if at all possible. Did you do any benchmark for performance impact in  case of 500 or so local users and groups when doing `etc.generate user`? I think we have one customer who has around 500 local users and 500 local groups and so that's not completely unrealistic target.

src/middlewared/middlewared/etc_files/web_ui_root_login_alert.py
```diff
@@ -0,0 +1,7 @@
+
+
+async def render(service, middleware, render_ctx):
```
yocalebo, line 3: Agree with Andrew. `render_ctx` should already have user/group query results so it'd be best to pass that through to the other functions to keep from having to unnecessary datastore.query calls.

yocalebo (changes requested):

yocalebo (approved):

yocalebo:
Wait for @anodos325 final review before merge, please.

anodos325 (approved):

PR #10115, opened 2022-11-21
NAS-119008 / 22.12 / properly send HA events on SCALE

yocalebo:


src/middlewared/middlewared/plugins/device_/vrrp_events.py
```diff
@@ -59,26 +60,26 @@ def run(self):
                     sleep(self._retry_timeout)
 
 
+def start_or_stop_vrrp_thread(middleware, licensed, shutting_down):
+    try:
+        vthread = [i for i in threading.enumerate() if i.name == VRRP_THREAD_NAME][0]
```
themylogin, line 65: It's a bit unusual to identify threads by their names and to use the thread name to ensure it is not already running. The classic approach is to just have a global variable with this thread (and start/stop thread and assign this variable using a shared lock to avoid race conditions).

src/middlewared/middlewared/plugins/failover_/event.py
```diff
@@ -588,11 +604,28 @@ def vrrp_backup(self, job, fobj, ifname, event):
...
+
+            # restarting keepalived sends a priority 0 advertisement
+            # which means any VIP that is on this controller will be
+            # migrated to the other controller
+            logger.info('Transitioning all VIPs off this node')
+            self.run_call('service.restart', 'keepalived')
+        finally:
+            if start_new_vrrp_thread:
+                logger.info('Restarting vrrp event thread')
+                vrrp_events.VrrpFifoThread(middleware=self.middleware).start()
```
themylogin, line 628: Should we have a single way to start/stop vrrp thread? We can even move it to its own `failover.vrrp` plugin with its own start/stop methods.

src/middlewared/middlewared/plugins/device_/vrrp_events.py
themylogin: I actually see this was approach that was used before, what's wrong with that?

src/middlewared/middlewared/plugins/failover_/event.py
yocalebo: yeah, I thought about that but got lazy. I'll create a separate plugin

src/middlewared/middlewared/plugins/device_/vrrp_events.py
themylogin: If you move managing this thread to a plugin, you won't even need to access the thread variable from other modules

yocalebo: yeah agree, that was the biggest problem with the way I was doing it previously

anodos325 (approved):

PR #10117, opened 2022-11-22
`user.setup_local_administrator`

themylogin:


anodos325:
Why are we using UID 1000? A lot of kubernetes apps use this same UID. Potential for collisions is high.

yocalebo (changes requested):
This is guaranteed to fail on most installations. We can't use 1000 as the uid, we need to come up with something else.

themylogin:
@yocalebo this won't collide with existing installations, we create this `admin` user only on fresh installations. Existing installation users will have to create `admin` (or whatever they like) themselves.

yocalebo:
> @yocalebo this won't collide with existing installations, we create this `admin` user only on fresh installations. Existing installation users will have to create `admin` (or whatever they like) themselves.

Okay, we should still choose a UID < 1000 since it lessens the probability of collisions.

yocalebo:
We need to change `truenas-set-authentication-method.py` to also use this new UID right?

yocalebo:
> We need to change `truenas-set-authentication-method.py` to also use this new UID right?

I'm blind and not paying attention. Ignore me.

src/middlewared/middlewared/plugins/account.py
```diff
@@ -878,26 +899,35 @@ async def has_local_administrator_set_up(self):
...
                 if options['ec2']['instance_id'] != await self.middleware.call('ec2.instance_id'):
                     raise CallError('Incorrect EC2 instance ID', errno.EACCES)
 
-        root = await self.middleware.call('user.query', [('username', '=', 'root')], {'get': True})
-        await self.middleware.call('user.update', root['id'], {'password': password})
+        if await self.middleware.call('user.has_local_administrator_set_up'):
+            raise CallError('Local administrator is already set up', errno.EALREADY)
+
+        if username == 'admin':
+            if await self.middleware.call('user.query', [['uid', '=', 1000]]):
```
yocalebo, line 922:
```suggestion
            if await self.middleware.call('user.query', [['uid', '=', 950]]):
```

yocalebo (changes requested):

src/middlewared/middlewared/plugins/account.py
```diff
@@ -878,26 +899,35 @@ async def has_local_administrator_set_up(self):
...
                     raise CallError('Incorrect EC2 instance ID', errno.EACCES)
 
-        root = await self.middleware.call('user.query', [('username', '=', 'root')], {'get': True})
-        await self.middleware.call('user.update', root['id'], {'password': password})
+        if await self.middleware.call('user.has_local_administrator_set_up'):
+            raise CallError('Local administrator is already set up', errno.EALREADY)
+
+        if username == 'admin':
+            if await self.middleware.call('user.query', [['uid', '=', 950]]):
+                raise CallError('A user with uid=1000 already exists, setting up local administrator is not possible',
```
themylogin, line 923:
```suggestion
                raise CallError('A user with uid=950 already exists, setting up local administrator is not possible',
```

yocalebo (approved):
Looks good, however, please wait for @anodos325 final review before merge, please.

anodos325:
There are ways (unsupported) via auxiliary parameters that users can intentionally enable mappings for LDAP and AD that go below UID 1000. I think a periodic alert (maybe once per day) that calls `user.get_user_obj` for the admin uid and compares NSS results (username, gecos, path, gid) with that which is expected based on our db entry and alert if there's a mismatch would be a good idea.

src/freenas/usr/local/bin/truenas-set-authentication-method.py
```diff
@@ -36,12 +34,16 @@
...
         user["bsdusr_unixhash"] = password
+        user["bsdusr_smbhash"] = "*"
         user["bsdusr_home"] = home
         user["bsdusr_full_name"] = "Local Administrator"
         user["bsdusr_builtin"] = 0
+        user["bsdusr_smb"] = 0
+        user["bsdusr_password_disabled"] = 0
+        user["bsdusr_locked"] = 0
         user["bsdusr_sudo"] = 1
         user["bsdusr_group_id"] = group_id
```
anodos325, line 48: Maybe we need hard-coded `gid` ?

anodos325:
When this user is generated (perhaps in our installer), we should make the directory /home/admin, chown to correct uid/gid, then chmod to 0700. The filesystem ACL API won't allow changes to /home/admin permissions.

anodos325:
Alert can also use `filesystem.stat` for `/home/admin` and verify that permissions are `0700` and `acl` is False.

src/freenas/usr/local/bin/truenas-set-authentication-method.py
anodos325: Perhaps `gid 0` is okay?

anodos325:
Do we have any validation that prevents user from creating a second user with uid 950? If not we would possibly want to do that, otherwise a user with permissions to create new local users could create duplicate uid 950, then SSH in as that user and append another key to authorized_keys in /home/admin, and gain admin access.

I think that perhaps with multi-user access to APIs we may want to prevent creation of users with duplicate IDs.

anodos325 (approved):

anodos325:
Oh. Just noticed we're missing `filesystem.stat` call in alert for `/home/admin` to check permissions.

src/middlewared/middlewared/plugins/account.py
```diff
@@ -878,26 +901,46 @@ async def has_local_administrator_set_up(self):
...
-        root = await self.middleware.call('user.query', [('username', '=', 'root')], {'get': True})
-        await self.middleware.call('user.update', root['id'], {'password': password})
+        if await self.middleware.call('user.has_local_administrator_set_up'):
+            raise CallError('Local administrator is already set up', errno.EALREADY)
+
+        if username == 'admin':
+            if await self.middleware.call('user.query', [['uid', '=', ADMIN_UID]]):
+                raise CallError(
+                    f'A user with uid={ADMIN_UID} already exists, setting up local administrator is not possible',
+                    errno.EALREADY,
```
anodos325, line 927: errno.EEXIST perhaps?

PR #10120, opened 2022-11-23
NAS-118966 / 22.12 / Refactor retrieving items logic from middleware to catalog validation

sonicaj:
This PR introduces changes to move out reading catalog changes from middleware to catalog_validation repo. Motivation behind this is to be able to use a single file to read the entirety of the catalog instead of traversing the entire catalog and reading it's contents in realtime whenever a catalog sync is initiated.

These performance improvements are intended for official catalog only now but a subsequent PR will be made which will allow other catalogs to benefit from the improvements.


src/middlewared/middlewared/plugins/catalogs_linux/items.py
```diff
@@ -146,58 +136,46 @@ def items_internal(self, job, label, options):
...
+
+        return self.get_trains_impl(job, catalog, options)
 
     @private
-    def get_trains(self, job, catalog, options):
+    def retrieve_trains_data_from_json(self, catalog, options):
+        trains_to_traverse = retrieve_train_names(
+            catalog['location'], options['retrieve_all_trains'], options['trains']
+        )
+        with open(os.path.join(catalog['location'], CATALOG_JSON_FILE), 'r') as f:
```
yocalebo, line 154: Should we catch `FileNotFoundError` here?

yocalebo (approved):
Brilliant!

Qubad786 (approved):

Ornias1993:
We'll sync with this change within the week :)

PR #10125, opened 2022-11-25
NAS-119136 / 22.12 / Add glusterfs.filesystem tests

anodos325:
This commit adds basic CI tests for private glusterfs filesystem API endpoints that will be used by middleware to interact with contents of glusterfs filesystems.

### DIFF
--- a/cluster-tests/tests/filesystem/test_002_glusterfs_filesystem_plugin.py
+++ b/cluster-tests/tests/filesystem/test_002_glusterfs_filesystem_plugin.py
@@ -1,0 +1,275 @@
+import pytest
+import stat
+
+from config import CLUSTER_INFO, CLUSTER_IPS, TIMEOUTS
+from contextlib import contextmanager
+from copy import deepcopy
+from exceptions import JobTimeOut
+from utils import make_ws_request, wait_on_job
+from pytest_dependency import depends
+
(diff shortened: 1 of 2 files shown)


cluster-tests/tests/filesystem/test_002_glusterfs_filesystem_plugin.py
```diff
@@ -0,0 +1,275 @@
...
+    depends(request, ['HAS_ROOT_UUID'])
+    ip = CLUSTER_IPS[0]
+
+
+    with create_test_object(ip, True, **{
+        'volume_name': CLUSTER_INFO['GLUSTER_VOLUME'],
+        'path': 'GFAPI_TESTDIR_SETATTRS',
+    }) as gl_obj:
+        # create a bunch of test files
+        for i in range(0, 100):
```
william-gr, line 203: Is this a magic number? Do we need that many for some special reason? Just thinking of 200+ calls for this single test

anodos325: Magic. I can reduce to something smaller if you wish.

william-gr: I just thought it would generate too many calls for no good reason. But if you think thats fine, so be it

PR #10127, opened 2022-11-28
NAS-118895 / 22.12 / Handle edge case for missing manifests in helm chart

Qubad786:
## Context

If an application's helm chart does not contain any kubernetes resources it will result in middleware not properly able to report what apps are installed as it expects manifests in helm secrets to be always present which is not the case.


Ornias1993:
Clean and simple fix, lgtm!

sonicaj (approved):

PR #10129, opened 2022-11-28
Auto 312

alexandrabain:
/user/has_root_password replaced by user/has_local_administrator_set_up

### DIFF
--- a/tests/api2/test_011_user.py
+++ b/tests/api2/test_011_user.py
@@ -273,7 +273,7 @@ def test_24_look_user_is_delete(request):
 
 def test_25_has_root_password(request):
     depends(request, ["user_02", "user_01"])
-    results = GET('/user/has_local_administrator_set_up/', anonymous=True).json() is True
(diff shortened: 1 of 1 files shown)


tests/api2/test_011_user.py
```diff
@@ -273,7 +273,7 @@ def test_24_look_user_is_delete(request):
 
 def test_25_has_root_password(request):
     depends(request, ["user_02", "user_01"])
-    assert GET('/user/has_root_password/', anonymous=True).json() is True
+    assert GET('/user/has_local_administrator_set_up/', anonymous=True).json() is True
```
ericbsd, line 276: Should this be account installed of user?

ericbsd (changes requested):

tests/api2/test_011_user.py
alexandrabain: No I checked in the api docs, user is corrected. tested in postman before updating

ericbsd (approved):
LGTM

PR #10138, opened 2022-11-30
NAS-118929 / 23.10 / Allow binding authorization token to its origin

themylogin:
We do not always require to match token with its origin (namely, remote IP address) because some use cases (e.g. semiautomatic remote SSH connection setup) will break if we do so.

### DIFF
--- a/src/middlewared/middlewared/plugins/auth.py
+++ b/src/middlewared/middlewared/plugins/auth.py
@@ -1,8 +1,6 @@
 from datetime import datetime, timedelta
 import errno
 import re
-import socket
-import struct
 import time
 import warnings
 
@@ -19,7 +17,8 @@ from middlewared.service import (
 )
 from middlewared.service_exception import MatchNotFound
 import middlewared.sqlalchemy as sa
-from middlewared.utils.nginx import get_peer_process, get_remote_addr_port
+from middlewared.utils.nginx import get_peer_process
+from middlewared.utils.origin import UnixSocketOrigin, TCPIPOrigin
 from middlewared.utils.crypto import generate_token
 from middlewared.validators import Range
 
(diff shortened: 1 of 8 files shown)


themylogin:
time 3:00

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -194,18 +185,18 @@ def dump(self):
...
-    if host in ["127.0.0.1", "::1"]:
-        return True
+    if isinstance(session.app.origin, TCPIPOrigin):
+        if session.app.origin.addr in ["127.0.0.1", "::1"]:
+            return True
 
-    if host in ["169.254.10.1", "169.254.10.2", "169.254.10.20", "169.254.10.80"] and port <= 1024:
-        return True
+        if (
+                session.app.origin.addr in ["169.254.10.1", "169.254.10.2", "169.254.10.20", "169.254.10.80"] and
```
yocalebo, line 196: .20 and .80 are not used on SCALE.

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -333,10 +321,14 @@ async def check_password(self, username, password):
         return await self.middleware.call('auth.authenticate', username, password) is not None
 
     @no_auth_required
-    @accepts(Int('ttl', default=600, null=True), Dict('attrs', additional_attrs=True))
+    @accepts(
+        Int('ttl', default=600, null=True),
+        Dict('attrs', additional_attrs=True),
+        Bool('match_origin', default=False),
```
yocalebo, line 327: Can you put a doc string for the new arg?

yocalebo (approved):

PR #10148, opened 2022-11-30
NAS-119138 / 23.10 / Improve validation checks for certs

Qubad786:
## Problem

If a CSR was used for FTP service which is not a valid selection - validation would error out with a `KeyError`.

## Solution

Method used to validate if a cert is valid or not was not safely checking for keys in case of a CSR as they would not be present.
It is already validated that a CSR cannot be used but additional validation is still done to outline any other potential problems which the cert/CSR might have.

### DIFF
--- a/src/middlewared/middlewared/plugins/crypto_/certificates.py
+++ b/src/middlewared/middlewared/plugins/crypto_/certificates.py
@@ -80,7 +80,6 @@ class CertificateService(CRUDService):
         # General method to check certificate health wrt usage in services
         cert = await self.middleware.call('certificate.query', [['id', '=', id]])
         verrors = ValidationErrors()
-        valid_key_size = {'EC': 28, 'RSA': 2048}
         if cert:
             cert = cert[0]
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/crypto_/certificates.py
```diff
@@ -110,25 +110,28 @@ async def cert_services_validation(self, id, schema_name, raise_verrors=True):
...
-                    schema_name,
-                    'Please use a certificate whose digest algorithm has at least 112 security bits'
-                )
-
-            if cert['revoked']:
-                verrors.add(
-                    schema_name,
-                    'This certificate is revoked'
-                )
+            if not cert['cert_type_CSR']:
```
sonicaj, line 113: I think it's better to just not validate rest of the stuff at all if the cert is not okay - like why should we care what the key size if the selection in question is a CSR which is wrong ?

sonicaj (approved):

Qubad786:
backport

PR #10161, opened 2022-12-01
NAS-119007 / 23.10 / Present zvol snapshot count in dataset details

anodos325:
This adds a new method zfs.dataset.snapshot_count that maintains a persistent tdb-backed cache of snapshot count for datasets that gets invalidated based on the snapshots_changed timestamp presented by libzfs.

### DIFF
--- a/src/middlewared/middlewared/alert/source/snapshot_count.py
+++ b/src/middlewared/middlewared/alert/source/snapshot_count.py
@@ -1,5 +1,3 @@
-from collections import defaultdict
-
 from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, Alert, AlertSource
 from middlewared.alert.schedule import CrontabSchedule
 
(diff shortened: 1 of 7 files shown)


yocalebo (approved):
Nice!

src/middlewared/middlewared/alert/source/snapshot_count.py
```diff
@@ -33,10 +31,10 @@ async def check(self):
         max_total = await self.middleware.call("pool.snapshottask.max_total_count")
 
         total = 0
-        datasets = defaultdict(lambda: 0)
-        for snapshot in await self.middleware.call("zfs.snapshot.query", [], {"select": ["name"]}):
-            total += 1
-            datasets[snapshot["name"].split("@")[0]] += 1
+        datasets = await self.middleware.call("zfs.snapshot.query", [], {"extra": {"count_only": True}})
```
sonicaj, line 34: It feels weird that a query endpoint is giving a dict - even if we want a count, getting an integer makes more sense rather then a dict. List should do the job nicely ?

PR #10171, opened 2022-12-02
NAS-119237 / 23.10 / Correctly retrieve attachments of a path

sonicaj:
This commit adds changes to also cover the scenario where we might be provided with a path which is child of configured path and in that case we should be correctly retrieving the attachment in question.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/dataset_attachments.py
+++ b/src/middlewared/middlewared/plugins/pool_/dataset_attachments.py
@@ -36,13 +36,13 @@ class PoolDatasetService(Service):
         return []
 
     @private
-    async def attachments_with_path(self, path, check_path_child_of_attachment=False):
+    async def attachments_with_path(self, path, check_parent=False):
         result = []
         if isinstance(path, str) and not path.startswith('/mnt/'):
-            self.logger.warning('%s: uexpected path not located within pool mountpoint', path)
+            self.logger.warning('%s: unexpected path not located within pool mountpoint', path)
 
         if path:
-            options = {'check_path_child_of_resource': check_path_child_of_attachment}
+            options = {'check_parent': check_parent}
             for delegate in self.attachment_delegates:
                 attachments = {'type': delegate.title, 'service': delegate.service, 'attachments': []}
                 for attachment in await delegate.query(path, True, options):
--- a/src/middlewared/middlewared/plugins/pool_/dataset_attachments.py
+++ b/src/middlewared/middlewared/plugins/pool_/dataset_attachments.py
@@ -36,15 +36,16 @@ class PoolDatasetService(Service):
         return []
 
     @private
-    async def attachments_with_path(self, path):
+    async def attachments_with_path(self, path, check_path_child_of_attachment=False):
         result = []
         if isinstance(path, str) and not path.startswith('/mnt/'):
             self.logger.warning('%s: uexpected path not located within pool mountpoint', path)
 
         if path:
+            options = {'check_path_child_of_resource': check_path_child_of_attachment}
             for delegate in self.attachment_delegates:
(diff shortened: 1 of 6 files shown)


themylogin (changes requested):
Please provide an integration test that fails without this change

tests/api2/test_attachment_querying.py
```diff
@@ -0,0 +1,34 @@
...
+        assert call('pool.dataset.attachments_with_path', parent_path) == []
+
+        with nfs_share(parent_dataset):
+            attachments = call('pool.dataset.attachments_with_path', parent_path)
+            assert len(attachments) > 0, attachments
+            assert attachments[0]['type'] == 'NFS Share', attachments
+
+            with dataset(CHILD_DATASET) as child_dataset:
+                child_path = f'/mnt/{child_dataset}'
+                call('sharing.nfs.create', {'path': child_path})
```
themylogin, line 31: Why not also use a context manager that will remove the test share upon exiting?

tests/api2/test_attachment_querying.py
```diff
@@ -0,0 +1,34 @@
...
+                attachments = call('pool.dataset.attachments_with_path', child_path)
+                assert len(attachments) > 0, attachments
```
themylogin, line 33: Will this test fail if you remove your patch? I am not nitpicking, I am genuinely trying to understand which type of incorrect behavior are you fixing.

themylogin (changes requested):

tests/api2/test_attachment_querying.py
sonicaj: Ah sorry, missed this - this is not intended

tests/api2/test_attachment_querying.py
sonicaj: Yes it will - so the problem is that when we query attachments, we are tackling the following case:

If `/mnt/pool/smb` is a SMB share and if we were to see attachments of `/mnt/pool`, we would correctly retrieve that there is a SMB attachment involved.

Now if we check `/mnt/pool/smb` to see if there are attachments we would find that yes there is a SMB attachment involved.

However if we were to check any child of the configured path i.e `/mnt/pool/smb/child` middleware would not report any attachments for this path which is wrong as the path is being consumed by SMB.

This PR fixes/addresses that

themylogin: So with your changes if we try to delete `pool/smb/child` dataset, the system will destroy the share for `pool/smb`. Won't that be a little bit too much?

themylogin (changes requested):

tests/api2/test_attachment_querying.py
anodos325: I agree, that's very undesirable.

themylogin (approved):

src/middlewared/middlewared/plugins/pool_/dataset_attachments.py
```diff
@@ -36,15 +36,16 @@ async def attachments(self, oid):
...
 
     @private
-    async def attachments_with_path(self, path):
+    async def attachments_with_path(self, path, check_path_child_of_attachment=False):
         result = []
         if isinstance(path, str) and not path.startswith('/mnt/'):
             self.logger.warning('%s: uexpected path not located within pool mountpoint', path)
 
         if path:
+            options = {'check_path_child_of_resource': check_path_child_of_attachment}
```
yocalebo, line 45: `check_path_child_of_resource` isn't a good name and doesn't actually explain what's being done. I think we should probably do something less wordy and call it something like `check_parent` or something along those lines. @anodos325 what do you think?

yocalebo (changes requested):

yocalebo (approved):
I'll approve for now but the entire child path checking logic needs to be simplified and fixed since the names of the functions don't actually convey what they're doing.

PR #10173, opened 2022-12-02
improve efficiency of reporting.iostat.DiskStats

yocalebo:
Investigating scalability problems on a large internal system showed that an abnormal amount of time was spent in `psutil.disk_io_counters`. Specifically, it was busy doing `readlines()` on `/proc/diskstats`. This fixes that problem by transposing the same logic to this class but instead of reading the entire file into memory, iterate over it. Generating flamegraph afterwards has shown that this is no longer a problem.

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/iostat.py
+++ b/src/middlewared/middlewared/plugins/reporting/iostat.py
@@ -1,4 +1,11 @@
-from psutil import disk_io_counters
+from collections import namedtuple
+
+dstats = namedtuple('dstats', [
+    'read_count', 'write_count',
+    'read_bytes', 'write_bytes',
+    'read_time', 'write_time',
+    'busy_time',
+])
 
 
 class DiskStats:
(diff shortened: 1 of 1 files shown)


themylogin (approved):
If it works faster now I am fine with it, but why is reading the file line-by-line is more efficient than reading it all at once?

yocalebo:
> If it works faster now I am fine with it, but why is reading the file line-by-line is more efficient than reading it all at once?

Here is results of simple test using `timeit` for `readlines()` or `for line in f`. It shows ~155% difference in time.
```
python3 time.py
readlines() 0.04570064700965304
for_lines() 0.0050824890058720484
```

It's more efficient since `readlines()` reads entire contents of file into memory formatting it as a list of strings while `line in f` iterates over the file line by line.

yocalebo:
The performance difference is negligible when the file is small but becomes more noticeable when it's larger. (which can be from any numerous reaons like zvols, dm, md, disks, partitions for each disk, etc)

PR #10174, opened 2022-12-02
NAS-119055 / 23.10 / Add kubernetes passthrough mode

Qubad786:
## Context

It was requested that we allow kubernetes passthrough mode which will disable applications functionality and only expose the bare kubernetes cluster for consumption.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
@@ -134,6 +134,13 @@ class KubernetesService(ConfigService):
                 'System is not licensed to use Applications'
             )
 
+        license = await self.middleware.call('system.license')
+        if data['passthrough_mode'] and (not license or '-MINI-' in license['system_product']):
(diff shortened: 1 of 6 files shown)


src/middlewared/middlewared/plugins/kubernetes_linux/update.py
```diff
@@ -132,6 +134,14 @@ async def validate_data(self, data, schema, old_data):
                 'System is not licensed to use Applications'
             )
 
+        if data['passthrough_mode'] and (
+            license := self.middleware.call('system.license')
+        ) and '-MINI-' not in license['system_product']:
+            verrors.add(
+                f'{schema}.passthrough_mode',
+                'Can only be enabled on licensed iX enterprise hardware'
+            )
```
Ornias1993, line 143: Can't we just like: Checkbox this away, with a very clear warning popup.
A tad like what we will do with disabling hostPath validation?

yocalebo (approved):

PR #10176, opened 2022-12-03
NAS-119254 / 13.0 / keep references for created asyncio tasks

rkojedzinszky:
According to https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task, references to created tasks should be held until the task terminates.


yocalebo (approved):
Fantastic, thanks for the contribution!

PR #10180, opened 2022-12-05
NAS-119258 / 23.10 / Disable k3s/kubelet dataset snapshot

Qubad786:
## Context

On backup of kubernetes, we take snapshot of complete ix-applications dataset (except for a few) so that we can rollback the snapshots when a cluster restore is taking place.

Similar to how we ignore some other datasets on backup we should be ignoring `ix-applications/k3s/kubelet` snapshot as it's redundant/useless because we are essentially re-initializing docker.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/backup.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/backup.py
@@ -89,7 +89,7 @@ class KubernetesService(Service):
 
     @private
     def to_ignore_datasets_on_backup(self, k8s_dataset):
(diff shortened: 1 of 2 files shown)


sonicaj (commented):
LGTM, however let's add integration tests for this please which ensure we are not taking snapshot of these datasets on backup. Thanks!

Ornias1993:
Thanks, this should save a lot of storage bloat :)

Ornias1993:
@Qubad786 If you've the spare time before launch, it might be worth to nuke existing snapshots on kubelet on upgrade to bluefin in another PR.

sonicaj (approved):

PR #10182, opened 2022-12-05
NAS-119350 / 13.0 / devd: read and parse events in asyncio loop

rkojedzinszky:


src/middlewared/middlewared/plugins/device_/devd_events.py
```diff
@@ -51,26 +50,23 @@ def parse_devd_message(msg):
...
     middleware.logger.info('devd connection established')
     DEVD_CONNECTED = True
 
     while True:
-        # line = await reader.readline()
-        line = await middleware.run_in_thread(reader.readline)
+        line = await reader.readline()
         line = line.decode(errors='ignore')
         if line == "":
+            writer.close()
```
yocalebo, line 61:
```suggestion
            writer.close()
            await writer.wait_closed()
```

yocalebo (changes requested):

src/middlewared/middlewared/plugins/device_/devd_events.py
```diff
@@ -51,26 +50,24 @@ def parse_devd_message(msg):
...
```
themylogin, line 61: Should we make a try/finally block that will close writer even if an exception occurs?

themylogin (changes requested):

src/middlewared/middlewared/plugins/device_/devd_events.py
rkojedzinszky: Enclosed the relevant code in try/finally block.

themylogin:
I approve this; however, I am in doubt whether this should be merged to the branch as stable as 13.0

themylogin (approved):

yocalebo:
@rkojedzinszky I'm just curious but was there a particular improvement or did this resolve a particular problem on your systems? While we appreciate the contribution, we're hesitant to merge anything into `truenas/13.0-stable` unless it's fixing something pretty serious.

rkojedzinszky:
> @rkojedzinszky I'm just curious but was there a particular improvement or did this resolve a particular problem on your systems? While we appreciate the contribution, we're hesitant to merge anything into `truenas/13.0-stable` unless it's fixing something pretty serious.

I was just reading code, and found that a write was missing, potentionally leaking an fd. While there, I thought to improve resource usage by running such simple code in the main asyncio loop rather than in a dedicated thread.

yocalebo:
@rkojedzinszky as always, thanks for the contribution. While this is definitely more efficient, we're going to forgo adding this to 13.0 since we don't like to change logic in our stable codebase for methods that are already working (and have been for years).

rkojedzinszky:
I understand. However a possible leak was also fixed.

yocalebo:
@rkojedzinszky you're right. I've opened a new PR with your changes (giving you credit of course) here: https://github.com/truenas/middleware/pull/10233

PR #10193, opened 2022-12-06
NAS-118887 / 23.10 / Properly report datastore name for services

Qubad786:
## Problem

When deleting certificates, sometimes it is not clear which service is consuming the certificate as the `CRUDService` has `check_dependencies` implementation which checks usages of the database row in question but for `SystemServiceService` children it only reports datastore names as it was not able to map the service name from datastore name which is not intuitive and user can not understand which service is actually consuming the certificate.

## Solution

`SystemServiceService` is special and we are not setting `datastore` variable for that but instead are relying on custom logic where we check for `service_model` or `service` and the name is made up dynamically.

Setting `datastore` in `Config` for `SystemServiceService` children handles the case and `check_dependencies` is able to report the services consuming the certificate accurately.

### DIFF
--- a/src/middlewared/middlewared/service.py
+++ b/src/middlewared/middlewared/service.py
@@ -356,8 +356,6 @@ class ServiceBase(type):
       - datastore_extend: datastore `extend` option used in common `query` method
       - datastore_prefix: datastore `prefix` option used in helper methods
       - service: system service `name` option used by `SystemServiceService`
-      - service_model: system service datastore model option used by `SystemServiceService`
-                       (`service` if used if not provided)
       - service_verb: verb to be used on update (default to `reload`)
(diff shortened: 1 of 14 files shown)


themylogin:
Why have `service_model` and `datastore` then if they are duplicating each other?

src/middlewared/middlewared/service.py
```diff
@@ -521,6 +521,9 @@ def __new__(cls, name, bases, attrs):
         ):
             return klass
 
+        if 'SystemServiceService' in [b.__name__ for b in bases] and not klass._config.datastore:
+            klass._config.datastore = f'services.{klass._config.service_model or klass._config.service}'
```
themylogin, line 525: Can we just get rid of `service_model` completely?

PR #10195, opened 2022-12-07
NAS-119178 / 13.0 / Use html2text in `format_alert`

themylogin:
No SCALE backport is needed, we have a more generic solution there


themylogin:
time 0:20

yocalebo (approved):

PR #10206, opened 2022-12-08
NAS-119298 / 23.10 / retry pool import on invalid cachefile

yocalebo:


sonicaj (commented):
I wonder if we should keep libzfs usages isolated to zfs plugin only ?

yocalebo:
> I wonder if we should keep libzfs usages isolated to zfs plugin only ?

You mean put the custom exception in the zfs plugin?

sonicaj:
Probably catch it in zfs plugin and raise appropriately from there ?

Motivation is to have py-libzfs usages in just one plugin only :)

sonicaj (approved):

PR #10210, opened 2022-12-08
NAS-119302 / 23.10 / `disk.sync_all_zfs_guid` for use in `vrrp_master`

themylogin:


src/middlewared/middlewared/plugins/disk_/zfs_guid.py
```diff
@@ -14,6 +14,20 @@ async def disk_by_zfs_guid(self, guid):
...
+    @private
+    async def sync_all_zfs_guid(self):
+        for pool in await self.middleware.call(
+            "zfs.pool.query",
+            [["name", "!=", await self.middleware.call("boot.pool_name")]],
+        ):
+            try:
+                await self.sync_zfs_guid({
+                    **pool,
+                    "topology": await self.middleware.call("pool.transform_topology", pool["groups"])
```
yocalebo, line 26: We should make the various topology transforming functions staticmethods because this is really painful on a large zpool. Our schema does too many deepcopies when we go through self.middleware and then inside sync_zfs_guid it then throws the pool groups object through self.middleware which does even more deepcopies etc etc. It'd be nice to make those functions staticmethods and then import them into various modules so we don't have to go through self.middleware

themylogin: We don't have `@schema` decorator for `transform_topology` so no copying is performed.

yocalebo (approved):

PR #10214, opened 2022-12-11
NAS-119241 / 23.10 / Add snapshot query tests.

bmeagherix:
These tests will query snapshots filtered by dataset name, and verify the returned data, exercising both simple (fast-path) and regular queries.

Also refactored to extract the functions `dataset`, `snapshot` and `snapshot_rollback` to facilitate their re-use.

### DIFF
--- a/tests/api2/assets/REST/pool.py
+++ b/tests/api2/assets/REST/pool.py
@@ -1,0 +1,24 @@
+import contextlib
+
+from functions import DELETE, GET, POST, PUT, wait_on_job
+
+
+@contextlib.contextmanager
+def dataset(pool_name, dataset_name):
+
+    dataset = f"{pool_name}/{dataset_name}"
+
+    payload = {
+        'name': dataset,
+    }
+    results = POST("/pool/dataset/", payload)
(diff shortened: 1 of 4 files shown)


tests/api2/assets/REST/pool.py
```diff
@@ -0,0 +1,24 @@
+import contextlib
+
+from functions import DELETE, GET, POST, PUT, wait_on_job
+
+
+@contextlib.contextmanager
+def dataset(pool_name, dataset_name):
```
anodos325, line 7: I think from standpoint of common test assets, it would be useful to allow specifying dataset properties as `options` kwarg and acl or mode as kwarg.

C.f.
https://github.com/truenas/middleware/blob/019654f9895901e48f02c528cca5fef66a32157d/tests/api2/test_300_nfs.py#L80

anodos325:
Test changes LGTM. I'll defer to Eric regarding details of REST assets (naming, etc).

anodos325 (approved):

yocalebo (approved):
Looks great!

ericbsd (approved):

PR #10215, opened 2022-12-12
NAS-118999 / 13.0 / Gracefully raise exception when deleting CA

Qubad786:
## Problem

If a CA has issued certificates and we try to delete the CA, users are greeted with a traceback where database does not allow deleting the certificate because of foreign key references.

## Solution

Neatly determine if CA has issued any certificates and raise a validation error before trying to delete it.

### DIFF
--- a/src/middlewared/middlewared/plugins/crypto.py
+++ b/src/middlewared/middlewared/plugins/crypto.py
@@ -2789,6 +2789,21 @@ class CertificateAuthorityService(CRUDService):
                     f'This certificate authority is in use by {text} service and cannot be deleted.'
                 )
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/crypto.py
```diff
@@ -2789,6 +2789,12 @@ async def do_delete(self, id):
                     f'This certificate authority is in use by {text} service and cannot be deleted.'
                 )
 
+        if certs := (await self.middleware.call('certificate.query', [['issuer.id', '=', id]])):
```
sonicaj, line 2792: We can also have intermediate CAs, so let's please add that check as well. Thanks!

sonicaj (approved):

PR #10219, opened 2022-12-12
NAS-119351 / 23.10 / Sanitse app errors from manual helm template catches

Ornias1993:
**Background**

With Helm templating, we can design custom faults and error messages, for example: If users enter the wrong input or settings cannot be combined.

With the current App-template Common rework by @stavros-k , the Apps will (likely) more heavily include these errors, to give as many user-friendly hints on how-to-fix common issues.

Hence it's crucial that those errors are displayed in a way that is as "not scary" as possible for the end-user

**Current Behavior**

Currently these would, for example, display like this:
```
[EFAULT] Failed to install chart release: Error: INSTALLATION FAILED: execution error at (plex/templates/deployment.yaml:1:3): Invalid hostpath /mnt/media_box. Path must be a valid path under a given pool e.g `/mnt/tank/somepath` is valid whereas `/mnt` or `/mnt/tank` are invalid examples.
```

This input is not clear (enough) and needlessly technical for the end user. 

**Wanted Behavior**
As these errors are generated with "user clarity" in mind. it would be prefered if the sentence here would be as clear as possible and leave any relevant technical details under `More info...`.

**Applied Changes**

To accomplice this this PR does:

- This rephrases `chart release` to `app`

- Removes `Error: INSTALLATION FAILED: execution error at (plex/templates/deployment.yaml:1:3): ` in above example. The format should be the same in other cases and is verified other types of helm errors.

- It does not change more than one (the first) match of the regex, so we don't by accident delete parts of custom errors for example.

- In case there is nothing to clean up, the original message is displayed unaltered

- In case the cleanup goes wrong, it does not touch the traceback available under `More info...`

- I've verified that this code portion is not covered by unittests yet.

**Potential other solutions**

1. Do so on the GUI side
   *pros:* This has the benefit of not altering the middleware and also being able to remove the `[EFAULT]` portion. 
   *cons:* It would require altering this in multiple places to counter all cases where something is upgraded/changed/installed and does not include GUI ran updates for example

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
@@ -18,7 +18,7 @@ class ChartReleaseService(Service):
     @private
     def helm_action(self, chart_release, chart_path, config, tn_action):
         args = ['-f']
-        CLEANUPREGEX = re.compile(r'^Error: .[)]: ')
+        CLEANUPREGEX = re.compile(r'^Error: .+[)]: ')
         
         if os.path.exists(os.path.join(chart_path, 'ix_values.yaml')):
             args.extend([os.path.join(chart_path, 'ix_values.yaml'), '-f'])
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
@@ -19,7 +19,7 @@ class ChartReleaseService(Service):
     def helm_action(self, chart_release, chart_path, config, tn_action):
         args = ['-f']
         CLEANUPREGEX = re.compile(r'^Error: .+[)]: ')
-        
+
         if os.path.exists(os.path.join(chart_path, 'ix_values.yaml')):
             args.extend([os.path.join(chart_path, 'ix_values.yaml'), '-f'])
 
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
@@ -2,6 +2,7 @@ import os
 import subprocess
 import tempfile
 import yaml
+import re
 
 from middlewared.plugins.kubernetes_linux.yaml import SafeDumper
 from middlewared.service import CallError, private, Service
@@ -41,6 +42,10 @@ class ChartReleaseService(Service):
             )
             stderr = cp.communicate()[1]
             if cp.returncode:
-                raise CallError(f'Failed to {tn_action} chart release: {stderr.decode()}')
+                errmsg = stderr.decode()
+                if tn_action == 'upgrade' or tn_action == 'install':
+                    cleanupregex = re.compile(r'^Error: .+?[)]: ')
+                    errmsg = re.sub(cleanupregex, '', errmsg, 1)
+                raise CallError(f'Failed to {tn_action} App: {errmsg}')
 
         self.middleware.call_sync('chart.release.clear_chart_release_portal_cache', chart_release)
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
@@ -18,6 +18,8 @@ class ChartReleaseService(Service):
     @private
     def helm_action(self, chart_release, chart_path, config, tn_action):
         args = ['-f']
+        CLEANUPREGEX = re.compile(r'^Error: .[)]: ')
+        
         if os.path.exists(os.path.join(chart_path, 'ix_values.yaml')):
             args.extend([os.path.join(chart_path, 'ix_values.yaml'), '-f'])
 
@@ -44,8 +46,7 @@ class ChartReleaseService(Service):
             if cp.returncode:
                 errmsg = stderr.decode()
                 if tn_action == 'upgrade' or tn_action == 'install':
-                    cleanupregex = re.compile(r'^Error: .[)]: ')
-                    errmsg = re.sub(cleanupregex, '', errmsg, 1)
+                    errmsg = re.sub(CLEANUPREGEX, '', errmsg, 1)
                 raise CallError(f'Failed to {tn_action} App: {errmsg}')
 
         self.middleware.call_sync('chart.release.clear_chart_release_portal_cache', chart_release)
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
@@ -45,8 +45,7 @@ class ChartReleaseService(Service):
             )
             stderr = cp.communicate()[1]
             if cp.returncode:
-                errmsg = stderr.decode()
-                errmsg = re.sub(RE_ERRCLEANUP, '', errmsg, 1)
+                errmsg = re.sub(RE_ERRCLEANUP, '',  stderr.decode(), 1)
                 raise CallError(f'Failed to {tn_action} App: {errmsg}')
 
         self.middleware.call_sync('chart.release.clear_chart_release_portal_cache', chart_release)
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
@@ -45,7 +45,7 @@ class ChartReleaseService(Service):
             )
             stderr = cp.communicate()[1]
             if cp.returncode:
-                errmsg = re.sub(RE_ERRCLEANUP, '',  stderr.decode(), 1)
+                errmsg = re.sub(RE_ERRCLEANUP, '', stderr.decode(), 1)
                 raise CallError(f'Failed to {tn_action} App: {errmsg}')
 
         self.middleware.call_sync('chart.release.clear_chart_release_portal_cache', chart_release)
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
@@ -9,6 +9,8 @@ from middlewared.service import CallError, private, Service
 
 from .utils import get_namespace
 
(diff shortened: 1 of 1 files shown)


Ornias1993:
@Qubad786 and @sonicaj This is your "department" as discussed.

Ornias1993:
@yocalebo thanks! :)

src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
```diff
@@ -41,6 +42,10 @@ def helm_action(self, chart_release, chart_path, config, tn_action):
             )
             stderr = cp.communicate()[1]
             if cp.returncode:
-                raise CallError(f'Failed to {tn_action} chart release: {stderr.decode()}')
+                errmsg = stderr.decode()
+                if tn_action == 'upgrade' or tn_action == 'install':
```
Qubad786, line 46: Is this conditional required?

src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
```diff
@@ -41,6 +42,10 @@ def helm_action(self, chart_release, chart_path, config, tn_action):
...
+                    cleanupregex = re.compile(r'^Error: .+?[)]: ')
+                    errmsg = re.sub(cleanupregex, '', errmsg, 1)
```
Qubad786, line 48: You can do this like `re.sub(r'^Error: .+?[)]: ', '', s, 1)`

Also i believe the regex you are looking at is `r'^Error: .+[)]?: '` (with question mark at the end)

src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
Ornias1993: We, @stavros-k and I, wanted to ensure it only gets used for user-facing helm 'commands' (via the middleware).
So we wanted to ensure it doesn't, by accident, gets used to sanitise other errors as well.

src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
Ornias1993: @stavros-k noted that the use of re- compile, was a practice used by iX in all other cases where `re.sub` was used, so we picked that for consistency.

stavros-k: Regarding `+?`:
image 

But `[)]?` It will capture also this:
`Error: INSTALLATION FAILED: execution error at (plex/templates/deployment.yaml:1:3): For example do this and that (carefully): ` 
And remove it.

TL;DR `?` can be removed and will still work

Qubad786: @stavros-k either ways having it after `+` does not make sense?
Using it means safely handling closing bracket that it might not exist sometime although I assume this is always going to exist as otherwise having the colon after this does not make sense so let's just remove it in that case then.

Qubad786: @Ornias1993 compiling is better but then it should be done only once after the imports by declaring it like this https://github.com/truenas/middleware/blob/c80c7bc7e091c17c0252d6c34c8abbfc18669bfb/src/middlewared/middlewared/plugins/iscsi_/iscsi_global.py#L10.
This way the regex is only compiled once and not each time the function is called..
Then you can do `RE_VARNAME.sub()`

stavros-k: > @stavros-k either ways having it after `+` does not make sense? Using it means safely handling closing bracket that it might not exist sometime although I assume this is always going to exist as otherwise having the colon after this does not make sense so let's just remove it in that case then.

The only different is that it will stop "parsing" a bit faster in this case, If I'm reading in correctly.
But for the sake of readability, and not having someone scratch his header down the road, and considering the parse time is not that HUGE to make a difference.

Let's remove it.

src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
```diff
@@ -41,6 +42,10 @@ def helm_action(self, chart_release, chart_path, config, tn_action):
             )
             stderr = cp.communicate()[1]
             if cp.returncode:
-                raise CallError(f'Failed to {tn_action} chart release: {stderr.decode()}')
+                errmsg = stderr.decode()
+                if tn_action == 'upgrade' or tn_action == 'install':
+                    cleanupregex = re.compile(r'^Error: .+?[)]: ')
```
stavros-k, line 47:
```suggestion
                    cleanupregex = re.compile(r'^Error: .+[)]: ')
```

src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
Ornias1993: DOne deal!

src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
```diff
@@ -41,6 +42,10 @@ def helm_action(self, chart_release, chart_path, config, tn_action):
             )
             stderr = cp.communicate()[1]
             if cp.returncode:
-                raise CallError(f'Failed to {tn_action} chart release: {stderr.decode()}')
+                errmsg = stderr.decode()
+                if tn_action == 'upgrade' or tn_action == 'install':
+                    cleanupregex = re.compile(r'^Error: .+[)]: ')
```
Ornias1993, line 47:
```suggestion
                    cleanupregex = re.compile(r'^Error: .[)]: ')
```

src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
```diff
@@ -41,6 +42,10 @@ def helm_action(self, chart_release, chart_path, config, tn_action):
             )
             stderr = cp.communicate()[1]
             if cp.returncode:
-                raise CallError(f'Failed to {tn_action} chart release: {stderr.decode()}')
+                errmsg = stderr.decode()
+                if tn_action == 'upgrade' or tn_action == 'install':
+                    cleanupregex = re.compile(r'^Error: .[)]: ')
```
Ornias1993:
```suggestion
                    cleanupregex = re.compile(r'^Error: .+[)]: ')
```

src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
```diff
@@ -17,6 +18,8 @@ class Config:
     @private
     def helm_action(self, chart_release, chart_path, config, tn_action):
         args = ['-f']
+        CLEANUPREGEX = re.compile(r'^Error: .[)]: ')
```
Ornias1993, line 21:
```suggestion
        CLEANUPREGEX = re.compile(r'^Error: .+[)]: ')
```

src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
sonicaj: All of this function's usages are user facing, so let's just remove the conditional please as all of those actions can benefit from the standardisation

src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
```diff
@@ -17,6 +18,8 @@ class Config:
     @private
     def helm_action(self, chart_release, chart_path, config, tn_action):
         args = ['-f']
+        CLEANUPREGEX = re.compile(r'^Error: .+[)]: ')
```
sonicaj, line 21: Please take the declaration of this variable to line 12 and then add a couple of new lines for flake8. Also we have an unspoken norm for naming scheme like `RE_CLEANUP` or something like that - which is basically prefixing `RE_` with regex compiles done.

Ornias1993: @sonicaj The condition is there, because it is only working (correctly) for a single(!) specific type of errors.
It is not, and will not be, working (correctly) for others.

You're basically pushing data it's not, validated to be, compatible with, through this regex.
Are you 100% sure that you want to be doing that?

Ornias1993: The other issue (constant defining at top of the file) has been fixed as `RE_ERRCLEANUP`

sonicaj: > @sonicaj The condition is there, because it is only working (correctly) for a single(!) specific type of errors. It is not, and will not be, working (correctly) for others.
> 
> You're basically pushing data it's not, validated to be, compatible with, through this regex. Are you 100% sure that you want to be doing that?

@Ornias1993 I don't follow ? We only have 3 possible values for tn actions i.e install/upgrade/update and for helm that just translates to upgrade/install which means that there is no other case ?

Ornias1993: As it's technically a general Helm-Call function, I personally find it nicer to assume it will be used for more than those two commands.

But As you want it changed i'll remove the if :)

src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
```diff
@@ -41,6 +45,9 @@ def helm_action(self, chart_release, chart_path, config, tn_action):
             )
             stderr = cp.communicate()[1]
             if cp.returncode:
-                raise CallError(f'Failed to {tn_action} chart release: {stderr.decode()}')
+                errmsg = stderr.decode()
+                if tn_action == 'upgrade' or tn_action == 'install':
+                    errmsg = re.sub(RE_ERRCLEANUP, '', errmsg, 1)
```
Ornias1993, line 50:
```suggestion
                errmsg = re.sub(RE_ERRCLEANUP, '', errmsg, 1)
```

src/middlewared/middlewared/plugins/chart_releases_linux/helm.py
Ornias1993: Resolved :)

sonicaj (approved):
Thank you @Ornias1993

PR #10224, opened 2022-12-13
NAS-119356 / 23.10 / Avoid python DeprecationWarning wrt import of Callable.

bmeagherix:
Modify import to avoid "DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is
deprecated since Python 3.3, and in 3.10 it will stop working"

Example context:
```
api2/test_480_system_advanced.py:60: AssertionError
=============================== warnings summary ===============================
../env/lib/python3.9/site-packages/middlewared-0.0.0-py3.9.egg/middlewared/client/client.py:4
  /root/jenkins/workspace/TrueNAS_SCALE_-_Unstable/Test_development/API_Tests_-_TrueNAS_SCALE_from_branch_-_Incremental/env/lib/python3.9/site-packages/middlewared-0.0.0-py3.9.egg/middlewared/client/client.py:4: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
    from collections import defaultdict, namedtuple, Callable
```

Used `try ... except ImportError` to preserve existing behavior in case this code is also still being used somewhere with older python.

Also used [isort](https://pycqa.github.io/isort/index.html) to clean up some import ordering.

### DIFF
--- a/src/middlewared/middlewared/client/client.py
+++ b/src/middlewared/middlewared/client/client.py
@@ -1,24 +1,31 @@
+from collections import defaultdict, namedtuple
+
 from . import ejson as json
 from .protocol import DDPProtocol
-from .utils import MIDDLEWARE_RUN_DIR, undefined, ProgressBar
-from collections import defaultdict, namedtuple, Callable
-from threading import Event as TEvent, Lock, Thread
-from ws4py.client.threadedclient import WebSocketClient
+from .utils import MIDDLEWARE_RUN_DIR, ProgressBar, undefined
+
+try:
+    from collections.abc import Callable
+except ImportError:
+    from collections import Callable
 
 import argparse
-from base64 import b64decode
 import errno
 import logging
 import os
 import pickle
+import platform
 import pprint
+import random
 import socket
 import sys
 import time
 import uuid
-import random
-import platform
+from base64 import b64decode
+from threading import Event as TEvent
+from threading import Lock, Thread
 
+from ws4py.client.threadedclient import WebSocketClient
 
 try:
     from libzfs import Error as ZFSError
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/client/client.py
```diff
@@ -1,24 +1,31 @@
+from collections import defaultdict, namedtuple
```
yocalebo, line 1: Since you're already here, can we fix the imports in this file? We _try_ to follow the common paradigm of 

1. python stdlib imports go first
--- (import blah) goes at top
--- (from blah import blah) goes next
--- use any `try: except ImportError` last
2. 3rd party modules get imported next with 1 space between the stdlib imports

3. our modules get imported with 1 space between the 3rd party modules

For example
```
import time
import os
from collections import defaultdict
try:
    from collections.abc import Callable
except ImportError:
    from collections import Callable

import psutil
from pyroute2 import Context

from middlewared.service import CallError
from . import ejson as json

**Review by yocalebo — CHANGES_REQUESTED** [review]

File: src/middlewared/middlewared/client/client.py (hunk 1 above)
Comment (bmeagherix) [reply src/middlewared/middlewared/client/client.py]: Thanks, I wasn't aware of the convention.

FWIW, I was able to get _close_ using isort with some customization:
```
$ sections=FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
$ no_lines_before=FUTURE,STDLIB,LOCALFOLDER
$ isort client.py
```
Then removed some unwanted newlines.

PR #10225, opened 2022-12-13
NAS-119357 / 23.10 / Fix regression when setting different quota types on same id(s)

yocalebo:
Fix setting (for example) a USER and USEROBJ quota for the same uid. We were creating a dictionary object with the key being the `xid` which means the previous entry in the dictionary would be overwritten with whatever the last entry was in the for loop.

Fixes implemented here:
- simplify the for loop by ignoring `PROJECT` and `PROJECTOBJ` quota types since those aren't implemented
- remove unnecessary newlines
- changes `quotas` object to a list of dictionaries so we don't overwrite previous `xid` entries
- improve the integration tests to (ideally) catch this from happening again in the future

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/dataset_quota_and_perms.py
+++ b/src/middlewared/middlewared/plugins/pool_/dataset_quota_and_perms.py
@@ -295,7 +295,7 @@ class PoolDatasetService(Service):
                     verrors.add(f'quotas.{i}.id', 'id for quota_type DATASET must be either "QUOTA" or "REFQUOTA"')
                 else:
                     xid = q['id'].lower()
-                    if xid in quotas:
+                    if any((i.get(xid, False) for i in quotas)):
                         verrors.add(
                             f'quotas.{i}.id',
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/pool_/dataset_quota_and_perms.py
```diff
@@ -280,66 +280,49 @@ async def set_quota(self, ds, data):
...
-                xid = q['id'].lower()
-                if xid in quotas:
-                    verrors.add(
-                        f'quotas.{i}.id',
-                        f'Setting multiple values for {xid} for quota_type "DATASET" is not permitted'
-                    )
-                    continue
-
-            elif q['quota_type'] not in ['PROJECT', 'PROJECTOBJ']:
+                if q['id'] not in ('QUOTA', 'REFQUOTA'):
```
anodos325, line 294: Aren't those quota types?

anodos325 (approved):

yocalebo:
build failure is unrelated to my changes

PR #10227, opened 2022-12-13
Fix cluster path handling in attachment delegates

anodos325:
Plugins containing shares have a special method to retrieve the path underlying the share. This is because in some case the path may be a composite (for instance SMB share path on a glusterfs volume is relative to the root of the volume rather than the root of the local filesystme). In this case self.path_field() returns a string containing the special CLUSTER prefix.


src/middlewared/middlewared/plugins/pool_/dataset_attachments.py
```diff
@@ -36,6 +36,9 @@ async def attachments(self, oid):
     @private
     async def attachments_with_path(self, path):
```
yocalebo, line 37: Is `path` here always a string type? If not, then `path.startswith` will crash.

src/middlewared/middlewared/common/attachment/__init__.py
```diff
@@ -130,7 +130,8 @@ async def remove_alert(self, attachment):
         await self.middleware.call(f'{self.namespace}.remove_locked_alert', attachment['id'])
 
     async def is_child_of_path(self, resource, path):
-        return is_child(resource[self.path_field], path)
+        share_path = await self.service_class.get_path_field(self.service_class, resource)
```
yocalebo, line 133: Maybe I'm blind but `get_path_field` only takes 1 argument?

yocalebo (approved):

PR #10230, opened 2022-12-14
NAS-119339 / 23.10 / Do not allow changing PCI device for active VM

Qubad786:
## Problem

2 problems were identified with the PCI device logic:

1. A running VM can have it's PCI device updated in database which will mean we won't be able to safely re-attach the old device to host.
2.  When a VM is stopped, we already handle the case where the PCI device is re-attached to host but we were also trying the same on PCI device deletion which is not required and doing it unconditionally there. It should be fine generally but for one user who had a critical PCI device passthrough done - re-attaching the device bricked the system..There have been changes since then already where we disallow passthrough for critical PCI devices.

## Solution

Validation has been added to not allow changing PCI device when a VM is in an active state and re-attaching logic on PCI device deletion has been removed.


sonicaj (approved):

Qubad786:
backport

PR #10234, opened 2022-12-14
NAS-119335 / 23.10 / Add validation for host path fields in apps

Qubad786:
## Context

Currently we have validation in CRI where we don't let containers deploy which are consuming sensitive paths..
However the feedback loop is not intuitive and users can't quickly see what is wrong with their path before clicking save on the update/install form of the app and later have to check app events to figure out what might have happened.

## Solution

Port validation logic from CRI to middleware and try to validate host path fields in middleware. This is not bullet proof as if some app developer did not conform to standard practices, then CRI will still not let the container be deployed but in most cases app devs use `hostpath` field in questions schema which in turn will make sure that middleware validates the value specified.


src/middlewared/middlewared/plugins/chart_releases_linux/host_path_attachments.py
```diff
@@ -1,6 +1,73 @@
...
+        if in_use_attachments:
+            return f'Invalid mount {path!r} {path_type}. Following service(s) use this ' \
+                   f'path: {", ".join(attachment["type"] for attachment in in_use_attachments)}'
+
+    @private
+    async def validate_host_source_path(self, path):
+        paths = {
+            'path': path,
+        }
+        real_path = os.path.realpath(path)
```
anodos325, line 57: This is blockign

sonicaj (approved):
LGTM, however we would need to remove this validation from containerd now as well ( once the changes have been finalized there ) so we can have all that logic here only and call this endpoint from containerd.

Ornias1993 (commented):
Handling errors in the middleware is a good solution to throw out errors earlier.

However, it would be even more helpfull to the end-user, to also give the a red warning (like how certain values are required) in the UI when a known "bad path" is used. As this is even more visual for the end-user.

Also note:
It's important though, to also keep this on the CRI level.
Keep it there is crucial to prevent container-escape using hostPaths, which can be done if this is ever removed from the CRI, as those escapes would bypass the middleware.

So in TLDR:
CRI, Middleware and UI errors all serve a purpose and all should be implemented for the best user experience.

yocalebo:
retest this please

Ornias1993:
> LGTM, however we would need to remove this validation from containerd now as well ( once the changes have been finalized there ) so we can have all that logic here only and call this endpoint from containerd.

You need to be carefull with this, as you also want validation to prevent container escapes through kubernetes directly!

src/middlewared/middlewared/plugins/chart_releases_linux/host_path_attachments.py
```diff
@@ -1,6 +1,73 @@
...
+    class Config:
+        namespace = 'chart.release'
+
+    @private
+    async def ix_mount_validation(self, path, path_type):
+        path_list = [p for p in path.split('/') if p.strip()]
+        if safe_to_ignore_path(path):
+            if path_list and len(path_list) < 3 and path_list == 'mnt':
+                return f'Invalid path {path}. Mounting root dataset or path outside a pool is not allowed'
+        elif path_list and path_list[0] == 'cluster':
```
anodos325, line 21: Why are we using the fuse mount via host path instead of something like glusterfs CSI?

There is no protection for ctdb_shared_vol (which is an internal volume that should _never under any circumstances_ be accessed by any app).

src/middlewared/middlewared/plugins/chart_releases_linux/host_path_attachments.py
```diff
@@ -1,6 +1,73 @@
...
+        }
+        real_path = await self.middleware.run_in_thread(os.path.realpath, path)
+        if real_path != path:
+            paths[f'path (real path of {path})'] = real_path
+
+        for path_type, path_to_test in paths.items():
+            if err := await self.ix_mount_validation(path_to_test, path_type):
+                return err
+
+            if path_to_test.startswith('/mnt/'):
```
anodos325, line 65: Since we're apparently allowing cluster volumes, you need to validate whether SMB share is attached to it.

src/middlewared/middlewared/plugins/chart_releases_linux/host_path_attachments.py
anodos325: For that matter, we probably want to discourage users from exposing their entire glusterfs volume via host path.

`/cluster/glvol/` is probably a bad idea because it removes the ability of admin to be able to pick / choose which data in volume to expose to kubernetes. `/cluster/gvol/k8s` is a safer bet.

anodos325: There are some plans to move contents of ctdb_shared_vol to special dir inside root of first created gluster volume. This means actually we _must_ prevent exposing `/cluster/glvol` to apps.

Ornias1993: > Why are we using the fuse mount via host path instead of something like glusterfs CSI?
> 
> There is no protection for ctdb_shared_vol (which is an internal volume that should _never under any circumstances_ be accessed by any app).

I agree here, there are a number of ways to mount a gluster cluster directly using a CSI and those also integrate better in potential future usecases for actually clusterised kubernetes.

yocalebo (changes requested):
hold off on merge

src/middlewared/middlewared/plugins/chart_releases_linux/utils.py
```diff
@@ -62,6 +62,42 @@ def get_chart_release_from_namespace(namespace):
...
+def safe_to_ignore_path(path: str) -> bool:
+    # "/" and "/home/keys/" are added for openebs use only, regular containers can't mount "/" as we have validation
+    # already in place by docker elsewhere to prevent that from happening
+    if path == '/':
+        return True
+
+    for ignore_path in (
+        '/etc/',
+        '/sys/',
+        '/proc/',
```
anodos325, line 74: I think we need to be quite careful about whether we allow /proc access. It is very easily abused.

src/middlewared/middlewared/plugins/chart_releases_linux/utils.py
```diff
@@ -62,6 +62,42 @@ def get_chart_release_from_namespace(namespace):
...
+        '/var/lib/kubelet/',
+        '/dev/',
+        '/mnt/',
+        '/home/keys/',
+        '/run/',
+        '/var/run/',
```
anodos325, line 80: Ditto for /run and /var/run

src/middlewared/middlewared/plugins/chart_releases_linux/utils.py
Ornias1993: I think this is just a copy-paste from the current validator.

However, you're 100% correct... The validation should (aim to) prevent both user-error and potential container escapes (which is a known issue with hostPaths on kubernetes)

I don't see a usecases where it's really required to have host-level access here.

src/middlewared/middlewared/plugins/chart_releases_linux/utils.py
Ornias1993: - Correct: at least `/var/run` should only be needed in containers within their local context, so shouldn't be fully whitelisted.
- I'm going to check how loki and other container-log fetchers behave with these mountpaths, i'll get back to you shortly!

Ornias1993: Checked, promtail/loki (One of the Apps that requires the biggest host file access of all our Apps) requires:
`/var/log/pods` and `/run/promtail`

The later can be problematic, I think it's best to check weither this `/run/promtail` access thing is a forced requirement though...

yocalebo (approved):

Qubad786:
backport

PR #10235, opened 2022-12-14
NAS-119206 / 23.10 / In NFSModel replace nfs_srv_v4 with nfs_srv_protocols.

bmeagherix:
Previously `nfs_srv_v4` (a boolean) determined whether NFSv4 would be active in addition to NFSv3.

`nfs_srv_protocols` (a list of strings) will instead allow any non-empty combination of "NFSV3" and "NFSV4".


src/middlewared/middlewared/alembic/versions/22.12/2022-12-14_00-45_nfs_protocols.py
```diff
@@ -0,0 +1,45 @@
+"""empty message
```
anodos325, line 1: Let's put a message here

src/middlewared/middlewared/plugins/nfs_/debug.py
```diff
@@ -39,7 +39,7 @@ def get_debug(self):
                 if not svc.endswith("debug"):
                     continue
 
-                with open(f"proc/sys/sunrpc/{svc}", "r") as f:
+                with open(f"/proc/sys/sunrpc/{svc}", "r") as f:
```
anodos325, line 42: Good catch. Whoops!

src/middlewared/middlewared/alembic/versions/22.12/2022-12-14_00-45_nfs_protocols.py
bmeagherix: Does "Replace v4 with protocols in nfs service configuration." seem sufficient?

src/middlewared/middlewared/alembic/versions/23.10/2022-12-14_00-45_nfs_protocols.py
```diff
@@ -0,0 +1,45 @@
...
+    with op.batch_alter_table('services_nfs', schema=None) as batch_op:
+        if share_count == 0 or nfs_srv_v4:
+            batch_op.add_column(sa.Column('nfs_srv_protocols', sa.TEXT(), nullable=False, server_default='["NFSV3", "NFSV4"]'))
+        else:
+            batch_op.add_column(sa.Column('nfs_srv_protocols', sa.TEXT(), nullable=False, server_default='["NFSV3"]'))
+        batch_op.drop_column('nfs_srv_v4')
+
+
+def downgrade():
+    conn = op.get_bind()
```
themylogin, line 38: @bmeagherix for the future: we don't do downgrades so this function can be ignored

themylogin (approved):
Great job! I changed `down_revision` (we merged a new migration meanwhile) and moved new migration to a new directory.

anodos325 (approved):

ericbsd (approved):

PR #10236, opened 2022-12-15
NAS-119399 / 23.10 / Fix "KeyError: 'remove_certificates' not found" exception.

Jah-On:
This error occurred when trying to download a client config after selecting the client certificate. The current code does not check if the key exists before popping thus throwing an error when the 'remove_certificates' key is not found.

### DIFF
--- a/src/middlewared/middlewared/plugins/vpn.py
+++ b/src/middlewared/middlewared/plugins/vpn.py
@@ -196,7 +196,8 @@ class OpenVPN:
                 'Please specify a valid authentication_algorithm.'
             )
 
-        if data.pop('remove_certificates'):
+        if 'remove_certificates' in data:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/vpn.py
```diff
@@ -196,7 +196,8 @@ async def common_validation(middleware, data, schema, mode):
                 'Please specify a valid authentication_algorithm.'
             )
 
-        if data.pop('remove_certificates'):
+        if 'remove_certificates' in data:
```
sonicaj, line 199: You can just do `data.pop('remove_certificates', None)`

sonicaj:
Thank you @Jah-On for your contribution

yocalebo (approved):

PR #10241, opened 2022-12-15
NAS-119380 / 23.10 / Do not treat `legacy` as mountpoint

themylogin:


themylogin:
time 0:10

yocalebo:
Should we really crash here? This will break `pool.dataset.unlock` when `toggle_attachments` is set.

themylogin:
@yocalebo why will it break? We don't have legacy mountpoints anywhere besides boot-pool which is not affected

Ornias1993 (commented):
I'm not 100% sure if this would trigger this codepath, but a LOT of App related paths are, in fact, legacy....

src/middlewared/middlewared/plugins/pool_/dataset_processes.py
```diff
@@ -41,9 +41,12 @@ async def processes(self, oid):
         dataset = await self.middleware.call('pool.dataset.get_instance_quick', oid, {'encryption': True})
         if dataset['locked']:
             return []
-        path = attachments_path(dataset)
-        zvol_path = f'/dev/zvol/{dataset["name"]}'
-        return await self.middleware.call('pool.dataset.processes_using_paths', [path, zvol_path])
+
+        paths = [f'/dev/zvol/{dataset["name"]}']
```
anodos325, line 45: Hmm.... is this a place where we missed conversion to using `zvol_name_to_path()` ? 

https://github.com/truenas/middleware/blob/master/src/middlewared/middlewared/plugins/zfs_/utils.py#L19

themylogin: Great catch! We actually had a few places like this

anodos325 (approved):

PR #10247, opened 2022-12-15
NAS-109044 / 23.10 / Phase out docker from middleware

sonicaj:
This PR phases out docker from SCALE and we now move in with containerd. Following are some notable changes which have been introduced:

1. NVIDIA gpu support has been added but now app devs need to use a `RuntimeClass` in their apps apart from the resources they already specify. This will be present already in the `iXChartContext` we provide on apps so app devs can consume that.
2. There are some bits which cannot be retrieved now like `labels` or `created` attrs for containerd images and the return type has been updated appropriately.


src/middlewared/middlewared/etc_files/rancher/k3s/flags.py
```diff
@@ -50,3 +51,10 @@ def render(service, middleware):
             'protect-kernel-defaults': True,
             'disable': [] if config['servicelb'] else ['servicelb'],
         }))
+
+    # TODO: See what we have to do make proxy work for pulling images
```
Ornias1993, line 55: Seems to be doable with an env-var:

https://github.com/rancher/rke2/issues/517

src/middlewared/middlewared/plugins/container_runtime_interface/images.py
```diff
@@ -117,66 +115,34 @@ async def pull(self, job, data):
...
-        await self.docker_checks()
-        # TODO: Have job progress report downloading progress
-        async with aiodocker.Docker() as docker:
+        self.middleware.call_sync('kubernetes.validate_k8s_setup')
+        with ContainerdClient('image') as client:
             try:
-                response = await docker.images.pull(
-                    from_image=data['from_image'], tag=data['tag'], auth=data['docker_authentication']
+                client.pull_image(
+                    f'{data["from_image"]}:{data["tag"]}', auth_config=data['docker_authentication'],
```
Ornias1993, line 122: Docker is basically "dying", it might be best to start using non-branded terminology for things like these internally at least.

src/middlewared/middlewared/plugins/container_runtime_interface/images.py
```diff
@@ -196,18 +162,17 @@ def normalise_tag(self, tag):
 
     @private
     def get_system_images_tags(self):
-        with open(DEFAULT_DOCKER_IMAGES_LIST_PATH, 'r') as f:
-            images = [i for i in map(str.strip, f.readlines()) if i]
+        # TODO: Let's add coredns/pause image etc
 
-        images.extend([
+        images = [
             'nvidia/k8s-device-plugin:1.0.0-beta6',
```
Ornias1993, line 168: This should be moved to a non-docker host, as this might hit docker rate-limits.

(OpenEBS is excluded from dockerhub rate limits, this one is not)

src/middlewared/middlewared/plugins/container_runtime_interface/prune.py
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+    ))
+    @job(lock='container_prune')
+    def prune(self, job, options):
+        """
+        Prune unused images/containers. This will by default remove any dangling images.
+
+        `prune_options.remove_unused_images` when set will remove any docker image which is not being used by any
+        container.
```
Ornias1993, line 41:
```suggestion
        `prune_options.remove_unused_images` when set will remove any container image which is not being used by any
        container.
```

Ornias1993 (commented):
Some small comments after a quick glance, nice work though...

sonicaj:
Unit test CI is failing because CRI python package is not available and is being added as a dependency in this PR

src/middlewared/middlewared/alert/source/cores.py
```diff
@@ -24,12 +24,6 @@ class CoreFilesArePresentAlertSource(AlertSource):
         "/usr/sbin/smartctl",
     )
     ignore_units = (
-        # Unit: "containerd.service"/"docker.service" is related to k3s.
-        # users are free to run whatever they would like to in containers
-        # and we don't officially support all the apps themselves so we
-        # ignore those core dumps
-        "containerd.service",
-        "docker.service",
```
anodos325, line 32: Do we have some other heuristic we're using to determine whether a corefile is generated by an app instead of base system?

src/middlewared/middlewared/etc_files/rancher/k3s/flags.py
```diff
@@ -50,3 +51,9 @@ def render(service, middleware):
             'protect-kernel-defaults': True,
             'disable': [] if config['servicelb'] else ['servicelb'],
         }))
+
+    with open('/etc/containerd.json', 'w') as f:
```
anodos325, line 55: Out of curiosity, what consumes this file?

sonicaj: Containerd consumes this file where we tell it to validate host paths or not

src/middlewared/middlewared/alert/source/cores.py
sonicaj: Containers deployed via containerd have core dumps disabled with their soft/hard limits set to 1 so this is not required anymore.

src/middlewared/middlewared/plugins/container_runtime_interface/images.py
```diff
@@ -35,7 +31,6 @@ class Config:
         Bool('dangling'),
         Bool('update_available'),
         Bool('system_image'),
-        Datetime('created', null=True),
```
Qubad786, line 38: `labels` are not available as well..

src/middlewared/middlewared/plugins/container_runtime_interface/images.py
```diff
@@ -47,44 +42,38 @@ class Config:
...
-                    'created': created_at,
+                    'repo_digests': image.get('repoDigests') or [],
+                    'size': int(image['size']),
                     'dangling': len(repo_tags) == 1 and repo_tags[0] == '<none>:<none>',
                     'update_available': not system_image and any(update_cache[r] for r in repo_tags),
                     'system_image': system_image,
                 }
                 if parse_tags:
-                    result['parsed_repo_tags'] = await self.middleware.call('container.image.parse_tags', repo_tags)
+                    result['parsed_repo_tags'] = self.middleware.call_sync('container.image.parse_tags', repo_tags)
```
Qubad786, line 76: We can use the util directly here instead of calling middleware?

Qubad786 (commented):
Excellencio

Qubad786 (approved):

PR #10248, opened 2022-12-16
NAS-119374 / 22.12.1 / Allow sudo-enabled users to run virsh/kubectl in web shell with root privileges

themylogin:


themylogin:
@yocalebo we can also add an explicit "allow virsh" and "allow kubectl" options to the privilege model to bypass this check.

src/middlewared/middlewared/main.py
```diff
@@ -550,36 +552,44 @@ class ShellWorkerThread(threading.Thread):
...
 
         if options.get('vm_id'):
-            return [
-                '/usr/bin/sudo', '-H', '-u', username,
+            command = [
                 '/usr/bin/virsh', '-c', 'qemu+unix:///system?socket=/run/truenas_libvirt/libvirt-sock',
                 'console', f'{options["vm_data"]["id"]}_{options["vm_data"]["name"]}'
             ]
+
+            if not can_sudo:
```
yocalebo, line 576: The logic seems off to me? Should it not be
```
if can_sudo:
    command = ['/usr/bin/sudo'
```

themylogin:
@yocalebo the logic is correct. If user can sudo then we run this command as root (because they have the permission to do that anyway). If not then we run the command as user and hope they have necessary permissions.

yocalebo:
> @yocalebo the logic is correct. If user can sudo then we run this command as root (because they have the permission to do that anyway). If not then we run the command as user and hope they have necessary permissions.

That's terribly confusing to name it `can_sudo`. Can we inverse the boolean value and name it `try_sudo` instead? So that way it doesn't confuse everyone who reads it?

anodos325:
Maybe `as_root` would be clearer because what we're really doing when it's True is we're running the shell as root.

yocalebo (approved):
thanks!

themylogin:
time 2:00

PR #10252, opened 2022-12-16
NAS-119251 / 23.10 / * Do not set ZFS guid for expired disks with the same name

themylogin:
* When retrieving disk by ZFS guid prioritize present disks over expired


themylogin:
time 14:00

yocalebo (approved):
Great job!

PR #10261, opened 2022-12-18
NAS-119398 / 23.10 / Fix backup/restore of k8s

Qubad786:
## Problem

With kubernetes api client re-write a regression was introduced where upstream kubernetes asyncio package was normalizing fields to snake case which kubernetes API provides as camel case and we had some leftover usages uncovered. This resulted in backups not storing persistent volume state.

## Solution

Cover the snake case usages to correctly retrieve them from the new k8s api and on restore make sure we handle both snake/camel case correctly.


Ornias1993 (commented):
Confirmed to be creating correctly filled `pv_info.json` when applied as a direct BlueFin hotpatch :)

sonicaj (approved):

sonicaj:
Also i think we should have integration tests for this, going to create a ticket for that

Ornias1993:
> Also i think we should have integration tests for this, going to create a ticket for that

Thanks! :)

PR #10066, opened 2022-11-04
NAS-114479 / 23.10 / Shrink swap when formatting a disk in `pool.replace` to fit a data pa…

themylogin:
…rtiton of required size

PR #10267, opened 2022-12-19
NAS-119473 / 23.10 / Use pyparted for `disk.format`

themylogin:
Two concerns:
* Exact partition sizes do not match with this approach (on 8GB HDD swap gets 900k smaller and data partition gets 900k larger), I guess it's alright?
* With incorrect arguments middleware will now SEGFAULT


themylogin:
time 5:00

yocalebo (changes requested):
`ZFS_PARTHEX` and `SWAP_PARTHEX` can be removed from what I can see

src/middlewared/middlewared/plugins/disk_/format.py
```diff
@@ -28,46 +25,58 @@ def validate_disk(self, disk, swapgb):
...
         # TODO: Install a dummy boot block so system gives meaningful message if booting from a zpool data disk.
 
-        for cmd in commands:
-            cp = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
-            if cp.returncode != 0:
-                raise CallError(f'Unable to GPT format the disk "{disk}": {cp.stderr}')
-
         self.middleware.call_sync('device.settle_udev_events')
 
         for partition in self.middleware.call_sync('disk.list_partitions', disk):
```
yocalebo, line 79: running `disk.list_partitons` for every disk and then calling `zfs.pool.clear_label` is unnecessary because `disk.wipe, QUICK` wipes the first and last 32MB of the disk.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/disk_/format.py
themylogin: Why is this code here then?
And what happens if we had a swap + partition; in that case, `disk.wipe` will only be able to erase partition end, but won't affect partition start?

yocalebo: We wipe the first and last 32MB of a disk regardless of what is on the disk in `disk.wipe`? Also, when I was investigating this, `device.clobber()` removes partition tables off the disk so I'm not even sure we need the call to `disk.wipe`. I can't remember if `device.clobber()` removed table entries at end of disk.

themylogin: @sonicaj you introduced this in e0d120f4284d139c989b8776452617a57272c79f - why?

themylogin: https://ixsystems.atlassian.net/browse/NAS-103069 that was the original issue @yocalebo

yocalebo: I understand but that change was made before we wiped the first and last 32MB of a disk IIRC which means the zfs partition label will always be removed.

Also, have you confirmed that the `clobber()` method removes all partition labels off the disk?

yocalebo:
@themylogin can you fix the conflicts?

themylogin:
time 7:00

themylogin:
@yocalebo I had to rewrite it almost from scratch to adapt to https://github.com/truenas/middleware/pull/10066 :(

themylogin:
time 4:00

PR #10269, opened 2022-12-19
NAS-117470 / 23.10 / Query formatted disks gptids after mass formatting to prevent race co…

themylogin:


themylogin:
time 0:30

yocalebo (changes requested):
1. update docstring
2. `system.is_ha_capable` can be called outside the inner `format_disk` method and simply passed as another arg
3. call `len(disks)` outside inner `format_disk` method and pass it in as an arg too

src/middlewared/middlewared/plugins/pool_/format_disks.py
```diff
@@ -7,30 +7,32 @@ class PoolService(Service):
     @private
     async def format_disks(self, job, disks):
         """
-        Format all disks, putting all freebsd-zfs partitions created
-        into their respective vdevs and encrypting disks if specified for FreeBSD.
+        Format all disks, putting all ZFS partitions created  into their respective vdevs.
```
yocalebo, line 10:
```suggestion
        Format all disks, putting all ZFS partitions created into their respective vdevs.
```

yocalebo (approved):
Confirmed on internal hardware that this works 😄

PR #10274, opened 2022-12-21
NAS-119427 / 23.10 / fix snmp-agent.py

yocalebo:
To quote Nietzsche, "Battle not with monsters, lest ye become a monster, and if you gaze into the abyss, the abyss gazes also into you.". Well, this is the abyss and I gazed into it. The ticket started off as seemingly innocent. `snmp-agent.py` uses 80-100% of a cpu core. Having the user generate a flamegraph of the process it becomes very apparent why it was occurring. I'll try and describe the fixes I've employed with this PR.

1. fix a core dump `libzfs` is not thread safe and even though we were only using that object for read-only operations, it still doesn't prevent us from potentially hitting reentrancy problems and crashing. With this fix, I only instantiate 1 `libzfs` object in the parent process.
2. remove an unnecessary thread (`ZpoolOThread`) (and very expensive, (friends don't let friends use `deepcopy` 😄 ))
3. remove unnecessary newlines and use a `prefix` variable for the arc stats `kstat.zfs.misc.arcstats`
4. All zfs stats are reported as uint64_t and we were representing these numbers as a `Integer32` type which meant we were having to calculate our own scale `allocation_units` which have nothing to do with zfs. This is unnecessary and very confusing because snmpv2/v3 have a `Counter64` type which we were already using. Convert ALL the snmp tables to `Counter64` so we can represent the numbers as-is without any unnecessary conversion.
5. this program was spiking a cpu because we're iterating (recursively) all datasets and zvols for all zpools and accessing the various space related properties. Accessing the properties is what takes the most time. To resolve this, I now generate mountinfo from procfs (this only shows zfs datasets not zvols) and use `statvfs` to calculate the size of datasets and only use `libzfs` for the zvols. This is a dramatic performance and time improvement.
6. fix `FREENAS-MIB.txt` to reflect the new structure(s)
7. `ZpoolHealthType` object was wrong on multiple accounts so I simply report the health of the zpool as a string

snmpwalk'ing our top-level OID on a system with ~1k datasets and ~1k zvols, it took approximately 60 seconds. With my changes it takes about 1 second. After my changes we're no longer spiking a cpu core (even on this multi-thousand zfs dataset/zvol system that I tested my changes on)


themylogin (changes requested):
IIRC when changing MIB.txt file you also have to re-generate python version http://docs.middleware.ixsystems.net/services/snmp.html

yocalebo:
> IIRC when changing MIB.txt file you also have to re-generate python version http://docs.middleware.ixsystems.net/services/snmp.html

oops, didn't push that in this PR. Using `smidump` on debian allows me to automatically generate that file based off FREENAS-MIB.txt but it changes format dramatically (in a better way). Will push shortly

src/freenas/usr/local/bin/snmp-agent.py
```diff
@@ -437,19 +369,116 @@ def run(self):
...
+        except Exception:
+            # don't crash, it's snmp so just continue
+            continue
+        else:
+            total_bytes = st.f_blocks * st.f_frsize
+            free_bytes = st.f_bfree * st.f_frsize
+            avail_bytes = st.f_bavail * st.f_frsize
+
+            idx += 1
+            fill_in_dataset_snmp_row_info(idx, info['mount_source'], total_bytes, free_bytes, avail_bytes)
```
themylogin, line 470: Does this mean the first row will have `idx=2`?

themylogin (changes requested):

src/freenas/usr/local/bin/snmp-agent.py
yocalebo: Pushing a round of fixes which resolves this.

anodos325:
Is it worth enhancing to include the object kstat counters for datasets?
```
root@TN3[~]# cat /proc/spl/kstat/zfs/dozer/objset-0x1069
46 1 0x01 7 2160 22676661853 504310406747386
name                            type data
dataset_name                    7    dozer/nfs
writes                          4    0
nwritten                        4    0
reads                           4    0
nread                           4    0
nunlinks                        4    0
nunlinked                       4    0
```

yocalebo:
> Is it worth enhancing to include the object kstat counters for datasets?
> 
> ```
> root@TN3[~]# cat /proc/spl/kstat/zfs/dozer/objset-0x1069
> 46 1 0x01 7 2160 22676661853 504310406747386
> name                            type data
> dataset_name                    7    dozer/nfs
> writes                          4    0
> nwritten                        4    0
> reads                           4    0
> nread                           4    0
> nunlinks                        4    0
> nunlinked                       4    0
> ```

Not for 22.12.1, maybe for 23.10. I'm going to actually push another PR to get rid of the `FREENAS-MIB*` related files and convert everything to `TRUENAS-MIB` and probably structure it a bit better (i.e. `TRUENAS-ZFS-MIB`, `TRUENAS-DISKTEMP-MIB`, etc)

yocalebo:
> IIRC when changing MIB.txt file you also have to re-generate python version http://docs.middleware.ixsystems.net/services/snmp.html

We don't need to run this through `libsmi2pysnmp` in my testing. If you just run `smidump -f python > blah.py` it generates a file that has json formatted structure and it worked on the machine that I tested these changes on. However, in interest of keeping changes as minimal as possible I went ahead and ran through `libsmi2pysnmp`.

themylogin (approved):

PR #10277, opened 2022-12-21
NAS-119507 / 23.10 / Don't delete builtin idmap table entries on constraint violation

anodos325:
It's not altogether clear how a user got in this situation, but the production impact of nuking the default domain idmap entry is rather profound. In the case of idmap-related tables we should remove the foreign key rather than deleting the table entry (at least if it's a builtin entry).

^^^ early stages of investigation

In commit https://github.com/truenas/middleware/commit/5265c8c49f84e3bf2bcd2bfcc30e3559019b0ca3 a migration was written to use AUTOINCREMENT to ensure id uniqueness.
In commit https://github.com/truenas/middleware/commit/85f5b97ec9a388b04ed50ff2c3c6528d0b3de994 the aforementioned migration was modified to also fix potential constraint
violation in this field.

Since there was a gap between these two commits, it is impossible that the original
migration without the subsequent revision and therefore the user's DB still contains the original
constraint violation. This table entry is critical to the proper function of the AD
plugin and since it is user-configurable, deletion cannot be repaired without manual
intervention.

The practical impact of this is that AD users who had installed SCALE during the 22.02
BETA cycle are broken on upgrade to 22.12.0. This is the first step in a two-part fix
for the issue.

### DIFF
--- a/src/middlewared/middlewared/plugins/datastore/connection.py
+++ b/src/middlewared/middlewared/plugins/datastore/connection.py
@@ -1,5 +1,7 @@
 from concurrent.futures import ThreadPoolExecutor
 import re
+import shutil
+import time
 
 from sqlalchemy import create_engine
 
@@ -28,6 +30,32 @@ class DatastoreService(Service):
     connection = None
 
     @private
+    def handle_constraint_violation(self, row, journal):
+        self.logger.warning("Row %d in table %s violates foreign key constraint on table %s.",
+                            row.rowid, row.table, row.parent)
+
+        if row.table == "directoryservice_idmap_domain" and row.rowid <= 5 and row.parent == "system_certificate":
+            """
+            In commit 5265c8c49f8 a migration was written to use AUTOINCREMENT to ensure id uniqueness.
+            In commit 85f5b97ec9a the aforementioned migration was modified to also fix potential constraint
+            violation in this field.
+
+            Since there was a gap between these two commits, it is impossible that the original
+            migration without the subsequent revision and therefore the user's DB still contains the original
+            constraint violation. This table entry is critical to the proper function of the AD
+            plugin and since it is user-configurable, deletion cannot be repaired without manual
+            intervention.
+            """
+            self.logger.warning("Removing certificate id for default idmap table entry.")
+            self.connection.execute(f"UPDATE {row.table} SET idmap_domain_certificate_id = NULL WHERE rowid = {row.rowid}")
+            return
+
+        self.logger.warning("Deleting row %d from table %s.", row.rowid, row.table)
+        op = f"DELETE FROM {row.table} WHERE rowid = {row.rowid}"
+        self.connection.execute(op)
+        journal.write(f'{op}\n')
+
+    @private
     def setup(self):
         if self.engine is not None:
             self.engine.dispose()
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/datastore/connection.py
```diff
@@ -27,6 +29,19 @@ class Config:
     engine = None
     connection = None
 
+    @private
+    def handle_constraint_violation(self, row):
+        self.logger.warning("Row %d in table %s violates foreign key constraint on table %s.",
+                            row.rowid, row.table, row.parent)
+
+        if row.table == "directoryservice_idmap_domain" and row.rowid <= DSType.DS_TYPE_DEFAULT_DOMAIN.value:
```
themylogin, line 37: We should probably check `row.parent` also to ensure that the violating column is indeed `idmap_domain_certificate_id` and not some other column

themylogin (changes requested):

src/middlewared/middlewared/plugins/datastore/connection.py
william-gr: Isnt this code violating the database layer and the database should be extended so the logic is on idmap side?

src/middlewared/middlewared/plugins/datastore/connection.py
```diff
@@ -27,6 +27,19 @@ class Config:
     engine = None
     connection = None
 
+    @private
+    def handle_constraint_violation(self, row):
+        self.logger.warning("Row %d in table %s violates foreign key constraint on table %s.",
+                            row.rowid, row.table, row.parent)
+
+        if row.table == "directoryservice_idmap_domain" and row.rowid <= 5 and row.parent == "system_certificate":
+            self.logger.warning("Removing certificate id for default idmap table entry.")
```
yocalebo, line 36: Let's put a comment explaining why we short-circuit here. (i.e. the fallout of deleting an entry from idmap table)

yocalebo (changes requested):

anodos325:
I believe I have a plausible causal-chain for the user's DB issue and wrote comment / updated commit message.

themylogin (approved):

src/middlewared/middlewared/plugins/datastore/connection.py
```diff
@@ -27,6 +27,30 @@ class Config:
...
+
+        if row.table == "directoryservice_idmap_domain" and row.rowid <= 5 and row.parent == "system_certificate":
+            """
+            In commit 5265c8c49f8 a migration was written to use AUTOINCREMENT to ensure id uniqueness.
+            In commit 85f5b97ec9a the aforementioned migration was modified to also fix potential constraint
+            violation in this field.
+
+            Since there was a gap between these two commits, it is impossible that the original
+            migration without the subsequent revision and therefore the user's DB still contains the original
+            constraint violation. This table entry is critical to the proper function of the AD
```
themylogin, line 43: I can disagree.

Commit 85f5b97ec9a was added because there was a migration crash described here https://ixsystems.atlassian.net/browse/NAS-111944. Recreating the table with new data was failing because the original table was containing invalid data (specifically `idmap_domain_certificate_id = ''`).

If the user had such an invalid data in their database before it's not possible they had applied `2021-01-20_10-19_autoincrement` successfully before my fix was made. If they did not have such data then both unpatched and patched migration will (or would have) yield the same result.

anodos325: I reproduced the error by installing 21.08-BETA1 on a VM and upgrading to 22.02.4.

yocalebo (approved):

anodos325:
I'll make an additional DB migration to re-add a default entry if it was dropped during the 22.12.0 upgrade. This will in most cases un-break AD.

yocalebo:
@anodos325 ping. What are we doing with this one?

yocalebo (approved):

PR #10287, opened 2022-12-23
NAS-118588 / 23.10 / Fix kerberos ticket renewals

anodos325:
Renewal logic was from TrueNAS 11.3, which was a refactorig of renewal logic from 9.10. It was somewhat broken on SCALE (relied on crontab entry to perform checks).

This PR simplifies how tickets are managed internally. I don't think we actually need to cache ticket info, but it is useful to explicitly indicate the default prinicpal. Timestamps are now converted into unix timestamps rather than python time struct.

A timestamp is also added to indicate the maximum time for which ticket can be renewed. This is also now evaluated to determine whether to get fresh ticket or attempt renewal.

Ticket flags are now parsed and evaluated to determine whether the ticket we're looking at is actually renewable. If it's not then we do fresh kinit rather than try to renew it.

Instead of having a crontab entry to check on the status of kerberos ticket, instead we now have a middleware job that keeps status info on when next renewal will be performed.


anodos325:
Will add tests once I'm done refactoring AD tests to be more reliable.

yocalebo (approved):

PR #10292, opened 2022-12-23
NAS-119558 / 23.10 / Refactor AD usage in test_030 and test_007

anodos325:
Add new activedirectoy context manager to ensure that we properly clean up AD state after tests complete. Simplify test_030 file to cover AD join / leave and SMB ops while joined to AD. Kerberos-specific are in separate file and will be updated in next PR.

This PR also expands the python SMB client to allow connections using AD credentials. In this case it uses NTLMv2 auth since the testing box is not joined to the AD domain.

### DIFF
--- a/tests/api2/assets/REST/directory_services.py
+++ b/tests/api2/assets/REST/directory_services.py
@@ -1,0 +1,60 @@
+import contextlib
+import urllib.parse
+from time import sleep
+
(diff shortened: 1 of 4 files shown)


anodos325:
These are being thrown into a context manager so that we have fewer false-positives for failure (and perhaps run tests more efficiently). This is also paring down / simplifying test cases.

yocalebo (approved):

PR #10295, opened 2022-12-26
fix apps backup lambda function

stavros-k:
Not sure if this is the right fix.. If not feel free to close :P


stavros-k:
Closing in favor of #10296

PR #10296, opened 2022-12-26
NAS-119511 / 23.10 / Fix filtering of secrets

Qubad786:


stavros-k (approved):
Even better solution than mine.
`sorted` should be able to iterate over a filter just fine!

Ornias1993 (commented):
LGTM :)

sonicaj (approved):

PR #10298, opened 2022-12-26
NAS-119595 / 23.10 / Default to depth=1 for catalog clone

Ornias1993:
We just need the "latest and greatest" when cloning an Apps catalog. As there is literally no advanced git operations likely to be ran that require full history (or any history at-all)

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/utils.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/utils.py
@@ -20,9 +20,9 @@ def convert_repository_to_path(git_repository_uri, branch):
     return git_repository_uri.split('://', 1)[-1].replace('/', '_').replace('.', '_') + f'_{branch}'
 
 
-def clone_repository(repository_uri, destination, depth):
+def clone_repository(repository_uri, destination, depth, branch):
     shutil.rmtree(destination, ignore_errors=True)
-    return git.Repo.clone_from(repository_uri, destination, env=os.environ.copy(), depth=depth)
+    return git.Repo.clone_from(repository_uri, destination, env=os.environ.copy(), depth=depth, branch=branch)
 
 
 def get_repo(destination):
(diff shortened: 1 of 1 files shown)


stavros-k (approved):

g-a-c:
You can't leave comments against unchanged lines, but I think there may be an issue here...https://github.com/truenas/middleware/pull/10298/files#diff-15359a7e2673706309c322c8d58ebfa2a0b1271ba7752efba95b7e2b10f38fe0R50-R55

If the repo is not currently available locally, then this will do a shallow clone. This will pull down the default branch for the repo OK, but if you want to use a different branch (which is a user-editable option in the web UI) then the shallow clone will not know of the existence of non-default branches (because shallow clone) and the `checkout` operation to switch from the default branch to the user-specified branch will fail

g-a-c:
```python
        if clone_repo:
            try:
                repo = clone_repository(repository_uri, destination, depth, branch)
```
and
```python
def clone_repository(repository_uri, destination, depth, branch):
    shutil.rmtree(destination, ignore_errors=True)
    return git.Repo.clone_from(repository_uri, destination, env=os.environ.copy(), depth=depth, branch=branch)
```
seem to work OK for me

Ornias1993:
> ```python
> def clone_repository(repository_uri, destination, depth, branch):
>     shutil.rmtree(destination, ignore_errors=True)
>     return git.Repo.clone_from(repository_uri, destination, env=os.environ.copy(), depth=depth, branch=branch)
> ```

This definately is fair and useable review feedback :)

Ornias1993:
> Passing the `branch` parameter through to `clone_repository()` like this:
> 
> ```python
>         if clone_repo:
>             try:
>                 repo = clone_repository(repository_uri, destination, depth, branch)
> ```
> 
> and:
> 
> ```python
> def clone_repository(repository_uri, destination, depth, branch):
>     shutil.rmtree(destination, ignore_errors=True)
>     return git.Repo.clone_from(repository_uri, destination, env=os.environ.copy(), depth=depth, branch=branch)
> ```
> 
> seem to work OK for me

Implemented!

sonicaj (approved):
Thank you for your contribution

PR #10299, opened 2022-12-27
NAS-119464 / 23.10 / Add integration test for validating k8s backup

Qubad786:
## Context

A test has been added to validate that the contents of apps backup are valid including backing up the app secrets and persistent volume information so we can use that information to restore the backup safely.

### DIFF
--- a/tests/api2/test_026_kubernetes_backup_chart_releases.py
+++ b/tests/api2/test_026_kubernetes_backup_chart_releases.py
@@ -1,3 +1,4 @@
+import json
 import os
 import pytest
 import sys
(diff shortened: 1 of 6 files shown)


tests/api2/test_026_kubernetes_backup_chart_releases.py
```diff
@@ -239,3 +241,78 @@ def test_24_get_k3s_logs():
...
+        depends(request, ['release_truechart_plex'])
+        global backup_name
+        backup_name = call('kubernetes.backup_chart_releases', job=True)
+        assert backup_name != ''
+
+
+    def test_28_validate_backup_structure(request):
+        depends(request, ['ix_app_truecharts_backup'])
+
+        def is_file_exist(file_path: str) -> bool:
```
sonicaj, line 282: We can test this with `filesystem.listdir` instead of using SSH

tests/api2/test_026_kubernetes_backup_chart_releases.py
```diff
@@ -239,3 +241,78 @@ def test_24_get_k3s_logs():
...
+            'label': 'TRUECHARTS',
+            'repository': 'https://github.com/truecharts/catalog.git',
+            'branch': 'main'
+        }
+        job = call('catalog.create', payload)
+        job_status = wait_on_job(job, 600)
+        assert job_status['state'] == 'SUCCESS', str(job_status['results'])
+
+    @pytest.mark.dependency(name='release_truechart_plex')
+    def test_26_create_plex_chart_release(request):
```
sonicaj, line 260: Instead of creating separate tests for these as they are not really what we want to test, i believe you should use some sort of try/finally or context manager approach where you can setup the pre-reqs before validating the structure and then after you have done the tests ( regardless of failure ) cleanup on exist.

sonicaj (approved):

Ornias1993 (commented):
This should, at least, catch some of the issues we've all encountered in Bluefin. :)
LGTM.

PR #10302, opened 2022-12-27
NAS-119610 / 23.10 / Remove unnecessary method

yocalebo:
Remove `filesystem.file_get_contents` since it doesn't have any type of validation and doesn't mix well with our rootless login functionality. This was a private method (only available across websocket connections) and was only used in our integration test suite, so I've fixed those tests accordingly.

### DIFF
--- a/tests/api2/test_cloud_sync.py
+++ b/tests/api2/test_cloud_sync.py
@@ -3,7 +3,8 @@ import time
 
 import pytest
 from pytest_dependency import depends
-from middlewared.test.integration.assets.cloud_sync import credential, task, local_s3_credential, local_s3_task, run_task
+from middlewared.test.integration.assets.cloud_sync import credential, task, local_s3_credential
+from middlewared.test.integration.assets.cloud_sync import local_s3_task, run_task
 from middlewared.test.integration.assets.ftp import anonymous_ftp_server, ftp_server_with_user_account
 from middlewared.test.integration.assets.pool import dataset
 from middlewared.test.integration.utils import call, pool, ssh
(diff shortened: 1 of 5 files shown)


anodos325 (approved):

tests/api2/test_cloud_sync.py
```diff
@@ -218,21 +219,29 @@ def test_state_persist():
...
-                local_logs = call("filesystem.file_get_contents", local_logs_path)
-
-                remote_job = call("failover.call_remote", "cloudsync.get_instance", [task["id"]])["job"]
-                remote_logs_path = remote_job["logs_path"]
-                remote_logs = call("failover.call_remote", "filesystem.file_get_contents", [remote_logs_path])
-
-                assert local_logs == remote_logs
+                path1 = call("cloudsync.get_instance", task["id"])["job"]["logs_path"]
+                path2 = call("failover.call_remote", "cloudsync.get_instance", [task["id"]])["job"]["logs_path"]
+                assert verify_file_contents_match(path1, path2)
```
themylogin, line 247: The issue with this approach is that the test results will not contain the difference between these two strings

src/middlewared/middlewared/test/integration/utils/files.py
```diff
@@ -0,0 +1,40 @@
...
+    if not local.is_absolute():
+        assert False, 'An absolute path for where the file is downloaded must be provided'
+    elif local.is_dir():
+        local_absolute = local.joinpath(remote_file).as_posix()
+    else:
+        local_absolute = local.as_posix()
+
+    gotit = None
+    try:
+        gotit = get_file(remote_file.as_posix(), local_absolute, user, password, _ip)
```
themylogin, line 33: We only use this for small text files that fit into memory, can we just run `cat /file/path` via SSH in order to avoid using temporary files in the filesystem? That will result in a much more easy-to-use API

themylogin (changes requested):

tests/api2/test_cloud_sync.py
yocalebo: Why would we want to see the difference between the files in this test? What will that show us?

src/middlewared/middlewared/test/integration/utils/files.py
yocalebo: And what about files that are bigger?

tests/api2/test_cloud_sync.py
themylogin: If we assert two things to be equal it's usually useful to see how exactly are they not different, that's what pytest does by default on assertion failure

src/middlewared/middlewared/test/integration/utils/files.py
themylogin: We can keep the method that will retrieve them into a temporary path, but a method to retrieve to RAM will be more convenient for the vast majority of the test cases. Also let's remember that this is a test runner so even 100MB file is fine to be read in RAM.

yocalebo:
@themylogin ping now that the HA websocket connection has been fixed. snmp tests pass

themylogin:
time 6:00

PR #10303, opened 2022-12-28
NAS-119618 / 22.12.1 / Use `$` for `admin` user zsh prompt

themylogin:


themylogin:
time 0:30

anodos325 (approved):

yocalebo (changes requested):
How does this effect `admin` user since you're changing `/root/.zshrc` file?

themylogin:
@yocalebo `/home/admin` is a copy of `/root` https://github.com/truenas/middleware/blob/master/src/freenas/debian/rules#L25

yocalebo (approved):

PR #10304, opened 2022-12-28
NAS-119619 / 23.10 / Fix scaling apps after restore

Qubad786:
## Problem

When a backup is restored all apps (including stopped ones) are started automatically.
What happens is that `0` was being treated as boolean false which resulted in falling back to default value when scaling workloads which in most cases is 1.

## Solution

Explicitly make sure that specified value is `None` and only in that case use the configured default value.


yocalebo (approved):

Ornias1993 (commented):
LGTM, looks good and an understandable bug :)

PR #10306, opened 2022-12-28
NAS-119621 / 23.10 / dont spam logs

yocalebo:
Our BHYVE SCALE HA VMs don't have nvdimm devices so the failure is expected.


themylogin (changes requested):
Should we skip running this alert on BHYVE at all?

Or does presence of something in the logs on the test runner even worth littering production code with checks like this?

yocalebo:
retest this please

themylogin (approved):

PR #10309, opened 2022-12-28
NAS-119570 / 23.10 / Do not reset `/home/admin` permissions

themylogin:


anodos325:
Since these permissions are rather important, maybe add some tests to ensure that we don't have another regression.

yocalebo (changes requested):
Agree, should be relatively easy to add tests to check perms and owner info for this dir

yocalebo (approved):
Great!

PR #10324, opened 2022-12-28
NAS-119423 / 23.10 / Strip trailing '/32' when writing /etc/initiators.allow

bmeagherix:
When writing `/etc/initiators.allow` the netmask should only be included when not dealing with a specific address.

Added some additional tests to further exercise this functionality.

### DIFF
--- a/src/middlewared/middlewared/etc_files/initiators.allow.mako
+++ b/src/middlewared/middlewared/etc_files/initiators.allow.mako
@@ -1,7 +1,20 @@
 <%
+    import ipaddress
     base_name = middleware.call_sync('iscsi.global.config')['basename']
     targets = middleware.call_sync('iscsi.target.query', [['auth_networks', '!=', []]])
+
+    def parse_auth(auth):
+        s = auth.split('/')
+        if len(s) == 2:
+            try:
+                ipobj = ipaddress.ip_interface(s[0])
+            except ValueError:
+                middleware.logger.warning(f"Invalid IP address: {s[0]}", exc_info=True)
+            else:
+                if (ipobj.version == 4 and s[1] == '32') or (ipobj.version == 6 and s[1] == '128'):
+                    return str(ipobj.ip)
+        return auth
 %>\
 % for target in targets:
-${base_name}:${target['name']} ${', '.join(target['auth_networks'])}
+${base_name}:${target['name']} ${', '.join([parse_auth(auth) for auth in target['auth_networks']])}
 % endfor
--- a/src/middlewared/middlewared/etc_files/initiators.allow.mako
+++ b/src/middlewared/middlewared/etc_files/initiators.allow.mako
@@ -9,7 +9,7 @@
             try:
                 ipobj = ipaddress.ip_interface(s[0])
             except ValueError:
-                middleware.logger.warning(f"Invalid IP address: {s[0]}", exc_info=True)
+                middleware.logger.warning("Invalid IP address: %s", s[0], exc_info=True)
             else:
                 if (ipobj.version == 4 and s[1] == '32') or (ipobj.version == 6 and s[1] == '128'):
                     return str(ipobj.ip)
--- a/src/middlewared/middlewared/etc_files/initiators.allow.mako
+++ b/src/middlewared/middlewared/etc_files/initiators.allow.mako
@@ -4,16 +4,14 @@
     targets = middleware.call_sync('iscsi.target.query', [['auth_networks', '!=', []]])
 
     def parse_auth(auth):
-        s = auth.split('/')
-        if len(s) == 2:
-            try:
-                ipobj = ipaddress.ip_interface(s[0])
-            except ValueError:
-                middleware.logger.warning("Invalid IP address: %s", s[0], exc_info=True)
-            else:
-                if (ipobj.version == 4 and s[1] == '32') or (ipobj.version == 6 and s[1] == '128'):
-                    return str(ipobj.ip)
-        return auth
+        try:
+            ipobj = ipaddress.ip_interface(auth)
+        except ValueError:
+            middleware.logger.warning("Invalid IP address: %s", auth, exc_info=True)
+        else:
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/etc_files/initiators.allow.mako
```diff
@@ -1,7 +1,18 @@
 <%
+    import socket
     base_name = middleware.call_sync('iscsi.global.config')['basename']
     targets = middleware.call_sync('iscsi.target.query', [['auth_networks', '!=', []]])
+
+    def parse_auth(auth):
+        s = auth.split('/')
+        try:
+            if len(s) == 2 and s[1] == '32':
+                socket.inet_pton(socket.AF_INET, s[0])
```
yocalebo, line 10: What about IPv6 addresses? I know we can't test this (still waiting on some internal infrastructure changes).

yocalebo (changes requested):

src/middlewared/middlewared/etc_files/initiators.allow.mako
yocalebo: Also, I believe using the builtin `ipaddress.ip_interface()` method would be a bit easier to use here.

yocalebo: Something like
```
try:
    ipobj = ipaddress.ip_interface('address-here')
except ValueError:
    log-something-about-invalid-ipaddress
else:
    return str(ipobj.ip)

Comment (bmeagherix) [reply src/middlewared/middlewared/etc_files/initiators.allow.mako:10]: Agreed, it doesn't handle IPv6 properly yet, but the `inet_pton` check is there so that we can add another check to the `parse_auth` function to do so (e.g. another try..except) when we get it plumbed thru.

We can add code now, but as you say can't test yet.

File: src/middlewared/middlewared/etc_files/initiators.allow.mako [hunk 2]
```diff
@@ -1,7 +1,20 @@
...
     base_name = middleware.call_sync('iscsi.global.config')['basename']
     targets = middleware.call_sync('iscsi.target.query', [['auth_networks', '!=', []]])
+
+    def parse_auth(auth):
+        s = auth.split('/')
+        if len(s) == 2:
+            try:
+                ipobj = ipaddress.ip_interface(s[0])
+            except ValueError:
+                middleware.logger.warning(f"Invalid IP address: {s[0]}", exc_info=True)
```
anodos325, line 12: Typically we don't use fstring for logger messages (just style thing)
```
middleware.logger.warning("Invalid IP address: %s", s[0], exc_info=True)
```

anodos325:
@yocalebo Do you think we should create an alert if there is an invalid IP address here (in addition to logging)?

src/middlewared/middlewared/etc_files/initiators.allow.mako
```diff
@@ -1,7 +1,20 @@
 <%
+    import ipaddress
     base_name = middleware.call_sync('iscsi.global.config')['basename']
     targets = middleware.call_sync('iscsi.target.query', [['auth_networks', '!=', []]])
+
+    def parse_auth(auth):
```
yocalebo, line 6: I was thinking more like this (instead of doing `auth.split`). This also has benefit of "normalizing" the network entries. i.e. if someone does something like `192.168.1.128/16`, it'll return `192.168.0.0/16`.

```
def parse_auth(auth):
    try:
        ipobj = ipaddress.ip_interface(auth)
    except ValueError:
        middleware.logger.warning("Invalid IP address: %s", auth, exc_info=True)
    else:
        if ipobj.network.prefixlen in (32, 128):
            return str(ipobj.ip)
        return str(ipobj.network)

**Comment by yocalebo** [comment]
> @yocalebo Do you think we should create an alert if there is an invalid IP address here (in addition to logging)?

Don't believe so in this instance. API uses `IPAddr` class to validate input so if this fails it's because someone manually edited db or we have some ancient legacy software that had invalid values long before we had this code in place.

**Review by yocalebo — APPROVED** [review]
Great job! I left one final comment on the `def parse_auth` function on how I perceived it to operate :) I'll let you decide if you want to implement it.

File: src/middlewared/middlewared/etc_files/initiators.allow.mako (hunk 3 above)
Comment (yocalebo) [reply src/middlewared/middlewared/etc_files/initiators.allow.mako:6]: I actually looked at the API and we use the same python module to validate the input so it should be an extreme edge-case if this function fails. (i.e. someone made manual changes to db or we have some ancient legacy version that like 5 people (total) are using).

File: src/middlewared/middlewared/etc_files/initiators.allow.mako [hunk 4]
```diff
@@ -1,7 +1,18 @@
...
+        except ValueError:
+            middleware.logger.warning("Invalid IP address: %s", auth, exc_info=True)
+        else:
+            if ipobj.network.prefixlen in (32, 128):
+                return str(ipobj.ip)
+            return str(ipobj.network)
 %>\
 % for target in targets:
-${base_name}:${target['name']} ${', '.join(target['auth_networks'])}
+${base_name}:${target['name']} ${', '.join([parse_auth(auth) for auth in target['auth_networks']])}
```
yocalebo, line 17: Sorry, I just noticed this possible failure condition. If `parse_auth` returns `None` then `.join` will fail with `TypeError`

yocalebo (changes requested):

src/middlewared/middlewared/etc_files/initiators.allow.mako
yocalebo: Think it might be best to just pass `target['auth_networks']` in it's entirety to `parse_auth` and make `parse_auth` return the formatted string.

bmeagherix: Sounds good.  See below.

yocalebo (approved):
Awesome, thanks!

PR #10327, opened 2022-12-29
NAS-119649 / 23.10 / fix and improve user.shell_choices

yocalebo:
1. still referenced `netcli` which has been removed
2. Took a `user_id` argument which is non-intuitive and confusing because that's not the uid, it's the database id. This argument isn't needed anymore
3. remove `IS_FREEBSD` checks in this file
4. simplify the method
5. add proper `@returns` information


yocalebo:
Waiting on associated webUI change since this is a backwards incompatible change.

yocalebo:
backport

anodos325 (approved):

PR #10335, opened 2022-12-30
NAS-119667 / 22.12.1 / remove dangling references to netcli

yocalebo:
In the rare case where a CORE user has upgraded to SCALE and 1 of their local users has `/etc/netcli` set as the user shell, we'll set it to `/usr/sbin/nologin`. We default to nologin because it's a fail-safe way of making sure we don't grant extra permissions than what the end-user expects.


src/middlewared/middlewared/alembic/versions/22.12/2022-12-30_13-21_fix_netcli_refs.py
```diff
@@ -0,0 +1,28 @@
...
+revision = '5cc601ce9a8e'
+down_revision = '136adf794fed'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    conn = op.get_bind()
+    conn.execute(
+        'UPDATE account_bsdusers SET bsdusr_shell = "/usr/sbin/nologin" WHERE instr("bsdusr_shell", "netcli") > 1'
```
themylogin, line 22: `instr("bsdusr_shell", "netcli")` will check if `"netcli"` is in literal string `"bsdusr_shell"`which is always `0`.  You probably meant `instr(bsdusr_shell, "netcli") > 1`. Or, better, just `bsdusr_shell LIKE "%netcli%"`

themylogin (changes requested):

src/middlewared/middlewared/alembic/versions/22.12/2022-12-30_13-21_fix_netcli_refs.py
yocalebo: No, this works as intended

themylogin (approved):

src/middlewared/middlewared/alembic/versions/22.12/2022-12-30_13-21_fix_netcli_refs.py
yocalebo: I looked at the sqlite official documentation for `instr` and wow it's bad. I've moved to using `WHERE bsdusr_shell LIKE "%netcli%"` since it's much easier to understand.

PR #10351, opened 2023-01-03
NAS-119702 / 23.10 / Add nameserver checks for activedirectory plugin

anodos325:
Some users may get confused about their AD domain name and how to configure DNS when TrueNAS is joined to active directory.

This merge request attempts to take some of the burden off of the administrator by:

1. Validating that each of our configured nameservers can successfully look up relevant SRV records for our AD domain. This will hopefully reduce occurance of administrators putting non-AD nameservers (e.g. Google) in their resolv.conf.

2. Validating that each of the nameservers can actually resolve our AD domain name. This will hopefully catch users doing something like setting the FQDN of a domain controller as the active directory domain name.

3. Incorporating nameserver checks into our hourly AD health alert checks.

### DIFF
--- a/src/middlewared/middlewared/alert/source/active_directory.py
+++ b/src/middlewared/middlewared/alert/source/active_directory.py
@@ -3,6 +3,7 @@ import logging
 from middlewared.alert.base import AlertClass, AlertCategory, Alert, AlertLevel, AlertSource
 from middlewared.alert.schedule import CrontabSchedule, IntervalSchedule
 from middlewared.plugins.directoryservices import DSStatus
+from middlewared.service_exception import CallError
 
 log = logging.getLogger("activedirectory_check_alertmod")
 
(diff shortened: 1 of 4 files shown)


yocalebo (approved):

src/middlewared/middlewared/alert/source/active_directory.py
```diff
@@ -39,6 +40,15 @@ async def check(self):
                 key=None
             )
 
+        try:
+            await self.middleware.call("activedirectory.check_nameservers")
```
TheJulianJES, line 44: Might be missing something, but are the domain and site parameters missing from this call?

```python
Failed to check for alert ActiveDirectoryDomainHealth: Traceback (most recent call last): File "/usr/lib/python3/dist-packages/middlewared/plugins/alert.py", line 776, in __run_source alerts = (await alert_source.check()) or [] File "/usr/lib/python3/dist-packages/middlewared/alert/source/active_directory.py", line 44, in check await self.middleware.call("activedirectory.check_nameservers") File "/usr/lib/python3/dist-packages/middlewared/main.py", line 1386, in call return await self._call( File "/usr/lib/python3/dist-packages/middlewared/main.py", line 1335, in _call return await methodobj(*prepared_call.args) TypeError: check_nameservers() missing 1 required positional argument: 'domain'
```

anodos325: Doh, looks like you're right. I'll get a fix prepped. Thanks for pointing this out.

TheJulianJES: Awesome, thanks!
The error happened with today's release of SCALE 22.12.1, but I didn't see an issue in the tracker yet. If needed, I can create one.

anodos325: https://github.com/truenas/middleware/pull/10729

Already created. NAS-120430

PR #10359, opened 2023-01-04
NAS-119518 / 23.10 / Refine whitelisted host paths for apps validation

Qubad786:
## Context

In docker we were not able to distinguish between a host path as required by the container itself to function or if it being a user specified host path.

With the change to containerd we can now make that distinction and paths have been updated accordingly so as to narrow down whitelisted paths which were allowed earlier.


yocalebo:
retest this please

yocalebo (approved):

PR #10363, opened 2023-01-06
NAS-119577 / 22.12.1 / Ignore container coredumps in bluefin

Qubad786:
## Context

We are ignoring container coredumps but had missed to include `k3s.service` unit to list of ignores as coredumps generated can also report as being generated by `k3s.service`.


yocalebo (approved):

yocalebo:
Is this needed in 23.10? I assume so

PR #10367, opened 2023-01-07
NAS-119763 / 23.10 / Expand tests for adding kerberos principal

anodos325:
This ensures that NFS principal in kerberos
keytab is properly visible when LDAP is configured.

This is prep step for adding functional testing for 
krb5+NFS with both AD and LDAP plugins.

### DIFF
--- a/tests/api2/test_036_ad_ldap.py
+++ b/tests/api2/test_036_ad_ldap.py
@@ -24,8 +24,16 @@ else:
     pytestmark = pytest.mark.skipif(dev_test, reason='Skip for testing')
 
 @pytest.fixture(scope="module")
-def kerberos_config():
-    return {}
+def kerberos_config(request):
+    payload = {"v4_krb": True}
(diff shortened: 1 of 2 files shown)


tests/api2/test_036_ad_ldap.py
```diff
@@ -24,8 +24,16 @@
...
 
 @pytest.fixture(scope="module")
-def kerberos_config():
-    return {}
+def kerberos_config(request):
+    payload = {"v4_krb": True}
+    results = PUT("/nfs/", payload)
+    assert results.status_code == 200, results.text
+    try:
+        yield(request)
```
yocalebo, line 32: This looks strange? just yield request?

anodos325: I don't really need the config info, but yielding lets me check the status of request.

yocalebo (approved):

PR #10379, opened 2023-01-09
NAS-119733 / 23.10 / Save empty iSCSI target alias as NULL

bmeagherix:
When saving an empty string iSCSI target alias store it as NULL.

The field in the underlying database is unique, so saving as "" is problematic.


bmeagherix:
Removed WIP after successful Jenkins run.

yocalebo (approved):

PR #10381, opened 2023-01-09
NAS-119789 / 23.10 / Shift more middleware-related sentinels to rundir

anodos325:
We should try to keep middleware-state-related items in /var/run/middleware rather than in /tmp

### DIFF
--- a/src/middlewared/middlewared/logger.py
+++ b/src/middlewared/middlewared/logger.py
@@ -65,7 +65,7 @@ class CrashReporting(object):
 
     def __init__(self):
         if sw_version_is_stable():
-            self.sentinel_file_path = '/tmp/.crashreporting_disabled'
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/logger.py
```diff
@@ -94,7 +94,7 @@ def is_disabled(self):
             bool: True if crash reporting is disabled, False otherwise.
         """
         # Allow report to be disabled via sentinel file or environment var,
-        # if FreeNAS current train is STABLE, the sentinel file path will be /tmp/,
+        # if FreeNAS current train is STABLE, the sentinel file path will be /var/run/middleware/,
```
yocalebo, line 97: FreeNAS to TrueNAS since you're already here 😄

yocalebo (approved):

PR #10384, opened 2023-01-10
NAS-119795 / 23.10 / Fix progress reporting for NTFS disk import

anodos325:
NTFS disk import jobs were stuck at 0% complete.


anodos325:
> I don't even like the fact this method exists.

 In general this method is flawed. It doesn't check for / preserve streams and is a sort of data migration anti-pattern. Migration should be done via file sharing protocols.

themylogin (approved):

PR #10391, opened 2023-01-10
NAS-119801 / 23.10 / Add `--returncode` parameter to `runtest.py` so it returns a failure …

themylogin:
…exit code if a test fails


themylogin:
time 0:20

yocalebo (approved):

PR #10396, opened 2023-01-10
NAS-119452 / 23.10 / Allow re-initializing ix-applications when force flag is specified

Qubad786:
## Context

Allow re-initializing ix-applications dataset if the existing setup has missed datasets which are required for apps to function (in the case when force flag is specified). This will destroy the current corrupt ix-applications setup and re-initialize the dataset as a fresh start for kubernetes.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
@@ -86,20 +86,26 @@ class KubernetesService(ConfigService):
         return can_run_apps
 
     @private
+    def missing_apps_dataset(self, apps_ds):
+        required_datasets = self.middleware.call_sync('kubernetes.kubernetes_datasets', apps_ds)
+        existing_datasets = set([
+            ds['id'] for ds in self.middleware.call_sync(
+                'zfs.dataset.query', [
+                    ['id', 'in', required_datasets]
+                ], {'extra': {'retrieve_children': False, 'properties': []}}
+            )
+        ])
+        return list(set(required_datasets) ^ existing_datasets)
+
+    @private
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/kubernetes_linux/update.py
```diff
@@ -86,20 +86,24 @@ async def license_active(self):
...
-        if force or not self.middleware.call_sync(
-            'zfs.dataset.query', [['id', '=', apps_ds]], {'extra': {'retrieve_children': False}}
-        ):
-            return
-
+    def missing_apps_dataset(self, apps_ds):
         missing = []
         for ds_name in self.middleware.call_sync('kubernetes.kubernetes_datasets', apps_ds):
             if not self.middleware.call_sync(
                 'zfs.dataset.query', [['id', '=', ds_name]], {'extra': {'retrieve_children': False}}
```
anodos325, line 93: Since we're just checking for dataset existence from the look of it, this can be optimized to not retrieve properties.

anodos325: We might also be able to do this in a single dataset query, right?

src/middlewared/middlewared/plugins/kubernetes_linux/update.py
```diff
@@ -440,9 +446,15 @@ async def do_update(self, job, data):
         old_config.pop('dataset')
         config = old_config.copy()
         config.update(data)
+        force_flag = config.get('force', False)
```
Ornias1993, line 449: Not described in function description...

Ornias1993: Resolved :)

Ornias1993 (commented):
Looking good LGTM!

PR #10399, opened 2023-01-11
NAS-119810 / 23.10 / fix ipmi sensor matching

yocalebo:
As pointed out by @rapperskull in PR: https://github.com/truenas/middleware/pull/10089, when calling `ipmitool` with the `elist` command, the hexadecimal value after the sensor name is removed. This fixes the sensor matching logic.

### DIFF
--- a/src/middlewared/middlewared/alert/source/ipmi_sel.py
+++ b/src/middlewared/middlewared/alert/source/ipmi_sel.py
@@ -124,12 +124,12 @@ class IPMISELAlertSource(AlertSource):
             record for record in records
             if (
                 (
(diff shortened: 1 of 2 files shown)


rapperskull:
Thank you for looking into it. I also would like to point out that this affects TrueNAS 13.0 too, so it needs to be backported.

yocalebo:
> Thank you for looking into it. I also would like to point out that this affects TrueNAS 13.0 too, so it needs to be backported.

Great, I'll confirm on some internal hardware and make sure it's included in next 13 release.

anodos325 (approved):

PR #10404, opened 2023-01-11
Add additional validation related to symlinks

anodos325:
SMB share paths that contain components that are
symbolic links may be unstable (as user may remove symlink breaking the share). Generally speaking,
we don't support the creation of symbolic links
through our API / webui and so this requires user
to manually do such things through the shell. This means additional validation shouldn't break users
who are using the appliance within bounds that
we set up, but should prevent some problem reports where users do things totally off the wall.


anodos325:
```root@truenas[/mnt/dozer/middleware/src/middlewared]# midclt call sharing.smb.create '{"name": "BOB", "path": "/mnt/dozer/symlink/foo"}'
[EINVAL] sharingsmb_create.path_local: Mountpoint /mnt/dozer/SHARE not within path /mnt/dozer/symlink/foo. This may indicate that the path of the SMB share contains a symlink component.
```

Sample error message.

yocalebo (approved):

yocalebo (changes requested):
Actually, flake8 issue and wondering if a test can be added?

anodos325:
Perhaps we need an over-arching ticket to test validation errors.

yocalebo (approved):

PR #10409, opened 2023-01-12
NAS-119829 / 23.10 / Ensure that DNS is updated on domain join

anodos325:
Add private internal API endpoint to use nsupdate
to update DNS A and AAAA records for the TrueNAS
server. Use this when joining AD to ensure that
TrueNAS server functions properly in an AD
environment. This PR also extends the IPAddr
schema so that we can optionally filter based
on whether we want to disallow certain types of
IP addresses (like ones reserved for multicast).

Validation in AD plugin is also expanded to check
whether the specified netbios name is already in
use by another server on the network (to prevent
us from clobbering it on domain join).

### DIFF
--- a/src/middlewared/middlewared/plugins/activedirectory.py
+++ b/src/middlewared/middlewared/plugins/activedirectory.py
@@ -213,6 +213,17 @@ class ActiveDirectoryService(TDBWrapConfigService):
 
     @private
     async def common_validate(self, new, old, verrors):
+        try:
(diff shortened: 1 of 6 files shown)


anodos325:
Flagged WIP so that I can add / test PTR handling.

yocalebo (changes requested):

yocalebo (approved):
Awesome!

PR #10411, opened 2023-01-12
NAS-119833 / 23.10 / Enable Nvidia GPU sharing for Apps

Ornias1993:
In contrast to Intel, the Nvidia device plugin, only accepts device sharing configuration via a configfile.
This is the vital component to support multiple Apps using the same (supported) Nvidia GPU.

As the existing code did not allow the creation, updating and removal of configmaps, this PR adds a section to define said configmaps.

To prevent confusion between the config(maps) and the (device)plugin daemonset, things have been renamed accordingly. As the previous (device)plugin daemonsets where named "config".

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/gpu.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/gpu.py
@@ -137,16 +137,16 @@ GPU_CONFIG = {
             'namespace': 'kube-system'
         },
         'data': {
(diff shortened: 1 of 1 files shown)


Ornias1993:
@yocalebo Should be ready for review now, so you can remove the labels :)

Ornias1993:
Potential reviewers:
@sonicaj @Qubad786 @stavros-k

Ornias1993:
succeeded by: #10439

PR #10421, opened 2023-01-13
NAS-119841 / 23.10 / prevent breaking smartd service with extra options

yocalebo:
Seen in the wild.  Someone had put `-d nvme` as an extra smart option in our API. Surprisingly, we do 0 validation here. We statically generate the smartd config and don't "overwrite" the config with any of these flags so it's best to prevent them here to prevent the smartd.conf from being generated incorrectly which ultimately prevents the smartd service from starting.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk.py
+++ b/src/middlewared/middlewared/plugins/disk.py
@@ -4,7 +4,7 @@ import subprocess
 from sqlalchemy.exc import IntegrityError
 
 from middlewared.schema import accepts, Bool, Datetime, Dict, Int, Patch, Str
-from middlewared.service import filterable, private, CallError, CRUDService
+from middlewared.service import filterable, private, CallError, CRUDService, ValidationError
 import middlewared.sqlalchemy as sa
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/disk.py
```diff
@@ -280,6 +280,11 @@ async def do_update(self, id, data):
             else:
                 await self.middleware.call('disk.toggle_smart_off', new['name'])
 
+            invalid_smart_flags = ['-a', '-d', '-n', '-W', '-m', '-M', 'exec']
```
themylogin, line 283: It's too late to validate at this point. The disk has already been updated at L263

themylogin (changes requested):

src/middlewared/middlewared/plugins/disk.py
yocalebo: oof, thank you. good catch, will fix

themylogin (approved):

PR #10428, opened 2023-01-13
NAS-119337 / 23.10 / Add ALUA support for iSCSI in SCALE

bmeagherix:


src/middlewared/middlewared/plugins/dlm.py
```diff
@@ -0,0 +1,443 @@
...
+    Simple interface with the kernel dlm.
+    """
+
+    SYSFS_DIR = '/sys/kernel/dlm'
+    CLUSTER_DIR = '/sys/kernel/config/dlm/cluster'
+    SPACES_DIR = CLUSTER_DIR + '/spaces'
+    COMMS_DIR = CLUSTER_DIR + '/comms'
+
+    def __init__(self, name):
+        if not os.path.isdir(KernelDlm.CLUSTER_DIR):
```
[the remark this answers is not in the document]
yocalebo: Or another way is to `import contextlib` and do something like
```
with contextlib.supporess(FileExistsError):
    for i in (blah):
        os.mkdir(i)
```

Up to you really 😄

yocalebo (commented):
Just some things I've noticed along the way.

src/middlewared/middlewared/plugins/dlm.py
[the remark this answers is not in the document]
bmeagherix: FWIW, went with the contextlib suppress

src/middlewared/middlewared/plugins/dlm.py
```diff
@@ -0,0 +1,504 @@
...
+    Simple interface with the kernel dlm.
+    """
+
+    SYSFS_DIR = '/sys/kernel/dlm'
+    CLUSTER_DIR = '/sys/kernel/config/dlm/cluster'
+    SPACES_DIR = CLUSTER_DIR + '/spaces'
+    COMMS_DIR = CLUSTER_DIR + '/comms'
+
+    def __init__(self, name):
+        with contextlib.suppress(FileExistsError):
```
anodos325, line 50: Wouldn't you want to suppress the os.mkdir() call rather than the iterator?

src/middlewared/middlewared/plugins/dlm.py
```diff
@@ -0,0 +1,298 @@
...
+        await middleware.call('dlm.leave_lockspace', lockspace)
+
+
+def remote_down_event(middleware, *args, **kwargs):
+    middleware.call_sync('dlm.remote_down')
+
+
+async def setup(middleware):
+    middleware.register_hook('udev.dlm', udev_dlm_hook)
+    # await middleware.call('failover.remote_on_connect', remote_status_event)
```
anodos325, line 297: Should this be commented out?

src/middlewared/middlewared/plugins/dlm.py
```diff
@@ -0,0 +1,298 @@
...
+            # Start the kernel lockspace on remaining nodes
+            await asyncio.gather(*[self.start_kernel_lockspace(nodeid, lockspace_name) for nodeid in nodeIDs])
+
+        except Exception:
+            self.logger.error('Failed to leave lockspace %s', lockspace_name, exc_info=True)
+            await self.middleware.call('dlm.kernel.lockspace_start', lockspace_name)
+            await self.middleware.call('dlm.kernel.set_sysfs_event_done', lockspace_name, 1)
+
+    @private
+    async def add_node(self, nodeid):
```
anodos325, line 235: I think it's okay to leave place-holder methods for future enhancements, but we probably need to note that it's a placeholder and raise a NonImplementedError at beginning of method

src/middlewared/middlewared/plugins/dlm_/kernel.py
```diff
@@ -0,0 +1,215 @@
...
+    if addr_obj.version == 4:
+        addr = sockaddr_in()
+        addr.sa_family = ctypes.c_ushort(socket.AF_INET)
+        if port:
+            addr.sin_port = ctypes.c_ushort(socket.htons(port))
+        if address:
+            bytes_ = [int(i) for i in address.split('.')]
+            addr.sin_addr = (ctypes.c_byte * 4)(*bytes_)
+    else:
+        raise NotImplementedError('Not implemented')
```
anodos325, line 33: Maybe explicitly state that IPv6 is not supported (clearer exception)

src/middlewared/middlewared/plugins/dlm_/kernel.py
```diff
@@ -0,0 +1,215 @@
...
+                    f.write(str(weight))
+
+    def lockspace_remove_node(self, lockspace_name, nodeid):
+        """
+        Remove the specified nodeid from the lockspace.
+        """
+        self.logger.debug('Removing node %s from lockspace %s', nodeid, lockspace_name)
+        node_path = os.path.join(KernelDistributedLockManagerService.SPACES_DIR, lockspace_name, 'nodes', '%d' % nodeid)
+        with contextlib.suppress(FileNotFoundError):
+            os.rmdir(node_path)
```
anodos325, line 183: Might want to double-check that this won't fail with ENOTEMPTY

src/middlewared/middlewared/plugins/failover_/event.py
```diff
@@ -728,8 +728,13 @@ def vrrp_backup(self, job, fobj, ifname, event):
...
 
-        # TODO: ALUA on SCALE??
-        # do something with iscsi service here
+        # If ALUA is configured reload the iscsitarget service (to regen config) and then start SCST
+        if self.run_call('iscsi.global.alua_enabled'):
+            if not self.run_call('service.reload', 'iscsitarget'):
+                timeout = 5
+                while not self.run_call('service.start', 'iscsitarget') and timeout > 0:
+                    logger.warning('Waiting one second to allow iscsitarget to start')
+                    sleep(1)
```
anodos325, line 737: need to decrement `timeout`

src/middlewared/middlewared/service/service_mixin.py
```diff
@@ -2,7 +2,7 @@
 
 
 class ServiceChangeMixin:
-    async def _service_change(self, service, verb):
+    async def _service_change(self, service, verb, options={}):
```
anodos325, line 5: I thing general convention would be `options=None` rather than an empty dict.

src/middlewared/middlewared/service/system_service.py
```diff
@@ -22,12 +22,12 @@ async def config(self):
         )
 
     @private
-    async def _update_service(self, old, new, verb=None):
+    async def _update_service(self, old, new, verb=None, options={}):
```
anodos325, line 25: ditto regarding `options=None`

yocalebo (changes requested):
Overall looks good, I made a few notes and asked for a minor change in the etc generation portion of the code.

src/middlewared/middlewared/etc_files/scst.conf.mako
```diff
@@ -4,6 +4,44 @@
...
+    # - Add them to copy_manager
+    # - reexport them on the same IQNs as the master, but with different
+    #   rel_tgt_id.
+    #
+    # BOTH:
+    # - Write a DEVICE_GROUP section with two TARGET_GROUPs
+    # - TARGET GROUPs and rel_tgt_id are tied to the controller,
+    #   *not* to whether it is currently the MASTER or BACKUP
+    #
+    ha_capable = middleware.call_sync("system.is_ha_capable")
```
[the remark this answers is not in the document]
yocalebo: I shouldn't have said never. What I meant is that we have customers that buy an HA capable system that is only populated with the A controller and have plans to upgrade to HA in the future. (That's what we typically see, at least)

yocalebo (approved):
Fantastic!

PR #10432, opened 2023-01-13
NAS-119707 / 23.10 / fix applying sysctl tunables on upgrade from 22.02

yocalebo:
In Angelfish we were writing the `tun_type` column in lower-case but this plugin was fixed in Bluefin and that subtle (confusing) detail was missed. This makes it so that anyone upgrading from 22.02 with sysctl tunables set will be recognized and applied on upgrade to 22.12.1+


yocalebo:
backport

bmeagherix (approved):

PR #10443, opened 2023-01-15
NAS-119681 / 23.10 / Add guest agent channel in devices

Qubad786:
## Problem

With some guests proper shutdown cannot be initiated unless libvirt is able to communicate with the guest libvirt agent.

## Solution

Allow libvirt to interact with guest agent by opening a unix channel so that proper shutdown can be initiated in guests.


yocalebo (approved):

yocalebo (changes requested):
Seems like this broke a unit test.

yocalebo (approved):

Qubad786:
backport

PR #10450, opened 2023-01-16
NAS-119664 / 23.10 / Do not show WebUiRootLogin alert if root password is not disabled

themylogin:


themylogin:
time 00:30

src/middlewared/middlewared/etc_files/web_ui_root_login_alert.py
```diff
@@ -1,10 +1,12 @@
...
 async def render(service, middleware, render_ctx):
-    if await middleware.call(
-            'privilege.always_has_root_password_enabled',
-            render_ctx['user.query'],
-            render_ctx['group.query'],
+    root_user = await middleware.call(
+        'datastore.query', 'account.bsdusers', [('username', '=', 'root')], {'get': True, 'prefix': 'bsdusr_'},
+    )
+
+    if root_user['password_disabled'] and await middleware.call(
```
yocalebo, line 6: Can you get the same root user information from render_ctx["user.query"] instead of making another db query?

yocalebo (approved):

PR #10461, opened 2023-01-17
NAS-119882 / 22.12.1 / Re-add mistakenly deleted idmap table entry

anodos325:
In commit 5265c8c a migration was written to use AUTOINCREMENT to ensure id uniqueness. In commit 85f5b97 the aforementioned migration was modified to also fix potential constraint violation in this field.

Since there was a gap between these two commits, it is impossible that the original migration without the subsequent revision and therefore the user's DB still contains the original constraint violation. This table entry is critical to the proper function of the AD plugin and since it is user-configurable, deletion cannot be repaired without manual intervention.

The practical impact of this is that AD users who had installed SCALE during our early ALPHA and BETA cycles prior to 22.02 released are broken on upgrade to 22.12.0.

This PR re-adds the DB table entry with its default value if possible. There is a chance that user had customized this range. In this case the configuration is already lost, but at least we can unbreak the idmap table and allow manual fix via webui.


src/middlewared/middlewared/alembic/versions/22.12/2023-01-17_15-29_fix-broken-idmap-table.py
```diff
@@ -0,0 +1,65 @@
...
+def upgrade():
+    # This applies a DB change that may have been skipped during SCALE alpha / beta cycle.
+    # The situation arose because we altered an existing migration between two pre-release versions
+    # rather than writing a new migration.
+    #
+    # The first option is that this is a new upgrade from Angelfish (broken) to BlueFin. In this
+    # case run the migration from NAS-111944.
+    conn = op.get_bind()
+    op.execute("""
+        UPDATE directoryservice_idmap_domain SET idmap_domain_certificate_id = NULL WHERE idmap_domain_certificate_id = ''
```
themylogin, line 28: How can it be `''`?

anodos325: Do note that this op is copied from here:
https://github.com/truenas/middleware/blob/b5d341c7ee16479dc4ad8e8f73c2e4d6d10fda6c/src/middlewared/middlewared/alembic/versions/22.02/2021-01-20_10-19_autoincrement.py#L24

I do not know the background for the investigation that led to this being included. Perhaps full root-cause was not investigated.

TL;DR, there was a bug in an earlier migration that could result in it being `''` instead of `NULL`. It wasn't caught in review. That can happen.

The above code would have fixed the issue if it had been added as a separate migration file, but instead it was appended a migration file after-the-fact, which means for some users (who were already on versions that had run the migration) the fix was never run (because migrations do not get re-run).

anodos325: That said, I don't think that this is material to the PR.

themylogin: How can I repeat the steps you are talking about here https://github.com/truenas/middleware/pull/10277/files#r1059801158 ?

anodos325: Sure. You can download old ISOs from here: https://download.truenas.com/TrueNAS-SCALE-Angelfish-BETA/
And run through the steps I mentioned.

themylogin: @anodos325 do I need to run additional steps (connect to AD domain) or just install - upgrade - upgrade and the row will be there (and then will go away)?

anodos325: IIRC no special setup is required. Install - upgrade - upgrade.

themylogin: Ok.
1. `''` value is introduced here https://github.com/truenas/middleware/blob/master/src/middlewared/middlewared/alembic/versions/12.0/2020-01-24_13-34_simplify_idmap_backends.py#L86 (should be `params.pop('certificate_id', None)`. SQLite allows to insert anything to a foreign key column when foreign_keys pragma is not set (default).
2. The first part here https://github.com/truenas/middleware/commit/85f5b97ec9a388b04ed50ff2c3c6528d0b3de994#diff-ccd1e1d3980a936667753318327c3253755ed0ab7b6684de05addd4bacbf2248R22 seems to be irrelevant. If I remove that part of the patch I can still upgrade the original offending database without any errors.
Conclusion: I was wrong in this comment https://github.com/truenas/middleware/pull/10277/files#r1059785360 and the line we are discussing has total sense.

themylogin (approved):

themylogin:
time 1:00

PR #10463, opened 2023-01-17
NAS-119359 / 23.10 / Remove ws4py

themylogin:


yocalebo (commented):
So this is a fun rabbit hole that I went down. I'm looking at these changes and saw that we're subclassing `threading.Event` on the justification that acquiring a `threading.Lock` is actually using `sem_timedwait` which is based off system clock time. This means, in theory, while waiting for a lock to be acquired a change in system clock will effect the time for acquiring the lock...

We apparently hit this bug on CORE when we were running django over 5 years ago. According to python documentation python 3.9.8 has the support for detecting and using `sem_clockwait` automatically if the gcc that gets bundled with the OS supports it. Sure enough, the python version that we ship in bullseye (3.9.2) doesn't automatically detect `sem_clockwait` so we're still susceptible to this bug. However, I've confirmed that python3.10 (which ships in next major debian release) automatically detects `sem_clockwait`. So I'd like to remove this class while you're here 😄

src/middlewared/middlewared/client/client.py
```diff
@@ -19,10 +18,12 @@
 except ImportError:
```
yocalebo, line 18: Let's also fix this import while you're here
```
from collections.abc import Callable
```

src/middlewared/middlewared/client/client.py
```diff
@@ -19,10 +18,12 @@
...
+from websocket import WebSocketApp
+from websocket._abnf import STATUS_NORMAL
+from websocket._http import connect, proxy_info
+from websocket._socket import sock_opt
 
 from . import ejson as json
-from .protocol import DDPProtocol
 from .utils import MIDDLEWARE_RUN_DIR, ProgressBar, undefined
 try:
     from libzfs import Error as ZFSError
```
yocalebo, line 29: Same with this import, let's fix it since `libzfs` should always be available

themylogin: Someone might theoretically be using this client from a machine where libzfs is not available, so it might have sense to keep it.

yocalebo: This code has rarely changed in 5+ years. If you're concerned about 3rd party consumers, then that's something they'll have to patch on their side. We never wrote this to be used outside of our ecosystem. However, if you think we'll break code that is internal then I agree to keep it.

yocalebo (changes requested):
This breaks HA in ways that I don't immediately understand. This is logged on active controller.
```
[2023/01/23 05:45:14] (WARNING) websocket.warning():66 - websocket connected
[2023/01/23 05:45:14] (WARNING) websocket.warning():66 - websocket connected
[2023/01/23 05:45:17] (WARNING) websocket.warning():66 - websocket connected
[2023/01/23 05:45:19] (WARNING) websocket.warning():66 - websocket connected
[2023/01/23 05:45:50] (WARNING) websocket.warning():66 - websocket connected

**Comment by themylogin** [comment]
time 2:00

File: src/middlewared/middlewared/logger.py [hunk 3]
```diff
@@ -17,8 +17,8 @@
 logging.getLogger('aiohttp.internal').setLevel(logging.WARN)
 # We dont need internal botocore debug logging
 logging.getLogger('botocore').setLevel(logging.WARN)
-# we dont need ws4py close debug messages
-logging.getLogger('ws4py').setLevel(logging.WARN)
+# we dont need websocket debug messages
+logging.getLogger('websocket').setLevel(logging.WARN)
```
yocalebo, line 21: This spams logs way too much for us to merge. We'll just have to bump this until upstream pushes new release to pypy.
```suggestion
logging.getLogger('websocket').setLevel(logging.CRITICAL)
```

yocalebo (changes requested):

yocalebo:
@themylogin now that https://github.com/truenas/middleware/pull/10548 has been merged, can you rebase this on latest master so I can run through test suite again?

yocalebo (approved):
No immediate regressions found in the CI on single node or HA node! Great job!

PR #10471, opened 2023-01-18
NAS-119196 / 23.10 / fix webui dashboard caching issues on HA

yocalebo:
The webUI team subscribes to `failover.disabled.reasons` but when the standby controller reboots, the last message they get on the active controller is `{"msg": "changed", "collection": "failover.disabled.reasons", "fields": {"disabled_reasons": ["REM_FAILOVER_ONGOING"]}}`. After the failover job on the standby controller completes, we need to refresh the failover status on the active controller to ensure the webUI team receives an event with an empty (hopefully) `disabled_reasons` field so that the dashboard widget(s) and the HA status icon can reflect reality. Without these changes, rebooting the standby will never show back up on the dashboard which leads to perceiving a false reality.

### DIFF
--- a/src/middlewared/middlewared/plugins/failover_/event.py
+++ b/src/middlewared/middlewared/plugins/failover_/event.py
@@ -123,7 +123,7 @@ class FailoverEventsService(Service):
             wait_id = await self.middleware.call('core.job_wait', jobid)
             await wait_id.wait(raise_error=True)
         except (CallError, KeyError):
-            # `CallErro` means the failover job didn't complete successfully
+            # `CallError` means the failover job didn't complete successfully
             # but we still want to refresh status in this scenario
             # `KeyError` shouldn't be possible but there exists a hypothetical
             # race condition...but we still want to refresh status


src/middlewared/middlewared/plugins/failover_/event.py
```diff
@@ -114,6 +115,34 @@ async def background(self):
...
 
+    async def refresh_failover_status(self, jobid, event):
+        # this is called in a background task so we need to make sure that
+        # we wait on the current failover job to complete before we try
+        # and update the failover status
+        try:
+            wait_id = await self.middleware.call('core.job_wait', jobid)
+            await wait_id.wait(raise_error=True)
+        except (CallError, KeyError):
+            # `CallErro` means the failover job didn't complete successfully
```
anodos325, line 126: Typo?

anodos325 (approved):

src/middlewared/middlewared/plugins/failover_/event.py
```diff
@@ -127,14 +156,15 @@ def event(self, ifname, event):
...
+            job = self._event(ifname, event)
+            return job
         except IgnoreFailoverEvent:
             refresh = False
         finally:
             # refreshing the failover status can cause delays in failover
             # there is no reason to refresh it if the event has been ignored
             if refresh:
-                self.run_call('failover.status_refresh')
+                self.middleware.create_task(self.refresh_failover_status(job.id, event))
```
themylogin, line 167: If another exception happens (not IgnoreFailoverEvent) then `refresh` will be `True` but `job` won't be defined

PR #10475, opened 2023-01-18
NAS-119902 / 23.10 / Remove some legacy handling from NFS and DS

anodos325:
The NFS plugin has overly-complicated check for
whether we need to change the `use_default_domain` parameter for the AD plugin. This PR simplifies and removes potentially unnecessary winbindd restarts.

Likewise, the directoryservices plugin has a check for kerberos method. This was due to legacy AD
configuration support in TrueNAS 11.3. Prior to
TrueNAS 11.3, the controllers were joined to AD
separately (two computer objects in AD vs a single one for the appliance). When we transitioned to
having a single computer account, we left legacy
shim in place so that users upgrading wouldn't be
immediately broken.

Over time, support migrated legacy systems to the
new unified design. Since this configuration has
never existed on SCALE, it is safe to remove
whatever vestiges of it still exist in the code base.

### DIFF
--- a/src/middlewared/middlewared/plugins/directoryservices.py
+++ b/src/middlewared/middlewared/plugins/directoryservices.py
@@ -466,22 +466,12 @@ class DirectoryServices(Service):
         has_secrets = await self.middleware.call("directoryservices.secrets_has_domain", workgroup)
 
         if ad_enabled and not has_secrets:
-            kerberos_method = await self.middleware.call("smb.getparm", "kerberos method", "GLOBAL")
             self.logger.warning("Domain secrets database does not exist. "
                                 "Attempting to restore.")
-            ok = await self.middleware.call("directoryservices.restore_secrets")
-            if not ok:
+
+            if not await self.middleware.call("directoryservices.restore_secrets"):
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/directoryservices.py
```diff
@@ -469,20 +469,11 @@ async def initialize(self, data=None):
             kerberos_method = await self.middleware.call("smb.getparm", "kerberos method", "GLOBAL")
             self.logger.warning("Domain secrets database does not exist. "
                                 "Attempting to restore.")
-            ok = await self.middleware.call("directoryservices.restore_secrets")
-            if not ok:
+
+            if not await self.middleware.call("directoryservices.restore_secrets")
```
yocalebo, line 473:
```suggestion
            if not await self.middleware.call("directoryservices.restore_secrets"):
```

anodos325: Whoops

yocalebo:
retest this please

yocalebo (approved):

PR #10485, opened 2023-01-18
NAS-119605 / 22.12.2 / sudo rework + replication sudo

themylogin:
* Change our sudo commands API. Previously we had a flag to enable sudo for the user/group,
  a list of allowed commands (an empty list meaning all commands are allowed) and a flag
  that makes all these commands being allowed to be executed without entering password.

  Now we have a nicer API that has a list of commands that can be executed as root (an
  empty list meaning no commands can be executed and `["ALL"]` meaning all commands can be
  executed) and another list for commands that can be executed without entering password.

  Not only this API is nicer, it also allows to allow executing certain commands without
  entering password and make others still require password (or forbid them completely).
  This is required for sudo replication and is the only real reason why we change our API
  in a stable branch.
* Allow using zettarepl `sudo` option to perform replication using sudo. Also allow passing `sudo` to our semi-automatic replication setup methods so it can add necessary commands to the `sudo_commands_nopasswd` option of the remote user profile.


themylogin:
time 12:00

yocalebo:
retest this please

yocalebo:
retest this please

anodos325 (approved):

PR #10494, opened 2023-01-19
NAS-119919 / 23.10 / Fix regex for retrieving valid usb passthrough choices

sonicaj:
This commit fixes an issue where if a USB hub is being used, the naming scheme for the devices change from usb_5_1 to usb_5_1_1 / usb_5_1_2.

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/usb.py
+++ b/src/middlewared/middlewared/plugins/vm/usb.py
@@ -10,7 +10,7 @@ from .devices.usb import USB_CONTROLLER_CHOICES
 from .utils import get_virsh_command_args
 
 
-RE_VALID_USB_DEVICE = re.compile(r'^usb_\d+_\d+$')
+RE_VALID_USB_DEVICE = re.compile(r'^usb_\d+_\d+(_\d)*$')
 
 
 class VMDeviceService(Service):


yocalebo (approved):

Qubad786 (approved):
Have you verified the passthrough itself still works in such cases?

sonicaj:
Yes @Qubad786, @william-gr tested it for me ;)

src/middlewared/middlewared/plugins/vm/usb.py
```diff
@@ -10,7 +10,7 @@
 from .utils import get_virsh_command_args
 
 
-RE_VALID_USB_DEVICE = re.compile(r'^usb_\d+_\d+$')
+RE_VALID_USB_DEVICE = re.compile(r'^usb_\d+_\d+(_\d)*$')
```
william-gr, line 13: Forgot a + sign? (_\d+)

sonicaj:
```
bool(re.findall(r'^usb_\d+_\d+(_\d)*$', 'usb_5_1'))
True
bool(re.findall(r'^usb_\d+_\d+(_\d)*$', 'usb_5_1_'))
False
bool(re.findall(r'^usb_\d+_\d+(_\d)*$', 'usb_5_1_1'))
True
```

It is good as the whole group has `*` so `_\d` will be ensured but i think i should make it a non-capturing group. Will make that change :P

william-gr:
```
>>> bool(re.findall(r'^usb_\d+_\d+(_\d)*$', 'usb_5_1_15'))
False
```

sonicaj: Ah shoot :D Like that 😂

PR #10500, opened 2023-01-19
NAS-119905 / 13.0 / Remove duplicate atrun cron job

freqlabs:
There is an atrun job in /etc/cron.d/at by default since 2017.


freqlabs:
time 1:30

amotin (approved):
Makes sense.  It seems we've missed to update this file 5 years ago when migrated to 12.

PR #10513, opened 2023-01-20
TE-167 / 22.12.1 / Fixing changing disk['size'] from string to int in optimal_disk_usage

ericbsd:
To avoid this error `TypeError: unsupported operand type(s) for -: 'str' and 'int'`


yocalebo (changes requested):
`disk['size']` should already be an integer

yocalebo:
huh....it's not an integer in 22.12 but is in master....

yocalebo (approved):

PR #10514, opened 2023-01-21
NAS-119933 / 23.10 / Enhance validation for licensed systems in apps

sonicaj:
This commit adds changes preventing licensed systems to create more catalogs or update their preferred trains to anything other then enterprise train.

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/update.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/update.py
@@ -174,6 +174,13 @@ class CatalogService(CRUDService):
                 f'{schema}.preferred_trains',
                 'At least 1 preferred train must be specified for a catalog.'
             )
+        if await self.middleware.call(
+            'kubernetes.license_active'
+        ) and data['preferred_trains'] != [OFFICIAL_ENTERPRISE_TRAIN]:
+            verrors.add(
+                f'{schema}.preferred_trains',
+                f'Licensed systems can only consume {OFFICIAL_ENTERPRISE_TRAIN!r} train'
+            )
 
         verrors.check()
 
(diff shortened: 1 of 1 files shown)


Ornias1993:
This is pretty-much a complete and utter dick-head move towards other projects wanting to release enterprise-level Apps.

This should be "the default" for licensed systems that want to use Apps, with a license flag to override (with thorough warnings that systems will not get support form iX for App issues if the license flag is set).

At this point, it basically invalidates **ALL** the work we, at @truecharts have been putting into preparing for SMB/Enterprise clients/users. It's simply not-done and brings up the question why I didn't even get a headsup when LITERALLY asking morgan about Enterprise train plans a 1-2 months ago.

yocalebo:
Do you know what this check actually does?

Ornias1993:
> Do you know what this check actually does?

Wait, deleted my comment… because the code shows it DOES offer an “enabled” flag. >.<

yocalebo:
Right, so all this does is prevent our customers that have purchased officially supported hardware from changing the train on the chart. Idea is to keep them from running 3rd party apps since we are contractually obligated to supporting them.

Ornias1993:
> Right, so all this does is prevent our customers that have purchased officially supported hardware from changing the train on the chart. Idea is to keep them from running 3rd party apps since we are contractually obligated to supporting them.

Okey, so it DOES prevent third party catalogs after all?

yocalebo:
Why do you care about that? Are you expecting to offer paid support for truechart apps to our enterprise customers?

Ornias1993:
> Why do you care about that? Are you expecting to offer paid support for truechart apps to our enterprise customers?

TLDR: yes, our at least to our enterprise train. Lets assume the target audience has an overlap for now. details are still a fair bit TBD, when it comes to scope and pricing though…

An idea originally brought up by Morgen and (re)discussed only 1-3 months ago.

Ornias1993 (commented):
I've thoroughly checked the code by now.
This PR should not cause needless issues, even for TrueCharts.

- The 'train' limitation only required trains for enterprise users to be explicitly called "enterprise", which is a good way to ensure train creators explicitly need to define enterprise users.
-  The JAILS license flag, enables all catelogs regardless of origin


Note:
In the future an additional flag might need to be added to seperate "JAILS" (AKA: DIY applications and external catalogs) from the iX Official Enterprise Apps.

But after latest update by @sonicaj I see no issues having this merged anymore.
LGTM/Approved.

yocalebo (approved):

PR #10515, opened 2023-01-22
NAS-119932 / 23.10 / Allow tying app version to SCALE releases

sonicaj:
This PR adds functionality to be able to specify app version to SCALE releases so for example if an app has specified that it can only be installed on Bluefin or above, Angelfish users should not be able to install or upgrade to that app version.
This essentially adds a baseline version which devs can specify so systems on that version or above that version can proceed with installation or upgrade of the app version in question.

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/items_util.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/items_util.py
@@ -7,7 +7,7 @@ from catalog_validation.items.items_util import (
 
 from middlewared.plugins.chart_releases_linux.schema import construct_schema
 from middlewared.plugins.update_.utils import can_update
-from middlewared.utils import manifest_version
+from middlewared.utils import sw_version
 
 
 def get_item_default_values(version_details: dict) -> dict:
@@ -30,8 +30,8 @@ def minimum_scale_version_check_update_impl(
         not check_supported_key or version_details['supported']
     ):
         try:
-            if manifest_version() != version_details['chart_metadata']['minimum_scale_version'] and not can_update(
-                version_details['chart_metadata']['minimum_scale_version'], manifest_version()
+            if sw_version(False) != version_details['chart_metadata']['minimum_scale_version'] and not can_update(
+                version_details['chart_metadata']['minimum_scale_version'], sw_version(False)
             ):
                 return False, False
         except Exception:
--- a/src/middlewared/middlewared/plugins/catalogs_linux/items_util.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/items_util.py
@@ -7,7 +7,7 @@ from catalog_validation.items.items_util import (
 
 from middlewared.plugins.chart_releases_linux.schema import construct_schema
 from middlewared.plugins.update_.utils import can_update
-from middlewared.utils import sw_version
+from middlewared.utils import sw_info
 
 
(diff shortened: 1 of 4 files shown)


Ornias1993 (commented):
LGTM, this is extremely usefull :)

Altough you should really start providing docs for this in the `charts` repo in the docs folder there. Because you should provide devdocs for this so people can actually use it without being platform experts.

src/middlewared/middlewared/utils/__init__.py
```diff
@@ -254,6 +255,11 @@ def sw_version():
     return VERSION
 
 
+@functools.cache
+def manifest_version():
```
yocalebo, line 259: Please check the various system plugins. I believe we already have a method that returns the information you want. If there isn't, please rename this function to what it's actually doing. It's reading the manifest file to determine the scale release, not return the manifest file version?

yocalebo (changes requested):

src/middlewared/middlewared/utils/__init__.py
sonicaj: We don't unfortunately have that method already - we have a method which returns the verbose name or Full Name as we call it.
I'll rename the method to reflect reality.

yocalebo: With https://github.com/truenas/middleware/commit/e42cd6e1da74b1bace5aa83a68db9b6dd8b02c31 merged, you can call `system.version False` so you don't have to add this helper function anymore.

src/middlewared/middlewared/plugins/catalogs_linux/items_util.py
```diff
@@ -1,23 +1,59 @@
...
+) -> typing.Tuple[bool, bool]:
+    # `check_supported_key` is used because when catalog validation returns the data it only checks the
+    # missing features and based on that makes the decision. So if something is not already supported
+    # we do not want to validate minimum scale version in that case. However, when we want to report to
+    # the user as to why exactly the app version is not supported, we need to be able to make that distinction
+    if version_details['healthy'] and version_details['chart_metadata'].get('minimum_scale_version') and (
+        not check_supported_key or version_details['supported']
+    ):
+        try:
+            if sw_version(False) != version_details['chart_metadata']['minimum_scale_version'] and not can_update(
```
yocalebo, line 33:
```suggestion
            if sw_version(fullname=False) != version_details['chart_metadata']['minimum_scale_version'] and not can_update(
```

src/middlewared/middlewared/plugins/update_/install_linux.py
```diff
@@ -27,15 +28,12 @@ def install_manual_impl(self, job, path, dest_extracted, options=None):
...
-        with open(SCALE_MANIFEST_FILE) as f:
-            old_manifest = json.load(f)
-
         progress_callback(0, "Reading update file")
         with mount_update(path) as mounted:
             with open(os.path.join(mounted, "manifest.json")) as f:
                 manifest = json.load(f)
 
-            old_version = old_manifest["version"]
+            old_version = sw_version(False)
```
yocalebo, line 36:
```suggestion
            old_version = sw_version(fullname=False)
```

yocalebo (changes requested):

yocalebo (approved):

PR #10516, opened 2023-01-22
Remove old logic to remove daemonsets etc on post k8s start

Qubad786:
## Context

On angelfish we had an issue where k3s was not deleting daemonsets it deployed against built in load balancer klipper based services even when the service was deleted. That issue is not happening anymore in k3s currently and we are removing code where we tried to work around the upstream issue.


yocalebo (approved):

Ornias1993 (commented):
Good call, totally forgot this TODO till this PR.
LGTM.

PR #10518, opened 2023-01-23
NAS-119975 / 23.10 / Fix filtering service.query based on state

anodos325:
We were passing query-filters directly to datastore.query without checking whether they were stored in the table.

This caused query-filters such as [['state', '=', 'RUNNING']] to fail with KeyError. This PR adds handling for these cases by performing unfiltered datastore.query in cases where filter cannot be reasonable applied at SQL-level.

This PR also adds a performance optimization to skip the service state check via an `extra` option.

This PR also slightly refactors service.query to utilize a `datastore_extend` and `datastore_extend_context` to perform service state retrieval.

### DIFF
--- a/src/middlewared/middlewared/plugins/service.py
+++ b/src/middlewared/middlewared/plugins/service.py
@@ -10,7 +10,7 @@ from middlewared.schema import accepts, Bool, Dict, Int, List, Ref, returns, Str
 from middlewared.service import filterable, CallError, CRUDService, periodic, private
 from middlewared.service_exception import MatchNotFound
 import middlewared.sqlalchemy as sa
-from middlewared.utils import filter_list
+from middlewared.utils import filter_list, filter_getattrs
 
 
 class ServiceModel(sa.Model):
@@ -25,6 +25,9 @@ class ServiceService(CRUDService):
 
     class Config:
         cli_namespace = "service"
+        datastore_prefix = 'srv_'
+        datastore_extend = 'service.service_extend'
+        datastore_extend_context = 'service.service_extend_context'
 
     ENTRY = Dict(
         'service_entry',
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/service.py
```diff
@@ -74,16 +71,43 @@ def result(task):
...
+        """
+        default_options = {
+            'prefix': self._config.datastore_prefix,
+            'extend': self._config.datastore_extend,
+            'extend_context': self._config.datastore_extend_context
+        }
 
-            return entry
+        if set(filter_getattrs(filters)) & {'state', 'pids'}:
+            services = await self.middleware.call('datastore.query', 'services.services', [], default_options)
```
themylogin, line 107: Why do you discard other options (e.g. `get`) and why is this a special case?

themylogin (changes requested):

src/middlewared/middlewared/plugins/service.py
anodos325: They're not discarded. They're picked up in the next line.

anodos325:
```
root@carl[/mnt/dozer/middleware/src/middlewared]# midclt call service.query '[["state", "=", "RUNNING"]]' '{"get": true}' | jq
{
  "id": 4,
  "service": "cifs",
  "enable": true,
  "state": "RUNNING",
  "pids": [
    533202
  ]
}
```
Seems like `get` is working as expected in this case. How can we apply `options` correctly if `filters` are not passed to datastore? Filters are not passed to datastore plugin because they fail with KeyError.

themylogin: Still will not contain `get` or other options and I am still don't understand why is this a special case worth special handling.

themylogin (changes requested):

themylogin (approved):

PR #10519, opened 2023-01-23
NAS-119977 / 23.10 / remove get_app_version method and replace with sw_info

yocalebo:
This does a few things:
1. remove `osc.get_app_version` and move the same logic to `utils/__init__.py:sw_info`
2. cache `sw_info` using `functools.cache` decorator
3. remove various `global` variables from `utils/__init__.py`
4. make `sw_buildinfo`, `sw_version` and `sw_version_is_stable` call `sw_info`
5. make `sw_version` take a kwarg (`fullname`) which when set to False, will return the software version without the `TrueNAS-SCALE-` prefix.
6. update `system.version` public API endpoint to accept a new boolean argument that makes the kwarg for `sw_info` default to `True` as to not change any functionality by default.
7. `mail.py` plugin was making an unnecessary websocket call just to lower-case `TrueNAS` but this string is always static so remove the websocket call and define a static string
8. `support.py` and `update.py` were calling builtin methods like `split` and `join` just to add `SCALE` prefix to the string. Instead use a f-string to prepend `SCALE-` to the string since it's always static.

There should be no change in functionality.

### DIFF
--- a/src/middlewared/middlewared/plugins/support.py
+++ b/src/middlewared/middlewared/plugins/support.py
@@ -17,6 +17,7 @@ from middlewared.service import CallError, ConfigService, job, ValidationErrors
 import middlewared.sqlalchemy as sa
 from middlewared.utils.network import INTERNET_TIMEOUT
 from middlewared.validators import Email
+from middlewared.utils import PRODUCT
 
 ADDRESS = 'support-proxy.ixsystems.com'
 
@@ -223,7 +224,7 @@ class SupportService(ConfigService):
             if i not in data:
                 raise CallError(f'{i} is required', errno.EINVAL)
 
-        data['version'] = f'SCALE-{await self.middleware.call("system.version", False)}'
+        data['version'] = f'{PRODUCT}-{await self.middleware.call("system.version", False)}'
         debug = data.pop('attach_debug')
 
         type_ = data.get('type')
--- a/src/middlewared/middlewared/plugins/update.py
+++ b/src/middlewared/middlewared/plugins/update.py
@@ -2,6 +2,7 @@ from middlewared.schema import accepts, Bool, Dict, Str
 from middlewared.service import job, private, CallError, Service
 import middlewared.sqlalchemy as sa
 from middlewared.plugins.update_.utils import UPLOAD_LOCATION
+from middlewared.utils import PRODUCT
 
 from datetime import datetime
 import enum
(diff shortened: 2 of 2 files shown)


anodos325 (approved):

src/middlewared/middlewared/plugins/mail.py
```diff
@@ -283,11 +283,9 @@ def send_raw(self, job, message, config):
         else:
             interval = timedelta(seconds=interval)
 
-        sw_name = self.middleware.call_sync('system.version').split('-', 1)[0]
-
         channel = message.get('channel')
         if not channel:
-            channel = sw_name.lower()
+            channel = 'truenas'
```
bmeagherix, line 288: You PR comment says "define a static string".  Perhaps that should actually be a definition in a file that gets imported and used here, so that we don't have constants sprinkled throughout our code.

bmeagherix (changes requested):

src/middlewared/middlewared/plugins/mail.py
yocalebo: Agree. Thanks, I've pushed a fix addressing your review.

src/middlewared/middlewared/plugins/support.py
```diff
@@ -223,7 +223,7 @@ async def new_ticket(self, job, data):
             if i not in data:
                 raise CallError(f'{i} is required', errno.EINVAL)
 
-        data['version'] = (await self.middleware.call('system.version')).split('-', 1)[-1]
+        data['version'] = f'SCALE-{await self.middleware.call("system.version", False)}'
```
themylogin, line 226: Is that `PRODUCT` constant is really necessary? If yes, why don't we use it here?

src/middlewared/middlewared/plugins/system/product.py
```diff
@@ -74,13 +74,16 @@ async def product_name(self):
         """
         return "TrueNAS"
 
-    @accepts()
+    @accepts(Bool('fullname', default=True))
     @returns(Str('truenas_version'))
-    def version(self):
-        """
-        Returns software version of the system.
+    def version(self, fullname):
```
themylogin, line 79: Should we offer `version` and `full_version` methods instead? This returns completely different things depending on a boolean flag

bmeagherix (approved):

src/middlewared/middlewared/plugins/support.py
yocalebo: Just missed in it the original PR. I'll update to use it here.

src/middlewared/middlewared/plugins/system/product.py
themylogin: Ok, `version` and `version_number`

yocalebo: If it's a CUSTOM or nightly image then it won't be the `version_number`, it'll be `23.10-MASTER-timestamp-here` same thing for custom.

PR #10523, opened 2023-01-23
NAS-119983 / 23.10 / dont add pmem devices to smartd.conf

yocalebo:


anodos325 (approved):

themylogin (changes requested):
* This is the method to list disks that support SMART self-testing (for scheduling self-test). It is not related to smartd.conf generation.
* nvme devices still do not support smart self-testing, don't remove that condition

yocalebo:
> * This is the method to list disks that support SMART self-testing (for scheduling self-test). It is not related to smartd.conf generation.
> * nvme devices still do not support smart self-testing, don't remove that condition

sigh, so complex when self-testing and monitoring smart attributes typically infer the other is true. Thanks for pointing out, I'll fix...

themylogin (approved):

PR #10537, opened 2023-01-24
NAS-119998 / 23.10 / Allow multiple custom portals

Ornias1993:
This PR adds support for defining multiple (user and catalog definable) portals in values.yaml/questions.yaml.

It does not touch current code and creates a new dict in values.yaml/questions.yaml to ensure backwards compatibility with older versions of ix-chart.

Basically a quick and dirty example of new layout in values.yaml (or after rendering questions.yaml) will be:
```
iXPortals:
  open:
    enabled: true    # So users/catalog can dynamically enable and disable catalogs
    protocol: http   # Protocol used for the URL
    host: "192.168.10.1"   # Sets the host used for the URL, defaults to nodeIP if not set
    port: 80   # port used for url generation
  admin:
    enabled: true
    protocol: https
    path: "/admin"
    # In this case the host used would be the nodeIP
```

Most of the above settings are optional and all non-bool settings, are parsed to strings.
It also sures the url is sanitised when ports 80 or 443 are used, as that's what browsers and users expect.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
@@ -107,7 +107,6 @@ class ChartReleaseService(Service):
                 port = ''
             else:
                 port = f':{port}'
-
             stored_portals[name] = [f'{protocol}://{host}{port}/{path}']
 
         return stored_portals
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
@@ -48,6 +48,12 @@ class ChartReleaseService(Service):
         if not os.path.exists(questions_yaml_path):
             return {}
 
+        # If portals are defined in values.yaml/questions.yaml parse them
+        if len(release_data['config'].get('iXPortals')):
+            portals = release_data['config'].get('iXPortals')
+            return self.get_ixportals(portals, node_ip)
+
+        # TODO: remove some time after a release with NAS-119932 (limit app ver to os var) included
         if release_data['chart_metadata']['name'] == 'ix-chart':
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
```diff
@@ -83,6 +86,27 @@ def get_ix_chart_portal(self, release_data, node_ip):
...
+        portals = release_data['config'].get('portals')
+        
+        if not portals:
+            return {}
+        stored_portals = {}
+        for portal_config in portals:
+            if not portal_config.get('enabled'):
+                return {}
+            host = node_ip if portal_config['useNodeIP'] else portal_config['host']
+            if (portal_config['port'] == 80 AND portal_config['protocol'] == 'http') OR (portal_config['port'] == 443 AND portal_config['protocol'] ==  'https') OR (not portal_config['port']):
```
yocalebo, line 101: Just a note but `AND` and `OR` aren't valid python operators.

Ornias1993: Yeah, I moved it away from draft too soon, because for some how github didn't show the CI.

Ornias1993: Cleaned all-of-it-up, thanks for the headsup!

Ornias1993:
@sonicaj if we can have this reviewed and merged, @stavros-k and I can also implement this in the new common chart from the start.

Ornias1993:
@sonicaj Reworking into a dict as discussed.
Putting it back to draft untill done :)

yocalebo:
This should have been added in this PR: https://github.com/truenas/middleware/pull/10650

Ornias1993:
@yocalebo sorry, totally forgot to close earlier, thanks for closing :(

PR #10546, opened 2023-01-24
NAS-120003 / 23.10 / Add check for system.ready at beginning CI

anodos325:
There are indications during some custom
CI jobs that tests are running before system
is ready. This will help us identify when this
is an issue.


ericbsd:
Could we wait for system.ready is True.

yocalebo (approved):

PR #10548, opened 2023-01-25
NAS-119959 / 23.10 / fix HA integration suite

yocalebo:
On BHYVE HA VMs used to run our integration test suite, the boot drive does not support SMART. So `failover.status != MASTER` is always true (since we're initializing HA for the first time AND this is called too early in boot process). This means we will generate an empty `/etc/smartd.conf` which causes a cascade of failures in other parts of our code. Migrating system dataset, formatting a disk etc because those methods try to restart smartd service. Since we generate smartmontools with `-q nodev` it will exit with a returncode and stop the service and we raise an error.

To fix this, set it to `-q never` so that if we try to start smartd with an empty config file OR all the devices in the config file are not able to be monitored via smart, the service will still start, not return a returncode and will continue to monitor the config file for any changes.

### DIFF
--- a/src/middlewared/middlewared/etc_files/default/smartmontools.mako
+++ b/src/middlewared/middlewared/etc_files/default/smartmontools.mako
@@ -9,4 +9,4 @@
 #enable_smart="/dev/hda /dev/hdb"
 
 # uncomment to pass additional options to smartd on startup
-smartd_opts="--interval=${config["interval"] * 60} -q nodev"
+smartd_opts="--interval=${config["interval"] * 60} -q never"
--- a/src/middlewared/middlewared/etc_files/smartd.py
+++ b/src/middlewared/middlewared/etc_files/smartd.py
@@ -63,6 +63,11 @@ def get_smartd_schedule(disk):
     ])
 
 
+def write_config(config):
+    with open("/etc/smartd.conf", "w") as f:
+        f.write(config)
+
+
 async def render(service, middleware):
     smart_config = await middleware.call("datastore.query", "services.smart", [], {"get": True})
 
@@ -98,7 +103,4 @@ async def render(service, middleware):
     for disk in disks:
         config += get_smartd_config(disk) + "\n"
 
-    path = "/etc/smartd.conf"
-
-    with open(path, "w") as f:
-        f.write(config)
+    await middleware.run_in_thread(write_config, config)
--- a/tests/api2/test_001_ssh.py
+++ b/tests/api2/test_001_ssh.py
@@ -105,9 +105,10 @@ def test_00_firstboot_checks():
     for srv in services:
         if srv['service'] == 'smartd':
             assert srv['enable'] is True, str(srv)
-            if not ha:
-                # On our HA VM (bhyve) drives are S.M.A.R.T.-capable so it can start successfully.
-                assert srv['state'] == 'STOPPED', str(srv)
+            # SMART is started with "-q never" which means it should
+            # always start in all circumstances
+            # (even if there is an invalid (or empty) config)
+            assert srv['state'] == 'RUNNING', str(srv)
         else:
             assert srv['enable'] is False, str(srv)
             assert srv['state'] == 'STOPPED', str(srv)


themylogin:
Why rollback useful changes only because of the testing system quirks? Can't we just ignore smartd startup error on bhyve?

yocalebo:
> Why rollback useful changes only because of the testing system quirks? Can't we just ignore smartd startup error on bhyve?

Because we're not hiding legitimate errors if we do that? The problem is that we're generating an empty /etc/smartd.conf file and still trying to restart the service in that scenario.

yocalebo:
Also, this logic will never work when an HA system is being setup initially. `failover.status != MASTER` will always be true because the end-user has to setup, networking, the license, a zpool etc

themylogin:
> this logic will never work when an HA system is being setup initially

Do we need S.M.A.R.T. to work when HA system is being set up initially?

yocalebo:
> > this logic will never work when an HA system is being setup initially
> 
> Do we need S.M.A.R.T. to work when HA system is being set up initially?

We need SMART to work on HA systems once the system is configured appropriately.

themylogin:
> The problem is that we're generating an empty /etc/smartd.conf file and still trying to restart the service in that scenario.

It will never be empty on a real HA hardware, won't it? And on bhyve we can just skip the failing part. I don't understand what legitimate errors will be hidden in that case.

themylogin:
> We need SMART to work on HA systems once the system is configured appropriately.

Does it not work now?

yocalebo:
>

yocalebo:
> > We need SMART to work on HA systems once the system is configured appropriately.
> 
> Does it not work now?

It doesn't work now. We generate an empty /etc/smartd.conf file and when we do that, many other plugins fail to run because we restart smartd _everywhere_ and it fails.

yocalebo:
> > We generate an empty /etc/smartd.conf file
> 
> Do we do it on the real hardware?

I don't know. Did you test on real hardware when you made the changes? I recall that you did test it for quite awhile on real hardware. It seems, however, we didn't pay attention to the test suite which is what this breaks.

themylogin:
> Did you test on real hardware when you made the changes? 

Yes

And I still do not understand why do we allow the fact that SMART can't start on a VM prevent us from monitoring boot drives on backup node. Not raising any smart service start exceptions on BHYVE seems like a fine enough solution to me.

yocalebo:
> > Did you test on real hardware when you made the changes?
> 
> Yes
> 
> And I still do not understand why do we allow the fact that SMART can't start on a VM prevent us from monitoring boot drives on backup node. Not raising any smart service start exceptions on BHYVE seems like a fine enough solution to me.

Because the bhyve HA VMs have real disks being passthrough that do support smart.

themylogin:
> Because the bhyve HA VMs have real disks being passthrough that do support smart.

We just don't use them (temporarily) or why is this not working?

yocalebo:
> > Because the bhyve HA VMs have real disks being passthrough that do support smart.
> 
> We just don't use them (temporarily) or why is this not working?

Because we're generating an empty /etc/smartd.conf file which breaks numerous other parts of our middleware. Including, but not limited to, migrating system dataset after zpool creation, formatting a disk, etc.

themylogin:
> Because we're generating an empty /etc/smartd.conf

Why is it empty if the disks support SMART?

yocalebo:
> > Because we're generating an empty /etc/smartd.conf
> 
> Why is it empty if the disks support SMART?

Because `failover.status != MASTER` is always true. We generate etc before a failover job completes AND it'll never be true on an HA system that isn't fully setup (no license, no zpool, no networking)

themylogin:
> Because failover.status != MASTER is always true. We generate etc before a failover job completes AND it'll never be true on an HA system that isn't fully setup (no license, no zpool, no networking)

How does this lead to smartd config file being empty? We include boot drives in that case.

yocalebo:
> > Because failover.status != MASTER is always true. We generate etc before a failover job completes AND it'll never be true on an HA system that isn't fully setup (no license, no zpool, no networking)
> 
> How does this lead to smartd config file being empty? We include boot drives in that case.

The boot drive on a BHYVE HA system does not support smart. So `ensure_smart_enabled` returns False for the disk

themylogin:
Now I get it :) Then the solution is to use `-q never` in the same condition with bhyve in `smartmontools.mako`

yocalebo:
> Now I get it :) Then the solution is to use `-q never` under the same condition with bhyve in `smartmontools.mako`

I thought you already did this in commit: c431f4a4b248a0a397fecf8cbcfed875dd729759 but I'm assuming `-q nodev` doesn't take into account if the config file is empty.

themylogin:
`-q nodev` is "exit when config file is empty", what I am offering is "stay alive if we are on bhyve non-master node even if config file is empty"

yocalebo:
> `-q nodev` is "exit when config file is empty", what I am offering is "stay alive if we are on bhyve non-master node even if config file is empty"

Yeah not according to documentation. ```Exit if there are no devices to monitor, or if any errors are found at startup in the configuration file.```

themylogin:
When config file is empty, there are no devices to monitor

yocalebo:
> When config file is empty, there are no devices to monitor

tested on this box and the only difference is that `-q never` doesn't exit with a returncode and starts. `-q nodev` exits with a returncode and logs an error. I'll make the change.

themylogin:
With `-q never` the test that expects SMART to not be running on a non-HA testing VM will fail

yocalebo:
@themylogin I've addressed the concerns, fixed and verified the integration test works, and changed the git commit message to reflect why we decided on `-q never`. Should be ready for review.

themylogin (approved):
LGTM! But we also need to ask QE to update manual test cases e.g. https://ixsystems.atlassian.net/browse/NAS-119029

yocalebo:
> LGTM! But we also need to ask QE to update manual test cases e.g. https://ixsystems.atlassian.net/browse/NAS-119029

Thanks, i'll let them know!

PR #10555, opened 2023-01-25
NAS-119918 / 22.12.2 / `ssh_adminlogin`

themylogin:
Disable the builtin `admin` account from logging in using password authentication over SSH by default. This mirrors the behavior for the `root` account.

Because the `truenas-set-authentication-method.py` runs _AFTER_ the alembic migrations, the alembic migration is essentially a NO-OP by design. This does 2 things:

1. prevents the built-in admin account that is created on fresh installations (if user chooses this option) from logging in via SSH using a password by default (to follow root account behavior)
2. If a user has created a local user named `admin` and upgrades to this release, it should not touch the behavior for that account.

This is a stop-gap solution for the 22.12 release but a more thorough solution will be implemented in 23.10 release.


src/middlewared/middlewared/etc_files/local/ssh/sshd_config.mako
```diff
@@ -108,3 +108,7 @@ ChallengeResponseAuthentication yes
 PrintMotd no
 % endif
 SetEnv LC_ALL=C.UTF-8
+% if not ssh_config['adminlogin']:
+Match User admin
```
william-gr, line 112: Hmm, cant the admin user be called anything?

themylogin: @william-gr no, for the sake of simplicity it can't. In 23.10 we'll have per-user/group option, but here we'll only replicate root functionality for the built-in `admin` user we allow to create using 22.12 CD installer.

yocalebo (approved):

anodos325 (approved):

themylogin:
time 4:00

PR #10564, opened 2023-01-26
NAS-120040 / 23.10 / Allow RO access to clustered config if DB healthy

anodos325:
If cluster is overall unhealthy (nodes missing), some of the nodes may still have a healthy copy of our DBs. Since TC wants to present users with at least some config info in this unhealthy state, we can check on per-db basis whether the data is healthy and return what we have.

Updates will still fail as expected.


anodos325:
Needs tests

yocalebo (approved):

PR #10565, opened 2023-01-26
NAS-120042 / 23.10 / Add test for ACL inheritance on new datasets via SMB

anodos325:


bmeagherix (approved):
LGTM (although I'm not yet that familiar with our ACL impl)

PR #10568, opened 2023-01-27
NAS-120063 / 23.10 / Fix sorting in disk_by_zfs_guid

anodos325:
Earlier versions of filter_list method contained
a bug where specifying `get` with `order_by` would short-circuit before sorting occured. So this
particular method would always return first
instance in the database. When the filter_list bug was fixed, it caused a regression in disk_by_zfs_guid because the `expiretime` key has intentionally
inconsistent types between database entries.

This PR changes from offloading the filtering to
filter_list to performing it in-method.


yocalebo (approved):
Fantastic!

PR #10577, opened 2023-01-27
NAS-120071 / 23.10 / Add order_by nulls_first, nulls_last

anodos325:
This PR adds "nulls_first:" and "nulls_last:"
handling to filter_list order_by option in order
to make options behavior match what we do with
datastore operations. It also fixes API documentation
errors related to this query-option.


themylogin (approved):

yocalebo (approved):
Significantly easier to understand what's going on at a glance.

PR #10604, opened 2023-01-30
NAS-120106 / 23.10 / Add some basic detection of misconfigured DNS

anodos325:
During CI runs it was noted that some jenkins users weren't properly cleaning up after their runs had
completed. This PR expands on previous DNS cleanup "tests" to also look up stale entries via reverse pointers for the current VM's IP address. This cleanup failed due to misconfigured reverse zone in the AD that
the jenkins pipeline was using, and so more robust detection of broken reverse zones was added to
the AD plugin to avoid dynamic updates to networks where this is broken.


yocalebo (approved):

ericbsd (approved):
tests LGTM

yocalebo:
retest this please

PR #10610, opened 2023-01-30
NAS-120129 / 23.10 / Simplify nslcd state handling

anodos325:
Move nslcd service handling into standard methods of controlling systemd services, and restart nslcd if needed when checking ldap started state. Middleware LDAP state is determined by using libldap python
bindings, but nslcd proved NSS and PAM LDAP integration.


yocalebo (approved):

yocalebo:
backport

PR #10614, opened 2023-01-31
NAS-120081 / 22.12.1 / fix importing zpools on HA

yocalebo:
The standby controller must be made aware of partition changes or zpool import on failover event could fail.

This adds a new method `disk.retaste` that will open the disk(s) in write mode which will causes kernel to detect any partitions on said device so that the `/dev/disk/by-partuuid` sysfs structure is populated correctly. Without the sysfs structure being populated correctly, zpool import will fail to find any zpools because, on SCALE, zpool import does not scan the disks for any metadata.

I use multiprocessing.Pool so that this is truly parallel execution since this runs on our HA hardware which has plenty of CPU cores. On an internal M60 with 1250 disks, this takes ~1.1 seconds to retaste all the disks.

We retaste disks in 2 scenarios:
1. on zpool creation
2. when we're becoming the active controller

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/retaste.py
+++ b/src/middlewared/middlewared/plugins/disk_/retaste.py
@@ -1,0 +1,48 @@
+import multiprocessing
+
+from pyudev import Context
+
+from middlewared.service import Service, accepts, job
+from middlewared.schema import List, Str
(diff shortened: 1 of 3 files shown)


yocalebo:
retest this please

yocalebo:
backport

anodos325 (approved):

PR #10620, opened 2023-01-31
NAS-120162 / 23.10 / Simplified apps dataset preset

anodos325:


src/middlewared/middlewared/plugins/pool_/dataset.py
```diff
@@ -657,6 +662,24 @@ async def do_create(self, app, data):
...
             await acl_job.wait()
 
+        elif data['type'] == 'FILESYSTEM' and data['share_type'] == 'APPS':
+            acl = (await self.middleware.call('filesystem.acltemplate.by_path', {
+                'query-filters': [('name', '=', 'NFS4_RESTRICTED')],
+                'format-options': {'canonicalize': True, 'ensure_builtins': True},
+            }))[0]['acl']
+            acl.append({
+                'tag': 'USER',
+                'id': 568,
```
yocalebo, line 672: Just a nit, but maybe define this UID somewhere (maybe we already do and I'm just forgetting)

yocalebo (approved):
Minor nit

anodos325:
Purpose of this is to provide a sensible preset for data that will be shared between apps and system services.

src/middlewared/middlewared/plugins/pool_/dataset.py
```diff
@@ -657,6 +662,24 @@ async def do_create(self, app, data):
...
+                'perms': {'BASIC': 'MODIFY'},
+                'flags': {'BASIC': 'INHERIT'},
+                'type': 'ALLOW'
+            })
+            acl_job = await self.middleware.call('filesystem.setacl', {
+                'path': mountpoint,
+                'dacl': acl,
+            })
+            await acl_job.wait(raise_error=True)
```
sonicaj, line 681: Should we raise error at L663 too ?

sonicaj (approved):

src/middlewared/middlewared/plugins/pool_/dataset.py
anodos325: I refactored slightly so that we only have one acl_job now. We also now perform execute check on mountpoint and raise ValidationError if the ACL contains entries that won't be effective.

anodos325:
Since this was refactored since my last full CI run, will send it through the wringer again.

PR #10631, opened 2023-02-01
NAS-120193 / 23.10 / Fix cleanup of DNS reverse zone

anodos325:
* Fix trailing dot on ptr record string in DNS test
* Handle SERVFAIL while looking up netbiosname
* Fix formatting of messages to delete ptr records when
  deleting multiple incorrect ones.


ericbsd (approved):
Test LGTM

yocalebo (approved):

PR #10632, opened 2023-02-02
NAS-120186 / 23.10 / Wait for app resources to be deleted by helm

Qubad786:
## Context

Earlier helm did not provide a way to wait for pods deployed by a chart release to be stopped on uninstallation. That feature has been added to helm now and it can now wait for pods to be terminated which were deployed by the chart release.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
@@ -583,9 +583,9 @@ class ChartReleaseService(CRUDService):
         chart_release = await self.get_instance(release_name, {'extra': {'retrieve_resources': True}})
         namespace = get_namespace(release_name)
 
-        cp = await run(['helm', 'uninstall', release_name, '-n', namespace], check=False)
+        cp = await run(['helm', 'uninstall', release_name, '-n', namespace, '--wait'], check=False)
         if cp.returncode:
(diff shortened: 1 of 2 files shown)


sonicaj (approved):

src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
```diff
@@ -616,10 +616,12 @@ async def do_delete(self, job, release_name, options):
                 await self.middleware.call('k8s.pod.delete', pod['metadata']['name'], {'namespace': namespace})
 
         job.set_progress(75, f'Waiting for {release_name!r} pods to terminate')
+        # We still have a fallback for waiting for pods to terminate
+        # because of https://github.com/helm/helm/issues/10586
         await self.middleware.call('chart.release.wait_for_pods_to_terminate', get_namespace(release_name))
-        await self.middleware.call('chart.release.wait_for_namespace_to_terminate', get_namespace(release_name))
 
         await self.post_remove_tasks(release_name, job)
+        await self.middleware.call('chart.release.wait_for_namespace_to_terminate', get_namespace(release_name))
```
Ornias1993, line 624: Suggestion:
If it does not, check if there are any kubernetes objects within said namespace (including Custom Resource Objects)... If not, we can safely force nuke it.

Qubad786: We are already nuking it but the API server registers the request and does not wait for the termination to complete hence the wait

PR #10635, opened 2023-02-02
NAS-120196 / 23.10 / Stop kubernetes/VPN services when moving system dataset

sonicaj:
This PR adds changes to restart k8s/vpn services when system dataset is being moved as otherwise it is potentially possible that the service in question is consuming the logging dataset.

### DIFF
--- a/src/middlewared/middlewared/plugins/sysdataset.py
+++ b/src/middlewared/middlewared/plugins/sysdataset.py
@@ -670,7 +670,7 @@ class SystemDatasetService(ConfigService):
         simultaneous releases of system dataset.
         """
         with self.sysdataset_release_lock:
-            restart = ['collectd', 'rrdcached', 'syslogd']
+            restart = ['collectd', 'rrdcached', 'syslogd', 'kubernetes', 'openvpn_server', 'openvpn_client']
 
             if self.middleware.call_sync('service.started', 'cifs'):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -672,20 +672,23 @@ def release_system_dataset(self):
...
+                {'name': 'wsdd'},
+                {'name': 'kubernetes'},
+                {'name': 'openvpn_client'},
+                {'name': 'openvpn_server'},
+            ]:
+                if self.middleware.call_sync(svc_to_check.get('method', 'service.started'), svc_to_check['name']):
+                    if 'index' in svc_to_check:
+                        restart.insert(svc_to_check['index'], svc_to_check['name'])
+                    else:
+                        restart.append(svc_to_check['name'])
```
anodos325, line 691: You could probably simplify to:
```
restart.insert(svc_to_check.get('index', len(restart)), svc_to_check['name'])
```

yocalebo (changes requested):
Please don't do this on master as I'm working on a PR locally to remove /var/log from system dataset so this won't be needed. You can keep the changes to 22.12.2 and 22.12.1

sonicaj:
backport

PR #10644, opened 2023-02-03
NAS-119536 / 23.10 / keep /var/log on boot partition

yocalebo:
Having the capability to move /var/log caused despair and anguish to bore a hole through the heart of any developer that stumbled across its (code) path. Countless integration tests have failed, thousands of hours lost to troubleshooting, needlessly complex workarounds and, finally, the sanity forcefully taken from any developer that touched this area of our code. There was, at one point, a logical reason to add this functionality but that was years ago, and times have changed. This removes the /var/log from the system dataset. This allows us to dramatically simplify many parts of our code. The downstream benefits of this change are too numerous to put in a github comment so just trust me when I say this has been a long time coming.


themylogin (approved):
Great job!

Ornias1993 (commented):
Bonuspoints for the PR message :)

PR #10648, opened 2023-02-03
Move ZFS plugin into separate service parts

sonicaj:


anodos325:
Do we really need this? it makes git history more painful.

yocalebo (approved):
CI/CD looks clean with these changes

PR #10650, opened 2023-02-06
NAS-119998 / 23.10 / Allow users to define custom portals

sonicaj:
This PR adds ability for users to define custom web portals along with https://github.com/truenas/catalog_validation/pull/33. This works in 2 layers:

1. App developer specifies that app in question can have custom web portals which results in a set of questions being presented to the user so he can add these portals.
2. User has been exposed with an option to define custom portals and if he/she desires can do that appropriately.

There are 2 cases for app developers where they might want to allow using custom portals. They are as respectively:
1. A valid case where app dev wants the user to define what aspect of app to expose ( based on other configuration he could have possible done )
2. App developer not wanting to specify portals in `questions.yaml` and using `ix_valyes.yaml` for example.

For (1) what the app dev should do is set the boolean flag for `enableIXPortals` in questions.yaml and place it at top level. Once that is done, he/she should also specify `iXPortalsGroupName` key which would dictate in which group the portal question belongs to. The groups are already being defined in `groups` key. Setting these 2 keys in  `questions.yaml` will make middleware add another set of question dynamically which will be rendered in the UI and validated appropriately where user can specify portals he/she desires to expose.

For (2), there is no need to specify `enableIXPortals` flag in questions.yaml and app dev can just specify portals in the following manner:
```
iXPortals:
    - portalName: "some portal"
      protocol: "http"
      useNodeIP: false
      host: "192.168.0.123"
      port: 15000
      path: "/login"
```

Now changes introduced by these 2 PRs will basically allow app devs to specify if they want user to able custom user portals where these custom user portals schema would be automatically added by middleware and validated accordingly so that invalid values are not specified.

Co-authored-by: Kjeld Schouten-Lebbing <<kjeld@schouten-lebbing.nl>>

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
@@ -89,9 +89,12 @@ class ChartReleaseService(Service):
         for portal_config in release_data['config'].get(CUSTOM_PORTALS_KEY) or []:
             path = portal_config.get('path') or ''
             host = node_ip if portal_config['useNodeIP'] else portal_config['host']
-            portals[portal_config['portalName']] = [
-                f'{portal_config["protocol"]}://{host}:{portal_config["port"]}{path}'
-            ]
+            protocol = portal_config['protocol']
+            port = portal_config['port']
+            if (protocol == 'http' and port == 80) or (protocol == 'https' or port == 443):
+                port = ''
+
+            portals[portal_config['portalName']] = [f'{protocol}://{host}:{port}{path}']
         return portals
 
     @private
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
@@ -4,7 +4,7 @@ import threading
 import yaml
 
 from catalog_validation.items.questions_utils import CUSTOM_PORTALS_KEY
-from catalog_validation.items.ix_values_utils import IX_VALUES_JSON_SCHEMA
+from catalog_validation.items.ix_values_utils import CUSTOM_PORTALS_JSON_SCHEMA
 from jsonschema import validate as json_schema_validate, ValidationError as JsonValidationError
 
 from middlewared.service import private, Service
@@ -84,7 +84,7 @@ class ChartReleaseService(Service):
         if custom_portals is None:
             return portals
         try:
-            json_schema_validate({CUSTOM_PORTALS_KEY: custom_portals}, IX_VALUES_JSON_SCHEMA)
+            json_schema_validate(custom_portals, CUSTOM_PORTALS_JSON_SCHEMA)
         except JsonValidationError:
             return portals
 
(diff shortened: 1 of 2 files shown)


Ornias1993:
@sonicaj Some tips on correct use of co-authored-by tags, this is the format that needs to be on the final commit...
It's best to set them on the PR commits, as they should be included even when using github "squash commits"

This is the format:
`Co-authored-by: Kjeld Schouten-Lebbing <kjeld@schouten-lebbing.nl>`

The most important part being the email adress.
If you don't have that, this is an awesome and easy tool to generate co-author-tags using github usernames:
https://coauthoredby.netlify.app/

src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
```diff
@@ -71,8 +70,20 @@ def tag_func(key):
...
         return cleaned_portals
 
+    @private
+    def get_user_configured_portals(self, release_data, node_ip):
+        portals = {}
+        for portal_config in release_data['config'].get(CUSTOM_PORTALS_KEY) or []:
+            path = portal_config.get('path') or ''
+            host = node_ip if portal_config['useNodeIP'] else portal_config['host']
+            portals[portal_config['portalName']] = [
+                f'{portal_config["protocol"]}://{host}:{portal_config["protocol"]}{path}'
```
Ornias1993, line 83:
```suggestion
                f'{portal_config["protocol"]}://{host}:{portal_config["port"]}{path}'
```
I think this needs to be port?

Ornias1993:
It might be good remove port 80 and 443 for http and https protocol repectively, as that's basically the standard anno 2023.

src/middlewared/middlewared/plugins/chart_releases_linux/portals.py
```diff
@@ -42,45 +48,68 @@ def retrieve_portals_for_chart_release(self, release_data, node_ip):
...
+        if custom_portals is None:
+            return portals
+        try:
+            json_schema_validate({CUSTOM_PORTALS_KEY: custom_portals}, IX_VALUES_JSON_SCHEMA)
+        except JsonValidationError:
+            return portals
+
+        for portal_config in release_data['config'].get(CUSTOM_PORTALS_KEY) or []:
+            path = portal_config.get('path') or ''
+            host = node_ip if portal_config['useNodeIP'] else portal_config['host']
```
Ornias1993, line 93: if host is not given, you can just default to the NodeIP... there is not really a reason for the useNodeIP option I think?

sonicaj: We should be explicit where we can be and this helps a lot especially when user is filling the portal form. It would be very weird to say "Leave host empty and then Node IP will be used" as compared to having a Use Node IP checkbox which when unchecked shows the host field.

sonicaj: We do that but that has it's drawbacks, node ip is something which can change when user changes it and then we can't change that value in each app's configuration. So it's simpler to just let them specify if they want whatever is the node ip atm to be used :)

yocalebo (changes requested):
unit tests are broken with these changes?

sonicaj:
That is because catalog validation PR was required and how our unit tests run does not take into account changes in other repos having same branch names

yocalebo:
Cool, just making sure.

Ornias1993 (commented):
Good clean, LGTM.

yocalebo (approved):

sonicaj:
backport

PR #10659, opened 2023-02-07
NAS-120210 / 23.10 / fix enclosure mapping on MINI-R

yocalebo:
Slots 9-12 need to be reversed to match physical slot map.


yocalebo:
I found discrepancies with official documentation for this platform and the platform team has resolved this problem by fixing the cabling in-house. No code change needed.

PR #10664, opened 2023-02-08
NAS-120259 / 22.12.2 / `gpt_header_size` might vary between sector sizes; do not take it int…

themylogin:
…o account when validating disks


yocalebo (approved):

themylogin:
time 4:00

PR #10669, opened 2023-02-08
NAS-120261 / 23.10 / Fix tests

yocalebo:
Changes introduced here https://github.com/truenas/middleware/pull/10655 exposed the fact that our tests were written with the assumption that sqlite didn't autoincrement columns on insert. Well...turns out it wasn't actually doing that which was breaking cloud-sync tasks so we fixed the autoincrement problem but then it broke these tests.

Fixes employed here:
1. update the global `tp['id']` after we import the zpool since that translates to a database insert which increments the column `id` by design

### DIFF
--- a/tests/api2/test_008_pool.py
+++ b/tests/api2/test_008_pool.py
@@ -183,6 +183,15 @@ def test_08_test_pool_property_normalization(request):
         assert ds['properties']['mountpoint']['value'] != 'legacy', str(ds['properties'])
         assert ds['properties']['sharenfs']['value'] == 'off', str(ds['properties'])
 
+    # the very next test that runs after this one exports the pool that
+    # we re-imported but the database id has changed so we need to update
+    # the global `tp['id']` variable
+    res = make_ws_request(ip, {'msg': 'method', 'method': 'datastore.query', 'params': ['storage.volume']})
+    assert res.get('error') is None
+    assert res['result']
+    for i in filter(lambda x: x['vol_name'] == tp['name'], res['result']):
+        tp['id'] = i['id']
+
 
 def test_09_export_test_pool_with_destroy_true(request):
     depends(request, ["pool_04"])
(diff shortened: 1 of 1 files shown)


themylogin:
> I've, instead, generated a random zpool name each time another_pool() is called to prevent this cascading effect of failures from taking place.

Unfortunately it won't help. If we are unable to destroy the `test` pool then we won't be able to create a new `another_pool` because we have a limited amount of disks and they all will be busy, so pool name won't really matter here. Tests, on the other hand, get more complicated with the dynamic pool name.

src/middlewared/middlewared/test/integration/assets/pool.py
```diff
@@ -28,19 +30,29 @@
...
+def generate_random_zpool_name():
+    zpool_name_length = 12
+    letters_and_digits = string.ascii_letters + string.digits
+    rando = ''.join(random.choices(letters_and_digits, k=zpool_name_length))
+    return f'cool_zpool_name_{rando}'
+
+
 @contextlib.contextmanager
-def another_pool(data=None, topology=None):
+def another_pool(name=None, data=None, topology=None):
```
themylogin, line 41: It'll have more sense to omit this parameter and just to `data.setdefault("name", generate_random_zpool_name())` as `data` is indeed the dictionary used for `pool.create`. But I also don't think random name will be useful at all, please see my comment above.

themylogin (changes requested):

yocalebo:
> > I've, instead, generated a random zpool name each time another_pool() is called to prevent this cascading effect of failures from taking place.
> 
> Unfortunately it won't help. If we are unable to destroy the `test` pool then we won't be able to create a new `another_pool` because we have a limited amount of disks and they all will be busy, so pool name won't really matter here. Tests, on the other hand, get more complicated with the dynamic pool name.

Sure, the non-HA VM has like 10 or 11 unused disks so this wouldn't be a problem. Our HA VMs, however, are a different story. I've reverted to just fixing the original problem for now.

themylogin (approved):

PR #10678, opened 2023-02-09
NAS-120226 / 23.10 / fix zpool import on failover events

yocalebo:
Past few days I've investigated random occurrences of zpools being unhealthy after failover event. I've learned
1. using a singular zpool cachefile for all zpools isn't the best idea more than likely
2. managing the zpool cachefile is very fragile and will lead to despair
3. limit the search paths given to zpool import on our HA certified hardware line to lessen the possibility of importing a zpool and have random disks get imported using the raw device instead of the gpt device.

Tested extensively in house with QE team since they were the ones seeing this quite often. After these changes, they've had a singular occurrence of this after my changes. I will push a separate PR to, hopefully, fix the last remaining issue.

### DIFF
--- a/src/middlewared/middlewared/plugins/failover_/event.py
+++ b/src/middlewared/middlewared/plugins/failover_/event.py
@@ -393,8 +393,8 @@ class FailoverEventsService(Service):
             # error and move on
             logger.error('Failed to unlock SED disk(s) with error: %r', e)
 
-        # setup the zpool cachefile
-        self.run_call('failover.zpool.cachefile.setup', 'MASTER')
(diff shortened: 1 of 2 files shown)


bmeagherix (approved):
LGTM.

FWIW, from [OpenZFS FAQ](https://openzfs.github.io/openzfs-docs/Project%20and%20Community/FAQ.html#generating-a-new-etc-zfs-zpool-cache-file) wrt cachefile:

> Conversely the cache file can be disabled by setting cachefile=none. This is useful for failover configurations where the pool should always be explicitly imported by the failover software.

PR #10679, opened 2023-02-09
NAS-120280 / 23.10 / Add some validation for query-filters

anodos325:
`query-filters` are always a list of lists (or in some cases tuples) of the form `[<key>, <op>, <val>]` or `['OR', <list of filters>]`. We should perform some basic schema validation to give users more helpful messages.

Example of unhelpful message:
```
root@truenas[/mnt/dozer/middleware/src/middlewared]# midclt call user.query '["bob"]'
Invalid operation: o
```

Example of more helpful message:
```
root@truenas[~]# midclt call user.query '["bob"]'
[EINVAL] query-filters: Item#0 is not valid per list types: [query-filter] Not a list
```

### DIFF
--- a/src/middlewared/middlewared/schema.py
+++ b/src/middlewared/middlewared/schema.py
@@ -585,7 +585,7 @@ class List(EnumMixin, Attribute):
         value = super(List, self).clean(value)
         if value is None:
             return copy.deepcopy(self.default)
-        if not isinstance(value, list):
+        if not isinstance(value, (list, tuple)):
             raise Error(self.name, 'Not a list')
         if not self.empty and not value:
             raise Error(self.name, 'Empty value not allowed')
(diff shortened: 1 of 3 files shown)


anodos325:
Feel free to make recommendations if you think we should have more validation. In theory I could make use recursively validate the whole thing, but I was trying to balance speed / complexity, and at least giving some clue about what is wrong.

yocalebo (approved):
Definitely like the added validation but agree that we should keep this minimal since it directly impacts performance in a rather profound way. Let's get at least 1 other person to approve before merge.

src/middlewared/middlewared/schema.py
```diff
@@ -585,10 +587,17 @@ def clean(self, value):
         value = super(List, self).clean(value)
         if value is None:
             return copy.deepcopy(self.default)
-        if not isinstance(value, list):
+        if not isinstance(value, (list, tuple)) and not issubclass(type(value), list):
```
themylogin, line 590: What is the `value` for which `isinstance(value, (list, tuple))` is `False`, but `issubclass(type(value), list)` is `True`?

anodos325: query-filter type. Come to think of it, might be something wonky going on with our schema stuff...

themylogin: Sorry, what's query-filter type? What python type is it?

anodos325: Hmm... can't seem to reproduce the issue I was seeing that caused this. As I was looking at the general problem again, I realized the PR could be simplified and improved by just re-using validation used in filter_list.

themylogin (approved):
LGTM, thank you!

PR #10684, opened 2023-02-10
NAS-120291 / 23.10 / wait for disk events to settle on failover event

yocalebo:
Again, having worked with QE for many days on this problem, I found that by solving the problem on the active controller of not seeing the gpt labels (by "retasting" all the disks) we introduced another subtle issue. When disks are opened in write mode (to tell kernel to update sysfs structure so ZFS can import pools properly), 3 udev events are generated. REMOVE, CHANGE, ADD (in that order) which means we introduced a subtle race condition where we could retaste the disks and import the zpool BEFORE all the ADD events had been generated/processed by udev. This, ultimately, results in zpools being imported with certain disks being reported as "OFFLINE". (When in reality, they're there we just tried the zpool import too soon)


anodos325 (approved):

bmeagherix:
I have a question.  In the other place where we use `Monitor.from_netlink` (`udev_events`) you recently added `monitor.set_receive_buffer_size(_256MB)`

Is there a reason this is not applicable here too?

PR #10690, opened 2023-02-12
NAS-119492 / 23.10 / `ssh_password_enabled`

themylogin:


themylogin:
@anodos325 [api2.test_012_directory_service_ssh.test_08_test_ssh_ad](https://ci.tn.ixsystems.net/jenkins/job/TrueNAS%20SCALE%20-%20Unstable/job/API%20Tests%20-%20TrueNAS%20SCALE%20with%20middleware%20reinstalled%20from%20a%20branch%20Incremental/607/testReport/junit/api2/test_012_directory_service_ssh/test_08_test_ssh_ad/) and [api2.test_012_directory_service_ssh.test_09_test_ssh_ldap](https://ci.tn.ixsystems.net/jenkins/job/TrueNAS%20SCALE%20-%20Unstable/job/API%20Tests%20-%20TrueNAS%20SCALE%20with%20middleware%20reinstalled%20from%20a%20branch%20Incremental/607/testReport/junit/api2/test_012_directory_service_ssh/test_09_test_ssh_ldap/) fail. What do we do with AD/LDAP users being apple to password-auth on SSH now?

themylogin:
`api2.test_user_admin.*` fail because this pipeline does not update `truenas-set-authentication-method.py` script

anodos325:
What if we just have a list of groups (names) that are granted password access via SSH in services->SSH rather than modifying local users/groups?

Validate list on `ssh.update()` and during etc file generation.

themylogin:
@yocalebo thoughts on the above?

yocalebo:
> @yocalebo thoughts on the above?

I agree with @anodos325. I learned some implementation details with the pam module and directory service users that's kind of gross. Using groups is the path of least resistance and seems to be the better solution.

themylogin:
@yocalebo what do I do then? Remove boolean flag from the group, add "group names" SSH option? What should I do with the users — should I keep the flag there? If not, it's not clear how to migrate current `rootlogin`/`adminlogin` flags. Additionally, we might need to show a UI warning saying "this user is a member for the group XYZ which has SSH password login enabled"

yocalebo:
1. We should keep the user table changes since we don't store directory service users in our local db
2. Remove the changes to the tables related to groups
3. We should change it so that the ssh service can accept a list of groups (either local or from a directory service) that gets put into sshd_config.
4. I'm not sure an alert about a user being able to ssh into the system is valid? The user will have had to manually add this information so they'll probably expect the users to be able to login and the alert will just be frustrating. We can discuss this one a bit more.

themylogin:
@anodos325 I made the changes, but SSH AD/LDAP tests still fail https://ci.tn.ixsystems.net/jenkins/job/TrueNAS%20SCALE%20-%20Unstable/job/API%20Tests%20-%20TrueNAS%20SCALE%20with%20middleware%20reinstalled%20from%20a%20branch%20Incremental/636/#showFailuresLink

Which group name(s) should I put to SSH configuration to allow them SSH password login.

Also looks like this will break SSH password login for existing installations that use AD/LDP @yocalebo

yocalebo:
> Which group name(s) should I put to SSH configuration to allow them SSH password login.
Whichever groups are on the AD server. You can choose a well-known group like "Domain Users" since all newly created user objects get put into that group by default.

> Also looks like this will break SSH password login for existing installations that use AD/LDP @yocalebo
An unfortunate side-effect of having a more secure configuration out of the box. It'll be something we have to note in the release notes. No way around it, unfortunately.

anodos325:
> Which group name(s) should I put to SSH configuration to allow them SSH password login.
For AD and LDAP, I'd just do a call to `user.get_user_obj` for the AD username we use to authenticate via SSH. This will get you the gid of the primary group, then call `group.get_group_obj` to get the group name of the group.

> Also looks like this will break SSH password login for existing installations that use AD/LDP @yocalebo

Yup. Breaking change. Will need to be documented.

themylogin:
Tests pass @yocalebo @anodos325

themylogin:
@anodos325 https://ci.tn.ixsystems.com/jenkins/job/TrueNAS%20SCALE%20-%20Unstable/job/API%20Tests%20-%20TrueNAS%20SCALE%20with%20middleware%20reinstalled%20from%20a%20branch%20Incremental/664/#showFailuresLink

yocalebo (changes requested):
Could you please rebase this? We've had another migration get merged since you opened this up.

yocalebo (approved):
Thanks! Please wait for @anodos325 final review before merge.

anodos325 (approved):

PR #10691, opened 2023-02-12
Added OVH authenticator to list of ACME DNS providers

Dogeek:
I know this is a bit unprompted, but I couldn't get the shell script route to work to add my domain registered at OVH to work with SSL certificates.

The code is more or less identical to the Cloudflare Authenticator, so it should work out of the box. One caveat is that OVH can only guarantee propagation to all DNS within 24 hours, in practice though I haven't had any case happen where propagation wasn't done within the hour.

Any comment / review is greatly appreciated for my first contribution to TrueNAS

### DIFF
--- a/src/middlewared/middlewared/plugins/acme_protocol_/authenticators/factory.py
+++ b/src/middlewared/middlewared/plugins/acme_protocol_/authenticators/factory.py
@@ -5,6 +5,7 @@ from middlewared.service_exception import CallError
 from .cloudflare import CloudFlareAuthenticator
 from .route53 import Route53Authenticator
 from .shell import ShellAuthenticator
+from .ovh import OVHAuthenticator
 
 
 class AuthenticatorFactory:
@@ -28,6 +29,7 @@ auth_factory = AuthenticatorFactory()
 for authenticator in [
     CloudFlareAuthenticator,
     Route53Authenticator,
+    OVHAuthenticator,
     ShellAuthenticator,
 ]:
     auth_factory.register(authenticator)
--- a/src/middlewared/middlewared/plugins/acme_protocol_/authenticators/factory.py
+++ b/src/middlewared/middlewared/plugins/acme_protocol_/authenticators/factory.py
@@ -3,9 +3,9 @@ import errno
 from middlewared.service_exception import CallError
 
 from .cloudflare import CloudFlareAuthenticator
+from .ovh import OVHAuthenticator
 from .route53 import Route53Authenticator
 from .shell import ShellAuthenticator
-from .ovh import OVHAuthenticator
 
 
 class AuthenticatorFactory:
--- a/src/middlewared/middlewared/plugins/acme_protocol_/authenticators/ovh.py
+++ b/src/middlewared/middlewared/plugins/acme_protocol_/authenticators/ovh.py
@@ -1,0 +1,51 @@
+import logging
+
+from lexicon.providers.ovh import ENDPOINTS
+from certbot_dns_ovh._internal.dns_ovh import _OVHLexiconClient
+
+from middlewared.schema import accepts, Dict, Str, ValidationErrors
+from middlewared.service import skip_arg
+
+from .base import Authenticator
+
+
+logger = logging.getLogger(__name__)
+OVH_ENDPOINTS = tuple(ENDPOINTS.keys())
+
+class OVHAuthenticator(Authenticator):
+
+    NAME = 'OVH'
+    PROPAGATION_DELAY = 60
+    SCHEMA = Dict(
+        'OVH',
+        Str('application_key', empty=False, null=False, title='OVH Application Key'),
+        Str('application_secret', empty=False, null=False, title='OVH Application Secret'),
+        Str('consumer_key', empty=False, null=False, title='OVH Consumer Key'),
+        Str('endpoint', empty=False, default='ovh-eu', title='OVH Endpoint', enum=OVH_ENDPOINTS),
+    )
+
+    def initialize_credentials(self):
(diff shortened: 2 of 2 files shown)


Ornias1993:
I don't think "within the hour" is even fast enough to have it work within TrueNAS.
It should be "within a few minutes" at worst.

This might also be why industry leader Cert-Manager has not natively added it yet as well.

Dogeek:
I eventually managed to make the shell script work for OVH, and there were no issues (DNS propagation was pretty much instantaneous in my case, though I live in France, where OVH is based, so there might be a bias there)

I used the same library as here in this pull request, and pretty much the same code, here it is :

```python
#!/usr/bin/env python3

from lexicon.providers.ovh import ENDPOINTS
from certbot_dns_ovh._internal.dns_ovh import _OVHLexiconClient
import sys

application_key = "xxxxxxx"
application_secret = "xxxxxxxx"
consumer_key = "xxxxxx"
endpoint = "ovh-eu"

def _perform(domain, validation_name, validation_content):
    get_client().add_txt_record(domain, validation_name, validation_content)

def get_client():
    return _OVHLexiconClient(
        endpoint, application_key, application_secret,
        consumer_key, 600,
    )

def _cleanup(domain, validation_name, validation_content):
    get_client().del_txt_record(domain, validation_name, validation_content)

if sys.argv[1] == 'set':
    _perform(*sys.argv[2:])
elif sys.argv[1] == 'unset':
    _cleanup(*sys.argv[2:])

```

src/middlewared/middlewared/plugins/acme_protocol_/authenticators/ovh.py
```diff
@@ -0,0 +1,51 @@
...
+        self.application_key = self.attributes.get('application_key')
+        self.application_secret = self.attributes.get('application_secret')
+        self.consumer_key = self.attributes.get('consumer_key')
+        self.endpoint = self.attributes.get('endpoint')
+
+    @staticmethod
+    @accepts(SCHEMA)
+    @skip_arg(count=1)
+    async def validate_credentials(middleware, data):
+        verrors = ValidationErrors()
```
sonicaj, line 37: You can just return data here or maybe even better would be to modify the base class and return data there by default, so route53 authenticator and ovh authenticator can benefit from those changes.

This is assuming you don't have any validation to do here and all above fields are required.

src/middlewared/middlewared/plugins/acme_protocol_/authenticators/ovh.py
```diff
@@ -0,0 +1,51 @@
...
+logger = logging.getLogger(__name__)
+OVH_ENDPOINTS = tuple(ENDPOINTS.keys())
+
+class OVHAuthenticator(Authenticator):
+
+    NAME = 'OVH'
+    PROPAGATION_DELAY = 60
+    SCHEMA = Dict(
+        'OVH',
+        Str('application_key', empty=False, null=False, title='OVH Application Key'),
```
sonicaj, line 21: Are all these 4 attributes required ? If yes, please set `required=True` appropriately then

src/middlewared/middlewared/plugins/acme_protocol_/authenticators/ovh.py
```diff
@@ -0,0 +1,51 @@
...
+        verrors = ValidationErrors()
+        verrors.check()
+        return data
+
+    def _perform(self, domain, validation_name, validation_content):
+        self.get_client().add_txt_record(domain, validation_name, validation_content)
+
+    def get_client(self):
+        return _OVHLexiconClient(
+            self.endpoint, self.application_key, self.application_secret,
```
sonicaj, line 46: Looking at this, it seems all these attributes are required - please make the appropriate changes in the schema

src/middlewared/middlewared/plugins/acme_protocol_/authenticators/factory.py
```diff
@@ -5,6 +5,7 @@
 from .cloudflare import CloudFlareAuthenticator
 from .route53 import Route53Authenticator
 from .shell import ShellAuthenticator
+from .ovh import OVHAuthenticator
```
sonicaj, line 8: While we are making some other changes, can we please have this alphabetically ordered :D 

So it would come after `cloudflare`, similar changes would be required below

sonicaj (commented):
Thank you @Dogeek for your contribution. I have some minor concerns listed below

src/middlewared/middlewared/plugins/acme_protocol_/authenticators/ovh.py
Dogeek: Sure thing, I did it, it's just that since there wasn't much documentation about what a schema actually is (for truenas) I based myself on the cloudflare authenticator to integrate the OVH authenticator as closely as possible.

src/middlewared/middlewared/plugins/acme_protocol_/authenticators/ovh.py
Dogeek: I indeed do not have any validation to do here, after all, this method, at least on the cloudflare authenticator, is used to make sure that we either have a global API key, or a n API token to use, no API calls are made.

I could add some validation by trying to perform an API call to OVH, and see the HTTP response code, but it strays away from the behaviour of other authenticators.

Side note : I've noticed that certbot includes clients for other well-known registrars, maybe I can implement authenticators for them as well (like DigitalOcean, Linode, Google and others, that may help other people registered with these companies to set up their domain names)

sonicaj (approved):
Thank you @Dogeek for the changes!

PR #10693, opened 2023-02-13
feat(rclone): Add checks parameter for rclone

SISheogorath:
This patch integrates the parameter `--checkers=N` into the cloud_sync plugin in order to allow speeding up sync of small files.

References:
https://rclone.org/docs//#checkers-n


yocalebo (changes requested):
Hello! Thanks for the patch. An alembic migration has to be written before adding a new column to the table in the database.

yocalebo:
Please open a feature request [here](https://ixsystems.atlassian.net/) and we can consider adding this feature to the product.

PR #10711, opened 2023-02-15
23.10 / remove license requirement on k8s passthrough mode

Ornias1993:
While I 100% agree that "passthrough mode" should not exposed to normal end users directly in the GUI. Locking this behind enterprise hardware requirements, has no technical reasoning at all behind it. At the same time, it's already trivial to remove this needless requirement manually for people with the knowhow to even make the API call.

iX has always been clear that it only puts OS features and API calls behind a license if it's technically required.

**TLDR**
This limit isn't technically needed, it makes no sense for gatekeeping novice users AND it seems to be in contrast/violation of expected iX policies regarding licensing.

This is not likely to be used much at all, because the knowhow required to even correctly deploy this hidden gem, is huge.
Basically: lets keep it security through obscurity please.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
@@ -142,9 +142,6 @@ class KubernetesService(ConfigService):
                 'System is not licensed to use Applications'
             )
 
-        if data['passthrough_mode'] and not await self.middleware.call('system.license'):
-            verrors.add(f'{schema}.passthrough_mode', 'Can only be enabled on licensed iX enterprise hardware')
-
         if data['pool'] and not await self.middleware.call('pool.query', [['name', '=', data['pool']]]):
             verrors.add(f'{schema}.pool', 'Please provide a valid pool configured in the system.')
 


kmoore134:
We are not planning to remove this check. This is something we're only expecting to use for Enterprise Customers we are directly engaged with on their deployments.

Ornias1993:
> We are not planning to remove this check. This is something we're only expecting to use for Enterprise Customers we are directly engaged with on their deployments.

So this means iX is officially droping the old policy of "we only license restrict where technical issues arise"?

kmoore134:
I would argue that it fits this policy. It opens doors to a whole host of technical issues that we are not planning to support on the community side of the project, it would render the community edition unsupported.

Ornias1993:
> I would argue that it fits this policy. It opens doors to a whole host of technical issues that we are not planning to support on the community side of the project, it would render the community edition unsupported.

And how is that related to hardware?
Like the fiber support and such?

THAT was the policy untill now.
Technical as in: Hardware related.

Ornias1993:
From another side:
hostPath validation: "opens doors to a whole host of technical issues that we are not planning to support"(when disabled)
**Fully supported even in GUI**

This: "opens doors to a whole host of technical issues that we are not planning to support"
**Somehow super important to add license limitations for**

PR #10715, opened 2023-02-18
NAS-120391 / 23.10 / Cleanup and improve SentryCrashReporting class

yocalebo:
1. Rename `CrashReporting` to `SentryCrashReporting` to reflect reality
2. Move this class to its own file
3. We're currently running sentry 0.13.2 which means we can get rid of the monkey-patching to work-around a bug we saw in a version < 0.10.3
4. Stop loading, potentially, a large file into memory every time a crash occurs just so we can load the last `MAX_STRING_LEN` chars of the file. Seek to relative end of file, and then read.

### DIFF
--- a/src/middlewared/middlewared/logging/sentry_crash_reporting.py
+++ b/src/middlewared/middlewared/logging/sentry_crash_reporting.py
@@ -43,7 +43,8 @@ class SentryCrashReporting:
             # disabled via environment variable
             return True
         elif os.stat(__file__).st_dev != os.stat('/').st_dev:
-            # really??
+            # middlewared package directory is remotely mounted
+            # (i.e. done by some developers)
             return True
         else:
             return False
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/logging/sentry_crash_reporting.py
```diff
@@ -0,0 +1,86 @@
...
+            # disabed in db
+            return True
+        elif os.path.exists(os.path.join(MIDDLEWARE_RUN_DIR if sw_info()['stable'] else '/data', DISABLED_SENTINEL)):
+            # disabled because of presence of sentinel file
+            return True
+        elif os.environ.get(DISABLED_KEY, False):
+            # disabled via environment variable
+            return True
+        elif os.stat(__file__).st_dev != os.stat('/').st_dev:
+            # really??
```
themylogin, line 46: Middleware mounted from an external source http://docs.middleware.ixsystems.net/middleware/development.html#substituting-the-middleware-code-in-an-already-built-image

yocalebo: Heh, yeah I realized that and forgot to remove comment 😄

themylogin (approved):

themylogin (approved):

PR #10717, opened 2023-02-19
NAS-112859 / 23.10 / Refactor `system` event into `system.ready`, `system.reboot`, `system.shutdown`

themylogin:


src/middlewared/middlewared/plugins/device_/vrrp_events.py
```diff
@@ -56,29 +56,38 @@ def run(self):
...
+    await _start_stop_vrrp_thread(middleware, shutting_down=True)
+
+
+async def _post_license_update(middleware, *args, **kwargs):
+    await _start_stop_vrrp_thread(middleware)
 
 
 async def setup(middleware):
-    middleware.create_task(_event_system(middleware))  # start thread on middlewared service start/restart
-    middleware.register_hook('system', _event_system)  # catch shutdown event and clean up thread
```
themylogin, line 83: @yocalebo this never worked, `event_subscribe` should have been used instead. Thread was managed only on middleware startup or license update. If it was sufficient, maybe we don't need to manage thread on system boot / shutdown after all?

themylogin:
time 3:00

src/middlewared/middlewared/plugins/device_/vrrp_events.py
yocalebo: Probably not. The only use-cases that we need are:
1. on boot
2. on middlewared service restart
3. on license update from single to HA

yocalebo (approved):

PR #10723, opened 2023-02-21
NAS-120417 / 23.10 / Add ability to specify whether to create homedir

anodos325:
This switches implicit creation of non-existing path to having API user explicitly determine whether to create a homedir with the username.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -144,6 +144,7 @@ class UserService(CRUDService):
         ('rm', {'name': 'group'}),
         ('rm', {'name': 'group_create'}),
         ('rm', {'name': 'home_mode'}),
+        ('rm', {'name': 'home_create'}),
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/account.py
```diff
@@ -282,6 +284,47 @@ async def query(self, filters, options):
...
+        try:
+            setperm_job = self.middleware.call_sync('filesystem.setperm', {
+                'path': target,
+                'mode': mode,
+                'uid': uid,
+                'gid': gid,
+                'options': {'stripacl': True}
+            })
+            setperm_job.wait_sync()
+        except Exception:
```
yocalebo, line 321: This won't ever raise an exception since it's a job. You need to do `wait_sync(raise_error=True)`

yocalebo (changes requested):

yocalebo (approved):

PR #10737, opened 2023-02-23
NAS-120478 / 23.10 / Make sure domain object is active before consumption

Qubad786:
## Context

For each VM specified in database we keep a list of VM objects throughout middleware lifecycle which is what interacts with libvirt. In each VM object we have a domain instance which is returned by the libvirt socket and consumed to perform actions for that specific VM.

## Problem

If libvirt dies for whatever reason, we have logic in place which re-starts libvirt and re-initializes the libvirt connection but the domain object still points to the older libvirt socket which is no longer active.

## Solution

Before consuming domain object make sure it is healthy and re-initialize otherwise as it is guaranteed at this point that libvirt is active and the socket connection is healthy.


sonicaj (approved):
Good catch!

PR #10740, opened 2023-02-24
NAS-120458 / 23.10 / NAS-120458 Avoid parsing dates while reading yaml

zenware:
This is specifically for YAML which is then converted to JSON for delivery to a Kubernetes API endpoint. The YAML parser converts datetime strings into Python datetime objects by default, losing the original formatting, and the JSON library doesn't understand what format it should use to output dates.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/yaml.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/yaml.py
@@ -1,5 +1,27 @@
 import yaml
 
+class NoDatesFullLoader(yaml.FullLoader):
+    @classmethod
+    def remove_implicit_resolver(cls, tag_to_remove):
+        """
+        Remove implicit resolvers for a particular tag
+
+        Takes care not to modify resolvers in super classes.
+
+        We want to load datetimes as strings, not dates, because we
+        go on to serialise as json which doesn't have the advanced types
+        of yaml, and leads to incompatibilities down the track.
+        """
+        if not 'yaml_implicit_resolvers' in cls.__dict__:
+            cls.yaml_implicit_resolvers = cls.yaml_implicit_resolvers.copy()
+
+        for first_letter, mappings in cls.yaml_implicit_resolvers.items():
+            cls.yaml_implicit_resolvers[first_letter] = [(tag, regexp) 
+                                                         for tag, regexp in mappings
+                                                         if tag != tag_to_remove]
+
+NoDatesFullLoader.remove_implicit_resolver('tag:yaml.org,2002:timestamp')
+
 
 class SafeDumper(yaml.SafeDumper):
     pass
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/restore.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/restore.py
@@ -9,6 +9,7 @@ import yaml
 
 from middlewared.schema import Dict, Bool, returns, Str
 from middlewared.service import accepts, CallError, job, Service
+from middlewared.plugins.kubernetes_linux.yaml import NoDatesFullLoader
 
 
 RE_POOL = re.compile(r'^.*?(/.*)')
@@ -150,14 +151,15 @@ class KubernetesService(Service):
 
             # First we will restore namespace and then the secrets
             with open(os.path.join(r_backup_dir, 'namespace.yaml'), 'r') as f:
-                namespace_body = yaml.load(f.read(), Loader=yaml.FullLoader)
+                namespace_body = yaml.load(f.read(), Loader=NoDatesFullLoader)
                 namespace_body['metadata'].pop('resourceVersion', None)
+
                 self.middleware.call_sync('k8s.namespace.create', {'body': namespace_body})
 
             secrets_dir = os.path.join(r_backup_dir, 'secrets')
             for secret in sorted(os.listdir(secrets_dir)):
                 with open(os.path.join(secrets_dir, secret)) as f:
-                    secret_body = yaml.load(f.read(), Loader=yaml.FullLoader)
+                    secret_body = yaml.load(f.read(), Loader=NoDatesFullLoader)
                     secret_body['metadata'].pop('resourceVersion', None)
                     self.middleware.call_sync(
                         'k8s.secret.create', {


zenware:
I couldn't find documentation on how this project prefers to receive PRs, I did review some other open PRs as well as some git history in an attempt to conform to project standards, please advise if I have missed something.

There are likely more places in the `../kubernetes_linux/*.py` files that need this or a similar change, that I have not covered as I was specifically catering to my needs for calling the `midcli` capability `app kubernetes restore_backup`. Please feel free to modify these changes to encompass those locations or if you would prefer I do it, I will find some time early next week.

yocalebo:
@zenware you did great! Thanks for the PR, we'll review it ASAP 👍

sonicaj (commented):
@zenware the changes make sense but i have not been able to reproduce the problem where the dates actually were a problem ?
Could you share some sample yaml generated by TrueNAS when it created a backup which was problematic when trying to restore the cluster ? Thanks!

zenware:
Yep, here's an example of an offending file, hope it helps!
```yaml
apiVersion: v1
kind: Namespace
metadata:
  annotations: null
  cluster_name: null
  creation_timestamp: 2022-03-10 02:07:27+00:00
  deletion_grace_period_seconds: null
  deletion_timestamp: null
  finalizers: null
  generate_name: null
  generation: null
  labels:
    catalog: TRUECHARTS
    catalog_branch: main
    catalog_train: core
    kubernetes.io/metadata.name: ix-traefik
  managed_fields:
  - api_version: v1
    fields_type: FieldsV1
    fields_v1:
      f:metadata:
        f:labels:
          .: {}
          f:catalog: {}
          f:catalog_branch: {}
          f:catalog_train: {}
          f:kubernetes.io/metadata.name: {}
    manager: OpenAPI-Generator
    operation: Update
    subresource: null
    time: 2022-03-10 02:07:27+00:00
  name: ix-traefik
  namespace: null
  owner_references: null
  resource_version: '18932'
  self_link: null
  uid: 97c281a3-4cb7-4065-a5c1-31221c73fb67
spec:
  finalizers:
  - kubernetes
status:
  conditions: null
  phase: Active
```

It's `metadata.creation_timestamp` as well as `metadata.managed_fields.time` which are the offenders. And every yaml file in backups I've created has timestamps like this. Perhaps a fix that makes more sense is to ensure that they're encoded initially as strings when being backed up? Although I'm suspicious the YAML parser would still serialize them into date objects.

I think the most robust solution is likely having a Pydantic model of Kubernetes types, like "Namespace" which guarantees all the fields are appropriately formatted before sending a request to the k8s APIs, but that also is like... a very heavyweight solution, or at least it feels heavy. But it suits the problem domain which is encoding/decoding serializing/deserializing data.

I also have the sense that there should somehow be a single or very few locations where a change is needed to solve the issue of not being able to JSON serialize datetimes when writing to, any plugin API really but in this case specifically the kubernetes plugin, but I suspect that the change needs to be made everywhere there's a call to `yaml.load` throughout the kubernetes_linux plugin. (at least everywhere for which the results of that load are then used to write to a k8s api.) Yeah I guess I'm thinking the "few places to make the change" are either there should maybe be a single primary entrypoint for the yaml.load configuration that is then used by the whole plugin library, or the solution could exist between when a call to the API client is made, and when the API client actually sends that request to k8s. <- This is just me thinking aloud, if there is a particular architectural direction to which you're beholden to meet some requirement, then obviously we should follow a path towards the intended architecture. I'd be interested in hearing some additional thoughts on this.

zenware:
Oh and I left one of the deserialized namespace objects on a JIRA comment, but I'll reproduce part of it here: the yaml loader is deserializing YAML dates into Python datatime objects, and then the Python JSON.dump can't magically auto-select a format for serializing the datetime into JSON.
```
              'creation_timestamp': datetime.datetime(2023, 2, 9, 5, 45, 21, tzinfo=datetime.timezone.utc),                                                                       
```

sonicaj:
@zenware can you share the output of the following command ?
```
midclt call k8s.namespace.export_to_yaml ix-traefik
```

So the problem essentially is that why i am not able to reproduce is that backup files which are created right now have dates escaped by single quotes, but that is not the case with the example file you shared.

Dates are still not relevant when we are creating namespaces but i would like to know why the dump output is different for you. Also if you create a recent backup yourself using `midclt call -job kubernetes.backup_chart_releases test_backup_yaml`, can you confirm the issue exists when you try to restore to this backup ? ( easy way might just be looking at the namespace files and seeing if the dates are escaped by single quotes )

zenware:
Huh... They are quote escaped in both of those cases, I wonder what happened
```
apiVersion: v1
kind: Namespace
metadata:
  creationTimestamp: '2023-02-28T04:09:07Z'
  labels:
    catalog: TRUECHARTS
    catalog_branch: main
    catalog_train: enterprise
    kubernetes.io/metadata.name: ix-traefik
  managedFields:
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:labels:
          .: {}
          f:catalog: {}
          f:catalog_branch: {}
          f:catalog_train: {}
          f:kubernetes.io/metadata.name: {}
    manager: Python
    operation: Update
    time: '2023-02-28T04:09:07Z'
  name: ix-traefik
  resourceVersion: '407921'
  uid: 77447fcb-524e-46a5-9ac3-83a82948b61d
spec:
  finalizers:
  - kubernetes
status:
  phase: Active
```

sonicaj (commented):
@zenware can you please fix the flake8 issues ?

So i found out why it was dumping yamls like that before and why it does not right now and why we have not faced this issue before.

We were using upstream kubernetes client first which provided us with datetime objects and they were dumped like that and loading them did not make a difference. When we made the transition to our own k8s client, that stopped happening - so any backups created with the new client and restored with the new client work but older backups ( like the one you showed is a year old ) would not.

src/middlewared/middlewared/plugins/kubernetes_linux/yaml.py
```diff
@@ -1,5 +1,27 @@
 import yaml
 
+class NoDatesFullLoader(yaml.FullLoader):
```
sonicaj, line 3: Also let's please rename this to something which is more or less reflecting reality which means that dates are not being removed but just used as strings

sonicaj:
Made the required fixes in https://github.com/truenas/middleware/pull/10904

PR #10741, opened 2023-02-24
Update grub config generation for production and debug kernels

usaleem-ix:
This commit updates the grub.d to generate GRUB menu entries for both production and debug kernels. The default boot entry present on the main menu should boot TrueNAS SCALE with production kernel.

A boot entry has been added under Advanced sub-menu to boot TrueNAS SCALE with debug kernel.

The advanced sub-menu previously had the exactly the same boot entry as present on the main menu. The only difference was in the title string so that entry has been cleaned up. GRUB also provides support to enter into recovery mode that is presented to user under Advanced sub-menu. Currently this option is not useful to us since we have not set up any recovery mode for SCALE, so GRUB_DISABLE_RECOVERY has been set to disable the boot entry for entering into recovery mode.


src/freenas/etc/grub.d/10_truenas_linux
```diff
@@ -135,7 +136,7 @@ linux_entry ()
 	  init-*)
 	      title="$(gettext_printf "%s, with Linux %s (%s)" "${os}" "${version}" "${type#init-}")" ;;
 	  *)
-	      title="$(gettext_printf "%s, with Linux %s" "${os}" "${version}")" ;;
+	      title="$(gettext_printf "%s, with %s" "${os}" "${extitle}")" ;;
```
themylogin, line 139: I see there are other calls of `linux_entry` without `$5`, what will the title be in that case?

src/freenas/etc/grub.d/10_truenas_linux
```diff
@@ -314,13 +315,11 @@ for dataset in `echo $datasets`; do
   fi
 
   truenas_version=`echo "$dataset" | sed "s|$roots_dataset/||"`
-  kernel_version=`zfs get -H -o value truenas:kernel_version "$dataset" 2>/dev/null`
-  if [ "x$kernel_version" = "x" -o "x$kernel_version" = "x-" ] ; then
-    continue
-  fi
+  kernel_version_prod=`dpkg -s linux-image-truenas-production-amd64 | grep '^Source:' | sed 's|Source: linux-||' 2>/dev/null`
```
themylogin, line 318: We can't do that. If we generate this list from, e.g., 23.10, `dpkg` will return kernel version used in 23.10, which will break 22.12 entries

themylogin (changes requested):

src/freenas/etc/grub.d/10_truenas_linux
usaleem-ix: The title would be Linux Kernel version string, which can be too long in some cases to be covered completely in that space provided in GRUB menu.

themylogin: I mean if `$5` is empty, then the title would be just `"TrueNAS SCALE 23.10, with "` which looks broken

usaleem-ix: Thanks, I had not thought of that, I will update this.

src/freenas/etc/grub.d/10_truenas_linux
usaleem-ix: Thanks, I will update this.

sonicaj (commented):
@themylogin we also have `debugkernel` field in `system.advanced`, iirc if it is set - the expectation is that next time system boots it will boot in debug kernel. I think we should account for that as well ?

usaleem-ix:
I have update the PR such that `debugkernel` from `system.advanced.config` is taken into account and if it is set to `false`, the default boot entry present on main menu of GRUB would boot SCALE with production kernel. Advanced sub-menu would contain the boot entry to debug kernel.

If `debugkernel` is set to `true`, the default entry on main menu of GRUB should boot SCALE with debug kernel. Advanced sub-menu would contain the boot entry to boot production kernel.

If the kernel version string does not contain `production` and `debug`, it means BE is from an earlier release, the boot entries for those BE are added as they were added before.

I have tested the changes by building an image and applying an update to an existing 22.12 installation. It is working as expected.

src/freenas/etc/grub.d/10_truenas_linux
```diff
@@ -320,7 +320,40 @@ for dataset in `echo $datasets`; do
   fi
 
   boot_device_id=
-  list="/boot/vmlinuz-$kernel_version"
+  # See if debugkernel is set to boot
+  debugkernel=`midclt call system.advanced.config | jq '.debugkernel' 2>/dev/null`
```
themylogin, line 324: This script can be ran by ISO installer, in that case middleware will not be available. You'll need to manually read the database file using sqlite3 CLI

themylogin (changes requested):

src/freenas/etc/grub.d/10_truenas_linux
usaleem-ix: Thanks, I have moved to reading the database file `/data/freenas-v1.db`. I have tested it by replacing the script on running system. That is working.  Building an image to test the install and update process from the ISO.

usaleem-ix: I have tested the update and installation in following scenarios with latest changes:

- Installing the .update file from WebUI.
- Installing the update by booting the ISO and upgrading an existing installation and preserving the previous BE.
- Fresh install on a system, destroying the previous installations and formatting the boot device.

In all the scenarios, update/install is successful. We are able to boot into previous and new installations from GRUB after install/update is successful. All the boot entries are functioning correctly and boot the system with selected entry. if debugkernel is set to true and `update-grub` is run, system boots with debug kernel on next boot from the entry that is selected bu default present on main menu of GRUB.

src/freenas/etc/grub.d/10_truenas_linux
```diff
@@ -320,7 +322,40 @@ for dataset in `echo $datasets`; do
...
+      # Boot with debug kernel if debugkernel is set
+      list="/boot/vmlinuz-$kernel_version_dbg"
+    else
+      # Boot with production kernel if debugkernel is not set
+      list="/boot/vmlinuz-$kernel_version_prod"
+    fi
+  else
+    # The dataset was not created with debug or production kernels
+    list="/boot/vmlinuz-$kernel_version"
+    # Set debugkernel to flase for this iteration
```
themylogin, line 355:
```suggestion
    # Set debugkernel to false for this iteration
```

themylogin (approved):

yocalebo (approved):
I've confirmed these changes work as intended.

PR #10742, opened 2023-02-24
NAS-120488 / 23.10 / fix API tests

yocalebo:
https://github.com/truenas/middleware/pull/10723 caused API tests regressions. This fixes it.


bmeagherix (commented):
Seems reasonable to fix existing tests.

However, it does make me wonder how useful those tests are.  AFAICS there are no tests for `home_create=False` even though this is now the default in `UserService.do_create`

PR #10748, opened 2023-02-27
dropped superfluous _0 suffix from some openapi schemas

das-intensity:
Due to the logic in `OpenAPIResource._accepts_to_request`, if a route took exactly 1 schema, it would have the schema suffixed with `_0`, e.g. for interface.create it would be `interface_create_0`. This suffix isn't necessary, so removing it to make the interface cleaner.

NOTE: This is only changing the schema **$ref** name/path, the `_name_` and `title` params within the schema were always correct.

Before:
interface-create-suffix

After:
interface-create-nosuffix

This may seem trivial, but will help projects which use the API schemas like [truenas-go-sdk](https://github.com/dariusbakunas/truenas-go-sdk).

Here's the old/new API docs (generated on top of https://github.com/truenas/middleware/pull/10747, although this PR is independent of that one):
[before-10748.txt](https://github.com/truenas/middleware/files/10835207/before-10748.txt)
[after-10748.txt](https://github.com/truenas/middleware/files/10835208/after-10748.txt)

Example:
```
$ diff before-10748.txt after-10748.txt  | grep interface_create
<         "$ref": "#/components/schemas/interface_create_0"
>         "$ref": "#/components/schemas/interface_create"
<    "interface_create_0": {
>    "interface_create": {
```

### DIFF
--- a/src/middlewared/middlewared/restful.py
+++ b/src/middlewared/middlewared/restful.py
@@ -354,35 +354,31 @@ class OpenAPIResource(object):
         }
 
     def _accepts_to_request(self, methodname, method, schemas):
-
         # Create an unique ID for every argument and register the schema
-        ids = []
-        for i, schema in enumerate(schemas):
-            if i == 0 and method['item_method']:
-                continue
-            unique_id = f'{methodname.replace(".", "_")}_{i}'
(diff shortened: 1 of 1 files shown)


sonicaj (approved):
Thank you @das-intensity for your contribution!

PR #10749, opened 2023-02-27
NAS-120521 / 23.10 / Mark metadata file for all catalogs as mandatory

sonicaj:
This PR adds changes to mark metadata file for all catalogs to be mandatory moving on in Cobia as the metadata file gives performance boost which is many times better then actually syncing the catalog itself in reality from middleware side.

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/items.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/items.py
@@ -1,4 +1,3 @@
-import functools
 import json
 import os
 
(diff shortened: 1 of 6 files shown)


src/middlewared/middlewared/plugins/catalogs_linux/items.py
```diff
@@ -166,33 +143,6 @@ def retrieve_trains_data_from_json(self, catalog, options):
...
-        # In order to calculate job progress, we need to know number of items we would be traversing
-        items = get_items_in_trains(trains_to_traverse, location)
-
-        job.set_progress(8, f'Retrieving {", ".join(trains_to_traverse)!r} train(s) information')
-
-        trains, unhealthy_apps = retrieve_trains_data(
-            items, location, catalog['preferred_trains'], trains_to_traverse, job, questions_context
-        )
-
-        if unhealthy_apps:
```
Qubad786, line 186: We would not be raising alert for unhealthy apps now?

Qubad786 (approved):

PR #10762, opened 2023-02-27
NAS-120534 / 23.10 / Fix removal of ctdb_shared_volume files during teardown

anodos325:
Typo in how we generated our list of paths to remove caused stale nodes file to exist which triggered issues for QE workflows to involve building / tearing down clusters.


yocalebo (approved):

anodos325:
Waiting confirmation from TC team that it works to resolve their issue.

PR #10770, opened 2023-02-28
NAS-120563 / 23.10 / Ensure we append path separator in SMB validation

anodos325:
When determining whether a mountpoint is a child of the current path, we should append path separator so that we don't generate false-positives in the following case

```
/mnt/dozer/SHARE
/mnt/dozer/SHARE2
```

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -1476,7 +1476,7 @@ class SharingSMBService(SharingService):
             return
 
         current_acltype = get_acl_type(this_mnt['super_opts'])
-        child_mounts = filter_list(list(mntinfo.values()), [['mountpoint', '^', path]])
+        child_mounts = filter_list(list(mntinfo.values()), [['mountpoint', '^', f'{path}/']])
         for mnt in child_mounts:
             if '@' in mnt['mount_source']:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -1476,7 +1476,7 @@ def get_acl_type(sb_info):
             return
 
         current_acltype = get_acl_type(this_mnt['super_opts'])
-        child_mounts = filter_list(list(mntinfo.values()), [['mountpoint', '^', path]])
+        child_mounts = filter_list(list(mntinfo.values()), [['mountpoint', '^', f'{path}/']])
```
bmeagherix, line 1479: This is probably fine, although if path already has a trailing "/" would this complicate things?

Could always do
`child_mounts = filter_list(list(mntinfo.values()), [['mountpoint', '^', path if path.endswith('/') else f'{path}/']])
`

bmeagherix (approved):

src/middlewared/middlewared/plugins/smb.py
anodos325: Good point. I switched to using `Path` from our schema validators, which normalizes the path with `os.path.normpath()`.

PR #10785, opened 2023-03-02
NAS-106086 / 23.10 / Add support for iSNS servers in scst

bmeagherix:
Add iSNS support.

Some additional context ...
### Add iSNS test
- The existing Linux(/FreeBSD) iSNS client software (included in [open-isns](https://github.com/open-iscsi/open-isns)) and MS iSNS server do not play very well together, so implemented a _minimal_ subset of iSNS (client) in python to side-step the issue when running tests.  (Only used in tests.)
- Added parameter `--isns_ip` to _runtest.py_ (with a default value pointing to our MS iSNS server)
- While in there, likewise added `--pool` (defaults to "tank" as was previously hard-coded).  Also allow `-s` to be passed thru to `pytest`.

### Add validate_isns_server
- Previously we were validating the format of the input, but not checking connectivity.  Add this.
- At the moment SCST only supports a single iSNS server, whereas the UI allows multiple.  Have implemented the checks for multiple for now, but will only _use_ the first provided.  This will need to be reconciled at a later time (i.e. either SCST or UI/middleware change).

### DIFF
--- a/src/middlewared/middlewared/plugins/iscsi_/iscsi_global.py
+++ b/src/middlewared/middlewared/plugins/iscsi_/iscsi_global.py
@@ -1,10 +1,10 @@
 import re
 
 import middlewared.sqlalchemy as sa
-
 from middlewared.async_validators import validate_port
-from middlewared.schema import accepts, Bool, Dict, Int, List, Str
-from middlewared.service import private, SystemServiceService, ValidationErrors
+from middlewared.schema import Bool, Dict, Int, List, Str, accepts
+from middlewared.service import SystemServiceService, ValidationErrors, private
+from middlewared.utils import run
 from middlewared.validators import IpAddress, Range
 
 RE_IP_PORT = re.compile(r'^(.+?)(:[0-9]+)?$')
@@ -75,8 +75,8 @@ class ISCSIGlobalService(SystemServiceService):
             self.middleware, 'iscsiglobal_update.listen_port', new['listen_port'], 'iscsi.global'
         ))
 
-        if verrors:
-            raise verrors
+        verrors.check()
+        licensed = await self.middleware.call('failover.licensed')
 
         new['isns_servers'] = '\n'.join(servers)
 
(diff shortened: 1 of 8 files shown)


yocalebo:
retest this please

src/middlewared/middlewared/plugins/iscsi_/iscsi_global.py
```diff
@@ -31,6 +33,66 @@ class Config:
...
+
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        if timeout:
+            s.settimeout(timeout)
+
+        try:
+            s.connect((host, port))
+            ret = True
+        except Exception as e:
+            self.logger.debug("connection to %s failed with error: %s",
```
yocalebo, line 48: For logging exceptions we like to try and use `exc_info=True` i.e.
```
exception Exception:
    self.logger.debug('connection failed %s blah...', server, exc_info=True)

File: src/middlewared/middlewared/plugins/iscsi_/iscsi_global.py [hunk 2]
```diff
@@ -85,8 +145,27 @@ async def do_update(self, data):
...
+        """
+        Unfortunately a SCST reload does not stop a previously active iSNS config, so
+        need to be able to perform an explicit action.
+        """
+        cp = await run([
+            'scstadmin', '-force', '-noprompt', '-set_drv_attr', 'iscsi',
+            '-attributes', 'iSNSServer=""'
+        ], check=False)
+        if cp.returncode:
+            self.middleware.logger.warning('Failed to stop active iSNS: %s', cp.stderr.decode())
```
yocalebo, line 167: Let's use `self.logger.warning`

tests/runtest.py
```diff
@@ -28,19 +30,21 @@
     print(config_file_msg)
     exit(1)
 
-error_msg = """Usage for %s:
+error_msg = f"""Usage for %s:
 Mandatory option
-    --ip <###.###.###.###>     - IP of the FreeNAS
-    --password <root password> - Password of the FreeNAS root user
-    --interface <interface>    - The interface that FreeNAS is run one
+    --ip <###.###.###.###>      - IP of the FreeNAS
```
yocalebo, line 35:
```suggestion
    --ip <###.###.###.###>      - IP of the TrueNAS
```

tests/runtest.py
```diff
@@ -28,19 +30,21 @@
...
+    --password <root password>  - Password of the FreeNAS root user
```
yocalebo, line 36:
```suggestion
    --password <root password>  - Password of the TrueNAS root user
```

yocalebo (commented):
I've left a few comments but overall looks great. I'll re-review again in a bit to make sure I don't spot anything.

yocalebo:
retest this please

yocalebo (approved):
I've no more qualms. Great job! However, I'd like @anodos325 to review as well to make sure I didn't miss anything obvious.

anodos325 (approved):

PR #10797, opened 2023-03-03
NAS-120588 / 23.10 / add drive identification capability to R30

yocalebo:


src/middlewared/middlewared/plugins/enclosure_/r30_drive_identify.py
```diff
@@ -0,0 +1,84 @@
...
+    if status != 'CLEAR':
+        cmds.append(f'{base} {status_map} {bay}')
+
+    # always go back to auto mode (for now)
+    cmds.append(f'{base} 0x3c 0x00')
+
+    # now subprocess (once) for the commands
+    cmds = '; '.join(cmds)
+    ret = run(cmds, stdout=PIPE, stderr=STDOUT, shell=True)
+    if ret.returncode != 0:
```
themylogin, line 83: Only return code of the last command will be checked, is this desired?

yocalebo: Yeah, it's a compromise I'm willing to make for this specific platform.

PR #10804, opened 2023-03-06
Add some casefolded filters

anodos325:


anodos325:
This is probably a bad idea

anodos325:
Will expand regular filters rather than hav

PR #10805, opened 2023-03-07
NAS-120643 / 23.10 / Add local smbtorture tests

anodos325:
These are relatively quick tests that ascertain whether internal libraries that samba uses work properly.


ericbsd (approved):
LGTM

PR #10808, opened 2023-03-07
NAS-120647 / 23.10 / Add some casefolded filters

anodos325:
Add special behavior where capital C prefix in filterop performs a casefolded version of
the filter. This is useful in cases where we're dealing with external protocols that are
case-insensitive, and will hopefully lead to fewer bugs related to case-insensitive checks.

### DIFF
--- a/src/middlewared/middlewared/utils/__init__.py
+++ b/src/middlewared/middlewared/utils/__init__.py
@@ -111,6 +111,19 @@ def get(obj, path):
     return cur
 
 
+def casefold(obj):
+    if obj is None:
+        return None
+
+    if isinstance(obj, str):
+        return obj.casefold()
+
+    if isinstance(obj, (list, tuple)):
+        return [x.casefold() for x in obj]
+
+    raise ValueError(f'{type(obj)}: support for casefolding object type not implemented.')
+
+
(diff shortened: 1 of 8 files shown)


anodos325:
Throwing WIP back on because of some CI failures.

anodos325:
Last custom jenkins run came up clean.

src/middlewared/middlewared/utils/__init__.py
```diff
@@ -111,8 +111,33 @@ def get(obj, path):
...
+
+    if isinstance(obj, (list, tuple)):
+        return [x.casefold() for x in obj]
+
+    raise ValueError(f'{type(obj)}: support for casefolding object type not implemented.')
+
+
 class filters(object):
-    opmap = {
+    case_opmap = {
```
themylogin, line 128: Can we just use operators from `opmap` but run `casefold` against arguments first? That way we won't have duplicate code.

themylogin (changes requested):
Also `casefold` seems confusing here, like there is some reason to perform indeed `casefold` and not, for example, just `lower`

themylogin (approved):

PR #10825, opened 2023-03-09
Cleanup pool.import_disk

themylogin:


themylogin:
time 1:00

yocalebo:
retest this please

src/middlewared/middlewared/plugins/pool_/import_disk.py
```diff
@@ -51,67 +118,65 @@ async def import_disk(self, job, device, fs_type, fs_options, dst_path):
         try:
             os.makedirs(src)
```
yocalebo, line 119: I'd really rather make all of this synchronous. Forcing all of this to be asynchronous doesn't help us in any way and only complicates everything unnecessarily. All of this is file/disk I/O which is, inherently, non-friendly to asyncio.

Also, there are still tons of areas in these methods that are calling blocking I/O in the main event loop. This is also why I'd rather make this all synchronous.

yocalebo (changes requested):

themylogin:
@anodos325 ping

themylogin:
@anodos325 ping

anodos325:
```
    @accepts(
        Str('device'),
        Str('fs_type'),
        Dict('fs_options', additional_attrs=True),
        Str('dst_path')
    )
```

Doesn't look like we're validating these parameters. Can we place an enum for fs_type?

anodos325:
```
        if os.path.exists(src):
            os.rmdir(src)
        try:
            os.makedirs(src)
```
Any idea what we were trying to avoid? I have a feeling we could probably just replace with `os.makedirs(src, exist_ok=True)` since we're just going to mount over the top of it anyway.

src/middlewared/middlewared/plugins/pool_/import_disk.py
```diff
@@ -1,20 +1,85 @@
...
+
+    proc = subprocess.Popen(
+        [executable] + arguments + [device, path],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+    output = proc.communicate()
+
+    if proc.returncode != 0:
+        logger.debug("Mount failed (%s): %s", proc.returncode, output)
```
anodos325, line 53: Should this be higher than debug?

src/middlewared/middlewared/plugins/pool_/import_disk.py
```diff
@@ -1,20 +1,85 @@
...
+        raise ValueError("Mount failed (exit code {0}):\n{1}{2}" .format(
+            proc.returncode,
+            output[0].decode("utf-8"),
+            output[1].decode("utf-8"),
+        ))
+    else:
+        return True
+
+
+class MountFsContextManager:
```
anodos325, line 63: Wouldn't it make more sense to have the mountpoint creation and deletion be part of the context manager?

themylogin:
> ```
>     @accepts(
>         Str('device'),
>         Str('fs_type'),
>         Dict('fs_options', additional_attrs=True),
>         Str('dst_path')
>     )
> ```
> 
> Doesn't look like we're validating these parameters. Can we place an enum for fs_type? (And maybe some device validation)

I am unsure how can we validate `device`. Right now mounting it will be the first action, and if it does not mount then the mount error will be properly presented, seems like a valid behavior to me.

yocalebo (approved):
Please wait for @anodos325 final review before merge.

anodos325 (approved):

PR #10826, opened 2023-03-09
fix enclosure alerts on SCALE

yocalebo:
I traced the logic that was put into this alert back to a commit done in 2019. The changes done back then essentially ignored all alerts for an enclosure. I honestly couldn't deduce what the intended reasoning was for choosing to do it the way it was done. Maybe, it was a simple mistake, who knows? Either way, I've fixed the alerts for enclosures on SCALE with these changes. I referenced the logic that I put into CORE (when I rewrote all of this over there) so this should be mostly feature (and logic) parity.

Tested this by pulling a PSU from an enclosure and ensuring the proper alert was generated.


yocalebo:
retest this please

themylogin (approved):

yocalebo:
retest this please

PR #10827, opened 2023-03-09
NAS-120302 / 22.12.2 / Allow users to add TPM to VMs

Qubad786:
## Context

Windows 11 has made it mandatory for system to have trusted platform module available for installation to succeed/work.
Also secure boot needs to be enabled for the VM so windows 11 considers it a healthy system. Looking at different configuration knobs which should be configured before windows 11 can be installed - profiles for VMs have been added similar to how we have certificate profiles.
These will provide sane defaults which UI can pre-populate based on the guest OS selected.


yocalebo (changes requested):
Middlewared is a pig for memory already so let's not cache this information on the class. I'd rather just return it in the method itself

yocalebo (approved):
Thanks!

PR #10838, opened 2023-03-10
NAS-120955 / 23.10 / Ensure required cgroup controllers are always present before starting apps

sonicaj:
This PR adds changes to ensure that required cgroup controllers are always present before attempting to start apps. We have been seeing this issue for various users where apps refused to work and kubernetes was not really giving us any good information wrt why it fails to start. It has been isolated that it is a kernel/systemd issue and not related to apps and OS team is working on isolating/diagnosing that appropriately. For now these changes will ensure we refuse to start apps and also raise an alert for apps when an attempt is made to start them outlining why it did not start.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
@@ -430,10 +430,10 @@ class KubernetesService(Service):
                     'ApplicationsConfigurationFailed',
                     {'error': e.errmsg},
                 )
-            else:
-                await self.middleware.call('alert.oneshot_delete', 'ApplicationsConfigurationFailed', None)
 
             raise
+
+        await self.middleware.call('alert.oneshot_delete', 'ApplicationsConfigurationFailed', None)
 
 
 async def _event_system_ready(middleware, event_type, args):
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
@@ -2,6 +2,7 @@ import asyncio
 import errno
 import json
 import os
+import re
 import shutil
 import uuid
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
```diff
@@ -376,10 +414,27 @@ def kubernetes_dataset_custom_props(self, ds: str) -> Dict:
...
+        try:
+            await self.middleware.call('kubernetes.validate_k8s_fs_setup')
+        except CallError as e:
+            if e.errno != CallError.EDATASETISLOCKED:
+                await self.middleware.call(
+                    'alert.oneshot_create',
+                    'ApplicationsConfigurationFailed',
+                    {'error': e.errmsg},
+                )
+            else:
```
yocalebo, line 433: This `else:` branch should be moved out of the exception scope right? Otherwise, an alert will be generated and potentially never removed? i.e.
```
try:
    blah
 except:
     create alert
 else:
     delete alert

yocalebo (changes requested):

src/middlewared/middlewared/plugins/kubernetes_linux/lifecycle.py
sonicaj: This alert is dismissed in another place as well when k8s starts so we know we are good but for this specific case we should have it outside the except scope. Thanks for pointing it out!

yocalebo (approved):
Nice!

frunkAF:
which cgroup.controllers file is referenced for this?

The code specifies `cgroup_root_path = '/sys/fs/cgroup'` which I think is `/sys/fs/cgroup/cgroup.controllers`? But that file has all of the required controllers listed.

I suspect that the one being referenced is `/sys/fs/cgroup/system.slice/middlewared.service/cgroup.controllers` because the file only contains memory and pids and the error I'm getting is that it's "Missing 'cpu, cpuset, hugetlb' cgroup controller(s)". Is that correct?

sonicaj:
@frunkAF the change essentially ensures that we have required cgroups controllers available and being reflected in `/sys/fs/cgroup/cgroup.subtree_control`

PR #10839, opened 2023-03-10
NAS-119515 / 23.10 / fix hot-spare not auto-detaching after replace

yocalebo:
This was added over 4 years ago for freeBSD based product. It is now, currently, interfering with the proper operation of zed and zfsd on CORE and SCALE respectively.


ixhamza:
Validated this on both Scale and Core with Manual disk replacement.

ixhamza (approved):

PR #10840, opened 2023-03-10
NAS-120608 / 23.10 / fix alert being generated on user caused failover

yocalebo:
The following scenario caused the alert to be generated during QE test cycle.
1. controller A is master controller
2. user initiates a failover via the webUI
3. controller B takes over as master and generates this alert

This alert shouldn't be generated since, almost always, a failover means a controller has been rebooted. This changes the logic just a bit to only generate this alert if we fail to sync the database after 20mins since some of the hardware platforms we sell can take roughly that amount of time to boot up.

### DIFF
--- a/src/middlewared/middlewared/plugins/failover_/datastore.py
+++ b/src/middlewared/middlewared/plugins/failover_/datastore.py
@@ -37,8 +37,8 @@ class FailoverDatastoreService(Service):
         self.failure = True
         try:
             self.send()
-        except Exception:
-            self.logger.warning('Error sending database to remote node on first replication failure')
+        except Exception as e:
+            self.logger.warning('Error sending database to remote node on first replication failure: %r', e)
 
             def send_retry():
                 set_thread_name('failover_datastore')
@@ -65,7 +65,6 @@ class FailoverDatastoreService(Service):
                         pass
 
                     if raise_alert_time <= 0 and self.failure:
-                        self.middleware.call_sync('alert.oneshot_delete', 'FailoverSyncFailed', None)
                         self.middleware.call_sync('alert.oneshot_create', 'FailoverSyncFailed', {'mins': total_mins})
                         raise_alert_time = RAISE_ALERT_SYNC_RETRY_TIME
 
--- a/src/middlewared/middlewared/plugins/failover_/datastore.py
+++ b/src/middlewared/middlewared/plugins/failover_/datastore.py
@@ -37,6 +37,8 @@ class FailoverDatastoreService(Service):
         try:
             self.send()
         except Exception:
+            self.logger.warning('Error sending database to remote node on first replication failure')
+
             def send_retry():
                 set_thread_name('failover_datastore')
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/failover_/datastore.py
```diff
@@ -37,17 +37,14 @@ def set_failure(self):
         try:
             self.send()
         except Exception:
-            self.logger.warning(
```
themylogin, line 40: Why remove this logging message? (But I think we should remove the traceback though)

src/middlewared/middlewared/plugins/failover_/datastore.py
```diff
@@ -37,17 +37,14 @@ def set_failure(self):
...
-            self.middleware.call_sync('alert.oneshot_create', 'FailoverSyncFailed', None)
-
             def send_retry():
                 set_thread_name('failover_datastore')
 
-                while True:
-                    time.sleep(60)
+                max_time_to_retry = 1200  # 20mins (some platforms take 15-20mins to reboot....)
+                sleep_time = 30
+                while max_time_to_retry > 0:
```
themylogin, line 45: This converts an infinite loop to infinite, is this change intended?

src/middlewared/middlewared/plugins/failover_/datastore.py
yocalebo: Agree, I added it back (without traceback)

src/middlewared/middlewared/plugins/failover_/datastore.py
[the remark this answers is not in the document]
yocalebo: Yeah, I believe that's okay? If we fail to sync db after 20mins, I don't see a reason to continue on, right? The user will be alerted to try and sync the db manually and if it fails again, this code will get started again and repeat the process?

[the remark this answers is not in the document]
themylogin: It has sense to continue the automatic retry infinitely, it is cheap enough. I see no reason why would someone want to trigger manual sync after an extensive period of other node being offline. At least we should allow greater threshold for the boot time

src/middlewared/middlewared/plugins/failover_/datastore.py
```diff
@@ -37,26 +38,24 @@ def set_failure(self):
         try:
             self.send()
         except Exception:
```
themylogin, line 40:
```suggestion
        except Exception as e:
```

src/middlewared/middlewared/plugins/failover_/datastore.py
```diff
@@ -65,6 +64,11 @@ def send_retry():
                     except Exception:
                         pass
 
+                    if raise_alert_time <= 0 and self.failure:
+                        self.middleware.call_sync('alert.oneshot_delete', 'FailoverSyncFailed', None)
```
themylogin, line 68: This will delete/remove alert every `RAISE_ALERT_SYNC_RETRY_TIME` seconds which will lead to e-mails being sent. Is this intended?

yocalebo: No it's not, I was just trying to avoid opening a duplicate alert if one had already been generated

yocalebo: Hmm, maybe an email alert every 20mins is actually decent.

themylogin (approved):

PR #10845, opened 2023-03-10
NAS-120571 / 23.10 / Ensure that share ACL is preserved

anodos325:
During registry synchronization we may have to remove shares if they are disabled. The standard tools for removing a share from samba's registry via libsmbconf also automatically remove the share's ACL. We work around this in normal situations in our share CRUD methods, but if for some reason these methods are bypassed, then the share ACL may be inadvertently deleted when we catch a share that shouldn't exist during registry synchronization.

This PR adds detection for shares that exist on the system but are inactive and in this case toggles the share ACL before removal from registry.

CI test is added that uses datastore plugin directly to simulate the CRUD methods being bypassed.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -1777,20 +1777,26 @@ class SharingSMBService(SharingService):
                                 "This may indicate that SMB service has not completed initialization.")
             return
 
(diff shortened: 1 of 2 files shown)


anodos325:
Will merge once test is verified to work as expected.

yocalebo (approved):
I've no qualms

PR #10850, opened 2023-03-11
NAS-119809 / 23.10 / `/usr/local/bin/truenas-autotune.py`

themylogin:


src/freenas/usr/local/bin/truenas-autotune.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+@zfs_parameter("l2arc_write_boost")
+def guess_vfs_zfs_l2arc_write_boost(context):
+    return 40000000
+
+
+@zfs_parameter("zfs_vdev_async_read_max_active")
+def guess_vfs_zfs_vdev_async_read_max_active(context):
+    if context.hardware == "Z50":
```
yocalebo, line 67: `context.hardware` will never be `Z50` only. It will be `TRUENAS-{Z50/M30/X20}` etc and sometimes we even tagged SMBIOS ending with `-HA` (i.e. `TRUENAS-M50-HA`). The check needs to be `context.hardware.startswith('TRUENAS-Z50')`.

src/freenas/usr/local/bin/truenas-autotune.py
```diff
@@ -0,0 +1,159 @@
...
+@zfs_parameter("zfs_dirty_data_max_max")
+def guess_vfs_zfs_dirty_data_max_max(context):
+    if context.hardware.startswith("M"):
+        return 12 * GiB
+    else:
+        return None
+
+
+@zfs_parameter("zfs_arc_max")
+def guess_vfs_zfs_arc_max(context):
```
yocalebo, line 39: This function can be simplified to.
```
physmem = int(context.physmem * .9)
if context.physmem / GiB > 200:
    physmem = int(context.physmem * .92)

return max(physmem - (context.kernel_reserved + context.userland_reserved), MIN_ZFS_RESERVED_MEM)

**Review by yocalebo — CHANGES_REQUESTED** [review]

File: src/freenas/usr/local/bin/truenas-autotune.py [hunk 3]
```diff
@@ -0,0 +1,159 @@
...
+    def decorator(func):
+        zfs_parameters[tunable_name] = func
+        return func
+
+    return decorator
+
+
+@zfs_parameter("zfs_dirty_data_max_max")
+def guess_vfs_zfs_dirty_data_max_max(context):
+    if context.hardware.startswith("M"):
```
yocalebo, line 32: Same thing here. This needs to only be applied to our M50 and M60

```
context.hardware.startswith(('TRUENAS-M50', TRUENAS-M60'))

File: src/middlewared/middlewared/plugins/system/dmi.py [hunk 4]
```diff
@@ -18,18 +21,17 @@ class SystemService(Service):
         'has-ipmi': None,
     }
 
-    @private
-    def dmidecode_info(self):
-        if all(v is None for k, v in SystemService.CACHE.items()):
+    def info(self):
```
yocalebo, line 24: Since this is a standard class outside of middleware, we should probably just decorate this function with `functools.cached_property` so we don't have to do the whole `self.CACHE` dance. Do note, however, we'll probably have to look at the unit tests that I wrote because I'm not sure how doing that will affect those off the top of my head.

themylogin: Yes, but what's the point of fixing something that already works?

src/freenas/usr/local/bin/truenas-autotune.py
themylogin: Forgot to pre-process `hardware` variable like the previous autotune did

yocalebo (approved):

PR #10864, opened 2023-03-14
NAS-121023 / 23.10 / Improve user homedir validation errors

anodos325:
Give explicit guidance in validation error messages about non-existent homedirs and also run homedir path through our standard method of path validation (to catch any more subtle errors in the path).

This also fixes an error where non-existent home directories were bypassing validation and error was being caught by filesystem plugin.

The PR also addresses issues where validation for user creation and update could potentially block main loop by running the homedir validation in a separate thread.

New tests were added to validate the following behavior:

1) Update of user to change home directory to an existing path.
   We check that path owner changed to user in question.

2) Creation of user with non-existing home directory base dir fails
   with a Validation Error.

3) Update of user to change home directory to new base and create
   new directory works and sets ownership correctly.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -5,6 +5,7 @@ from middlewared.service import (
 import middlewared.sqlalchemy as sa
 from middlewared.utils import run, filter_list
 from middlewared.validators import Email
+from middlewared.async_validators import check_path_resides_within_volume
 from middlewared.plugins.smb import SMBBuiltin
 
 import binascii
@@ -27,6 +28,7 @@ from contextlib import suppress
 ADMIN_UID = 950  # When googled, does not conflict with anything
 ADMIN_GID = 950
 SKEL_PATH = '/etc/skel/'
+DEFAULT_HOME_PATH = '/nonexistent'
 
 
 def pw_checkname(verrors, attribute, name):
@@ -116,7 +118,7 @@ class UserModel(sa.Model):
     bsdusr_username = sa.Column(sa.String(16), default='User &', unique=True)
     bsdusr_unixhash = sa.Column(sa.String(128), default='*')
     bsdusr_smbhash = sa.Column(sa.EncryptedText(), default='*')
-    bsdusr_home = sa.Column(sa.String(255), default="/nonexistent")
+    bsdusr_home = sa.Column(sa.String(255), default=DEFAULT_HOME_PATH)
     bsdusr_shell = sa.Column(sa.String(120), default='/bin/csh')
     bsdusr_full_name = sa.Column(sa.String(120))
     bsdusr_builtin = sa.Column(sa.Boolean(), default=False)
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/account.py
```diff
@@ -287,6 +289,84 @@ async def query(self, filters, options):
...
+    @private
+    def validate_homedir_path(self, verrors, schema, data, users):
+        needs_additional_validation = True
+        p = Path(data['home'])
+
+        if not p.is_absolute():
+            verrors.add(f'{schema}.home', '"Home Directory" must be an absolute path.')
+            return False
+
+        if p.is_file():
```
yocalebo, line 301: `p.is_file()` will return False even if the file/dir doesn't exist so there is a potential the validation error that is returned to end-user leads them astray. Probably want to do `if not p.exists()` first so by the time `p.is_file()` is called, ideally, it exists 😄

src/middlewared/middlewared/plugins/account.py
```diff
@@ -287,6 +289,84 @@ async def query(self, filters, options):
...
+
+        if data['home'] == DEFAULT_HOME_PATH:
+            return False
+
+        if not p.exists():
+            if data.get('home_create', False):
+                verrors.add(
+                    f'{schema}.home',
+                    f'{data["home"]}: path specified to use for home directory creation does not '
+                    'exist. TrueNAS uses the provided path as the parent directory of the '
```
yocalebo, line 317: You've got `of the of the` twice in the validation error.

src/middlewared/middlewared/plugins/account.py
```diff
@@ -287,6 +289,84 @@ async def query(self, filters, options):
...
+        if not data['home'].startswith('/mnt/'):
+            verrors.add(
+                f'{schema}.home',
+                '"Home Directory" must begin with /mnt/ or set to '
+                f'{DEFAULT_HOME_PATH}.'
+            )
+            needs_additional_validation = False
+        elif not any(
+            data['home'] == i['path'] or data['home'].startswith(i['path'] + '/')
+            for i in self.middleware.call_sync('pool.query')
```
yocalebo, line 347: Probably want to do something like this instead of calling `pool.query` since it can be particularly expensive (last time I checked)
```
elif not any(
    data['home'] == f'/mnt/{v["name"]}' or data['home'].startswith(f'/mnt/{v["name"]}/'
    for k, v in self.middleware.call_sync('zfs.pool.query_imported_fast').items()
)

yocalebo (changes requested):

src/middlewared/middlewared/plugins/account.py
anodos325: Yeah, I felt yucky shifting this validation :)

src/middlewared/middlewared/plugins/account.py
anodos325: Except we want to error-out if `is_file()` is True (which only happens if the path exists and is a file -- and not a directory).

yocalebo (approved):

PR #10865, opened 2023-03-14
NAS-121034 / 23.10 / generate a udev event after disk has been wiped

yocalebo:
When we wipe partition information from a disk using this method, we must open the device again to trigger udev events so the udev cache for the partition information for the disk that we just wiped is updated to reflect reality. Without these changes, middleware will still show a device as having partition labels on it when it doesn't.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/wipe.py
+++ b/src/middlewared/middlewared/plugins/disk_/wipe.py
@@ -11,7 +11,8 @@ class DiskService(Service):
 
     @private
     def _wipe(self, data):
-        with open(os.open(f'/dev/{data["dev"]}', os.O_WRONLY | os.O_EXCL), 'wb') as f:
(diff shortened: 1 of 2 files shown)


themylogin (commented):
Is this a rare issue or 100% reproducible?

yocalebo:
> Is this a rare issue or 100% reproducible?

100% reproducible unfortunately.

themylogin (commented):
It’s fortunate because in that case we can add a reliable integration test 🙂 It’s weird though that this worked before.

yocalebo:
> It’s fortunate because in that case we can add a reliable integration test 🙂 It’s weird though that this worked before.

It's reproducible on real hardware. Virtual hardware is a different animal. I'll have to see if it's an issue there

themylogin (approved):

themylogin (approved):

PR #10872, opened 2023-03-15
prevent more unnecessary log spam in disk.sync_all

yocalebo:
Troubleshooting, yet another, unrelated ticket I saw this method spam log files unnecessarily. In this particular scenario, I created the 1st zpool on an HA system and, by design, the standby controller is rebooted. However, the reboot is happening before `disk.sync_all` takes place so it spams log with unnecessary, scary, errors.


anodos325 (approved):

yocalebo:
backport

PR #10878, opened 2023-03-15
NAS-121058 / 23.10 / Only register HA VIPs during AD dynamic DNS updates

anodos325:
We must restrict SMB bind ips and dynamically registered IP addresses to TrueNAS HA Virtual IPs. Otherwise clients may attempt to connect to the wrong storage node.


anodos325:
I will hold off on merge pending solution to NAS-121074.

yocalebo (approved):

PR #10879, opened 2023-03-16
NAS-120616 / 23.10 / Added support for wildcarded NFS hostnames

mgrimesix:
See `man 5 exports` for list of valid wildcards.
Rules:
* Wildcards are allowed only in the left most level
* If the name is multiple levels, the levels below must be resolvable

Updated _exports.mako_ to allow wildcards.
Extended test 32 in _test_300_nfs.py_ to include wildcard hostnames.
Fixed a few `flake8` complaints.

### DIFF
--- a/src/middlewared/middlewared/plugins/nfs.py
+++ b/src/middlewared/middlewared/plugins/nfs.py
@@ -13,6 +13,7 @@ from middlewared.validators import Match, Range
 from middlewared.service import private, SharingService, SystemServiceService, ValidationError, ValidationErrors
 import middlewared.sqlalchemy as sa
 from middlewared.utils.asyncio_ import asyncio_map
+from middlewared.plugins.nfs_.utils import get_domain, leftmost_has_wildcards, get_wildcard_domain
 
 
 class NFSProtocol(str, enum.Enum):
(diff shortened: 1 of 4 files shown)


anodos325:
This looks good. I think since we have code duplication in the exports etc_file and NFS plugin, it might be time to create a utils.py file in the `nfs_` directory and put the validation functions there. Maybe look around and see if there are any other candidates for putting there as well. Note that you can precompile the regex like this:

```
>>> RE_DOMAIN_WILDCARD = re.compile(r'\*|\?|\[|\]')
>>> domain = "*.billy.goat"
>>> RE_DOMAIN_WILDCARD.search(domain.split('.')[0])
<re.Match object; span=(0, 1), match='*'>
```

Not that this is a hot code path, but we generally precompile them where possible.

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -445,11 +446,40 @@ async def validate(self, data, schema_name, verrors, old=None):
                     f"The following security flavor(s) require NFSv4 to be enabled: {','.join(v4_sec)}."
                 )
 
+    @private
+    def get_domain(self, hostname):
+        lst = hostname.split('.')
```
anodos325, line 451: I think you can do:
```
>>> domain = "*.billy.goat"
>>> domain.split('.', 1)
['*', 'billy.goat']
```
Then check length and return None as intended.

anodos325 (approved):

PR #10888, opened 2023-03-16
NAS-121072 / 23.10 / Fix bug to output hactl in debug file

william-gr:


yocalebo (changes requested):
SCALE HA removed the use of `ha-journal` so you can simplify this logic even more.

yocalebo (approved):

PR #10898, opened 2023-03-17
NAS-121091 / 23.10 / Add tests for MS-RPC server

anodos325:
This series of tests is just a thin wrapper around rpclient, and is a preliminary step to having formal tests for FSS support.

Right now we only check that shares are properly enumerated via srvsvc endpoint and that users and RIDs are presented as expeceted.

### DIFF
--- a/tests/api2/test_428_smb_rpc.py
+++ b/tests/api2/test_428_smb_rpc.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+
+import pytest
+import sys
+import os
+apifolder = os.getcwd()
+sys.path.append(apifolder)
+from auto_config import (ip, dev_test, pool_name)
+from functions import GET
+from middlewared.test.integration.assets.account import user
+from middlewared.test.integration.assets.smb import smb_share
+from middlewared.test.integration.assets.pool import dataset
+from protocols import MS_RPC
+
+reason = 'Skipping for test development testing'
+# comment pytestmark for development testing with --dev-test
+pytestmark = pytest.mark.skipif(dev_test, reason=reason)
+
+SMB_USER = "smbrpcuser"
+SMB_PWD = "smb1234#!@"
+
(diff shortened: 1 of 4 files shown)


tests/api2/test_428_smb_rpc.py
```diff
@@ -0,0 +1,132 @@
...
+
+reason = 'Skipping for test development testing'
+# comment pytestmark for development testing with --dev-test
+pytestmark = pytest.mark.skipif(dev_test, reason=reason)
+
+SMB_USER = "smbrpcuser"
+SMB_PWD = "smb1234#!@"
+
+
+@contextlib.contextmanager
```
themylogin, line 30: You might want to move this to `assets` package so other tests could use this too

tests/api2/test_428_smb_rpc.py
```diff
@@ -0,0 +1,132 @@
...
+    assert results.status_code == 200, results.text
+    id = results.json()["id"]
+
+    try:
+        yield id
+    finally:
+        result = DELETE(f"/sharing/smb/id/{id}/")
+        assert result.status_code == 200, result.text
+
+    assert results.status_code == 200, results.text
```
themylogin, line 45: Did you mean to do another `results = ...` call before this?

tests/api2/test_428_smb_rpc.py
```diff
@@ -0,0 +1,132 @@
...
+
+    try:
+        yield request
+
+    finally:
+        results = POST("/service/stop/", {"service": "cifs"})
+        assert results.status_code == 200, results.text
+
+
+@pytest.mark.dependency(name="SMB_USER_CREATED")
```
themylogin, line 66: You can make this a fixture too, and it will allow to create and delete the user in the same method

themylogin (changes requested):

themylogin (approved):

PR #10903, opened 2023-03-18
NAS-120069 / 23.10 / Add 2FA support for multiple users

sonicaj:


themylogin (approved):

src/middlewared/middlewared/plugins/account_/2fa.py
```diff
@@ -0,0 +1,45 @@
...
+        Returns the provisioning URI for the OTP for `user_id`. This can then be encoded in a QR code and used
+        to provision an OTP app like Google Authenticator.
+        """
+        user = await self.middleware.call('user.get_instance', user_id)
+        twofactor_config = await self.middleware.call('auth.twofactor.config')
+        user_twofactor_config = await self.middleware.call('auth.twofactor.get_user_twofactor_config', user_id)
+        if not user_twofactor_config['secret']:
+            raise CallError(f'{user["username"]!r} user does not has two factor authentication configured')
+
+        return pyotp.totp.TOTP(
```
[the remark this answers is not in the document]
sonicaj: Getting provisioning URI is good, the implementation is quite straight forward and i don't believe we would be blocked because of that

src/middlewared/middlewared/plugins/account_/2fa.py
```diff
@@ -0,0 +1,45 @@
...
+    async def verify_twofactor_token(self, user_id, token):
+        """
+        Returns boolean true if provided `token` is successfully authenticated for `user_id`.
+        """
+        twofactor_config = await self.middleware.call('auth.twofactor.config')
+        if not twofactor_config['enabled']:
+            raise CallError('Please enable Two Factor Authentication first')
+
+        user_twofactor_config = await self.middleware.call('auth.twofactor.get_user_twofactor_config', user_id)
+        totp = pyotp.totp.TOTP(
```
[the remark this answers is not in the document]
sonicaj: Changed the method to be synchronous

src/middlewared/middlewared/alembic/versions/23.10/2023-03-24_22-57_2fa_users.py
```diff
@@ -0,0 +1,54 @@
...
+    # 1) Have a model where we can keep user 2fa data
+    # 2) Have a model where global 2fa configurations can be enforced
+    # 3) Migrate
+
+    op.create_table(
+        'account_twofactor_user_auth',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('secret', sa.String(length=16), nullable=True),
+        sa.Column('user_id', sa.Integer(), nullable=True),
+        sa.ForeignKeyConstraint(
```
anodos325, line 30: How will this work with users from directory services? They won't be present in our sqlite database.

sonicaj: They would be handled in a subsequent PR - these changes only account for local users

anodos325: Even if you're not implementing some AD / LDAP wrapper now, we shouldn't make implementation decisions that prevent it in the future.

sonicaj: Sure, the idea is that when we add supoprt for them we would be adding another column for a unique id which will point to the AD user and we will be making appropriate changes then. So the current workflow is handled in a way so we can account for AD users in the future.

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -424,11 +426,16 @@ async def login(self, app, username, password, otp_token):
...
         twofactor_auth = await self.middleware.call('auth.twofactor.config')
 
-        if twofactor_auth['enabled']:
-            # We should run auth.twofactor.verify nevertheless of check_user result to prevent guessing
+        if twofactor_auth['enabled'] and (
+            user_obj := await self.middleware.call(
+                'user.query', [['username', '=', username], ['twofactor_auth_configured', '=', True]]
+            )
+        ):
+            # We should run user.verify_twofactor_token nevertheless of check_user result to prevent guessing
```
anodos325, line 434: s/nevertheless/regardless perhaps?

sonicaj: Done

src/middlewared/middlewared/plugins/account_/2fa.py
```diff
@@ -0,0 +1,45 @@
...
+
+from middlewared.schema import accepts, Bool, Int, returns, Str
+from middlewared.service import CallError, Service
+
+
+class UserService(Service):
+
+    @accepts(Int('user_id'))
+    @returns(Str(title='Provisioning URI'))
+    async def provisioning_uri(self, user_id):
```
anodos325, line 11: Is there any way we can use uid values rather than DB IDs? Former are easier to get for AD / LDAP users (since they don't exist in our sqlite database).

sonicaj: I believe those changes we can account for when we add AD support ? Motivation is that we can comprehensively discuss how to go about AD and make the changes. These changes are not stable yet and we can get started with AD right away ?

sonicaj: By stable i mean they are not going in a stable release right now :)

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -403,13 +401,17 @@ def get_token_for_shell_application(self, token_id, origin):
...
     @returns(Bool('two_factor_auth_enabled', description='Is `true` if 2FA is enabled'))
-    async def two_factor_auth(self):
+    async def two_factor_auth(self, username):
         """
         Returns true if two factor authorization is required for authorizing user's login.
         """
-        return (await self.middleware.call('auth.twofactor.config'))['enabled']
+        return (await self.middleware.call('auth.twofactor.config'))['enabled'] and (
+            await self.middleware.call(
+                'user.query', [['username', '=', username], ['twofactor_auth_configured', '=', True]]
```
anodos325, line 412: We'll have easier time with directory services if you got through the procedure of `user.get_user_obj` -> `pw_uid` -> `<the thing>` because AD usernames are case-insensitive and may be of form `DOMAIN\\username` or `username@DOMAIN` (both are valid and will be converted into a passwd struct via get_user_obj

sonicaj: Roger that. Will make appropriate changes

anodos325:
OK. LGTM on the condition that AD support gets added properly later. Please make sure that migration properly references latest one merged into master (I think one was merged recently-ish). I'll defer to Caleb for final approval.

src/middlewared/middlewared/alembic/versions/23.10/2023-03-24_22-57_2fa_users.py
```diff
@@ -0,0 +1,54 @@
+"""
+2FA support for multiple usrers
```
yocalebo, line 2: Typo

yocalebo (approved):

droidben:
[NAS-120154](https://ixsystems.atlassian.net/browse/NAS-120154) is related and if not already fixed by this, should also be considered.

PR #10907, opened 2023-03-18
Account for IP as well when validating if a port is in use

sonicaj:


src/middlewared/middlewared/common/ports/__init__.py
```diff
@@ -19,6 +19,7 @@ async def get_ports(self):
 
 
 class ServicePortDelegate(PortDelegate):
+    bind_address_field = NotImplementedError
```
themylogin, line 22: `NotImplementedError` is an exception class. There is a `NotImplemented` constant just for this purpose (or why not just `None`?)

src/middlewared/middlewared/common/ports/__init__.py
```diff
@@ -27,10 +28,28 @@ async def basic_checks(self):
...
+            config.get(self.bind_address_field) or default
+        )
+
+    def get_bind_ip_port_tuple(self, config, port_field):
+        return self.bind_address(config), config[port_field]
+
+    async def config(self):
+        return await self.middleware.call(f'{self.namespace}.config')
+
+    async def get_ports_internal_override(self):
```
themylogin, line 43: Why not give it a more descriptive name that tells us that this method returns the ports that we are listening on all IP addresses?

src/middlewared/middlewared/common/ports/__init__.py
```diff
@@ -27,10 +28,28 @@ async def basic_checks(self):
...
+        return []
+
     async def get_ports_internal(self):
+        if override_ports := await self.get_ports_internal_override():
+            return [('0.0.0.0', port) for port in override_ports]
```
themylogin, line 48: Do we also need an IPv6 equivalent?

src/middlewared/middlewared/common/ports/__init__.py
```diff
@@ -3,34 +3,53 @@
...
+            config.get(self.bind_address_field) or default
+        )
+
+    def get_bind_ip_port_tuple(self, config, port_field):
+        return self.bind_address(config), config[port_field]
+
+    async def config(self):
+        return await self.middleware.call(f'{self.namespace}.config')
+
+    async def get_ports_binded_on_wildcards(self):
```
themylogin, line 43: `bound`

themylogin (approved):

yocalebo (changes requested):
I'll add my review in a bit with my suggested changes.

src/middlewared/middlewared/common/ports/__init__.py
yocalebo, line 43:
```suggestion
    async def get_ports_bound_on_wildcards(self):
```

src/middlewared/middlewared/plugins/ports/ports.py
```diff
@@ -1,9 +1,21 @@
...
 from middlewared.service import Service, ValidationErrors
 
+from .utils import WILDCARD_IPS
+
+
+SYSTEM_PORTS = [(wildcard, port) for wildcard in WILDCARD_IPS for port in [67, 123, 3702, 5353, 6000]]
+
 
-SYSTEM_PORTS = [67, 123, 3702, 5353, 6000]
+def check_ip_is_ipv6(ip: str) -> bool:
```
yocalebo, line 13: Could we use the standard `ipaddress` module to determine the ip address? This could be simplified to 1 function instead of 2.

Also, this is not returning the "protocol". This is returning the IP version. Something like:

```
def get_ip_version(ip: str) -> int:
    return ipaddress.ip_interface(ip).version
```

yocalebo (changes requested):

src/middlewared/middlewared/pytest/unit/plugins/test_port_attachments.py
```diff
@@ -0,0 +1,604 @@
+import contextlib
```
yocalebo, line 1: I see no tests for IPv6 addresses (or an IPv6 wildcard)

yocalebo (changes requested):

src/middlewared/middlewared/pytest/unit/plugins/test_port_attachments.py
sonicaj: Yup on it already

src/middlewared/middlewared/plugins/ports/ports.py
```diff
@@ -1,9 +1,18 @@
...
 from middlewared.service import Service, ValidationErrors
 
+from .utils import WILDCARD_IPS
+
+
+SYSTEM_PORTS = [(wildcard, port) for wildcard in WILDCARD_IPS for port in [67, 123, 3702, 5353, 6000]]
 
-SYSTEM_PORTS = [67, 123, 3702, 5353, 6000]
+
+def get_ip_protocol(ip: str) -> int:
```
yocalebo, line 14:
```suggestion
def get_ip_version(ip: str) -> int:
```

yocalebo (changes requested):

src/middlewared/middlewared/plugins/ports/ports.py
```diff
@@ -40,20 +52,51 @@ async def get_in_use(self):
...
+            return verrors
+
+        ip_errors = []
+        for index, port_detail in enumerate(port_attachment.items()):
+            ip, port_entry = port_detail
+            if get_ip_protocol(ip) != bindip_protocol:
+                continue
+
+            if bindip == wildcard_ip or ip == wildcard_ip or (bindip != wildcard_ip and ip == bindip):
+                entry = next(
```
yocalebo, line 75: This will raise a `StopIteration` if a match isn't found. Are you expecting this method to crash here with that type of traceback?

yocalebo (changes requested):

yocalebo (approved):

sonicaj:
backport

PR #10910, opened 2023-03-19
Add ACME DNS support for Google DNS and DigitalOcean

Dogeek:
Provided by the certbot packages.

NAS-121092 seems to mention the need for Google Domains support as well.

Since the `certbot_dns_google` and `certbot_dns_digitalocean` are available out of the box, I thought I'd kill two birds with one stone.

### DIFF
--- a/src/middlewared/middlewared/plugins/acme_protocol_/authenticators/google.py
+++ b/src/middlewared/middlewared/plugins/acme_protocol_/authenticators/google.py
@@ -1,0 +1,41 @@
+
+
+import logging
+
+from certbot_dns_google._internal.dns_google import _GoogleClient
+
+from middlewared.schema import accepts, Dict, File
+from middlewared.service import skip_arg
+
+from .base import Authenticator
+
+
+logger = logging.getLogger(__name__)
+
+
+class GoogleAuthenticator(Authenticator):
+
+    NAME = 'Google DNS'
+    PROPAGATION_DELAY = 60
+    SCHEMA = Dict(
+        'Google',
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/acme_protocol_/authenticators/google.py
```diff
@@ -0,0 +1,39 @@
...
+logger = logging.getLogger(__name__)
+
+
+class GoogleAuthenticator(Authenticator):
+
+    NAME = 'Google DNS'
+    PROPAGATION_DELAY = 60
+    SCHEMA = Dict(
+        'Google',
+        File('service_account_path', empty=False, null=False, title='Path to the service account JSON', required=True),
```
sonicaj, line 20: Let's please add this explicitly instead of a file path as this is not reliable and introduces other issues.

So what can be done is either have the user specify the complete json file here in the field as a string and we can do the magic to convert it to json before giving it to the client and also ofc validating it is valid json.
Another way might be if the json file is not huge, we can ask for explicit attrs in the json file and then pass them over to the google auth client.
Can you post a sample json file here as well for clarity please ? Thanks!

sonicaj (commented):
LGTM, thank you @Dogeek for your contribution. Can you confirm if you have tested both the authenticators and made some minor comment(s).

bamsammich:
@yocalebo @Dogeek what was the reason for this PR getting closed? I use Google DNS and would really like to see this added.

yocalebo:
> @yocalebo @Dogeek what was the reason for this PR getting closed? I use Google DNS and would really like to see this added.

No response from author after we asked for changes.

PR #10918, opened 2023-03-20
NAS-121017 / 23.10 / fix disk.retaste and, hopefully, SCALE HA failover events

yocalebo:
This does 2 primary things:
1. in `disk.retaste` method, I now call `BLKRRPART` which tells kernel to force reread partition tables on the disk. (I also log any type of errors that might occur while I'm there)
2. Remove the `disk.retaste` call on failover event since that seems to have caused all of the problems that we've been experiencing.

### DIFF
--- a/src/middlewared/middlewared/plugins/failover_/event.py
+++ b/src/middlewared/middlewared/plugins/failover_/event.py
@@ -417,7 +417,8 @@ class FailoverEventsService(Service):
             except Exception as e:
                 if e.errno == errno.ENOENT:
                     try_again = True
-                    logger.warning('Failed importing %r using cachefile so trying without it.', vol['name'])
+                    # logger.warning('Failed importing %r using cachefile so trying without it.', vol['name'])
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/failover_/event.py
```diff
@@ -443,13 +435,18 @@ def vrrp_master(self, job, fobj, ifname, event):
                     vol['error'] = str(e)
                     failed.append(vol)
                     continue
+
+                # TODO: come back and fix this once we figure out how to properly manage zpool cachefile
+                # (i.e. we need a cachefile per zpool, and not a global one)
```
themylogin, line 440: What prevents us from implementing per-pool cache file right now?

yocalebo: Time. This is currently blocking 22.12.2 release.

themylogin (approved):

PR #10920, opened 2023-03-21
NAS-121085 / 23.10 / Make `core.job_wait` forward the job progress

themylogin:


themylogin:
Tests pass https://ci.tn.ixsystems.net/jenkins/job/TrueNAS%20SCALE%20-%20Unstable/job/API%20Tests%20-%20TrueNAS%20SCALE%20with%20middleware%20reinstalled%20from%20a%20branch%20Incremental/728/

yocalebo (commented):
How does `self.wrapped` get cleared?

themylogin:
@yocalebo why does it need to be cleared? The whole job that contains `wrapped` will be cleared when 1000 newer jobs eject it from the deque.

yocalebo (approved):

PR #10937, opened 2023-03-22
Fix failover via rebooting active controller

anodos325:
When systemd is handling reboot of an HA server, a race condition may be exposed whereby fenced is stopped before pool is exported. This can result in the now-passive controller halting IO to the zpool and hanging in a state that requires server reset in order to recover. This PR makes a couple of minor changes:

1) On active (rebooting) controller, we no longer stop the fenced
   process when processing a shutdown event in middleware.

2) On standby (becoming active) controller, we poll the fenced for
   remote node for up to 10 seconds (repeating if other side has
   fenced running) to give time for fenced to stop / controller
   reboot.


yocalebo (approved):
I need to come up with a better way of managing keepalived and middlewared service when `shutdown` or `reboot` binaries are called from cli. However, this is least intrusive way of implementing a mitigation.

PR #10973, opened 2023-03-27
NAS-121032 / 23.10 / Use `lagg_ordernum` to determine the correct child interface to grab …

themylogin:
…the MAC address for bond


yocalebo (commented):
I need to think about this before merging.

yocalebo (approved):

PR #10984, opened 2023-03-28
NAS-121216 / 23.10 / properly enable/disable systemd units

yocalebo:
Don't enable/disable systemd units if they're already enabled/disabled respectively.

PR #10985, opened 2023-03-28
properly enable/disable systemd units

yocalebo:


yocalebo:
This was approved in https://github.com/truenas/middleware/pull/10984 but I named the branch wrong so merging on that condition.

PR #10994, opened 2023-03-28
NAS-121206 / 23.10 / Expand validation of user home directories

anodos325:
We should catch following cases and turn them into informative validation errors:

* New home directory doesn't exist and parent directory is on RO dataset
* New home directory doesn't exist and parent directory is immutable
* New home directory exists and is on RO dataset or immutable

Since we're already having to get mount information, this commit adds an additional check for filesystem type in the returned mount info. If any validation errors have been added, then skip more expensive checks using libzfs.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -302,8 +302,25 @@ class UserService(CRUDService):
         )
 
     @private
+    def validate_homedir_mountinfo(self, verrors, schema, dev):
+        mntinfo = self.middleware.call_sync(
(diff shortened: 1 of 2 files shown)


themylogin (changes requested):
We can easily add integration tests here

william-gr:
This has been lingering for some time.
Wonder if we should merge and create another ticket for the follow up.

anodos325:
I can add tests soon-ish. We were going through variations of having various jenkins things broken when I was preparing to write tests.

yocalebo (approved):

PR #11002, opened 2023-03-28
NAS-120957 / 23.10 / Update NFS /etc/exports validations

mgrimesix:
Updated validation checks:
* Allow sharing unrelated directories
* Prevent sharing related directories to 'same' hosts

### DIFF
--- a/src/middlewared/middlewared/plugins/nfs.py
+++ b/src/middlewared/middlewared/plugins/nfs.py
@@ -1,7 +1,9 @@
+import asyncio
 import contextlib
 import enum
 import ipaddress
 import os
+import socket
 
 from middlewared.common.attachment import LockableFSAttachmentDelegate
 from middlewared.common.listen import SystemServiceListenMultipleDelegate
@@ -467,7 +469,9 @@ class SharingNFSService(SharingService):
                         return None
                     else:
                         try:
-                            return (await self.middleware.call('dnsclient.forward_lookup', {"names": [hostname]}))[0]['address']
+                            return (
+                                await asyncio.wait_for(self.middleware.run_in_thread(socket.getaddrinfo, hostname, None), 5)
+                            )[0][4][0]
                         except Exception as e:
                             self.logger.warning("Unable to resolve host %r: %r", hostname, e)
                             return None
--- a/src/middlewared/middlewared/plugins/nfs.py
+++ b/src/middlewared/middlewared/plugins/nfs.py
@@ -1,15 +1,13 @@
-import asyncio
 import contextlib
 import enum
 import ipaddress
 import os
-import socket
 
 from middlewared.common.attachment import LockableFSAttachmentDelegate
 from middlewared.common.listen import SystemServiceListenMultipleDelegate
 from middlewared.schema import accepts, Bool, Dict, Dir, Int, IPAddr, List, Patch, returns, Str
 from middlewared.async_validators import check_path_resides_within_volume, validate_port
-from middlewared.validators import Match, Range
+from middlewared.validators import Match, Range, IpAddress
 from middlewared.service import private, SharingService, SystemServiceService, ValidationError, ValidationErrors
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -547,12 +554,47 @@ def validate_hosts_and_networks(self, other_shares, data, schema_name, verrors,
...
+            found_match = False
+            pp_oldshr = PurePath(share['path'])
+            if pp_newshr.is_relative_to(pp_oldshr):
+                lowerpath = pp_newshr
+                upperpath = pp_oldshr
+                found_match = True
+            if pp_oldshr.is_relative_to(pp_newshr):
+                lowerpath = pp_oldshr
+                upperpath = pp_newshr
+                found_match = True
```
anodos325, line 574: We can probably use `filesystem.is_child` for the path checks

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -14,6 +14,7 @@
 import middlewared.sqlalchemy as sa
 from middlewared.utils.asyncio_ import asyncio_map
 from middlewared.plugins.nfs_.utils import get_domain, leftmost_has_wildcards, get_wildcard_domain
+from pathlib import PurePath
```
yocalebo, line 17: For standard library modules, we like to keep these imports at the top of the file.

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -454,13 +457,17 @@ async def resolve(hostname):
...
-            try:
-                return (
-                    await asyncio.wait_for(self.middleware.run_in_thread(socket.getaddrinfo, hostname, None), 5)
-                )[0][4][0]
-            except Exception as e:
-                self.logger.warning("Unable to resolve host %r: %r", hostname, e)
+            if leftmost_has_wildcards(hostname):
+                # We know this will not resolve
                 return None
+            else:
```
yocalebo, line 463: Since you're here and making changes, we have a dns client that is async native so no reason to create a native thread to run `socket.getaddrinfo`. Name of the function is `dnsclient.forward_lookup` which takes a list of hostnames. You should be able to grep our code-base and see example usages of this.

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -547,12 +554,47 @@ def validate_hosts_and_networks(self, other_shares, data, schema_name, verrors,
...
-                verrors.add(
-                    f"{schema_name}.networks",
-                    "Another NFS share already exports this dataset for some network"
-                )
+    @private
+    async def validate_share_path(self, other_shares, data, schema_name, verrors):
+        """
+        Checks new share path against existing
+        """
+        realPath = os.path.realpath(data["path"])
```
yocalebo, line 562: Since this method is a coroutine (`async def validate_share_path`), performing file I/O will block the main event loop since `os.path.realpath` just calls `lstat` underneath the hood.

yocalebo (changes requested):
Great job! I've made a few suggestions.

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -451,15 +451,28 @@ async def resolve_hostnames(self, hostnames):
...
+                    # Not an IP address, should be a name
+                    if domain := get_wildcard_domain(hostname):
+                        hostname = domain
+
+                    if leftmost_has_wildcards(hostname):
+                        # We know this will not resolve
+                        return None
+                    else:
+                        try:
+                            return (await self.middleware.call('dnsclient.forward_lookup', {"names": [hostname]}))[0]['address']
```
anodos325, line 470: This API only allows searching one record type per query (defaulting with A), do we ned to be concerned about AAAA?

anodos325 (approved):

yocalebo (approved):

PR #11003, opened 2023-03-28
Enable cgroup controllers at boot

sonicaj:
This PR adds changes to enable cgroup controllers at boot. Motivation behind the change is that OS team found if some process has `PF_NO_SETAFFINITY ` flag set, we are not able to add cpuset controller. This happens when SMB service is started and it has shares mounted which starts a process which has the flag which means that after that we can't enable the `cpuset` controller.

There are 2 approaches to solve the problem:
1. Enable cgroup controllers before everything else
2. Have a firewall in place which does not allow users to connect before the system is ready

After discussion with Andrew, we have done point 1 and Andrew will be syncing with services team to see if we should factor in other services as well for the firewall and those changes can still be added separately as they still bring value to the table.


yocalebo (changes requested):
We should be more strict with this unit file and make it so that it can be called before middlewared.service explicitly.

This means we need to move the setup of the cgroup related items to a standalone python script and then call it in the `ix-cgroups` file. You can look in `middlewared/scripts/wait_on_disks.py` as a reference.

PR #11007, opened 2023-03-28
NAS-121195 / 23.10 / Sync catalogs if they have not been synced

sonicaj:
This PR adds changes to account for the fact that if catalogs have not been synced, we sync them when `app.available` is called. Why this is required is that we don't sync catalogs at all unless user configures apps or navigates to available apps. Earlier `catalog.items` used to account for this and it made sure catalog was synced before we retrieved it's contents but that is not the case anymore as `catalog.items` is not a job anymore.

Earlier, we used to ensure that official catalog is synced in such cases and similar changes have been introduced now so we keep backwards compatibility.

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/sync_catalogs.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/sync_catalogs.py
@@ -9,6 +9,8 @@ from .utils import pull_clone_repository
 
 class CatalogService(Service):
 
+    SYNCED = False
+
     @accepts()
     @returns()
     @job(lock='sync_catalogs')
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/catalogs_linux/sync_catalogs.py
```diff
@@ -70,3 +74,12 @@ def update_git_repository(self, catalog):
         return pull_clone_repository(
             catalog['repository'], os.path.dirname(catalog['location']), catalog['branch'],
         )
+
+    @private
```
yocalebo, line 78: We have a @cache decorator that can be used so we don't have to create a class attribute

yocalebo (approved):
Just a minor nit

PR #11033, opened 2023-04-03
NAS-120230 / 23.10 / Properly mark a gpu as critical if it is critical to the system

Qubad786:
## Context

Logic where we determine if a gpu is critical for the host to function had been outdated with what we have for PCI devices. The same logic is now used in both places to determine critical devices to make sure that a GPU is properly marked as critical.


src/middlewared/middlewared/plugins/vm/utils.py
```diff
@@ -5,6 +5,12 @@
 LIBVIRT_URI = 'qemu+unix:///system?socket=/run/truenas_libvirt/libvirt-sock'
 LIBVIRT_USER = 'libvirt-qemu'
 NGINX_PREFIX = '/vm/display'
+SENSITIVE_PCI_DEVICE_TYPES = (
+    'Host bridge',
```
sonicaj, line 9: This is not required as just validating `bridge` covers this entry as well

PR #11037, opened 2023-04-03
NAS-121304 / 23.10 / Improve logging in workflow for UI

sonicaj:
This PR adds changes so UI can consume the following workflow for logging in users:

1. Users present a login screen where username/password is only asked
2. UI calls `auth.two_factor_auth` where UI passes credentials and endpoint will return if the user has 2FA configured. It will only say 2FA is configured when the credentials are good and 2FA is actually configured.
3. Based on step 2 output, UI will show a new screen where it will ask for OTP
4. With/without OTP, after step 3 - UI will finally call `auth.login` to login the said user


themylogin (commented):
Timing attack again. We need to sleep a random time on invalid login/password here and everywhere else.

themylogin (approved):
5 seconds is too long

yocalebo (approved):

PR #11055, opened 2023-04-05
NAS-121348 / 23.10 / fix proactive support alert verbiage

yocalebo:


src/middlewared/middlewared/alert/source/proactive_support.py
```diff
@@ -34,5 +31,5 @@ async def check(self):
             if unfilled:
                 return Alert(
                     ProactiveSupportAlertClass,
-                    'Please complete these fields on the System/Proactive Support page: %s.' % ', '.join(unfilled)
+                    f'Please complete these fields on the {webui_page}: {", ".join(unfilled)}'
```
themylogin, line 34: It will be Please complete these fields on the System Settings->General->Support page.: (note the .:)

themylogin (changes requested):

themylogin (approved):

PR #11068, opened 2023-04-07
NAS-121376 / 23.10 / Add 2FA support for AD users

sonicaj:
This PR adds support for AD users to be able to consume 2FA support. The workflow now would be following for consuming 2fa:

1. 2FA is enabled/disabled globally
2. If a user does not has his secret configured before, he will be able to login via UI but he won't be able to use SSH unless he configures his 2FA secret (provided 2FA is enabled for ssh service)
3. AD/local users will consume the same endpoint to renew/configure their 2fa secret i.e `user.renew_2fa_secret`

For the UI, workflow would be the following:
1. User opens login screen
2. UI makes a call to `auth.two_factor_auth` and specifies username/password and the endpoint will tell if UI should be asking for OTP for the user in question or not
3. If in (2), it returned boolean true - then UI will ask for OTP from user and use `auth.login` or if OTP is not to be asked for it will just use `auth.login` directly
4. If an admin is creating another user, that user will not have 2FA secret configured and he will have to configure it himself.
5. There will be a dedicated screen for 2FA settings where global 2FA settings will be presented to the user (`auth.twofactor.config` basically) and he will be presented with his personal 2FA configuration which will be as follows:
      - a) A button which says `Renew 2FA secret` if `twofactor_auth_configured` was `true` for the user from `user.query` output and if it was not it will say `Configure 2FA secret`. Regardless of the text on button, when user presses the button - it will make a call to `user.renew_2fa_secret` which will configure the secret for the user in question. This applies to both AD and local users and UI does not has to distinguish between them.
      - b) If 2FA for the user is configured, he should be allowed to get a QR code which he will use in his/her authenticator app to configure 2FA (UI will consume `user.provisioning_uri` endpoint for this)

Implementation wise what is happening is that for all local users we have a record in the database which points to that user's 2fa config regardless if he has secret configured/unconfigured. Whenever an AD user configures a secret we add his record to the 2FA table. For local users, when they are deleted so are is their 2fa record and for AD users we have a periodic endpoint which runs once daily and prunes out users which are no longer available on the AD network.

Changes have also been added to make sure we are able to normalize any username provided to us and get that user's configuration i.e endpoint will get us user's config even if he is a local or an AD user based just on his username and it covers all the flavours of usernames which we find with AD users.
The `user.query` method is now capable to give us `sid` of AD users as well and when the user in question is local, the `sid` attribute points to SMB based sid and if it is not - it points out the AD sid of the user in question.

**Note:** This has been discussed already but adding an explicit note, if globally 2FA is enabled including SSH service - regardless of the user having configured 2FA he won't be able to SSH into the system unless he configures his 2FA secret. This only applies to SSH and UI/middleware work just fine. This is a catch based on the underlying pam module we use for validating OTP.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -647,7 +647,6 @@ class UserService(CRUDService):
         Patch(
             'user_create',
             'user_update',
-            ('add', Bool('renew_twofactor_secret', default=False)),
             ('attr', {'update': True}),
             ('rm', {'name': 'group_create'}),
         ),
@@ -768,7 +767,6 @@ class UserService(CRUDService):
                 data['home'] = os.path.join(data['home'], data.get('username') or user['username'])
 
         # After this point user dict has values from data
-        renew_2fa_secret = data.pop('renew_twofactor_secret', False)
         user.update(data)
 
         mode_to_set = user.get('home_mode')
@@ -826,15 +824,6 @@ class UserService(CRUDService):
         if 'groups' in user:
             groups = user.pop('groups')
             await self.__set_groups(pk, groups)
-
-        if renew_2fa_secret:
-            twofactor_auth_id = (await self.middleware.call('auth.twofactor.get_user_config', pk))['id']
-            await self.middleware.call(
-                'datastore.update',
-                'account.twofactor_user_auth',
-                twofactor_auth_id,
-                {'secret': await self.middleware.call('auth.twofactor.generate_base32_secret')}
-            )
 
         user = await self.user_compress(user)
         await self.middleware.call('datastore.update', 'account.bsdusers', pk, user, {'prefix': 'bsdusr_'})
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -181,7 +181,6 @@ class UserService(CRUDService):
 
         return {
             'memberships': memberships,
-            'global_2fa_configured': (await self.middleware.call('auth.twofactor.config'))['enabled'],
             'user_2fa_mapping': ({
                 entry['user']['id']: bool(entry['secret']) for entry in await self.middleware.call(
                     'datastore.query', 'account.twofactor_user_auth', [['user_id', '!=', None]]
@@ -211,7 +210,7 @@ class UserService(CRUDService):
         user['sshpubkey'] = await self.middleware.run_in_thread(self._read_authorized_keys, user['home'])
 
         user['immutable'] = user['builtin'] or (user['username'] == 'admin' and user['home'] == '/home/admin')
-        user['twofactor_auth_configured'] = ctx['global_2fa_configured'] and bool(ctx['user_2fa_mapping'][user['id']])
+        user['twofactor_auth_configured'] = bool(ctx['user_2fa_mapping'][user['id']])
 
         return user
 
@@ -280,12 +279,7 @@ class UserService(CRUDService):
         if dssearch:
             dssearch_results = await self.middleware.call('dscache.query', 'USERS', filters, options.copy())
             # For AD users, we will not have 2FA attribute normalized so let's do that
-            global_2fa_configured = (await self.middleware.call('auth.twofactor.config'))['enabled']
-            if global_2fa_configured:
-                ad_users_2fa_mapping = await self.middleware.call('auth.twofactor.get_ad_users')
-            else:
-                ad_users_2fa_mapping = {}
-
+            ad_users_2fa_mapping = await self.middleware.call('auth.twofactor.get_ad_users')
             for index, user in enumerate(filter(
                 lambda u: not u['local'] and 'twofactor_auth_configured' not in u, dssearch_results)
             ):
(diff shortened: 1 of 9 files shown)


src/middlewared/middlewared/plugins/auth_/2fa.py
```diff
@@ -88,23 +89,74 @@ async def do_update(self, data):
...
+
+        mapping = {
+            user['sid']: user for user in self.middleware.call_sync(
+                'user.query', [['local', '=', False], ['sid', '!=', None]], {
+                    'extra': {'additional_information': ['DS']},
+                }
             )
-        ]
+        }
+        for sid, entry in (await self.get_ad_users()).items():
```
anodos325, line 160: We can't rely on our cache to always be up-to-date. If you need to check whether a SID still maps to an AD user, then you can try `idmap.sid_to_name` call. libwbclient supports bulk conversion of sids to names. I can expose that API if you wish (then you can send it a large list of SIDs and only remove the ones that don't resolve).

anodos325: sid_to_name will raise `MatchNotFound` if it doesn't resolve.

src/middlewared/middlewared/plugins/account.py
```diff
@@ -278,7 +278,19 @@ async def query(self, filters, options):
                 }})
 
         if dssearch:
-            return await self.middleware.call('dscache.query', 'USERS', filters, options)
+            dssearch_results = await self.middleware.call('dscache.query', 'USERS', filters, options.copy())
```
yocalebo, line 281: Why are we doing `options.copy()`?

src/middlewared/middlewared/plugins/account_/2fa.py
```diff
@@ -13,9 +14,11 @@ async def provisioning_uri(self, username):
...
         """
-        user = await self.middleware.call('user.query', [['username', '=', username]], {'get': True})
+        user = await self.translate_username(username)
         twofactor_config = await self.middleware.call('auth.twofactor.config')
-        user_twofactor_config = await self.middleware.call('auth.twofactor.get_user_config', user['id'])
+        user_twofactor_config = await self.middleware.call(
+            'auth.twofactor.get_user_config', user['id' if user['local'] else 'sid'], user['local'],
+        )
         if not user_twofactor_config['secret']:
             raise CallError(f'{user["username"]!r} user does not has two factor authentication configured')
```
yocalebo, line 23:
```suggestion
            raise CallError(f'{user["username"]!r} user does not have two factor authentication configured')
```

src/middlewared/middlewared/plugins/account_/2fa.py
```diff
@@ -37,10 +40,67 @@ def verify_twofactor_token(self, username, token):
...
+        user = await self.translate_username(username)
+        twofactor_auth = await self.middleware.call(
+            'auth.twofactor.get_user_config', user['id' if user['local'] else 'sid'], user['local']
+        )
+
+        # Add some sanity checks here
+        if user['local'] and not twofactor_auth['exists']:
+            raise CallError(f'Unable to locate two factor authentication configuration for {username!r} user')
+
+        secret = await self.middleware.call('auth.twofactor.generate_base32_secret')
```
yocalebo, line 88: We should only generate the `secret` if the `if two factor_auth['exists']:` branch is true

src/middlewared/middlewared/plugins/account.py
sonicaj: Because we are popping attributes from `options` which is not desirable and can affect the result here in `user.query`

src/middlewared/middlewared/plugins/account_/2fa.py
sonicaj: @yocalebo for AD users we do not add their record unless they actually configure 2fa which is why the check is for local users only as all local users are mapped to the 2fa table because we have them in the database already.
I'll add some comments to outline this behaviour, pretty sure i'll forget this myself as well after a while

undsoft:
I voiced my concerns to Waqar that it would be better if global setting was "Require 2FA" instead of "Enable/disable" and we'd allow users to enable 2FA individually for their accounts. He explained that it was not possible.
Other than that the rest of UI description is okay. We'll always allow users to configure their 2FA settings and will try to communicate that they may not have any effect until 2FA is enabled globally.

src/middlewared/middlewared/plugins/account_/2fa.py
```diff
@@ -13,11 +14,13 @@ async def provisioning_uri(self, username):
         Returns the provisioning URI for the OTP for `username`. This can then be encoded in a QR code and used
         to provision an OTP app like Google Authenticator.
         """
-        user = await self.middleware.call('user.query', [['username', '=', username]], {'get': True})
+        user = await self.translate_username(username)
```
yocalebo, line 17: This now raises a `CallError` instead of a `MatchNotFoundError`. Is this desired?

sonicaj: I think it is fine, this is a new endpoint and not being used before so there is no backwards compatibility to lookout for.

anodos325 (approved):

yocalebo (approved):

PR #11075, opened 2023-04-07
NAS-121387 / 23.10 / fix hook_license_update

yocalebo:
The `failover.licensed` endpoint reads the license file from disk, deserializes it and then checks to see if this is an HA license. Since this endpoint is called _EVERYWHERE_ we moved to caching the return of this endpoint years ago. This has exposed a bug in the `hook_license_update` method that gets called when an HA license is updated via our API.

We send the license file to the other controller, but that is not enough to cache the license into memory. We must call the appropriate hook on the remote controller as well. However, after investigation, I realized that this hook would cause a recursion loop. Now I remove that hook and explicitly setup the license on the other controller. Also, while I'm here, fix a false log message to reflect reality.


sonicaj (approved):
That is a clean fix now ;)

themylogin (approved):

PR #11092, opened 2023-04-11
NAS-121309 / 23.10 / Update forward_lookup to support A/AAAA records by default

mgrimesix:
Input parameters:
* record_type changed to a list of record_types. Default is ['A', 'AAAA']. 'SRV' and 'CNAME' remain singular and cannot be requested in combination with the others. NOTE: With the default record_types selection, hosts that support both IPv4 and IPv6 will return entries for both 'A' and 'AAAA' records.
* Added 'throw_exception_on'. Default to 'host_failure'.  Options are 'never', 'any_failure', 'host_failure', all_failure'. Output parameters and exception reporting behavior (with default selection) remain unchanged.

### DIFF
--- a/src/middlewared/middlewared/plugins/dns_client.py
+++ b/src/middlewared/middlewared/plugins/dns_client.py
@@ -57,11 +57,11 @@ class DNSClient(Service):
             List('nameservers', items=[IPAddr("ip")], default=[]),
             Int('lifetime', default=12),
             Int('timeout', default=4),
+            Str('raise_error', default='HOST_FAILURE', enum=['NEVER', 'ANY_FAILURE', 'HOST_FAILURE', 'ALL_FAILURE']),
             register=True
         ),
         Ref('query-filters'),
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/dns_client.py
```diff
@@ -57,6 +61,7 @@ async def resolve_name(self, name, rdtype, options):
         ),
         Ref('query-filters'),
         Ref('query-options'),
+        Str('throw_exception_on', default='host_failure', enum=['never', 'any_failure', 'host_failure', 'all_failure']),
```
anodos325, line 64: We typically make enums all caps, e.g. NEVER. I believe our convention is `raise_error` for cases where caller can control exception handling.

themylogin (changes requested):

themylogin (approved):

anodos325 (approved):

PR #11094, opened 2023-04-11
NAS-121443 / 23.10 / fix legacy nvdimm alert

yocalebo:
After clarification from platform team, we do not need a recommended nvdimm firmware alert for version 2.6. Instead, we need to alert as if it's an invalid firmware. I've left the recommended firmware logic in place in-case we ever have more than 1 qualified firmware for an nvdimm.

### DIFF
--- a/src/middlewared/middlewared/alert/source/mseries_nvdimm_and_bios.py
+++ b/src/middlewared/middlewared/alert/source/mseries_nvdimm_and_bios.py
@@ -3,7 +3,9 @@ import datetime
 from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, Alert, ThreadedAlertSource
 from middlewared.alert.schedule import IntervalSchedule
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/alert/source/mseries_nvdimm_and_bios.py
```diff
@@ -3,7 +3,7 @@
 from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, Alert, ThreadedAlertSource
 from middlewared.alert.schedule import IntervalSchedule
 
-WEBUI_SUPPORT_FORM = 'Please contact iXsystems Support using the form in System Settings -> General -> Support'
+WEBUI_SUPPORT_FORM = 'Please contact iXsystems Support using the form in System Settings->General->Support'
```
bmeagherix, line 6: Which form is this?  Do you mean the File Ticket button/form?  If so, maybe spell that out.

bmeagherix (approved):

PR #11105, opened 2023-04-12
NAS-120980 / 23.10 / Add MAX_EXTENT_NAME_LEN for iSCSI

bmeagherix:
Currently the underlying iSCSI target code has a limitation wrt the maximum length of the extent name (64).  Enforce a check in middleware to prevent an invalid SCST configuration.


yocalebo (approved):
Easy enough!

PR #11107, opened 2023-04-12
NAS-121442 / 23.10 / Add resource statistics for apps

Qubad786:
## Context

New Apps UI would like to display some resource statistics for apps to better tell the user how much resources an app is consuming.

I have decided to follow through on 3 types of statistics:

1. CPU usage
2. Memory usage
3. Network usage

Kubernetes provides CPU usage stats in nano cores which when is going to be translated to actual core is something like `1e9 (1,000,000,000) nanocores`. We provide an overall percentage across all system cores that an app might be consuming.

Memory usage stats are provided in bytes are forwarded as such and the same applies to network stats.

It would have been nice if we could stream these statistics as well with an event source but apparently that is not possible from kubernetes side as it is not allowing us to stream all pods statistics so if UI wants to update the stats in realtime - they would need to make repeated calls.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/resources.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/resources.py
@@ -2,6 +2,7 @@ import collections
 import errno
 import os
 
+from middlewared.plugins.system.utils import cpu_info
 from middlewared.schema import Bool, Dict, Int, List, Ref, Str, returns
 from middlewared.service import accepts, CallError, job, private, Service
 from middlewared.validators import Range
@@ -13,6 +14,14 @@ class ChartReleaseService(Service):
 
     class Config:
         namespace = 'chart.release'
+
+    SYSTEM_CORES = None
+
+    @private
+    async def get_system_cores(self):
+        if cpu_info.core_count is None:
+            await self.middleware.call('system.cpu_info')
+        return cpu_info.core_count
 
     @private
     async def retrieve_pod_with_containers(self, release_name):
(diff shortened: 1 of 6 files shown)


src/middlewared/middlewared/plugins/chart_releases_linux/resources.py
```diff
@@ -14,6 +14,14 @@ class ChartReleaseService(Service):
     class Config:
         namespace = 'chart.release'
 
+    SYSTEM_CORES = None
```
yocalebo, line 17: No reason to cache cpu core count since we're already caching it in another plugin. Please see `system.cpu_info`.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/chart_releases_linux/resources.py
Qubad786: Yes but that would mean making repeated calls i.e 20 apps and 20 redundant middleware calls to get the same attr in single `chart.release.query` call..

src/middlewared/middlewared/plugins/chart_releases_linux/resources.py
```diff
@@ -303,3 +312,28 @@ async def get_consumed_host_paths(self):
...
+                ['podRef.name', 'in', [p['metadata']['name'] for p in pods]]
+            ]
+        ):
+            stats['cpu'] += int(pod_stats['cpu']['usageNanoCores'])
+            stats['memory'] += int(pod_stats['memory']['rssBytes'])
+            for interface in pod_stats['network']['interfaces']:
+                stats['network']['incoming'] += int(interface['rxBytes'])
+                stats['network']['outgoing'] += int(interface['txBytes'])
+
+        stats['cpu'] = ((stats['cpu'] / 1000000000) / await self.get_system_cores()) * 100
```
yocalebo, line 337: I still don't understand why you're changing all of this and adding another duplicate method `get_system_cores`. Why not do:
```suggestion
        stats['cpu'] = ((stats['cpu'] / 1000000000) / (await self.middleware.call('system.cpu_info'))['core_count']) * 100
```

Qubad786: @yocalebo that is to avoid doing `self.middleware.call` calls..I can change it to use this but this would be calling middleware quite a number of times and we can avoid it easily..?

Qubad786:
Closing this PR until we investigate cobia reporting re-write..

PR #11109, opened 2023-04-13
NAS-121477 / 23.10 / fix core.debug and remove dead code

yocalebo:
In preparation for Cobia, I'm going through and removing dead code and/or paying off technical debt.

1. `ptvsd` is abandoned since 4 years ago (replaced with `debugpy`). More importantly, however, this has never worked on SCALE because we don't even bring in that package in the build. (This was never caught since we did lazy import inside the method)
2. `pydevd` was forked by us for some strange reason and we haven't touched our repo in over 3 years. I'm not sure why that was the case, probably just initially creating SCALE and creating the build system I presume. There is a proper upstream debian package that we can bring in if we ever want to add this particular debugging tool again.

Keep the `remote_pdb` engine since, very VERY rarely have I used this debugging tool. It's up for debate on whether or not it's really that useful, but I'll leave it here just in case.


yocalebo:
> FWIW, currently I also see `python3-pydevd` listed twice in [middleware/src/middlewared/debian/control](https://github.com/truenas/middleware/blob/master/src/middlewared/debian/control)

thanks, something I missed. I'll remove it there too.

themylogin (approved):

PR #11111, opened 2023-04-13
NAS-121347 / 23.10 / Remove lvm on disks used for pool creation

Qubad786:
## Problem

If a disk device has LVM configured - wipe fails for that disk as the disk is in use.
The same logic we use for removing md mirrors on disks should be used for LVM as well to remove LVM on disks in question which are going to be wiped for pool creation.

## Solution

We generate a mapping of each disk and it's associated logical volumes and disk groups. A disk can have multiple partitions and each partition can have separate logical volumes configured - this has been taken care of as well with the mapping we generate.
Once the mapping is in place we remove logical volumes on those disks and pool creation succeeds.


src/middlewared/middlewared/pytest/unit/plugins/disk/test_normalized_disk_name.py
```diff
@@ -0,0 +1,32 @@
+import contextlib
+import pytest
+
+from unittest.mock import patch, mock_open
+
+from middlewared.plugins.disk_.disk_info import DiskService
+from middlewared.pytest.unit.middleware import Middleware
+
+
+@contextlib.contextmanager
```
themylogin, line 10: `contextlib.nullcontext()`

src/middlewared/middlewared/pytest/unit/plugins/disk/test_normalized_disk_name.py
```diff
@@ -0,0 +1,32 @@
...
+    ('nvme1n1p1', '1', True, 'nvme1n1'),
+    ('vdc', '1', False, 'vdc'),
+    ('nvme1n1', '1', False, 'nvme1n1'),
+])
+def test_normalize_device_to_disk_name(partition_name, partition_no, is_partition, expected_output):
+    m = Middleware()
+    disk_service = DiskService(m)
+
+    with patch('builtins.open', mock_open(read_data=partition_no)) if is_partition else null_context():
+        with patch('middlewared.plugins.disk_.disk_info.DiskService.is_partition') as is_partition_fuc:
```
themylogin, line 28: `with patch('middlewared.plugins.disk_.disk_info.DiskService.is_partition', Mock(return_value=is_partition))`

themylogin (changes requested):
How do LVMs appear there in the first place?

Can we somehow simulate that situation to have an integration test?

igpit:
> How do LVMs appear there in the first place?

E.g. when the admin plugs in a disk with pre-existing data which he then wants to use in TN ;)

src/middlewared/middlewared/plugins/disk_/lvm.py
```diff
@@ -0,0 +1,24 @@
...
+        """
+        Remove LVM from disks.
+        """
+        lvm_to_disk_mapping = await self.middleware.call('device.list_lvm_to_disk_mapping')
+        for disk in disks:
+            if disk not in lvm_to_disk_mapping:
+                continue
+
+            for entry in lvm_to_disk_mapping[disk]:
+                await run(
```
yocalebo, line 20: I don't like forking+execing in the main event loop. Especially in a for loop.

yocalebo (changes requested):

william-gr:
Closing this PR for now, please see ticket.

PR #11129, opened 2023-04-19
NAS-121556 / 23.10 / Integrate ixdiagnose in middleware

Qubad786:
## Context

Integrated `truenas/ixdiagnose` in middleware by consuming ixdiagnose python library directly adding proper callback support to ensure job progress works as intended.

Note: `freenas-debug` contents have not been removed and once this seems stable enough we can remove that code from our codebase


yocalebo (changes requested):
Let's clean `debug_generate` up since you're here.

1. `direc` should be moved to `MIDDLEWARE_RUN_DIR`
2. `if os.path.exists(directly): shutil.rmtree(directly)` seems redundant and unnecessary. Furthermore, it's prone to a race condition so let's remove this instance. I'm assuming it's not needed since we have `clean_debug_path: True`?

yocalebo (approved):

yocalebo (approved):

PR #11136, opened 2023-04-20
NAS-121581 / 23.10 / Replication manager role

themylogin:
This PR:
- Introduces the conception of an authorized user role
- Adds `role_prefix` and `role_separate_delete` attributes to the service classes configurations to allow binding them to the roles
- Creates a CRUD roles for keychain credentials, replication tasks and ZFS snapshots and complex "Replication Manager" role that allows user to manage backup replication tasks.
- Adds `roles` field to the privilege model, allowing system administrator to grant specific roles to user privileges (and, consequently, users)


themylogin:
No authentication/privilege-related test failures were found https://ci.tn.ixsystems.net/jenkins/job/TrueNAS%20SCALE%20-%20Unstable/job/API%20Tests%20-%20TrueNAS%20SCALE%20with%20middleware%20reinstalled%20from%20a%20branch%20Incremental/814/#showFailuresLink

src/middlewared/middlewared/role.py
```diff
@@ -0,0 +1,62 @@
...
+        for role in roles:
+            if role not in self.roles:
+                raise ValueError(f"Invalid role {role!r}")
+
+        self.methods[method_name] = roles
+
+        for role in roles:
+            self.allowlists_for_roles[role].append({"method": "CALL", "resource": method_name})
+
+    def roles_for_fole(self, role):
```
yocalebo, line 52: `fole` has negative connotation depending on the person's native tongue. I'm  not sure if it has another meaning but nonetheless, it's not a valid method name.

yocalebo (changes requested):
We need a comment describing what this does since it's a significant feature. We then need to copy whatever is written here and put it in an internal document (or NEP).

We also actually need documentation on how other developers can add these roles to other endpoints to get the desired effect. (again, putting it all in a document)

yocalebo:
> @yocalebo I added a middleware developer doc

Developer documentation is good but please also add a summary of what this does to the actual GitHub PR. This is also a new ticket and not associated to the ticket that I created which lists vastly more detail on why we are doing this and the general idea.

sonicaj (approved):

src/middlewared/middlewared/plugins/account_/privilege.py
```diff
@@ -334,10 +340,16 @@ async def privileges_for_groups(self, groups_key, group_ids):
     @private
     async def compose_privilege(self, privileges):
         compose = {
+            'roles': set(),
             'allowlist': [],
             'web_shell': False,
         }
         for privilege in privileges:
+            for role in privilege['roles']:
+                compose['roles'] |= self.middleware.role_manager.roles_for_fole(role)
```
yocalebo, line 349: `roles_for_fole` is back
```suggestion
                compose['roles'] |= self.middleware.role_manager.roles_for_fole(role)
```

yocalebo (changes requested):

anodos325 (approved):

themylogin:
time 30:00

PR #11138, opened 2023-04-20
NAS-121594 / 23.10 / fix unhandled exception in vrrp_backup

yocalebo:
Again, investigating an unrelated issue and noticed this in the logs on the system I'm investigating. Don't crash right here at the end. This failure scenario is present when an HA system is installed and 1 controller is powered off/removed from the head-unit while the end-user is configuring the A controller for the first time. Initial failover event that is received is BACKUP event, subsequently followed by a MASTER event.

### DIFF
--- a/src/middlewared/middlewared/plugins/failover_/event.py
+++ b/src/middlewared/middlewared/plugins/failover_/event.py
@@ -733,7 +733,12 @@ class FailoverEventsService(Service):
         # do something with iscsi service here
 
         logger.info('Syncing encryption keys from MASTER node (if any)')
-        self.run_call('failover.call_remote', 'failover.sync_keys_to_remote_node')
+        try:
+            self.run_call('failover.call_remote', 'failover.sync_keys_to_remote_node')
+        except Exception as e:
+            ignore = (errno.ECONNRESET, errno.ECONNREFUSED, errno.ECONNABORTED, errno.EHOSTDOWN)
+            if isinstance(e, CallError) and e.errno not in ignore:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/failover_/event.py
```diff
@@ -733,7 +733,12 @@ def vrrp_backup(self, job, fobj, ifname, event):
         # do something with iscsi service here
 
         logger.info('Syncing encryption keys from MASTER node (if any)')
-        self.run_call('failover.call_remote', 'failover.sync_keys_to_remote_node')
+        try:
+            self.run_call('failover.call_remote', 'failover.sync_keys_to_remote_node')
+        except Exception as e:
+            ignore = (errno.ECONNRESET, errno.ECONNREFUSED, errno.ECONNABORTED, errno.EHOSTDOWN)
+            if isinstance(e, CallError) and e.errno not in ignore:
```
themylogin, line 740: Do we silently ignore everything that is not `CallError`?

themylogin (commented):
We have a lot of similar errno lists for failover remote call error handing. Should we check for these errors in `failover.call_remote` itself and re-raise a specific type of exception that could be handled easier?

themylogin:
@yocalebo what about the other concern (failover remote host down checks duplicated all over the codebase)? Separate PR?

yocalebo:
> @yocalebo what about the other concern (failover remote host down checks duplicated all over the codebase)? Separate PR?

Yeah, separate PR

themylogin (approved):

PR #11143, opened 2023-04-21
NAS-121621 / 23.10 / remove onedrivesdk (deprecated a long time ago)

yocalebo:
This module hasn't seen a commit since 2016 so it's dead and it does not work with python3.11 (ships in Cobia). There doesn't seem to be a decent alternative either so we'll drop this for now and revisit if we find another solution.


bmeagherix (approved):

imkiva:
Hi all. I just **made onedrivesdk compatible with Python 3.11**. You can find the repo with my fixes here: https://github.com/imkiva/onedrive-sdk-python and a corresponding Pull Request has been made to the upstream: https://github.com/OneDrive/onedrive-sdk-python/pull/206

I have tested it locally and found it still works just like before. Is it possible to add OneDrive CloudSync back?

yocalebo:
> Hi all. I just **made onedrivesdk compatible with Python 3.11**. You can find the repo with my fixes here: https://github.com/imkiva/onedrive-sdk-python and a corresponding Pull Request has been made to the upstream: https://github.com/OneDrive/onedrive-sdk-python/pull/206
> 
> 
> 
> I have tested it locally and found it still works just like before. Is it possible to add OneDrive CloudSync back?

Hi @imkiva, great job on fixing this! However, we've got no plans on bringing this back. Unless your PR is merged upstream and a new version is published on pypi, I don't see this provider being added again. The last PR that was merged upstream was 7 years ago. I don't expect yours to be merged, unfortunately. Furthermore, this adds support for 3.11 but 3.12 is already released and we will try and keep up to date with the Python versions. Even if we were to bring this back, I suspect new versions of Python will render this library unusable and so we'll be forced, yet again, to remove it.

PR #11144, opened 2023-04-21
NAS-121633 / 23.10 / don't generate ctdb unless glusterd is enabled

yocalebo:
ctdb api endpoints makes the assumption that the glusterd service is started. ctdb.conf needs to be gated on whether or not glusterd is enabled.


mgrimesix (approved):
Looks good to me.

yocalebo:
backport

PR #11155, opened 2023-04-26
NAS-121677 / 23.10 / fix async_validators.py:resolve_hostnames()

yocalebo:
Python 3.10 removed the loop kwarg to asyncio.wait_for so by upgrading to python3.11, this raises an error. To my pleasant surprise, this is the only place that I could find where we passed a loop kwarg to asyncio.wait_for.

### DIFF
--- a/src/middlewared/middlewared/async_validators.py
+++ b/src/middlewared/middlewared/async_validators.py
@@ -87,9 +87,8 @@ async def resolve_hostname(middleware, verrors, name, hostname):
             return False
 
     try:
-        result = await asyncio.wait_for(middleware.create_task(
-            middleware.run_in_thread(resolve_host_name_thread, hostname)
-        ), 5, loop=asyncio.get_event_loop())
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/async_validators.py
```diff
@@ -87,9 +87,8 @@ def resolve_host_name_thread(hostname):
             return False
 
     try:
-        result = await asyncio.wait_for(middleware.create_task(
-            middleware.run_in_thread(resolve_host_name_thread, hostname)
-        ), 5, loop=asyncio.get_event_loop())
+        aw = middleware.create_task(middleware.run_in_thread(resolve_host_name_thread, hostname)),
```
bmeagherix, line 90: Trailing comma?

bmeagherix (approved):

src/middlewared/middlewared/async_validators.py
yocalebo: Good catch, thanks!

PR #11156, opened 2023-04-26
NAS-121558 / 23.10 / Fix kubernetes logs assertion error

Qubad786:
## Problem

`qbittorrent` application has been changed as to how it functions and it now has init container which exits and we don't have logs to follow/assert.

## Solution

Use a different application which solves this problem and has logs streaming which we can test/assert.


Qubad786:
backport

PR #11160, opened 2023-04-26
NAS-121693 / 23.10 / remove dead python3-prctl dependency

yocalebo:
Yet again, we're using a module that is crashing after bookworm transition. Luckily we're only using this module for setting process names and setting pthread_signals. This is easily achieved with native python using ctypes module. Below is a crash that is being seen.

```
subprocess.CalledProcessError: Command 'Failed to execute truenas-initrd: Failed to update initramfs
  File "/tmp/tmplil82rux/usr/local/bin/truenas-initrd.py", line 181, in <module>
    from middlewared.utils.db import FREENAS_DATABASE, query_config_table, query_table
  File "/usr/local/lib/python3.9/dist-packages/middlewared/utils/__init__.py", line 12, in <module>
    from middlewared.utils.threading import start_daemon_thread  # noqa
  File "/usr/local/lib/python3.9/dist-packages/middlewared/utils/threading.py", line 8, in <module>
    import prctl
  File "/tmp/tmplil82rux/usr/lib/python3/dist-packages/prctl.py", line 58, in <module>
    for i in range(_prctl.CAP_LAST_CAP+1):
AttributeError: module '_prctl' has no attribute 'CAP_LAST_CAP'
```


src/middlewared/middlewared/utils/prctl.py
```diff
@@ -0,0 +1,37 @@
...
+    # from linux/prctl.h
+    SET_PDEATHSIG = 1
+    SET_NAME = 15
+
+
+@contextlib.contextmanager
+def load_libc():
+    libc = None
+    try:
+        libc = ctypes.CDLL(ctypes.util.find_library('c'))
```
themylogin, line 19: Is this thread-safe?

themylogin (commented):
The root cause of the truenas-initrd issue is
```/tmp/tmplil82rux/usr/lib/python3/dist-packages```
Is it possible to run it from a chroot instead of mixing two python versions? @sonicaj

src/middlewared/middlewared/utils/prctl.py
yocalebo: Should be, documentation states that we only need to worry about tracked references when we pass callback functions (which we're not)

themylogin (approved):

PR #11170, opened 2023-04-27
NAS-121727 / 23.10 / Skip iSCSI CHAP tests for invalid initiatorname values

bmeagherix:
Currently `python-scsi` presents an **illegal** initiator name when CHAP is in use.  In bookworm `SCST` responds very badly to these illegal values, so skip these tests for now.

FWIW, plan to address two related aspects separately (and in this order):

1. Modify test-runner environment to use a more modern `python-scsi` that does not present an illegal initiator name
2. Robustize `SCST` to reject illegal initiator names.


mgrimesix (approved):
Looks good to me.

PR #11176, opened 2023-04-28
NAS-121740 / 23.10 / Add changes to allow renaming catalogs for installed apps

sonicaj:
This PR adds changes to allow changing catalog name for existing apps. So for example if we want to rename the official catalog, we need to make that happen in 2 places i.e available apps and installed apps, for available ones we would be updating the database and for installed we would be writing a migration which would make this work.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/app_migrations.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/app_migrations.py
@@ -9,6 +9,7 @@ from catalog_validation.schema.migration_schema import APP_MIGRATION_SCHEMA
 from middlewared.plugins.catalogs_linux.update import OFFICIAL_LABEL
 from middlewared.plugins.chart_releases_linux.utils import get_namespace
 from middlewared.service import Service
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/kubernetes_linux/app_migrations.py
```diff
@@ -143,7 +165,21 @@ def load_migrations(self, catalog):
 
             return migrations
         else:
-            return {}
+            return migrations
+
+    def official_migrations(self):
+        migrations = {}
```
yocalebo, line 171: This can be simplified and made more performant (there is no reason for sorting)
```suggestion
        migrations = {}
        app_migration_dir = f'{get_middleware_dir()}/plugins/kubernetes_linux/app_migrations'
        for migration in filter(lambda x: x.suffix == '.json', Path(app_migration_dir).iterdir():
            migrations[migration.name] = json.loads(migration.read_text())
        return migrations
```

yocalebo (changes requested):

yocalebo (approved):

PR #11190, opened 2023-05-01
NAS-121764 / 22.12.3 / add ipmi.sel plugin

yocalebo:
This adds an `ipmi.sel` specific plugin for directly managing the ipmi SEL portion of systems. The justification for doing this is to not crowd the main `ipmi` plugin with other ipmi sub-tasks that don't really make sense. Keeping each ipmi related "sub-task" in their own plugin makes it easier to maintain and extend.

### DIFF
--- a/src/middlewared/middlewared/plugins/ipmi_/sel.py
+++ b/src/middlewared/middlewared/plugins/ipmi_/sel.py
@@ -38,14 +38,8 @@ class IpmiSelService(Service):
         """Query IPMI System Event Log (SEL) extended list"""
         rv = []
         job.set_progress(78, 'Enumerating extended event log info')
-        output = get_sel_data('elist')
-        len_output = len(output) or 1
-        increment = round(79 / len_output, 3)  # remaining ~19% of progress
-        progress_percent = 79
-        for line in output:
+        for line in get_sel_data('elist'):
             if (values := line.strip().split(',')) and len(values) == 6:
-                progress_percent += increment
-                job.set_progress(progress_percent, 'Parsing extended event log info')
                 rv.append({
                     'id': values[0].strip(),
                     'date': values[1].strip(),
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/ipmi_/sel.py
```diff
@@ -0,0 +1,88 @@
...
+        rv = {}
+        job.set_progress(78, 'Enumerating general extended event log info')
+        output = get_sel_data('info')
+        len_output = len(output) or 1
+        increment = round(79 / len_output, 3)  # remaining ~19% of progress
+        progress_percent = 79
+        for line in output:
+            if (values := line.strip().split(':')) and len(values) == 2:
+                progress_percent += increment
+                job.set_progress(progress_percent, 'Parsing general extended event log info')
```
themylogin, line 75: Why is progress reporting necessary here, what is the data size? How much time does the entire method take and how much time passes between `set_progress` calls?

themylogin: I don't see any data streaming performed here, we wait for IPMI command to execute (which will look stalled) and then we parse its output (which will take microseconds, but yield a number of progress messages)

src/middlewared/middlewared/plugins/ipmi_/sel.py
```diff
@@ -0,0 +1,82 @@
...
+        cli_namespace = 'service.ipmi.sel'
+
+    @filterable
+    @filterable_returns(Dict('ipmi_elist', additional_attrs=True))
+    @job(lock=SEL_LOCK, lock_queue_size=1)
+    def elist(self, job, filters, options):
+        """Query IPMI System Event Log (SEL) extended list"""
+        rv = []
+        job.set_progress(78, 'Enumerating extended event log info')
+        output = get_sel_data('elist')
```
themylogin, line 41: Same here

themylogin (changes requested):

PR #11201, opened 2023-05-02
NAS-121781 / 23.10 / add -v and --no-header-output to ipmi.sel.elist

yocalebo:
This adds the `-v` command flag so we can get the event direction for each sel elist entry as well as added the `--no-header-output` flag so the column title line would not be returned in the results.


yocalebo:
retest this please

Qubad786 (approved):

PR #11206, opened 2023-05-02
NAS-121509 / 23.10 / Fix application pool name bug in kubernetes setup

Qubad786:
## Problem
There was a problem with how we removed the pool name from k3s datasets. Essentially we were relying on some custom logic which broke when pool was named `applications` and hence the resulting dataset was malformed i.e `ix-/kubectl` 

## Solution
To address this issue, we remove the complicated logic we had in place and just strip away the pool name cleanly.


sonicaj (approved):

Qubad786:
backport

PR #11217, opened 2023-05-03
NAS-121690 / 23.10 / Update NFS configuration for nfs-utils v2.6

mgrimesix:
nfs-utils has moved from v1.3 to v2.6.   One significant change is
is how the services are started and the management of the config files.

* Configuration parameters for the services are no longer passed in on the command line.
* nfs-common and nfs-kernel-server config files have been combined into a single 'ini' formatted file: /etc/nfs.utils.d/local.conf


yocalebo (approved):
Great job! This looks good but let's wait for @anodos325 final review before merge, please.

anodos325:
Looks like the test run failed test_036 (kerberos + NFS4). There may be additional configuration required to make that work, but this will be enough to at least let people start testing NFS in nightlies.

anodos325 (approved):

PR #11220, opened 2023-05-03
NAS-121696 / 23.10 / Properly cover the case where parent of VM disk device is encrypted

Qubad786:
## Problem
We have encountered a use case where the parent dataset is locked, but its child dataset is being used as a disk by a virtual machine (VM). This scenario has not been addressed before. The dataset logic checks for a mount path like `/mnt/tank/enc` in the pool, but the path on the VM is in the form of a zvol, such as `/zvol/tank/enc/vm-vol`.

## Solution
To resolve this issue, we need to verify whether the zvol disk is a child of the unlocked dataset or not. If it is a child, then we should restart the VM.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/unlock.py
+++ b/src/middlewared/middlewared/plugins/pool_/unlock.py
@@ -58,14 +58,15 @@ class PoolDatasetService(Service):
                 if not path:
                     continue
 
-                if (
-                    (
-                        dataset['type'] == 'FILESYSTEM' and
-                        (mountpoint := dataset_mountpoint(dataset)) and
-                        path.startswith(mountpoint + '/')
-                    ) or
-                    (dataset['type'] == 'VOLUME' and zvol_name_to_path(dataset['name']) == path)
-                ):
+                unlock = False
+                if dataset['type'] == 'FILESYSTEM' and (mountpoint := dataset_mountpoint(dataset)):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/pool_/unlock.py
```diff
@@ -61,8 +61,8 @@ async def unlock_restarted_vms(self, dataset):
                 if (
                     (
                         dataset['type'] == 'FILESYSTEM' and
-                        (mountpoint := dataset_mountpoint(dataset)) and
-                        path.startswith(mountpoint + '/')
+                        ((mountpoint := dataset_mountpoint(dataset)) and path.startswith(mountpoint + '/')) or
+                        (path.startswith(zvol_name_to_path(dataset['name']) + '/'))
```
anodos325, line 65: This conditional isn't particularly readable. I feel like it should probably be broken apart into separate checks to make it easier to understand and maintain.

Shouldn't in this case be converting the zvol path to a dataset name and comparing the string with the dataset name being unlocked.

`/dev/zvol/tank/encrypted/my_zvol` -> `tank/encrypted/my_zvol` then compare that with what's my unlocked `tank/encrypted`

yocalebo (approved):

yocalebo (approved):

PR #11222, opened 2023-05-03
NAS-121806 / 23.10 / Install relevant dependencies before running unit tests in CI

Qubad786:
## Context

We should be installing any relevant dependencies required for the testing environment to work before executing unit tests in the CI.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -29,6 +29,7 @@ jobs:
 
       - name: Deploy
         run: |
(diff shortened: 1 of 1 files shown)


yocalebo (approved):

yocalebo (changes requested):
I think this is being addressed properly here: https://github.com/truenas/middleware/pull/11223

yocalebo (changes requested):
Please update title of PR and the body of the commit message.

yocalebo (approved):

PR #11229, opened 2023-05-03
NAS-121813 / 23.10 / simplify and speed up gather_network() in usage.py

yocalebo:
- inner functions incur a non-trivial amount of overhead and there is almost 0 reason to use them
- simplify this for loop dramatically so we only iterate network interfaces once (instead of 4 times)


mgrimesix (approved):
Looks good to me.

PR #11231, opened 2023-05-03
NAS-122815 / 23.10 / Add 61-openmediavault-dev-disk-by-id.rules and serial_id from openmed…

MichelBr:
Hello,

I enjoy Truenas and I know you don't like usb :), but from a practical viewpoint the usb-bus is just that: a bus. usb 3 supports 5 Gbit and 10 Gbit and usb-c even more and they can be highly performant. I also have a multi-bay enclosure with automatic cooling and it also has esata, but my simple nas (a nice good nuc, only support usb-3, thunderbolt, ...). I read the discussions on openmediavault and even more expensive enclosures use a chipset that doesn't pass through the drives serial id, but represent them as "serial id and a counter" in udev as the serial id. 

The short serial id, as used internally by truenas as a drive id, then sees multiples drives, all with the same serial id and only shows the last one. The linux-system of course sees all the drives, but truenas not as it assumes that the short serial id should be unique, as many other projects seem to do. Anyway, not in this case.

Openmediavault fixed this by using a custom udev-program "serial_id" that uses hdparm to retrieve the drives real id and present that as the serial id for these usb-to-sata-chipsets that only present one serial-id themselves by default in udev. It calls the serial_id on the drives, if one of these chipsets is detected. 
The drives then report a unique serial id and truenas can work with them without further error.

So, I created this patch with integrates these files from omv into truenas. The licenses (GPL-3, BSD-3) seem compatible. If you wish, I can ask if they may be copied over/shared between the projects. I don't mind the usb-warning as I checked the bay uses at least the 10 Gbit usb bus using lshw, so performance should be a non-issue (which seems to be the case up till now) for this hdd-drives.

So, please merge :). Thank you.

…iavault project to support multibay-usb-enclosures that use a chipset that doesn't passthrough the drives serial-id, but only report 1 serial-id.

Reference: https://github.com/openmediavault/openmediavault/blob/master/deb/openmediavault/etc/udev/rules.d/61-openmediavault-dev-disk-by-id.rules - GPL 3 license seems to be used. Based on my reading, this is compatible with the bsd-3 licencse for middleware, but I can ask if we may copy it.


MichelBr:
Is there something I can do to advance this?

Kind regards,
Michel

yocalebo:
> Thanks for sharing this @MichelBr!
> 
> It seems that this workaround for known USB enclosures via Udev rules works like a charm!
> 
> In my tests using with an Orico enclosure (VL717 USB 3.1 Gen2 / 10Gbps), I had to swap lines 32 and 33 below on `get_disk_serial()` to give precedence to `ID_SERIAL_SHORT` over `ID_SCSI_SERIAL`, otherwise the disks would still not be recognized because `ID_SCSI_SERIAL` was still the same:
> 
> https://github.com/truenas/middleware/blob/b5571f0462ef009060cb56ab0c445c0355e5fab4/src/middlewared/middlewared/plugins/device_/device_info.py#L30-L34
> 
> I hope this get merged soon, as these udev rules are specific to a few USB devices and there are no side-effects for other users (not using these USB chipsets).
> 
> There are opportunities to simply rewrite/simplify these rules and also make a `serial_id` program based on `smartctl` instead of `hdparm`, which I believe has more compatibility with Seagate HDDs (and supports FARM).

We appreciate the PR, we're definitely going to review it! However, we do not use hdparm to retrieve serial information and `smartctl` will provide the exact same serial information in this scenario. Lastly, sub-processing (fork+exec)'ing for every hard-drive that we detect is very expensive and only hurts the performance of other users that have many 10's or even 100's (and for our enterprise users 1000's of hard drives) on a singular system. However, this script opens up some possibilities so we're definitely appreciative of it. We're hoping to get to this soon, will keep this thread updated when we get to it. Thanks again!

yocalebo:
Why does this need to run hdparm to get the serial number? The kernel already has this information so can this udev rule just simply append the ID_SERIAL_SHORT with the other unique identifiers?

MichelBr:
Because the USB enclosure is not reporting them as expected. It is
reporting them with an integer appended: 0, 1,... I expect the position in
the driveway, which isn't what one wants. I can check udev again, but I do
not remember a unique serial in there. I started this from the I'm project
as I noticed my bay chipset I'd in there, next to some other ones they
encountered and fixed. But it may certainly be improved upon.

Michel

Op vr 2 jun. 2023 15:56 schreef Caleb St. John ***@***.***>:

yocalebo:
> Because the USB enclosure is not reporting them as expected. It is reporting them with an integer appended: 0, 1,... I expect the position in the driveway, which isn't what one wants. I can check udev again, but I do not remember a unique serial in there. I started this from the I'm project as I noticed my bay chipset I'd in there, next to some other ones they encountered and fixed. But it may certainly be improved upon. Michel Op vr 2 jun. 2023 15:56 schreef Caleb St. John ***@***.***>:
> […](#)

So if the USB hub is reporting a serial with an integer appended (0, 1, 2) that makes the serial numbers unique? If the attached disks have unique serials, then we shouldn't need these udev rules? I was under the impression that the udev rules are for when a USB hub reports all disks as having the _same_ serial id?

MichelBr:
True as uses the short I'd which is the same for every disk causing the
issue. The long I'd has the integer. I wouldn't want this id as switching
the disk to another bay may cause issues. Hdparm I'd seems better

Op vr 2 jun. 2023 16:19 schreef Caleb St. John ***@***.***>:

MichelBr:
Sorry for that last response as I typed it from my cellphone.

Ok,


   -  Truenas uses the short serial id as the unique id for a drive. This
   id is the same for all drives of the usb enclosure. Using the long serial
   id adds the unique integer
   - I wouldn't use an ordering or usb-hub specific behaviour to identify a
   disk. Wouldn't that create problems if the enclosure or chips dies, you
   replace the enclosure/chip and suddenly your disks have a different id?
   Querying the correct id uses hdparm at least prevents this issue. Not
   saying it can be improved upon

Personally, I would use a software defined disk uuid that you can assign on
the first time you find a disk, similar to zfs and other mounting behaviour
can be done. So, it doesn't depend on a specific hardware location on your
system, but these scripts are a start to atleast decently support certain
external drive-enclosures over usb-busses. Preferably usb-c, thunderbolt
of course, but usb 3.x 5 Gbps should be a good start for normal harddrives.

Kind regards,
Michel

MichelBr:
I looked up a drive behind the bad usb controller. However, my truenas system has this change temporarily installed. So, it reports itself in udev as:

**drive 1 behind usb controller with fix applied:**

P: /devices/pci0000:00/0000:00:14.0/usb4/4-3/4-3:1.0/host3/target3:0:0/3:0:0:0/block/sde
N: sde
L: 0
S: disk/by-id/usb-TOSHIBA_HDWG180_80C0A08LFAUG-0:0
S: disk/by-path/pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:0
E: DEVPATH=/devices/pci0000:00/0000:00:14.0/usb4/4-3/4-3:1.0/host3/target3:0:0/3:0:0:0/block/sde
E: DEVNAME=/dev/sde
E: DEVTYPE=disk
E: DISKSEQ=6
E: MAJOR=8
E: MINOR=64
E: SUBSYSTEM=block
E: USEC_INITIALIZED=6724000
E: ID_VENDOR=TOSHIBA
E: ID_VENDOR_ENC=TOSHIBA\x20
E: ID_VENDOR_ID=152d
E: ID_MODEL=HDWG180
E: ID_MODEL_ENC=HDWG180\x20\x20\x20\x20\x20\x20\x20\x20\x20
E: ID_MODEL_ID=0567
E: ID_REVISION=0125
E: ID_SERIAL=TOSHIBA_HDWG180_80C0A08LFAUG-0:0
E: ID_SERIAL_SHORT=80C0A08LFAUG
E: ID_TYPE=disk
E: ID_INSTANCE=0:0
E: ID_BUS=usb
E: ID_USB_INTERFACES=:080650:
E: ID_USB_INTERFACE_NUM=00
E: ID_USB_DRIVER=usb-storage
E: ID_PATH=pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:0
E: ID_PATH_TAG=pci-0000_00_14_0-usb-0_3_1_0-scsi-0_0_0_0
E: ID_PART_TABLE_UUID=854d3aa8-f17c-48ab-bfe5-388089d64f7f
E: ID_PART_TABLE_TYPE=gpt
E: DEVLINKS=/dev/disk/by-id/usb-TOSHIBA_HDWG180_80C0A08LFAUG-0:0 /dev/disk/by-path/pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:0
E: TAGS=:systemd:
E: CURRENT_TAGS=:systemd:
---

**So, truenas uses the following info:

E: ID_SERIAL_SHORT=80C0A08LFAUG**

I have another drive installed in this enclosure, which, with the fix applied, reports itself in udev as (udevadm info /dev/sdf):

P: /devices/pci0000:00/0000:00:14.0/usb4/4-3/4-3:1.0/host3/target3:0:0/3:0:0:1/block/sdf
N: sdf
L: 0
S: disk/by-path/pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:1
S: disk/by-id/usb-TOSHIBA_HDWG180_80C0A08QFAUG-0:1
E: DEVPATH=/devices/pci0000:00/0000:00:14.0/usb4/4-3/4-3:1.0/host3/target3:0:0/3:0:0:1/block/sdf
E: DEVNAME=/dev/sdf
E: DEVTYPE=disk
E: DISKSEQ=7
E: MAJOR=8
E: MINOR=80
E: SUBSYSTEM=block
E: USEC_INITIALIZED=6618462
E: ID_VENDOR=TOSHIBA
E: ID_VENDOR_ENC=TOSHIBA\x20
E: ID_VENDOR_ID=152d
E: ID_MODEL=HDWG180
E: ID_MODEL_ENC=HDWG180\x20\x20\x20\x20\x20\x20\x20\x20\x20
E: ID_MODEL_ID=0567
E: ID_REVISION=0125
**E: ID_SERIAL=TOSHIBA_HDWG180_80C0A08QFAUG-0:1
E: ID_SERIAL_SHORT=80C0A08QFAUG**
E: ID_TYPE=disk
E: ID_INSTANCE=0:1
E: ID_BUS=usb
E: ID_USB_INTERFACES=:080650:
E: ID_USB_INTERFACE_NUM=00
E: ID_USB_DRIVER=usb-storage
E: ID_PATH=pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:1
E: ID_PATH_TAG=pci-0000_00_14_0-usb-0_3_1_0-scsi-0_0_0_1
E: ID_PART_TABLE_UUID=2d98f80a-9757-427a-afb3-8cc3727a52a8
E: ID_PART_TABLE_TYPE=gpt
E: DEVLINKS=/dev/disk/by-path/pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:1 /dev/disk/by-id/usb-TOSHIBA_HDWG180_80C0A08QFAUG-0:1
E: TAGS=:systemd:
E: CURRENT_TAGS=:systemd:

So, truenas uses

**E: ID_SERIAL_SHORT=80C0A08QFAUG**

Without this fix,

the second drive would have the same serial id as the first drive and I can't be sure without bringing my system into an error-state, but that reported id didn't come from any drive I think. It's the same regardless of the drive behind it. It doesn't matter anyway. 
**The problem is that without this fix, each drive reports the same ID_SERIAL_SHORT, which confuses truenas and it only reports one drive in the enclosure.**

You can also see that the value for the key-value-pair "ID_SERIAL", is the same as ID_SERIAL_SHORT, but then with an integer added to it. This would indeed solve the uniqueness-issue, but as I said I wouldn't do it, as the purpose of the id is to identify the disk and changing enclosures and probably even changing the order of the disks in the enclosures would make truenas

- loose track of the disk when the enclosure changes (and thus the logic)
- mix up statistics (if tracked) for 2 different disks, which actually just got swapped (let's say during a maintenance procedure).

This is all solved using this commit.

The commit could be improved upon later on by:

- automatically detecting this scenario and reporting the bad usb-chipset/bridge
- using the gpt table's ptuuid as there is no need for a universal unique id worldwide. Nice if it exists of course. Truenas only needs a unique id for each disk attached to its own system and the ptuuid can also be assigned manually, so you could try to initialize it from the reported disk id, ... and make sure it's always unique. In this case, we do not have to rely on external parties and the uniqueness constraint is fullfilled by the software.

Kind regards,

Michel

yocalebo:
> I looked up a drive behind the bad usb controller. However, my truenas system has this change temporarily installed. So, it reports itself in udev as:
> 
> 
> 
> **drive 1 behind usb controller with fix applied:**
> 
> 
> 
> P: /devices/pci0000:00/0000:00:14.0/usb4/4-3/4-3:1.0/host3/target3:0:0/3:0:0:0/block/sde
> 
> N: sde
> 
> L: 0
> 
> S: disk/by-id/usb-TOSHIBA_HDWG180_80C0A08LFAUG-0:0
> 
> S: disk/by-path/pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:0
> 
> E: DEVPATH=/devices/pci0000:00/0000:00:14.0/usb4/4-3/4-3:1.0/host3/target3:0:0/3:0:0:0/block/sde
> 
> E: DEVNAME=/dev/sde
> 
> E: DEVTYPE=disk
> 
> E: DISKSEQ=6
> 
> E: MAJOR=8
> 
> E: MINOR=64
> 
> E: SUBSYSTEM=block
> 
> E: USEC_INITIALIZED=6724000
> 
> E: ID_VENDOR=TOSHIBA
> 
> E: ID_VENDOR_ENC=TOSHIBA\x20
> 
> E: ID_VENDOR_ID=152d
> 
> E: ID_MODEL=HDWG180
> 
> E: ID_MODEL_ENC=HDWG180\x20\x20\x20\x20\x20\x20\x20\x20\x20
> 
> E: ID_MODEL_ID=0567
> 
> E: ID_REVISION=0125
> 
> E: ID_SERIAL=TOSHIBA_HDWG180_80C0A08LFAUG-0:0
> 
> E: ID_SERIAL_SHORT=80C0A08LFAUG
> 
> E: ID_TYPE=disk
> 
> E: ID_INSTANCE=0:0
> 
> E: ID_BUS=usb
> 
> E: ID_USB_INTERFACES=:080650:
> 
> E: ID_USB_INTERFACE_NUM=00
> 
> E: ID_USB_DRIVER=usb-storage
> 
> E: ID_PATH=pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:0
> 
> E: ID_PATH_TAG=pci-0000_00_14_0-usb-0_3_1_0-scsi-0_0_0_0
> 
> E: ID_PART_TABLE_UUID=854d3aa8-f17c-48ab-bfe5-388089d64f7f
> 
> E: ID_PART_TABLE_TYPE=gpt
> 
> E: DEVLINKS=/dev/disk/by-id/usb-TOSHIBA_HDWG180_80C0A08LFAUG-0:0 /dev/disk/by-path/pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:0
> 
> E: TAGS=:systemd:
> 
> E: CURRENT_TAGS=:systemd:
> 
> ---
> 
> 
> 
> **So, truenas uses the following info:
> 
> 
> 
> E: ID_SERIAL_SHORT=80C0A08LFAUG**
> 
> 
> 
> I have another drive installed in this enclosure, which, with the fix applied, reports itself in udev as (udevadm info /dev/sdf):
> 
> 
> 
> P: /devices/pci0000:00/0000:00:14.0/usb4/4-3/4-3:1.0/host3/target3:0:0/3:0:0:1/block/sdf
> 
> N: sdf
> 
> L: 0
> 
> S: disk/by-path/pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:1
> 
> S: disk/by-id/usb-TOSHIBA_HDWG180_80C0A08QFAUG-0:1
> 
> E: DEVPATH=/devices/pci0000:00/0000:00:14.0/usb4/4-3/4-3:1.0/host3/target3:0:0/3:0:0:1/block/sdf
> 
> E: DEVNAME=/dev/sdf
> 
> E: DEVTYPE=disk
> 
> E: DISKSEQ=7
> 
> E: MAJOR=8
> 
> E: MINOR=80
> 
> E: SUBSYSTEM=block
> 
> E: USEC_INITIALIZED=6618462
> 
> E: ID_VENDOR=TOSHIBA
> 
> E: ID_VENDOR_ENC=TOSHIBA\x20
> 
> E: ID_VENDOR_ID=152d
> 
> E: ID_MODEL=HDWG180
> 
> E: ID_MODEL_ENC=HDWG180\x20\x20\x20\x20\x20\x20\x20\x20\x20
> 
> E: ID_MODEL_ID=0567
> 
> E: ID_REVISION=0125
> 
> **E: ID_SERIAL=TOSHIBA_HDWG180_80C0A08QFAUG-0:1
> 
> E: ID_SERIAL_SHORT=80C0A08QFAUG**
> 
> E: ID_TYPE=disk
> 
> E: ID_INSTANCE=0:1
> 
> E: ID_BUS=usb
> 
> E: ID_USB_INTERFACES=:080650:
> 
> E: ID_USB_INTERFACE_NUM=00
> 
> E: ID_USB_DRIVER=usb-storage
> 
> E: ID_PATH=pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:1
> 
> E: ID_PATH_TAG=pci-0000_00_14_0-usb-0_3_1_0-scsi-0_0_0_1
> 
> E: ID_PART_TABLE_UUID=2d98f80a-9757-427a-afb3-8cc3727a52a8
> 
> E: ID_PART_TABLE_TYPE=gpt
> 
> E: DEVLINKS=/dev/disk/by-path/pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:1 /dev/disk/by-id/usb-TOSHIBA_HDWG180_80C0A08QFAUG-0:1
> 
> E: TAGS=:systemd:
> 
> E: CURRENT_TAGS=:systemd:
> 
> 
> 
> So, truenas uses
> 
> 
> 
> **E: ID_SERIAL_SHORT=80C0A08QFAUG**
> 
> 
> 
> Without this fix,
> 
> 
> 
> the second drive would have the same serial id as the first drive and I can't be sure without bringing my system into an error-state, but that reported id didn't come from any drive I think. It's the same regardless of the drive behind it. It doesn't matter anyway. 
> 
> **The problem is that without this fix, each drive reports the same ID_SERIAL_SHORT, which confuses truenas and it only reports one drive in the enclosure.**
> 
> 
> 
> You can also see that the value for the key-value-pair "ID_SERIAL", is the same as ID_SERIAL_SHORT, but then with an integer added to it. This would indeed solve the uniqueness-issue, but as I said I wouldn't do it, as the purpose of the id is to identify the disk and changing enclosures and probably even changing the order of the disks in the enclosures would make truenas
> 
> 
> 
> - loose track of the disk when the enclosure changes (and thus the logic)
> 
> - mix up statistics (if tracked) for 2 different disks, which actually just got swapped (let's say during a maintenance procedure).
> 
> 
> 
> This is all solved using this commit.
> 
> 
> 
> The commit could be improved upon later on by:
> 
> 
> 
> - automatically detecting this scenario and reporting the bad usb-chipset/bridge
> 
> - using the gpt table's ptuuid as there is no need for a universal unique id worldwide. Nice if it exists of course. Truenas only needs a unique id for each disk attached to its own system and the disk's gpt table's ptuuid can also be assigned manually, so you could try to initialize it from the reported disk id, ... and make sure it's always unique. In this case, we do not have to rely on external parties and the uniqueness constraint is fullfilled by the software.
> 
> 
> 
> Kind regards,
> 
> 
> 
> Michel

Thanks, I understand what you're saying but you're not answering my questions. What happens if you move one of the disks to another slot with these udev rules in place? Does the serial change? I predict it does because the only uniqueness this script is adding is the ID_INSTANCE key.

MichelBr:
I can try that and I'll report back the contents. I think it will be indeed
like that.


P.s.: I'm trying to answer your questions. Sorry, if I misunderstood.
Hopefully we can move forward now.

Op ma 5 jun. 2023 22:34 schreef Caleb St. John ***@***.***>:

MichelBr:
With the "fix" in place and switching the disks, udev reports the following info:

1.

root@truenas[~]# udevadm info /dev/sde
P: /devices/pci0000:00/0000:00:14.0/usb4/4-3/4-3:1.0/host4/target4:0:0/4:0:0:0/block/sde
N: sde
L: 0
S: disk/by-id/usb-TOSHIBA_HDWG180_80C0A08QFAUG-0:0
S: disk/by-path/pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:0
E: DEVPATH=/devices/pci0000:00/0000:00:14.0/usb4/4-3/4-3:1.0/host4/target4:0:0/4:0:0:0/block/sde
E: DEVNAME=/dev/sde
E: DEVTYPE=disk
E: DISKSEQ=7
E: MAJOR=8
E: MINOR=64
E: SUBSYSTEM=block
E: USEC_INITIALIZED=6600959
E: ID_VENDOR=TOSHIBA
E: ID_VENDOR_ENC=TOSHIBA\x20
E: ID_VENDOR_ID=152d
E: ID_MODEL=HDWG180
E: ID_MODEL_ENC=HDWG180\x20\x20\x20\x20\x20\x20\x20\x20\x20
E: ID_MODEL_ID=0567
E: ID_REVISION=0125
**E: ID_SERIAL=TOSHIBA_HDWG180_80C0A08QFAUG-0:0
E: ID_SERIAL_SHORT=80C0A08QFAUG**
E: ID_TYPE=disk
E: ID_INSTANCE=0:0
E: ID_BUS=usb
E: ID_USB_INTERFACES=:080650:
E: ID_USB_INTERFACE_NUM=00
E: ID_USB_DRIVER=usb-storage
E: ID_PATH=pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:0
E: ID_PATH_TAG=pci-0000_00_14_0-usb-0_3_1_0-scsi-0_0_0_0
E: ID_PART_TABLE_UUID=2d98f80a-9757-427a-afb3-8cc3727a52a8
E: ID_PART_TABLE_TYPE=gpt
E: DEVLINKS=/dev/disk/by-id/usb-TOSHIBA_HDWG180_80C0A08QFAUG-0:0 /dev/disk/by-path/pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:0
E: TAGS=:systemd:
E: CURRENT_TAGS=:systemd:

---


root@truenas[~]# udevadm info /dev/sdf
P: /devices/pci0000:00/0000:00:14.0/usb4/4-3/4-3:1.0/host4/target4:0:0/4:0:0:1/block/sdf
N: sdf
L: 0
S: disk/by-id/usb-TOSHIBA_HDWG180_80C0A08LFAUG-0:1
S: disk/by-path/pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:1
E: DEVPATH=/devices/pci0000:00/0000:00:14.0/usb4/4-3/4-3:1.0/host4/target4:0:0/4:0:0:1/block/sdf
E: DEVNAME=/dev/sdf
E: DEVTYPE=disk
E: DISKSEQ=8
E: MAJOR=8
E: MINOR=80
E: SUBSYSTEM=block
E: USEC_INITIALIZED=6555800
E: ID_VENDOR=TOSHIBA
E: ID_VENDOR_ENC=TOSHIBA\x20
E: ID_VENDOR_ID=152d
E: ID_MODEL=HDWG180
E: ID_MODEL_ENC=HDWG180\x20\x20\x20\x20\x20\x20\x20\x20\x20
E: ID_MODEL_ID=0567
E: ID_REVISION=0125
**E: ID_SERIAL=TOSHIBA_HDWG180_80C0A08LFAUG-0:1
E: ID_SERIAL_SHORT=80C0A08LFAUG**
E: ID_TYPE=disk
E: ID_INSTANCE=0:1
E: ID_BUS=usb
E: ID_USB_INTERFACES=:080650:
E: ID_USB_INTERFACE_NUM=00
E: ID_USB_DRIVER=usb-storage
E: ID_PATH=pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:1
E: ID_PATH_TAG=pci-0000_00_14_0-usb-0_3_1_0-scsi-0_0_0_1
E: ID_PART_TABLE_UUID=854d3aa8-f17c-48ab-bfe5-388089d64f7f
E: ID_PART_TABLE_TYPE=gpt
E: DEVLINKS=/dev/disk/by-id/usb-TOSHIBA_HDWG180_80C0A08LFAUG-0:1 /dev/disk/by-path/pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:1
E: TAGS=:systemd:
E: CURRENT_TAGS=:systemd:

---

The long serial id seems not ok at all as it changes when I switches disks. I think this comes from the usb-rules, but I would have to look at the source code to be sure. But switching disk-locations shouldn't affect the disk-id I would think.

Fyi:
- Truenas came up correctly after switching the disks, but that was expected as the pools use parition-id's. 
- The gui shows the disks below the pools using their linux disk sd-id (sde and sdf) and these changed of course after switching the disks. sde is now sdf and vice versa. But that behaviour is not specific to these udev-rules I suppose.

I hope this helps. Should you need any more info, let me know.

Michel

MichelBr:
Fiy, but maybe you noticed already, as it's a very minimal difference, the serial id of these drives (the short serial id) are different, but only in 1 letter Q<->L:

80C0A08**L**FAUG
80C0A08**Q**FAUG

yocalebo:
> > Thanks for sharing this @MichelBr!
> > It seems that this workaround for known USB enclosures via Udev rules works like a charm!
> > In my tests using with an Orico enclosure (VL717 USB 3.1 Gen2 / 10Gbps), I had to swap lines 32 and 33 below on `get_disk_serial()` to give precedence to `ID_SERIAL_SHORT` over `ID_SCSI_SERIAL`, otherwise the disks would still not be recognized because `ID_SCSI_SERIAL` was still the same:
> > https://github.com/truenas/middleware/blob/b5571f0462ef009060cb56ab0c445c0355e5fab4/src/middlewared/middlewared/plugins/device_/device_info.py#L30-L34
> >

The problem is having to make this change. That has a potentially profound impact on all the enterprise hardware that we sell unfortunately.

MichelBr:
@yocalebo This doesn't need to be changed as this change only adds udev-rules and those udev-rules only work on the usb-chipsets that convert sata to USB and which have this behaviour. 

If your enterprise-hardware would make use of this hardware, you would have noticed this issue already. So, this patch shouldn't affect any of your enterprise setups. They work as-is, and this code won't be invoked in their case. Only if they would add USB(-c) storage that connects to these kinds of drive-arrays and I suppose that they would be happy this code is in place then :).

So, this code doesn't affect your Enterprise customers. I'm aware that you don't want to interfere with them, but the udev-rules are very specific to the scenario's that currently don't work yet and make them work.

MichelBr:
@andersonfreitas What didn't work before you changed the preference in disk_serial? I didn't need to make this change. The long serial and the short serial are both unique for me with this patch. I suppose I need to see if the udev-rules can also just change the ID_SCSI_SERIAL-key. Then this doesn't impact the truenas-codebase as requested.

MichelBr:
Worst case this could be solved, but adding a customer property to the truenas codebase, before ID_SCSI_SERIAL, if that would be fine, but I'll first try with the udev rules.

yocalebo:
@MichelBr thanks, that was my next suggestion was to test to see if these udev rules could change the `ID_SCSI_SERIAL` (or set it) attribute instead. That could make it easier for me to push into our product.

One last item that bugs me. I don't believe running `hdparm` is necessary here since we've already got the `ID_SERIAL_SHORT` attribute being reported. I think these rules could be simplified greatly by simply taking `ID_SERIAL_SHORT` and adding `ID_INSTANCE` to it to make `ID_SCSI_SERIAL`

MichelBr:
just keeping this open. Didn't have time yet, but hope to get to it this weekend.

MichelBr:
@andersonfreitas  Could you test with the modified patch? The serial_id-script just adds/overwrites the extra ID_SCSI_SERIAL-variable. This should make it work for you without having to resort to changing the truenas-code. In my setup I have no ID_SCSI_SERIAL, but this could should overwrite it too normally, but it's nice if you could test this. Thanks.

@yocalebo I delved into the why of this whole issue a bit deeper, but first let me summarize it:

- the linux usb storage driver, disables the try_vpd-flag on usb-disks by default as some chipsets may have issues with it (in general). Because of this the scsi_serial is not available for usb disks by default. It can be overwritten using the scsi_mod-kernel-parameter "dev_flags". However, as this is on the scsi-level, you can't specify the usb-model- and vendor here I suppose (I didn't try this yet, but it seems logical at first sight). So, you would need to setup this parameter to enable the try_vpd-flag for each drive-vendor/model behind all usb-chipsets. From a user-standpoint this is not very userfriendly according to me.
- Although the usb storage driver disables try_vpd by default, one can still issue the scsi-command for it directly of course. This is why tools like hdparm or sg_vpd are able retrieve the correct serial id of the disk and you can import them into the udev-environment as this patch does.

So, this summarizes the issues on a technical level. Some thoughts from my side:

- I wouldn't use the INSTANCE_ID-udev environment variable as it depends on the location in the usb drive array and the goal is to have a unique disk-identifier, which is not dependend on its location, nor which chipset it is behind.
- So, I would use the current retrieval logic based on the scsi-vpd (0x80 inquiry)-command to retrieve this info.
- There are 2 ways to achieve the scsi-vpd logic: 
  - or I submit a patch to the linux kernel mailing list to add an extra parameter to the usb storage-driver to be able to change the default try_vpd-flag for certain usb-chipsets/Vendors in the same logic as the scsi_dev-structure currently allows, but then from the usb-storage drivers, which sets the initial values.
  -  or I fix up this script to the way you like, which is only triggered for usb drive enclosures. However, based on the current kernel logic, it is needed for any usb-based drive enclosure I suppose. Although I prefer to not duplicate code which is currently maintained already by somebody, I could for example write a small C- or Rust-program to do the vpd-fetch specifically an leave out the rest. I looked it up and the code is very minimal. Isuppose the program will be a but smaller than hdparm or sg_vpd, but it's specific to Truenas then and does the same. hdparm is 140Kb and sg_Vpd 115Kb currently.

I'll add some more specific details of the current kernel logic in a second post. But please tell me you thoughts. Thank you,
Michel

MichelBr:
Scsi-vpd pages contain the disk info like serial, vendor, ...

Current linux vpd-logic. To give credit where it's due, this page put me in the right direction: https://mpdesouza.com/blog/kernel-adventures-enabling-vpd-pages-for-usb-storage-devices-in-sysfs/

usb-disk-devices are initialized from the usb storage driver: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/usb/storage/scsiglue.c

By default "skip_vpd pages" is enabled by default, unless whitelisted for the specific scsi-devices being created:
image

Whitelisting function:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/scsi/scsi_devinfo.c
image

Via modinfo:
image

Suse has some examples: https://www.suse.com/ja-jp/support/kb/doc/?id=000016555

This parameter is also exported into sysfs, but unfortunately it's readonly "blacklist". It calls blacklist, but basically it's the same. It allows you to specify the allowed scsi-commands.

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/scsi/scsi_sysfs.c

So, I hope this sheds some light on the current Linux kernel logic for these devices, which seem to match the behaviour we're experiencing here.

MichelBr:
To be honest, I prefer fixing it specifically for Truenas using the script, or a C/Rust-program initially, as the solution is basically the same as the kernel patch on a functional level and is present (or almost). It's just where it's applied (udev vs kernel-parameters). 
If you really want the kernel-way, I can schedule it for the long run, but I hope you find the offered udev-possibilities for a faster inclusion into Truenas, which are technically also sound normally.

william-gr:
I would like to thank you for the insight and contribution.

However at this time we are not comfortable integrating this PR as-is since we would like to do our own diligence and develop our own solution rather than copying from OMV and for that to happen we will also need our own hardware to play with.

I am creating and scheduling that work through NAS-122815 and closing this PR for now.

Thank you for understanding!

MichelBr:
Thank you for the update. 

I understand your point of view to have more control, but I was hoping to share some effort between the projects as both are license-compatible and share the same goal. I'm happy the provided info may help. 
As I'm using a usb-based setup currently, I'll create the necessary config for now, so that I'm able to upgrade between releases and do not have to wait on your usb-support. Should you need a tester for your usb-feature, feel free to contact me.

I'm unable to follow the jira-ticket as it's not public. Not sure if you're willing to open it up, so I can follow its progress a bit.

Looking forward to further developments.

william-gr:
Didn't realize the ticket was private. Should be public now.

PR #11256, opened 2023-05-08
NAS-121871 / 23.10 / Make sure unencrypted dataset cannot be created inside an encrypted dataset

sonicaj:


themylogin (changes requested):
Tests

src/middlewared/middlewared/plugins/pool_/dataset.py
```diff
@@ -579,6 +579,14 @@ async def do_create(self, app, data):
         if not inherit_encryption_properties:
             encryption_dict = {'encryption': 'off'}
 
+        if parent_ds['encrypted'] and data['encryption'] is False and not inherit_encryption_properties:
```
yocalebo, line 582: We probably have to check the entire ancestor tree. (i.e. encrypted->parentA->childA)

yocalebo (changes requested):

themylogin (commented):
What are the issues?

We currently allow putting system dataset on a pool with an encrypted root dataset (system dataset will be unencrypted), will there be any problems with that?

sonicaj:
@themylogin when native zfs encryption support was added, motivation was to keep the design simple and not introduce undue complexity.
Wrt system dataset, no it is fine - in terms of immediate problem which occurs with this is child getting mounted where parent is locked i.e
`/mnt/pool/encrypted/child`, we will see people writing to `encrypted` even it is locked whereas child is mounted.

themylogin (approved):

yocalebo (approved):

PR #11257, opened 2023-05-08
NAS-112093 / 23.10 / Set ashift to 12 for pools which do not have it

sonicaj:
This PR adds changes to change `ashift` to `12` if it is at `DEFAULT` value whenever a zfs pool is updated or extended.


yocalebo:
> I had a ticket that looks related https://ixsystems.atlassian.net/browse/NAS-121321 should this be done with boot pool also?

@sonicaj we should definitely do this for boot drives too.

yocalebo (approved):

PR #11260, opened 2023-05-09
NAS-121881 / 13.0 / Fix flapping enclosure alerts

themylogin:
Enclosure elements can "flap" (i.e. report a bad status but then be cleared almost immediately). The alert that is flapping is a particular problem in and of itself. However, this fixes the scenario so that we wait a bit before we raise an alert to the end-user.


themylogin:
time 4:00

PR #11261, opened 2023-05-09
NAS-121632 / 23.10 / Don't create swap from data pool if swap partitions in boot pool exist

Qubad786:
## Context

Changes have been made to not use data pool disks for swap configuration if swap partitions exist in the boot-pool.
While at it have also done some optimizations to not list all disks partitions while configuring swap to handle cases where systems have thousands of disks which are not involved in pools or the system is HA capable and data pool based disks should not be touched at all. In cases like these we only take those disks into account for listing partitions which can actually be used in swap.


yocalebo (approved):
Great job!

PR #11298, opened 2023-05-12
Improve nfs CI and other nfs fixes

mgrimesix:
The goal was to enhance the existing tests and to fix any issues that were exposed.
* Fixed nfs_/debug.py as it was not functioning as expected
* Enhanced behavior testing in test_300_nfs.py
    - Added client testing where possible
    - Added independent verification of setting changes where possible

### DIFF
--- a/src/middlewared/middlewared/plugins/nfs_/debug.py
+++ b/src/middlewared/middlewared/plugins/nfs_/debug.py
@@ -149,13 +149,16 @@ class NFSService(Service):
                 return rv
 
             for x in opts:
-                rv += self.dbgcls[svc][x].value
+                rv = rv | self.dbgcls[svc][x].value
 
             return rv
 
         for svc, opts in services.items():
             if opts == []:
                 continue
+            if "NONE" in opts and len(opts) > 1:
+                raise ValueError(f"Cannot specify another value with NONE: {svc}={opts}")
+
             to_set = "0x%0.4X" % debug_level_to_int(svc, opts)
 
             with open(f"/proc/sys/sunrpc/{svc.lower()}_debug", "w") as f:
--- a/src/middlewared/middlewared/plugins/nfs_/debug.py
+++ b/src/middlewared/middlewared/plugins/nfs_/debug.py
@@ -1,4 +1,4 @@
-from middlewared.schema import accepts, List, Str
+from middlewared.schema import accepts, List, Str, Dict
 from middlewared.service import Service, private
 from contextlib import suppress
 
@@ -6,7 +6,7 @@ import os
 import enum
 
 
-class NFSDBG(enum.Enum):
+class NFS_DBGFLAGS(enum.Enum):
     # include/uapi/linux/nfs_fs.h
     NONE = 0x0000
     VFS = 0x0001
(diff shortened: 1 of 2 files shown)


tests/api2/test_300_nfs.py
```diff
@@ -681,6 +804,10 @@ def test_39_check_nfs_service_protocols_parameter(request):
     assert s['nfsd']["vers3"] == 'y', str(s)
     assert s['nfsd']["vers4"] == 'n', str(s)
 
+    # Confirm setting has taken effect: v4->off, v3->on
+    results = SSH_TEST("rpcinfo -s | grep 100003", user, password, ip)
```
bmeagherix, line 808: Hmmm.  Why 100003?  It's not clear what this is.  Would it be better to search for service name == nfs?
```
root@scale1[~]# rpcinfo -s | awk '$4 == "nfs"'
    100003  4,3       tcp6,tcp                         nfs         superuser
```

Or at least add a define if continuing to use 100003

Same comment applies to duplicate commands below.

tests/api2/test_300_nfs.py
```diff
@@ -713,32 +844,52 @@ def test_39_check_nfs_service_protocols_parameter(request):
...
+        #    DB == False, conf == 'n'
+        res = make_ws_request(ip, get_payload)
+        assert res['result']['udp'] is False, res
+        s = parse_server_config()
+        assert s['nfsd']["udp"] == 'n', str(s)
+
+        # Multiple restarts cause systemd failures.  Reset the systemd counters.
+        svcs_to_reset = "nfs-idmapd nfs-mountd nfs-server rpcbind rpc-statd"
+        results = SSH_TEST(f"systemctl reset-failed {svcs_to_reset}", user, password, ip)
+        assert results['result'] is True
```
bmeagherix, line 872: Maybe extract this out for reuse?  I see another call over in _test_015_services.py::test_01_non_silent_service_start_failure_

tests/api2/test_300_nfs.py
```diff
@@ -123,17 +181,44 @@ def nfs_share(path, options=None):
         assert result.status_code == 200, result.text
 
 
+@contextlib.contextmanager
+def nfs_config(options=None):
+    '''
+    Use this to restore settings after the test
+    '''
```
bmeagherix, line 188: This could be correct, but IMO it might be worthwhile to add a comment explain why there are **two** different NFS config save/restore mechanisms, and _why_ they are called from where they are.

bmeagherix (changes requested):
Looks very good.  Can only see a few minor items.

src/middlewared/middlewared/plugins/nfs_/debug.py
```diff
@@ -28,56 +28,137 @@ class NFSDBG(enum.Enum):
...
             rv = 0
-            if "NONE" in lvl:
+
+            if "NONE" in opts:
                 return rv
 
-            for x in lvl:
-                rv += NFSDBG[x].value
+            for x in opts:
+                rv += self.dbgcls[svc][x].value
```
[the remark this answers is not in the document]
mgrimesix: Agreed, bitwise OR is better.

tests/api2/test_300_nfs.py
mgrimesix: I agree that 100003 is not descriptive, but it's the well-known unique program number.  There are other services that include 'nfs' in their name, but ' nfs ' should work.  I'll test it.

tests/api2/test_300_nfs.py
mgrimesix: Agreed.

src/middlewared/middlewared/plugins/nfs_/debug.py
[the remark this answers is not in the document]
mgrimesix: done

tests/api2/test_300_nfs.py
```diff
@@ -80,6 +80,57 @@ def parse_server_config(fname="local.conf"):
...
+
+def confirm_rpc_processes(expected=['idmap', 'bind', 'statd']):
+    '''
+    Confirm the expected rpc processes are running
+    '''
+
+    rx_mountd = r"rpc\.mountd"
+    result = SSH_TEST(f"ps -ef | grep '{rx_mountd}' | wc -l", user, password, ip)
+    # We subtract one to account for the rpc.mountd thread manager
+    assert int(result['output']) - 1 == expected
```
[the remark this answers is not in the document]
mgrimesix: done

tests/api2/test_300_nfs.py
mgrimesix: done

tests/api2/test_300_nfs.py
mgrimesix: done

tests/api2/test_300_nfs.py
mgrimesix: done

bmeagherix (approved):

PR #11301, opened 2023-05-15
NAS-121954 / 23.10 / Skip read of authorized keys where possible

anodos325:
If user does not have have a homedir we should not try to read its authorized_keys file (which would be a read of /.ssh/authorized_keys).


yocalebo:
retest this please

anodos325:
This is basically superseded by different ticket I opened related paths.

PR #11309, opened 2023-05-15
NAS-121697 / 23.10 / Start service if it stopped for lockable fs attachment delegates

Qubad786:
## Problem

Services are not restarted when datasets are unlocked if they were already stopped at the time of unlocking. In a change introduced in https://github.com/truenas/middleware/pull/7388, motivation was that we should always be restarting/reloading necessary services and that does not take into account the case where service was stopped at the time dataset was unlocked.

## Solution

After discussing with Caleb, we decided that the service if stopped should be started on dataset unlock if the service is meant to be started on boot. Changes have been made to Lockable FS attachment delegate so any attachments defined using this will be making use of this change.

## Notes:

For testing this change, i configured different shares and tried following cases to see that we are not regressing:
1. Exporting pool
2. Deleting dataset
3. Locking dataset
4. Importing pool
5. Unlocking dataset

### DIFF
--- a/src/middlewared/middlewared/common/attachment/__init__.py
+++ b/src/middlewared/middlewared/common/attachment/__init__.py
@@ -89,6 +89,14 @@ class LockableFSAttachmentDelegate(FSAttachmentDelegate):
             filters += [[self.locked_field, '=', options['locked']]]
         return filters
 
+    async def start_service(self):
+        if not (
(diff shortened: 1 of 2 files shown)


themylogin (changes requested):
An integration test can easily be added here

themylogin (approved):

PR #11321, opened 2023-05-17
Fix openvpn server logs ownership

sonicaj:
This PR fixes the ownership of openvpn server based log files which happened to be `root/root`. A user reported issues with his OpenVPN server not accepting clients because of these permission issues which when he manually fixed resulted in successful connection.

The user in question had some extra parameters specified as well but i think it is generally speaking good practice to do this.

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/openvpn/server/perms.py
+++ b/src/middlewared/middlewared/etc_files/local/openvpn/server/perms.py
@@ -1,0 +1,29 @@
+import errno
+import os
+
+from middlewared.service import CallError
+
+
+LOGS_FOLDER = '/var/log/openvpn'
+
+
+def fix_perms(middleware):
+    if nobody_user := middleware.call_sync('user.query', [('username', '=', 'nobody')]):
+        nobody_uid = nobody_user[0]['uid']
+        nobody_gid = nobody_user[0]['group']['bsdgrp_gid']
+    else:
+        raise CallError('Unable to locate "nobody" user', errno=errno.ENOENT)
+
+    os.makedirs(LOGS_FOLDER, exist_ok=True)
+    for path in (
+        LOGS_FOLDER, *map(lambda file: os.path.join(LOGS_FOLDER, file), ('openvpn.log', 'openvpn-status.log'))
+    ):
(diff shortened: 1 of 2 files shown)


yocalebo (changes requested):

src/middlewared/middlewared/etc_files/local/openvpn/server/perms.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def fix_perms(middleware):
+    if nobody_user := middleware.call_sync('user.query', [('username', '=', 'nobody')]):
+        nobody_uid = nobody_user[0]['uid']
+        nobody_gid = nobody_user[0]['group']['bsdgrp_gid']
+    else:
+        raise CallError('Unable to locate "nobody" user', errno=errno.ENOENT)
+
+    log_dir = Path(LOGS_FOLDER)
```
yocalebo, line 19: There is no reason to define a global LOGS_FOLDER now. Just move it inside the function.

src/middlewared/middlewared/etc_files/local/openvpn/server/perms.py
```diff
@@ -0,0 +1,29 @@
...
+    log_dir.mkdir(parents=True, exist_ok=True)
+    for path_attr in (
```
yocalebo, line 21: Why are you still doing this complicated logic? A simple for loop will suffice....

sonicaj: It is a simple for loop ?
Ownership needs to be changed for the directory as well so it needs to be in the for loop as well

sonicaj: I don't mind either ways but curious why write the same chown statement 2 times instead of having log dir in the for loop as well ?

anodos325:
Why would log files need to be owned by `nobody`? That sounds wrong.

sonicaj:
@anodos325 that is because openvpn server runs as the following:
```
user nobody
group nogroup
```

yocalebo (approved):

PR #11346, opened 2023-05-18
NAS-122006 / 23.10 / add ipmi.lan plugin and hide ipmi lan channel 8 on F1 HA

yocalebo:
This does a few things:
- add an ipmi.lan plugin since the generic "ipmi" plugin isn't the proper namespace
- mask lan channel 8 on the F1 HA platform
- move the crud portion of the generic ipmi plugin to the ipmi.lan plugin
- use `ipmi-config` binary which prevents (potentially) 16x subprocesses when the channels function is first called

NOTE: I've kept the same logic for the ipmi.lan.update that was in ipmi.update. However, this could be simplified (I predict) by using `ipmi-config` as well. That will come in another PR.

### DIFF
--- a/src/middlewared/middlewared/plugins/ipmi.py
+++ b/src/middlewared/middlewared/plugins/ipmi.py
@@ -1,167 +1,19 @@
 import os
-from subprocess import run, DEVNULL
-from functools import cache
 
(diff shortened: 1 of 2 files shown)


Qubad786 (approved):

yocalebo:
retest this please

PR #11364, opened 2023-05-21
NAS-121957 / 23.10 / Set proper homedir for built-in users

themylogin:


anodos325:
Needs commit message of some sort.

anodos325:
Perhaps I'm missing something, but how does this fix case where DB insertion with wrong homedir has already occurred?

themylogin:
> Needs commit message of some sort.

What is wrong with the current commit message?

> Perhaps I'm missing something, but how does this fix case where DB insertion with wrong homedir has already occurred?

This method updates built-in users database on every TrueNAS version update, including homedirs.

yocalebo:
retest this please

yocalebo:
retest this please

yocalebo (changes requested):
While this might fix some, it doesn't fix all. For example, the `bin` user has `/` set as the home directory in the `/etc/passwd` db file. This is because that's the default on BSD* based systems, however, on linux its default is `/bin`. I think we should probably compare the passed file we ship with a stock debian install to make sure we've got them matched up as best as possible

themylogin:
@yocalebo where does `/bin` user have `/` set as their home directory?

yocalebo (approved):

themylogin:
time 2:00

PR #11375, opened 2023-05-23
NAS-122102 / 23.10 / Prefer sids privileges

anodos325:
SIDS are more stable (can't change value based on user configuration)
and so should be preferred for mapping privileges for directory
services groups.

This PR adds a bulk conversion API for mapping unix ids to SIDs, which
is used for privilege evaluation.


anodos325:
https://github.com/truenas/middleware/pull/11378 will add testing for setting privilege via SID.

PR #11385, opened 2023-05-24
NAS-120951 / 23.10 / add hardware.memory.error_info endpoint

yocalebo:
We currently have a mechanism to report on APEI or MCA events on CORE for our enterprise users. This brings us to mostly feature parity with CORE by getting the relevant information from sysfs.

### DIFF
--- a/src/middlewared/middlewared/alert/source/memory_errors.py
+++ b/src/middlewared/middlewared/alert/source/memory_errors.py
@@ -1,20 +1,11 @@
 from middlewared.alert.base import Alert, AlertCategory, AlertClass, AlertLevel, AlertSource
 
 
-class MemoryErrorsNoInfoAlertClass(AlertClass):
+class MemoryErrorsAlertClass(AlertClass):
     category = AlertCategory.HARDWARE
     level = AlertLevel.WARNING
     title = 'Uncorrected Memory Errors Detected'
-    text = '%(count)d total uncorrected errors detected for memory controller "%(mc)s".'
-    products = ('SCALE_ENTERPRISE',)
-    proactive_support = True
-
-
-class MemoryErrorsInfoAlertClass(AlertClass):
-    category = AlertCategory.HARDWARE
-    level = AlertLevel.WARNING
-    title = 'Uncorrected Memory Errors Detected'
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/alert/source/memory_errors.py
```diff
@@ -0,0 +1,42 @@
...
+
+            for dimm_key in filter(lambda x: x.startswith(('dimm', 'rank')), info):
+                if (value := info[dimm_key]['uncorrected_errors']) is not None and value > 0:
+                    # the specific dimm
+                    errors[mem_ctrl].update({dimm_key: {'uncorrected_errors': value}})
+                    alert = True
+
+        if alert:
+            serial = await self.middleware.call('system.dmidecode_info')['system-serial-number']
+            job = await self.middleware.call('support.new_ticket', {
```
themylogin, line 29: If you don't want this method to emit an alert that is visible in the alerts UI section, there is no reason to make this an alert source, it can be just a periodically ran job.

However, if you add an alert class and set it with `proactive_support = True`, a support ticket will be created automatically (moreover, it won't be creating new support tickets every 24 hours)

themylogin (approved):

src/middlewared/middlewared/alert/source/memory_errors.py
```diff
@@ -0,0 +1,43 @@
...
+                # Because of this fact, we'll just report the error count without the DIMM information.
+                alerts.append(Alert(MemoryErrorsNoInfoAlertClass, {'mc': mem_ctrl, 'count': val}))
+            elif (val := info['uncorrected_errors']) is not None and val > 0:
+                # this means that there were uncorrected errors where the dimm information was able
+                # to be obtained.
+                for dimm_key in filter(lambda x: x.startswith(('dimm', 'rank')), info):
+                    if (val2 := info[dimm_key]['uncorrected_errors']) is not None and val2 > 0:
+                        # the specific dimm
+                        alerts.append(Alert(
+                            MemoryErrorsInfoAlertClass, {'mc': mem_ctrl, 'count': val2, 'dimm': dimm_key}
```
themylogin, line 40:
```suggestion
                            MemoryErrorsNoInfoAlertClass, {'mc': f'{mem_ctrl} on {dimm_key}', 'count': val2}
```

PR #11391, opened 2023-05-25
NAS-121891 / 23.10 / Consume update warnings and allow resuming the update after warning

themylogin:
Warning will be returned as `EAGAIN` `CallError`. The UI will check for this specific `errno`, present a yes/no dialog, and, if "yes" is selected, will call the same method but with `{"resume": true}` option (no file re-upload will be necessary, no repeated network requests will be performed).


themylogin:
backport

PR #11397, opened 2023-05-25
NAS-117915 / 23.10 / Solve identical gpu passthrough problem

Qubad786:
## Problem

We were using product/vendor id to isolate GPUs which caused a problem when identical GPUs were present in the system. What happened in that case is that the product/vendor id is same for identical GPUs and it results in all the identical GPUs being isolated.

## Solution

Wrote a short script which explicitly consume the pci slots address and add a driver override for `vfio-pci` so that it is the only one consumed. Following is an output showing one of the identical GPUs being properly isolated from the host(you can check from the `Kernel driver in use` attribute):

```
03:00.0 VGA compatible controller [0300]: ASPEED Technology, Inc. ASPEED Graphics Family [1a03:2000] (rev 41)
	DeviceName: ASPEED Video AST2500
	Subsystem: Super Micro Computer Inc ASPEED Graphics Family [15d9:095d]
	Kernel driver in use: ast
--
17:00.0 VGA compatible controller [0300]: NVIDIA Corporation TU117GL [T400 4GB] [10de:1ff2] (rev a1)
	Subsystem: Lenovo TU117GL [T400 4GB] [17aa:1613]
	Kernel driver in use: vfio-pci
	Kernel modules: nouveau, nvidia_current_drm, nvidia_current
--
65:00.0 VGA compatible controller [0300]: NVIDIA Corporation TU117GL [T400 4GB] [10de:1ff2] (rev a1)
	Subsystem: Lenovo TU117GL [T400 4GB] [17aa:1613]
	Kernel driver in use: nvidia
	Kernel modules: nouveau, nvidia_current_drm, nvidia_current
```

I also verified that the new `truenas_bind_vfio.sh` script is properly added to initramfs image using `lsinitramfs`.

### DIFF
--- a/src/freenas/usr/local/bin/truenas-initrd.py
+++ b/src/freenas/usr/local/bin/truenas-initrd.py
@@ -68,21 +68,21 @@ def update_zfs_default(root):
 def get_current_gpu_pci_ids(root):
     adv_config = query_config_table("system_advanced", os.path.join(root, FREENAS_DATABASE[1:]), "adv_")
     to_isolate = [gpu for gpu in get_gpus() if gpu["addr"]["pci_slot"] in adv_config.get("isolated_gpu_pci_ids", [])]
-    return [dev["pci_id"] for gpu in to_isolate for dev in gpu["devices"]]
+    return [dev["pci_slot"] for gpu in to_isolate for dev in gpu["devices"]]
 
 
 def update_pci_module_files(root, config):
     def get_path(p):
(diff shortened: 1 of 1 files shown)


sonicaj (approved):
Good work!
When we initially added GPU passthrough support, it indeed was a bit weird using product/vendor id but apparently eveeryone was using the same workflow.
Nice to see this works!

PR #11411, opened 2023-05-30
NAS-122231 / 23.10 / Revert "Fix KeyError in API tests (#11404)

sonicaj:
This PR adds changes to make sure whenever an admin user is created, we also add a record of that user to our 2fa table as all local users have a 1-1 relation with the 2fa table and this was a case which we had missed where a user was being added but the relevant 2fa record was not being added.


yocalebo (changes requested):
Please add a proper git commit summary to this PR.

themylogin (approved):

yocalebo (approved):

PR #11413, opened 2023-05-31
NAS-121479 / 23.10 / Allow updating renew_days attribute for ACME certs

sonicaj:
This PR adds changes to allow updating renew days attribute for ACME certs so users can define when they expect the cert to be renewed i.e 5 days before expiry or 1 day before expiry etc.

### DIFF
--- a/src/middlewared/middlewared/plugins/crypto_/certificates.py
+++ b/src/middlewared/middlewared/plugins/crypto_/certificates.py
@@ -172,7 +172,7 @@ class CertificateService(CRUDService):
             Int('csr_id'),
             Int('signedby'),
             Int('key_length', enum=[2048, 4096]),
-            Int('renew_days'),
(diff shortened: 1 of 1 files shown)


themylogin (approved):

themylogin (approved):

src/middlewared/middlewared/plugins/crypto_/certificates.py
```diff
@@ -595,7 +595,7 @@ async def do_update(self, job, id, data):
 
         new.update(data)
 
-        if any(new[k] != old[k] for k in ('name', 'revoked')):
+        if any(new.get(k) != old.get(k) for k in ('name', 'revoked', 'renew_days')):
```
yocalebo, line 598: What if both new.get(k) and old.get(k) are None?

sonicaj: It means nothing changed, we are good

PR #11421, opened 2023-05-31
NAS-117546 / 23.10 / Use chronyd rather than ntpd for time keeping

bmeagherix:
Switch from `ntp` to `chrony` for extra robustness.

Added `NTPPeer` to reduce leakage of implementation details throughout the codebase.

----
The scale-build PR https://github.com/truenas/scale-build/pull/438 should be merged before this one.


yocalebo (approved):
Great job!

PR #11429, opened 2023-06-01
NAS-122211 / 23.10 / Improve display devices defaults

Qubad786:
## Context

`DISPLAY` devices defaults have been to not ask user to specify `password` as a hard requirement.
However that is not appropriate because if VNC server socket is not protected by password it just means that any malicious actor can just scan ports and then gain access to a VMs console which in most cases is also connected to the user's local network.
Now specifying `password` is a hard requirement for any one looking to create more `DISPLAY` devices or edit the ones they have already. Also by default we are going to bind the VNC/spice server to `localhost` instead of wildcard ip i.e `0.0.0.0` improving the defaults further so only if anyone really wants to expose their VNC server socket they can do so explicitly.

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -1822,6 +1822,7 @@ class InterfaceService(CRUDService):
         ignore_nics = self.middleware.call_sync('interface.internal_interfaces')
         if choices['loopback']:
             ignore_nics.remove('lo')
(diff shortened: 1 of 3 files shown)


sonicaj (approved):
To confirm, you have tested both SPICE/VNC to make sure they work nicely with localhost ?

yocalebo:
Let's make sure to open a webUI ticket if we need to so they can improve or make any changes if necessary. If one isn't needed, then disregard.

PR #11432, opened 2023-06-01
NAS-120654 / 23.10 / Add CI tests for FTP

mgrimesix:
* Change defaultroot 'default' setting from False to True to match actual
* Add 'clear' of anonpath when onlyanonymous is disabled

### DIFF
--- a/tests/api2/test_200_ftp.py
+++ b/tests/api2/test_200_ftp.py
@@ -742,7 +742,7 @@ def test_001_validate_default_configuration(request, ftp_init_db_dflt):
         diffs = {}
         for setting in set(DB_DFLT) & set(db):
             # Special cases: ftp_anonpath is 'nullable' in the DB, but the default is False
-            if setting == "anonpath" and db[setting] == '':
+            if setting == "anonpath" and (db[setting] == '' or db[setting] is None):
                 db[setting] = False
             # Special cases: Restore 'None' for empty string
             if setting in ['banner', 'options', 'masqaddress'] and db[setting] == '':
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/ftp.py
```diff
@@ -199,6 +199,9 @@ async def do_update(self, data):
                 verrors.add("ftp_update.anonpath", "This field is required for anonymous login")
             else:
                 await check_path_resides_within_volume(verrors, self.middleware, "ftp_update.anonpath", new["anonpath"])
+        else:
+            # onlyanonymous is disabled, clear anonpath
+            new["anonpath"] = ""
```
anodos325, line 204: Since value is nullable, perhaps `None` makes more sense?

tests/api2/test_200_ftp.py
```diff
@@ -7,48 +7,1457 @@
...
+        retry = 6
+        while res.get('error') is not None and retry > 0:
+            sleep(2)
+            res = make_ws_request(ip, payload)
+            retry -= 1
+        assert res.get('error') is None, res
+
+
+@contextlib.contextmanager
+def ftp_configure(changes={}):
```
anodos325, line 319: Let's avoid using `{}` as a default value for kwarg. `changes = None`.

tests/api2/test_200_ftp.py
```diff
@@ -7,48 +7,1457 @@
...
+    '''
+    Apply requested FTP configuration changes.
+    Restore original setting when done
+    '''
```
anodos325, line 323: Inside method you can do something like `changes = changes or {}`

anodos325 (approved):

PR #11437, opened 2023-06-02
NAS-122320 / 23.10 / Properly handle unattributed Storj buckets

jewharton:
In order to determine whether a Storj bucket was associated with iX Systems, the text contents of the bucket's `Attribution` tag in Storj's S3 XML response was checked to see whether it contained the iX-Storj user agent. However, a `TypeError` exception was raised if the bucket was unattributed because the content of the empty `Attribution` element was `None`. This issue has been fixed by gracefully handling the empty element.


themylogin (approved):

yocalebo:
@jewharton thanks for the contribution!

PR #11440, opened 2023-06-03
NAS-121867 / 23.10 / Fix CRUD service events/hooks not working with job based methods

sonicaj:
This PR fixes create/update/delete post execution events/hooks when any of those methods are jobs as we were not waiting for the job to actually finish before trying to fire events and it has never worked for job based methods. Now with the current changes we properly wait for the actual method to complete execution and then we fire events/hooks regardless of the actual method being a job or not.

### DIFF
--- a/tests/api2/test_crud_events.py
+++ b/tests/api2/test_crud_events.py
@@ -39,7 +39,7 @@ def gather_events(event_endpoint: str, context_args: dict = None):
         'timeout': 60,
         **(context_args or {})
     }
-    thread = threading.Thread(target=event_thread, args=(event_endpoint, context))
+    thread = threading.Thread(target=event_thread, args=(event_endpoint, context), daemon=True)
     thread.start()
     if not context['start_event'].wait(timeout=30):
         raise Exception('Timed out waiting for event thread to start')
@@ -53,7 +53,6 @@ def gather_events(event_endpoint: str, context_args: dict = None):
 
 
 def assert_result(context: dict, event_endpoint: str, oid: typing.Union[int, str], event_type: str) -> None:
-    assert context['result'] is not None
     assert context['result'] == {
         'msg': event_type,
         'collection': event_endpoint,
--- a/tests/api2/test_crud_events.py
+++ b/tests/api2/test_crud_events.py
@@ -64,3 +64,24 @@ def test_event_create_on_non_job_method():
                 'collection': 'certificateauthority.query',
                 'id': root_ca['id'],
             }, context['result']
+
+
+def test_event_create_on_job_method():
+    with gather_events('certificate.query') as context:
+        with root_certificate_authority('root_ca_create_event_test') as root_ca:
+            cert = call('certificate.create', {
+                'name': 'cert_test',
+                'signedby': root_ca['id'],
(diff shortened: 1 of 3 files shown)


themylogin (changes requested):
Can we have an integration test for the specific plugin that needs this?

sonicaj:
@themylogin with current changes, both job based and non-job based methods will be affected. I'll write an integration test covering both.

tests/api2/test_crud_events.py
```diff
@@ -0,0 +1,146 @@
...
+    try:
+        yield context
+    finally:
+        if context['event'] and context['event'].is_set() is False:
+            context['event'].set()
+        thread.join(timeout=5)
+
+
+def assert_result(context: dict, event_endpoint: str, oid: typing.Union[int, str], event_type: str) -> None:
+    assert context['result'] is not None, context
```
themylogin, line 67: Why do you need to specify stuff after comma?

tests/api2/test_crud_events.py
```diff
@@ -0,0 +1,135 @@
...
+
+def event_thread(event_endpoint: str, context: dict):
+    with client(py_exceptions=False) as c:
+        context['start_event'].set()
+        subscribe_payload = c.event_payload()
+        event = subscribe_payload['event']
+        context['event'] = event
+
+        def cb(mtype, **message):
+            if len(message) != 3 or not all(
```
themylogin, line 18: Will that be a problem if another extra key is present in the message?

tests/api2/test_crud_events.py
```diff
@@ -0,0 +1,135 @@
...
+@contextlib.contextmanager
+def gather_events(event_endpoint: str, context_args: dict = None):
+    context = {
+        'result': None,
+        'event': None,
+        'start_event': threading.Event(),
+        'timeout': 60,
+        **(context_args or {})
+    }
+    thread = threading.Thread(target=event_thread, args=(event_endpoint, context))
```
themylogin, line 42: The thread is not daemon so we'll wait indefinitely for it to end on the interpreter shutdown

tests/api2/test_crud_events.py
```diff
@@ -0,0 +1,135 @@
...
+        yield context
+    finally:
+        if context['event'] and context['event'].is_set() is False:
+            context['event'].set()
+        thread.join(timeout=5)
+
+
+def assert_result(context: dict, event_endpoint: str, oid: typing.Union[int, str], event_type: str) -> None:
+    assert context['result'] is not None
+    assert context['result'] == {
```
themylogin, line 57: Is this an error? You assert for the equality of the same variable

tests/api2/test_crud_events.py
sonicaj: That won't happen for CRUD automatic events ?
And if it does, this needs to be updated as well to account for that

sonicaj: It is to avoid other system events which would be emitting from middleware. There are certain APIs which send the same event but under different circumstances so like `replication.query` or `disk.query` send the same query event but it is not sent by middleware on automatic CRUD operations.

themylogin (approved):

PR #11447, opened 2023-06-05
NAS-122319 / 23.10 / Fix booting back to FreeBSD

themylogin:
/boot/grub was not mounted at the time this script was started which prevented
    the GRUB config from being updated and resulted in booting back to CORE not
    possible.
    
    Additionally, let's stop running update script when we just need to boot back
    to FreeBSD as this consumes too much time.


themylogin:
time 4:00

PR #11452, opened 2023-06-05
NAS-120267 / 23.10 / Remove geli related attributes from database

sonicaj:
This PR introduces changes to cover usages of encryption fields which we had in pool table.
We still have some leftover references of geli in failover plugin which will be handled in a separate ticket.


yocalebo:
retest this please

PR #11459, opened 2023-06-06
NAS-122332 / 23.10 / remove python3-ntplib dependency

yocalebo:
With the ongoing effort to try and reduce 3rd party dependencies, I've removed the `python3-ntplib` package. This is a very simple single python file module that sends and parses NTP requests.

I've written our own that is simpler and uses vastly less resources since we only need a subset of the functionality the original python module provides.


bmeagherix (approved):

src/middlewared/middlewared/plugins/ntp_/client.py
```diff
@@ -0,0 +1,49 @@
...
+        'recv_timestamp': recv_timestamp,
+        'recv_timestamp_formatted': recv_utc_time,
+        'tx_timestamp': unpacked[13] + float(unpacked[14]) / 2 ** 32,
+    }
+
+
+@dataclasses.dataclass(slots=True)
+class NTPClient:
+    host: str
+    timeout: int = 1  # second
```
bmeagherix, line 35: Maybe make the `timeout` a default parameter to `make_request`, just in case 1 second isn't enough.

src/middlewared/middlewared/plugins/system/info.py
```diff
@@ -192,14 +192,13 @@ def get_synced_clock_time(self):
...
-        client = ntplib.NTPClient()
         try:
-            response = client.request('localhost')
+            response = NTPClient('localhost').make_request()
         except Exception:
             # Cannot connect to NTP server
             self.logger.error('Error while connecting to NTP server', exc_info=True)
         else:
-            if response.version and response.leap != 3:
+            if response['version'] and response['leap'] != 3:
```
bmeagherix, line 201: Maybe we should return a (boolean) `valid` field from **inside** `format_ntp_packet` that encompasses this test

`response['version'] and response['leap'] != 3
`

Then we could use it both here, and in `test_ntp_server`

src/middlewared/middlewared/plugins/ntp_/client.py
yocalebo: When you instantiate the `NTPClient` you can pass a `timeout` argument to it to override it. `make_request` will reference `self.timeout` accordingly.

src/middlewared/middlewared/plugins/system/info.py
yocalebo: Funny you see this, I'm actually looking at this logic. The link to a random ntp program actually has the inverse of our logic. So I believe this is actually a bug. (i.e. it should be `response['leap'] == 3` based on what I'm seeing). Still doing some research on this front.

src/middlewared/middlewared/plugins/ntp_/client.py
bmeagherix, line 35: IIRC `ntplib` used a 5 second timeout, so I'm going to _suggest_ that here too.

bmeagherix (approved):

PR #11464, opened 2023-06-06
NAS-122344 / 23.10 / Rely exclusively on kerberos tickets for AD ops

anodos325:
For quite some time already we don't need the AD bindpw stored in our sqlite database. Pop it out of config dict before insertion in activedirectory.update (after we've performed kinit to get kerberos ticket as admin).


bmeagherix (approved):
I observe that the `bindpw` is used by `validate_credentials`, **before** the call to `ad_compress`, so I _think_ this is safe.

anodos325:
> I observe that the `bindpw` is used by `validate_credentials`, **before** the call to `ad_compress`, so I _think_ this is safe.

Right. we kinit during `validate_credentials` and then use the ticket to perform the required ops.

PR #11469, opened 2023-06-07
NAS-121315 / 23.10 / Cloud backup

themylogin:
Initial cloud backup support.

Introduces `cloud_backup` plugin with CRUD interface and two methods:
* `cloud_backup.init` - a job, should be called after the backup task is created
* `cloud_backup.sync` - a job, performs cloud backup


src/middlewared/middlewared/plugins/cloud/path.py
```diff
@@ -0,0 +1,26 @@
...
+def get_remote_path(provider, attributes):
+    remote_path = attributes["folder"].rstrip("/")
+    if not remote_path:
+        remote_path = "/"
+    if provider.buckets:
+        remote_path = f"{attributes['bucket']}/{remote_path.lstrip('/')}"
+    return remote_path
+
+
+async def check_local_path(middleware, path, *, check_mountpoint=True, error_text_path=None):
```
yocalebo, line 15: Has blocking I/O in this coroutine.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/cloud/model.py
```diff
@@ -0,0 +1,64 @@
...
+
+from middlewared.schema import Bool, Cron, Dict, Int, List, Str
+import middlewared.sqlalchemy as sa
+from middlewared.validators import Range, Time
+
+
+class CloudTaskModelMixin:
+    id = sa.Column(sa.Integer(), primary_key=True)
+    description = sa.Column(sa.String(150))
+
```
yocalebo, line 11: There is a bunch of whitespace in this mixin. Can we remove it? Obviously the credential_id attribute can have spaces around it but not sure about the other seemingly random whitespace.

src/middlewared/middlewared/plugins/cloud_backup/crud.py
```diff
@@ -0,0 +1,149 @@
...
+        if cloud_backup["credentials"]:
+            cloud_backup["credentials"] = cloud_backup["credentials"]["id"]
+
+        cloud_backup.update(data)
+
+        verrors = ValidationErrors()
+
+        await self._validate(verrors, "cloud_backup_update", cloud_backup)
+
+        if verrors:
```
yocalebo, line 98: verrors.check()

src/middlewared/middlewared/plugins/cloud_backup/crud.py
```diff
@@ -0,0 +1,149 @@
...
+        register=True,
+    ))
+    async def do_create(self, cloud_backup):
+        """
+        """
+        verrors = ValidationErrors()
+
+        await self._validate(verrors, "cloud_backup_create", cloud_backup)
+
+        if verrors:
```
yocalebo, line 70: verrors.check()

src/middlewared/middlewared/plugins/cloud/path.py
```diff
@@ -0,0 +1,26 @@
...
+    error_text_path = error_text_path or path
+
+    if not await middleware.run_in_thread(os.path.exists, path):
```
yocalebo, line 18: Instead of spinning up 2 new threads each time this is called, we can do something like
```
try:
    info = await middleware.run_in_thread(os.stat, path)
except FileNotFoundError:
    raise CallError(f"Directory {error_text_path!r} does not exist")
else:
    if not: stat.S_ISDIR(info.st_mode):
        raise CallError(f"{error_text_path!r} is not a directory")

themylogin: @yocalebo it does not spin new threads

yocalebo: Yeah, I used bad verbiage. Can we still consider taking my approach? It still only uses 1 thread in the thread-pool instead of 2.

yocalebo (approved):
Please make sure the db migration is still correct since some have been merged since this PR was opened up.

yocalebo (changes requested):
Actually sorry, looking at one more thing.

themylogin:
time 20:00

PR #11470, opened 2023-06-07
NAS-122331 / 23.10 / fix system.general.timezone_choices and test

yocalebo:
The test was failing which caused me to investigate and find a bunch of problems.

- performing blocking IO in the main event loop
- we are reading from a directory and that was producing a bunch of unexpected entries that differed from the command `timedatectl list-timezones`
- don't read from a directory, just run `timedatectl list-timezones` since it does exactly what we need
- cache the response of this method using `@cache` decorator instead of using a class attribute
- fix the test by removing it. We already call timezone_choices endpoint in this test so there is nothing else to do

### DIFF
--- a/src/middlewared/middlewared/plugins/system_general/timezone.py
+++ b/src/middlewared/middlewared/plugins/system_general/timezone.py
@@ -1,12 +1,11 @@
-import os
+from subprocess import run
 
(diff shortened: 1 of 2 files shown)


yocalebo:
wrong ticket

PR #11480, opened 2023-06-09
NAS-122396 / 23.10 / Fix anonpath validation hole

mgrimesix:
anonpath was allowed to be set unvalidated if onlyanonymous was not enabled.

Also included:
* Clearing anonpath if onlyanonymous is not enabled
* Change the default value of defaultroot from False to True.  This matches our actual default setting

### DIFF
--- a/src/middlewared/middlewared/plugins/ftp.py
+++ b/src/middlewared/middlewared/plugins/ftp.py
@@ -24,7 +24,7 @@ class FTPModel(sa.Model):
     ftp_dirmask = sa.Column(sa.String(3), default="022")
     ftp_fxp = sa.Column(sa.Boolean(), default=False)
     ftp_resume = sa.Column(sa.Boolean(), default=False)
-    ftp_defaultroot = sa.Column(sa.Boolean(), default=False)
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/ftp.py
```diff
@@ -197,8 +197,13 @@ async def do_update(self, data):
         if new["onlyanonymous"]:
             if not new["anonpath"]:
                 verrors.add("ftp_update.anonpath", "This field is required for anonymous login")
-            else:
-                await check_path_resides_within_volume(verrors, self.middleware, "ftp_update.anonpath", new["anonpath"])
+        else:
+            # Anonymous is disabled, clear the anonpath
+            if new["anonpath"] is not None:
+                new["anonpath"] = None
```
bmeagherix, line 203: So if anonymous ftp access is turned off and then turned back on again, then the administrator **must** to re-select a path?  Personally that's not the behavior that I'd prefer, but I suppose an argument could be made either way.

bmeagherix (approved):

PR #11495, opened 2023-06-13
NAS-121605 Don't delete UEFI vars when updating VM data

MJDSys:
Change the VM manager to perserve the UEFI variables when updating the VM data in libvirt.  This adds a new argument to undefine_domain for when update_domain is updating the domain's xml to perserve the NVRAM contents.  Also change update_domain to move the NVRAM file when the domain is renamed, to preserve its contents.

This also changes undefine_domain to use libvirt's functionality to remove the NVRAM storage.


yocalebo (dismissed):

MJDSys:
I've fixed the flake8 error as well.

Qubad786 (approved):
Thank you @MJDSys for your contribution 👍

PR #11497, opened 2023-06-13
NAS-122240 / 13.1 / Network interface `link_address_b`

themylogin:


src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
```diff
@@ -890,6 +892,8 @@ def carp_backup(self, fobj, ifname, vhid, event, user_override):
                 # Sync GELI and/or ZFS encryption keys from MASTER node
                 self.middleware.call_sync('failover.sync_keys_from_remote_node')
 
+                self.run_call('failover.call_remote', 'interface.persist_link_addresses')
```
yocalebo, line 895: Other node will be running the old version of software on upgrade so this will crash. It can also crash if the other side is down. Need to catch this and ignore it so we don't break failover.

yocalebo (changes requested):

src/freenas/usr/local/lib/middlewared_truenas/plugins/failover_/event.py
yocalebo: Oh, nvm I see it's handled the function itself.

yocalebo (approved):

themylogin:
backport

themylogin:
time 12:00

PR #11499, opened 2023-06-13
NAS-121926 / 23.10 / Add DRAID support

Qubad786:
## Context

ZFS allows creating a vdev of `DRAID` type. Basically each draid type vdev has certain attributes associated with it namely:

1. parity
2. spares
3. explicitly specifying data disks

After discussion with Alexander Motin we decided to at least expose specifying parity when creating a draid type vdev similar to how raidz vdev works. With the current changes as required we can further allow specifying the other configuration options for draid type vdev if requested.

### DIFF
--- a/src/middlewared/middlewared/plugins/zfs_/pool_utils.py
+++ b/src/middlewared/middlewared/plugins/zfs_/pool_utils.py
@@ -20,9 +20,19 @@ def convert_topology(zfs, vdevs):
             topology[vdev['root'].lower()].extend(children)
         else:
             z_vdev = libzfs.ZFSVdev(zfs, 'disk')
-            z_vdev.type = vdev['type'].lower()
             z_vdev.children = children
-            topology[vdev['root'].lower()].append(z_vdev)
+            if vdev['type'].startswith('DRAID'):
+                z_vdev.type = 'draid'
+                topology['draid'].append({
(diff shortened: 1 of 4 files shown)


yocalebo (changes requested):
Our non-HA VMs have around 10 disks I believe so we can add an integration test here to ensure at least creating the pool works as well as verifying the topology is what we expect.

src/middlewared/middlewared/plugins/zfs_/pool_utils.py
```diff
@@ -20,9 +20,20 @@ def convert_topology(zfs, vdevs):
             topology[vdev['root'].lower()].extend(children)
         else:
             z_vdev = libzfs.ZFSVdev(zfs, 'disk')
-            z_vdev.type = vdev['type'].lower()
             z_vdev.children = children
-            topology[vdev['root'].lower()].append(z_vdev)
+            if vdev['type'].startswith('DRAID'):
+                vdev_type = vdev['type'][:-1].lower()
```
yocalebo, line 25: There is no reason to do `vdev_type = vdev['type'][:-1].lower()`. Just do `z_vdev.type = 'drain'` as the latter is much more readable.

yocalebo (approved):

PR #11508, opened 2023-06-14
NAS-122467 / 23.10 / fix > 1k skipped ci tests

yocalebo:
I sped up the full CI run by over 2hrs! But I did it in an unexpected way :(. The "pool_04" string has been hard-coded _everywhere_ in our test modules so change all the references of this string to "pool_name" which is set in "auto_config" module. This will fix the > 1k tests that are being skipped.

### DIFF
--- a/tests/api2/test_006_pool_and_sysds.py
+++ b/tests/api2/test_006_pool_and_sysds.py
@@ -40,7 +40,7 @@ def test_001_check_sysdataset_exists_on_boot_pool(ws_client):
     assert bp_basename == sysds['basename']
 
 
-@pytest.mark.dependency(name='PERM_ZPOOL_CREATED')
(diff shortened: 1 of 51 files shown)


themylogin:
Can we replace `pool_04` everywhere then?

yocalebo:
> Can we replace `pool_04` everywhere then?

I could but I'd have to touch a lot of files. I also was going to hard-code it to `pool_name` which is imported from auto_config so that if we ever changed the name it wouldn't break > 1k api tests. I ended up just going this route as a path of least resistance.

yocalebo:
> @yocalebo what's wrong with touching a lot of files?

1. takes too much of my time (I'm lazy)
2. introduces more possibilities of regressions

Regardless of these reasons that I've listed above, I went ahead and changed it 👍

PR #11512, opened 2023-06-14
NAS-122467 / 23.10 / Reduce pytest_dependency usage (pool_04 and ssh_password)

themylogin:
I sped up the full CI run by over 2hrs! But I did it in an unexpected way :(. The "pool_04" string has been hard-coded everywhere in our test modules. Investigating this a bit further, we’ve decided to abort the entire CI session if we fail to create the zpool. This allows us to remove the pytest_dependency dance that we have to perform (which is inefficient and doesn’t scale). We’re doing the same for the “ssh_password” dependency as well.

### DIFF
--- a/tests/api2/test_001_ssh.py
+++ b/tests/api2/test_001_ssh.py
@@ -33,7 +33,8 @@ def test_001_is_system_ready(ws_client):
     # will be non-deterministic because middleware plugins
     # internally expect that the system is ready before
     # propertly responding to REST/WS requests.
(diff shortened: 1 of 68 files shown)


themylogin:
https://ci.tn.ixsystems.net/jenkins/job/TrueNAS%20SCALE%20-%20Unstable/job/API%20Tests%20-%20TrueNAS%20SCALE%20with%20middleware%20reinstalled%20from%20a%20branch%20Incremental/932/

yocalebo (approved):

PR #11522, opened 2023-06-15
NAS-119900 / 23.10 / use new cython module for getting ses info

yocalebo:
We've added a new Cython module which uses the sgutils2-dev api for sending commands to various end-devices. We've only added SES capability at the moment since this is the problem that we need to immediately solve. The library was written in such a way that it returns the exact same string output that `sg_ses` does which allows very minimal changes to take advantage of the library.

This cuts down on the amount of `subprocess`'ing that we do on our particularly large enterprise systems.


yocalebo:
CI failures are expected since new module was added and the latest incremental doesn't have it yet.

src/middlewared/middlewared/plugins/enclosure.py
```diff
@@ -1038,21 +1039,27 @@ def device_slot_set(self, status):
...
-            p = subprocess.run(cmd, encoding="utf-8", errors="ignore", stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-            if p.returncode != 0:
-                logger.warning("%r failed: %s", cmd, p.stderr)
-                ok = False
+        if status == 'clear':
+            for getit, setit in (('get=fault', 'clear=fault'), ('get=ident', 'clear=ident')):
+                if self.sg3obj.set_control(slot, getit):
+                    # fault or identify light is toggled so try and clear it
+                    self.sg3obj.set_control(slot, setit)
+                    if self.sg3obj.set_control(slot, getit):
```
themylogin, line 1049: Does it return some error number we can also log?

src/middlewared/middlewared/plugins/enclosure.py
```diff
@@ -1038,21 +1039,27 @@ def device_slot_set(self, status):
...
+                        # it shows as being turned on so at least log a message
+                        # to try and give some idea of what's going on
+                        self.logger.warning('Failed to %s slot %r on %r', status, slot, self.enclosure.devname)
+                        ok = False
+        else:
+            cmds = (('get=ident', 'set=ident'),) if status == 'identify' else ((f'get={status}', f'set={status}'),)
+            for getit, setit in cmds:
+                if self.sg3obj.set_control(slot, getit):
+                    self.sg3obj.set_control(slot, setit)
+                    if self.sg3obj.set_control(slot, getit):
```
themylogin, line 1060: I am confused by this logic. We expect it to be "off", then set it to "on" and then log an error if it is "on"?

themylogin (approved):

anodos325:
Does libsg3 need to be added as a build / runtime dependency for middleware package now?

PR #11529, opened 2023-06-15
NAS-122493 / 13.0 / Add test for stream writes on directories

anodos325:


bmeagherix (approved):
BTW, I see you cleanup _streamstestdir_.  Is there a reason why _streamstestfile_ is not also cleaned up?

anodos325:
> BTW, I see you cleanup _streamstestdir_. Is there a reason why _streamstestfile_ is not also cleaned up?

Gets reused in later tests

PR #11533, opened 2023-06-15
NAS-122495 / 23.10 / download update file once on HA systems

themylogin:
On an HA system, download the update file once (on the active controller) and then send that file to the other controller via the heartbeat interface. This prevents the standby from downloading the same files off the network.


yocalebo (approved):

themylogin:
time 2:00

PR #11534, opened 2023-06-15
NAS-122504 / 23.10 / Add raise_connect_error option to failover.call_remote

themylogin:
Add `raise_connect_error` boolean option to `failover.call_remote`. If this is set to false, then it will not raise an exception of the remote call fails in a known way (i.e. standby node is rebooting/upgrading/heartbeat issue/etc). Defaults to true otherwise.


yocalebo (approved):

themylogin:
time 3:00

PR #11537, opened 2023-06-16
NAS-121930 / 23.10 / Add FIPS support

sonicaj:
This PR adds FIPS support to middleware allowing it to be user configurable where users can just enable/disable it as per their needs.


src/middlewared/middlewared/plugins/security/update.py
```diff
@@ -0,0 +1,70 @@
...
+    )
+    async def do_update(self, data):
+        """
+        Update System Security Service Configuration.
+
+        `enable_fips` when set, enables FIPS mode.
+        """
+        old = await self.config()
+        new = old.copy()
+        new.update(data)
```
yocalebo, line 48: If no data has changed, return early here.

src/middlewared/middlewared/plugins/security/update.py
```diff
@@ -0,0 +1,70 @@
...
+class SystemSecurityService(ConfigService):
+
+    class Config:
+        cli_namespace = 'system.security'
+        datastore = 'system.security'
+        namespace = 'system.security'
+
+    ENTRY = Dict(
+        'system_security_entry',
+        Bool('enable_fips', required=True),
```
yocalebo, line 29: Do we need to add "reboot" Boolean and "reboot_timeout" Int schema objects to automatically reboot system if they enable it? If we decide to do this, we need to handle HA systems.

sonicaj: I don't think we should reboot ourselves, we don't do the same when grub configuration is changed 🤔 
Probably we can make that clear in the UI

src/middlewared/middlewared/scripts/configure_fips.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def validate_system_state() -> None:
+    for path in (FIPS_MODULE_FILE, OPENSSL_CONFIG_FILE, OPENSSL_FIPS_FILE):
+        if not os.path.exists(path):
+            raise Exception(f'{path!r} does not exist')
+
+
+def modify_openssl_config(enable_fips: bool) -> None:
+    with open(OPENSSL_CONFIG_FILE, 'r') as f:
```
themylogin, line 23: Does anybody else modify `OPENSSL_CONFIG_FILE`? Why do we have to edit it instead of generating from scratch every time like we do with other config files?

tests/api2/test_fips.py
```diff
@@ -0,0 +1,17 @@
+import os
+import pytest
+import sys
+
+from middlewared.test.integration.utils import call
+
+
+sys.path.append(os.getcwd())
```
themylogin, line 8: Why?

themylogin (changes requested):

src/middlewared/middlewared/scripts/configure_fips.py
sonicaj: Will make the changes consuming it from `/conf/base` itself as discussed. Thanks!

themylogin (approved):

src/middlewared/middlewared/plugins/security/update.py
```diff
@@ -0,0 +1,73 @@
...
+        if new['enable_fips'] != old['enable_fips']:
+            # TODO: We likely need to do some SSH magic as well
+            #  let's investigate the exact configuration there
+            await self.middleware.call('etc.generate', 'fips')
+
+        return await self.config()
+
+    @private
+    async def fips_enabled(self):
+        cp = await run(['openssl', 'list', '-providers'], check=False)
```
yocalebo, line 69: I don't like fork+exec'ing in main event loop. Please make this a non-coroutine.

src/middlewared/middlewared/plugins/security/update.py
```diff
@@ -0,0 +1,73 @@
...
+        if cp.returncode:
+            raise CallError(f'Failed to determine if fips is enabled: {cp.stderr.decode()}')
+
+        return bool(RE_FIPS.search(cp.stdout.decode()))
```
yocalebo, line 73: Why do we need a regex and a boolean cast and decoding the returned output? Just do `return b'OpenSSL FIPS Provider' in cp.stdout`

yocalebo (changes requested):

yocalebo (approved):
Thanks!

PR #11540, opened 2023-06-16
NAS-122512 / 23.10 / Only send the necessary update file to the standby controller

themylogin:
Before these changes, this failed with the following error:
```
....
  File "/usr/lib/python3/dist-packages/middlewared/main.py", line 1425, in call_sync
    return methodobj(*prepared_call.args)
  File "/usr/lib/python3/dist-packages/middlewared/plugins/failover_/remote.py", line 269, in send_file
    self.CLIENT.send_file(token, src, dst)
  File "/usr/lib/python3/dist-packages/middlewared/plugins/failover_/remote.py", line 180, in send_file
    raise CallError(
middlewared.service_exception.CallError: [EFAULT] Failed to send /var/db/system/update/updatefile.sqsh to Standby Controller: [Errno 28] No space left on device.
```

This is because we were sending the downloaded update file and the manual update file to the remote path which is tmpfs. The remote path is on tmpfs on HA systems and so it ran out of space. This changes it to only send the necessary file to remote side.


yocalebo (approved):

themylogin:
time 4:00

PR #11541, opened 2023-06-16
NAS-122511 / 23.10 / Better progress report for failover upgrade

themylogin:
This improves the progress reporting when upgrading an HA system. Instead of showing 0% for the entire process, it'll update accordingly.


yocalebo (approved):

themylogin:
time 2:30

PR #11546, opened 2023-06-16
NAS-122513 / 23.10 / fix build by renaming python3-pysg3 to python3-sg3

yocalebo:
This package was using a non-standard package name. I've fixed that here: https://github.com/truenas/py-sg3/pull/2 while also fixing the typo I made here.


themylogin:
Can we fix the package name instead to match Debian naming conventions?

themylogin (approved):

PR #11559, opened 2023-06-21
NAS-122561 / 23.10 / Remove SSH credentials cipher field, but force FIPS-compliant cipher when FIPS compliance is enabled

themylogin:


src/middlewared/middlewared/plugins/zettarepl.py
```diff
@@ -862,12 +862,17 @@ async def _define_ssh_transport(self, credentials):
...
+        if (await self.middleware.call("system.security.config"))["enable_fips"]:
+            cipher = "fips"
+
         return {
             "hostname": credentials["attributes"]["host"],
             "port": credentials["attributes"]["port"],
             "username": credentials["attributes"]["username"],
             "private-key": key_pair["attributes"]["private_key"],
             "host-key": credentials["attributes"]["remote_host_key"],
+            "cipher": cipher,
```
themylogin, line 875: @yocalebo we never used this option value (by accident I think) and no one has complained. `none` cipher is still unsupported in SCALE and no one has ever tried to use it. Should we just remove this option altogether?

yocalebo: Yeah, just remove it.

yocalebo (approved):

yocalebo (approved):

themylogin:
time 2:00

PR #11567, opened 2023-06-22
NAS-122594 / 23.10 / deepcopy `@cache`d function return value in case caller modifies it

themylogin:
If the function that is being cached returns a mutable python object, then the caller of said function can alter the object. This will cause all other subsequent callers to receive the altered response.


yocalebo (approved):

themylogin:
time 2:00

PR #11572, opened 2023-06-23
NAS-122482 / 23.10 / Add SNMP tests that verify the response uses the same IP as request

bmeagherix:
Add SNMP tests that verify the response uses the same IP as request.

Use the new utility functions `async_SSH_start` and `async_SSH_done` to run a `tcpdump` on the TrueNAS system to capture a SNMP Get/Response (issued using `pysnmplib`).  Verify that the destination of the GET is the same as the source of the RESPONSE.

### DIFF
--- a/tests/api2/test_440_snmp.py
+++ b/tests/api2/test_440_snmp.py
@@ -1,21 +1,77 @@
 #!/usr/bin/env python3
 # License: BSD
 
-import pytest
-import sys
 import os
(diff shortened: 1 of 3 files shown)


bmeagherix:
Since the original defect was targeted to 13.0 we're likely going to backport there, and presumably to 13.1 too?

TBD whether we also want to backport to 22.12.x.  (No code changes were required for SCALE.)

PR #11573, opened 2023-06-25
NAS-119538 / 23.10 / Integrate Netdata in TrueNAS

Qubad786:
## Context

It was decided that we should add netdata to the product and phase out collectd/rrdcached.
This PR is a first iteration which focuses on adding following changes:

1. Integrate bare minimums for netdata by installing `netdata-core` and `netdata-plugins-python` packages only
2. Reporting section can be mainly divided in 2 aspects:
    a) Real time reporting
    b) Historical data reporting
    
    Real time reporting is planned to be completely backwards compatible with the metrics we are reporting currently so it has not been refactored much to consume netdata instead.
    
    Historical data reporting has some changes to what the collectd/rdcached API's return currently. The historical data section was divided in plugins with each plugin retrieving some aspect of the system information i.e cpu/memory etc. Plugins have been re-written to have parity with the ones we have already (ZFS plugins need to be re-visited to make sure we have the exact same metrics as with collectd and disk temperature too as `netdata` user cannot access the database - also nfsstats plugin is still left to be ported over to netdata).
    
    The current changes do not remove collectd/rrdcached but new endpoints have been added for retrieving historical reporting so that we can give some time for the netdata integration to mature.

3. Preserving historical reporting data
    Netdata uses `dbengine` by default which is a custom time series database to store historical data. There are different tiers at which data is stored. After discussion with Caleb, we initially want to save past 7 days of per second data.
    Based on this requirement, per second data come under `tier 0` of dbengine so the changes only enable 1 tier of the database and have a disk space consumption logic in place which approximately figures out how much space the metrics will take on disk for tier 0 and past 7 days (about 7 days, currently the changes use `graph_age` from `reporting.config` but we can discuss the best way forward for that as that field is used by collectd implementation and depicts the time in months).
    Netdata state is being stored inside system dataset so it can be preserved across boot environments.


yocalebo (approved):
I've tested this on a real system and since this is being installed side-by-side with existing API, I believe this is safe to merge. We can iterate and make improvements in separate PR's to keep the amount of changes smaller.

src/middlewared/middlewared/pytest/unit/plugins/disk/test_temperature.py
```diff
@@ -1,6 +1,6 @@
 import pytest
 
-from middlewared.plugins.disk_.temperature import get_temperature, DiskService
+from middlewared.utils.disks import parse_sata_or_sas_disk_temp, parse_smartctl_for_temperature_output
```
themylogin, line 3: Please also rename test files to match the modules they are testing

src/middlewared/middlewared/utils/cpu.py
```diff
@@ -0,0 +1,99 @@
...
+        for name, temps in value.items():
+            if not (m := RE_CORE.match(name)):
+                continue
+            for temp, value in temps.items():
+                if 'input' in temp:
+                    temperatures[chip][int(m.group(1))] = value
+                    break
+
+    return {
+        str(k): v for k, v in dict(enumerate(sum(
```
themylogin, line 61: Here you return string keys in AMD implementation you return integer keys

src/freenas/usr/lib/netdata/python.d/disktemp.chart.py
```diff
@@ -0,0 +1,36 @@
+from bases.FrameworkServices.SimpleService import SimpleService
+from copy import deepcopy
+
+from middlewared.utils.disks import get_disks_for_temperature_reading, get_disks_temperatures
```
themylogin, line 4: Why not use middleware client for that like we did with collectd plugins?

Qubad786: @themylogin netdata user cannot access the middleware client sadly..

src/middlewared/middlewared/utils/cpu.py
Qubad786: Good catch..fixing this asap

src/freenas/usr/lib/netdata/python.d/disktemp.chart.py
themylogin: @Qubad786 why?

themylogin (approved):
I do not approve querying temperatures outside of the middleware, but providing API to the UI team in time is more important now.

PR #11607, opened 2023-06-28
NAS-122696 / 23.10 / Use pyglfs bindings for permissions ops

anodos325:
This changes how permissions-related operations are handled on glusterfs volumes so that we use libgfapi directly rather than shelling out to iterate in glusterfs fuse mount.


anodos325:
This functionality already has test coverage via cluster_tests/tests/filesystem/test_001_base_filesystem_tests.py

yocalebo (approved):
Does kind of suck to list xattrs just to strip it but this looks good for now.

anodos325:
> Does kind of suck to list xattrs just to strip it but this looks good for now.

Yeah, I just have to make change to pyglfs before I can remove that logic.

PR #131, opened 2016-04-06
Bug 4247 syslog level

rohitkeshri1986:
Ticket: #4247.
Added "Syslog level" checkbox in GUI.
User menu selection will be used as a filter for logs that are sent to remote server with "Syslog server" defined IP.
This does not affect local logging.

PR #11616, opened 2023-07-03
NAS-122707 / 23.10 / Add iSCSI XCOPY unit tests

bmeagherix:
Add unit tests to exercise EXTENDED COPY (XCOPY) between pairs of iSCSI targets.

Because of the dynamic nature of the tests, it was discovered that the `scstadmin` component of `scst` was eliminating some extent LUNs that were auto-added to the copy manager.  This would **not** manifest if the targets were defined, and then `scst` was restarted, but it was a problem during the unit tests as two targets were added in order to exercise the XCOPY.

The simplest solution was to avoid the issue by explicitly writing an existing copy manager LUNs when rewriting the _scst.conf_

### DIFF
--- a/tests/api2/test_261_iscsi_cmd.py
+++ b/tests/api2/test_261_iscsi_cmd.py
@@ -49,6 +49,9 @@ except ImportError:
 skip_persistent_reservations = pytest.mark.skipif(not pyscsi_has_persistent_reservations,
                                                   reason="PYSCSI does not support persistent reservations")
 
+skip_extended_copy4 = pytest.mark.skipif(not hasattr(pyscsi.pyscsi.scsi.SCSI, 'extendedcopy4'),
+                                         reason="PYSCSI does not support extended copy (3/4)")
+
 skip_multi_initiator = pytest.mark.skipif(not initiator_name_supported(),
                                           reason="PYSCSI does not support persistent reservations")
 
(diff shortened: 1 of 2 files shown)


tests/api2/test_261_iscsi_cmd.py
```diff
@@ -49,6 +49,9 @@
 skip_persistent_reservations = pytest.mark.skipif(not pyscsi_has_persistent_reservations,
                                                   reason="PYSCSI does not support persistent reservations")
 
+skip_extended_copy4 = pytest.mark.skipif(not hasattr(pyscsi.pyscsi.scsi.SCSI, 'extendedcopy4'),
```
themylogin, line 52: Why do we have to support different versions of pyscsi? I don't like the idea of skipping tests because the Jenkins pipeline will be green despite some tests not being ran but I might be missing something.

bmeagherix: Jenkins **shouldn't** skip anything as I have recently added extendedcopy4 support to the pyscsi '[tester](https://github.com/truenas/python-scsi/tree/tester)' branch being used to run our tests.  However, despite raising a PR ([#131](https://github.com/python-scsi/python-scsi/pull/131)) against upstream several weeks ago I've had zero response so far.  (BTW I consulted @yocalebo & @anodos325 and verified that they agreed that I should proceed with the tests.)

So, the `skip_extended_copy4` is just a nod to the fact that this code is still _somewhat_ in flux.  I can certainly remove it.

Please let me know if you still think it should be removed after reading the above context.

yocalebo (approved):

PR #11622, opened 2023-07-05
NAS-122813 / 23.10 / Add more protection for JWT usage in cluster event

anodos325:
Initial implementation of JWT usage was to encode static dummy payload with pre-shared key. Presence of decodable header was used to authenticate the command / payload sent to the webhook. Since the encoded payload and secret were both static, the JWTOKEN used for all commands was identical.

This PR replaces the static dummy payload with a dict containing timestamp and uuid for message id. If clock slew from message we receive is more than 10 seconds or if the message id is reused, then we reject the event request.

### DIFF
--- a/src/middlewared/middlewared/plugins/gluster_linux/local_events.py
+++ b/src/middlewared/middlewared/plugins/gluster_linux/local_events.py
@@ -4,12 +4,14 @@ import jwt
 import enum
 import asyncio
 import os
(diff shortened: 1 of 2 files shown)


yocalebo (approved):

anodos325:
Next PR will encrypt transport for Cobia.

PR #11623, opened 2023-07-05
NAS-122875 / 23.10 / [vm] Be more specific about which PCI device/controllers are "sensitive"

ownaginatious:
(apologies if this isn't the right way to propose changes; I'm not familiar with the JIRA workflow and am external to iX systems)

I recently upgraded to TrueNAS 22.12.3.1 from 22.12.2.x and suddenly I was no longer able to pass an NVMe drive into one of my VMs as it's now considered "critical".

image

This is PCI device 0000:04:00.0.
```
$ lspci -nnD
...
0000:00:14.2 RAM memory [0500]: Intel Corporation Device [8086:7aa7] (rev 11)
...
0000:04:00.0 Non-Volatile memory controller [0108]: Phison Electronics Corporation PS5013 E13 NVMe Controller
...
```

I believe the catch-all for "memory" controller types was probably intended specifically for "RAM memory" PCI devices/controllers.

Fortunately, the existing code easily supports that with minimal changes (the regular expression already captures the full device/controller type). This PR simply specifies more specific sensitive device types to avoid accidentally catching those which aren't really sensitive at all.

### DIFF
--- a/src/middlewared/middlewared/utils/gpu.py
+++ b/src/middlewared/middlewared/utils/gpu.py
@@ -7,8 +7,9 @@ from middlewared.service_exception import CallError
 
 RE_PCI_ADDR = re.compile(r'(?P<domain>.*):(?P<bus>.*):(?P<slot>.*)\.')
 SENSITIVE_PCI_DEVICE_TYPES = (
-    'Bridge',
-    'memory',
+    'PCI Bridge',
+    'ISA Bridge',
+    'RAM memory',
     'SMBus',
 )
 


yocalebo:
This is great, thanks @ownaginatious. I definitely agree with you and I'm actually in this same portion of code trying to optimize another function so I do believe this change is correct. However, the machines that I have access to are not listing a "RAM Memory" device. Do you have an example of one that shows that?

ownaginatious:
@yocalebo thanks for taking a look!

The example output from `lspci -nnD` I posted has the RAM memory device/controller from my personal NAS server. Here are the specs, if it helps: 
- 13th Gen Intel(R) Core(TM) i5-13400 (4x32 GB RAM)

I also have another NUC ([this](https://www.amazon.com/gp/product/B0B8NLZVZ1/ref=ppx_yo_dt_b_search_asin_title?ie=UTF8&psc=1) specifically) not running TrueNAS, but it too shows a "RAM memory" device.

```
$ lspci -nnD
...
0000:00:14.2 RAM memory [0500]: Intel Corporation Device [8086:4def] (rev 01)
...
```

Let me know if you need anything else!

yocalebo (approved):
While working on https://github.com/truenas/middleware/pull/11644, I investigated these changes and I agree that this is being a little bit more specific with the "sensitive" devices. Thanks for the contribution!

PR #11635, opened 2023-07-06
NAS-122831 / 23.10 / Fix rsync tests

themylogin:


sonicaj (approved):

yocalebo (commented):
I think we should just remove this test since it's not actually testing anything of value here. It's only testing that basic CRUD operations work and doesn't test that the rsync task actually copies the data properly. What do you think?

themylogin:
@yocalebo CRUD operations can also fail so they are of value too. We can add testing rsync operation itself later.

themylogin:
time 0:30

PR #11642, opened 2023-07-07
NAS-122859 / 23.10 / Improve schema handling for private items

anodos325:
OROperator lacked has_private() method which broke dumping values in some edge cases.

Allow dumping of Lists where single item schema is set for the list and the schema contains private entries. If we don't do this then our job log is relatively useless.


themylogin (commented):
Se can have a unit test for this

sonicaj (approved):

yocalebo (changes requested):
flake8 and unit tests?

yocalebo (approved):

PR #11643, opened 2023-07-07
NAS-122919 / 23.10 / Improve ctdb private ips

anodos325:
Improve CTDB private IPs handling

This PR makes several critical changes to how the ctdb nodes
file (private IPs) are managed in middleware. Some types of
issues that are addressed by this PR are:

* Inconsistencies in nodes configuration file on different cluster nodes
* Duplicate nodes entries being added via ctdb.shared.volume.create
  (e.g. same node twice with different IPs)
* Lack of clear mapping between ctdb nodes and gluster peers

Originally there was easy mapping of entries in the ctdb nodes
file back to the originating gluster peer. This PR changes the
nodes files entry from form of
```
<ipaddress>
<ipaddress>
<ipaddress>
```

to
```
<ipaddress>#{<peer information>}
<ipaddress>#{<peer information>}
```

Above required a change to nodes file parsing function in ctdbd,
which means this PR may not be backported into stable/bluefin.

Currently only peer UUID is stored in the file, but in future
we can potentially expand to include additional information.
Store gluster node UUID in ctdb nodes file. This allows us to
more tightly couple gluster TSP configuration with our CTDB nodes
configuration (e.g. include peer UUID in ctdb.private.ips.query
return).

Since this change requires that the backend maintain the mapping
between nodes and peers, the ctdb private ips API was changed to
require caller to provide a gluster peer UUID for the nodes entry.

The ctdb.shared.volume.create method originally would automatically
append nodes file entries for missing gluster peers based on DNS
lookup results, but this has over time proven to be somewhat less
than reliable (users may have DNS misconfigurations that result
in multiple nodes files entries or incorrect interfaces being used).

This PR allows caller of ctdb.shared.volume.create to optionally
include a list of private address + gluster peer UUID mappings
to be added (if necessary) to the nodes file. This method will now
explicitly fail in the following situations:

* gluster peers that are not present in resulting file
* peer UUIDs in payload that do not exist
* entries in nodes file that do not map back to gluster UUID

This has somewhat wide-ranging impact for how our backend APIs are
used. For instance, if a new gluster peer is added to an existing
TSP, then a private IP for the nodes file must also be supplied.

A new cluster management service is also being added via this
pull request to enforce consistency in how SCALE clusters are
configured (and ensure that proper validation takes place to prevent
issue reports on drastically misconfigured clutsers). It also
provides a stub of an API for adding new cluster nodes (expansion).

The end-goal of this API is to force the caller to provide us with
a payload containing gluster peer, brick, and private address (nodes
file) information for each proposed cluster node.


src/middlewared/middlewared/plugins/cluster_linux/ctdb_shared_volume.py
```diff
@@ -456,25 +458,43 @@ async def create(self, job):
...
+        private_ips_nodes_uuid = set([x['node_uuid'] for x in priv_ctdb_ips])
+        gluster_peers_uuid = set([x['uuid'] for x in gluster_peers])
+
+        if not ip_info and not priv_ctdb_ips:
+            raise CallError(
+                'No private IPs were specified in payload and no private IPs are currently '
+                'set within the CTDB system directory.'
+            )
+        elif not ip_info:
+            private_ips_nodes_uuid = set([x['node_uuid'] for x in priv_ctdb_ips])
```
yocalebo, line 475: Lines 475 and 476 can be removed since you're generating the same list on lines 466 and 467 respectively.

cluster-tests/tests/directoryservices/test_001_activedirectory.py
```diff
@@ -286,29 +286,24 @@ def test_010_xattrs_writable_via_smb(request):
 
 def test_030_test_share_failover(request):
     depends(request, ['DS_SMB_SHARE_IS_WRITABLE'])
+    # temporarily disabled while refactoring test code
```
yocalebo, line 289: What's going on here?

cluster-tests/tests/interfaces/test_001_base_interface_tests.py
```diff
@@ -28,10 +28,12 @@ def node_by_public_address(addr):
     res = None
 
     for private_ip, data in ifaces_nodemap.items():
-        if addr not in data['active_interfaces']:
+        if addr not in data.get('active_interfaces', []):
             continue
 
-        res = (private_ip, data)
+        #cluster_ip = CLUSTER_IPS[PRIVATE_IPS.index(private_ip)]
```
yocalebo, line 34: What's going on here?

src/middlewared/middlewared/plugins/cluster_linux/ctdb_root_dir.py
```diff
@@ -14,44 +28,595 @@ class Config:
...
+                    continue
+
+                peer_uuid = file_name.strip('_UPDATE_IN_PROGRESS')
+                wait_for_nodes.append(peer_uuid[1:])
+
+            if not wait_for_nodes:
+                break
+
+            job.set_progress(80, f'Waiting for peers [{", ".join(wait_for_nodes)}] to finish update.')
+            sleep(10)
```
yocalebo, line 392: Do we need to wait 10 seconds each iteration or can we check at quicker intervals?

yocalebo (changes requested):

cluster-tests/tests/directoryservices/test_001_activedirectory.py
anodos325: Initially refactored to use different network for cluster backend traffic and didn't fully back-out changes when I realized it might cause problems for releng.

src/middlewared/middlewared/plugins/cluster_linux/ctdb_root_dir.py
anodos325: I don't mind doing 1 second intervals

PR #11644, opened 2023-07-07
NAS-122865 / 23.10 / fix vm.device.passthrough_device_choices

yocalebo:
This fixes many numerous problems. First and foremost, this method was subprocessing N * 2 times. "N" being the total number of pci devices on the system. This, obviously, becomes painful on certain systems rather quickly. Furthermore, this was doing xml parsing and calling blocking I/O in the main event loop. On a system with 331 pci devices, this takes ~0.22 seconds and without my changes it was taking ~32.22 seconds. This is roughly ~197% faster to run. The returned data is the exact same with the exception that my changes are actually pulling a bit more verbose information for the vendor/product etc. (This doesn't matter since it's shown to end-user and not actually depended upon).

Fixes employed:
- stop sub processing for pci devices and use direct db bindings to udev database
- stop parsing xml in the main event loop and, better yet, get rid of xml entirely (it's unneeded)
- stop calling blocking I/O in the main event loop
- simplify the logic as best as possible while maintaining as much performance as possible
- add basic unit test

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/pci.py
+++ b/src/middlewared/middlewared/plugins/vm/pci.py
@@ -1,16 +1,13 @@
+import collections
 import os
+import pathlib
 import re
 
 from pyudev import Context
-from xml.etree import ElementTree as etree
 
 from middlewared.schema import accepts, Bool, Dict, List, Ref, returns, Str
-from middlewared.service import CallError, private, Service
-from middlewared.utils import run
+from middlewared.service import private, Service
 from middlewared.utils.gpu import SENSITIVE_PCI_DEVICE_TYPES
-
-from .utils import get_virsh_command_args
-
 
 RE_DEVICE_PATH = re.compile(r'pci_(\w+)_(\w+)_(\w+)_(\w+)')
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/vm/pci.py
```diff
@@ -70,109 +143,24 @@ def retrieve_node_information(self, xml):
...
-        # e.g 86:00.0 3D Controller: Nvidia corporation gp100GL []
-        # We say string before ":" is prefix and the one after is suffix
-        prefix = str(pci_id) + (f' {controller_type!r}' if controller_type else '')
-        vendor = (node_info['capability'].get('vendor') or '').strip()
-        suffix = (node_info['capability'].get('product') or '').strip()
-        if vendor:
-            suffix = f'{suffix} by {vendor!r}'
-        return f'{prefix}: {suffix}' if suffix else prefix
+        return await self.middleware.run_in_thread(
+            self.get_single_pci_device_details, RE_DEVICE_PATH.sub(r'\1:\2:\3.\4', device)
```
themylogin, line 153: This method can easily be made synchronous

src/middlewared/middlewared/plugins/vm/pci.py
```diff
@@ -70,109 +143,24 @@ def retrieve_node_information(self, xml):
...
-        return mapping
-
-    @private
-    def get_pci_devices(self):
-        return {
-            i.sys_name: i.properties.get('ID_PCI_SUBCLASS_FROM_DATABASE', 'Unassigned class')
-            for i in Context().list_devices(subsystem='pci')
-        }
+        """Available choices for PCI passthru devices"""
+        return await self.middleware.run_in_thread(self.get_all_pci_devices_details)
```
themylogin, line 160: And this one too

src/middlewared/middlewared/plugins/vm/pci.py
```diff
@@ -27,28 +24,104 @@ async def iommu_enabled(self):
...
+                    'function': f'0x{func}',
+                })
+                final[i.name] = {'number': iommu_group, 'addresses': addresses[iommu_group]}
+        except FileNotFoundError:
+            pass
+
+        return final
+
+    @private
+    def get_pci_device_details(self, obj, iommu_info):
```
themylogin, line 51: This can be easily unit-tested

yocalebo:
@themylogin added unit tests

src/middlewared/middlewared/plugins/vm/pci.py
```diff
@@ -70,109 +143,22 @@ def retrieve_node_information(self, xml):
         Bool('reset_mechanism_defined', required=True),
         Str('error', null=True, required=True),
         Str('device_path', null=True, required=True),
+        Str('description', empty=True, required=True),
         register=True,
     ))
-    async def passthrough_device(self, device):
```
Qubad786, line 75: Can we add `controller_type` above to `returns` schema please?
Also with the new changes `controller_type` can be `null` - this is for my case where the old implementation reported `Unassigned class` - i believe it can be `null` and we should just reflect that properly?
```
{
  "capability": {
    "class": "0x130000",
    "domain": "0",
    "bus": "46",
    "slot": "0",
    "function": "0",
    "product": "Starship/Matisse PCIe Dummy Function",
    "vendor": "Advanced Micro Devices, Inc. [AMD]"
  },
  "iommu_group": {
    "number": 30,
    "addresses": [
      {
        "domain": "0x0000",
        "bus": "0x2e",
        "slot": "0x00",
        "function": "0x0"
      }
    ]
  },
  "controller_type": "Unassigned class",
  "critical": false,
  "device_path": "/sys/bus/pci/devices/0000:2e:00.0",
  "drivers": [],
  "available": true,
  "error": null,
  "reset_mechanism_defined": true,
  "description": "0000:2e:00.0 'Unassigned class': Starship/Matisse PCIe Dummy Function by 'Advanced Micro Devices, Inc. [AMD]'"
}
```

This is from the older implementation ^^^

themylogin (approved):

PR #11652, opened 2023-07-12
Separately check home_mode for immutable user

anodos325:
This fixes KeyError when user submits payload that includes home_mode for immutable user.


themylogin:
We can have an integration test for this

themylogin (approved):
Thank you!

PR #11655, opened 2023-07-12
NAS-122974 / 23.10 / Shift automation workloads to dedicated AD OU

anodos325:
Only install new computer objects in a dedicated OU in our AD domain during CI runs. This helps us to keep track of what was added via CI and what has been manually added (as well as simplifying permissions management).


anodos325:
depends on related change in jenkinsfile

yocalebo (approved):

PR #11657, opened 2023-07-13
NAS-119331 / 23.10 / On bridge update, do not put its member down if it already was a member of the bridge before update.

themylogin:
This bug was introduced by 97ff05a74845dacbd2974ef8c03a846e060986aa


yocalebo (approved):

themylogin:
time 4:00

PR #11661, opened 2023-07-14
NAS-122697 / 23.10 / prevent crash when users doing things via the cli

yocalebo:
User ssh'ed into a box and tried to "dd" a zvol to another zvol (to copy it) but made a typo in the command. This lead to dd creating a regular file in /dev/zvol directory. When we tried to "readlink" this file, it crashed with "Invalid Argument". This prevented the webUI from showing zvol choices in the VM creation window. To fix this, simply skip files that we crash on when trying to "readlink"


anodos325 (approved):

themylogin:
It’s easy to write an integration test for this

PR #11668, opened 2023-07-14
NAS-123054 / 23.10 / Use preshared secret for localevents for clpwenc

anodos325:
Since ctdb_shared_vol is not longer being used for ctdb configuration, shift our clwpenc plugin to use localevents token for our secret seed.

This reduces risk of exposure via glusterfs clients and also makes removes dependency on health of gluster volume.

### DIFF
--- a/src/middlewared/middlewared/plugins/pwenc.py
+++ b/src/middlewared/middlewared/plugins/pwenc.py
@@ -130,7 +130,7 @@ class CLPWEncService(PWEncService):
 
     def _read_secret(self):
         with open(self.secret_path, 'r', opener=self._secret_opener) as f:
-            self.secret = bytes.from_hex(f.read())
+            self.secret = bytes.fromhex(f.read())
 
     def encrypt(self, data):
         return encrypt(data, True)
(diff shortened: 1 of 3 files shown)


yocalebo (approved):

src/middlewared/middlewared/plugins/pwenc.py
```diff
@@ -120,101 +115,22 @@ class Config:
...
-                    raise
-                check_file = secret_dir.create('.check_file', os.O_RDWR)
+    def _reset_passwords(self):
+        raise NotImplementedError
 
-            with lock_file_open(check_file, os.O_RDWR) as fd:
-                fd.ftruncate(0)
-                fd.pwrite(self.encrypt(PWENC_CHECK), 0)
+    def _read_secret(self):
+        with open(self.secret_path, 'r', opener=self._secret_opener) as f:
```
yocalebo, line 132: This will crash if file doesn't exist. Is that expected?

anodos325: Not a hot code path and we probably want crash here (since will only happen if JWT is missing).

PR #11674, opened 2023-07-15
NAS-123038 / 23.10 / Remove novnc from SCALE

sonicaj:
This PR adds changes to remove `novnc` package from SCALE. We are seeing some CVE's wrt `nodejs` which are being brought in by `novnc` package and it has been decided that we remove `VNC` display type and just have `SPICE`. A migration has been added which will convert VNC display device to SPICE display device but if there are multiple display devices, we remove VNC display device as we can't have multiple SPICE display devices in a single VM.


sonicaj:
retest this please

yocalebo (approved):

PR #11675, opened 2023-07-15
NAS-121071 / 23.10 / add gssproxy support

mgrimesix:
### Add support for large kerberos tickets.

Add gssproxy package.


src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -166,6 +172,24 @@ async def do_update(self, data):
...
         await self.middleware.call('etc.generate', 'kerberos')
+
+        # gssproxy provides support for large kerberos tickets.
+        # It does not replace the kernel gssapi. The kernel will use gssproxy
+        # if it is available.  This is determined during the boot process.
+        # Currently, the status of gssproxy cannot be changed outside of boot time.
+        if new['enable_gssproxy'] != old['enable_gssproxy']:
+            # gssproxy seems to ignore the 'disable' setting and requires masking
+            if new['enable_gssproxy'] is True:
+                await systemd_unit("gssproxy", "unmask")
```
themylogin, line 183: There are various places all over our codebase where systemd service enable/disable is handled:
* etc_files/systemd (middleware)
* config upload (middleware)
* installer code (scale-build)

themylogin (changes requested):

src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -166,6 +172,24 @@ async def do_update(self, data):
...
+                await systemd_unit("gssproxy", "enable")
+                # Reboot is required to enact the change
+                self.logger.info("Processed gssproxy enable. Reboot is required to enact change.")
+            else:
+                await systemd_unit("gssproxy", "disable")
+                await systemd_unit("gssproxy", "mask")
+                # Reboot is required to enact the change
+                self.logger.info("Processed gssproxy disable. Reboot is required to enact change.")
```
anodos325, line 191: I think we should also add a "reboot required" FilePresenceAlert so that it gets visibility with end-user. c.f. UpdateFailedAlertSource

You'll probably want to put the sentinel file in the middleware rundir (tmpfs) so that it gets auto-cleared on reboot.

themylogin: Or make it a one-shot alert that is removed on boot, more modern approach 🙂

anodos325: Okay. I pointed Mark at some examples of one-shot alerts. @themylogin do you think it makes sense to create the alert on system update if we're not automatically rebooting?

themylogin: @anodos325 how is system update involved in this?

anodos325: Another situation where we might want to generate an alert "You should reboot your server" :)

anodos325:
Revised version looks good. Don't forget to fix the PR comment to reflect what you're doing.

src/middlewared/middlewared/plugins/kerberos.py
mgrimesix: removed enable/disable capability

src/middlewared/middlewared/plugins/kerberos.py
mgrimesix: Removed enable/disable capability

mgrimesix:
Simplified PR to install of gssproxy only.

themylogin (approved):

PR #11693, opened 2023-07-17
NAS-122666 / 23.10 / Use netdata for realtime event source

Qubad786:
## Context

Changes have been made to use netdata instead of the custom implementation we have had to retrieve realtime event statistics.
It has been ensured that we stay backwards compatible so as to not break existing consumers by ensuring the units and values we retrieve reflect what we were retrieving before.


yocalebo (approved):
Looks really clean. Great job! I'll approve this for now so we can meet the freeze window. I'll open a separate ticket for adding unit tests for the new standalone functions you've added.

PR #11695, opened 2023-07-18
NAS-123073 / 23.10 / Also catch self-emited `CallError` when `raise_connect_error` is spec…

themylogin:
…ified


themylogin:
time 0:20

PR #11698, opened 2023-07-18
NAS-123088 / 23.10 / Use  flake8-builtins

themylogin:
It will say if
```
+src/middlewared/middlewared/role.py:71:1: A001 variable "id" is shadowing a python builtin
```


yocalebo (approved):

themylogin:
time 1:00

PR #11701, opened 2023-07-19
NAS-123089 / 23.10 / Do not spam upsd errors

sonicaj:
This commit adds changes to not spam middleware logs again and again if the error has already been logged once.

### DIFF
--- a/src/middlewared/middlewared/plugins/ups.py
+++ b/src/middlewared/middlewared/plugins/ups.py
@@ -75,6 +75,7 @@ class UPSService(SystemServiceService):
         Str('shutdowncmd', null=True, required=True),
         Str('complete_identifier', required=True),
     )
+    LOGGED_ERRORS = []
 
     class Config:
         datastore = 'services.ups'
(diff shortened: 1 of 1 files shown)


bmeagherix:
I'm writing this as a comment / **_possible_** suggestion as the current PR code looks fine, but I'm wondering about the behavior.

Agree that spamming needs to be avoided, but I'm wondering whether silencing altogether after the first log might mean that the event gets overlooked.  If you disagree, then please disregard the rest of this comment.

A couple of other options might be to either maintain a count of how often the even occurs and log every e.g. 100 (or ???) occurrences.  Yielding something like

```
[2023/07/17 09:28:46] (ERROR) UPSService.upssched_event():288 - Failed to retrieve ups information: Error: Connection failure: Connection refused
[2023/07/18 04:28:46] (ERROR) UPSService.upssched_event():288 - Failed to retrieve ups information: Repeated 100 times.  Error: Connection failure: Connection refused
[2023/07/19 06:18:46] (ERROR) UPSService.upssched_event():288 - Failed to retrieve ups information: Repeated 200 times.  Error: Connection failure: Connection refused
```

Alternately, could store a timestamp when logged, and only log at most e.g. once a day.

(Or store both: LOGGED_ERRORS could be a dict keyed by error string, with value dict {'lastlog': ..., 'count': ...})

There's likely many other options to reduce the spam without making the issue disappear.

Anyway, just my 2 cents.

mgrimesix:
An 'Alert' might be the best solution for this sort of reporting.

sonicaj:
@mgrimesix we already have alerts for UPS service - the exact stderr will not i think help the user much, he already has alerts when ups status changes. Do let me know if you think we should add some more metadata to the logging statement and i can add appropriate changes.

mgrimesix:
If alerts are already generated on these events, then yes, we should squash repeating messages.  Even slow trickle messages will over time cause important messages to rotate out.

bmeagherix (approved):
> we already have alerts for UPS service

In that case my other concern is not valid.  Thanks.

PR #11708, opened 2023-07-20
Silence `Encoding detection: ascii is most likely the one.`

themylogin:


anodos325:
Wonderful. That was beginning to annoy me :)

PR #11713, opened 2023-07-20
NAS-123024 / 23.10 / FTP service does not automatically start upon system reboot

mgrimesix:
FTP service was not starting after reboot despite the service being enabled.

The source of the problem was that `proftpd.socket` was also enabled.  This is not correct or supposed to happen.  The default state is for `proftpd.socket` to be disabled. 

The fix is to disable `proftpd.socket` in our `postinst` module.

Also added a CI test to detect regression.

### DIFF
--- a/tests/api2/test_200_ftp.py
+++ b/tests/api2/test_200_ftp.py
@@ -1423,6 +1423,7 @@ class TestFTPSUser(UserTests):
             assert login_error is None
 
 
+@pytest.mark.skip(reason="Enable this when Jenkins infrastructure is better able to handle this test")
 def test_085_ftp_service_starts_after_reboot():
     '''
     NAS-123024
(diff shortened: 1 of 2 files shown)


anodos325:
Please fix PR title.

tests/api2/test_200_ftp.py
```diff
@@ -1396,6 +1423,76 @@ def setup(self, request):
...
 
+def test_085_ftp_service_starts_after_reboot():
+    '''
+    NAS-123024
+    There is a bug in the Debian Bookwork proftpd install package
+    that enables proftpd.socket which blocks proftpd.service from starting.
+
+    We fixed this by disabling proftpd.socket. There is a different fix
+    in a Bookworm update that involves refactoring the systemd unit files.
+    '''
```
anodos325, line 1434: I like this test in principle, but we probably need it disabled for now in our jenkins pipeline until we can get shifted to better infrastructure (it will be disruptive and slow things down).

anodos325 (approved):

PR #11717, opened 2023-07-21
NAS-123148 / 23.10 / Use directories rather than mountpoints for bricks

anodos325:
Attempts to access ZFS ctldir via libgfapi can cause gluster volume mount to freeze. We should prevent this variant of foot-shooting by adding robust checks to ensure that only directories within ZFS datasets are used. Since we already have filesystem.stat output as part of brick validation we can easily expand existence check to also cover things like fstype, whether the brick path is empty, dataset mount options, etc.


src/middlewared/middlewared/plugins/cluster_linux/management.py
```diff
@@ -80,6 +84,63 @@ def __validate_brick_path(self, schema_name, verrors):
...
+            )
+            return
+
+        if stat_info['is_mountpoint']:
+            verrors.add(
+                f'{schema_name}.brick_path',
+                f'{self.brick_path}: path is mountpoint on host {self.hostname}.'
+            )
+
+        if stat_info['nlink'] != 2:
```
yocalebo, line 103: You could put this as an `or` operand on line 97 to remove code duplication
```
if stat_info['is_mountpoint'] or stat_info['link'] != 2

anodos325: Derp. I meant to remove that.

yocalebo (approved):

PR #11723, opened 2023-07-24
NAS-123187 / 23.10 / add system.release_notes_url

yocalebo:
This adds `system.release_notes_url` endpoint that returns the public documentation link based on the major version of the SCALE version that is running. This will be used by webUI to display the URL to end-user.

### DIFF
--- a/src/middlewared/middlewared/plugins/system/product.py
+++ b/src/middlewared/middlewared/plugins/system/product.py
@@ -96,6 +96,12 @@ class SystemService(Service):
         return '.'.join(self.version_short().split('-')[0].split('.', 2)[0:2])
 
     @accepts()
+    @returns(Str('truenas_release_notes_url'))
+    def release_notes_url(self):
+        """Returns the release notes URL for the SCALE version that is currently installed"""
+        return f'https://truenas.com/docs/scale/{self.version_major()}'
+
+    @accepts()
     @returns(Str('truenas_version'))
     def version(self):
         """Returns the full name of the software version of the system."""


src/middlewared/middlewared/plugins/system/product.py
```diff
@@ -95,6 +95,11 @@ def version_major(self):
         # major is "23.10"
         return '.'.join(self.version_short().split('-')[0].split('.', 2)[0:2])
 
+    @accepts()
+    @returns(Str('truenase_release_notes_url'))
+    def release_notes_url(self):
```
Qubad786, line 100: Missing docstring

yocalebo: oops, thanks. Fixed!

src/middlewared/middlewared/plugins/system/product.py
```diff
@@ -98,6 +98,7 @@ def version_major(self):
     @accepts()
     @returns(Str('truenase_release_notes_url'))
```
bmeagherix, line 99: truenase ?

bmeagherix (changes requested):

src/middlewared/middlewared/plugins/system/product.py
yocalebo: yeah! it's the new "hip" way of spelling 😉 ....thanks fixed it

src/middlewared/middlewared/plugins/system/product.py
```diff
@@ -95,6 +95,12 @@ def version_major(self):
         # major is "23.10"
         return '.'.join(self.version_short().split('-')[0].split('.', 2)[0:2])
 
+    @accepts()
+    @returns(Str('truenas_release_notes_url'))
+    def release_notes_url(self):
+        """Returns the release notes URL for the SCALE version that is currently installed"""
+        return f'https://truenas.com/docs/scale/{self.version_major()}'
```
bmeagherix, line 102: Not sure if we're planning to change the URL to match the code, but currently it is
https://www.truenas.com/docs/scale/scale22.12/

(so and extra "scale" in there, as IIRC `version_major` will just return "22.12")

yocalebo: No the documentation team is setting it up differently starting in Cobia. The link will be just /23.10

bmeagherix (approved):

PR #11724, opened 2023-07-24
NAS-108381 / 23.10 / Add support for clustered SMB accounts

anodos325:
This commit adds basic support for clustered SMB accounts. Address space is carved out of the high end of what is supportable for idmaps in general. Validation for AD idmap settings is tweaked to prevent users from carving into this range.

The accounts have are disabled via shadow file, but have a valid NT hash for SMB access. This allows them to be used in situations where clustering user needs non-AD clustered SMB access, a primary example of where this may be useful is for micro-benchmarking cluster performance in an automated manner. The accounts are not valid for SSH, shell, or API access.

CI tests are also added to validate that clustered users can access SMB shares via any node's public IPs and that they cannot access shares if they have been administratively disabled (locked).


cluster-tests/tests/smb/test_002_smb_accounts.py
```diff
@@ -0,0 +1,282 @@
...
+    assert res.status_code == 200, res.text
+    share = res.json()[0]
+    assert share['cluster_volname'] == CLUSTER_INFO["GLUSTER_VOLUME"], str(share)
+    assert share['name'] == 'CL_SMB2', str(share)
+    assert share['path_local'] == SHARE_FUSE_PATH, res.text
+
+
+@pytest.mark.dependency(name="SMB_SERVICE2_STARTED")
+@pytest.mark.parametrize('ip', CLUSTER_IPS)
+def test_004_start_smb_service(ip, request):
```
mgrimesix, line 56: Isn't the start of SMB a cluster-wide event?  e.g. make the call to one node, then confirm the state on all nodes.

cluster-tests/tests/smb/test_002_smb_accounts.py
```diff
@@ -0,0 +1,282 @@
...
+        'msg': 'method',
+        'method': 'cluster.accounts.group.query',
+        'params': []
+    })
+    assert res.get('error') is None, str(res['error'])
+
+    groups = res['result']
+    # We should only have one group left (primary groups of users should be deleted)
+    assert len(groups) == 1, str(groups)
+
```
mgrimesix, line 224: Maybe `assert groups[0]['group'] == 'clustered_users'` ?

cluster-tests/tests/smb/test_002_smb_accounts.py
```diff
@@ -0,0 +1,282 @@
...
+    assert res.status_code == 200, res.text
+
+    assert len(res.json()['gr_mem']) == 5, res.text
+
+    for i in range(0, 5):
+        payload = {'username': f'clustered_user_{i}'}
+        url = f'http://{ip}/api/v2.0/user/get_user_obj/'
+        res = make_request('post', url, data=payload)
+        assert res.status_code == 200, res.text
+
```
mgrimesix, line 136: Confirm the UIDs are in the specified range?

cluster-tests/tests/smb/test_002_smb_accounts.py
anodos325: No. It's not cluster-wide with this particular API. If I were enabling via `ctdb.services`, then it would be, but this is quick-and-dirty test.

anodos325: Simpler to just start and stop by separately hitting all nodes.

yocalebo (approved):

PR #11725, opened 2023-07-25
Remove collectd and rrdcached

themylogin:
This does not remove any API, only removed the services themselves. Let's remove API after the new reporting API is merged (I don't want to introduce any merge conflicts for Rehan)


yocalebo (approved):
Fantastic!

PR #11726, opened 2023-07-25
Do not show feature disabled alert for pool status as we have PoolUpg…

themylogin:
…radedAlertClass for that


yocalebo (approved):

themylogin:
time 1:30

PR #11727, opened 2023-07-25
NAS-122774 / 23.10 / Use netdata for retrieving cpu temperatures

Qubad786:
## Context

Netdata cpu temperature plugin has been updated to consume the library/logic netdata uses to retrieve cpu temperatures.

Reasoning for using our own implementation of cpu temperatures is for 2 reasons:
1. Only retrieve metrics we are interested in as otherwise sensors retrieve lots of other metrics which will consume space unnecessarily
2. `sensors` netdata plugin retrieves temperatures under different charts which makes it harder to normalize/properly retrieve when trying to get historical cpu temperature statistics.


yocalebo:
retest this please

yocalebo (approved):

PR #11744, opened 2023-07-27
NAS-116349 / 23.10 / Revert "Do not send events for transient jobs"

yocalebo:
This reverts commit 66442f016bfb62c6e57db2d35566b1cb6e3620f4.

This allows a deadlock to occur when using our websocket client and calling a method with the `job=True` argument. Since no events are emitted for the job, the websocket client will wait indefinitely but since no events are emitted it will never complete.

A subsequent change was made in https://github.com/truenas/middleware/pull/10714 which exposes the `transient` flag in `core.get_jobs` which is what the webUI team should use to filter jobs from the task UI.

### DIFF
--- a/src/middlewared/middlewared/job.py
+++ b/src/middlewared/middlewared/job.py
@@ -96,7 +96,8 @@ class JobsQueue(object):
         self.deque.add(job)
         self.queue.append(job)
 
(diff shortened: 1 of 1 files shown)


yocalebo:
Used wrong branch name. Will re-open with proper name.

PR #11747, opened 2023-07-27
NAS-122665 / 23.10 / Properly calculate netdata disk usage

Qubad786:
## Context

To properly calculate how much space netdata database is going to consume we now make note of all the stats which we are retrieving at per second interval for 7 days and based on that we calculate how much disk space will be required.


yocalebo (changes requested):
get_disks is expensive on large systems. I made a "serial_only" option which is faster for situations like this.

yocalebo (approved):

PR #11755, opened 2023-07-31
NAS-123000 / 24.04 / SATA: Wait for udev events to settle before setting max sector count

ixhamza:
When a process opens a device for a write operation and then closes it, udevd synthesizes a change event, which results in several ATA commands. The ATA_CMD_SET_MAX_EXT command requires ATA_CMD_READ_NATIVE_MAX_EXT to be executed exactly as the previous command. sginfo operations open the device with O_RDWR. Setting sense data before setting max sectors causes udevd to inject an ATA command between ATA_CMD_READ_NATIVE_MAX_EXT and ATA_CMD_SET_MAX_EXT, leading to the eventual failure of ATA_CMD_SET_MAX_EXT. To fix this issue, we should wait for the udev event queue to drain before setting max sectors.

Validated on both Bluefin and Cobia by Jeff Ervine.

### DIFF
--- a/src/freenas/usr/local/sbin/disk_resize
+++ b/src/freenas/usr/local/sbin/disk_resize
@@ -1,5 +1,6 @@
 #!/bin/sh
 
+udevd_timeout=10
 dev=$1
 osize=$2
 
@@ -120,6 +121,11 @@ ata|sata)
 
 	if [ -z "${size}" ]; then
 		size=${maxsect}
+	fi
+	echo "Waiting up to ${udevd_timeout} seconds for udevd events to settle"
+	udevadm settle -t ${udevd_timeout}
+	if [ $? -ne 0 ]; then
+		echo "Resize may fail. Udevd events failed to settle within ${udevd_timeout} seconds."
 	fi
 	echo "Setting Max Sectors to ${size}"
 	hdparm --yes-i-know-what-i-am-doing -N p${size} /dev/${dev}	# NB: the p is for persistence


ixhamza:
Thanks, @yocalebo. Let me check if I could use `udevadm wait` for this particular scenario.

ixhamza:
@yocalebo - `udevadm wait /dev/sdY --settle` still waits for all the events to settle. I have tested it by opening and closing the `/dev/sdX` in a loop with the `O_RDWR` to generate continuous udev events for `/dev/sdX`. However, `udevadm wait /dev/sdY --settle` does not return until the whole event queue is drained. Moreover, the `udevadm wait ...` version is not available until `v251`, and we are currently using `v247` in Bluefin and v252 in Cobia.

yocalebo:
> --

Thanks for testing this, however, `udevadm settle` waits for _all_ hardware (not just the disk device) to settle. So are you saying that `udevadm wait` works (so long as >= v251 of systemd is present)?

ixhamza:
There does not seem to be a `udevadm wait /dev/sdX` version that could wait for the partition re-read to complete. After a discussion with @yocalebo on Slack, added a timeout version of `udevadm settle`.

src/freenas/usr/local/sbin/disk_resize
```diff
@@ -121,6 +122,12 @@ ata|sata)
 	if [ -z "${size}" ]; then
 		size=${maxsect}
 	fi
+	echo "Waiting up to ${udevd_timeout} seconds for udevd events to settle"
+	udevadm settle -t ${udevd_timeout}
+	if [ $? -ne 0 ]; then
+		echo "Resize failed. Udevd events failed to settle within ${udevd_timeout} seconds."
+		exit 1
```
amotin, line 129: I don't think you are required to fail here.  It may fail later if still hit the race, or it may succeed.  Warning message may be enough.

ixhamza: Ah, right. Thank you. Updated.

amotin (approved):

yocalebo (approved):
Thanks

PR #11756, opened 2023-07-31
Fix remaining items from reporting framework rewrite

yocalebo:
This does a few things:
- fix the failing integration tests by removing tests that no longer test anything (since migration to netdata)
- remove the remaining public API endpoints (and make existing NO-OPs where necessary) since the webUI team is no longer consuming them (neither is TC)
- fix a crash in disk.temperature_agg since it is referencing the old implementation of the reporting framework


Qubad786 (commented):
We have failing unit tests

yocalebo:
meh, too many changes and too much scope. I'll close and open one to fix immediate proboems.

PR #11767, opened 2023-08-01
NAS-123338 / 24.04 / change the release notes URL

yocalebo:
Now that we've branched for Cobia, the release notes webpage will be published soon and these are the URLs that will be used.


Mrt134 (approved):
Looks great, thanks for the sync up on this!

PR #11781, opened 2023-08-02
NAS-122711 / 24.04 / add ES102S enclosure support

yocalebo:
This adds ES102S enclosure support to our product. I had to change some logic because this enclosure seems to enumerate disk slot directories starting at 1 instead of 0. Every other enclosure (including head-units) that we sell enumerate starting at 0 so this exposed the fact that our logic was flawed and we would be off by 1 when mapping the disks to their respective enclosure slots. I also noticed some pretty severe inefficiencies while I was here. We are parsing stdout for the enclosure and for each line in the output, we were iterating every enclosure disk slot directory in procfs and opening/closing files.

Fixes and improvements I've made here:
- do not iterate every enclosure slot dir for every line of stdout for the enclosure (we just skip the disk slot element types now)
- add a new method that generates the slot to disk object only once
- inside the new method utilize the new `libsg3.ses.EnclosureDevice` class (written in cython) to enumerate the enclosure status information to easily get the enclosure slot status
- always ensure that the disk mapping starts at 1 to, in theory, prevent this from occurring again when we add new enclosures to our portfolio

### DIFF
--- a/src/middlewared/middlewared/plugins/enclosure.py
+++ b/src/middlewared/middlewared/plugins/enclosure.py
@@ -1,7 +1,6 @@
 import asyncio
 import errno
 import logging
-import os
 import re
 import pathlib
 from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


yocalebo:
ping @truenas/middleware

sonicaj (approved):

yocalebo:
backport

PR #11784, opened 2023-08-03
NAS-123267 / 24.04 / Use the correct namespace id for disk resize

ixhamza:
In nvmeXnY, 'Y' does not represent the namespace ID, but rather the namespace instance. Reading the correct namespace ID allows us to resize nvmeXn2 disks, which happens in the case of HA where fenced does not release a disk, making it reappear as nvmeXn2.

Tested by Jeff Ervin on f60-141.dc1.ixsystems.net.

### DIFF
--- a/src/freenas/usr/local/sbin/disk_resize
+++ b/src/freenas/usr/local/sbin/disk_resize
@@ -167,7 +167,7 @@ sas|scsi)
 nvme)
 	# Get controller and current namespace ID.
 	ctrlr=/dev/${dev%n*}
-	nsid=${dev##nvme*n}
+	nsid=`nvme get-ns-id /dev/${dev} | awk -F 'namespace-id:' '{ print $2 }'`
 
 	# Identify controller properties.
 	idctrl=`nvme id-ctrl ${ctrlr}`


yocalebo (approved):

usaleem-ix (approved):

src/freenas/usr/local/sbin/disk_resize
```diff
@@ -167,7 +167,7 @@ sas|scsi)
 nvme)
 	# Get controller and current namespace ID.
 	ctrlr=/dev/${dev%n*}
-	nsid=${dev##nvme*n}
+	nsid=`nvme id-ns /dev/${dev} | awk '/^NVME Identify Namespace/ { print $4 }' | xargs | tr -d ': '`
```
amotin, line 170: Why not `nvme get-ns-id`, that seems to be specially made for it?

ixhamza: @amotin - Thank you for pointing it out. I totally missed that. Let me update.

ixhamza: Thanks, updated.

amotin (approved):

PR #11807, opened 2023-08-07
NAS-123440 / 24.04 / Allow user to unset his 2fa secret

sonicaj:
This PR adds changes to allow user to unset his 2fa secret if 2fa is disabled globally.


yocalebo (approved):

yocalebo:
Im adding this for 23.10-RC.1 just to prevent as much divergence with master as possible.

PR #11809, opened 2023-08-07
NAS-123443 / 24.04 / Handle netdata response in chunks

Qubad786:
## Problem

On a m60 system with 1200 disks we saw asyncio logging warnings with the request saying that the task is blocked.

## Solution

The problem was occurring when we tried to load all of the content at once it blocked the event loop and breaking that in chunks solved the problem with the warning being no longer visible.

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/netdata/client.py
+++ b/src/middlewared/middlewared/plugins/reporting/netdata/client.py
@@ -2,6 +2,7 @@ import aiohttp
 import aiohttp.client_exceptions
 import asyncio
 import contextlib
+import json
 
 from .exceptions import ApiException, ClientConnectError
 from .utils import NETDATA_URI, NETDATA_REQUEST_TIMEOUT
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/reporting/netdata/client.py
```diff
@@ -34,6 +35,9 @@ async def request(
     async def api_call(cls, resource: str, timeout: int = NETDATA_REQUEST_TIMEOUT, version: str = 'v1') -> dict:
         try:
             async with cls.request(resource, timeout, version) as resp:
-                return await resp.json()
+                output = ''
+                async for line in resp.content.iter_any():
+                    output += line.decode(errors='ignore')
+                return json.loads(output)
```
themylogin, line 41: Which exact operation did block the loop? The parsing of (seemingly) large JSON will still happen in the main loop

Qubad786: @themylogin difference is when `json` method is called directly on the response - it reads the whole response content in a single go i.e
```
    async def json(
        self,
        *,
        encoding: Optional[str] = None,
        loads: JSONDecoder = DEFAULT_JSON_DECODER,
        content_type: Optional[str] = "application/json",
    ) -> Any:
        """Read and decodes JSON response."""
        if self._body is None:
            await self.read()
```

Whereas in this case we don't do that all at once and instead read it in chunks which results in us seeing the task being blocked anymore.

PR #11810, opened 2023-08-07
NAS-123216 / 24.04 / Do not crash on sysfs write failure

themylogin:


src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -60,8 +60,10 @@ def get_sysctl(self, var):
 
     @private
     def set_sysctl(self, var, value):
-        with open(f'/proc/sys/{var.replace(".", "/")}', 'w') as f:
-            f.write(value)
+        path = f'/proc/sys/{var.replace(".", "/")}'
```
sonicaj, line 63: Do you think we should perhaps have validation to now allow such tunables in the first place ?

src/middlewared/middlewared/plugins/tunables.py
```diff
@@ -60,8 +60,10 @@ def get_sysctl(self, var):
...
+        if os.access(path, os.W_OK):
```
yocalebo, line 64: It's pretty confusing and not intuitive that `os.access` implicitly returns false when the path doesn't exist. This is also, technically, a race condition since these path's are dynamic and can change based on hardware being inserted/removed.

I'd much rather do
```
try:
    with open(path, 'w') as f:
        f.write(value)
except FileNotFoundError:
    (do we raise a validation error?)
except PermissionDeniedError:
    (do we raise a validation error?)
```

We need to be careful if we choose to raise a validation error since this is called in the installer/upgrade image on our enterprise platforms potentially.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/tunables.py
themylogin: > this is called in the installer/upgrade image on our enterprise platforms potentially.

middleware is not running when the installer is being executed so this will not be used in the installer image code

yocalebo (approved):
Thanks!

sonicaj (approved):

PR #11815, opened 2023-08-08
NAS-123419 / 13.0 / Add overwrites for TrueCommand service start and get_state

ZackaryWelch:
The `truecommand` service, aka `wireguard`, isn't a real service but is instead a rc wrapper around wg-quick, which simply opens or closes the VPN connection. There is no need to close the connection on start, and `wireguard` will not show up in the list of running processes, as it is not a real daemon. This is only an issue on CORE, so it does not need a SCALE backport.

Added debugs because `wireguard`'s output is singled out to be redirected to /dev/null for whatever reason, so this will help understand why the link might go up or down.

### DIFF
--- a/src/middlewared/middlewared/plugins/service_/services/truecommand.py
+++ b/src/middlewared/middlewared/plugins/service_/services/truecommand.py
@@ -13,8 +13,8 @@ class TruecommandService(SimpleService):
     systemd_unit = 'wg-quick@wg0'
 
     async def _start_freebsd(self):
-        await self._freebsd_service(self.freebsd_rc, "start")
+        await self._freebsd_service(self.freebsd_rc, 'start')
 
     async def _get_state_freebsd(self):
-        status = (await self._freebsd_service(self.freebsd_rc, "status")).stdout
+        status = (await self._freebsd_service(self.freebsd_rc, 'status')).stdout
         return ServiceState("Device not configured" not in status, [])
--- a/src/middlewared/middlewared/plugins/service_/services/truecommand.py
+++ b/src/middlewared/middlewared/plugins/service_/services/truecommand.py
@@ -12,9 +12,6 @@ class TruecommandService(SimpleService):
 
     systemd_unit = 'wg-quick@wg0'
 
-    async def _start_freebsd(self):
-        await self._freebsd_service(self.freebsd_rc, 'start')
-
     async def _get_state_freebsd(self):
         status = (await self._freebsd_service(self.freebsd_rc, 'status')).stdout
         return ServiceState("Device not configured" not in status, [])
--- a/src/middlewared/middlewared/plugins/service_/services/truecommand.py
+++ b/src/middlewared/middlewared/plugins/service_/services/truecommand.py
@@ -1,4 +1,4 @@
-from .base import SimpleService
+from .base import SimpleService, ServiceState
 
 
 class TruecommandService(SimpleService):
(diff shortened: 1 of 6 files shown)


src/middlewared/middlewared/plugins/truecommand/__init__.py
```diff
@@ -6,6 +6,7 @@ async def _event_system(middleware, event_type, args):
         if await middleware.call('failover.licensed'):
             return
 
+        middleware.logger.debug('Starting TrueCommand service on system boot')
```
sonicaj, line 9: This is not true, the method below determines if it is going to be started

src/middlewared/middlewared/plugins/truecommand/update.py
```diff
@@ -170,6 +170,8 @@ async def do_update(self, data):
             # can happen on update
             # 1) Service enabled/disabled
             # 2) Api Key changed
+
+            self.middleware.logger.debug('Stopping TrueCommand service on config change')
```
sonicaj, line 174: Is this really required ? If TC was not running before, this is misleading ?

src/middlewared/middlewared/plugins/truecommand/wireguard.py
```diff
@@ -127,7 +127,12 @@ async def start_truecommand_service(self):
             if Status(config['api_key_state']) == Status.CONNECTED and all(
                 config[k] for k in ('wg_private_key', 'remote_address', 'endpoint', 'tc_public_key', 'wg_address')
             ):
-                await self.middleware.call('service.start', 'truecommand')
+                if not await self.middleware.call('service.started', 'truecommand'):
```
sonicaj, line 130: Is this required ? Just want to account for random race conditions

src/middlewared/middlewared/plugins/service_/services/truecommand.py
```diff
@@ -11,3 +11,10 @@ class TruecommandService(SimpleService):
     freebsd_proc_arguments_match = True
 
     systemd_unit = 'wg-quick@wg0'
+
+    async def _start_freebsd(self):
+        await self._freebsd_service(self.freebsd_rc, "start")
```
sonicaj, line 16: Please use single quotes as the whole file uses single quotes

sonicaj (commented):
Have these changes been tested on 13.1 as well ?

src/middlewared/middlewared/plugins/truecommand/wireguard.py
ZackaryWelch: I'm not sure. Testing without it.

ZackaryWelch: Working fine now, but needs a one line patch in the port. I'll link that to this issue.

sonicaj (approved):
Thanks!

ZackaryWelch:
@sonicaj Removed a couple of lines that I've been able to do without.

sonicaj (approved):

ZackaryWelch:
https://github.com/truenas/ports/pull/1300

yocalebo:
@ZackaryWelch does this also need to be in 13.1?

ZackaryWelch:
> @ZackaryWelch does this also need to be in 13.1?

Yes. Do you want me to create another branch/PR with cherry picks?

sonicaj:
@ZackaryWelch bug clerk should do that for us, Caleb was asking for confirmation if that needs to happen. Thanks for confirming!

PR #11816, opened 2023-08-09
NAS-123479 / 24.04 / port CORE enclosure plugin to SCALE (stage 1)

yocalebo:
### Introduction
Before the reviewer(s) go(es) into a panic because of the ~2k lines of changes, it's important to note a few things.
1. this is NOT being used by the front-end and isn't even being called by back-end. This is incomplete right now (still need to handle mapping nvme drives on our enclosures, and drive identification) but is fully functional otherwise
2. the `element_types.py`, `enclosure_class.py` and `identification.py` files are essentially copy and pasted directly from CORE with a few minor tweaks.
3. 869 lines of changes are from unit tests (yayyy, unit tests are better than no tests 😄 )
4. MOST of this logic is essentially what we do on CORE so I'm not diverging significantly
5. I've added a pure python (ctypes) script (`scsi_generic.py`) for sending inquiry commands to the enclosure devices (will work with SAS drives too) to more efficiently return the vendor, product, revision (and serial if we wanted) data from said device.

### Description
The changes employed here were made because of learning of some fundamental design flaws with how our current enclosure operates. Frankly, our existing logic is working by pure luck. "Failing successfully" essentially. Our old logic references "num" as an item in the namedtuple which, historically, was an attribute added in CORE because on freeBSD the enclosure devices are enumerated as devices starting at `/dev/ses0` and incrementing by 1. The `0` is what we referenced as the `num` in said namedtuple. Barring myself from writing a novel to explain on how that is actually flawed logic as well, I'll cut to the point. SCALE (linux) doesn't have such a notion but even more so enclosure device attributes will change (and often do) depending on what else is plugged into the system. We have to have some form of unchanging uniquely identifying property of the enclosure to ensure we can properly find said enclosure and map the drive slots accordingly. I've managed to do this in the `slot_mapping.py` file. The "unchanging" unique identifiers for our enclosure devices are what our drive mapping logic is hinged upon. This varies between platforms, obviously, but it solves the overarching problem with how we've designed our enclosure plugin currently.

### Unit Tests
I've added unit tests because our current enclosure logic isn't tested in any fashion. While unit tests don't guarantee success, they provide other developers an idea of what particular functions do and how data should be formatted. Some tests are better than no tests 😄 

### Maintainability
Code maintainability has, hopefully, been improved as well. While this is highly subjective, the current code was unnecessarily complex and fragile. With the changes employed here, I can write developer documentation on how to add a new platform and what all changes are required for this to happen. While this could have been done with the existing logic, the rewrite has allowed me to have a clear picture of exactly what needs to change.

### Performance
Finally performance has been lacking in this particular area because of the complexity of our old logic. I have tested the changes on the largest system we sell (~1.2k HDDs with 12x ES102 shelves) and it is 133% faster (2.5 seconds to 0.5 seconds for `enclosure.query` to complete)

### To-do
1. fix the `map_nvme` logic
2. add drive identification to `enclosure2.py` (and test by physical going into DC and visually inspecting drive leds)
3. add docstrings to _EVERYTHING_ where necessary
4. add developer documentation for adding a new enclosure
5. add moarrrr tests

### DIFF
--- a/src/middlewared/middlewared/plugins/enclosure_/map2.py
+++ b/src/middlewared/middlewared/plugins/enclosure_/map2.py
@@ -49,7 +49,7 @@ def map_enclosures(enclosures):
                     'original': {
                         'enclosure_id': enclosure['id'],
                         'enclosure_sg': enclosure['sg'],
-                        'encllsure_bsg': enclosure['bsg'],
+                        'enclosure_bsg': enclosure['bsg'],
                         'descriptor': orig_info['descriptor'],
                         'slot': orig_slot
                     }
--- a/src/middlewared/middlewared/plugins/enclosure_/map2.py
+++ b/src/middlewared/middlewared/plugins/enclosure_/map2.py
@@ -1,0 +1,59 @@
+from .slot_mappings import MAPPINGS
+
+
+def map_enclosures(enclosures):
+    mapped = [{'id': 'mapped_enclosure_0', 'elements': {'Array Device Slot': {}}}]
+    jbods = []
+    for enclosure in enclosures:
+        if not enclosure['controller']:
+            jbods.append(enclosure)
+            continue
+
+        try:
+            mapped_dict = MAPPINGS[enclosure['dmi']]['mapping_info']
+        except KeyError:
+            # X, M, F series don't need to be mapped so this is expected
+            return enclosures
+
+        vers_key = 'DEFAULT'
+        if not MAPPINGS[enclosure['dmi']]['any_version']:
+            # platforms can have different "versions" which
+            # means that they can be ever so slightly cabled
+            # differently which leads to us having to map the
+            # drives based on how they're cabled
+            # (hence the "version")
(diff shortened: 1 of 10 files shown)


yocalebo:
retest this please

src/middlewared/middlewared/plugins/enclosure_/map2.py
```diff
@@ -0,0 +1,59 @@
+from .slot_mappings import MAPPINGS
+
+
+def map_enclosures(enclosures):
```
bmeagherix, line 4: It might be appropriate to add a comment here explaining the whole `map_enclosures` mechanism, along with some explanation of the input and output of the function.

bmeagherix (approved):
I'm approving because none of my comments _require_ changes.

As mentioned inline I would suggest that a brief general overview of the enclosure mapping goals & mechanism be provided (and preferably some example input/output).

PR #9742, opened 2022-08-30
NAS-117931 / 22.12 / HTTP Basic Auth is unavailable when OTP is enabled

themylogin:

PR #11818, opened 2023-08-09
NAS-123289 / 13.1 / HTTP Basic Auth is unavailable when OTP is enabl…

sonicaj:
…ed (#9742)

* HTTP Basic Auth is unavailable when OTP is enabled

* OTP integration tests

### DIFF
--- a/tests/api2/test_auth_otp.py
+++ b/tests/api2/test_auth_otp.py
@@ -18,12 +18,3 @@ def test_otp_http_basic_auth(otp_enabled):
         r = s.get(f"{url()}/api/v2.0/system/info/")
         assert r.status_code == 401
         assert r.text == "HTTP Basic Auth is unavailable when OTP is enabled"
-
-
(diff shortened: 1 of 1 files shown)


sonicaj:
backport

tests/api2/test_auth_otp.py
```diff
@@ -0,0 +1,29 @@
...
+def otp_enabled():
+    call("auth.twofactor.update", {"enabled": True})
+
+    try:
+        yield
+    finally:
+        ssh("midclt call auth.twofactor.update '{\"enabled\": false}'")
+
+
+def test_otp_http_basic_auth(otp_enabled):
```
yocalebo, line 16: We need to test the specific scenario that was highlighted in the ticket. (i.e. using root credentials, try to disable 2fa globally over REST)

yocalebo (changes requested):

PR #11822, opened 2023-08-10
NAS-123074 / 24.04 / Add unit tests for realtime statistics

Qubad786:
## Context

We have util functions which determine reporting realtime statistics and unit tests have been added to test the reporting in question to make sure it is accurate.


yocalebo (approved):
Nice!

PR #11825, opened 2023-08-10
Fix new port assigment bug

Qubad786:
## Problem

When assigning a port number to `web_port`, we are not taking into account the display port number that will be utilized by SPICE. Consequently, there are instances where the `web_port` and display ports end up being the same. This issue disrupts the entire virtual machine display configuration.

## Solution

If user has already specified a port we should remove that from the new ports we are getting from our port wizard so that if one of the attrs has been specified already and it matches to what the port wizard provided us, we remove it from there and make sure we don't end up using it.


sonicaj (approved):
Interesting catch :P

PR #11848, opened 2023-08-15
NAS-123488 / 24.04 / Fix bug in interface stats

Qubad786:
## Problem

The ifstat function is currently returning data on the first iteration, which prevents it from processing other interfaces.

## Solution

Make sure all interfaces are processed when getting realtime interface stats.


themylogin (approved):

Qubad786:
backport

PR #11849, opened 2023-08-15
NAS-121442 / 24.04 / Gather kubernetes statistics for UI

Qubad786:
## Context
At present, k3s statistics are obtained during runtime using the k3s node summary API. However, this approach introduces notable sluggishness into the process, particularly due to the extensive number of API calls needed to access and update data in real-time. To tackle this performance bottleneck, a novel solution has been implemented in the form of a netdata python.d plugin. This plugin systematically collects data at set intervals, enabling us to efficiently retrieve data through streamlined queries. Consequently, this transition in data collection effectively distributes the computational load, significantly alleviating the previous strain on the middleware. It's important to note that netdata does not offer alternative methods that can be seamlessly integrated into the middleware for accessing API statistics. To leverage netdata's kubectl API functionality, either netdata needs to be executed within a Kubernetes environment or the entire Go plugin has to be installed, both of which are impractical. Thus, the introduction of the python.d plugin serves as a pragmatic solution to this issue.

## References
https://learn.netdata.cloud/docs/miscellaneous/kubernetes-monitoring-with-netdata-overview-and-visualizations

https://learn.netdata.cloud/docs/data-collection/kubernetes/kubelet

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
@@ -136,6 +136,7 @@ class ChartReleaseService(CRUDService):
         extra = copy.deepcopy(options.get('extra', {}))
         retrieve_schema = extra.get('include_chart_schema')
         retrieve_stats = extra.get('stats', False)
+        netdata_metrics = await self.middleware.call('netdata.get_all_metrics') if retrieve_stats else None
         get_resources = extra.get('retrieve_resources')
         get_locked_paths = extra.get('retrieve_locked_paths')
         locked_datasets = await self.middleware.call('zfs.dataset.locked_datasets') if get_locked_paths else []
@@ -219,7 +220,7 @@ class ChartReleaseService(CRUDService):
 
             if retrieve_stats:
                 release_data['stats'] = await self.middleware.call(
-                    'chart.release.stats_internal', resources[Resources.POD.value][name]
+                    'chart.release.stats_internal', resources[Resources.POD.value][name], netdata_metrics,
                 )
 
             container_images_normalized = {
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
@@ -1,4 +1,5 @@
 import collections
+import contextlib
 import copy
 import errno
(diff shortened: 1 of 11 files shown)


src/middlewared/middlewared/plugins/reporting/stats_utils.py
```diff
@@ -0,0 +1,23 @@
...
+            netdata_metrics, f'k3s_stats.{pod_name}.mem', f'{pod_name}.mem', default=0
+        ))
+        stats['network']['incoming'] += safely_retrieve_dimension(
+            netdata_metrics, f'k3s_stats.{pod_name}.net', f'{pod_name}.net.incoming', default=0
+        )
+        stats['network']['outgoing'] += safely_retrieve_dimension(
+            netdata_metrics, f'k3s_stats.{pod_name}.net', f'{pod_name}.net.outgoing', default=0
+        )
+
+    stats['cpu'] = ((stats['cpu'] / 1000000000) / cpu_info()['core_count']) * 100
```
yocalebo, line 22: Why are we dividing by `1000000000`? Add a comment here so we know what's going on.

yocalebo (approved):

src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
```diff
@@ -135,6 +135,8 @@ async def query(self, filters, options):
         options = options or {}
         extra = copy.deepcopy(options.get('extra', {}))
         retrieve_schema = extra.get('include_chart_schema')
+        retrieve_stats = extra.get('stats', False)
+        netdata_metrics = await self.middleware.call('netdata.get_all_metrics') if retrieve_stats else None
```
yocalebo, line 139: This seems heavy-handed to query all stats on the system just for apps? Can we not query the k3s specific stats explicitly here?

yocalebo (changes requested):

src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
Qubad786: @yocalebo i have made changes so that instead of having k8s data split in multiple charts - we gather it in one single chart and query accordingly..

yocalebo (approved):
Thanks

PR #11850, opened 2023-08-15
NAS-123466 / 24.04 / Make sure containerd has initialized itself before declaring apps are ready

sonicaj:
This PR adds changes which makes sure that containerd socket is available and has initialized itself appropriately as we are seeing in some cases that we mark the node as ready but that is not the case with containerd socket still not available for consumption.

### DIFF
--- a/src/middlewared/middlewared/utils/socket.py
+++ b/src/middlewared/middlewared/utils/socket.py
@@ -1,16 +1,13 @@
-import os
 import socket
 
 
 def is_socket_available(socket_path: str) -> bool:
     """Check if a Unix socket is available."""
-    if not os.path.exists(socket_path):
-        return False
     s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
     try:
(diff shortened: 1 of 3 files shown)


Qubad786 (approved):

src/middlewared/middlewared/utils/socket.py
```diff
@@ -0,0 +1,16 @@
...
+import socket
+
+
+def is_socket_available(socket_path: str) -> bool:
+    """Check if a Unix socket is available."""
+    if not os.path.exists(socket_path):
+        return False
+    s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    try:
+        s.connect(socket_path)
```
yocalebo, line 11: Why not just try to connect and catch FileNotFoundError?

sonicaj: Good thinking, will make the changes

yocalebo (approved):

PR #11856, opened 2023-08-16
NAS-123560 / 24.04 / Fix `get_cloudsync_tasks_count` and `get_rsync_tasks_count`

themylogin:


yocalebo (changes requested):
I was still investigating and working on a PR (is why the ticket was still assigned to me) but I was trying to find:
- what are all the possible values that `mountpoint` can be set to? (i.e. `-`, `legacy`, literal `none`).
- we can move the filter logic in both of these methods to a single function that both `get_cloudsync_tasks..` and `get_rsync_tasks..` could use.

themylogin:
@yocalebo `-` is `None`

yocalebo (approved):

themylogin:
time 1:00

PR #11860, opened 2023-08-16
NAS-123572 / 24.04 / Update OTP test to try and disable 2FA

sonicaj:
This commit adds changes to try and disable 2fa instead of querying system info as a customer had faced this scenario in previous versions and we want to be sure we are not allowing that to happen.


yocalebo (approved):
Thank you!

PR #11868, opened 2023-08-16
NAS-123570 / 24.04 / Extending a file-based iSCSI target has no effect

bmeagherix:
Allow the size of a file-based extent for an iSCSI target to be increased.

This includes adding `resync_lun_size_for_file` (similar to `resync_lun_size_for_zvol`) to notify SCST and any connected clients about the change.

---
Currently the API ignores attempts to reduce the size of the extent (although the size will be saved in the database).  I would like to tweak the API to also allow reducing the size of the file, however ...

For ZVOL based targets, the webui code currently traps attempting to reduce the size of a target.  It presents the error message:
> Shrinking a ZVOL is not allowed in the User Interface. This can lead to data loss.

@undsoft do you think the webui should add a similar block to prevent reducing the size of a file extent?  (If so, I can then tweak this PR to allow any size change.)  If not, should I throw a `ValidationError` if the user attempts to reduce the size?

### DIFF
--- a/src/middlewared/middlewared/plugins/iscsi_/extents.py
+++ b/src/middlewared/middlewared/plugins/iscsi_/extents.py
@@ -141,6 +141,7 @@ class iSCSITargetExtentService(SharingService):
         verrors.check()
 
         await self.middleware.call('iscsi.extent.save', new, 'iscsi_extent_create', verrors, old)
+        verrors.check()
         new.pop(self.locked_field)
 
         await self.middleware.call(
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/iscsi_/extents.py
```diff
@@ -411,9 +411,20 @@ def save(self, data, schema_name, verrors):
...
+            # create the extent, or perhaps extend it
             if not os.path.exists(path):
+                # create the extent
                 subprocess.run(['truncate', '-s', str(data['filesize']), path])
+            else:
+                if old:
+                    old_size = int(old['filesize'])
+                    new_size = int(data['filesize'])
+                    # For now only allow expansion, we might remove this
+                    # restriction later if the webui adds a check.
```
yocalebo, line 423: We should raise a `ValidationError` to prevent foot-shooting in the event uses the webUI to shrink an extent. There are too many "gotchas" there and too many edge-case scenarios where that can get out of hand.

bmeagherix: Thanks.  Done.

yocalebo (approved):
Thanks

bmeagherix:
time 4:00

PR #11871, opened 2023-08-16
NAS-123498 / 24.04 / Allow NFS server to start with empty or missing exports

mgrimesix:
In releases prior to Cobia, the NFS server start was dependent on the existence of NFS shares and would potentially complain if the exports file was empty.  This is no longer the case in Cobia. 
What this changes:
    - NFS server start is independent of NFS shares
 
Additions:
    - Entries in /etc/exports.d are not tolerated and a systemd start condition was added to that effect
    - At manual NFS start, we raise an error alert if entries are detected in /etc/exports.d

Added a CI test for NFS service start independent of shares
Added a CI test for NFS start with rogue entries in /etc/exports.d

### DIFF
--- a/src/middlewared/middlewared/plugins/service_/services/nfs.py
+++ b/src/middlewared/middlewared/plugins/service_/services/nfs.py
@@ -1,8 +1,9 @@
-import errno
+# import errno
 import os
 
 from .base import SimpleService
-from middlewared.service_exception import CallError
+# from middlewared.service_exception import CallError
+poison_exports_marker = "/var/local/bad_exports_block_nfsd"
 
 
 class NFSService(SimpleService):
@@ -14,13 +15,17 @@ class NFSService(SimpleService):
     systemd_unit = "nfs-server"
 
     async def check_configuration(self):
-        if not await self.middleware.run_in_thread(os.path.exists, '/etc/exports'):
-            raise CallError(
-                'At least one configured and available NFS export is required '
-                'in order to start the NFS service. Check the NFS share configuration '
-                'and availability of any paths currently being exported.',
-                errno.EINVAL
-            )
+        # NAS-123498: Eliminate requirement to have shares configured to start NFS
+        # But, raise an alarm if there are entries in /etc/exports.d
+        exportsd = '/etc/exports.d'
+        if os.path.exists(exportsd) and not os.path.isfile(exportsd):
+            exportsdList = os.listdir(exportsd)
+            if len(exportsdList) > 0:
+                await self.middleware.call('alert.oneshot_create', 'NFSblockedByExportsDir', {
+                    'entries': exportsdList
+                })
+            else:
+                await self.middleware.call('alert.oneshot_delete', 'NFSblockedByExportsDir', None)
 
     async def after_start(self):
         await self._systemd_unit("rpc-statd", "start")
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/service_/services/nfs.py
```diff
@@ -14,13 +12,17 @@ class NFSService(SimpleService):
...
-            raise CallError(
-                'At least one configured and available NFS export is required '
-                'in order to start the NFS service. Check the NFS share configuration '
-                'and availability of any paths currently being exported.',
-                errno.EINVAL
-            )
+        # NAS-123498: Eliminate requirement to have shares configured to start NFS
+        # But, raise an alarm if there are entries in /etc/exports.d
+        exportsd = '/etc/exports.d'
+        if os.path.exists(exportsd) and not os.path.isfile(exportsd):
```
yocalebo, line 18: This is calling blocking I/O in an async method so it'll block the event loop. Lastly, I don't like checking for existence of things before performing I/O on them if we don't have to. We should create a synchronous method and call it inside a thread. i.e.

```
def check_exportsd_dir(self):
    exports = list()
    try:
        with os.scandir('/etc/exports.d') as scan:
            for i in filter(lambda x: x.is_file() and not x.name.startswith('.'), scan):
                exports.append(i.name)
    except (FileNotFoundError, NotADirectoryError):
        pass
   return exports
   
async def check_configuration(self):
    if (exportsdList := await self.middleware.run_in_thread(self.check_exports_dir)):
        await self.middleware.call('alert.oneshot_create', 'NFSblockedByExportsDir', {'entries': exportsdList})
    else:
        await self.middleware.call('alert.oneshot_delete', 'NFSblockedByExportsDir')

**Review by yocalebo — CHANGES_REQUESTED** [review]

File: tests/api2/test_300_nfs.py [hunk 2]
```diff
@@ -1161,3 +1161,30 @@ def test_55_checking_nfs_disable_at_boot(request):
 def test_56_destroying_smb_dataset(request):
     results = DELETE(f"/pool/dataset/id/{dataset_url}/")
     assert results.status_code == 200, results.text
+
+
+def test_60_start_nfs_service_with_empty_exports():
```
yocalebo, line 1166: The tests are a bit confusing to me but maybe I'm misunderstanding 😄 With your changes here, there is no reason to create an empty exports file? Simply starting the nfs service will generate an export file and the service will start, right?

Lastly, should we test the scenario of creating a file in `/etc/exports.d/` and making sure the alert is raised appropriately? Furthermore ensuring that the alert is removed if there are no files in `/etc/exports.d`?

mgrimesix: > Simply starting the nfs service will generate an export file and the service will start

That is correct. I'll modify the test to run both empty and missing.

Testing files actually in /etc/exports.d is a bit more complicated as our code makes it _really_ hard to actually put anything in that  directory.   I would have to create a file that 'root' is unable to remove.  This could be tricky and problematic in our automated testing.   Let me know if you know of a safe was to do this.    I did my local testing by pointing the check to a different directory.

mgrimesix: immutable files work and it exposed that I'm not getting a companion alert set.  
Going to place this in WIP until I get that ironed out.
Thanks, Caleb, for being detailed!

yocalebo (approved):
Great! thanks

PR #11874, opened 2023-08-17
NAS-123597 / 24.04 / Add an additional attribute in replication task

sonicaj:
This PR adds an additional attribute in replication tasks which can be used to determine if the task has associated datasets which have encryption keys which can be downloaded by the user. This attribute can be consumed by UI to display a button next to replication tasks so that it is only displayed when there are actually datasets available which match the criteria.

While at it, performance for gathering keys which will be exported for a replication task has been improved as well with previous timing for me being on average 4 seconds for a task i have and now i am seeing 1.5 seconds.

For `replication.query` i see that when the extra flag is not specified the time for me is `0.09` seconds and with the flag it is 0.2 seconds where i have 3 replication tasks.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/dataset_encryption_info.py
+++ b/src/middlewared/middlewared/plugins/pool_/dataset_encryption_info.py
@@ -367,7 +367,7 @@ class PoolDatasetService(Service):
 
         for source_ds in task['source_datasets']:
             for ds_name, key in mapping[source_ds].items():
-                for dataset in dataset_mapping[ds_name] if include_encryption_root_children else [{'id': ds_name}]:
+                for dataset in (dataset_mapping[ds_name] if include_encryption_root_children else [{'id': ds_name}]):
                     result[dataset['id'].replace(source_ds, source_mapping[source_ds], 1)] = key
 
         return result
@@ -379,6 +379,8 @@ class PoolDatasetService(Service):
             'pool.dataset.query', [], {'extra': {'properties': ['encryptionroot']}}
         ):
             dataset_encryption_root_mapping[dataset['encryption_root']].append(dataset)
+
+        return dataset_encryption_root_mapping
 
     @accepts(
         Str('id'),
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/replication.py
```diff
@@ -97,8 +97,15 @@ class Config:
 
     @private
     async def extend_context(self, rows, extra):
+        if extra.get("check_dataset_encryption_keys", False) and any(row["direction"] == "PUSH" for row in rows):
+            (await self.middleware.call("pool.dataset.sync_db_keys")).wait()
```
themylogin, line 101: Why does this have to happen every time?

src/middlewared/middlewared/plugins/pool_/dataset_encryption_info.py
```diff
@@ -356,17 +364,28 @@ async def export_keys_for_replication_internal(self, replication_task_id):
...
+                for dataset in (dataset_mapping[ds_name] if include_encryption_root_children else [{'id': ds_name}]):
                     result[dataset['id'].replace(source_ds, source_mapping[source_ds], 1)] = key
 
         return result
 
+    @private
+    async def dataset_encryption_root_mapping(self):
+        dataset_encryption_root_mapping = collections.defaultdict(list)
+        for dataset in await self.middleware.call(
+            'pool.dataset.query', [], {'extra': {'properties': ['encryptionroot']}}
```
yocalebo, line 383: So we now will query every dataset on the system if this arg is given to replication.query?

sonicaj: That unfortunately needs to be done - on the positive side we are not getting all the properties so it does not has that much of an impact i.e my VM has 114 datasets and `replication.query` is taking now 0.16 seconds approximately.

sonicaj: It was `0.09` before this change

themylogin (approved):

PR #11906, opened 2023-08-22
NAS-123374 / 24.04 / Fix `set_slot_status` not unmapping the physical enclosure bsg

themylogin:


src/middlewared/middlewared/plugins/enclosure.py
```diff
@@ -206,18 +206,18 @@ def _get_ses_slot_for_disk(self, disk):
 
         return self._get_ses_slot(enclosure, element)
 
-    def _get_original_disk_slot(self, enclosure_id, slot, info):
+    def _unmap_enclosure(self, enclosure_id, slot, info):
```
yocalebo, line 209: Changing the name of this function doesn't make sense to me. Im not sure what "unmap_enclosure" means so please go back to using the original function name. You can name it something like "get_orig_enclosure_and_disk"

yocalebo (changes requested):

themylogin:
time 4:00

yocalebo (approved):
thanks!

PR #11931, opened 2023-08-24
NAS-123688 / 24.04 / Fix AttributeError crash in _doesVMDependOnDataStore

themylogin:


themylogin:
time 0:05

sonicaj (approved):

src/middlewared/middlewared/plugins/vmware.py
```diff
@@ -596,6 +596,8 @@ def _doesVMDependOnDataStore(self, vm, dataStore):
                 if device.backing is None:
```
yocalebo, line 596: Does it make sense to just to a try: except here?
```
try:
    return device.backing.datastore.info.name == dataStore
except AttributeError:
    continue

PR #11934, opened 2023-08-24
NAS-123692 / 24.04 / Add additional draid vdev stats when querying pool data

Qubad786:
## Context

It was requested by UI that middleware should report draid vdev statistics like parity/spares/data disks etc when querying for pool information.


yocalebo (approved):
Thanks!

PR #11936, opened 2023-08-24
NAS-123612 / 24.04 / add fseries drive identification logic

yocalebo:
This was missed when I initially added the new F series product. This is very similar to what we do with the r30.


bmeagherix (approved):
Looks good.

PR #11937, opened 2023-08-24
NAS-119369 / 24.04 / Properly account for the case when TC container might not be running

sonicaj:
This PR adds changes to raise an alert for the user in the case when wireguard service on truenas is running as desired but the the TC container is actually down.

### DIFF
--- a/src/middlewared/middlewared/plugins/truecommand/portal.py
+++ b/src/middlewared/middlewared/plugins/truecommand/portal.py
@@ -44,10 +44,12 @@ class TruecommandService(Service, TruecommandAPIMixin):
                 )
                 if status['tc_state'] == 'running':
                     await self.middleware.call('truecommand.dismiss_alerts')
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/truecommand/portal.py
```diff
@@ -42,8 +42,14 @@ async def poll_api_for_status(self, job):
                 self.middleware.send_event(
                     'truecommand.config', 'CHANGED', fields=(await self.middleware.call('truecommand.config'))
                 )
-                await self.middleware.call('truecommand.dismiss_alerts')
+                if status['tc_state'] == 'running':
```
yocalebo, line 45: We should get this key safely since we've already seen it not exist in some circumstances.

yocalebo (changes requested):

yocalebo (approved):

PR #11963, opened 2023-08-28
NAS-123700 / 24.04 / strip trailing newline/whitespace on sshpubkey

yocalebo:
Strip trailing newlines/whitespace in `user.query` for the `sshpubkey` attribute.


sonicaj (approved):
As discussed this does not handle the case when multiple keys are present but that is then a totally new problem :P

anodos325 (approved):

PR #11964, opened 2023-08-28
NAS-123755 / 24.04 / Collect disk temperature stats at 30 min intervals

Qubad786:
## Problem

Currently, smartd collects data at a 30-minute interval. However, our current data collection process from smartd involves allocating disk space for a 1-second interval. This mismatch between the collection intervals and disk space allocation intervals is causing excessive disk space allocation and unnecessarily cpu is being consumed as well.

## Solution

To address this, we need to adjust our data collection process to match the 30-minute interval used by smartd for data collection.

### DIFF
--- a/src/freenas/usr/lib/netdata/python.d/smart_log.chart.py
+++ b/src/freenas/usr/lib/netdata/python.d/smart_log.chart.py
@@ -258,6 +258,9 @@ class Service(SimpleService):
         self.disks = list()
         self.runs = 0
         self.do_force_rescan = False
+        # smartd daemon only queries drive temps every 30mins so the files won't be updated
+        # but once every 30ish minutes - we should change this if at any point we change smartd interval
+        self.update_every = 30 * 60
 
     def check(self):
         return self.scan() > 0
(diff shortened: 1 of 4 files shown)


src/freenas/usr/lib/netdata/python.d/smart_log.chart.py
```diff
@@ -258,6 +258,7 @@ def __init__(self, configuration=None, name=None):
         self.disks = list()
         self.runs = 0
         self.do_force_rescan = False
+        self.update_every = 30 * 60
```
yocalebo, line 261: Let's add a comment here explaining why we're doing this please 👍 (i.e. "smartd daemon only queries drive temps every 30mins so the files won't be updated but once every 30ish minutes") or something alone those lines 😄

themylogin (changes requested):

src/freenas/usr/lib/netdata/python.d/smart_log.chart.py
yocalebo: @themylogin no it's not.

PR #11971, opened 2023-08-29
Improve disk sizes tests

themylogin:


tests/api2/test_device_get_disks_size.py
```diff
@@ -1,15 +1,4 @@
...
-import pytest
-
 from middlewared.test.integration.utils import call, ssh
-from auto_config import dev_test
-
-# comment pytestmark for development testing with --dev-test
-pytestmark = pytest.mark.skipif(dev_test, reason='Skip for testing')
 
 
 def test_device_get_disks_size():
```
themylogin, line 4: @yocalebo I don't want to remove this test, it tests if `device.get_disks` actually works which is not covered by any other test in the suite

yocalebo: Then we should combine these files into 1. You also need to update the ticket to reflect reality since this PR is doing nothing related to what the original ticket asks.

yocalebo (changes requested):
replied to your comment

tests/api2/test_device_get_disks_size.py
themylogin: @yocalebo why would we want to combine these files into one? They test not only different methods, but even different plugins

themylogin: @yocalebo  we might potentially want to test other values returned by `device` or `disk` plugin which might be completely unrelated. The current situation is just an accident.

yocalebo: Seems like that was never thought about when these tests were written considering they're named the exact function they're testing. So adding any other tests to either of these files would mean that we would need to rename the test files themselves so they're not so narrow-scoped in their name. Either way, please update the ticket with what you've done and why so it reflects reality.

themylogin: @yocalebo done. Files will be renamed when the time comes.

themylogin:
time 0:30

yocalebo:
backport

PR #11972, opened 2023-08-29
NAS-123654 / 24.04 / Use only one encrypted pool for system dataset on encrypted pool tests

themylogin:


yocalebo (approved):

yocalebo (changes requested):
Was there no way to move some of the system dataset functionality to the other test like I asked in the original ticket?

themylogin:
@yocalebo I would rather move migration tests that test preserving dataset contents into a 3rd file than trying to mix this into test_006 which is really complex already.

yocalebo:
> @yocalebo I would rather move migration tests that test preserving dataset contents into a 3rd file than trying to mix this into test_006 which is really complex already.

The entire justification for making these changes is to lessen the amount of time our test suite runs. Adding, yet another, file that tests "system dataset functionality" when we're already testing it in other places is going against exactly what we're trying to achieve.

Using 1 zpool is a good first start but please update the ticket with why you decided not to merge some of the functionality into the other test so we don't have a ticket that doesn't reflect reality.

themylogin:
@yocalebo I don't see any `systemdataset.update` calls in test_006 so I fail to understand why is my test that runs this method is redundant.

Can I simply remove "We need to move the test_system_migrate functionality into the test_006 script" request from the ticket body?

themylogin:
time 0:30

themylogin:
@yocalebo ticket updated

yocalebo (approved):

PR #11983, opened 2023-08-29
NAS-123585 / 22.12.4 / Add validation for share ACL payload

anodos325:
This add checks that SID, domain, and name is known to us on an as-needed basis. Checking early in middleware gives better control of error messages before we pass along to the sharesec command. This endpoint has been substantially rewritten for Cobia and so the change only applies to BlueFin.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb_/sharesec.py
+++ b/src/middlewared/middlewared/plugins/smb_/sharesec.py
@@ -66,7 +66,7 @@ class ShareSec(CRUDService):
             return
 
         if not (await self.middleware.call('idmap.name_to_sid', f'{domain}\\{name}'))['sid']:
-            verrors.add(f'{schema}.ae_who_name.name', f'{domain}\{name}: unknown user or group.')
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/smb_/sharesec.py
```diff
@@ -42,6 +42,40 @@ class Config:
         'data_type': 'BYTES'
     }
 
+    @private
+    async def validate_share_ace(self, schema, ace, verrors):
+        if sid := ace.get('ae_who_sid'):
+            if not await self.middleware.call('idmap.sid_to_unixid', sid):
+                verrors.add(f'{schema}.sid', f'{sid}: SID is unknown')
+
+            return
```
yocalebo, line 51: Does this return need to be here or in scope of the verrors?

anodos325: Nah, If we have a valid SID we don't need to check the username because it will be ignored.

yocalebo (approved):

PR #11993, opened 2023-08-30
NAS-123821 / 24.04 / Fix zpool status alerts

yocalebo:
We're unnecessarily beating up our process pool by calling `pool.query` every minute in these 2 alerts. More importantly, however, we're doing it in a very inefficient manner.

This does a few things:
- get rid of the boot pool volume alert and move it into volume_status.py alert
- remove the 1hr interval so it goes back to running every 60 seconds
- before we do anything expensive, check to see if we've already generated an alert for the zpool
- gate all of our logic by running the `query_imported_fast_impl` method. This allows us to get the relevant information in a lockless manner without doing anything within our process pool
- remove unnecessary "self.enabled" method since it's not needed now that we're using `query_imported_fast_impl`


src/middlewared/middlewared/alert/source/volume_status.py
```diff
@@ -1,62 +1,70 @@
...
+    title = 'Boot Pool Is Not Healthy'
+    text = 'Boot pool status is %(status)s: %(status_detail)s.'
     proactive_support = True
 
 
 class VolumeStatusAlertSource(AlertSource):
+    async def check(self):
+        alerts = []
+        boot_name = await self.middleware.call('boot.pool_name')
+        for guid, info in (await self.middleware.call('zfs.pool.query_imported_fast')).items():
```
sonicaj, line 24: This is consuming the process pool ?

anodos325: The function is actually simple enough that we could probably move it to the ZFS utils file and import directly into the volume status alert.

yocalebo: sorry, I didn't see these but that's exactly what I'm doing.

src/middlewared/middlewared/alert/source/volume_status.py
```diff
@@ -1,62 +1,71 @@
...
+    def check_sync(self):
+        alerts = []
+        boot_name = self.middleware.call_sync('boot.pool_name')
+        for guid, info in query_imported_fast_impl().items():
+            if info['status'] == 'ONLINE':
+                continue
 
-    schedule = CrontabSchedule(hour=1)
+            try:
+                pool = self.middleware.call_sync('pool.query', [['name', '=', info['name']]], {'get': True})
```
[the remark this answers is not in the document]
yocalebo: done

sonicaj (approved):

PR #12000, opened 2023-08-30
NAS-123614 / 24.04 / Have a retry mechanism in realtime event source

Qubad786:
## Problem

We saw a rogue case where a timeout error happened for a user when trying retrieve netdata metrics which resulted in realtime event source erroring out.

## Solution

This error could not be reproduced but it was agreed with Caleb that we can have at least a retry logic in place to account for such weird cases in realtime reporting event source.

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/events.py
+++ b/src/middlewared/middlewared/plugins/reporting/events.py
@@ -63,7 +63,18 @@ class RealtimeEventSource(EventSource):
 
         while not self._cancel_sync.is_set():
             # this gathers the most recent metric recorded via netdata (for all charts)
-            netdata_metrics = self.middleware.call_sync('netdata.get_all_metrics')
+            retries = 2
+            while retries > 0:
+                try:
+                    netdata_metrics = self.middleware.call_sync('netdata.get_all_metrics')
+                except Exception:
+                    retries -= 1
+                    if retries <= 0:
+                        raise
+
+                    time.sleep(0.5)
+                else:
+                    break
 
             data = {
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/reporting/events.py
```diff
@@ -63,7 +63,16 @@ def run_sync(self):
 
         while not self._cancel_sync.is_set():
             # this gathers the most recent metric recorded via netdata (for all charts)
-            netdata_metrics = self.middleware.call_sync('netdata.get_all_metrics')
+            retries = 2
+            while retries > 0:
+                try:
+                    netdata_metrics = self.middleware.call_sync('netdata.get_all_metrics')
+                except Exception:
+                    retries -= 1
```
yocalebo, line 71: We should sleep an arbitrary amount of time instead of trying as fast as a cpu core allows.

src/middlewared/middlewared/plugins/reporting/events.py
```diff
@@ -63,7 +64,18 @@ def run_sync(self):
...
+                    if retries <= 0:
+                        raise
+
+                    time.sleep(random.randint(0, interval))
```
themylogin, line 76: Why sleep for 0 seconds?

yocalebo: agree, we don't need to do `random.randint(0, interval)`. We just need to sleep for arbitrary number each time, maybe 0.5 seconds

yocalebo (approved):

PR #12006, opened 2023-08-31
NAS-123733 / 24.04 / Check for host keys in user home directory when setting up a rsync task

themylogin:
with SSH key in user home directory.
Add `ssh_keyscan` option to automatically run `ssh_keyscan` when setting up a rsync task with SSH key in user home directory.


yocalebo (approved):

themylogin:
time 5:00

PR #12013, opened 2023-08-31
NAS-123980 / 24.04 / Add syslog-ng mako files and syslog tests

anodos325:
This will make it easier to precisely understand logging behavior and validate that logs are behaving properly.


tests/api2/test_475_syslog.py
```diff
@@ -0,0 +1,87 @@
...
+from time import sleep
+
+pytestmark = pytest.mark.skipif(dev_test, reason='Skipping for test development testing')
+
+
+def do_syslog(ident, message, facility='syslog.LOG_USER', priority='syslog.LOG_INFO'):
+    cmd = 'python3 -c "import syslog;'
+    cmd += f'syslog.openlog(ident=\\\"{ident}\\\", facility={facility});'
+    cmd += f'syslog.syslog({priority},\\\"{message}\\\");syslog.closelog()"'
+    results = SSH_TEST(cmd, user, password, ip)
```
themylogin, line 14: Why this and not `middlewared.test.integration.utils.ssh` ?

tests/api2/test_475_syslog.py
```diff
@@ -0,0 +1,87 @@
...
+    assert results['result'] is True, results['output']
+
+
+def check_syslog(log_path, message):
+    cmd = f'grep -R "{message}" {log_path}'
+    results = SSH_TEST(cmd, user, password, ip)
+    assert results['result'] is True, results['output']
+
+
+def validate_syslog_filter(data):
```
themylogin, line 24: Why define a function only used once?

anodos325: That's a fair criticism. I wrote the functions before writing tests.

tests/api2/test_475_syslog.py
anodos325: Wasn't aware you wrote a wrapper around it.

tests/api2/test_475_syslog.py
```diff
@@ -0,0 +1,92 @@
...
+    '/var/log/messages',
+    '/var/log/syslog',
+    '/var/log/daemon.log'
+])
+def test_filter_leak(request, log_path):
+    """
+    This test validates that our exclude filter works properly and that
+    particularly spammy applications aren't polluting useful logs.
+    """
+    results = ssh(f'grep -R "ZZZZ:" {log_path}', complete_response=True)
```
themylogin, line 91: `ssh` helper asserts successful exit code, pass `check=False` to skip this

tests/api2/test_475_syslog.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    Common function to check whether a particular message exists in a log file.
+    This will be used to check local and remote syslog servers.
+
+    Current implementation performs simple grep through the log file, and so
+    onus is on test developer to not under-specify `message` in order to avoid
+    false positives.
+    """
+    cmd = f'grep -R "{message}" {log_path}'
+    ssh(cmd, user=target_user, password=target_passwd, ip=target_ip)
```
themylogin, line 31: You don't need to pass this

themylogin (changes requested):

tests/api2/test_475_syslog.py
anodos325: We should have remote syslog server available in CI pipeline soon. Once this is in place the tests will be expanded to validate that messages are appearing properly on remote server.

themylogin (approved):
Thank you!

PR #12022, opened 2023-08-31
NAS-118579 / 24.04 / avahi-daemon is not announcing services

bmeagherix:
This PR is predominately to add some _basic_ unit tests wrt mDNS service announcement.  More may be added at a later date.

It includes one to check announcements following a reboot, but this is currently turned off (being perhaps too heavyweight for everyday use).

The python [`zeroconf`](https://pypi.org/project/zeroconf/) package is used to discover announcements from the TrueNAS under test.

### DIFF
--- a/src/middlewared/middlewared/plugins/service_/services/cifs.py
+++ b/src/middlewared/middlewared/plugins/service_/services/cifs.py
@@ -23,8 +23,16 @@ class CIFSService(SimpleService):
 
         await self._systemd_unit("smbd", "start")
 
+    async def after_start(self):
+        # We reconfigure mdns (add SMB service, possibly also ADISK)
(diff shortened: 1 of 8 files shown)


src/middlewared/middlewared/plugins/service_/services/cifs.py
```diff
@@ -23,8 +23,16 @@ async def start(self):
 
         await self._systemd_unit("smbd", "start")
 
+    async def after_start(self):
+        # We reconfigure mdns (add SMB service, possibly also ADISK)
+        await self.middleware.call('service.restart', 'mdns')
```
anodos325, line 28: I believe we already restart mDNS as needed in sharing.smb.update.

anodos325 (approved):

PR #12023, opened 2023-09-01
NAS-123865 / 24.04 / fix init_shutdown_script.py

yocalebo:
This plugin hasn't been touched in awhile and had some obvious problems.

- lots of blocking I/O in the main event loop
- self.middleware.logger instead of self.logger
- f-string in the self.logger calls
- unnecessary blank lines

### DIFF
--- a/src/middlewared/middlewared/plugins/init_shutdown_script.py
+++ b/src/middlewared/middlewared/plugins/init_shutdown_script.py
@@ -1,5 +1,6 @@
 import asyncio
 import os
+import stat
 import subprocess
 import tempfile
 
@@ -146,6 +147,7 @@ class InitShutdownScriptService(CRUDService):
     def write_script_text(self, script_text):
         with tempfile.NamedTemporaryFile(delete=False, mode='w') as f:
             f.write(script_text)
+            os.fchmod(f.fileno(), stat.S_IRWXU)
             return f.name
 
     @private
--- a/src/middlewared/middlewared/plugins/init_shutdown_script.py
+++ b/src/middlewared/middlewared/plugins/init_shutdown_script.py
@@ -1,6 +1,5 @@
 import asyncio
 import os
-import stat
 import subprocess
 import tempfile
 
@@ -58,8 +57,8 @@ class InitShutdownScriptService(CRUDService):
 
         There are three choices for `when`:
 
-        1) PREINIT - This is early in the boot process before all the services / rc scripts have started
-        2) POSTINIT - This is late in the boot process when most of the services / rc scripts have started
+        1) PREINIT - This is early in the boot process before all the services have started
+        2) POSTINIT - This is late in the boot process when most of the services have started
         3) SHUTDOWN - This is on shutdown
 
         `timeout` is an integer value which indicates time in seconds which the system should wait for the execution
@@ -69,16 +68,13 @@ class InitShutdownScriptService(CRUDService):
         by the base OS's limit.
         """
         await self.validate(data, 'init_shutdown_script_create')
-
         await self.init_shutdown_script_compress(data)
-
         data['id'] = await self.middleware.call(
             'datastore.insert',
             self._config.datastore,
             data,
             {'prefix': self._config.datastore_prefix}
         )
-
         return await self.get_instance(data['id'])
 
     async def do_update(self, id_, data):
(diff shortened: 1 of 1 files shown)


anodos325:
Webui doesn't consume the script text option (doesn't look like there are any consumers. Maybe remove from Cobia / Dragonfish.

src/middlewared/middlewared/plugins/init_shutdown_script.py
```diff
@@ -156,72 +143,71 @@ async def validate(self, data, schema_name):
...
 
         if task_type == 'COMMAND':
             cmd = task['command']
         elif task_type == 'SCRIPT' and task['script_text']:
-            fd, tmp_script = tempfile.mkstemp(text=True)
-            os.close(fd)
-            os.chmod(tmp_script, stat.S_IRWXU)
-            with open(tmp_script, 'w') as f:
-                f.write(task['script_text'])
+            tmp_script = await self.middleware.run_in_thread(self.write_script_text, task['script_text'])
```
themylogin, line 174: Why not just `middleware.call`?

src/middlewared/middlewared/plugins/init_shutdown_script.py
```diff
@@ -156,72 +143,71 @@ async def validate(self, data, schema_name):
...
+        return os.path.exists(script) and os.access(script, os.X_OK)
+
+    @private
+    def rem_tmp_script(self, script):
+        try:
+            os.unlink(script)
+        except FileNotFoundError:
+            pass
+        except Exception:
+            self.logger.debug('Unexpected failure removing %r', script, exc_info=True)
```
themylogin, line 164: No need to log the entire traceback here

themylogin (changes requested):

src/middlewared/middlewared/plugins/init_shutdown_script.py
yocalebo: blocking I/O in main event loop?

src/middlewared/middlewared/plugins/init_shutdown_script.py
```diff
@@ -156,72 +143,71 @@ async def validate(self, data, schema_name):
...
         elif task_type == 'SCRIPT' and task['script_text']:
-            fd, tmp_script = tempfile.mkstemp(text=True)
-            os.close(fd)
-            os.chmod(tmp_script, stat.S_IRWXU)
-            with open(tmp_script, 'w') as f:
-                f.write(task['script_text'])
+            tmp_script = await self.middleware.run_in_thread(self.write_script_text, task['script_text'])
             cmd = f'exec {tmp_script}'
-        elif os.path.exists(task['script'] or '') and os.access(task['script'], os.X_OK):
+        elif await self.middleware.run_in_thread(self.script_already_exists, task['script']):
```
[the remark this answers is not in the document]
yocalebo: it exists??

src/middlewared/middlewared/plugins/init_shutdown_script.py
yocalebo: Why?

src/middlewared/middlewared/plugins/init_shutdown_script.py
themylogin: `middleware.call` is an async method and it does not block I/O

yocalebo: `write_script_text` calls blocking I/O?

src/middlewared/middlewared/plugins/init_shutdown_script.py
themylogin: Because we are not interested in who called `rem_tmp_script` (we only have one caller), we are only interested in file path and I/O error.

src/middlewared/middlewared/plugins/init_shutdown_script.py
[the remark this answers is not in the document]
themylogin: Why `script_already_exists` and not just `script_exists`? `already` makes reader think we do this check to avoid creating something that already exists which is not the case here.

src/middlewared/middlewared/plugins/init_shutdown_script.py
yocalebo: I don't like adding the overhead of `middleware.call` for a class method that sits right next to the calling method

yocalebo:
superseded by https://github.com/truenas/middleware/pull/12030

PR #12026, opened 2023-09-01
NAS-123782 / 24.04 / Update default recordsize for draid pool based datasets/zvols

Qubad786:
## Context

A request was made to redefine default values and minimum values for recordsize/volblocksize attrs of a dataest/zvol for draid based pools.

### DIFF
--- a/tests/api2/test_draid_record_and_block_size.py
+++ b/tests/api2/test_draid_record_and_block_size.py
@@ -1,0 +1,177 @@
+import pytest
+
+from middlewared.service_exception import ValidationErrors
+from middlewared.test.integration.assets.pool import another_pool
+from middlewared.test.integration.utils import call
+
+
+if len(call('disk.get_unused')) < 2:
+    pytest.skip('Insufficient number of disk to perform these test')
+
+
+@pytest.fixture(scope='module')
+def draid_pool():
+    unused_disks = call('disk.get_unused')
+    with another_pool({
+        'name': 'test_draid_pool',
+        'topology': {
+            'data': [{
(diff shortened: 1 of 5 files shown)


yocalebo:
@truenas/middleware ping

themylogin (approved):

yocalebo:
retest this please

tests/api2/test_draid_record_and_block_size.py
```diff
@@ -0,0 +1,188 @@
...
+                'type': pool_type,
+                **({'draid_data_disks': 1} if pool_type.startswith('DRAID') else {})
+            }],
+        },
+        'allow_duplicate_serials': True,
+    }) as pool_name:
+        assert call('pool.dataset.get_instance', pool_name['name'])['recordsize']['value'] == record_size
+
+
+@pytest.mark.parametrize(
```
themylogin, line 36: We will have a relatively large number of pool creation/deletion operations here, running these tests might take up a very long time. Can we test the complete recordsize logic with an integration test and here only test a part of it?

themylogin (changes requested):

tests/api2/test_draid_record_and_block_size.py
Qubad786: I'll use a fixture here so we avoid creating a pool each time these tests are run and that should solve this problem ✅

yocalebo: Agree with @themylogin's review. When you use a fixture, make sure that the pool is cleaned up after this module is run. Thanks!

themylogin (approved):

PR #12035, opened 2023-09-04
NAS-123916 / 24.04 / Fix failing integration tests

sonicaj:


src/middlewared/middlewared/plugins/rsync.py
```diff
@@ -291,6 +292,9 @@ async def validate_rsync_task(self, data, schema):
                                 errors='ignore',
                             )).stdout
 
+                            # If for whatever reason the dir does not exist, let's create it
+                            # An example of this is when we run rsync tests we nuke the directory
+                            await self.middleware.run_in_thread(ssh_dir_path.mkdir, 0o700, False, True)
```
themylogin, line 297: Should we chown the directory too?

tests/api2/test_380_rsync.py
```diff
@@ -31,7 +31,8 @@ def test_04_Creating_rsync_task(rsynctask_dict):
         'path': '/mnt/tank/share',
         "remotepath": "/share",
         "remoteport": 22,
-        "validate_rpath": False
+        "validate_rpath": False,
+        "ssh_keyscan": True,
```
themylogin, line 35: Why?

tests/api2/test_rsync_ssh_authentication.py
```diff
@@ -229,6 +228,7 @@ def test_ssh_credentials_key(cleanup, localuser, remoteuser, src, dst, ssh_crede
         "ssh_credentials": ssh_credentials["credentials"]["id"],
         "mode": "SSH",
         "remotepath": dst,
+        "ssh_keyscan": True,
```
themylogin, line 231: Why? Should not be necessary because we are using a keychain credential that already contains the host key.

themylogin (changes requested):

src/middlewared/middlewared/plugins/pool_/dataset.py
```diff
@@ -881,6 +881,9 @@ async def do_delete(self, id_, options):
         `recursive` will also delete/destroy all children datasets.
         `force` will force delete busy datasets.
 
+        When root dataset is specified as `id` with `recursive`, it will destroy all the children of the
```
yocalebo, line 884:
```suggestion
        When root dataset is specified as `id_` with `recursive`, it will destroy all the children of the
```

yocalebo (changes requested):

src/middlewared/middlewared/plugins/pool_/dataset.py
sonicaj: We would be using `id` because that is what schema reflects. It does not matter what we are using in the actual parameter name in the function.

src/middlewared/middlewared/plugins/rsync.py
```diff
@@ -294,7 +294,13 @@ async def validate_rsync_task(self, data, schema):
 
                             # If for whatever reason the dir does not exist, let's create it
                             # An example of this is when we run rsync tests we nuke the directory
-                            await self.middleware.run_in_thread(ssh_dir_path.mkdir, 0o700, False, True)
+                            if not await self.middleware.run_in_thread(ssh_dir_path.exists):
```
yocalebo, line 297: We're creating 5 separate threads to perform operations that can be done in a single thread. Also, this is a race condition to check for existence before creating.

```
def handle_ssh_dir(self, ssh_dir_path, known_hosts_path, known_hosts_text, uid, gid):
    try:
        ssh_dir_path.mkdir(0o700)
        os.chown(ssh_dir_path, uid, gid)
    except FileExistsError:
        pass
  
    known_hosts_path.write_text(known_hosts_text)
    os.chown(known_hosts_path, uid, gid)

await self.middleware.run_in_thread(
    self.handle_ssh_dir, ssh_dir_path, known_hosts_path, known_hosts_text, user['pw_uid'], user['pw_gid']
)

**Review by yocalebo — CHANGES_REQUESTED** [review]

File: src/middlewared/middlewared/plugins/rsync.py [hunk 6]
```diff
@@ -294,16 +294,15 @@ async def validate_rsync_task(self, data, schema):
...
-                            if not await self.middleware.run_in_thread(ssh_dir_path.exists):
-                                # We still do a safe mkdir to avoid any race conditions
-                                await self.middleware.run_in_thread(ssh_dir_path.mkdir, 0o700, False, True)
-                                await self.middleware.run_in_thread(
-                                    os.chown, ssh_dir_path.absolute(), user['pw_uid'], user['pw_gid']
-                                )
+                            def handle_ssh_dir():
+                                with contextlib.suppress(FileExistsError):
+                                    ssh_dir_path.mkdir(0o700)
+                                    os.chown(ssh_dir_path.absolute(), user['pw_uid'], user['pw_gid'])
```
yocalebo, line 300: This os.chown call will have to be moved outside of the scope of the `with contextlib` since if it crashes with `FileExistsError`, the `os.chown` won't be made

yocalebo (changes requested):

src/middlewared/middlewared/plugins/rsync.py
themylogin: This is intended, why chown directory if it already exists?

yocalebo: Because it could be owned by a different uid/gid

src/middlewared/middlewared/plugins/rsync.py
```diff
@@ -291,9 +292,17 @@ async def validate_rsync_task(self, data, schema):
...
+                            # An example of this is when we run rsync tests we nuke the directory
+                            def handle_ssh_dir():
+                                with contextlib.suppress(FileExistsError):
+                                    ssh_dir_path.mkdir(0o700)
+                                    os.chown(ssh_dir_path.absolute(), user['pw_uid'], user['pw_gid'])
+
+                                known_hosts_path.write_text(known_hosts_text)
+                                os.chown(known_hosts_path.absolute(), user['pw_uid'], user['pw_gid'])
+
+                            await self.middleware.run_in_thread(handle_ssh_dir)
```
themylogin, line 305: It was actually not necessary to complicate things here, `run_in_thread` reuses threads

tests/api2/test_rsync_ssh_authentication.py
```diff
@@ -244,6 +243,7 @@ def test_ssh_credentials_delete(cleanup, localuser, remoteuser, src, dst):
             "ssh_credentials": c["credentials"]["id"],
             "mode": "SSH",
             "remotepath": dst,
+            "ssh_keyscan": True,
```
themylogin, line 246: Please remove it everywhere where `ssh_credentials` are used

themylogin (changes requested):

src/middlewared/middlewared/plugins/rsync.py
yocalebo: "complicate" is relative here. `run_in_thread` reusing threads implies the developer has a deep understanding of how our threadpool operates. What I suggested, instead, clearly shows that a single function will be called in a single thread.

yocalebo (approved):
Looks good to me but please wait for Vlad's final review since he's much more knowledgeable in this area than I am 😄

themylogin (approved):

src/middlewared/middlewared/plugins/rsync.py
```diff
@@ -310,7 +310,7 @@ def handle_ssh_dir():
                         try:
                             async with await asyncssh.connect(
                                 **connect_kwargs,
-                                known_hosts=known_hosts,
+                                known_hosts=known_hosts if not data['ssh_credentials'] else None,
```
themylogin, line 313: We need to specify known_hosts in any case so that we verify that they are valid

themylogin (changes requested):

src/middlewared/middlewared/plugins/rsync.py
```diff
@@ -599,6 +615,7 @@ def commandline(self, id_):
                         for host_key in credentials['remote_host_key'].split("\n")
                         if host_key.strip() and not host_key.strip().startswith("#")
                     ]))
+                    host_key_file.write(get_host_key_file_contents_from_ssh_credentials(credentials))
```
themylogin, line 618: What's wrong with the code above?

tests/api2/test_rsync_ssh_authentication.py
```diff
@@ -199,7 +199,6 @@ def test_ssh_keyscan_does_not_duplicate_host_keys(cleanup, localuser, remoteuser
 
 
 def test_home_directory_key(cleanup, localuser, remoteuser, src, dst, ssh_credentials):
-    ssh(f"mkdir {localuser['home']}/.ssh")
```
themylogin, line 202: Why? Won't `file_receive` below crash if the directory does not exist?

themylogin (changes requested):

tests/api2/test_rsync_ssh_authentication.py
```diff
@@ -199,7 +199,7 @@ def test_ssh_keyscan_does_not_duplicate_host_keys(cleanup, localuser, remoteuser
...
+    ssh(f"mkdir -p {localuser['home']}/.ssh")
```
themylogin, line 202: Why `-p`, how can `localuser['home']` be non-existent?

themylogin (approved):

PR #12042, opened 2023-09-06
NAS-123942 / 24.04 / Remove older implementation of ixdiagnose

Qubad786:
## Context

With truenas/ixdiagnose now being used to generate debug, we should remove older implementation of ixdiagnose from middleware.


yocalebo (approved):
SHIP IT! 😄

PR #12047, opened 2023-09-06
NAS-123905 / 24.04 / add blackhole route for heartbeat iface on HA

yocalebo:
The fseries platform uses the AMD ntb driver. This driver behaves differently than the intel NTB driver in that the B controller is active and the A controller reboots, the `ntb0` interface is removed from the B controller. When this happens, any traffic on the 169.254.10/24 subnet will be forwarded to the default gateway. We don't want to do this so we add a RTN_BLACKHOLE route for 169.254.10.0/23. This route is 1 bit larger than the subnet we put on the interface so it will only take effect when the B controller is active and the `ntb0` interface is removed from the OS.

### DIFF
--- a/src/middlewared/middlewared/plugins/failover_/internal_interface.py
+++ b/src/middlewared/middlewared/plugins/failover_/internal_interface.py
@@ -33,7 +33,6 @@ class InternalInterfaceService(Service):
             return []
 
     async def pre_sync(self):
-
         if not await self.middleware.call('system.is_enterprise'):
             return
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/failover_/internal_interface.py
```diff
@@ -56,10 +55,26 @@ async def pre_sync(self):
...
+            # ip address we put on the internal interface. We do this because the
+            # f-series platform uses AMD ntb driver and the behavior for when the
+            # B controller is active and the A controller reboots, is that the ntb0
+            # interface is removed from the B controller. This means any src/dst
+            # traffic on the 169.254.10/24 subnet will be forwarded out of the gateway
+            # of last resort (default route). Since this is internal traffic, we
+            # obviously don't want to forward this traffic to the default gateway.
+            # This just routes the data into oblivion (drops it).
+            try:
+                ndb.routes.create(dst='169.254.10.0/23', table=default_table, type=rtn_blackhole).commit()
```
mgrimesix, line 77: Should we manipulate `internal_ip` instead of hard-coding the address?

mgrimesix (approved):
excellent!

PR #12049, opened 2023-09-06
NAS-123731 / 24.04 / Remove haproxy from TrueNAS

sonicaj:
This PR adds changes removing haproxy from TN and instead we now rely directly on nginx to handle vm display devices uris. Motivation behind this change is to avoid resource usage and also reduce dependancies of TN so we reduce the surface area of security attacks as well.


yocalebo (approved):
This looks good from a code standpoint. I trust you'll merge after you test 😄

PR #12050, opened 2023-09-06
NAS-123960 / 24.04 / Add ability to convert mountinfo to tree

anodos325:
This provides validated method of generating tree for checking whether unexpected filesystems or ZFS dataset parameters are mounted under a given path. Example of how one would do this would be:

getmnttree(os.lstat('/mnt/dozer/SHARE').st_dev)

### DIFF
--- a/src/middlewared/middlewared/utils/osc/linux/mount.py
+++ b/src/middlewared/middlewared/utils/osc/linux/mount.py
@@ -62,6 +62,9 @@ def __create_tree(info, dev_id):
         else:
             parent['children'].append(entry)
 
+    if dev_id and not mount_id:
+        raise KeyError(f'{dev_id}: device not in mountinfo')
+
     return info[mount_id or root_id]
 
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/utils/osc/linux/mount.py
```diff
@@ -27,30 +28,79 @@ def __parse_mntent(line, out_dict):
...
+            callback(line, private_data)
+
+
+def getmntinfo(dev_id=None):
+    """
+    Get mount information. returns dictionary indexed by dev_t.
+    User can optionally specify dev_t for faster lookup of single
+    device.
+    """
+    out = {}
```
yocalebo, line 92: Let's make this similar to how it's being done in `getmnttree`.

- remove unnecessary empty newlines
- change `out` to info

yocalebo (changes requested):

yocalebo (approved):

PR #12073, opened 2023-09-11
NAS-123910 / 24.04 / Roles

themylogin:
We would like to have 3 different roles in API to begin with:

- Full Admin: just like root/admin
- Read-only: Access to all readonly endpoints (e.g. query, choices, get, etc). No access to any data changing methods (create, update, delete)
- Shares Manager: Be able to create/edit datasets and shares. No be able to do any other system admin tasks nor delete/destroy anything.

More information is in the ticket.


themylogin:
time 4:00

themylogin:
@bmeagherix @anodos325 ping

anodos325:
Does this mean that a readonly user will have read access to the user DB table?

anodos325 (changes requested):
We should have some handling for private fields in READONLY role otherwise there is potential for privilege escalation / unintended disclosure of information.

bmeagherix (changes requested):
What is there looks very good, but I agree with Andrew's comment:
> We should have some handling for private fields in READONLY role otherwise there is potential for privilege escalation / unintended disclosure of information.

(For example, once we add a READ role covering it, I could envisage wanting to redact CHAP credentials from `iscsi.auth.query`.)

Have we worked out yet how the webui will behave for a user with READONLY role?

Wrt tests, IMO it would be good to add success/failure method calls in nfs, smb & iscsi for SHARING_READ role.  Also success calls (using the same methods) for SHARING_WRITE role.

yocalebo:
After some discussion, hiding `private=True` fields from `query` endpoints is a huge non-trivial change that needs to happen in another PR. This PR just adds the exact items that were requested from management so we're going to hold off on hiding sensitive fields from query endpoints. I will open a separate ticket and link it to this one to implement said feature. For now, let's forgo that thought process and get another round of reviews please @anodos325 @bmeagherix

bmeagherix (changes requested):
Suggest adding tests for success/failure method calls in nfs, smb & iscsi for SHARING_READ role.

Also success calls (using the same methods) for SHARING_WRITE role.

anodos325:
I think we should add privilege table entries for each of the roles that don't already exist (we already have an admin one with full control).

anodos325:
I think we should also expand the dump of logged-in user `auth.me` to include information about privileges assigned to user (based on group membership).

themylogin:
> I think we should add privilege table entries for each of the roles that don't already exist (we already have an admin one with full control).

@yocalebo what do you think of this? Looks questionable, we'll have a ton of useless entries by default.

anodos325:
> > I think we should add privilege table entries for each of the roles that don't already exist (we already have an admin one with full control).
> 
> @yocalebo what do you think of this? Looks questionable, we'll have a ton of useless entries by default.

Well, maybe not all of them, but one or two that cover some basic common use-cases (like the READONLY access). In that way it can kind of act as template for users to think "oh this is how it's done". Just a thought.

themylogin:
I agree on adding "readonly user" built-in privilege. @yocalebo ?

yocalebo:
@themylogin yeah adding read only is a good first step for now

src/middlewared/middlewared/alembic/versions/24.04/2023-10-30_21-02_readonly_privilege.py
```diff
@@ -0,0 +1,31 @@
...
+down_revision = 'e3436f15aa9c'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.execute(
+        "INSERT INTO account_privilege (builtin_name, name, local_groups, ds_groups, allowlist, web_shell, roles) "
+        "VALUES ('READONLY', 'Readonly Users', '[]', '[]', '[]', 0, '[\"READONLY\"]')"
```
anodos325, line 23: I believe per our previous conversations about sensitive info being exposed to READONLY users that we may want to explicitly name something like "READONLY ADMINISTRATORS", correct?

yocalebo (approved):
Great job but please wait for @anodos325 and @bmeagherix approval before merge.

anodos325 (approved):

themylogin:
time 8:00

PR #12075, opened 2023-09-11
NAS-124058 / 24.04 / Stop doing runtime modifications of files in /usr

anodos325:
This PR makes several changes in preparation for using separate datasets for relevant portions of the root filesystem heirarchy.

* Stop permissions checks on stub SMB configuration
* Make static netdata configuration changes at install time rather than runtime.
* Move local ca certificates path to /var/local rather than /usr/local in accordance with Linux FHS guidelines. Original path (/usr/local/share/ca-certificates) will be replaced with symlink to /var/local/ca-certificates.


anodos325:
This is related to https://github.com/truenas/scale-build/pull/494

src/freenas/debian/preinst
```diff
@@ -3,6 +3,8 @@
 mkdir -p /var/trash
 for file in \
     /etc/nsswitch.conf \
+    /etc/nsswitch.conf \
```
themylogin, line 6: Duplicate line?

src/freenas/usr/bin/unlock-root-fs.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/python3
+
+import json
+import sys
+from subprocess import run
+
+
+def unlock_dataset(entry):
```
themylogin, line 8: Lock/unlock are the terms used for encryption. Can we come up with another term for toggling readonly flag on the filesystem?

src/middlewared/middlewared/etc_files/generate_ssl_certs.py
```diff
@@ -26,9 +26,8 @@ def write_certificates(certs: list, cacerts: list) -> set:
             with open(cert['csr_path'], 'w') as f:
                 f.write(cert['CSR'])
 
-    trusted_cas_path = '/usr/local/share/ca-certificates'
+    trusted_cas_path = '/var/local/ca-certificates'
     shutil.rmtree(trusted_cas_path, ignore_errors=True)
-    os.makedirs(trusted_cas_path)
```
themylogin, line 31: Why? How will `f.write` call below work if we just removed the directory above?

src/freenas/usr/bin/unlock-root-fs.py
```diff
@@ -0,0 +1,38 @@
...
+        return
+
+    print(f'Setting readonly=off on dataset {entry["ds"]}')
+    run(['zfs', 'set', 'readonly=off', entry['ds']])
+
+
+if __name__ == '__main__':
+    datasets = None
+    try:
+        with open('/conf/truenas_root_ds.conf', 'r') as f:
```
themylogin, line 23: What is this file? How is it created?

src/middlewared/middlewared/etc_files/generate_ssl_certs.py
anodos325: It's a dataset.

src/freenas/usr/bin/unlock-root-fs.py
anodos325: It is created during TrueNAS install process c.f. scale-build PR.

src/freenas/usr/bin/unlock-root-fs.py
anodos325: Sure.

anodos325: I renamed the script as `truenas-developer-enable.py`

src/freenas/usr/bin/truenas-developer-enable.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # There shouldn't be a legitimate reason to edit files in /conf
+    if entry['fhs_entry']['name'] == 'conf':
+        return
+
+    print(f'Setting readonly=off on dataset {entry["ds"]}')
+    run(['zfs', 'set', 'readonly=off', entry['ds']])
+
+
+def chmod_files():
```
themylogin, line 25: Why do we need 20 files of python code instead of 2 lines of bash code?

src/middlewared/middlewared/etc_files/generate_ssl_certs.py
themylogin: Please leave a comment then, it's not obvious what is happening here

anodos325: Okay. Done.

src/freenas/usr/bin/truenas-developer-enable.py
anodos325: My early morning thought (pre-coffee) was that users may want to enable apt, dpkg, and set readonly=off independently of installing the dev tools (idea being that keeping actions to absolute minimum will encourage tinkerers to do this rather than writing a separate how-to), which will make it easier to identify via custom ZFS property. Is there a particular reason this is unacceptable (other than using more lines of code)?

themylogin: Let's do it with subprocess call then. The less code the clearer in this case. Current python code forces all permissions to 0755 which is definitely not the same as `chmod +x`

themylogin (changes requested):

src/freenas/usr/bin/truenas-developer-enable.py
anodos325: I changed the os.chmod call to do +x.

anodos325:
The zectl flag change depends on 
https://github.com/truenas/zectl/pull/12
and
https://github.com/truenas/zectl/pull/11

Which should be merged tomorrow.

themylogin (approved):

yocalebo:
@Qubad786 can you review this one since it's making changes to netdata configuration? This is the better approach to take, but I want to make sure you look at it.

Qubad786 (approved):

PR #12077, opened 2023-09-11
NAS-123668 / 24.04 / Enable exporting reporting data to graphite

Qubad786:
## Context

It was requested that we add ability to export reporting data to graphite similar to what we used to do before and while we are doing this also make sure the approach is implementation agnostic so we can add more relevant export types as well.

Here changes have been made to the reporting system to ensure that the configuration aligns with the use of Netdata for reporting. Specifically, the reporting system is updated to support Netdata's metrics export to Graphite, as Collectd is no longer used. This involves creating a new endpoint within the reporting system, such as `reporting.exporters`, which exposes methods for creating, deleting, and updating Netdata export configurations.

The following parameters should be specified to enable Graphite as the reporting destination for Netdata:

1. **enable**: This parameter enables a specific instance of the export.
2. **name**: It specifies a unique instance name for Netdata. If multiple instances are configured, each must have a distinct name.
3. **attributes.destination_ip**: This parameter sets the IP address of the Graphite server.
4. **attributes.destination_port**: It specifies the port on which the Graphite server is listening.
5. **attributes.prefix**: This parameter defines a prefix that will be prepended to all Netdata metrics sent to Graphite.
6. **attributes.hostname**: It is used to specify the hostname to append to the metrics after the prefix.
7. **attributes.update_every**: This parameter determines the interval at which metrics are updated.
8. **attributes.buffer_on_failures**: It specifies the tolerance for update failures.
9. **attributes.send_names_instead_of_ids**: This option controls whether to send chart names instead of IDs.
10. **attributes.matching_charts**: This parameter allows specifying a regular expression for Netdata charts that should be sent to the Graphite server.
11. **type**: Currently, only `GRAPHITE` is supported. This parameter simplifies the integration of other exporters available in Netdata.

By implementing these changes and ensuring that the reporting system is aligned with Netdata's Graphite export configuration, the reporting process will be accurate and efficient and the approach is generic too so we can add more exporters in the future with ease.

## References

- [Netdata Documentation](https://learn.netdata.cloud/docs/agent/collectors/plugins.d.plugin/graphite.plugin)
- [Netdata Exporting Ref](https://learn.netdata.cloud/docs/exporting/exporting-reference)

### DIFF
--- a/src/middlewared/middlewared/alembic/versions/24.04/2023-09-07_19-50_add_exporting_table_for_reporting.py
+++ b/src/middlewared/middlewared/alembic/versions/24.04/2023-09-07_19-50_add_exporting_table_for_reporting.py
@@ -1,0 +1,61 @@
+"""Add exporting table for reporting
+
+Revision ID: 3e16b0a74d78
+Revises: e915a3b8fff6
+Create Date: 2023-09-07 19:50:42.678757+00:00
+
+"""
+import json
+import sqlalchemy as sa
+
+from alembic import op
+from sqlalchemy.sql import text
+
+
+revision = '3e16b0a74d78'
+down_revision = 'e915a3b8fff6'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'reporting_exporters',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('enabled', sa.Boolean(), nullable=False),
+        sa.Column('type', sa.String(), nullable=False),
+        sa.Column('name', sa.String(), nullable=False),
+        sa.Column('attributes', sa.TEXT(), nullable=False),
+        sa.PrimaryKeyConstraint('id', name=op.f('pk_reporting_exports'))
+    )
+
+    conn = op.get_bind()
+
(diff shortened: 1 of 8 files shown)


yocalebo (approved):
This looks good to me, however, please wait for Vlad's review since it's touching database.

src/middlewared/middlewared/alembic/versions/24.04/2023-09-07_19-50_add_exporting_table_for_reporting.py
```diff
@@ -0,0 +1,53 @@
...
+            'prefix': 'dragonfish',
+            'hostname': 'truenas',
+            'update_every': 1,
+            'buffer_on_failures': 10,
+            'send_names_instead_of_ids': True,
+            'matching_charts': '*'
+        }
+        op.execute(
+            'INSERT INTO reporting_exporters (enabled, type, name, attributes) VALUES'
+            f"(true, 'GRAPHITE', 'netdata', '{json.dumps(attributes)}')"
```
themylogin, line 46: If `graphite_ip[0]` for some reason contains `'`, it will be an SQL error. Please use bind parameters instead

themylogin (changes requested):

yocalebo:
@themylogin ping

themylogin (approved):

Qubad786:
backport

PR #12078, opened 2023-09-11
NAS-124053 / 24.04 / Fix `script` schema arg in initshutdownscript plugin

yocalebo:
The integration tests were failing because an empty `File` was being passed to our schema class and it was running `isfile` on an empty string. This is a regression but it also pointed out that we don't do near enough validation in this plugin when a script is provided so this does 2 primary things:
1. convert `script` to just a plain `Str` and do proper validation on this value inside `validate` method
2. remove the old integration tests for this plugin and add integration tests that test 1 of the failure scenarios when a bad script path is given


tests/api2/test_initshutdownscript.py
```diff
@@ -0,0 +1,81 @@
...
+_775 = stat.S_IRWXU | stat.S_IRWXG | stat.S_IROTH | stat.S_IXOTH
+
+
+@pytest.fixture(scope='module')
+def ip_to_use():
+    return ip if not ha else os.environ['controller1_ip']
+
+
+@pytest.fixture(scope='module')
+def ws_client(ip_to_use):
```
themylogin, line 23: Why is this necessary?

tests/api2/test_initshutdownscript.py
```diff
@@ -0,0 +1,81 @@
...
+    with client(host_ip=ip_to_use) as c:
+        yield c
+
+
+def test_initshtudown_script(ws_client):
+    # create the test script file first
+    ws_client.call(
+        'filesystem.file_receive',
+        TEST_SCRIPT_FILE,
+        base64.b64encode(b'echo "testing"').decode(),
```
themylogin, line 33: You can just use `ssh` command to create simple files on the NAS

tests/api2/test_initshutdownscript.py
yocalebo: So we only establish 1 websocket session for the scope of the module since there are 2 tests being executed (`test_initshutdown_script` and `test_initshutdown_script_bad`)

tests/api2/test_initshutdownscript.py
yocalebo: That's true but I'm trying to stick to only using websocket client for this module

tests/api2/test_initshutdownscript.py
yocalebo: Do you mean why is `ip_to_use` fixture necessary?

yocalebo: Should be fixed now.

themylogin: YOu forgot to remove `ip_to_use` argument

themylogin (changes requested):

tests/api2/test_initshutdownscript.py
yocalebo: oops, fixed

themylogin (approved):

PR #12093, opened 2023-09-13
NAS-124088 / 24.04 / Fix dumping `List` that is `None`

themylogin:


themylogin:
time 0:10

yocalebo:
Should we add unit tests to test this specific scenario?

src/middlewared/middlewared/pytest/unit/test_schema.py
```diff
@@ -496,6 +496,17 @@ def listnotnull(self, data):
...
 
+def test__schema_list_null_items():
+
+    @accepts(List('data', null=True, items=[Str('item')]))
+    def listnull(self, data):
+        return data
+
+    self = Mock()
+
+    listnull(self, None)
```
yocalebo, line 507: Sorry, I must me misunderstanding but how does this test anything if we're not checking return value of `listnull`?

yocalebo: Ever other test in this file checks return values and/or if an error is raised. Either:
1. add a comment on why this test is different
2. or actually do `listnull(self, None) == []` and add a `TODO:` comment so we can come back to this and fix it later

yocalebo (approved):
Thank you!

PR #12103, opened 2023-09-14
NAS-124118 / 24.04 / optimize reporting/events.py

yocalebo:
I generated a flame graph of the main middleware process on an internal system with ~1.2k disks. The flame graph included 40,293 samples. Of those samples, 27,837 (69.09%) of them were spent on line 85 in the `reporting/events.py` script.

After looking at line 85, I realized that we're casting the dictionary to a list which just pulls out the keys of the dictionary. The keys of the dictionary from `device.get_disks` are simply block device names for the disks on the system. This is too expensive to call just to get the names of the disks so I added a new function `device.get_disk_names`.

The function that I've added is ~150% faster than the old method. The returned output is the exact same.

### DIFF
--- a/src/middlewared/middlewared/plugins/device_/device_info.py
+++ b/src/middlewared/middlewared/plugins/device_/device_info.py
@@ -1,7 +1,9 @@
+import pathlib
 import re
 
-import libsgio
 import pyudev
+
+import libsgio
 from middlewared.plugins.disk_.enums import DISKS_TO_IGNORE
 from middlewared.schema import Dict, returns
 from middlewared.service import Service, accepts, private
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/device_/device_info.py
```diff
@@ -41,6 +44,34 @@ def get_disk_serial(self, dev):
...
         )
 
+    @private
+    def get_disk_names(self):
+        """This endpoint serves almost exclusively to be called in our
+        reporting plugin. It just needs the block device names
+        (sda/nvme0n1/pmem0/etc) and so this will very quickly enumerate
+        that information.
+
+        NOTE: The return of this method should match the keys retrieved
```
themylogin, line 54: Can you add an integration test for that?

themylogin (changes requested):

themylogin (approved):

PR #12122, opened 2023-09-18
NAS-124155 / 24.04 / Adapt `test_system_general_ui_allowlist` to VM changes

themylogin:
We used expect GET /vm/display to return 403 when Web UI access is restricted. It does not return it anymore because this URL is not served by nginx (we now have individual rules per-vm). That's why we are now probing /vm/display/1 (and we mock the VM to have this rule in nginx configuration)

### DIFF
--- a/tests/api2/test_system_general_ui_allowlist.py
+++ b/tests/api2/test_system_general_ui_allowlist.py
@@ -4,7 +4,7 @@ import time
 import requests
 import websocket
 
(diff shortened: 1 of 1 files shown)


themylogin:
time 1:00

yocalebo (changes requested):
Need a comment, a link, a past PR or some general statement on what this is doing and why it fixes the problem.

yocalebo (approved):
Thank you!

PR #11820, opened 2023-08-09
NAS-123999 / 24.04 / Add parser for audit messages

anodos325:
This PR adds a syslog-ng parser that converts audit messages
generated by different applications into insertions to
per-service auditing sqlite3 databases.

Messages should be generated

```
syslog(config->syslog_priority | config->syslog_facility,
       "@cee:{\"TNAUDIT\": %s}", msg);
```

with syslog ident set to "TNAUDIT_<service name>". It would
probably be best to keep the name short. Example `TNAUDIT_SMB`.

Once message is received it will be inserted into the per-service
database. Example entry:

```
       aid = 0cc9306c-51f1-48aa-ac2c-3fffb2975e26
      vers = 0.1
      time = 2023-08-30 14:19:20.564154Z
      addr = 127.0.0.1
      user = smbuser
      sess = aff12102-14e5-4bc7-b94c-3e6e06049233
       svc = SMB
  svc_data = {"vers": {"major": 0, "minor": 1}, "service": "AUDIT_TEST", "session_id": "112026400", "tcon_id": "2906164554"}
     event = CONNECT
event_data = {"host": "127.0.0.1", "unix_token": {"uid": 1000, "gid": 1000, "groups": [545, 1000, 90000005, 90000012, 90000017]}, "result": {"type": "UNIX", "value_raw": 0, "value_parsed": "SUCCESS"}, "vers": {"major": 0, "minor": 1}}
   success = 1
```

PR #12129, opened 2023-09-19
NAS-123955 / 24.04 / Expose SMB auditing feature

anodos325:
This PR expands the SMB share schema to add SMB auditing configuration parameters:

`enable` - enable auditing for the SMB share
`watch_list` - optional parameter to restrict auditing to
    a list of groups. SMB ops will be audited if user token
    contains one of listed groups.
`ignore_list` - optional parameter to skip auditing for a
    a list of groups. SMB ops will _not_ be audited if user
    token contains one of listed groups.

NOTE: if user token contains group in watch_list and in ignore_list, then watch_list takes precedence.


anodos325:
Depends on this PR https://github.com/truenas/middleware/pull/11820

yocalebo (changes requested):
Will need to have updated db migration.

yocalebo (approved):

PR #12140, opened 2023-09-21
NAS-121831 / 24.04 / Add script explaining that package management tools are disabled

bmeagherix:
Add a script `pkg_mgmt_disabled` and suitable soft links for the following executables:

```
root@truenas[/usr/local/bin]# ls -l /usr/bin | grep '^-rw-' | awk '{print $9}'
apt
apt-cache
apt-cdrom
apt-config
apt-extracttemplates
apt-ftparchive
apt-get
apt-key
apt-mark
apt-sortpkgs
dpkg
```


bmeagherix:
Removing WIP now that https://github.com/truenas/scale-build/pull/520 is merged

yocalebo (approved):

PR #12173, opened 2023-09-25
NAS-124312 / 24.04 / Add flag outlining if cpu temperatures are being retrieved

Qubad786:
**Motivation:**

The UI necessitates a flag to indicate the availability of cpu temperature data.

**Change:**

This PR addresses the requirement by setting the `cpu_temperatures_retrieved` flag to True when temperature values are available for machine.

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/events.py
+++ b/src/middlewared/middlewared/plugins/reporting/events.py
@@ -88,8 +88,7 @@ class RealtimeEventSource(EventSource):
             }
 
             # CPU temperature
-            data['cpu']['temperature_celsius'] = self.middleware.call_sync('reporting.cpu_temperatures')
-            data['cpu']['temperature'] = {k: 2732 + int(v * 10) for k, v in data['cpu']['temperature_celsius'].items()}
+            data['cpu']['temperature_celsius'] = self.middleware.call_sync('reporting.cpu_temperatures') or None
 
             self.send_event('ADDED', fields=data)
             time.sleep(interval)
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/reporting/events.py
```diff
@@ -91,6 +91,9 @@ def run_sync(self):
             data['cpu']['temperature_celsius'] = self.middleware.call_sync('reporting.cpu_temperatures')
             data['cpu']['temperature'] = {k: 2732 + int(v * 10) for k, v in data['cpu']['temperature_celsius'].items()}
 
+            # Flag telling whether the CPU temperatures' data is available
+            data['cpu_temperatures_retrieved'] = bool(sum(v for v in data['cpu']['temperature_celsius'].values()))
```
yocalebo, line 95: The name of the key doesn't match your comment. I'd rather name it `_available` to reflect reality.
```suggestion
            data['cpu_temperatures_available'] = bool(sum(v for v in data['cpu']['temperature_celsius'].values()))
```

Finally, there has to be a better way of knowing whether or not we have cpu temperatures then doing `sum` again since we're storing this information in the `['cpu']['temperature_celsius']` key up above, right?

yocalebo (changes requested):

themylogin (changes requested):
This is a trivial calculation, why bother with it on our side?

Or we can even remove `temperature` key completely if it has no temperatures (or have it set to `None`)

Current API is bloated: three values for temperature (why do we still have it in Kelvins?) where it could be one.

themylogin (approved):

PR #12178, opened 2023-09-25
NAS-111198 / 24.04 / Add some return type documentation for AD plugin

anodos325:


src/middlewared/middlewared/plugins/activedirectory.py
```diff
@@ -297,6 +297,30 @@ async def common_validate(self, new, old, verrors):
         Bool('enable'),
         update=True
     ))
+    @returns(Dict(
```
themylogin, line 300: We should probably add `ENTRY` variable like other plugins do and use it instead of copy-paste

themylogin (changes requested):

yocalebo (approved):

PR #12184, opened 2023-09-25
NAS-124326 / 24.04 / Fix API key based authentication cluster management

anodos325:
A typo in credential setup broke authentication call for the remote node connection in add_node API.


anodos325:
This fixes code-path that I added for completeness. I advised TC team to use auth tokens with adequate TTL for op completion rather than passing around actual API keys.

yocalebo (approved):

PR #12188, opened 2023-09-25
NAS-124328 / 24.04 / Fix missing ip following removal of dev_test

bmeagherix:
Also some flake8

### DIFF
--- a/tests/api2/test_470_system.py
+++ b/tests/api2/test_470_system.py
@@ -4,10 +4,12 @@
 # License: BSD
 # Location for tests into REST API of FreeNAS
 
-import sys
 import os
+import sys
+
 apifolder = os.getcwd()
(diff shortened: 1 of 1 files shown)


bmeagherix:
```
def test_06_check_system_set_time():
        """
        This test intentionally slews our clock to be off
        by 300 seconds and then verifies that it got set
        """
        results = GET("/system/info/")
    
        # Convert to seconds
        datetime = results.json()['datetime']['$date']/1000
    
        # hop 300 seconds into the past
        target = datetime - 300
>       res = make_ws_request(ip, {
            'msg': 'method',
            'method': 'system.set_time',
            'params': [int(target)]
        })
E       NameError: name 'ip' is not defined
```

anodos325 (approved):

PR #12200, opened 2023-09-26
NAS-124348 / 24.04 / Add flag showing if a display device is available for a VM

sonicaj:
This commit adds changes for UI to be able to see if a display device is available for a VM. Consumer can use following call to retrieve this information:

```
midclt call vm.query | jq .
```


yocalebo (changes requested):
I don't think adding yet another query parameter to this plugin is useful. We've already got the information from the database, so we should probably just add the `display_available` key by default.

yocalebo (approved):

PR #12202, opened 2023-09-26
NAS-123504 / 24.04 / fix ix-zfs.service

yocalebo:
Users are seeing ix-zfs.service take many 10's of minutes to complete while sometimes not even completing after letting it sit for many hours. This is probably an esoteric zfs/hardware problem for the extreme cases but we're really hurting ourselves by doing a bunch of unnecessary zpool calls. This limits the amount of zpool/zfs I/O that is performed in the `pool.import_on_boot` method since it's called in the `ix-zfs.service` systemd unit file which blocks boot-up from completing. Please note that most of the lines changed here are actually me documenting this madness since there were none and a small part of my soul was stolen from me when investigating this area of our code. P.S. I've added logging to _EVERYTHING_. Troubleshooting this is a nightmare and using `job.set_progress` is inadequate for what we need. The solution is to simply add a bunch of `self.logger` statements so we can at least look at `middlewared.log` and get a general idea of what could potentially be going wrong. There are still some significant design problems with some core functionality of our middleware (I'm looking at you `call_hook` 🤮 ) but this tries to work-around/mitigate as best as possible.

Things implemented to try and improve this method:
1. stop using `py-libzfs` in this method. We're doing a `zpool import` equivalent on _every_ zpool created on the system. This means we're telling zfs to scan all disks metadata for every zpool when `zpool import` scans all the disks by itself. Instead shift to using the good ole fashioned `subprocess` module and shell out to do `zpool import` with the necessary flags. Again, this function is only called at boot time, so I'm not too concerned about the potential drawbacks.
2. The non-standard zpool cachefile location and the surrounding logic was flawed and was never working as intended. Instead simplify to just creating our non-standard cachefile (if it doesn't already exist) and then when importing the zpool pass the `-o cachefile=` flag to update it on import. This removes an extra call to `zfs set` which is what we were doing before my changes.
3. Separate the zpool import calls into its own method while also splitting the pool unlock/umount logic into its own method so its easier to investigate/follow/maintain etc.
4. Document ALL THE THINGS. I got irrationally frustrated trying to understand our pool.unlock/umount logic and why we were doing it the way we were doing it. This brings me to my next point.
5. Simplify the unlocking/umounting logic related to when we unlock zpools whose root dataset is encrypted and/or if there are any number of child datasets that are encrypted.
6. call `disk.swaps_configure` in the background since nothing at boot is dependent on that finishing

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/import_pool.py
+++ b/src/middlewared/middlewared/plugins/pool_/import_pool.py
@@ -1,12 +1,12 @@
 import contextlib
 import errno
 import os
-import shutil
+import subprocess
 
 from middlewared.schema import accepts, Bool, Dict, List, returns, Str
 from middlewared.service import CallError, job, private, Service
 
-from .utils import ZPOOL_CACHE_FILE, ZPOOL_KILLCACHE
+from .utils import ZPOOL_CACHE_FILE
 
 
 class PoolService(Service):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/pool_/import_pool.py
```diff
@@ -204,117 +204,172 @@ async def import_pool(self, job, data):
...
-                try:
-                    self.middleware.call_sync('zfs.pool.import_pool', pool['name'], {
-                        'altroot': '/mnt',
-                        'cachefile': 'none',
-                    })
-                except Exception:
-                    self.logger.error('Failed to import %s', pool['name'], exc_info=True)
-                    continue
-
+            self.logger.debug('Creating path components for %r', ZPOOL_CACHE_FILE)
```
themylogin, line 341: Are there any statements in this `try` block that could cause an exception?

src/middlewared/middlewared/plugins/pool_/import_pool.py
```diff
@@ -204,117 +204,172 @@ async def import_pool(self, job, data):
...
-            os.unlink(ZPOOL_KILLCACHE)
-
-        if os.path.exists(ZPOOL_CACHE_FILE):
-            shutil.copy(ZPOOL_CACHE_FILE, zpool_cache_saved)
-
-        # Now finally configure swap to manage any disks which might have been removed
-        self.middleware.call_sync('disk.swaps_configure')
-        self.middleware.call_hook_sync('pool.post_import', None)
-        job.set_progress(100, 'Pools import completed')
+        # no reason to wait on these to complete before returning so call them
```
themylogin, line 373: There are reasons. For example, we generate a huge part of `etc` files in this hook. If we allow the boot process to continue, we'll be prone to a number of race conditions with services starting with invalid configuration.

src/middlewared/middlewared/plugins/pool_/import_pool.py
sonicaj: Which path components are we referring to btw here when we talk about their creation ?

src/middlewared/middlewared/plugins/pool_/import_pool.py
yocalebo: @themylogin I'm not changing anything with the order of this hook. I'm simply not blocking the entire boot-up process waiting for it to complete. The huge part of `etc` is still generated _after_ the zpools have been fully imported

themylogin: @yocalebo but we have to wait for this hook to complete before continuing with the boot process, otherwise we leave a bunch of files in /etc not generated

yocalebo: hence my comment in the body of this PR about significant design problems. This is boot up, so we need to do as little as possible so we don't unnecessarily halt boot. Trying to do _everything_ at boot is really, really hurting us. I'll move that call_hook back out of the background but this is something that we need to fix. There is no reason to halt boot-up just so we can generate all these configurations.

src/middlewared/middlewared/plugins/pool_/import_pool.py
yocalebo: > Which path components are we referring to btw here when we talk about their creation ?

That's pretty obvious if you look at the `ZPOOL_CACHE_FILE` value 😄

yocalebo: > Are there any statements in this `try` block that could cause an exception?

good catch....have no idea how the `os.makedirs` call got lost in my commit 🤷

src/middlewared/middlewared/plugins/pool_/import_pool.py
```diff
@@ -204,117 +204,166 @@ async def import_pool(self, job, data):
...
+            self.logger.debug('Importing %r with guid: %r', vol_name, vol_guid)
+            cp = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+            if cp.returncode != 0:
+                self.logger.error(
+                    'Failed to import %r with guid: %r with error: %r',
+                    vol_name, vol_guid, cp.stdout.decode()
+                )
+                return False
+        except Exception:
+            self.logger.error('Unhandled exception importing %r', vol_name, exc_info=True)
```
anodos325, line 226: I wonder if we should generate a one-shot alert if pool fails to import.

themylogin: We already have "a pool is offline" alert and it'll be re-checked at boot time a few moments later.

themylogin (approved):

PR #12203, opened 2023-09-26
NAS-124352 / 24.04 / Add iSCSI initiator group test

bmeagherix:
This was inspired by some recent discussion on the SCST mailing list.

### DIFF
--- a/tests/api2/test_261_iscsi_cmd.py
+++ b/tests/api2/test_261_iscsi_cmd.py
@@ -2401,9 +2401,10 @@ def test_26_resize_target_file(request):
 def test_27_initiator_group(request):
     depends(request, ["iscsi_cmd_00"], scope="session")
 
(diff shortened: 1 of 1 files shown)


tests/api2/test_261_iscsi_cmd.py
```diff
@@ -2377,6 +2397,58 @@ def test_26_resize_target_file(request):
                 assert MB_512 == _read_capacity16(s)
 
 
+@skip_multi_initiator
+def test_27_initiator_group(request):
+    depends(request, ["iscsi_cmd_00"], scope="session")
+
+    initiator_iqn1 = f"iqn.2018-01.org.pyscsi:{socket.gethostname()}:one"
```
yocalebo, line 2404: Minor nit but instead of calling `socket.gethostname()` 3 times, could just call it once since it (shouldn't) be changing between the invocations.

yocalebo (approved):
Cool, thanks!

PR #12207, opened 2023-09-27
Add private backend plugin for auditing framework

anodos325:
This PR adds a plugin for interacting with service-specific auditing via TrueNAS middleware ("auditbackend"). The plugin is private and will be used by to-be-developed frontend plugin with public APIs.

What we get in this PR is:
* method to query audit DBs
* periodic task to remove old entries
* updates to syslog-ng config to accomodate the new changes

Future work will add public plugin that provide:
- unified query method for multiple DBs
- configuration parameters for retention period and such
- other items that come up in design phase


anodos325:
Will begin to add tests once I get some feedback on this.

src/middlewared/middlewared/etc_files/syslog-ng/conf.d/tnaudit.conf.mako
```diff
@@ -2,11 +2,11 @@
...
     import textwrap
 
-    AUDITABLE_SERVICES = ['SMB']
-    AUDIT_DIR = '/audit'
+    from middlewared.plugins.audit.utils import AUDITED_SERVICES, audit_file_path, DATASET_PATH
 
     COLUMNS = textwrap.dedent('''
         "aid varchar",
+        "msg_ts INT",
         "vers FLOAT",
```
themylogin, line 10: Sorry, but I just noticed this. Float version, why?

src/middlewared/middlewared/etc_files/syslog-ng/conf.d/tnaudit.conf.mako
```diff
@@ -43,7 +44,7 @@
 
     def get_db(svc):
         sql = 'sql(type(sqlite3)'
-        db = f'database("{os.path.join(AUDIT_DIR, svc)}.db")'
+        db = f'database("{audit_file_path(svc)}")'
         table = 'table("audit_${TNAUDIT.svc}_${TNAUDIT.vers.major}_${TNAUDIT.vers.minor}")'
```
themylogin, line 48: Why do we distinguish versions in table names? Do we plan to copy audit logs to new BE on upgrade?

src/middlewared/middlewared/plugins/audit/backend.py
```diff
@@ -0,0 +1,179 @@
...
+        if not days or days < 0:
+            raise ValueError("Days must be positive value greater than zero.")
+
+        secs = days * 86400
+        cutoff_ts = int(time.time()) - secs
+        with self.lock:
+            with Session(self.engine) as s:
+                expired = s.query(self.table).filter(self.table.c.msg_ts < cutoff_ts)
+                expired.delete(synchronize_session=False)
+                s.commit()
```
themylogin, line 107: Do we need VACUUM here?

src/middlewared/middlewared/etc_files/syslog-ng/conf.d/tnaudit.conf.mako
anodos325: At present there are no plans to copy them forward between versions. The idea being that local audit records are somewhat ephemeral and if someone is really serious about auditing they should configure to send them off remotely (TODO). But plans can always change. By putting them in separate table we avoid having syslog-ng try to automatically create new columns in the DB if we decide to change schema (indicated by version bump).

src/middlewared/middlewared/etc_files/syslog-ng/conf.d/tnaudit.conf.mako
anodos325: Because I probably needed to look at it a different way. We already store version in the table name. We can save a column and just delete.

src/middlewared/middlewared/plugins/audit/backend.py
anodos325: That's a good idea

themylogin (approved):

PR #12209, opened 2023-09-27
No longer shortcut alua_enabled when license contains FIBRECHANNEL

bmeagherix:
This is not valid for SCALE as currently fibrechannel is not supported.


yocalebo (approved):

yocalebo:
backport

PR #12213, opened 2023-09-28
NAS-124388 / 24.04 / Refactor and simplify test_420_smb

anodos325:
This PR simplifies many test cases in the test_420_smb file, removes various sleep() calls, and replaces calls to midclt via SSH with websocket client calls.

### DIFF
--- a/tests/api2/test_420_smb.py
+++ b/tests/api2/test_420_smb.py
@@ -12,13 +12,14 @@ import uuid
 from time import sleep
 apifolder = os.getcwd()
 sys.path.append(apifolder)
-from functions import PUT, POST, GET, DELETE, SSH_TEST
+from functions import PUT, SSH_TEST
 from protocols import smb_connection
 from utils import create_dataset
 from auto_config import ip, pool_name, password, user, hostname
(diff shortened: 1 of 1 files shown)


yocalebo (changes requested):
Let's write a module level fixture that instantiates a single websocket session for entirety of these tests.

```
@pytest.fixture(scope='module')
def ws_client():
    # by the time this test is called in the pipeline,
    # the HA VM should have networking configured so
    # we can use the VIP
    with client(host_ip=os.environ['virtual_ip'] if ha else ip) as c:
        yield c

File: tests/api2/test_420_smb.py [hunk 1]
```diff
@@ -12,13 +12,14 @@
 from time import sleep
 apifolder = os.getcwd()
 sys.path.append(apifolder)
-from functions import PUT, POST, GET, DELETE, SSH_TEST
+from functions import PUT, SSH_TEST
```
yocalebo, line 15: There are only 4 more instances of using the `PUT` REST interface, why not remove them all since you're doing it mostly everywhere with these changes?

yocalebo (approved):

PR #12221, opened 2023-09-29
When possible reuse websocket sessions in tests

anodos325:
This is a minor optimization to open and use a single websocket client connection for the duration of the test. API consumers (for instance for HA tests) can import the PersistentCtx and run setup method with relevant arguments.

### DIFF
--- a/src/middlewared/middlewared/test/integration/utils/client.py
+++ b/src/middlewared/middlewared/test/integration/utils/client.py
@@ -13,21 +13,32 @@ __all__ = ["client", "host", "host_websocket_uri", "password", "session", "url",
 logger = logging.getLogger(__name__)
 
 class PersistentClient(Client):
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/test/integration/utils/client.py
```diff
@@ -1,17 +1,92 @@
...
+        return self.setup(*args, **kwargs)
+
+
+PersistentCtx = ClientCtx()
 
 
 @contextlib.contextmanager
-def client(*, auth=undefined, auth_required=True, py_exceptions=True, log_py_exceptions=True, host_ip=None):
+def client(*, auth=undefined, auth_required=True, py_exceptions=True, log_py_exceptions=True, host_ip=None, reuse_conn=True):
+    if reuse_conn and auth is undefined and host_ip is None and auth_required:
```
themylogin, line 86: What about other parameters (e.g. related to exceptions)?

PR #12229, opened 2023-10-02
NAS-124400 / 24.04 / Remove pre-SQLite 3.30 workaround

themylogin:
It was introduced by e1054f7b2e04ce75225742a1b3c5f02644829c6a along with some tests that now pass even with this bit removed


themylogin:
time 0:15

anodos325 (approved):

PR #12232, opened 2023-10-02
NAS-124425 / 24.04 / Fatal Error message appears post start of iSCSI service containing multiple targets and file extents

bmeagherix:
Starting in COBIA we require a valid serial number to be specified for each iSCSI target extent, in order to support ALUA.  This gets applied to the scst config as `t10_dev_id`.

Previously _some_ paths through our code auto-generated a serial number, but another path did not.  This PR rectifies this deficiency by enhancing `extent_serial` to also generate on the empty string.

Furthermore, we also now ensure that serial numbers are unique.

Finally, add some tests.

### DIFF
--- a/tests/api2/test_261_iscsi_cmd.py
+++ b/tests/api2/test_261_iscsi_cmd.py
@@ -2494,7 +2494,7 @@ def test_28_portal_access(request):
                             assert MB_100 == _read_capacity16(s)
 
 
-def test_29_multiple_extents(request):
+def test_29_multiple_extents():
     """
     Verify that an iSCSI client can access multiple target LUNs
     when multiple extents are configured.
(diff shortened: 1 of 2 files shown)


tests/api2/test_261_iscsi_cmd.py
```diff
@@ -2491,6 +2494,48 @@ def test_28_portal_access(request):
...
+                                    TUR(s)
+                                    s.blocksize = 512
+                                    assert MB_256 == _read_capacity16(s)
+
+                                # Now try to create another extent using the same serial number
+                                # We expect this to fail.
+                                with pytest.raises(AssertionError) as ve:
+                                    with file_extent(pool_name, dataset_name, "target.extent3", filesize=MB_512,
+                                                     extent_name="extent3", serial=extent1_config['serial']):
+                                        assert False, "Should not have been able to duplicate extent serial."
```
themylogin, line 2530: You don't need to do this, `pytest.raise` will fail the test if the exception was not raised by the end of the block

themylogin (changes requested):

tests/api2/test_261_iscsi_cmd.py
bmeagherix: Agreed.  Thanks.

themylogin (approved):
Thank you!

PR #12233, opened 2023-10-02
NAS-124168 / 24.04 / Properly raise an exception if snapshot has dependents and defer is not set

sonicaj:


themylogin (commented):
Now we are listing snapshot dependents on each delete operation which can negatively impact the performance.

IMO it will be more efficient to only check for dependents if that specific error was raised by `snap.delete`

yocalebo (approved):

PR #12240, opened 2023-10-03
NAS-123930 / 24.04 / Add audit plugin

anodos325:
Add a public audit plugin for API consumers. This provides `config` and
`update` methods that allow configuration of auditing details as well
as a `query` endpoint for retrieving information from the auditing
databases.

No API endpoints are provided for audit database entry deletion or
modification (apart from altering the auditing retention).

The `audit.config` endpoint also provides information about space
utilization for the auditing dataset to simplify implementation
for API consumers, and whether the we are sending the audit
entries to a remote syslog server.


src/middlewared/middlewared/plugins/audit/audit.py
```diff
@@ -0,0 +1,347 @@
...
+from middlewared.service import private, ConfigService
+from middlewared.service_exception import ValidationErrors
+from middlewared.utils import filter_list
+from middlewared.utils.functools import cache
+from middlewared.utils.osc import getmntinfo
+from middlewared.validators import Range
+
+
+ALL_AUDITED = [svc[0] for svc in AUDITED_SERVICES]
+QUOTA_WARN = 'org.freenas:refquota_warning'
```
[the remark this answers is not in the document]
anodos325: I'm plumbing into the existing quota / refquota alert framework here, which means I can't really change it. I did notice that we're using magic strings all over the place for this stuff and so I'm putting all of the custom ZFS user properties that we create into a single enum class.

yocalebo (approved):
Cool!

src/middlewared/middlewared/plugins/audit/utils.py
```diff
@@ -6,7 +6,12 @@
 AUDIT_DATASET_PATH = '/audit'
 AUDITED_SERVICES = [('SMB', 0.1)]
 AUDIT_TABLE_PREFIX = 'audit_'
-AUDIT_LIFETIME = 7  # Temporary hard-coded retention until public APIs developed
+AUDIT_LIFETIME = 7
+AUDIT_DEFAULT_RESERVATION = 0
+AUDIT_DEFAULT_QUOTA = 0
+AUDIT_FILL_CRITICAL = 95
+AUDIT_FILL_WARNING = 80
```
bmeagherix, line 13: Should these now be renamed to `AUDIT_DEFAULT_LIFETIME`, `AUDIT_DEFAULT_FILL_CRITICAL` & `AUDIT_DEFAULT_FILL_WARNING`

src/middlewared/middlewared/plugins/audit/utils.py
```diff
@@ -6,7 +6,12 @@
...
+AUDIT_MIN_FREE = 1  # base minimum free size to keep availalable at all times
```
bmeagherix, line 14: Is `AUDIT_MIN_FREE` used anywhere?

If so, maybe fix the spelling "availalable".  If not, maybe remove.

src/middlewared/middlewared/schema/string_schema.py
```diff
@@ -340,6 +340,20 @@ def validate(self, value):
...
 
+class UUID(Str):
+
+    def validate(self, value):
+        if value is None:
+            return
+
+        try:
+            uuid.UUID(value)
+        except TypeError:
```
bmeagherix, line 351: Is `uuid` imported?

tests/api2/test_audit_basic.py
```diff
@@ -0,0 +1,109 @@
...
+            c.read(fd, 0, 3)
+        c.close(fd, True)
+
+    sleep(10)
+    ops = call('audit.query', {
+        'services': ['SMB'],
+        'query-filters': [['user', '=', SMBUSER]],
+        'query-options': {'count': True}
+    })
+    assert ops > 0
```
bmeagherix, line 109: Would it be a good idea to also create _another_ share which is **NOT** being audited, and then write to both shares, and verify that you only get events from the expected share?

PR #12252, opened 2023-10-05
NAS-124486 / 22.12.5 / Update "toggled" share acl on sharesec update

anodos325:
There are various ways that users can encrypt, decrypt, enable and disable SMB paths and shares. Some operations require removing the SMB share from our shares list, which automatically removes the share acl entry. To prevent this we store a copy of it prefixed by a `#` character when share path is encrypted and when share is decrypted we restore from the backup copy. This logic had some gaps in it whereby a reboot when the backup was out-of-sync from the current one, the current share ACL could be replaced. This commit plugs the gap by immediately duplicating the ACL on setacl calls.

This issue does not impact Cobia and later since SMB share ACL handling was substantially refactored.


yocalebo:
process

yocalebo (approved):

PR #12256, opened 2023-10-06
fix: https://ixsystems.atlassian.net/browse/NAS-124487

exander77:


yocalebo:
As I said in the ticket, we do not expect disks to be formatted by hand. We expect gptid partitions on the disks. You've created your pool manually by using /by-id/ which is not what we expect. It will cause u defined behavior. In the ticket, you actually state that with these changes, the disks still show up as disks that can be used in a new zpool when they're already a part of one. Hence why I said "undefined behavior"

PR #12258, opened 2023-10-06
fix: https://ixsystems.atlassian.net/browse/NAS-124487

exander77:
I am making another pull request with this after the previous one was cancelled.

Whatever you think about the issue, it is not acceptable to read file without checking if it is present and then crashing storage UI, k3s and docker in the process. How it happens is of no consequence.

This is not feature request, this pull request is fixing unsanitized code. You may 100% ignore how I discovered it, but the code is unsanitized in a major way.

I also do not think that importing zpool should crash storage UI, k3s and docker under any circumstances.

I am willing to improve it any way, so it passes and even track why disks are reported as unused, it will be similar issue.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/disk_info.py
+++ b/src/middlewared/middlewared/plugins/disk_/disk_info.py
@@ -125,11 +125,15 @@ class DiskService(Service):
     def get_disk_from_partition(self, part_name):
         if not os.path.exists(os.path.join('/dev', part_name)):
             return None
-        with open(os.path.join('/sys/class/block', part_name, 'partition'), 'r') as f:
-            part_num = f.read().strip()
-        if part_name.startswith(('nvme', 'pmem')):
-            # nvme/pmem partitions would be like nvmen1p1 where disk is nvmen1
-            part_num = f'p{part_num}'
+        try:
+            with open(os.path.join('/sys/class/block', part_name, 'partition'), 'r') as f:
+                part_num = f.read().strip()
+        except FileNotFoundError:
+            return part_name
+        else:
+            if part_name.startswith(('nvme', 'pmem')):
+                # nvme/pmem partitions would be like nvmen1p1 where disk is nvmen1
+                part_num = f'p{part_num}'
         return part_name.rsplit(part_num, 1)[0].strip()
 
     @private


yocalebo (changes requested):
You've introduced a race condition by checking for a files existence before reading it.

Furthermore, you do not give any indication what this is exactly fixing. I've also given you a reason why I closed the original PR. You're trying to fix our code because you've manually partitioned your disks and created your zpool without using our API.

exander77:
If such race condition happens the behavior will fall back to its original, so I think this is not valid argument.

I do not think it is the issue, how the pool was created, but nowhere in docs, I see I cannot import pool made elsewhere. But as I say, it is not the issue.

Issue is that file is read before its existence is checked:
```
        with open(os.path.join('/sys/class/block', part_name, 'partition'), 'r') as f:
```

When disk is not portioned it should just return `part_name`.
```
        disk = os.path.join('/sys/class/block', part_name, 'partition')
        if not os.path.exists(disk):
            return part_name
        with open(disk, 'r') as f:
```

Btw, it works fine if the pool is imported by `/dev/name` (but this is unstable as device names are not fixed) instead of `/dev/disk/by-id/name`. So this is purely a sanitization issue.

exander77:
Btw, If you want to remove race condition, I can just catch the exception there, but I do not think partitions just pop up between checking and accessing the partitions file, but even if they were, it would not be an issue.

yocalebo:
You still have introduced a race condition by checking for the files existence.
```
        if not os.path.exists(disk):
             return part_name
         with open(disk, 'r') as f:
```
`disk` can be created between the existence of when that file is checked and the time it's opened for a read.

exander77:
It is the same as if the disk were created after it tried to read partitions file. Where is the difference?

But I can easily alter it to:
```
        try:
            with open(os.path.join('/sys/class/block', part_name, 'partition'), 'r') as f:
                part_num = f.read().strip()
        except:
            return part_name
```

yocalebo:
Can you do this instead please
```
try:
    with open(os.path.join('/sys/class/block', part_name, 'partition') as f:
        part_num = f.read().strip()
except FileNotFoundError:
    return part_name
else:
    ...

**Comment by exander77** [comment]
```
    @private
    def label_to_dev(self, label, *args):
        dev = os.path.realpath(os.path.join('/dev', label)).split('/')[-1]
        if dev == label and os.path.exists(os.path.join('/sys/block', label)):
            # This is to cater for a case where `label` is a complete disk
            # instead of something like disk/by-partuuid/some-uuid-here
            return dev
        else:
            return dev if dev != label.split('/')[-1] else None

    @private
    def label_to_disk(self, label, *args):
        part_disk = self.label_to_dev(label)
        if part_disk == label:
            return label
        else:
            return self.get_disk_from_partition(part_disk) if part_disk else None
```

Also, you have same case here, existence of the directory is checked the same way, before the file is read.  So, you can rewrite it all if that's your issue, `label_to_dev` checks if directory is there and then it calls `get_disk_from_partition` to read the file.

But as long as it passes, I can rewrite by any way. I actually checked the code around to see how you are doing it, so that's why there is `os.path.exists`.

**Comment by yocalebo** [comment]
I'm glad you're looking through the code, we have bugs everywhere (just like any software project) so PRs fixing these are always welcome. For this PR, however, can you do implement the changes like so
```
try:
    with open(os.path.join('/sys/class/block', part_name, 'partition') as f:
        part_num = f.read().strip()
except FileNotFoundError:
    return part_name
else:
    ...

yocalebo (approved):
Thank you!

exander77:
See: https://github.com/truenas/middleware/pull/12258/files

exander77:
I thank you, it is super hard to get pull request pass you. I will try to track why the disk are put in unused, it would some name mismatch somewhere. But I have to find out where are those unused disk even comming from. It is much easier to track exception in the code than weird behavior.

yocalebo:
> I thank you, it is super hard to get pull request pass you. I will try to track why the disk are put in unused, it would some name mismatch somewhere. But I have to find out where are those unused disk even comming from. It is much easier to track exception in the code than weird behavior.

No problem. Unfortunately for you, the disk management logic is some of the most fragile, complex code we have so I wish you good luck in tracking it down. If you manage to come across the problem and you open another PR, it will be reviewed with the same type of scrutiny because this is such a critical portion of our code :semi

PR #12260, opened 2023-10-06
fix: https://ixsystems.atlassian.net/browse/NAS-124487 (#12258)

yocalebo:
(cherry picked from commit 700d4401d26defb8e3c8c20c39332a915582769f)


exander77:
Btw, if anybody can point be out where unused disks are comming from, I would be glad. Is it `plugins/disk_/availability.py`?

PR #12271, opened 2023-10-06
NAS-124503 / 24.04 / Add more error handling for reinit of SMB passdb

anodos325:
If we hit some error while attempting to rename the passdb.tdb file, we should still go ahead with attempting to sync it up with our config database.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb_/passdb.py
+++ b/src/middlewared/middlewared/plugins/smb_/passdb.py
@@ -165,9 +165,15 @@ class SMBService(Service):
         private_dir = SMBPath.PASSDB_DIR.platform()
         ts = int(time.time())
         old_path = f'{private_dir}/passdb.tdb'
-        new_path = f'{private_dir}/passdb.{ts}.corrupted'
-        os.rename(old_path, new_path)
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/smb_/passdb.py
```diff
@@ -165,9 +165,15 @@ async def passdb_reinit(self, conf_users):
         private_dir = SMBPath.PASSDB_DIR.platform()
         ts = int(time.time())
         old_path = f'{private_dir}/passdb.tdb'
-        new_path = f'{private_dir}/passdb.{ts}.corrupted'
-        os.rename(old_path, new_path)
-        self.logger.debug("Backing up original passdb to [%s]", new_path)
+        new_path = f'{private_dir}/passdb.{ts}.old'
+        try:
+            os.rename(old_path, new_path)
+            self.logger.debug("Backing up original passdb to [%s]", new_path)
```
yocalebo, line 171: Let's move this before the os.rename

yocalebo (changes requested):

yocalebo (approved):

PR #12282, opened 2023-10-09
NAS-124565 / 24.04 / Adjust parameter names per feedback from ui team

anodos325:
Have more descriptive column names for audit databases.


tests/api2/test_420_smb.py
```diff
@@ -312,8 +312,8 @@ def check_groupmap_for_entry(groupmaps, nt_name):
 
 
 AUDIT_FIELDS = [
-    'aid', 'vers', 'time', 'addr', 'user', 'sess', 'svc',
-    'svc_data', 'event', 'event_data', 'success'
```
bmeagherix, line 316: I'll approve, but I do note that we used to have `vers` here (and _only_ here?).  Presumably it was originally planned, but got dropped?

bmeagherix (approved):

tests/api2/test_420_smb.py
anodos325: Yeah, it got removed as part of review during second part of merges.

PR #12286, opened 2023-10-09
NAS-127148 / 24.10 / Include information about allocation size

anodos325:
Add allocation_size key to filesystem.stat and filesystem.listdir output. This commit also expands documentation for filesystem.stat output.


src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -222,6 +222,7 @@ def statx_entry_impl(self, entry, options=None):
         Path('realpath', required=True),
         Str('type', required=True, enum=['DIRECTORY', 'FILE', 'SYMLINK', 'OTHER']),
         Int('size', required=True, null=True),
+        Int('allocation_size', required=True, null=True),
```
sonicaj, line 225: Will it be null though seeing the below diff ?

sonicaj (approved):

yocalebo (approved):

PR #12303, opened 2023-10-10
NAS-122626 / 24.04 / Allow specifying interval/otp digits at per user level for 2FA

sonicaj:
This PR adds changes to have `interval` and `otp_digits` be specified at user level for both AD and local users. This helps making it more configurable and a user can change his own settings instead of this being applied at a global level which resulted in all the existing users secret to be renewed because secret depends on this.


anodos325:
Is it possible to add test for this?

anodos325 (approved):

PR #12304, opened 2023-10-10
NAS-124588 / 24.04 / Fix LDAP bindpw handling

mgrimesix:
Remove code that was preventing the clearing of the LDAP 'bindpw' field.
If keytab is available, we prefer that 'bindpw' not be populated.

Changed the nag message to a user action request and reduced it's priority from warning to info.

### DIFF
--- a/src/middlewared/middlewared/plugins/ldap.py
+++ b/src/middlewared/middlewared/plugins/ldap.py
@@ -375,9 +375,6 @@ class LDAPService(TDBWrapConfigService):
         for key in ["ssl", "schema"]:
             data[key] = data[key].lower()
 
-        if not data['bindpw']:
-            data.pop('bindpw')
-
         data.pop('uri_list')
         data.pop('cert_name')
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -457,8 +454,8 @@ async def common_validate(self, new, old, verrors):
                 "Bind credentials or kerberos keytab are required for an authenticated bind."
             )
         if new["bindpw"] and new["kerberos_principal"]:
-            self.logger.warning("Simultaneous keytab and password authentication "
-                                "are selected. Clearing LDAP bind password.")
+            self.logger.info("Simultaneous keytab and password authentication "
+                             "are selected. Please clear LDAP bind password.")
```
anodos325, line 458: If this is called when validating ldap.update() then we in theory will now update with the bindpw removed. I think it's probably fine to just remove logging here.

mgrimesix: Agreed

anodos325 (approved):

PR #12307, opened 2023-10-11
NAS-124605 / 24.04 / Add API for generating audit reports

anodos325:
This provides an API for generating an audit report based on our query-filters and query-options for services (same API details as audit.query), but adds an additional parameter to specify format. When the endpoint is called a file in the specified format is written to a file in a directory (based on currently-authenticated user's username) within /audit/reports. The path for the newly-created file is return to caller, and may be used to fetch file either through scp or middleware filesystem.get API.

### DIFF
--- a/src/middlewared/middlewared/plugins/audit/audit.py
+++ b/src/middlewared/middlewared/plugins/audit/audit.py
@@ -1,5 +1,10 @@
+import csv
+import errno
 import middlewared.sqlalchemy as sa
 import os
+import time
+import uuid
+import yaml
 
 from .utils import (
     AUDIT_DATASET_PATH,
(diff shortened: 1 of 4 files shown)


bmeagherix (approved):

src/middlewared/middlewared/plugins/audit/audit.py
```diff
@@ -190,6 +199,103 @@ async def query(self, data):
 
         return filter_list(results, data['query-filters'], data['query-options'])
 
+    @accepts(Patch(
+        'audit_query', 'audit_export',
+        ('add', Str('export_format', enum=['CSV', 'JSON', 'YAML'], default='JSON')),
+    ))
+    @returns(Str('audit_file_path'))
+    @pass_app()
+    def export(self, app, data):
```
yocalebo, line 208: So we have a limit option that can be passed? Do we expect to honor that here?

Lastly, are we concerned about the amount of data that can be queried and then written to disk? Would this potentially block for a large amount of time based on the number of entries that are being exported? Just food for thought

yocalebo (approved):
really cool!

PR #12312, opened 2023-10-12
NAS-124429 / 24.04 / Add flag to reporting realtime event source determining netdata's health

Qubad786:
## Problem

The current user interface lacks the ability to verify whether Netdata is operational and correctly receiving data from the `reporting.realtime` event source.

## Solution

To address this limitation, a flag is added to the response from `reporting.realtime` event source. This flag will be set to true if Netdata is not running. Furthermore, changes are made to the Netdata unit file's restart policy, setting it to "always." This ensures that Netdata will automatically restart in case it fails to start for any reason. These improvements enhance the UI's capability to verify and maintain Netdata's functionality.

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/events.py
+++ b/src/middlewared/middlewared/plugins/reporting/events.py
@@ -76,19 +76,23 @@ class RealtimeEventSource(EventSource):
                 else:
                     break
 
-            data = {
-                'zfs': get_arc_stats(netdata_metrics),  # ZFS ARC Size
-                'memory': get_memory_info(netdata_metrics),
-                'virtual_memory': psutil.virtual_memory()._asdict(),
-                'cpu': get_cpu_stats(netdata_metrics, cores),
-                'disks': get_disk_stats(netdata_metrics, self.middleware.call_sync('device.get_disk_names')),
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/reporting/events.py
```diff
@@ -76,19 +76,23 @@ def run_sync(self):
...
+                data = {
+                    'zfs': get_arc_stats(netdata_metrics),  # ZFS ARC Size
+                    'memory': get_memory_info(netdata_metrics),
+                    'virtual_memory': psutil.virtual_memory()._asdict(),
+                    'cpu': get_cpu_stats(netdata_metrics, cores),
+                    'disks': get_disk_stats(netdata_metrics, self.middleware.call_sync('device.get_disk_names')),
+                    'interfaces': get_interface_stats(
+                        netdata_metrics, [i['name'] for i in self.middleware.call_sync('interface.query')]
+                    ),
+                    'failed_to_connect': failed_to_connect,
```
yocalebo, line 91: Just explicitly set this to `False`

yocalebo (changes requested):

yocalebo (approved):

PR #12314, opened 2023-10-12
NAS-124585 / 24.04 / Ensure we always generate debug

Qubad786:
## Problem
In certain scenarios (which i have not been able to reproduce), the `shutils.rmtree` operation fails to remove an existing debug directory. This failure results in a "Compressed path already exists" call error.

## Solution
To mitigate this error and ensure that we always generate a debug no matter what, a new directory will be created if an unexpected issue occurs during the removal of the debug directory. This adjustment ensures the smooth execution of the operation and prevents the error from being triggered.

### DIFF
--- a/src/middlewared/middlewared/plugins/system/debug.py
+++ b/src/middlewared/middlewared/plugins/system/debug.py
@@ -1,3 +1,4 @@
+import contextlib
 import io
 import os
 import requests
@@ -25,11 +26,24 @@ class SystemService(Service):
         Result value will be the absolute path of the file.
         """
         system_dataset_path = self.middleware.call_sync('systemdataset.config')['path']
-        execution_dir = get_debug_execution_dir(system_dataset_path)
-        dump = os.path.join(execution_dir, 'ixdiagnose.tgz')
 
-        # Be extra safe in case we have left over from previous run
-        shutil.rmtree(execution_dir, ignore_errors=True)
+        i = 0
+        while True:
+            execution_dir = get_debug_execution_dir(system_dataset_path, i)
+            dump = os.path.join(execution_dir, 'ixdiagnose.tgz')
+
+            # Be extra safe in case we have left over from previous run
+            try:
+                with contextlib.suppress(FileNotFoundError):
+                    shutil.rmtree(execution_dir, ignore_errors=False)
+            except Exception as e:
+                i += 1
+                if i >= 5:
+                    raise CallError(f'Failed to generate ixdiagnose debug: {e!r}')
+                else:
+                    self.logger.warning('Failed to generate ixdiagnose debug: %r', str(e))
+            else:
+                break
 
         conf.apply({
             'compress': True,
--- a/src/middlewared/middlewared/plugins/system/utils.py
+++ b/src/middlewared/middlewared/plugins/system/utils.py
@@ -1,6 +1,9 @@
 import enum
 import os
 import re
+import typing
+
+from middlewared.utils import MIDDLEWARE_RUN_DIR
 
 
 DEBUG_MAX_SIZE = 30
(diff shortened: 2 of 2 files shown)


sonicaj (approved):

yocalebo (changes requested):
self.logger.warning??

yocalebo (changes requested):
This could retry an indefinite amount of times so we need to cut it off after 5 times.

We also need to change the log message so that it lets end-user we're retrying without exc_info=True. On the last retry attempt if we fail, then we can log a separate message with exc_info=True

anodos325:
Since debugs can have sensitive information, I don't think it's a great idea to have a code path that can result in writing it to /tmp.

src/middlewared/middlewared/plugins/system/debug.py
```diff
@@ -25,11 +26,26 @@ def debug_generate(self, job):
         Result value will be the absolute path of the file.
         """
         system_dataset_path = self.middleware.call_sync('systemdataset.config')['path']
-        execution_dir = get_debug_execution_dir(system_dataset_path)
-        dump = os.path.join(execution_dir, 'ixdiagnose.tgz')
 
-        # Be extra safe in case we have left over from previous run
-        shutil.rmtree(execution_dir, ignore_errors=True)
+        i = 0
```
yocalebo, line 30: Sorry I wasn't clear enough with my review. This is what I meant.
```
max_retries = 5
retry_attempts = 0
while retry_attempts <= max_retries:
    try:
        ...
    except Exception as e:
        retry_attempts += 1
        if retry_attempts >= max_retries:
            raise CallError(f'Failed to generate debug information {str(e)}')
        else:
            self.logger.warning('Failed to generate debug information with error %s, retrying.', e)

File: src/middlewared/middlewared/plugins/system/utils.py [hunk 2]
```diff
@@ -23,8 +24,10 @@ def __init__(self):
         self.SYSTEM_SHUTTING_DOWN = False
 
 
-def get_debug_execution_dir(system_dataset_path: str) -> str:
-    return '/var/tmp/ixdiagnose' if system_dataset_path is None else os.path.join(system_dataset_path, 'ixdiagnose')
+def get_debug_execution_dir(system_dataset_path: str, iteration: typing.Optional[int] = 0) -> str:
+    return f'/var/tmp/ixdiagnose-{iteration}' if system_dataset_path is None else os.path.join(
```
yocalebo, line 28: I do agree with @anodos325 's comment about putting this in `/tmp`. We're trying to not perform operations within that filesystem for obvious reasons. We have a middleware run directory that is somewhat more secure than `/tmp` that we can put these in.

yocalebo (changes requested):

yocalebo (commented):
This looks good but 1 last issue I see is that we're using /var/run/middleware but I believe we have a global variable that defines this path. Can we use that instead since it's possible this dir could change

yocalebo (approved):

src/middlewared/middlewared/plugins/system/debug.py
```diff
@@ -25,11 +26,24 @@ def debug_generate(self, job):
...
+        while True:
+            execution_dir = get_debug_execution_dir(system_dataset_path, i)
```
anodos325, line 32: It doesn't look like you're specifying any permissions in `os.makedirs` call in `generate_debug() in ixdiagnose. You'll want some mechanism to set permissions so as to ensure that final path component here prevents unprivileged user from accessing the debug info.

Qubad786: @anodos325 made those changes in https://github.com/truenas/ixdiagnose/pull/99

anodos325 (approved):

PR #12316, opened 2023-10-12
NAS-124598 / 24.04 / Optimize retrieving shares/tasks information

sonicaj:
This PR adds various changes to optimize retrieval of shares/tasks after a discussion with @anodos325.

(TODO: Fill this in)

### DIFF
--- a/src/middlewared/middlewared/service/sharing_service.py
+++ b/src/middlewared/middlewared/service/sharing_service.py
@@ -27,11 +27,12 @@ class SharingTaskService(CRUDService):
         return {
             'locked_datasets': await self.middleware.call(
                 f'zfs.dataset.locked_datasets{"_cached" if extra.get("use_cached_locked_datasets", True) else ""}'
-            ),
+            ) if extra.get('retrieve_locked_info', True) else [],
             'service_extend': (
                 await self.middleware.call(self._config.datastore_extend_context, rows, extra)
                 if self._config.datastore_extend_context else None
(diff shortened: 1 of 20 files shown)


src/middlewared/middlewared/service/sharing_service.py
```diff
@@ -24,17 +24,19 @@ async def get_path_field(self, data):
...
-
+        if extra.get('use_cached_locked_datasets', True):
+            locked_ds_endpoint = 'pool.dataset.locked_datasets_cached'
+        else:
+            locked_ds_endpoint = 'zfs.dataset.locked_datasets'
         return {
-            'locked_datasets': await self.middleware.call('zfs.dataset.locked_datasets', datasets) if datasets else [],
+            'locked_datasets': await self.middleware.call(
+                locked_ds_endpoint
+            ) if extra.get('retrieve_locked_info', True) else [],
```
anodos325, line 34: I think instead of using `extra.retrieve_locked_info` it would be easier and less error-prone to omit locked info lookup in cases where `select` is specified and doesn't include `locked`. We'd have to update existing callers, but this would hopefully make future code using fast-path easier to write.

anodos325:
Okay. This looks good. We'll probably need to carefully review middlewared logs in CI run to make sure it doesn't introduce regressions.

yocalebo (approved):

anodos325 (approved):

PR #12332, opened 2023-10-13
Add share_type preset for mixed SMB/NFS

anodos325:
In principle in Linux this is less of a problem than in FreeBSD because of proper inotify support, and kernel oplocks. Add a share_type preset that provides user with dataset settings that are likely to make SMB and NFS clients happier.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/dataset.py
+++ b/src/middlewared/middlewared/plugins/pool_/dataset.py
@@ -432,7 +432,7 @@ class PoolDatasetService(CRUDService):
         Inheritable(Str('casesensitivity', enum=['SENSITIVE', 'INSENSITIVE']), has_default=False),
         Inheritable(Str('aclmode', enum=['PASSTHROUGH', 'RESTRICTED', 'DISCARD']), has_default=False),
         Inheritable(Str('acltype', enum=['OFF', 'NFSV4', 'POSIX']), has_default=False),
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/pool_/dataset.py
```diff
@@ -563,6 +563,11 @@ async def do_create(self, app, data):
                 'flags': {'BASIC': 'INHERIT'},
                 'type': 'ALLOW'
             })
+        elif data['share_type'] == 'MULTIPROTOCOL':
+            data['casesensitivity'] = 'SENSITIVE'
```
sonicaj, line 567: Just curious on what happens when user explicitly specifies these attrs to something else ? Do we have validation in place to prevent them from doing that or are these defaults which we want to apply and if user overrides, he/she can happily do it ?

anodos325: It currently overrides inherited or manually specified parameters. This is consistent with other `share_type` entries that we have. We could redesign this field for DragonFish or Electic Eel if needed.

yocalebo (approved):

src/middlewared/middlewared/plugins/pool_/dataset.py
sonicaj: @anodos325 we should have validation in place to not allow users to specify these attrs then as this creates a false impression to the user where he assumes he has set something which is not being actually set

PR #12333, opened 2023-10-14
Fix filtering behavior for user and group queries

anodos325:
This commit removes a legacy behavior for the private method dscache.query. In this case dscache.query calls would also retrieve the local users datastore and return unified results. This was due to some design constraints when replacing the user and group cache in our legacy django webui in TrueNAS 11.2. As of SCALE the webui no longer directly calls dscache.query to get user / group lists and so we can safely remove this oddity.

The primary motivation for removing this legacy behavior is to fix how query-options are applied to query results.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -245,9 +245,7 @@ class UserService(CRUDService):
 
         `"extra": {"search_dscache": true}` is a legacy method of querying for directory services users.
         """
-        if not filters:
-            filters = []
-
+        ds_users = []
         options = options or {}
         options['extend'] = self._config.datastore_extend
         options['extend_context'] = self._config.datastore_extend_context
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/account.py
```diff
@@ -277,15 +276,13 @@ async def query(self, filters, options):
         if dssearch:
             ds_state = await self.middleware.call('directoryservices.get_state')
             if ds_state['activedirectory'] == 'HEALTHY' or ds_state['ldap'] == 'HEALTHY':
-                dssearch_results = await self.middleware.call('dscache.query', 'USERS', filters, options.copy())
+                ds_users = await self.middleware.call('dscache.query', 'USERS', filters, options.copy())
```
sonicaj, line 279: Just to be sure, the users we get here will have all attrs our regular local db users have (normalized and everything) ?

anodos325: Yes.

sonicaj (approved):

PR #12364, opened 2023-10-19
NAS-124719 / 24.04 / Fix cases where Str is used instead of Password

anodos325:
The Password class ensures that the string being handled is treated as private and redacted from relevant logs. This commit fixes many cases where the incorrect class was used for hanlding passwords. In some of the instances, the Str had the private parameter set and so there is no functional change, but it's beneficial to clearly mark places where passwords are being handled as such in the codebase.

For the purpose of this PR, API keys and tokens are also treated as password-equivalent strings.

### DIFF
--- a/src/middlewared/middlewared/alert/service/aws_sns.py
+++ b/src/middlewared/middlewared/alert/service/aws_sns.py
@@ -1,7 +1,7 @@
 import boto3
 
 from middlewared.alert.base import ThreadedAlertService
-from middlewared.schema import Dict, Str
+from middlewared.schema import Dict, Password, Str
 
 
 class AWSSNSAlertService(ThreadedAlertService):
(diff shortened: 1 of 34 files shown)


bmeagherix (changes requested):
All looks good so far.

Would suggest adding `secret` and `peersecret` from [src/middlewared/middlewared/plugins/iscsi_/auth.py](https://github.com/truenas/middleware/blob/master/src/middlewared/middlewared/plugins/iscsi_/auth.py)

I also **wonder** about:

-  `secret` in [add_jwt_secret](https://github.com/truenas/middleware/blob/master/src/middlewared/middlewared/plugins/gluster_linux/local_events.py) and [GlusterEventsdService.create](https://github.com/truenas/middleware/blob/master/src/middlewared/middlewared/plugins/gluster_linux/eventsd.py) (and `bearer_token`)
- a few `client_secret` wrt rclone/remote/*.py
- [application_secret](https://github.com/truenas/middleware/blob/master/src/middlewared/middlewared/plugins/acme_protocol_/authenticators/ovh.py) and [secret_access_key](https://github.com/truenas/middleware/blob/master/src/middlewared/middlewared/plugins/acme_protocol_/authenticators/route53.py)

bmeagherix (approved):
LGTM.  Nice!

sonicaj (approved):

PR #12368, opened 2023-10-20
NAS-124691 / 24.04 / Add changes to hide sensitive parameters in case of exception

sonicaj:
This commit adds changes to hide sensitive details passed on to parameters of certain functions in case an exception is raised from them.


yocalebo (changes requested):
Right idea but this should be a decorator to reduce the probability of human error. (i.e. if we rename a method or replace it, the hard-coded list has to be updated. If it's not, then there is possibility that new method will suffer from same problem)

PR #12373, opened 2023-10-23
NAS-124752 / 24.04 / Add additional metadata to chart context

stavros-k:
See ticket for more information

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
@@ -560,15 +560,20 @@ class ChartReleaseService(CRUDService):
 
         await self.perform_actions(context)
 
+        # When a user edits a chart release, pods will be scaled up again.
+        # This will cause pods to spin up even for a few seconds, before we have a chance to scale it down,
+        # if the previous status was STOPPED. By adding in the context the "desiredStatus" as "STOPPED",
+        # we are letting the app developer to take extra action if wanted. Like explicitly setting the replica count to 0.
         config = await add_context_to_configuration(config, {
             CONTEXT_KEY_NAME: {
                 **get_action_context(chart_release),
                 'operation': 'UPDATE',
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
```diff
@@ -560,15 +560,20 @@ async def do_update(self, job, chart_release, data):
...
             CONTEXT_KEY_NAME: {
                 **get_action_context(chart_release),
                 'operation': 'UPDATE',
+                'desiredStatus': release_orig['status'],
                 'isUpdate': True,
             }
         }, self.middleware, chart_release)
 
-        await self.middleware.call('chart.release.helm_action', chart_release, chart_path, config, 'update')
+        await self.middleware.call('chart.release.helm_action', chart_release, chart_path, config, 'upgrade')
```
stavros-k, line 576: This does not change functionality, just a typo, it was falling into the else case anyway

https://github.com/truenas/middleware/blob/e1f3091ced8c94ab9035a62ff20f8240ec8109c4/src/middlewared/middlewared/plugins/chart_releases_linux/helm.py#L27

src/middlewared/middlewared/plugins/chart_releases_linux/utils.py
```diff
@@ -37,6 +37,7 @@ def get_action_context(release_name):
         'isInstall': False,
         'isUpdate': False,
```
sonicaj, line 38: Let's make changes here as we discussed. Thanks!

src/middlewared/middlewared/plugins/chart_releases_linux/chart_release.py
sonicaj: Please keep it as it was before because it is used in error message and in truenas context it is update and not upgrade which is bound to cause confusion

sonicaj (approved):

stavros-k:
backport

PR #12375, opened 2023-10-23
NAS-124758 / 24.04 / NFS and directory service should wait for dependent services

mgrimesix:
Under some configurations, NFS can be dependent on nslcd and directory service, AD or LDAP (IPA) are nearly always dependent on the network subsystem.
Under some conditions and configurations, NFS and directory service would start before their dependent services.

NFS and nslcd are mostly managed by systemd.
To address NFS dependency on nslcd:  Added `After=nslcd` criteria to nfs-server unit file.

Directory Service is managed by middleware and, if enabled, is started as part of smb.configure.
To address Directory Service dependency on the network subsystem:  Added a sentinel file generated in the ix-netif unit file.  SMB `configure` will wait for the sentinel file before starting Directory Service.

Also did some cosmetic clean up of smb.py for flake8.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -24,10 +24,11 @@ from middlewared.service import TDBWrapConfigService, ValidationErrors, filterab
 from middlewared.service_exception import CallError, MatchNotFound
 from middlewared.plugins.smb_.smbconf.reg_global_smb import LOGLEVEL_MAP
 import middlewared.sqlalchemy as sa
-from middlewared.utils import filter_list, osc, Popen, run
+from middlewared.utils import filter_list, osc, Popen, run, MIDDLEWARE_RUN_DIR
 from middlewared.utils.osc import getmnttree
 from middlewared.utils.path import FSLocation, path_location, is_child_realpath
 
+NETIF_COMPLETE_SENTINEL = f"{MIDDLEWARE_RUN_DIR}/ix-netif-complete"
 CONFIGURED_SENTINEL = '/var/run/samba/.configured'
 SMB_AUDIT_DEFAULTS = {'enable': False, 'watch_list': [], 'ignore_list': []}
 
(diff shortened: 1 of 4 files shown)


yocalebo (changes requested):
Sentinel file won't work because of 2 reasons:
1. we have a `-t 120` (timeout) option in the `ExecStart` section of the ix-netif systemd unit file which means it's not guaranteed that network interfaces are up by the time it's done
2. I was just on a customer system where this systemd service "completed" but the NICs took another 30+ seconds to actually enter an `UP` state.

We'll probably have to have something in middleware that actually does something quite a bit more involved that actually checks to see if the interface that the service is bound to is actually `UP`. But even then, we'll have to have some timeout because we can't actually predict if the NIC will _EVER_ come `UP`. (Maybe switch port is down or someone pulled a cable)

mgrimesix:
Unfortunate, but not surprising. 
A fail-safe timeout is already part of the 'wait'.   I'll craft a better network test.

mgrimesix:
retest this please

mgrimesix:
retest this please

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -494,6 +495,42 @@ async def ctdb_wait(self):
...
+    @private
+    async def netif_wait(self, timeout=120):
+        """
+        Wait for for the ix-netif sentinel file
+        and confirm connectivity with some targeted tests.
+        All must be completed before the timeout.
+        """
+        found_sentinel = False
+        while timeout >= 0 and not found_sentinel:
+            if os.path.exists(NETIF_COMPLETE_SENTINEL):
```
yocalebo, line 507:
```suggestion
            try:
                os.remove(NETIF_COMPLETE_SENTINEL)
                found_sentinel = True
            except FileNotFoundError:
                pass
```

src/middlewared/middlewared/plugins/smb.py
```diff
@@ -494,6 +495,42 @@ async def ctdb_wait(self):
...
+                self.logger.warning('Failed to get netif completion sentinal.')
+            elif not found_sentinel:
+                await asyncio.sleep(1)
+
+        """
+        Confirm at least one network interface is UP
+        """
+        if_UP = False
+        while timeout >= 0 and not if_UP:
+            netif = await self.middleware.call('interface.query')
```
yocalebo, line 522: There is no reason to generate lists of interface names when we just need to detect if any of them are up. I suggest doing something like this
```suggestion
    while timeout >= 0:
        if any((
            i['state']['link_state'] == 'LINK_STATE_UP' for i in await self.middleware.call('interface.query')
        )):
            break
        else:
            timeout -= 1
            if timeout <= 0:
                self.logger.warning('Failed to detect any connected network interfaces.')
            else:
                await asyncio.sleep(1)
```

yocalebo (changes requested):

src/middlewared/middlewared/plugins/smb.py
mgrimesix: 👍

yocalebo (approved):
Thanks!

src/middlewared/middlewared/plugins/smb.py
mgrimesix: 👍

PR #12376, opened 2023-10-23
NAS-124370 / 24.04 / dont start dhcp on eno1 on fseries

yocalebo:
When an f-series is installed cleanly and boots up the 1st time, by design , we start DHCP on all the physical interfaces on the system. With the f-series, we mask the `eno1` interface since it's shared with IPMI.

To ensure this interface is ignored, I added `eno1` check in the `interface.internal_interfaces` method which is what all consumers should be calling. Since I added this logic to that method, I was able to simplify `interface.query` but more importantly, the `interface.sync` method will now ignore `eno1` on the f-series platform when the system boots for the first time.


yocalebo:
sigh, wrong ticket number

PR #12380, opened 2023-10-24
NAS-124620 / 24.04 / Remove /var/tmp usages in middleware

sonicaj:
This PR adds changes to use middlewared run directory instead of consuming /var/tmp directly. Also while at it, changes have been made to remove outdated source alerts which are not being consumed anymore.

### DIFF
--- a/src/middlewared/middlewared/plugins/failover_/fenced.py
+++ b/src/middlewared/middlewared/plugins/failover_/fenced.py
@@ -10,7 +10,7 @@ from middlewared.utils.cgroups import move_to_root_cgroups
 from fenced.fence import ExitCode as FencedExitCodes
 
 
-PID_FILE = os.path.join(MIDDLEWARE_RUN_DIR, '.fenced-pid')
+PID_FILE = os.path.join(MIDDLEWARE_RUN_DIR, 'fenced.pid')
 IS_ALIVE_SIGNAL = 0
 
 


src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
```diff
@@ -118,7 +120,7 @@ http {
 
     gzip  on;
     #upload_store /var/tmp/firmware;
```
themylogin, line 122: This should be removed

src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
```diff
@@ -118,7 +120,7 @@ http {
...
-    client_body_temp_path /var/tmp/firmware;
+    client_body_temp_path ${os.path.join(MIDDLEWARE_RUN_DIR, 'nginx_firmware')};
```
themylogin, line 123: Why is it "firmware"?

themylogin (changes requested):

src/middlewared/middlewared/plugins/update_/utils.py
```diff
@@ -1,14 +1,17 @@
...
 
+from middlewared.utils import MIDDLEWARE_RUN_DIR
+
 DEFAULT_SCALE_UPDATE_SERVER = "https://update.ixsystems.com/scale"
 SCALE_MANIFEST_FILE = "/data/manifest.json"
 
 DOWNLOAD_UPDATE_FILE = "update.sqsh"
 
-UPLOAD_LOCATION = "/var/tmp/firmware"
+UPLOAD_LOCATION = os.path.join(MIDDLEWARE_RUN_DIR, "upload_firmware")
```
themylogin, line 14: Again, why firmware?

src/middlewared/middlewared/plugins/failover_/fenced.py
```diff
@@ -5,11 +5,12 @@
 import signal
 
 from middlewared.service import Service, CallError
+from middlewared.utils import MIDDLEWARE_RUN_DIR
 from middlewared.utils.cgroups import move_to_root_cgroups
 from fenced.fence import ExitCode as FencedExitCodes
 
 
-PID_FILE = '/tmp/.fenced-pid'
+PID_FILE = os.path.join(MIDDLEWARE_RUN_DIR, '.fenced-pid')
```
themylogin, line 13: Why a dotfile, why are we hiding it?

sonicaj: Keeping backwards compatibility but will change that to make it prominent so it reflects what it is

themylogin (approved):

PR #12398, opened 2023-10-27
NAS-124776 / 24.04 / Make sure we don't mount any unencrypted datasets on pool import

sonicaj:
This PR fixes a case where any unencrypted datasets under encrypted root dataset get mounted before we actually get the chance to unlock root dataset.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/import_pool.py
+++ b/src/middlewared/middlewared/plugins/pool_/import_pool.py
@@ -4,7 +4,7 @@ import os
 import subprocess
 
 from middlewared.schema import accepts, Bool, Dict, List, returns, Str
-from middlewared.service import CallError, job, private, Service
+from middlewared.service import CallError, InstanceNotFound, job, private, Service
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/pool_/import_pool.py
```diff
@@ -227,6 +232,7 @@ def import_on_boot_impl(self, vol_name, vol_guid, set_cachefile=False):
             self.logger.error('Unhandled exception importing %r', vol_name, exc_info=True)
             return False
 
+        self.middleware.call_sync('pool.handle_unencrypted_datasets_on_import', vol_name)
```
yocalebo, line 235: Why would you call this here and not in `unlock_on_boot_impl`? We're making 2 duplicate calls to `pool.dataset.get_instance_quick`

yocalebo (changes requested):

yocalebo (approved):

PR #12411, opened 2023-10-29
NAS-124896 / 24.04 / add HA KVM detection

yocalebo:
Similar to what we do when we run our CI against HA VMs running on BHYVE hypervisors, we now have HA VMs working on KVM hypervisor. This adds the necessary changes to our code to detect such an environment and configure the system accordingly.

### DIFF
--- a/src/middlewared/middlewared/plugins/failover_/detect_enclosure.py
+++ b/src/middlewared/middlewared/plugins/failover_/detect_enclosure.py
@@ -22,10 +22,21 @@ class EnclosureDetectionService(Service):
     @cache
     def detect(self):
         HARDWARE = NODE = 'MANUAL'
-        product = self.middleware.call_sync('system.dmidecode_info')['system-product-name']
+        dmi = self.middleware.call_sync('system.dmidecode_info')
+        product = dmi['system-product-name']
         if not product:
             # no reason to continue since we've got no path forward
             return HARDWARE, NODE
+        elif dmi['system-manufacturer'] == 'QEMU':
+            serial = dmi['system-serial-number']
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/failover_/detect_enclosure.py
```diff
@@ -22,10 +22,21 @@ class Config:
...
+        product = dmi['system-product-name']
         if not product:
             # no reason to continue since we've got no path forward
             return HARDWARE, NODE
+        elif dmi['system-manufacturer'] == 'QEMU':
+            serial = dmi['system-serial-number']
+            if not serial.startswith('ha') and not serial.endswith(('_c1', '_c2')):
+                # truenas is often installed in KVM so we need to check our specific
+                # strings in DMI and bail out early here
+                return HARDWARE, NODE
```
anodos325, line 35: Should we have a persistent alert that this is unsupported configuration for production workloads in case someone sees the source and decides to write a howto?

src/middlewared/middlewared/plugins/failover_/detect_enclosure.py
```diff
@@ -22,10 +22,21 @@ class Config:
...
+            else:
+                HARDWARE = 'IXKVM'
+                NODE = 'A' if serial[-1] == '1' else 'B'
+                return HARDWARE, NODE
```
anodos325, line 39: Should we have a persistent alert that this is unsupported configuration for production workloads in case someone sees the source and decides to write a howto?

anodos325 (approved):

PR #12422, opened 2023-10-31
NAS-124842 / 24.04 / Load API keys on failover event

themylogin:


themylogin:
time 0:10

src/middlewared/middlewared/plugins/failover_/event.py
```diff
@@ -581,6 +581,8 @@ def vrrp_master(self, job, fobj, ifname, event):
         self.run_call('alert.block_failover_alerts')
         self.run_call('alert.initialize', False)
 
+        self.run_call('api_key.load_keys')
```
yocalebo, line 584: Does this need to be loaded before we start the truecommand service?

yocalebo (approved):

PR #12423, opened 2023-10-31
NAS-124935 / 24.04 / Add SMB AUTHENTICATION audit events

anodos325:
Now that we have centralized place for audit-style information, remove the handling whereby we wrote auth_audit events into a separate samba log file in favor of direct insertion into our auditing DB. If we need auth info from users, it is simple enough to just request a dump of it from them when troubleshooting.

### DIFF
--- a/src/middlewared/middlewared/plugins/audit/schema/smb.py
+++ b/src/middlewared/middlewared/plugins/audit/schema/smb.py
@@ -333,7 +333,7 @@ AUDIT_EVENT_SMB_SCHEMAS.append(audit_schema_from_base(
 
 AUDIT_EVENT_SMB_SCHEMAS.append(audit_schema_from_base(
     AUDIT_EVENT_SMB_BASE_SCHEMA,
-    'audit_entry_smb_setattr',
+    'audit_entry_smb_set_attr',
     Str(AuditEventParam.EVENT.value, enum=[AuditSmbEventType.SET_ATTR.name]),
     AUDIT_EVENT_DATA_SMB_SET_ATTR
 ))
--- a/src/middlewared/middlewared/plugins/audit/schema/smb.py
+++ b/src/middlewared/middlewared/plugins/audit/schema/smb.py
@@ -39,6 +39,7 @@ class AuditSmbEventType(AuditEnum):
     This enum contains all possible SMB audit events. Values correspond with
     `event` written to auditing SQLite database.
     """
+    AUTHENTICATION = 'AUTHENTICATION'
     CONNECT = 'CONNECT'
     DISCONNECT = 'DISCONNECT'
     CREATE = 'CREATE'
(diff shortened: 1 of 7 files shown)


src/middlewared/middlewared/plugins/smb_/smbconf/reg_global_smb.py
```diff
@@ -86,13 +86,12 @@ def log_level_transform(entry, conf):
 
     def set_log_level(entry, val, data_in, data_out):
         loglevelint = int(LOGLEVEL_MAP.inv.get(val, 1))
```
bmeagherix, line 88: So, we **silently** default to MINIMUM if an invalid `val` is supplied?  Hmm.

src/middlewared/middlewared/plugins/smb_/status.py
```diff
@@ -54,33 +54,11 @@ def status(self, info_level, filters, options, status_options):
...
-
-            except FileNotFoundError:
-                self.logger.warning("SMB auth audit log does not exist "
-                                    "this is expected if users have never "
-                                    "authenticated to this server.")
-
-            return filter_list(ret, filters, options)
+            return self.middleware.call_sync('audit.query', {
+                'services': ['SMB'],
+                'query-filters': filters + [['event', '=', 'AUTHENTICATION']],
```
bmeagherix, line 59: Hmmm.  So you've decided to _special case_ AUTHENTICATION events and automatically add them to the filter.

Won't argue against it, but perhaps a comment with some justification would be useful.  Further, do the method docs need to be updated?

src/middlewared/middlewared/plugins/audit/schema/smb.py
```diff
@@ -65,6 +66,33 @@ class AuditSetattrType(AuditEnum):
 """
 
 
+AUDIT_EVENT_DATA_SMB_AUTHENTICATION = Dict(
+    str(AuditEventParam.EVENT_DATA),
+    Str('logonId'),
+    Int('logonType'),
+    Str('localAddress'),
```
bmeagherix, line 73: `status` appears to be present in samba/auth/auth_log.c

If it's not supposed to be present here, then please add a comment explaining why.

Also wondering about `clientPolicyAccessCheck` and `serverPolicyAccessCheck.`

tests/api2/test_420_smb.py
```diff
@@ -410,16 +410,18 @@ def get_event(event_list, ev_type):
             assert new_data['audit']['ignore_list'] == ['builtin_users'], str(new_data['audit'])
 
             # Verify that being member of group in ignore list is sufficient to avoid new messages
-            assert len(do_audit_ops(s['name'])) == len(events)
+            # By default authentication attempts are always logged
+            assert len(do_audit_ops(s['name'])) == len(events + 1)
```
bmeagherix, line 414: These are fine, but I do also wonder whether it might make sense to ensure the events look right, rather than just relying upon a count.

src/middlewared/middlewared/plugins/smb_/smbconf/reg_global_smb.py
anodos325: Old behavior. Validation occurs earlier.

src/middlewared/middlewared/plugins/smb_/status.py
anodos325: This is actually functionally identical to previous code. With the log level change (no longer writing to this file) and a samba-level change, authentication messages are now written to the audit database. If we want output / API consistency with previous version then we need the extra filter.

src/middlewared/middlewared/plugins/audit/schema/smb.py
anodos325: We use a slightly different function for this than what is in auth/auth_log.c. Status is remapped to `result` and `clientPolicyAccessCheck` and `serverPolicyAccessCheck` are dropped due to being AD DC specific.

tests/api2/test_420_smb.py
anodos325: That's a good idea.

bmeagherix (approved):

PR #12424, opened 2023-10-31
NAS-124782 / 24.04 / Make `disk.label_to_dev` and `disk.label_to_disk` work for cases when…

themylogin:
… label is a partition


themylogin:
time 3:00

yocalebo (approved):

PR #12428, opened 2023-11-01
NAS-124981 / 24.04 / Allow query of roles via account plugin

anodos325:
This commit adds a new option for `additional_information` from user.query and group.query whereby API user can request roles of which the user or group is a member.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -7,6 +7,7 @@ from middlewared.utils import run, filter_list
 from middlewared.validators import Email, Range
 from middlewared.async_validators import check_path_resides_within_volume
 from middlewared.plugins.smb import SMBBuiltin
+from middlewared.plugins.account_.privilege_utils import privileges_group_mapping
 from middlewared.plugins.idmap_.utils import TRUENAS_IDMAP_DEFAULT_LOW
 
 import binascii
@@ -175,6 +176,8 @@ class UserService(CRUDService):
             [], {'prefix': 'bsdgrpmember_'}
         )
 
+        group_roles = await self.middleware.call('group.query', [], {'select': ['id', 'roles']})
+
         for i in res:
             uid = i['user']['id']
             if uid in memberships:
@@ -189,6 +192,7 @@ class UserService(CRUDService):
                     'datastore.query', 'account.twofactor_user_auth', [['user_id', '!=', None]]
                 )
             }),
+            'roles_mapping': {i['id']: i['roles'] for i in group_roles}
         }
 
     @private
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/account.py
```diff
@@ -273,6 +277,8 @@ async def query(self, filters, options):
                     'sid': u['User SID'],
                 }})
 
+        group_role_info = {g['id']: g['roles'] for g in await self.middleware.call('group.query')}
```
themylogin, line 280: We can user `user_extend_context` and `user_extend` methods to join this information using a more uniform way

src/middlewared/middlewared/plugins/account.py
```diff
@@ -1555,6 +1573,22 @@ async def query(self, filters, options):
             else:
                 entry.update({'nt_name': None, 'sid': None})
 
+            roles = []
```
themylogin, line 1576: Can we avoid duplicating this logic by using `privileges_for_groups`? Possibly refactoring some logic out there for performance reasons.

src/middlewared/middlewared/plugins/account.py
```diff
@@ -301,6 +307,13 @@ async def query(self, filters, options):
             else:
                 entry.update({'nt_name': None, 'sid': None})
 
+            user_roles = []
+
+            for group_pk in entry['groups']:
+                user_roles.extend(group_role_info.get(group_pk, []))
```
sonicaj, line 313: Is it possible that `group_pk` does not exist in `group_role_info` ? I guess it's good we are still playing it safe because this endpoint is critical but just curious

src/middlewared/middlewared/plugins/account.py
```diff
@@ -301,6 +307,13 @@ async def query(self, filters, options):
...
+
+            entry['roles'] = list(set(user_roles))
```
sonicaj, line 315: We can probably have `user_roles` as set initially

themylogin (changes requested):

src/middlewared/middlewared/plugins/account.py
anodos325: Ah, good point. I had initially made this configurable.

src/middlewared/middlewared/plugins/account.py
anodos325: We can perhaps have hundreds of local users (I think we had a user with over 500). I'd rather not do N calls to privilege endpoint.

anodos325: Okay. I see how I can do this without performance hit.

src/middlewared/middlewared/plugins/account.py
anodos325: Good idea

themylogin (approved):

PR #12435, opened 2023-11-02
NAS-125005 / 24.04 / Fix periodic sync of SMB share ACL with datastore

anodos325:
SMB share ACLs may be modified by by SMB clients or locally outside of our APIs by intrepid administrators. Currently we periodically check whether the file has changed behind our back and update stored copies of the SMB share ACL (in our config file). The share ACL stored in DB is base64 encoded security descriptor blob from share_info.tdb.

This edge case did not have adequate test coverage and regressed after an API redesign for the cobia release.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb_/sharesec.py
+++ b/src/middlewared/middlewared/plugins/smb_/sharesec.py
@@ -44,7 +44,6 @@ class ShareSec(CRUDService):
         'data_type': 'BYTES'
     }
 
-    @private
     async def fetch(self, share_name):
         return await self.middleware.call('tdb.fetch', {
             'name': LOCAL_SHARE_INFO_FILE,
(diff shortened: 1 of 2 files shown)


mgrimesix (approved):

src/middlewared/middlewared/plugins/smb_/sharesec.py
```diff
@@ -44,15 +44,13 @@ class Config:
         'data_type': 'BYTES'
     }
 
-    @private
```
bmeagherix, line 47: If you are removing private from this (and others), then should some documentation be provided?

TBH, I'd lean more towards keeping these private.  Is there a reason you're changing it?

bmeagherix (changes requested):

src/middlewared/middlewared/plugins/smb_/sharesec.py
anodos325: At some point in the past I made the whole plugin private. These are redundant.

bmeagherix (approved):

PR #12446, opened 2023-11-03
NAS-125043 / 24.04 / Add Sharing Manager builtin privilege

anodos325:
This adds a new builtin privilege with SHARING_MANAGER per requirements for roles epic.


anodos325:
There will be a follow-on PR that adds more plumbing for roles in filesystem / sharing plugins with tests.

themylogin (approved):

PR #12449, opened 2023-11-03
NAS-125052 / 24.04 / Add roles to filesystem plugin

anodos325:
This commit adds filesystem specific roles, and grants FILESYSTEM_READ_ATTRS to the READONLY role since this is required to render many webui components. The SHARING_MANAGER role is granted FILESYSTEM_WRITE_ATTRS so that it is allowed to chown, setperm, setacl, etc.

### DIFF
--- a/src/middlewared/middlewared/plugins/filesystem.py
+++ b/src/middlewared/middlewared/plugins/filesystem.py
@@ -30,7 +30,7 @@ class FilesystemService(Service):
     class Config:
         cli_private = True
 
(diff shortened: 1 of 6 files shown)


themylogin (commented):
Current names follow the `OBJECT_ACTION` pattern, can we add `FILESYSTEM_ATTRS_READ` and `FILESYSTEM_DATA_READ` instead?

themylogin (approved):

PR #12450, opened 2023-11-03
NAS-125055 / 24.04 / Add roles for SMB share ACL and status

anodos325:
This allows properly authorized users to view / set SMB share ACL and view SMB status.


bmeagherix (changes requested):
Add test for `client_count`?

Also, do the `get_remote_acl` and `presets` APIs need to have a role set?

bmeagherix (approved):

mgrimesix (approved):

PR #12456, opened 2023-11-06
NAS-125085 / 24.04 / Do not apply no_authz_required for API keys

anodos325:
When an administrator generates an API key they explicity whitelist one or more API methods that may be called via the key. This means that authorizations checks for API keys must always be strict to avoid unexpected access being granted.

### DIFF
--- a/src/middlewared/middlewared/auth.py
+++ b/src/middlewared/middlewared/auth.py
@@ -4,6 +4,8 @@ from middlewared.utils.allowlist import Allowlist
 
 
 class SessionManagerCredentials:
+    is_user_session = False
+
     @classmethod
     def class_name(cls):
         return re.sub(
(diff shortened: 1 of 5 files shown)


themylogin (changes requested):
Add a test please

anodos325:
Sure thing.

src/middlewared/middlewared/auth.py
```diff
@@ -38,6 +40,7 @@ class UserSessionManagerCredentials(SessionManagerCredentials):
     def __init__(self, user):
         self.user = user
         self.allowlist = Allowlist(user["privilege"]["allowlist"])
+        self.is_user_session = True
```
themylogin, line 43: This will fail for UI as they generate a token for currently authenticated user and then use that token (so it'll be `TokenSessionManagerCredentials`.

Let's make this a method instead, say, `use_no_authz_required`. Default implementation will return `False`, `UserSessionManagerCredentials` will return `True`, and `TokenSessionManagerCredentials` will return the value for its parent credentials.

themylogin (changes requested):

src/middlewared/middlewared/auth.py
anodos325: Okay. Probably I'll update PR tomorrow.

anodos325:
Added tests (which passed) and fixed is_user_session for auth tokens.

themylogin (approved):

PR #12460, opened 2023-11-07
NAS-125056 / 24.04 / Set up roles for iSCSI-related API endpoints

bmeagherix:
At the moment this INCOMPLETE PR contains a **tentative** direction of travel wrt adding iSCSI roles - for discussion purposes.

Previously `SHARING_ISCSI_EXTENT_[READ|WRITE]` had been added & it _could_ make sense to retain a **fine-grained** level of control to the various iSCSI-related APIs.

After all, grouping them as subroles (if that's the right term) under `SHARING_ISCSI_READ` & `SHARING_ISCSI_WRITE` doesn't cost us any real additional effort, and does allow us to split things later if we so choose.  (For example, we _could_ allow a user to change the initiators allowed to talk to a target, but prevent the user from being able to remove the target or its underlying extent.)

---

Add several role_prefix statements for various iSCSI APIs

Add the following role_prefixes:
- `SHARING_ISCSI_AUTH`
- `SHARING_ISCSI_GLOBAL`
- `SHARING_ISCSI_INITIATOR`
- `SHARING_ISCSI_PORTAL`
- `SHARING_ISCSI_TARGET`
- `SHARING_ISCSI_TARGETEXTENT`

The new `SHARING_ISCSI_READ` role will include `SHARING_ISCSI_AUTH_READ`, `SHARING_ISCSI_EXTENT_READ`, etc.

Likewise, `SHARING_ISCSI_WRITE` role will include `SHARING_ISCSI_AUTH_WRITE`, `SHARING_ISCSI_EXTENT_READ`, etc.


src/middlewared/middlewared/plugins/iscsi_/portal.py
```diff
@@ -64,7 +65,7 @@ async def config_extend(self, data, context):
         data['discovery_authgroup'] = data.pop('discoveryauthgroup')
         return data
 
-    @accepts()
+    @accepts(roles=['SHARING_ISCSI_PORTAL_READ'])
```
themylogin, line 68: This is not necessary, because, as the doc states https://github.com/truenas/middleware/blob/master/docs/source/middleware/roles.rst#id4, `SERVICE_READ` (and it will have access to calling `service.config`, `service.get_instance`, `service.query` and all  `service.*_choices` methods)

src/middlewared/middlewared/plugins/iscsi_/targets.py
```diff
@@ -256,7 +257,9 @@ async def __validate(self, verrors, data, schema_name, old=None):
                             f'Authentication group {group["auth"]} does not support CHAP Mutual'
                         )
 
-    @accepts(Str('name'), Int('existing_id', null=True, default=None))
+    @accepts(Str('name'),
+             Int('existing_id', null=True, default=None),
+             roles=['SHARING_ISCSI_TARGET_READ'])
```
themylogin, line 262: Does `READ` role need this? Or only `WRITE` one?

themylogin (commented):
Received your mail, this approach looks good to me!

src/middlewared/middlewared/plugins/iscsi_/portal.py
bmeagherix: Right.  Thanks!

src/middlewared/middlewared/plugins/iscsi_/targets.py
bmeagherix: You are correct, even though this is a READ operation, it should only be needed by the `WRITE` role.  Thanks.

bmeagherix:
Wrong branch name.  Closing PR, deleting branch and will recreate.

PR #12461, opened 2023-11-07
NAS-125099 / 24.04 / Document fields for get_nfsX_clients

mgrimesix:
Describe the fields in the midclt calls get_nfs[34]_clients


anodos325:
SMB has a public `smb.status` endpoint. I think these should be public as well (for consistency sake). This means that you'll need to add an accepts and returns decorator. The latter can be used to document expected return value for webui folks (they have open ticket to add NFS client info via these endpoints).

bmeagherix (approved):

PR #12462, opened 2023-11-07
NAS-123854 / 24.04 / add hseries enclosure mapping

yocalebo:
Add h-series enclosure management to legacy enclosure code. This is a bit hacky but this platform uses a Virtual SES driver, for which we have outright ignored for the entirety of our enclosure plugin. To try and keep the potential for regression to a minimum, I've hidden the new logic (most of it) behind a `if is_hseries` check.

### DIFF
--- a/src/middlewared/middlewared/plugins/enclosure.py
+++ b/src/middlewared/middlewared/plugins/enclosure.py
@@ -98,8 +98,11 @@ class EnclosureService(CRUDService):
                         "descriptor": elem.descriptor,
                         "status": elem.status,
                         "value": elem.value,
-                        "value_raw": hex(elem.value_raw),
+                        "value_raw": 0x0,
                     }
+                    if isinstance(elem.value_raw, int):
+                        element["value_raw"] = hex(elem.value_raw)
+
                     if hasattr(elem, "device_slot_set"):
                         has_slot_status = True
                         element["fault"] = elem.fault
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/enclosure.py
```diff
@@ -98,8 +98,11 @@ def query(self, filters, options):
                         "descriptor": elem.descriptor,
                         "status": elem.status,
                         "value": elem.value,
-                        "value_raw": hex(elem.value_raw),
+                        "value_raw": 0x0,
                     }
+                    if isinstance(elem.value_raw, int):
+                        element["value_raw"] = hex(elem.value_raw)
```
bmeagherix, line 104: Might be an idea to add a comment explaining **why** we could get a non-int as the `value_raw`.

src/middlewared/middlewared/plugins/enclosure.py
```diff
@@ -478,19 +504,27 @@ def map_disks_to_enclosure_slots(self):
...
+                info = self._enclosure_element(slot, 'Array Device Slot', slot_status, None, '', disk, ident, fault)
+            else:
+                info = self._enclosure_element(
+                    slot,
+                    'Array Device Slot',
+                    self._parse_raw_value(disk_raw_values.get(slot, [5, 0, 0, 0])),
+                    None,
+                    '',
+                    disk
+                )
```
bmeagherix, line 524: Possibly worthwhile to comment **why** we didn't supply the `ident` and `fault` here too, or instead **why** we did include them for H-series.

bmeagherix (approved):
Approving with a slight caveat.

There are a couple of places in the code where you have made accommodations for the unusual behavior of VirtualSES.  IMO for clarity it would be good to spell out in comments the different behavior of VirtualSES - rather than having to work backwards from the code.

PR #12463, opened 2023-11-07
NAS-125056 / 24.04 / Set up roles for iSCSI-related API endpoints

bmeagherix:
Add iSCSI roles.

(Had previously generated a PR (#12460) using the wrong branch name.)


bmeagherix:
Removing WIP following jenkins run [#1294](https://ci.tn.ixsystems.net/jenkins/job/TrueNAS%20SCALE%20-%20Unstable/job/API%20Tests%20-%20TrueNAS%20SCALE%20with%20middleware%20reinstalled%20from%20a%20branch%20Incremental/1294/)

yocalebo:
@truenas/middleware ping

yocalebo:
@themylogin the original PR here: https://github.com/truenas/middleware/pull/12460 was something that you reviewed but it was under wrong branch name. However, this one was opened up and I think he implemented the changes you requested. Can you spot check this to make sure it looks good.

tests/api2/test_261_iscsi_cmd.py
```diff
@@ -285,6 +285,21 @@ def zvol_resize(zvol, volsize, expected_status_code=200):
     assert result.status_code == expected_status_code, result.text
 
 
+def get_iscsi_sessions(payload=None, check_length=None):
+    results = GET('/iscsi/global/sessions', payload=payload)
```
yocalebo, line 289: Why not use websocket calls here in `get_iscsi_sessions` and `get_client_count`?

yocalebo (approved):

src/middlewared/middlewared/service/decorators.py
```diff
@@ -25,6 +25,19 @@ def filterable(fn):
     return accepts(Ref('query-filters'), Ref('query-options'))(fn)
 
 
+# When we want to specify roles with filterable call this
+# i.e. @filterable -> @filterable_with_roles(['role1',...])
+# MUCH simplier than the alternatives.
```
themylogin, line 30: IMO you write the implementation once, but you use it multiple times, so the simplicity of the usage shall be preferred over the simplicity of implementation. E.g. built-in `@dataclass` decorator can be used both as `@dataclass` and `@dataclass()`, that's what we can use as an example.

bmeagherix: Good point, thanks.  I'll address.

themylogin (approved):
Thank you!

PR #12464, opened 2023-11-08
NAS-124351 / 24.04 / Add UPS reporting

Qubad786:
## Context

It was requested that UPS reporting be added back to TrueNAS as it was removed when we moved to netdata for gathering reporting information.


yocalebo (approved):
Be sure and get with webUI team so that they know this is now being brought back. I'm not sure if they have to make changes on their side or not.

PR #12472, opened 2023-11-09
NAS-124416 / 24.04 / Add an endpoint to retrieve PCI devices which are to be isolated for VM GPU isolation

sonicaj:
This PR adds a new endpoint which UI should use when trying to determine which PCI devices are to be created for isolation when user selects a GPU to be attached to a VM.

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/pci.py
+++ b/src/middlewared/middlewared/plugins/vm/pci.py
@@ -1,4 +1,5 @@
 import collections
+import errno
 import os
 import pathlib
 import re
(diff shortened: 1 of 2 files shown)


Qubad786 (approved):

src/middlewared/middlewared/plugins/vm/pci.py
```diff
@@ -186,3 +189,27 @@ def passthrough_device_choices(self):
...
+
+    @accepts(Str('gpu_pci_id', empty=False))
+    @returns(List(Str('pci_ids')))
+    def get_pci_ids_for_gpu_isolation(self, gpu_pci_id):
+        """Get PCI IDs for GPU isolation"""
+        gpu = next(
+            (gpu for gpu in get_gpus() if convert_pci_id_to_vm_pci_slot(gpu['addr']['pci_slot']) == gpu_pci_id), None
+        )
+        if not gpu:
+            raise CallError(f'GPU {gpu_pci_id} not found', errno=errno.ENOENT)
```
yocalebo, line 201: If this is to be called by webui then we should raise a validation error.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/vm/pci.py
sonicaj: Sure, although this won't be a user input - so no form is involved. Will make relevant changes, thanks!

yocalebo (approved):
Thanks!

PR #12473, opened 2023-11-09
NAS-124864 / 24.04 / Update VM teardown memory changes

sonicaj:
This PR adds changes to remove the default arc max size which was defined earlier as half of what the memory is but with recent changes in openzfs, that is not the case now and we should make appropriate amends.

### DIFF
--- a/src/middlewared/middlewared/plugins/sysctl/sysctl_info.py
+++ b/src/middlewared/middlewared/plugins/sysctl/sysctl_info.py
@@ -20,9 +20,7 @@ class SysctlService(Service):
         return cp.stdout.decode().split('=')[-1].strip()
 
     def get_default_arc_max(self):
-        # Default value of arc max on systems
-        # https://github.com/truenas/zfs/blob/f38f3a4a3ee430a15fa221ba9fc2b6b8fd17c040/module/os/linux/zfs/arc_os.c#L85
-        return psutil.virtual_memory().total // 2
+        return psutil.virtual_memory().total
 
     def get_arc_max(self):
         return self.get_arcstats()['c_max']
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/sysctl/sysctl_info.py
```diff
@@ -20,9 +20,8 @@ async def get_value(self, sysctl_name):
         return cp.stdout.decode().split('=')[-1].strip()
 
     def get_default_arc_max(self):
-        # Default value of arc max on systems
-        # https://github.com/truenas/zfs/blob/f38f3a4a3ee430a15fa221ba9fc2b6b8fd17c040/module/os/linux/zfs/arc_os.c#L85
-        return psutil.virtual_memory().total // 2
+        # This is the default value for arc_max on linux (https://github.com/openzfs/zfs/pull/15437)
+        return psutil.virtual_memory().total
```
yocalebo, line 24: This isn't right according to the PR you linked. Furthermore, I'm not sure that we need to be setting/changing the arc sysctls anymore since it seems like libvirt/zfs arc are aware of each other?

We need to check `zfs_arc_max` sysctl and if it's 0, it means that ZFS is implementing the `default_arc_max` function that you link in the PR. Essentially we need to be doing something like

```
size = sysctl/zfs_arc_max
if  size == 0:
    totalmem = psutil.virtual_memory().total
    _32MB = 2 << 24
    _1GB = 1 << 30
    arc_min = max(totalmem / 32, _32MB)
    if totalmem >= _1GB:
        size = totalmem - _1GB
    else:
        size = arc_min
    size = max(totalmem * (5 / 8), size)

return size

yocalebo (changes requested):

PR #12482, opened 2023-11-11
NAS-125187 / 24.04 / Allow limited users to interact with own jobs

anodos325:
This commit alters behavior of job-related core plugin methods in the following ways:

1. Authorization is no longer required for core.get_jobs, core.job_wait, and core.job_abort.
2. Non-admin users calling core.get_jobs will only receive info about their own jobs
3. Non-admin users will only be able to wait for, abort, or modify their own jobs

To achieve this a few minor changes were made:
1. Allowlist object now stores boolean indicating whether the allow list contained an unrestrained full access entry
2. TokenSessionManagerCredentials now contains a reference to the user info from the token's root (originating) credentials.
3. Job encoding now includes a `user` key that contains the username for user sessions. In case of API key or internal jobs, this will be None / null.

### DIFF
--- a/src/middlewared/middlewared/job.py
+++ b/src/middlewared/middlewared/job.py
@@ -527,11 +527,6 @@ class Job:
 
     def __encode__(self, raw_result=True):
         exc_info = None
-        if self.credentials and self.credentials.is_user_session:
-            username = self.credentials.user['username']
-        else:
-            username = None
-
         if self.exc_info:
             etype = self.exc_info[0]
             evalue = self.exc_info[1]
@@ -569,7 +564,6 @@ class Job:
             'state': self.state.name,
             'time_started': self.time_started,
             'time_finished': self.time_finished,
-            'user': username,
             'credentials': (
                 {
                     'type': self.credentials.class_name(),
--- a/src/middlewared/middlewared/job.py
+++ b/src/middlewared/middlewared/job.py
@@ -86,6 +86,19 @@ class JobsQueue(object):
     def all(self):
         return self.deque.all()
 
+    def for_username(self, username):
+        out = {}
+        for jid, job in self.all().items():
+            if job.credentials is None or not job.credentials.is_user_session:
+                continue
+
+            if job.credentials.user['username'] != username:
+                continue
+
+            out[jid] = job
+
+        return out
+
     def add(self, job):
         self.handle_lock(job)
         if job.options["lock_queue_size"] is not None:
@@ -514,6 +527,11 @@ class Job:
 
     def __encode__(self, raw_result=True):
         exc_info = None
+        if self.credentials and self.credentials.is_user_session:
+            username = self.credentials.user['username']
+        else:
+            username = None
+
         if self.exc_info:
             etype = self.exc_info[0]
             evalue = self.exc_info[1]
@@ -551,6 +569,7 @@ class Job:
             'state': self.state.name,
             'time_started': self.time_started,
             'time_finished': self.time_finished,
+            'user': username,
             'credentials': (
                 {
                     'type': self.credentials.class_name(),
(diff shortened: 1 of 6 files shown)


anodos325:
Will add docs and tests shortly.

src/middlewared/middlewared/job.py
```diff
@@ -551,6 +569,7 @@ def __encode__(self, raw_result=True):
             'state': self.state.name,
             'time_started': self.time_started,
             'time_finished': self.time_finished,
+            'user': username,
```
themylogin, line 572: We already have this information in the `credentials` field below

src/middlewared/middlewared/service/core_service.py
```diff
@@ -134,6 +135,32 @@ def get_tasks(self):
                 'frames': frames,
             }
 
+    @private
```
themylogin, line 138: This decorator is not needed for methods that are already private

src/middlewared/middlewared/service/core_service.py
```diff
@@ -172,25 +199,52 @@ def get_tasks(self):
...
+            i.__encode__(raw_result) for i in jobs
         ], filters, options)
         return jobs
 
+    @no_authz_required
     @accepts(Int('id'))
     @job()
     async def job_wait(self, job, id_):
-        return await job.wrap(self.middleware.jobs[id_])
+        target_job = self.middleware.jobs[id_]
```
themylogin, line 225: Should we add a method that gets the job by ID and checks that current session can see it?

src/middlewared/middlewared/service/core_service.py
```diff
@@ -172,25 +199,52 @@ def get_tasks(self):
...
+    @no_authz_required
     @accepts(Int('id'), Dict(
         'job-update',
         Dict('progress', additional_attrs=True),
     ))
-    def job_update(self, id_, data):
+    @pass_app()
+    def job_update(self, app, id_, data):
         job = self.middleware.jobs[id_]
+        self.__check_job_owned_by_session(job, app)
```
themylogin, line 247: This is called by worker from another process pool. It should be private and this check is not needed (no users will update their jobs). `@no_authz_required` should also not be necessary

src/middlewared/middlewared/service/core_service.py
```diff
@@ -593,6 +650,7 @@ async def download(self, app, method, args, filename, buffered):
         return job.id, f'/_download/{job.id}?auth_token={token}'
 
     @private
+    @no_authz_required
```
themylogin, line 653: This is a private method for testing, this decorator is not necessary there

themylogin (changes requested):

src/middlewared/middlewared/job.py
anodos325: I thought maybe it would be nice to have it explicitly listed (for filter usage) for case where `credentials` is token.

src/middlewared/middlewared/service/core_service.py
anodos325: Whoops. Don't know why I did that.

src/middlewared/middlewared/service/core_service.py
anodos325: Good to know. I'll back out those changes

src/middlewared/middlewared/service/core_service.py
anodos325: I was removing authorization requirement so that I could use it for testing job behavior with various limited users.

src/middlewared/middlewared/job.py
anodos325: Per our conversation, I reworked to include in credential dump.

themylogin (approved):
Thank you!

PR #12485, opened 2023-11-13
NAS-125204 / 24.04 / Add cobia reporting exporter migration to dragonfish

Qubad786:
## Context

With reporting exporter changes backported to cobia - a merge migration has been added to master and the migration also added to dragonfish.

### DIFF
--- a/docs/source/database/migrations.rst
+++ b/docs/source/database/migrations.rst
@@ -122,7 +122,15 @@ you have to do the following:
 
 * Switch to the stable branch, cherry-pick the code, remove master migration file and generate the same migration (it
   will be the same `upgrade` code, but different file name, `revision` and `down_revision` values).
-* Change the code to be idempotent so it won't fail if the corresponding DB changes were already made.
+* Changes to the code are to be idempotent so it won't fail if the corresponding DB changes were already made.
+  This can be done like the example given below from `src/middlewared/middlewared/alembic/versions/23.10/2023-11-10_23-19_add_exporting_table_for_reporting.py`
+  where we check if the table already exists and return early if it does, not doing anything else:
+
+.. code-block:: python
+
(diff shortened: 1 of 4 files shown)


yocalebo (changes requested):
Check the build log but this fails during alembic migration phase. Stating it can't find `system_reporting` table.

src/middlewared/middlewared/alembic/versions/23.10/2023-11-10_23-19_add_exporting_table_for_reporting.py
```diff
@@ -0,0 +1,63 @@
...
+revision = '8f8942557260'
+down_revision = '304e43883592'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    conn = op.get_bind()
+    inspector = Inspector.from_engine(conn)
+    if 'reporting_exporters' in inspector.get_table_names():
```
themylogin, line 23: Good job encountering this! Please also add this to the appropriate docs section for a specific example on how to make an idempotent migration

themylogin (changes requested):

docs/source/database/migrations.rst
```diff
@@ -126,6 +126,9 @@ you have to do the following:
 * Merge newly created stable branch migration to the master branch.
 * Also make the original master migration code idempotent (highly likely, it'll just be the same code as in stable
   branch).
+* An example for such a case is `src/middlewared/middlewared/alembic/versions/23.10/2023-11-10_23-19_add_exporting_table_for_reporting.py`
```
themylogin, line 129: Be more specific please. “For example, one can check if a table already exists by doing xxx, you can see file yyy as an example’.

themylogin (approved):

yocalebo (approved):

PR #12488, opened 2023-11-13
NAS-125109 / 24.04 / Prevent password_disable for SMB users

anodos325:
At present SMB authentication for local users requires some form of NT hash. User expectation of disabling password-based auth for local users on SMB access is ambiguous at best, and so we should just raise a ValidationError and have user make some other choices regarding parameter combination.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -687,6 +687,14 @@ class UserService(CRUDService):
             except CallError as e:
                 verrors.add('user_update.password_disabled', e.errmsg)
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/account.py
```diff
@@ -687,6 +687,14 @@ def do_update(self, app, pk, data):
             except CallError as e:
                 verrors.add('user_update.password_disabled', e.errmsg)
 
+            if (has_smb := data.get('smb')) is None:
```
themylogin, line 690: Why do we have this validation duplicated here and there?

themylogin (approved):

PR #12489, opened 2023-11-13
NAS-125078 / 24.04 / Fix swap partition being created after the data partition when the da…

themylogin:
…ta partition size was specified explicitly


yocalebo (approved):
Impressive

PR #12490, opened 2023-11-14
NAS-125212 / 24.04 / fix smb.update stall waiting for sentinel file

mgrimesix:
At boot, smb.configure holds up waiting for the network subsystem to complete configuration.  To achieve this a sentinel file is generated at the completion of network configuration.  The sentinel file was removed after smb.configure detected it's presence.  However, without the sentinel subsequent calls to smb.update would stall (timeout waiting for the sentinel)

The sentinel file is created in /var/run/middleware. The /var/run/middleware directory is tmpfs and starts with zero files on every boot. We do not need to actively manage the removal of the ix-netif-sentinel. 

This change eliminates the stall on calling smb.update during runtime

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -504,7 +504,7 @@ class SMBService(TDBWrapConfigService):
         """
         found_sentinel = False
         while timeout >= 0 and not found_sentinel:
-            if os.path.exists(NETIF_COMPLETE_SENTINEL):
+            if await self.middleware.run_in_thread(os.path.exists, NETIF_COMPLETE_SENTINEL):
                 found_sentinel = True
 
             timeout -= 1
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -504,11 +504,8 @@ async def netif_wait(self, timeout=120):
         """
         found_sentinel = False
         while timeout >= 0 and not found_sentinel:
-            try:
-                os.remove(NETIF_COMPLETE_SENTINEL)
+            if os.path.exists(NETIF_COMPLETE_SENTINEL):
```
anodos325, line 507: We shouldn't be doing file IO in async coroutine. You can either make method synchronous or run this in thread via self.middleware.run_in_thread.

anodos325 (approved):

PR #12491, opened 2023-11-14
NAS-125225 / 24.04 / finalize enclosure2 (and fix many issues)

yocalebo:
I've rewritten this about 4 times at this point. The changes introduced here do many things, the primary one being making the enclosure2.* related endpoints actually function and return _correct_ drive slot mapping information for our platforms. This all spawned out of necessity because of the new H series platform that I'm working on. The new platform diverges pretty dramatically from what we traditionally do and it exposed many numerous issues with how we had designed our enclosure plugin. To our own defense, it's virtually impossible to write "future proof" code in this area. Ideally, however, with hindsight being 20-20, this new code has been written in a way that is as clear and concise as possible while also "future proofing" (if that's even possible). The simple fact of the matter is that we're dealing with complicated logic and it's going to be somewhat complex no matter how we slice it.

Highlights of what this does:

1. stop treating each platform differently and mapping some but not all, this maps ALL of our enclosures/JBODs for every platform we sell. This necessitates, that when adding a new JBOD/platform, the changes be explicit instead of a large amount of the information being filled in implicitly
2. move the array device (disk slot) mapping logic into the `enclosure_class.py:Enclosure` class
3. simplify `map2.py` and also remove most of its logic by adding a singular function `combine_enclosures`
4. move the drive slot mapping logic for our "fake" nvme enclosures into `nvme2.py`
5. add docstrings _everywhere_
6. fix and add unit tests for all the things

With these changes, I'm able to populate a M60 with 12x ES102 JBODs (fully populated) in ~0.69 seconds. Compared to what we're currently getting, it takes about ~1.8 seconds. Finally, and arguably the most important, is that this now returns correct disk slot mapping information and is much more easily understood how we got those mappings.


src/middlewared/middlewared/plugins/enclosure_/map2.py
```diff
@@ -1,71 +1,59 @@
...
-            # we're on a platform that has a JBOD attached so we
-            # want to place these in another list so we don't lose
-            # them during the mapping process
-            jbods.append(enclosure)
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def to_ignore(enclosure, model):
```
anodos325, line 6: We should probably return False on default case here so that type returned remains consistent.

yocalebo: Good catch, fixed it.

src/middlewared/middlewared/plugins/enclosure_/map2.py
```diff
@@ -1,71 +1,59 @@
...
+    if not enclosure['controller']:
+        # this is a JBOD and doesn't need to
+        # be "combined" into any other object
+        return True
+    elif model.startswith(('F60', 'F100', 'F130', 'R30'))
```
anodos325, line 11: Missing colon

src/middlewared/middlewared/plugins/enclosure_/map2.py
```diff
@@ -1,71 +1,59 @@
...
-                        'enclosure_bsg': enclosure['bsg'],
-                        'descriptor': orig_info['descriptor'],
-                        'slot': orig_slot
-                    }
-                }
-            })
-
-    # we've mapped the head-unit but we don't want to lose the JBOD information
-    return mapped + jbods
+        elif enclosure['elements']['Array Device Slot'].get(1):
```
anodos325, line 41: Might improve readability if we define constant like `HEAD_UNIT_DISK_SLOT_NUMBER = 1` then get it here.

yocalebo: done

src/middlewared/middlewared/plugins/enclosure_/identification.py
```diff
@@ -1,46 +1,42 @@
...
+                return '', False
         case 'ECStream_FS1' | 'ECStream_FS2' | 'ECStream_DSS212Sp' | 'ECStream_DSS212Ss':
             # R series
             return model, True
-        case 'iX_FS1' | 'iX_FS2' | 'iX_DSS212Sp' | 'iX_DSS212Ss':
+        case 'iX_FS1L' | 'iX_FS2' | 'iX_DSS212Sp' | 'iX_DSS212Ss':
             # more R series
             return model, True
-        case 'iX_TrueNAS R20p' | 'iX_TrueNAS 2012Sp' | 'iX_TrueNAS SMC SC826-P':
+        case 'iX_TrueNASR20p' | 'iX_2012Sp' | 'iX_TrueNASSMCSC826-P':
```
anodos325, line 28: Doesn't really have to be fixed here, but we need to consider creating / using an enum for all these strings. Looks like it would be quite easy to typo some of them somewhere.

src/middlewared/middlewared/plugins/enclosure_/slot_mappings.py
```diff
@@ -121,11 +155,35 @@ def get_slot_info(model, enclosures=None):
                     'id': {
                         ses_logical_ids[min_id]: {
                             # 1 - 24
-                            i: {'orig_slot': i, 'mapped_slot': i} for i in range(1, 25)
+                            i: {'sysfs_slot': i, 'mapped_slot': i} for i in range(1, 25)
```
anodos325, line 158: I think strings `sysfs_slot` and `mapped_slot` should be defined as constants, e.g. `SYSFS_SLOT_KEY` and `MAPPED_SLOT_KEY` just because they're used so many times it will be easy to gloss over a typo in this file.

yocalebo: Good point, fixed.

yocalebo:
> To add my 'two-cents' to the comment by Andrew...
> 
> * In general,  I would avoid using 'marketing' names.  Marketing could decide to change the names, but the underlying hardware is essentially the same.  Use platform 'code names' that we control.  We probably don't care if a unit has 4 cores or 24 cores, but that could result in a 'marketing' name change to what is essentially the same platform.
> * The 'startswith' tests indicate there might be a need to create some attributes attached to the enclosures.  Platform with a list of attributes.  The list could be extended dynamically if add-on hardware was detected.  For hardware, PCI IDs are good identifiers (again avoiding using marketing names).

Those are valid points but which marketing names are you referring to specifically? The `model`? (i.e M60, R30, etc?)

mgrimesix:
Yeah.  My suggestion is probably not a 'quick change' fix.  🤷‍♂️

yocalebo:
> Yeah. My suggestion is probably not a 'quick change' fix. 🤷‍♂️

Those values are actually determined by SMBIOS information (burned in by our production team before we ship the product).  That's about as close to determining the product by a hardware attribute as you can get, I would imagine. (Unless I'm completely misunderstanding haha)

mgrimesix:
On the flip side, the underlying hardware might change, but to marketing it's the same unit.  If we had platform code names we would then map any number of platforms to the same 'public' name.

yocalebo:
> On the flip side, the underlying hardware might change, but to marketing it's the same unit. If we had platform code names we would then map any number of platforms to the same 'public' name.

That's correct and this has happened. But that's why this code exists in get_slot_info for which I wrote this docstring
```
        `versions->[vers_key]` is a dictionary that represents the
            version. So, for example, if we ship an R20 and the
            `any_versions` key is True, then we will access the
            `versions->DEFAULT` key by "default". However, if
            `any_versions` is False, then there should be another
            top-level key that represents the identifier for that
            version on the platform.
            (i.e. {'versions':
                      'DEFAULT': ...
                      '1.0': ...
                      '2.0': ...
                      etc ...
                  }
            )
```

If an underlying hardware change is made to existing platform that is transparent to the end-user, then there are internal processes by which all new units using said hardware change have to be flashed in SMBIOS with a new version. This allows us to make any changes on our side (if required). For the most part, nothing is required but there could be a case where an internal wiring change has to be made which requires us to change the way the drives are mapped. In that scenario, we have to be able to maintain the original version with the new version.

bmeagherix (commented):
Obviously this is a huge PR (well done!).  However, this means I could have missed some of the high-level view while reviewing the details.

IMO it would be useful to have a middleware/docs document that presented a high-level (developer) view of enclosure2.  (Similar to the one for [Roles](https://github.com/truenas/middleware/blob/master/docs/source/middleware/roles.rst).)

src/middlewared/middlewared/plugins/enclosure_/nvme2.py
```diff
@@ -35,12 +39,13 @@ def fake_nvme_enclosure(model, num_of_nvme_slots, mapped):
         device = mapped.get(slot, None)
         if device is not None:
             status = 'OK'
-            value_raw = '0x1000000'
+            value_raw = 16777216
         else:
-            status = 'Not Installed'
-            value_raw = '0x0500000'
+            status = 'Not installed'
+            value_raw = 83886080
```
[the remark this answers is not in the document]
yocalebo: The old values were actually bugged in 2 ways:
1. they were a string and so were never being converted to their integer value
2. they were the wrong hex value

This value is what should be returned from sg3_utils api when we're talking to proper SES. Since this is NVMe (which uses PCIe), we "fake" the value. I can go back to using the proper hex value.

src/middlewared/middlewared/plugins/enclosure_/slot_mappings.py
```diff
@@ -87,14 +91,46 @@ def get_slot_info(model, enclosures=None):
...
+                2: {'descriptor': '<empty>', 'status': [5, 0, 0, 0], 'type': 23},
+                3: {'descriptor': '<empty>', 'status': [1, 0, 0, 0], 'type': 23},
+                4: {'descriptor': '<empty>', 'status': [5, 0, 0, 0], 'type': 23},
+            }}
+
+            If we take look at the elements[1] key we might think that the device slot is 1 but
+            it's not guaranteed. If we compare what sysfs gives us for slot 1 we see something
+            like this:
+            root@truenas[~]# cat /sys/class/enclosure/0:0:0:0/9/slot
+            root@truenas[~]# 1
```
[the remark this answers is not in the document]
yocalebo: Yes that's correct. I explain what is needed to be provided from the platform team right below this section
```
            If we compare the sysfs output with the dictionary response, we can see that the directory
            name actually is `9` which represents the drive slot reported by the HBA. This means
            what we get from sg3_utils does not match 1 to 1 with sysfs. So how we determine how to
            "map" the drives to their "original slots" is doing 2 things:

                1. platform team needs to give us a sysfs `slot` mapping (i.e. what `slot` maps to
                    what physical slot in the enclosure) (i.e. `slot` 0 is physical slot 8, etc)
                2. take the `EnclosureDevice().status()` output and map the index values to their
                    respective sysfs `slot` files

File: src/middlewared/middlewared/plugins/enclosure_/identification.py (hunk 4 above)
Comment (yocalebo) [reply src/middlewared/middlewared/plugins/enclosure_/identification.py:28]: I can't use a traditional python enum for these strings. It's a 1-to-many relationship. (i.e. ES24 enclosure has 2 different strings that map back to it. Enums won't work for this) Furthermore, they're only used in this function.

File: src/middlewared/middlewared/plugins/enclosure_/slot_mappings.py [hunk 8]
```diff
@@ -87,14 +148,46 @@ def get_slot_info(model, enclosures=None):
             {'versions': {
                 'DEFAULT': {
                     'product': {'eDrawer4048S1' : {
-                        1: {'orig_slot': 1, 'mapped_slot': 1},
-                        5: {'orig_slot': 5, 'mapped_slot': 2},
+                        1: {'sysfs_slot': 1, 'mapped_slot': 1},
+                        5: {'sysfs_slot': 5, 'mapped_slot': 2},
```
anodos325, line 152: Should these be `SYSFS_SLOT_KEY` and `MAPPED_SLOT_KEY`?

anodos325: Nevermind. Lost that this is in a comment

src/middlewared/middlewared/plugins/enclosure_/slot_mappings.py
```diff
@@ -142,290 +238,320 @@ def get_slot_info(model, enclosures=None):
                     'product': {
                         'eDrawer4048S1': {
                             # 1 - 24
-                            i: {'orig_slot': i, 'mapped_slot': i} for i in range(1, 25)
+                            i: {SYSFS_SLOT_KEY: i - 1, MAPPED_SLOT_KEY: i} for i in range(1, 25)
                         },
                         'eDrawer4048S2': {
                             # 25 - 48
-                            i: {'orig_slot': i, 'mapped_slot': j} for i, j in zip(range(1, 25), range(25, 49))
+                            i: {SYSFS_SLOT_KEY: i - 1, MAPPED_SLOT_KEY: j} for i, j in zip(range(1, 25), range(25, 49))
```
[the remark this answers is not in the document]
yocalebo: These changed because sysfs reports the device slot numbers that the HBA reports by default. The device slot numbers for 99% of the HBAs/enclosures that we sell all start at index 0. The only exception is the ES102G2 which starts at 1. So the changes here are fixing a problem 😄 I've added more docs to try and explain this.

src/middlewared/middlewared/plugins/enclosure_/map2.py
```diff
@@ -1,71 +1,65 @@
...
+            k: v for k, v in sorted(enclosures[head_unit_idx]['elements']['Array Device Slot'].items())
+        }
 
-    # we've mapped the head-unit but we don't want to lose the JBOD information
-    return mapped + jbods
+    for idx in reversed(to_remove):
+        # we've combined the enclosures into the
+        # main "head-unit" enclosure object so let's
+        # remove the objects we combined from
+        enclosures.pop(idx)
```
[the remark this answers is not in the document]
yocalebo: Done

src/middlewared/middlewared/plugins/enclosure_/enclosure_class.py
```diff
@@ -48,75 +83,151 @@ def _get_vendor_product_revision_and_encname(self):
...
+                self.controller = False
+            case 'HGST_H4060-J':
+                self.model = JbodModels.ES60G2.value
+                self.controller = False
+            case 'HGST_H4102-J':
+                self.model = JbodModels.ES102.value
+                self.controller = False
+            case 'VikingES_NDS-41022-BB':
+                self.model = JbodModels.ES102G2.value
+                self.controller = False
```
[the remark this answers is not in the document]
yocalebo: I thought about it, but I'll do that in a separate ticket.

bmeagherix (approved):

PR #12492, opened 2023-11-14
NAS-125216 / 24.04 / Switch default home path to /var/empty

anodos325:
PAM session checks require that home directory actually exist, which is causing various CI tests to fail because the PAM check prevents su to the limited user account. /var/empty exists and is immutable, this has the side-benefit of preventing situations where pam_mkhomedir may try to create the directory /nonexistent leading to users sharing same writable home directory.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -31,7 +31,15 @@ from contextlib import suppress
 ADMIN_UID = 950  # When googled, does not conflict with anything
 ADMIN_GID = 950
 SKEL_PATH = '/etc/skel/'
-DEFAULT_HOME_PATH = '/nonexistent'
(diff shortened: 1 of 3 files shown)


yocalebo (approved):

src/middlewared/middlewared/plugins/account.py
```diff
@@ -358,7 +366,7 @@ def validate_homedir_path(self, verrors, schema, data, users):
             verrors.add(f'{schema}.home', '"Home Directory" cannot contain colons (:).')
             return False
 
-        if data['home'] == DEFAULT_HOME_PATH:
+        if data['home'] in DEFAULT_HOME_PATHS:
```
yocalebo, line 369: This looks wrong? `data['home']` has to be a literal `/var/empty` before this condition returns `True`. It's essentially the exact same operation before you made the change `==`

PR #12502, opened 2023-11-15
NAS-125139 / 24.04 / Fix network send/receive rate

Qubad786:
**Problem:**

When reporting network statistics in the `realtime.reporting` event, the rate of change is miscalculated. Currently, the calculation involves dividing the current value by the interval, but it should be the difference between the current and old values at the current time minus the interval. The interval should remain constant.

**Solution:**

To address this, the solution involves querying the current value from Netdata along with the old value, which is from seconds equal to the interval in the past. The difference between these values is then calculated to accurately report the change in network statistics.

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/events.py
+++ b/src/middlewared/middlewared/plugins/reporting/events.py
@@ -86,7 +86,7 @@ class RealtimeEventSource(EventSource):
                     'cpu': get_cpu_stats(netdata_metrics, cores),
                     'disks': get_disk_stats(netdata_metrics, self.middleware.call_sync('device.get_disk_names')),
                     'interfaces': get_interface_stats(
-                        netdata_metrics, [i['name'] for i in self.middleware.call_sync('interface.query')]
+                        netdata_metrics, self.middleware.call_sync('interface.query_names_only')
                     ),
                     'failed_to_connect': False,
                 }
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -175,6 +175,10 @@ class InterfaceService(CRUDService):
         additional_attrs=True,
     )
 
+    @private
+    async def query_names_only(self):
+        return [i['int_interface'] for i in await self.middleware.call('datastore.query', 'network.interfaces')]
+
     @filterable
     def query(self, filters, options):
         """
--- a/src/middlewared/middlewared/plugins/reporting/realtime_reporting/ifstat.py
+++ b/src/middlewared/middlewared/plugins/reporting/realtime_reporting/ifstat.py
@@ -1,8 +1,6 @@
 import collections
 import typing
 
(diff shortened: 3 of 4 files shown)


Qubad786:
@yocalebo pushed a commit

src/middlewared/middlewared/plugins/reporting/events.py
```diff
@@ -64,9 +65,22 @@ def run_sync(self):
         while not self._cancel_sync.is_set():
             # this gathers the most recent metric recorded via netdata (for all charts)
             retries = 2
+            interfaces_names = [i['name'] for i in self.middleware.call_sync('interface.query')]
```
yocalebo, line 68: This is too expensive to call every 0.5 seconds. That endpoint does a TON of other things that just doesn't make sense when all we're doing is getting the interface name.

We should extend `interface.query` with an `{'extra': {'names_only': True}}` argument that short-circuits very early and returns the names of the interfaces instead of doing everything else

yocalebo (changes requested):

ylu19870105:
I found that the 6999 api has already returned a value in units of speed in kilobits/s! Why did you still decide to calculate on your own?

Qubad786:
@ylu19870105 we want to represent in bytes/sec where we do the first conversion and then we want to keep backward compatibility with how we reported stats earlier (https://github.com/truenas/middleware/blob/5af57fe7ff07c2344cf631b64ba02a563b9cb655/src/middlewared/middlewared/plugins/reporting/realtime_reporting/ifstat.py#L16 - this should address your query)

src/middlewared/middlewared/plugins/reporting/events.py
```diff
@@ -86,7 +86,11 @@ def run_sync(self):
                     'cpu': get_cpu_stats(netdata_metrics, cores),
                     'disks': get_disk_stats(netdata_metrics, self.middleware.call_sync('device.get_disk_names')),
                     'interfaces': get_interface_stats(
-                        netdata_metrics, [i['name'] for i in self.middleware.call_sync('interface.query')]
+                        interval, netdata_metrics, [
+                            i['name'] for i in self.middleware.call_sync(
+                                'interface.query', [], {'extra': {'names_only': True}}
```
yocalebo, line 91: Where is this being handled in `interface.query`? Did you forget a commit?

yocalebo (changes requested):

ylu19870105:
> @ylu19870105 we want to represent in bytes/sec where we do the first conversion and then we want to keep backward compatibility with how we reported stats earlier (
> 
> https://github.com/truenas/middleware/blob/5af57fe7ff07c2344cf631b64ba02a563b9cb655/src/middlewared/middlewared/plugins/reporting/realtime_reporting/ifstat.py#L16
> 
> - this should address your query)

Compared to speed, 'received_bytes' and 'sent_bytes' are even less important. No need for backward compatibility.

PR #12503, opened 2023-11-15
NAS-124864 / 24.04 / properly toggle ARC sysctl values

yocalebo:
Upstream ZFS removed the ARC MAX == 50% of available RAM recently. We need to update the VM plugin as well. Instead of halving memory in the VM plugin, we simply store the "default" ARC max value _before_ we initialize any VMs. We use that as a high water mark since duplicating the openzfs logic inside middleware isn't the right approach (since upstream can change this calculation at any time).

### DIFF
--- a/src/middlewared/middlewared/plugins/sysctl/sysctl_info.py
+++ b/src/middlewared/middlewared/plugins/sysctl/sysctl_info.py
@@ -20,9 +20,7 @@ class SysctlService(Service):
         return cp.stdout.decode().split('=')[-1].strip()
 
     def get_default_arc_max(self):
-        # Default value of arc max on systems
-        # https://github.com/truenas/zfs/blob/f38f3a4a3ee430a15fa221ba9fc2b6b8fd17c040/module/os/linux/zfs/arc_os.c#L85
-        return psutil.virtual_memory().total // 2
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/sysctl/sysctl_info.py
```diff
@@ -19,10 +18,21 @@ async def get_value(self, sysctl_name):
             raise CallError(f'Unable to retrieve value of "{sysctl_name}" sysctl : {cp.stderr.decode()}')
         return cp.stdout.decode().split('=')[-1].strip()
 
+    def store_default_arc_max(self):
+        """This method should be called _BEFORE_ we initialize any VMs
+        so that we can capture what the ARC max value was before we start
+        changing the various ARC sysctls based on VM memory configurations"""
+        val = self.get_arcstats()['c_max']
+        with open(DEFAULT_ARC_MAX_FILE, 'w') as f:
+            f.write(str(val))
```
anodos325, line 27: Should we flush to be extra certain?

yocalebo: done

PR #12504, opened 2023-11-15
NAS-124864 / 24.04 / properly toggle ARC sysctls

yocalebo:
Upstream ZFS removed the ARC MAX == 50% of available RAM recently. We need to update the VM plugin as well. Instead of halving memory in the VM plugin, we simply store the "default" ARC max value before we initialize any VMs. We use that as a high water mark since duplicating the openzfs logic inside middleware isn't the right approach (since upstream can change this calculation at any time)

### DIFF
--- a/tests/api2/test_007_early_settings.py
+++ b/tests/api2/test_007_early_settings.py
@@ -1,7 +1,9 @@
 from middlewared.test.integration.utils import call
+from middlewared.plugins.sysctl.sysctl_info import DEFAULT_ARC_MAX_FILE
 
 
 def test_sysctl_arc_max_is_set():
-    """Middleware should have set this value early in boot phase
-    and this should return a number"""
-    assert call('sysctl.get_default_arc_max')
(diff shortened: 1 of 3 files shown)


anodos325:
What do we do about case where user changes arc_max via tunable?

tests/api2/test_007_early_settings.py
```diff
@@ -0,0 +1,7 @@
+from middlewared.test.integration.utils import call
+
+
+def test_sysctl_arc_max_is_set():
+    """Middleware should have set this value early in boot phase
+    and this should return a number"""
+    assert call('sysctl.get_default_arc_max')
```
anodos325, line 7: Won't this write the file if it doesn't exist? We should be able to `filesystem.stat` it to check for existence.

yocalebo: derp, fixed

anodos325 (approved):

PR #12515, opened 2023-11-17
NAS-124930 / 24.04 / Properly validate interface selected with apps

sonicaj:
This PR adds changes to properly validate that the interface specified in apps settings for routing traffic, is actually "UP" at the time it is intended to be used as if the interface link state is down, that is bound to result in a failure when apps are going to try and configure themselves.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
@@ -362,12 +362,13 @@ class KubernetesService(ConfigService):
             i['id']: i.get('state', {}).get('link_state') == 'LINK_STATE_UP'
             for i in await self.middleware.call('interface.query')
         }
+        valid_choices = [i for i in filter(lambda k: interface_states.get(k), interfaces)]
         for k in filter(lambda k: data[k], ('route_v4_interface', 'route_v6_interface')):
             err_str = ''
             if data[k] not in interfaces:
-                err_str = f'Please specify a valid interface (i.e {", ".join(interfaces)!r}).'
+                err_str = f'Please specify a valid interface (i.e {", ".join(valid_choices)!r}).'
             elif not interface_states.get(data[k]):
-                err_str = 'Specified interface is not active'
+                err_str = f'Please specify a valid interface which is active (i.e {", ".join(valid_choices)!r}).'
             if err_str:
                 errors.append((k, data[k], err_str))
         return errors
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/kubernetes_linux/update.py
```diff
@@ -359,10 +358,19 @@ async def validate_data(self, data, schema, old_data):
...
+            for i in await self.middleware.call('interface.query')
+        }
+        valid_choices = [i for i in filter(lambda k: interface_states.get(k), interfaces)]
+        for k in filter(lambda k: data[k], ('route_v4_interface', 'route_v6_interface')):
+            err_str = ''
+            if data[k] not in interfaces:
+                err_str = f'Please specify a valid interface (i.e {", ".join(valid_choices)!r}).'
+            elif not interface_states.get(data[k]):
+                err_str = f'Please specify a valid interface which is active (i.e {", ".join(valid_choices)!r}).'
+            if err_str:
```
yocalebo, line 372: No reason to have this check at the end. Just `errors.append` in each branch inside the loop. We should also return early on the first true evaluation so it removes risk of having verrors that have basically duplicate data. Finally `data[k]` isn't used in the caller so no reason to pass it back i.e.
```
if data[k] not in interfaces:
    errors.append((k, f'Please specify a valid interface (i.e {", ".join(valid_choices)!r}).'))
    break
elif not interfaces_states.get(data[k]):
    errors.append((k, f'Please specify a valid interface which is active (i.e {", ".join(valid_choices)!r}).'))
    break

Comment (sonicaj) [reply src/middlewared/middlewared/plugins/kubernetes_linux/update.py:372]: @yocalebo that is used but in a different place and also we don't want to break, we are looping over basically attributes which we want to validate as they are not related to each other in the sense of duplication and each should be validated separately

Comment (yocalebo) [reply src/middlewared/middlewared/plugins/kubernetes_linux/update.py:372]: The caller is doing this
```
        for k, _, err_s in await self.validate_interfaces(data):
             verrors.add(f'{schema}.{k}', err_s)
```
So how is `data[k]` being used anywhere else except inside that function?

sonicaj: It's being used in another file - when we start k8s basically

yocalebo: so there is another caller, i understand now 😄

yocalebo (approved):

PR #12528, opened 2023-11-19
NAS-125274 / 24.04 / add serial number information to smart alert (if possible)

yocalebo:
The automatic alert that we generate in `smart_alert.py` (called by `smartd`) only gives the block device name (i.e. `/dev/sda`). Those names change frequently so we need to try and append the serial number of the drive. This does that by doing 2 things:
1. fix the `device.get_disk` api to match `device.get_disks`
2. call `device.get_disk` in `smart_alert.py` and add the serial number information to the alert

### DIFF
--- a/src/freenas/usr/local/libexec/smart_alert.py
+++ b/src/freenas/usr/local/libexec/smart_alert.py
@@ -21,7 +21,7 @@ def main():
         dev_name = device.removeprefix("/dev/")
         info = c.call("device.get_disk", dev_name, False, True)
         if info is not None and (serial := info.get(dev_name)):
-            " ".join([device, f"({serial!r})"])
+            device = " ".join([device, f"({serial!r})"])
 
         c.call("alert.oneshot_create", "SMART", {"device": device, "message": message})
 
--- a/src/freenas/usr/local/libexec/smart_alert.py
+++ b/src/freenas/usr/local/libexec/smart_alert.py
@@ -18,6 +18,11 @@ def main():
         return
 
     with Client() as c:
+        dev_name = device.removeprefix("/dev/")
+        info = c.call("device.get_disk", dev_name, False, True)
+        if info is not None and (serial := info.get(dev_name)):
+            " ".join([device, f"({serial!r})"])
+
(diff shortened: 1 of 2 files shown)


src/freenas/usr/local/libexec/smart_alert.py
```diff
@@ -18,6 +18,11 @@ def main():
         return
 
     with Client() as c:
+        dev_name = device.removeprefix("/dev/")
+        info = c.call("device.get_disk", dev_name, False, True)
+        if info is not None and (serial := info.get(dev_name)):
+            " ".join([device, f"({serial!r})"])
```
sonicaj, line 24: I suppose you wanted to add `device = " ".join`

yocalebo: Good catch!

src/middlewared/middlewared/plugins/device_/device_info.py
```diff
@@ -226,14 +226,18 @@ def safe_retrieval(self, prop, key, default, asint=False):
         return default
 
     @private
-    def get_disk(self, name):
+    def get_disk(self, name, get_partitions=False, serial_only=False):
         context = pyudev.Context()
         try:
             block_device = pyudev.Devices.from_name(context, 'block', name)
+            if serial_only:
+                return {name: self.get_disk_serial(block_device)}
```
sonicaj, line 234: Btw if we are only retrieving disk here, should we be adding `name` to the dict at all ?

yocalebo: It kind of follows the returned object if serial_only=False

sonicaj: Sure, my idea was that if it's only 1 disk which this endpoint is going to give data of - probably no need to use that name in the dict at all and just return the information directly or use `{"serial": value}` here

sonicaj (approved):

src/middlewared/middlewared/plugins/device_/device_info.py
yocalebo: That's a good idea. I changed it to return `{'serial': ''}`

PR #12543, opened 2023-11-20
NAS-125048 / 24.04 / Fix NFS bindip configuration at boot.

mgrimesix:
### The problem
The NFS bindip cannot be configured if the selected IP address is not functional.  This setting is in `/etc/rc.conf.freebsd`.  NFS is started and configured _before_ the network is started.  NFS has a post network start event handler that will restrart NFS if bindip is configured in the DB.

However, the NFS bindip _setting_ was blocked at boot during the initial configuration run because the network subsystem had not yet completed startup.  This resulted in no NFS bindip configuration.

### The fix
In the post network start handler, regenerate `rc.conf.freebsd` before restarting NFS.


mgrimesix:
PR against wrong branch

PR #12544, opened 2023-11-20
NAS-125048 / 13.1 / Fix NFS bindip configuration at boot

mgrimesix:
### The problem
The NFS bindip cannot be configured if the selected IP address is not functional.  This setting is in `/etc/rc.conf.freebsd`.  NFS is started and configured _before_ the network is started.  NFS has a post network start event handler that will restrart NFS if bindip is configured in the DB.

However, the NFS bindip _setting_ was blocked at boot during the initial configuration run because the network subsystem had not yet completed startup.  This resulted in no NFS bindip configuration.

### The fix
In the post network start handler, regenerate `rc.conf.freebsd` before restarting NFS.

### DIFF
--- a/src/middlewared/middlewared/plugins/nfs.py
+++ b/src/middlewared/middlewared/plugins/nfs.py
@@ -564,7 +564,6 @@ async def pool_post_import(middleware, pool):
     path = f'/mnt/{pool["name"]}'
     for share in await middleware.call('sharing.nfs.query'):
         if any(filter(lambda x: x == path or x.startswith(f'{path}/'), share['paths'])):
-            await middleware.call('etc.generate', 'rc')
             middleware.create_task(middleware.call('service.reload', 'nfs'))
             break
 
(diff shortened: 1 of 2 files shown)


mgrimesix:
retest this please

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -564,6 +564,7 @@ async def pool_post_import(middleware, pool):
     path = f'/mnt/{pool["name"]}'
     for share in await middleware.call('sharing.nfs.query'):
         if any(filter(lambda x: x == path or x.startswith(f'{path}/'), share['paths'])):
+            await middleware.call('etc.generate', 'rc')
```
anodos325, line 567: Is this the right place to re-generate the rc.conf? Maybe we need to re-generate after bind ip change in NFS plugin?

anodos325: Nevermind, poor reading comprehension. Maybe also review whether it's also needed on nfs.update for bindip.

mgrimesix:
A middleware re-install CI test was not run, that capability is currently not available.

anodos325 (approved):

PR #12553, opened 2023-11-21
add new firmware for ES60G2 to enclosure2

yocalebo:
This adds the new firmware (identified during QE cycle) to the enclosure2 namespace. Was explicitly done in a separate PR so ticket tracking could be kept separate since the original PR was back ported all the way back to 13.0 but enclosure2 only exists in Cobia/DF.


yocalebo:
backport

PR #12564, opened 2023-11-22
NAS-125335 / 24.04 / Add validation for SMB share name

anodos325:
TrueNAS SMB share name validation allowed invalid characters, which
resulted in share that was created being rejected by libsmbconf.

This commit validates that share name minimally meets requirements
from MS-FSCC 2.1.6

```
* A share name MUST be no more than 80 characters in length.

* The following characters are illegal in a share name:
  " \ / [ ] : | < > + = ; , * ?

* Control characters in range 0x00 through 0x1F, inclusive, are illegal in a share name.

* All other Unicode characters are legal.
```

Extended requirements for TrueNAS are as follows:
* the % (percent) character is also illegal
* all unicode control characters are prohibited

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -7,6 +7,7 @@ import re
 from pathlib import Path
 import stat
 import subprocess
+import unicodedata
 import uuid
 
 from copy import deepcopy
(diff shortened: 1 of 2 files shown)


sonicaj (approved):

yocalebo (approved):
Test?

anodos325:
Looks like I need to grab full list from MS-FSCC 2.1.6.

PR #12585, opened 2023-11-28
Fix audit filter in syslog-ng config

anodos325:


sonicaj (approved):

bmeagherix:
Do we have any tests that'd check the audit filter was operating as expected?

anodos325:
> Do we have any tests that'd check the audit filter was operating as expected?

I'll work on these tests in separate PR once this one is merged.

PR #12586, opened 2023-11-28
NAS-125106 / 24.04 / Allow specifying force flag for ACL based questions

Qubad786:
### Context

Applying ACL to a non-empty path with force flag disabled gives a traceback of error message.

### Change

This PR will add the change to handle validation more appropriately and raise a friendly error message for user.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/validation.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/validation.py
@@ -17,6 +17,7 @@ validation_mapping = {
     'validations/nodePort': 'port_available_on_node',
     'validations/hostPath': 'custom_host_path',
     'normalize/ixVolume': 'ix_mount_path',
+    'normalize/acl': 'acl_entries',
     'validations/lockedHostPath': 'locked_host_path',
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/chart_releases_linux/validation.py
```diff
@@ -191,6 +194,14 @@ async def validate_ix_mount_path(self, verrors, value, question, schema_name, re
         elif not validate_dataset_name(path):
             verrors.add(schema_name, f'Invalid dataset name {path}. "test1, ix-test, ix_test" are valid examples.')
 
+    @private
+    async def validate_acl_entries(self, verrors, value, question, schema_name, release_data):
+        try:
+            if value.get('path') and os.listdir(value['path']) and not value.get('options', {}).get('force'):
```
yocalebo, line 200: 2 problems:
1. blocking I/O in a coroutine
2. os.listdir just to see if there are contents is inappropriate

yocalebo (changes requested):

yocalebo (approved):

PR #12592, opened 2023-11-28
Add private endpoint for webui to pre-validate SMB share

anodos325:
This endpoint was requested by webui team to facilitate simpler share configuration and have validation that is more restrictive than our normal validation (in that it requires local SMB user to exist if AD isn't configured).

This PR also slightly refactors SMB share name validation.


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -1657,6 +1650,29 @@ async def validate_cluster_path(self, verrors, name, volname, path):
...
+        invalid_characters = INVALID_SHARE_NAME_CHARACTERS & set(name)
+        if invalid_characters:
+            verrors.add(
+                f'{schema_name}.name',
+                f'Share name contains the following invalid characters: {", ".join(invalid_characters)}'
+            )
+
+        if any(unicodedata.category(char) == 'Cc' for char in name):
+            verrors.add(
+                f'{schema_name}.name', 'Share name contains unicode control characters.'
```
yocalebo, line 1673:
```suggestion
                f'{schema_name}.name', 'Share name cannot contain unicode control characters.'
```

yocalebo (approved):

PR #12606, opened 2023-11-30
NAS-125261 / 24.04 / add tests for enclosure2.query

yocalebo:
This is, arguably, the most important method for the enclosure2 namespace. It's responsible for querying, formatting, and structuring the final data to be returned. This adds a unit test that actually tests this functionality with real world data. I've added 5 different platforms that follow very specific code-paths in the enclosure2 namespace. The idea is, in theory, to stress all parts of this query method. The most important part of this is the fact that `combine_enclosures` is being tested which is of vital importance.

1. add `see_enclosures2.py` and move `get_ses_enclosures` there to maintain uniformity with how we're doing in the legacy code base
2. don't call `map_nvme` directly in `enclosure2.query`, instead call `self.map_nvme()`
3. fix a bug that I spotted in `enclosure2.set_slot_status`
4. add `test_enclosure2_query.py` which adds functionality so that the various hardware platforms can be tested
5. add `test-cases` for 5 different hardware platforms that we sell


src/middlewared/middlewared/plugins/enclosure_/ses_enclosures2.py
```diff
@@ -0,0 +1,27 @@
...
+def get_ses_enclosure_status(bsg_path):
+    try:
+        return EnclosureDevice(bsg_path).status()
+    except OSError:
+        logger.error('Error querying enclosure status for %r', bsg_path, exc_info=True)
+
+
+def get_ses_enclosures(dmi):
+    rv = list()
+    with suppress(FileNotFoundError):
```
anodos325, line 20: Should we be suppressing FileNotFoundError here or below where we get the enclosure status? Do we want to break iteration early?

themylogin (approved):

anodos325 (approved):

PR #12614, opened 2023-12-01
NAS-125487 / 24.10 / Use stderr from zectl failures for CallError

anodos325:
The zectl command on failure conditions will print to stderr rather than stdout. This means we were losing information about errors on bootenv operations.


bmeagherix (approved):

sonicaj (approved):

src/middlewared/middlewared/plugins/bootenv.py
```diff
@@ -165,7 +165,7 @@ def activate(self, oid):
         try:
             subprocess.run([self.BE_TOOL, 'activate', oid], capture_output=True, text=True, check=True)
```
yocalebo, line 166: We should do `stdout=subprocess.PIPE, stderr=subprocess.STDOUT` and check the return code because IIRC, zectl can dump errors to stdout or stderr.

PR #12618, opened 2023-12-01
fix Email schema class

yocalebo:
Who would have thought how miserable email address validation could be....I sure didn't until I started doing research. Instead of trying to validate all conceivable incantations of the various RFC's wrt to email addresses, we take the other approach of being less strict so we don't have a false-negative (i.e. preventing a valid email address) which is what our current code does.


themylogin (approved):
The best regex for an email address is just .+@.+

PR #12627, opened 2023-12-01
NAS-125533 / 24.04 / Relax authorization requirements

anodos325:
* Allow all authorized users to gather some basic hardware info.
* Allow all authenticated users to see their own 2fa config via auth.me.


sonicaj (approved):

src/middlewared/middlewared/plugins/system/info.py
```diff
@@ -130,6 +130,7 @@ async def build_time(self, app):
         return datetime.fromtimestamp(int(buildtime)) if buildtime else buildtime
 
     @accepts()
+    @no_authz_required
```
themylogin, line 133: Is this all really needed? Things like loadavg/uptime can be seen as sort of private.

themylogin (approved):

PR #12645, opened 2023-12-05
NAS-125562 / 24.04 / Fix more log spam (network activity feature)

yocalebo:
This entire feature wasn't designed/thought about very well. It's flawed in many aspects but without the changes in this PR we're spamming our log files when someone does `Deny All` (even though that's not actually happening) in the webUI.

This squashes the most obvious log spam use-cases. (This was found on an internal HA system for which I'm troubleshooting an unrelated issue)

### DIFF
--- a/src/middlewared/middlewared/plugins/acme_protocol.py
+++ b/src/middlewared/middlewared/plugins/acme_protocol.py
@@ -115,10 +115,7 @@ class ACMERegistrationService(CRUDService):
         # 3) SAVE REGISTRATION OBJECT
         # 4) SAVE REGISTRATION BODY
 
-        self.middleware.call_sync('network.general.will_perform_activity', 'acme')
-
         verrors = ValidationErrors()
-
         directory = self.get_directory(data['acme_directory_uri'])
         if not isinstance(directory, messages.Directory):
             verrors.add(
--- a/src/middlewared/middlewared/plugins/mail.py
+++ b/src/middlewared/middlewared/plugins/mail.py
@@ -1,7 +1,7 @@
 from middlewared.schema import accepts, Bool, Dict, Int, List, Password, Patch, Ref, returns, Str
 from middlewared.service import CallError, ConfigService, ValidationErrors, job, periodic, private
 import middlewared.sqlalchemy as sa
-from middlewared.utils import osc, BRAND
+from middlewared.utils import BRAND
 from middlewared.utils.mako import get_template
 from middlewared.validators import Email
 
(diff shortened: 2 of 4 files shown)


src/middlewared/middlewared/plugins/acme_protocol.py
```diff
@@ -115,10 +115,7 @@ def do_create(self, data):
         # 3) SAVE REGISTRATION OBJECT
         # 4) SAVE REGISTRATION BODY
 
-        self.middleware.call_sync('network.general.will_perform_activity', 'acme')
```
themylogin, line 118: Why?

src/middlewared/middlewared/plugins/mail.py
```diff
@@ -24,6 +24,10 @@
 import syslog
 
 
+class DenyNetworkActivity(Exception):
```
themylogin, line 27: Why here and who raises this exception?

src/middlewared/middlewared/plugins/mail.py
```diff
@@ -455,8 +464,12 @@ def send_mail_queue(self):
                                         queue.message['To'].split(', '),
                                         queue.message.as_string())
                         server.quit()
-                except Exception:
-                    self.logger.debug('Sending message from queue failed', exc_info=True)
+                except Exception as e:
+                    if isinstance(e, DenyNetworkActivity):
+                        self.logger.warning('Sending message from queue denied')
```
themylogin, line 469: Should we stop trying in this case?

src/middlewared/middlewared/plugins/truecommand/connection.py
```diff
@@ -9,7 +9,11 @@ class TruecommandAPIMixin:
     PORTAL_URI = 'https://portal.ixsystems.com/api'
 
     async def _post_call(self, options=None, payload=None):
-        await self.middleware.call('network.general.will_perform_activity', 'truecommand')
+        try:
+            await self.middleware.call('network.general.will_perform_activity', 'truecommand')
+        except Exception:
+            return {'error': 'Network activity denied for TrueCommand service'}
```
themylogin, line 15: We can use `can_perform_activity` instead of `try`/`catch` block

src/middlewared/middlewared/plugins/acme_protocol.py
yocalebo: Because `self.get_directory()` makes the exact same call so this was unnecessary.

sonicaj (approved):

PR #12663, opened 2023-12-06
NAS-125586 / 24.04 / Allow authenticated users nenew own 2fa secret

anodos325:
Authenticated users should always be able to renew their own 2fa secret.


src/middlewared/middlewared/utils/privilege.py
```diff
@@ -0,0 +1,57 @@
...
+    if credential is None:
+        return False
+
+    elif credential_has_full_admin(credential):
+        return True
+
+    return credential.user['username'] == username
+
+
+def app_username_check(app, username):
```
themylogin, line 18: Can we have a more descriptive name? Like `can_modify_user`

src/middlewared/middlewared/utils/privilege.py
```diff
@@ -0,0 +1,57 @@
...
+    * credential username matches `username` passed into this method
+    """
+    if app is None:
+        return True
+
+    return credential_username_check(app.authenticated_credentials, username)
+
+
+def privileges_group_mapping(
+    privileges: list,
```
sonicaj, line 38: We are partially doing type hinting, can we perhaps do it on the above methods as well in the same file ?

themylogin (approved):

PR #12666, opened 2023-12-06
NAS-125593 / 24.04 / Make SHARING_MANAGER role a superset of READONLY

anodos325:
Per request from webui team, grant the SHARING_MANAGER role all rights that READONLY has.


themylogin (commented):
Seems a pretty serious design change from the original idea...

anodos325:
> Seems a pretty serious design change from the original idea...

This design change was made pursuant to higher-level discussion regarding concept of how the roles would be used.

yocalebo (approved):

themylogin (approved):

PR #12667, opened 2023-12-06
Improve resilience of mountd syslog filter test

mgrimesix:
Syslog can be slow to flush.
Added simple retries to get the mountd messages when they are supposed to exist.
Added a 2 second sleep before reading from the log when messages are _not_ supposed to exist.


bmeagherix (approved):
I'll approve despite some possible tweaks.  Your call if you want to address.

PR #12670, opened 2023-12-07
NAS-125618 / 24.04 / Allow users to set own passwords

anodos325:
This commit adds a new API endpoint user.set_password that may be used to change the password of a given user.

NOTE: when authenticated session has less than FULL_ADMIN role, password changes will be rejected if the payload does not match the currently-authenticated user.

API keys granting access to this endpoint will be able to reset password of any user.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -1484,7 +1484,10 @@ class UserService(CRUDService):
 
         verrors = ValidationErrors()
         is_full_admin = credential_has_full_admin(app.authenticated_credentials)
-        authenticated_user = app.authenticated_credentials.user['username']
+        authenticated_user = None
+
+        if app.authenticated_credentials.is_user_session:
+            authenticated_user = app.authenticated_credentials.user['username']
 
         username = data['username']
         password = data['new_password']
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -1,10 +1,13 @@
 from middlewared.schema import accepts, Bool, Dict, Int, List, Password, Patch, returns, Str, LocalUsername
 from middlewared.service import (
-    CallError, CRUDService, ValidationErrors, no_auth_required, pass_app, private, filterable, job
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/account.py
```diff
@@ -1445,6 +1448,132 @@ def update_sshpubkey(self, homedir, user, group):
...
+        if entry['locked']:
+            verrors.add(
+                'user.reset_password.username',
+                f'{username}: user account is locked.'
+            )
+
+        new_hash = crypted_password(password)
+        verrors.check()
+
+        if entry['smb']:
```
themylogin, line 1556: Code repetition for `__set_password`

src/middlewared/middlewared/plugins/account.py
```diff
@@ -1445,6 +1448,132 @@ def update_sshpubkey(self, homedir, user, group):
...
+        Str('username', required=True),
+        Password('old_password', default=''),
+        Password('new_password', required=True),
+        Dict(
+            'options',
+            Bool('skip_password_check', default=False)
+        )
+    ))
+    @pass_app(require=True)
+    async def password_reset(self, app, data):
```
themylogin, line 1463: Why `reset` if we do `set`?

src/middlewared/middlewared/plugins/account.py
```diff
@@ -1445,6 +1448,132 @@ def update_sshpubkey(self, homedir, user, group):
...
 
+    @no_authz_required
+    @accepts(Dict(
+        'password_reset_data',
+        Str('username', required=True),
+        Password('old_password', default=''),
+        Password('new_password', required=True),
+        Dict(
+            'options',
+            Bool('skip_password_check', default=False)
```
themylogin, line 1459: We can have a simpler API interface: make `old_password` not required on schema level, but check that it is set if we are not a full admin. Then this flag won't be necessary.

src/middlewared/middlewared/plugins/account.py
anodos325: I'm pretty bad at naming things.

PR #12676, opened 2023-12-08
Add SHARING_MANAGER to builtin privilege enum

anodos325:
This fixes an oversight during intial work to add new builtin roles / privileges.


themylogin (changes requested):
We also need a database migration that inserts the default undeletable privilege to the table (same as `middlewared/alembic/versions/24.04/2023-11-03_11-35_sharing_manager_privilege.py`)

PR #12682, opened 2023-12-08
NAS-125631 / 24.10 / Disable mocking support in stable releases

anodos325:
Development and testing tools should be disabled for published releases.


themylogin (changes requested):
This will prevent us from running integration tests on stable releases, what is the purpose of this?

PR #12684, opened 2023-12-08
NAS-125573 / 24.04 / Add webui_access flag to composed privilege

anodos325:
This adds flag to composed privilege (for example in auth.me output) that can be used by the webui team to determine whether and authenticated session should have webui access. Current algorithm is to check whether credential has one of non-builtin roles (READONLY, FULL_ADMIN, SHARING_MANAGER).


anodos325:
This provides hint to webui team that they should deny web access (in terms of limiting scope of oddball configurations that they have to support).

sonicaj (approved):

PR #12685, opened 2023-12-08
NAS-125501 / 24.04 / Manage number of nfsd and mountd

mgrimesix:
Previously the default number of nfsd was 16.    We also started an equal number of mountd.  This benefitted systems with many cores and resources, but potentially penalized systems with fewer resources.  The server administrator could change the number of nfsd via the UI and the number of mountd would follow, but most customers use the default setting.

This PR addresses that situation by introducing a platform sensitive calculation to the number of nfsd and reduces the number of mountd to 1/4 the number of nfsd.

After this PR, the default setting for the number of nfsd will be `null`, an unset value.  This will indicate we are to optimally determine the number of nfsd based on the capabilities of the host system.  Currently, the criteria is the number of cores.

The user may override the automatic generation via the UI by selecting the desired number of nfsd threads.   The user may return to the automatic generation by clearing the field.

Additionally, all systems that upgrade and are using the previous default setting of 16 nfsd will be converted to the new platform sensitive calculation method.


src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -84,13 +85,27 @@ async def nfs_extend(self, nfs):
...
         nfs["keytab_has_nfs_spn"] = keytab_has_nfs
+
+        # 'None' indicates we are to dynamically manage the number of nfsd
+        if nfs['servers'] is None:
+            nfs['managed_nfsd'] = True
+            cpu_info = await self.middleware.call("system.cpu_info")
+
+            # Default calculation:
+            #     Number of nfsd == number of cores, but not zero or greater than 32
+            nfs['servers'] = min(max(cpu_info['core_count'], 1), 32)
```
sonicaj, line 96: If we have `servers` set to `null` and you update any other attribute in the NFS configuration, you will set `servers` value automatically which won't be desired behavior as user is not setting it

mgrimesix: Good catch!  You are correct!

sonicaj (approved):
Thanks!

mgrimesix:
Added more CI tests to confirm proper behavior.

anodos325 (approved):

PR #12692, opened 2023-12-09
NAS-124988 / 24.04 / fix faulty HA logic

yocalebo:
This fixes 2 primary issues with HA logic on SCALE.

1. receiving a rapid succession of failover events
2. BACKUP failover loop

This fixes both situations by adding a pub/sub model to vrrp_events.py and checking the timeframe for when we receive VRRP events. The reason why this needs to be fixed is that without these changes, when both controllers are boot up simultaneously (for example), we don't handle the events properly and leave the system with the zpool imported without fenced running. That's a worst case scenario since fenced guards us from having the zpool imported on both controllers simultaneously.

Have tested this myself on 2 different internal systems extensively.

### DIFF
--- a/src/middlewared/middlewared/plugins/device_/vrrp_events.py
+++ b/src/middlewared/middlewared/plugins/device_/vrrp_events.py
@@ -17,7 +17,7 @@ LOGGER = getLogger('failover')  # so logs show up in /var/log/failover.log
 class VrrpObjs:
     fifo_thread = None
     event_thread = None
-    event_queue = None
+    event_queue = deque(maxlen=1)
 
 
 class VrrpThreadService(Service):
@@ -40,10 +40,10 @@ class VrrpThreadService(Service):
 
 class VrrpEventThread(Thread):
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, **kwargs):
         super(VrrpEventThread, self).__init__()
         self.middleware = kwargs.get('middleware')
-        self.event_queue = kwargs.get('queue')
+        self.event_queue = VrrpObjs.event_queue
         self.shutdown_event = Event()
         self.pause_event = Event()
         self.grace_period = 0.5
@@ -201,7 +201,7 @@ class VrrpFifoThread(Thread):
         self._vrrp_file = '/var/run/vrrpd.fifo'
         self.pause_event = Event()
         self.middleware = kwargs.get('middleware')
-        self.event_queue = kwargs.get('queue')
+        self.event_queue = VrrpObjs.event_queue
         self.shutdown_line = '--SHUTDOWN--'
 
     def shutdown(self):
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/device_/vrrp_events.py
```diff
@@ -1,10 +1,196 @@
...
+                self.event_queue.clear()
+                sleep(0.2)
+
+            try:
+                event = self.event_queue[-1]
+                this_event = self.format_fifo_msg(event)
+            except IndexError:
+                # loop is started but we've received no events
+                sleep(0.2)
+                continue
```
anodos325, line 119: Not a requirement for this PR, but you could probably avoid this part of loop by doing a blocking poll (with timeout) on an event_fd, and write to it from VrrpFifoThread.

src/middlewared/middlewared/plugins/device_/vrrp_events.py
```diff
@@ -1,10 +1,196 @@
...
+                # so we will pause this thread while we become
+                # the backup controller and then unpause after
+                last_event = None
+                self.event_queue.clear()
+                sleep(0.2)
+
+            try:
+                event = self.event_queue[-1]
+                this_event = self.format_fifo_msg(event)
+            except IndexError:
```
anodos325, line 116: For future, you can probably create a generic "wakeup" Event() and block loop with `wait(timeout=self.max_wait)` to avoid having to sleep. Then have it wake up either via shutdown() or being set from the VrrpFifoThread.

src/middlewared/middlewared/plugins/device_/vrrp_events.py
```diff
@@ -58,28 +255,47 @@ def run(self):
...
+            VrrpObjs.event_queue = None
+    else:
         # if this is a system that is being licensed for HA for the
         # first time (without being rebooted) then we need to make
-        # sure we start this.
-        VRRP_THREAD = VrrpFifoThread(middleware=middleware)
-        VRRP_THREAD.start()
+        # sure we start these threads
+        if VrrpObjs.event_queue is None:
+            VrrpObjs.event_queue = deque(maxlen=1)
```
anodos325, line 283: In future consider increasing deque size so that we can keep internal history of past events to better detect flapping state.

src/middlewared/middlewared/plugins/device_/vrrp_events.py
```diff
@@ -13,13 +199,20 @@ def __init__(self, *args, **kwargs):
         super(VrrpFifoThread, self).__init__()
         self._retry_timeout = 2  # timeout in seconds before retrying to connect to FIFO
         self._vrrp_file = '/var/run/vrrpd.fifo'
+        self.pause_event = Event()
         self.middleware = kwargs.get('middleware')
-        self.logger = self.middleware.logger
-        self.shutdown_line = '--SHUTDOWN--\n'
+        self.event_queue = kwargs.get('queue')
```
anodos325, line 204: What is impact of `self.event_queue = None` (e.g. not in kwargs)?

anodos325: If it's mandatory, then we should have it be mandatory kwarg.

yocalebo: agree, fixed

src/middlewared/middlewared/plugins/device_/vrrp_events.py
```diff
@@ -1,10 +1,196 @@
...
+        super(VrrpEventThread, self).__init__()
+        self.middleware = kwargs.get('middleware')
+        self.event_queue = VrrpObjs.event_queue
+        self.shutdown_event = Event()
+        self.pause_event = Event()
+        self.grace_period = 0.5
+        self.user_provided_timeout = kwargs.get('timeout') or 2
+        self.max_wait = self.user_provided_timeout + self.grace_period
+        self.settle_time = (self.max_wait / 2) + self.grace_period
+        self.max_rapid_settle_time = 5
```
anodos325, line 53: It would probably be good at some point to have a file containing failover-related constants.

anodos325 (approved):

PR #12696, opened 2023-12-11
Remove `test_exclude_recycle_bin` as it is not possible to run this t…

themylogin:
…est anymore without local S3 service


themylogin:
time 0:05

yocalebo (approved):

PR #12699, opened 2023-12-11
NAS-124692 / 24.04 / Changes to bug reporting API

themylogin:
* Remove `category` (for non-enterprise systems) and `type` for `support.new_ticket`
* Allow adding extra files to debug via `debug_extra` option to `support.new_ticket`
* Allow retrieving similar issues via `support.similar_issues`


src/middlewared/middlewared/plugins/system/debug.py
```diff
@@ -62,16 +63,21 @@ def progress_callback(percent, desc):
         except Exception as e:
             raise CallError(f'Failed to generate debug: {e!r}')
 
-    @accepts()
+    @accepts(
+        Dict(
+            'debug_options',
+            Dict('extra', additional_attrs=True)
```
anodos325, line 69: Can we define what the supported debug options are?

yocalebo, line 69: Don't like this option because of how it's implemented. This allows arbitrary data (unbounded in size) to be written to an arbitrary path inside the debug archive (so it could overwrite existing files).

1. We should add a top-level directory to `ixdiagnose` (i.e. another Plugin) that gets added by default (maybe like `WebUIDir` or something). It should capture all data in there by default. (glob pattern)
2. The top-level directory name should be statically defined in middleware repo or in ixdiagnose repo and then imported accordingly
3. The api args that we accept should be a file name, and then a Pipe object
4. We need to truncate the contents of the file to be written to 50KB (Per request of webUI)

yocalebo (changes requested):

src/middlewared/middlewared/plugins/system/debug.py
themylogin: > so it could overwrite existing files

It can't (ixdiagnose won't do it)

> It should capture all data in there by default. (glob pattern)

I don't understand what does this mean in this context, what's "all data"?

themylogin: Done

yocalebo: Because the webUI team is not the only consumer of this endpoint. If we still want to add this feature then we need to restrict the starting component of the path to `plugins/`.

src/middlewared/middlewared/plugins/support.py
```diff
@@ -172,10 +169,9 @@ async def fetch_categories(self, token):
         'new_ticket',
         Str('title', required=True, max_length=None),
         Str('body', required=True, max_length=None),
-        Str('category', required=True),
```
denysbutenko, line 175: The enterprise systems still use the `category` field to submit new tickets. It may be better to make it optional.

src/middlewared/middlewared/plugins/support.py
```diff
@@ -172,10 +169,9 @@ async def fetch_categories(self, token):
...
         Bool('attach_debug', default=False),
+        Dict('debug_extra', additional_attrs=True),
         Password('token'),
-        Str('type', enum=['BUG', 'FEATURE']),
```
denysbutenko, line 178: Why is it removed? We use this field to allow users to choose between Bug or Feature they want to create:

image

src/middlewared/middlewared/plugins/system/debug.py
themylogin: @yocalebo I don't understand what part of my message are you replying to. So, to conclude, what needs to be done now?

src/middlewared/middlewared/plugins/support.py
themylogin: @yocalebo we remove Bug/Feature distinction, don't we?

themylogin:
@yocalebo sorry, any news here?

themylogin:
time 0:30

src/middlewared/middlewared/plugins/support.py
```diff
@@ -169,14 +169,36 @@ async def fetch_categories(self, token):
...
+        data = await post(
+            f'https://{ADDRESS}/freenas/api/v1.0/similar_issues',
+            data=json.dumps({
+                'token': token,
+                'query': query,
+            }),
+        )
+
+        if 'error' in data:
+            raise CallError(data['message'], errno.EINVAL)
```
yocalebo, line 191: Should we raise `ValidationError` here so it can be pretty printed in the UI?

src/middlewared/middlewared/plugins/support.py
```diff
@@ -169,14 +169,36 @@ async def fetch_categories(self, token):
...
 
+    @accepts(Password('token'), Str('query'), roles=['SUPPORT_READ'])
+    @returns(List('similar_issues', items=[Dict(
+        'similar_issue',
+        Str('url'),
+        Str('summary'),
+        additional_attrs=True,
+    )]))
+    async def similar_issues(self, token, query):
+        await self.middleware.call('network.general.will_perform_activity', 'support')
```
yocalebo, line 180: Same here as well, should probably raise a ValidationError if network activity is blocked.

src/middlewared/middlewared/plugins/support.py
themylogin: Why can't `CallError` be pretty-printed?

yocalebo (approved):

PR #12700, opened 2023-12-11
NAS-125673 / 24.04 / Configure roles for service plugin

anodos325:
This commit adds SERVICE_READ and SERVICE_WRITE roles that are allowed to read / write services table configuration and administer services. SERVICE_READ is able to determine service running state. If user has SHARING_NFS_WRITE, SHARING_SMB_WRITE, and SHARING_ISCSI write they will be able to start, stop, restart, etc the respective service. If user has SERVICE_WRITE, then they will be able to manipulate running state of all services.

### DIFF
--- a/src/middlewared/middlewared/plugins/service_/utils.py
+++ b/src/middlewared/middlewared/plugins/service_/utils.py
@@ -15,7 +15,7 @@ def app_has_write_privilege_for_service(
         # Internal middleware call
         return True
 
-    if not app.authenticated_credentials is None:
+    if app.authenticated_credentials is None:
         return False
 
     if credential_has_full_admin(app.authenticated_credentials):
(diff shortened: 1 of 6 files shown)


src/middlewared/middlewared/plugins/service_/utils.py
```diff
@@ -0,0 +1,28 @@
...
+    CIFS = 'SHARING_SMB_WRITE'
+    NFS = 'SHARING_NFS_WRITE'
+    ISCSITARGET = 'SHARING_ISCSI_WRITE'
+
+
+def app_has_write_privilege_for_service(
+    app: object | None,
+    service: str
+) -> bool:
+    if app_credential_full_admin_or_user(app, None):
```
yocalebo, line 14: It's not immediately obvious to me what happens when the `username` (2nd arg to this function) is `None`.

anodos325: It just isn't evaluated. I refactored to use `credential_has_full_admin`

yocalebo (approved):

PR #12702, opened 2023-12-12
Do not scale workloads which are using hostpaths which don't exist

sonicaj:
This PR adds changes to not scale workloads which have host paths which don't exist as in this case k3s automatically creates those paths which can be problematic as if a user tries to create a dataset which has the same mountpoint as the path, dataset creation will fail as we make sure that the dataset's mountpoint does not exist when we attempt to create it.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/scale_workload.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/scale_workload.py
@@ -1,5 +1,6 @@
 import asyncio
 import errno
+import os
 
 from collections import defaultdict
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/chart_releases_linux/scale_workload.py
```diff
@@ -67,6 +68,13 @@ async def scale(self, job, release_name, options):
                     f'{release_name!r} cannot be started as it is consuming following host path(s) '
                     f'which are locked: {", ".join(release["resources"]["locked_host_paths"])}'
                 )
+            elif non_existent_host_paths := [path for path in release['resources']['host_path_volumes'] if not (
```
yocalebo, line 71: It's not really a big deal right now since most apps are only going to have 1 or 2 host paths configured for an app. However, spinning up background threads for every path seems a little much. Would be better, imo, to have a synchronous function that does `os.path.exists` calls and only 1 thread get used. Doesn't need to be fixed in this PR but something to consider in future.

yocalebo (approved):

PR #12703, opened 2023-12-12
NAS-125600 / 24.04 / Mark chart release logs test as flaky

sonicaj:
This commit makes 2 changes, first it adds changes to use an image which is lightweight and can be pulled relatively quickly so that we don't timeout and secondly it marks the test as flaky as sometimes on the CI the test can take some time to complete because of the I/O invovled.


themylogin:
Why not just  wait more?

yocalebo (approved):

themylogin (approved):

PR #12704, opened 2023-12-12
NAS-125636 / 24.04 / Restrict config reset to root / admin account

anodos325:
Configuration reset is an incredibly destructive action that is inappropriate to expose to more than the root account.

### DIFF
--- a/src/middlewared/middlewared/plugins/config.py
+++ b/src/middlewared/middlewared/plugins/config.py
@@ -246,6 +246,13 @@ class ConfigService(Service):
         If `reboot` is true this job will reboot the system after its completed with a delay of 10
         seconds.
         """
+        job.set_progress(0, 'Performing credential check')
+        if job.credentials is None:
+            raise CallError('Unable to check credentials')
+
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/config.py
```diff
@@ -241,11 +242,29 @@ def validate_alembic(self, alembic_version):
...
+        either root or admin accounts that were set up during initial install.
 
         If `reboot` is true this job will reboot the system after its completed with a delay of 10
         seconds.
+
+        NOTE: In case of clustered TrueNAS this may unrecoverably destroy cluster-wide
+        configuration.
         """
+        job.set_progress(0, 'Performing credential check')
+        if job.credentials is not None and job.credentials.is_user_session:
```
yocalebo, line 255: What does it mean if `job.credentials is None` ?

anodos325: Internal job. I suppose here it would be rather unexpected since we shouldn't have anyone writing a method that calls into this.

yocalebo:
Looks like conflicts now.

yocalebo (approved):

PR #12709, opened 2023-12-12
NAS-123832 / 24.04 / Setup initiator and configure interface to use NVME over RoCE

bmeagherix:
This PR adds an API to allow ES24N NVMe/RDMA expansion units (JBOFs) to be attached to SCALE, either single node or HA-pair.

The currently supported configuration is for the unit to be **directly-attached** to the host RDMA-capable interfaces (i.e. no switches).

When a JBOF is added by API, the IP address and credentials of the redfish interface are supplied.  The JBOF will - internal to SCALE - be assigned a unique (and sequential from 0) `jbof_index`.  This index will then be used to assign static IP addresses to all of the dataplane interfaces on the JBOF - accomplished using the redfish interface.  These addresses are fixed & deterministic (`169.254.{20 + shelf_index}.(interface_index <<2)+1`)

Next, all available host RDMA interfaces will be stepped thru the various possible corresponding initiator IP addresses & connectivity checked.  This is done so that we can be tolerant wrt wiring.  Saved assignments are stored in another table (corresponding to `rdma.interface`).  This table is HA aware (i.e. has a `node` column) as the IPs will **NOT** be mirrored on both controllers.

----
In order to best support Engineering testing of the hardware, we currently have the following caveats:
- Allow **any** wiring, so long as each node has *at least one* dataplane path to the expansion.
- Makes no changes wrt NVMe multipath (i.e. currently this is disabled, but we may wish to either enable it or change the above to only allow one dataplane path).
- Provide, but do not call a method to restart the config following a reboot (I believe a ticket already exists that'll handle this).


src/middlewared/middlewared/plugins/interface/rdma.py
```diff
@@ -0,0 +1,114 @@
...
+            result.append({'rdma': link['ifname'], 'netdev': link['netdev']})
+        return result
+
+    @accepts()
+    @returns(List(items=[Dict(
+        'rdma_card_config',
+        Str('serial'),
+        Str('product'),
+        Str('part_number'),
+        List('links', items=[Dict('rdma_link_config')])
```
anodos325, line 70: Are you wanting to `items=[Ref('rdma_link_config')]`?

bmeagherix: Yes, thanks.

anodos325:
Are there any alerts that would be beneficial to add?

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,219 @@
...
+        'jbof_entry',
+        Int('id', required=True),
+
+        Str('description'),
+
+        # Redfish
+        IPAddr('mgmt_ip1', required=True),
+        IPAddr('mgmt_ip2', required=False),
+        Str('mgmt_username', required=True),
+        Password('mgmt_password', required=True),
```
anodos325, line 46: Not saying that this is a problem, just noting for reference that you will need to possibly think through behavior of when user restores server configuration from backup without the 'secret seed'. In this case mgmt_password will be empty string.

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,434 @@
...
+        """Synchronous function to ensure we have a redfish client in cache."""
+        try:
+            RedfishClient.cache_get(self.middleware, mgmt_ip)
+        except KeyError:
+            # This could take a while to login, etc ... hence synchronous wrapper.
+            redfish = RedfishClient(f'https://{mgmt_ip}', username, password)
+            RedfishClient.cache_set(mgmt_ip, redfish)
+
+    @private
+    async def licensed(self):
```
anodos325, line 224: Let's include this in either `system.info` output or some other appropriate public API endpoint for webui consumption.

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,434 @@
...
+    jbof_mgmt_password = sa.Column(sa.EncryptedText())
+
+
+class JBOFService(CRUDService):
+
+    class Config:
+        service = 'jbof'
+        datastore = 'storage.jbof'
+        datastore_prefix = "jbof_"
+        cli_private = True
```
anodos325, line 36: Let's assign this an appropriate `role_prefix` assuming that we want to allow delegating JBOF management to particular privileged groups.

src/middlewared/middlewared/plugins/jbof/redfish/target.py
```diff
@@ -0,0 +1,110 @@
...
+REDFISH_ROOT_PATH = '/redfish/v1'
+
+
+class JBOFRedfishService(Service):
+
+    class Config:
+        namespace = 'jbof.redfish'
+        private = True
+
+    @private
```
anodos325, line 17: Class is already private, we don't need the private decorators.

src/middlewared/middlewared/plugins/rdma/interface/crud.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+class RDMAInterfaceService(CRUDService):
+
+    class Config:
+        namespace = 'rdma.interface'
+        private = True
+        cli_private = True
+        datastore = 'rdma.interface'
+        datastore_prefix = "rdmaif_"
```
anodos325, line 42: You might be able to re-use the `NETWORK_INTERFACE_` role prefix here (assuming that's appropriate).

src/middlewared/middlewared/plugins/jbof/crud.py
bmeagherix: Thanks, added `JBOF` role.

src/middlewared/middlewared/plugins/rdma/interface/crud.py
bmeagherix: Currently this API is not public, so I believe that means a role is not appropriate.

src/middlewared/middlewared/plugins/jbof/redfish/target.py
bmeagherix: Thanks, removed.

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,548 @@
...
+        role_prefix = 'JBOF'
+
+    ENTRY = Dict(
+        'jbof_entry',
+        Int('id', required=True),
+
+        Str('description'),
+
+        # Redfish
+        IPAddr('mgmt_ip1', required=True),
```
anodos325, line 56: Are there any address types we want to exclude via `excluded_address_types=`?

bmeagherix: Debatable.  _Ideally_ we'd have a statically assigned IP address, but IMO if the shelf were configured with DHCP then a hostname tied to that would be preferable.

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,546 @@
...
+        Str('mgmt_username', required=True),
+        Password('mgmt_password', required=True),
+    )
+
+    @private
+    async def add_index(self, data):
+        """Add a private unique index (0-255) to the entry if not already present."""
+        if 'index' not in data:
+            index = await self.middleware.call('jbof.next_index')
+            if index is not None:
```
anodos325, line 67: Would an index of None be considered an error condition?

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,546 @@
...
+    async def hardwire_dataplane(self, mgmt_ip, shelf_index, schema, verrors):
+        """Hardware the dataplane interfaces of the specified JBOF.
+
+        Configure the data plane network interfaces on the JBOF to
+        previously determined subnets.
+
+        Then attempt to connect using all the available RDMA capable
+        interfaces.
+        """
+        # shelf_interfaces = await self.middleware.call('jbof.fabric_interface_choices', mgmt_ip)
```
anodos325, line 277: Remove?

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,546 @@
...
+        Update JBOF of `id`
+        """
+        old = await self.get_instance(id_)
+        new = old.copy()
+        new.update(data)
+        verrors, data = await self.validate(new, 'jbof_update', old)
+        verrors.check()
+
+        if old['uuid'] != new['uuid']:
+            self.middleware.logger.debug('Changed UUID of JBOF')
```
anodos325, line 192: Would it be useful to log the old / new UUIDs?

src/middlewared/middlewared/plugins/jbof/redfish/client.py
```diff
@@ -0,0 +1,300 @@
...
+        r = self.get(uri)
+        if r.ok:
+            self.cache[uri] = r.json()
+            return self.cache[uri]
+
+    def get_root_object(self):
+        try:
+            return self.get(self.prefix).json()
+        except Exception as excp:
+            raise excp
```
anodos325, line 163: Did we want to do something with this exception?

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,546 @@
...
+                count = await self.middleware.call('jbof.query', [], {'count': True})
+                if count >= license_count:
+                    verrors.add(f"{schema_name}.mgmt_ip1", f"Already configured the number of licensed emclosures: {license_count}")
+
+        # Ensure redfish connects to mgmt1 (incl login)
+        mgmt_ip1 = data.get('mgmt_ip1')
+        if not RedfishClient.is_redfish(mgmt_ip1):
+            verrors.add(f"{schema_name}.mgmt_ip1", "Not a redfish management interface")
+        else:
+            redfish1 = RedfishClient(f'https://{mgmt_ip1}')
```
anodos325, line 91: Is there any chance of having `is_redfish`, but subsequent request to get redfish root failing here?

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,546 @@
...
+        return macs
+
+    @private
+    def hardwire_shelf(self, mgmt_ip, shelf_index):
+        redfish = RedfishClient.cache_get(self.middleware, mgmt_ip)
+        shelf_interfaces = redfish.fabric_ethernet_interfaces()
+        for (eth_index, uri) in enumerate(shelf_interfaces):
+            address = jbof_static_ip(shelf_index, eth_index)
+            redfish.configure_fabric_interface(uri, address, static_ip_netmask_str(address), mtusize=static_mtu())
+        time.sleep(20)
```
anodos325, line 306: Should we have this as constant somewhere?

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,546 @@
...
+
+        # Setup a dict with the expected IP pairs
+        shelf_ip_to_host_ip = {}
+        for eth_index in range(0, len(shelf_interfaces)):
+            shelf_ip_to_host_ip[jbof_static_ip(shelf_index, eth_index)] = initiator_static_ip(shelf_index, eth_index)
+
+        # Let's check that we have the expected hardwired IPs on the shelf
+        if set(shelf_ips) != set(list(shelf_ip_to_host_ip.keys())):
+            # This should not happen
+            verrors.add(schema, 'JBOF does not have expected IPs.')
```
anodos325, line 330: Would it be helpful to admin to include what IPs were expected and which ones were actually set?

src/middlewared/middlewared/plugins/jbof/redfish/client.py
```diff
@@ -0,0 +1,300 @@
...
+    @classmethod
+    def setup(cls):
+        # Silence InsecureRequestWarning: Unverified HTTPS request is being made to host
+        requests.packages.urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+    @classmethod
+    def ping(cls, mgmt_ip, timeout=DEFAULT_REDFISH_TIMEOUT_SECS):
+        mgmt_ip = mgmt_ip.rstrip('/')
+        url = f'https://{mgmt_ip}{REDFISH_ROOT_PATH}'
+        r = requests.get(url, verify=False, timeout=timeout)
```
anodos325, line 42: Is there a reason we're not checking the HTTP status code `r.status_code` here? Should we log an error message if this fails?

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,546 @@
...
+                else:
+                    self.logger.info('Removing RDMA interface that cannot connect to JBOF')
+                    await self.middleware.call('rdma.interface.delete', interface['id'])
+                    dirty = True
+        for shelf_ip in shelf_ips:
+            if shelf_ip in connected_shelf_ips:
+                continue
+            # Try each remaining interface
+            if dirty:
+                configured_interfaces = await self.middleware.call('rdma.interface.query')
```
yocalebo, line 400: Is there a reason to call `rdma.interface.query` for every `shelf_ip` in `shelf_ips` or can this be done once outside the for loop?

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,546 @@
...
+            connected_shelf_ips = await self.hardwire_node('', shelf_index, shelf_ip_to_mac)
+            if not connected_shelf_ips:
+                # Failed to connect any IPs => error
+                verrors.add(schema, 'Must be able to communicate with at least one interface on the expansion shelf.')
+                return
+
+    @private
+    async def hardwire_node(self, node, shelf_index, shelf_ip_to_mac):
+        localnode = not node or node == await self.middleware.call('failover.node')
+        shelf_ips = list(shelf_ip_to_mac.keys())
```
yocalebo, line 361: There is no reason to instantiate this variable at the top of the method when it's not used until you get to line 376. Furthermore, there is no reason to do `list(dict.keys())`. You can remove this line and then on line 375 do
```
for shelf_ip in shelf_ips:
    ...
```

src/middlewared/middlewared/plugins/jbof/redfish/target.py
```diff
@@ -0,0 +1,108 @@
...
+
+        auth = kwargs.get('auth', None)
+        if not auth:
+            auth = (kwargs.get('username', 'Admin'), kwargs.get('password', ''))
+
+        # print(url)
+        # print(auth)
+        # return req(url, auth=auth, verify=False)
+        return req(url, auth=auth, verify=False, data=kwargs.get('data', {})).json()
+        # return req(url, auth=auth, **{'headers': HEADER, 'data': json.dumps(kwargs.get('data', {}))}, verify=False)
```
anodos325, line 103: Remove commented code?

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,546 @@
...
+                if interface['node'] != node:
+                    continue
+                jbof_ip = jbof_static_ip_from_initiator_ip(interface['address'])
+                await self.middleware.call('jbof.nvme_connect', jbof_ip)
+        else:
+            for interface in configured_interfaces:
+                if interface['node'] != node:
+                    continue
+                jbof_ip = jbof_static_ip_from_initiator_ip(interface['address'])
+                await self.middleware.call('failover.call_remote', 'jbof.nvme_connect', [jbof_ip])
```
yocalebo, line 462: This call remote will fail on HA systems that are upgraded. We're going to have to catch ENOMETHOD (you can grep for it in our code-base and find other examples)

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,546 @@
...
+
+    @private
+    def nvme_connect(self, ip, nr_io_queues=16):
+        command = ['nvme', 'connect-all', '-t', 'rdma', '-a', ip, '--persistent', '-i', f'{nr_io_queues}']
+        ret = subprocess.run(command, capture_output=True)
+        if ret.returncode:
+            self.logger.debug(f'Failed to execute "{" ".join(command)}": {ret.stderr.decode()}')
```
yocalebo, line 469: We need to check for stderr and/or stdout (unless you're 100% positive nvme binary always reports error to stderr). We also need to decode the stderr once. (i.e.
```
if ret.returncode:
    error = ret.stderr.decode() if stderr else ret.stdout.decode()
    if not error:
        error = 'No error message reported'
```
We also need to use `%s` operators in the logger calls for better efficiency i.e
```
self.logger.debug('Failed to execute command: %r with error: %r', " ".join(command), error)

File: src/middlewared/middlewared/plugins/jbof/redfish/target.py [hunk 21]
```diff
@@ -0,0 +1,108 @@
...
+
+
+async def setup(middleware):
+    # Silence InsecureRequestWarning: Unverified HTTPS request is being made to host
+    requests.packages.urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
anodos325, line 108: Let's see about doing something less impactful than disabling globally (if possible).

src/middlewared/middlewared/plugins/jbof/crud.py
bmeagherix: Logically it should work, but there _could_ suddenly be a network outage, etc.  I was going to allow the exception to bubble up in that case.  Do you think I should catch instead?

src/middlewared/middlewared/plugins/jbof/redfish/target.py
```diff
@@ -0,0 +1,108 @@
...
+            if url.startswith(REDFISH_ROOT_PATH):
+                url = f'https://{mgmt_ip}{url}'
+            else:
+                url = f'https://{mgmt_ip}{REDFISH_ROOT_PATH}{url}'
+
+        auth = kwargs.get('auth', None)
+        if not auth:
+            auth = (kwargs.get('username', 'Admin'), kwargs.get('password', ''))
+
+        # print(url)
```
yocalebo, line 99: Leftover and can be removed?

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,546 @@
...
+        # To do that we first need to fetch the config.
+        data = await self.get_instance(id_)
+        await self.middleware.run_in_thread(self.ensure_redfish_client_cached,
+                                            data['mgmt_ip1'],
+                                            data.get('mgmt_username'),
+                                            data.get('mgmt_password'))
+        try:
+            await self.middleware.call('jbof.unwire_dataplane', data['mgmt_ip1'], data['index'])
+        except Exception:
+            self.middleware.logger.debug('Unable to unwire JBOF @%r', data['mgmt_ip1'])
```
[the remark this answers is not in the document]
bmeagherix: Depending on where the failure occurred it could mean:
- nvme disks don't get cleanly removed (very unlikely that everything fails, so they'll possibly get _uncleanly_ removed)
- host RDMA IPs don't get torn down
- ES24N networking doesn't get torn down.

I suppose the best clean up would be a reboot.

src/middlewared/middlewared/service/core_service.py
```diff
@@ -543,18 +543,29 @@ def ping(self):
...
         return 'pong'
 
-    def _ping_host(self, host, timeout):
-        return run(['ping', '-4', '-w', f'{timeout}', host]).returncode == 0
-
-    def _ping6_host(self, host, timeout):
-        return run(['ping6', '-w', f'{timeout}', host]).returncode == 0
+    def _ping_host(self, version, host, timeout, count=None, interface=None, interval=None):
+        if version == 4:
+            command = ['ping', '-4', '-w', f'{timeout}']
```
yocalebo, line 548: Just something to note:
If you do `ping -4 -w 2 -c 4`, the return code will not be `0` like you expect (regardless if the 1st 2 packets go through successfully.

src/middlewared/middlewared/utils/license.py
```diff
@@ -10,4 +10,5 @@
     9: "ES102",
     10: "ES102G2",
     11: "ES60G2",
+    12: "ES24N",
```
yocalebo, line 13: I'm assuming this has not been added to the license server, yet? This key in this dict matches the unique id for the database table on an internal license server that we have. I don't expect you to know that since it's tribal knowledge.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/jbof/crud.py
bmeagherix: We need to update `configured_interfaces` when a change is made (`rdma.interface.delete` or `rdma.interface.create`) and this needs to be done before the 
`if ifname not in configured_interface_names:
`

We could refactor to not use the `dirty` variable to control that, but it'll mean we would replace this call site with two.

src/middlewared/middlewared/plugins/jbof/crud.py
anodos325: No. I just wanted to make sure that's intentional.

src/middlewared/middlewared/plugins/jbof/crud.py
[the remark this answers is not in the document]
anodos325: Maybe generate an alert then that reboot may be required due to potentially incomplete teardown?

src/middlewared/middlewared/utils/license.py
bmeagherix: Correct.  I have zero knowledge wrt the license server, so the above may need to be changed.

src/middlewared/middlewared/plugins/jbof/crud.py
bmeagherix: Added handling

src/middlewared/middlewared/service/core_service.py
bmeagherix: Understood.  FWIW, adding an appropriate interval to the above would allow all 4 pkts to go thru in the allotted time.

src/middlewared/middlewared/plugins/jbof/redfish/client.py
bmeagherix: Logged

src/middlewared/middlewared/plugins/jbof/crud.py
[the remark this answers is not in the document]
bmeagherix: Added `JBOFTearDownFailureAlertClass`

src/middlewared/middlewared/plugins/jbof/crud.py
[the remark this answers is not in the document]
bmeagherix: Added a poll loop instead.  Checks link status on interfaces whose link was up before the config change.

src/middlewared/middlewared/plugins/rdma/interface/crud.py
```diff
@@ -0,0 +1,251 @@
...
+                return False
+        return True
+
+    async def internal_interfaces(self):
+        links = await self.middleware.call('rdma.get_link_choices')
+        ifname_to_netdev = {}
+        for link in links:
+            ifname_to_netdev[link['rdma']] = link['netdev']
+
+        if await self.middleware.call('system.product_type') in ['ENTERPRISE', 'SCALE_ENTERPRISE']:
```
yocalebo, line 228: This will only ever be `SCALE_ENTERPRISE`

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,598 @@
...
+            # Really only expect 4, but we'll over-estimate for now, as we check them anyway
+            return 6
+
+    @private
+    async def unwire_host(self, mgmt_ip, shelf_index):
+        """Unware the dataplane interfaces of the specified JBOF."""
+        possible_host_ips = []
+        possible_shelf_ips = []
+        shelf_interface_count = await self.shelf_interface_count(mgmt_ip)
+        for eth_index in range(0, shelf_interface_count):
```
yocalebo, line 561: This will not iterate over all of `shelf_interface_count` since range() is half-open. Are you sure this doesn't need to be `for eth_index in range(shelf_interface_count + 1)`?

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,598 @@
...
+            error = ret.stderr.decode() if ret.stderr else ret.stdout.decode()
+            if not error:
+                error = 'No error message reported'
+            self.logger.debug('Failed to execute command: %r with error: %r', " ".join(command), error)
+            raise CallError(f'Failed connect NVMe disks: {error}')
+        return True
+
+    @private
+    def nvme_disconnect(self, ips):
+        command = ['nvme', 'list-subsys', '-o', 'json']
```
yocalebo, line 519: Just something that I'm going to note here but not require to be changed in this PR. This command seems to be a bit heavy (and takes a surprising amount of time to return) for the information that we're getting out of it. I was able to write about 10 lines using `pathlib.Path` iterating over sysfs to get this information. I also keyed the returned object on nvme character device name (i.e. `nvme0`, `nvme1` (since those have to be unique)). It would simplify the logic in this function considerably. Anyways, just FYI. If you want the code (I didn't test too thoroughly) just DM me.

src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -0,0 +1,598 @@
...
+    def ensure_redfish_client_cached(self, mgmt_ip, username=None, password=None):
+        """Synchronous function to ensure we have a redfish client in cache."""
+        try:
+            RedfishClient.cache_get(self.middleware, mgmt_ip)
+        except KeyError:
+            # This could take a while to login, etc ... hence synchronous wrapper.
+            redfish = RedfishClient(f'https://{mgmt_ip}', username, password)
+            RedfishClient.cache_set(mgmt_ip, redfish)
+
+    @accepts()
```
anodos325, line 242: Needs `accepts(roles=['JBOF_READ'])`?

src/middlewared/middlewared/plugins/rdma/interface/crud.py
anodos325: Maybe simply `system.is_enterprise`?

src/middlewared/middlewared/plugins/jbof/crud.py
bmeagherix: Updated to make clearer.

src/middlewared/middlewared/plugins/jbof/crud.py
bmeagherix: Updated.

src/middlewared/middlewared/plugins/jbof/functions.py
```diff
@@ -0,0 +1,91 @@
...
+        return jbof_static_ip(*result)
+
+
+def initiator_ip_from_jbof_static_ip(ip):
+    result = decode_static_ip(ip)
+    if result:
+        return initiator_static_ip(*result)
+
+
+def get_sys_class_nvme():
```
yocalebo, line 55: Awesome! However, `/sys/class/nvme` could not exist so if someone were to call this on a platform that doesn't have an nvme drive, this could raise `FileNotFoundError`. You probably just want to handle this error condition in the caller.

bmeagherix: Added `try..except`, although this should only be called when the directory **_is_** in place.

yocalebo (approved):
SHIP IT!

P.S. wait for Andrew's final review before merge.

anodos325 (approved):

PR #12715, opened 2023-12-13
NAS-125700 / 24.04 / Enhance testing of NFS 'allow_nonroot' setting

mgrimesix:
Part of NAS-125674

'allow_nonroot' translates to the `insecure` option on shares defined in /etc/exports.
This commit is to improve the CI testing for this setting.

### DIFF
--- a/tests/api2/test_300_nfs.py
+++ b/tests/api2/test_300_nfs.py
@@ -930,31 +930,75 @@ def test_38_check_nfs_allow_nonroot_behavior(request):
     we append "insecure" to each exports line.
     Since this is a global option, it triggers an nfsd restart
     even though it's not technically required.
+    Linux will, by default, mount using a priviledged port (1..1023)
+    MacOS NFS mounts do not follow this 'standard' behavior.
 
     Sample:
     "/mnt/dozer/NFSV4"\
         *(sec=sys,rw,insecure,no_subtree_check)
     """
 
+    def get_client_nfs_port():
+        '''
+        Output from netstat -nt looks like:
+            tcp        0      0 127.0.0.1:50664         127.0.0.1:6000          ESTABLISHED
+        The client port is the number after the ':' in the 5th column
+        '''
+        rv = (None, None)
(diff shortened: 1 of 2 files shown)


tests/api2/test_300_nfs.py
```diff
@@ -930,31 +930,75 @@ def test_38_check_nfs_allow_nonroot_behavior(request):
...
+        '''
+        Output from netstat -nt looks like:
+            tcp        0      0 127.0.0.1:50664         127.0.0.1:6000          ESTABLISHED
+        The client port is the number after the ':' in the 5th column
+        '''
+        rv = (None, None)
+        res = ssh("netstat -nt")
+        for line in str(res).splitlines():
+            # The server will listen on port 2049
+            if f"{ip}:2049" in line.split()[3]:
```
bmeagherix, line 951: Should `in` be `==` ?
Otherwise, e.g.
```
>>> "127.0.0.0:2049" in "127.0.0.0:20499"
True
```

tests/api2/test_300_nfs.py
```diff
@@ -930,31 +930,75 @@ def test_38_check_nfs_allow_nonroot_behavior(request):
...
+        assert new_nfs_conf['allow_nonroot'] is True, new_nfs_conf
 
-    results = PUT("/nfs/", {"allow_nonroot": False})
-    assert results.status_code == 200, results.text
+        parsed = parse_exports()
+        assert len(parsed) == 1, str(parsed)
+        assert 'insecure' in parsed[0]['opts'][0]['parameters'], str(parsed)
+
+        # Confirm we allow mounts from 'non-root' ports
+        with SSH_NFS(ip, NFS_PATH, vers=4, options=['noresvport'],
```
bmeagherix, line 993: Would it be a good idea to **also** add a test to ensure can can still mount without `noresvport`?  AFAIK, `insecure` does not preclude that.

bmeagherix (changes requested):

bmeagherix (approved):

PR #12716, opened 2023-12-14
NAS-125710 / 24.04 / Add account_attributes to auth.me output

anodos325:
This adds specific account-related flags to user credentials that webui can evaluate whether user is local, our special system administrator account, etc. This is implemented as a list so that we have capability to expand what is presented without having to modify API.

This commit also removes two indirect calls to `user.query` during `auth.authenticate`.

### DIFF
--- a/src/middlewared/middlewared/plugins/auth.py
+++ b/src/middlewared/middlewared/plugins/auth.py
@@ -11,7 +11,7 @@ from middlewared.auth import (SessionManagerCredentials, UserSessionManagerCrede
                               UnixSocketSessionManagerCredentials, RootTcpSocketSessionManagerCredentials,
                               LoginPasswordSessionManagerCredentials, ApiKeySessionManagerCredentials,
                               TrueNasNodeSessionManagerCredentials)
-from middlewared.schema import accepts, Any, Bool, Datetime, Dict, Int, Password, Patch, returns, Str
+from middlewared.schema import accepts, Any, Bool, Datetime, Dict, Int, List, Password, Patch, Ref, returns, Str
 from middlewared.service import (
     Service, filterable, filterable_returns, filter_list, no_auth_required, no_authz_required,
     pass_app, private, cli_private, CallError,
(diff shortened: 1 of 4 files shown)


tests/api2/test_audit.py
```diff
@@ -259,7 +259,7 @@ def test_token_login():
                         "credentials_data": {
                             "parent": {
                                 "credentials": "LOGIN_PASSWORD",
-                                "credentials_data": {"username": "root"},
+                                "credentials_data": {"username": "root", "account_attributes": ["LOCAL", "SYS_ADMIN"]},
```
themylogin, line 262: Do we need this in the audit logs?

anodos325: I'm kind of torn, there won't be many of these, but if you want me to I can remove from the credentials.dump() output.

themylogin (approved):

yocalebo (approved):

src/middlewared/middlewared/plugins/account_/2fa.py
```diff
@@ -29,7 +29,9 @@ async def provisioning_uri(self, username):
         """
         user = await self.translate_username(username)
         user_twofactor_config = await self.middleware.call(
-            'auth.twofactor.get_user_config', user['id' if user['local'] else 'sid'], user['local'],
+            'auth.twofactor.get_user_config',
+            user['id' if user['local'] else 'sid'],
```
yocalebo, line 33: I don't like these types of comprehensions when we're checking the same variable twice. It's also kind of messy code. What do you think about doing something like:
```
user_id, local_user = 'id', 'DB_ID'
if not user['local']:
    user_id, local_user = 'sid', 'SID'

user_twofactor_config = await self.middleware.call(
    'auth.twofactor.get_user_config', user_id, local_user
)

anodos325: I'm actually testing backing out the 2fa changes right now.

PR #12717, opened 2023-12-14
NAS-125713 / 23.10.2 / Do not store and use hardware addresses for non-physical interfaces

themylogin:
VLANs have the same hardware address as their parent interface which will cause confusion in the network interface migration process:

```
[2023/12/14 05:24:45] (INFO) middlewared.setup():202 - Interface 'ens5f4' is now 'vlan370' (matched by link address '00:07:43:40:60:50')
[2023/12/14 05:24:45] (INFO) middlewared.rename():111 - Renaming interface 'ens5f4' to 'vlan370'
```


themylogin:
time 02:00

yocalebo (approved):

yocalebo (approved):

PR #12723, opened 2023-12-14
NAS-125720 / 24.04 / NAS-125720 - Add License text to repo

kmoore134:


src/middlewared/middlewared/plugins/enclosure_/enclosure2.py
```diff
@@ -1,3 +1,8 @@
+# Copyright (c) 2023 - iXsystems Inc.
```
yocalebo, line 1: Is it possible to add copyright to `LICENSE.IX` so we just, in theory, have to update it in 1 place instead of everywhere 😄 ?

kmoore134: Not a bad idea ;) I'll do that

kmoore134: Done, thanks for the feedback, that will save us some time later.

yocalebo (approved):
LGTM

PR #12727, opened 2023-12-15
NAS-125714 / 24.04 / Add more error handling for broken system dataset

anodos325:
There are some edge cases with regard to SMB passdb and groupmap interactions where user.query and group.query will fail if the system dataset is broken. Rather than failing these calls we should simply omit the SMB-related information so that dependent calls do not fail.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -299,11 +299,16 @@ class UserService(CRUDService):
 
         username_sid = {}
         if 'SMB' in additional_information:
-            for u in await self.middleware.call("smb.passdb_list", True):
-                username_sid.update({u['Unix username']: {
-                    'nt_name': u['NT username'],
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/account.py
```diff
@@ -299,11 +299,16 @@ async def query(self, filters, options):
...
+            try:
+                for u in await self.middleware.call("smb.passdb_list", True):
+                    username_sid.update({u['Unix username']: {
+                        'nt_name': u['NT username'],
+                        'sid': u['User SID'],
+                    }})
+            except Exception:
+                # Failure to retrieve passdb list often means that system dataset is
+                # broken
+                self.logger.error('Failed to retrieve passdb information', exc_info=True)
```
yocalebo, line 311: Should we short-circuit and return early at this exception?

yocalebo (approved):

PR #12731, opened 2023-12-15
NAS-124823 / 24.04 / Make sure we mount all datasets we can when doing recursive mount

sonicaj:
This commit adds changes to ensure that when we want to recursively mount datasets and if for some reason any dataset fails to mount - we do not stop in mounting the subsequent datasets as shown by the behavior of 'zfs mount -a' as well.

### DIFF
--- a/src/middlewared/middlewared/plugins/zfs_/dataset_actions.py
+++ b/src/middlewared/middlewared/plugins/zfs_/dataset_actions.py
@@ -57,7 +57,7 @@ class ZFSDatasetService(Service):
             with libzfs.ZFS() as zfs:
                 dataset = zfs.get_dataset(name)
                 if options['recursive']:
-                    dataset.mount_recursive()
+                    dataset.mount_recursive(ignore_errors=True)
                 else:
                     dataset.mount()
         except libzfs.ZFSException as e:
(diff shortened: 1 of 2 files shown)


themylogin (commented):
Can we have an integration test for that?

yocalebo:
@themylogin ping

tests/api2/test_legacy_dataset_mount.py
```diff
@@ -0,0 +1,65 @@
...
+        call('pool.snapshottask.delete', snap_task['id'])
+
+
+def test_replicated_dataset_legacy_mount():
+    with dataset(SRC_PARENT_DATASET) as parent_src:
+        with dataset(DST_PARENT_DATASET) as parent_dst:
+            with dataset(f'{SRC_PARENT_DATASET}/child1'):
+                with dataset(f'{SRC_PARENT_DATASET}/child1/legacy') as src_legacy:
+                    ssh(f'zfs set mountpoint=legacy {src_legacy}')
+                    with replication_task(parent_src, parent_dst) as rep_task:
```
themylogin, line 56: Can’t you repeat this zfs filesystem structure without using replication?

themylogin (approved):
Thank you!

PR #12733, opened 2023-12-15
NAS-125732 / 24.04 / Handle and log exception while normalizing cpu temp

Qubad786:
## Problem
The netdata python.d plugin is presently not handling and logging exceptions that may occur during the normalization process of CPU temperature data. As a consequence, the CPU temperature fails to execute due to an unhandled exception.

## Solution
To address this, exceptions that may arise during the normalization of CPU temperature data are now handled and logged appropriately. In case of an error, default data is returned, and the error is logged to facilitate debugging in the normalization process.

### DIFF
--- a/src/freenas/usr/lib/netdata/python.d/cputemp.chart.py
+++ b/src/freenas/usr/lib/netdata/python.d/cputemp.chart.py
@@ -68,10 +68,15 @@ class Service(SimpleService):
                 cpu_data[chip_name] = cpu_d
         except sensors.SensorsError as error:
             self.error(error)
-            return None
+
+        try:
+            cpu_temps = cpu_temperatures(cpu_data)
+        except Exception as error:
+            self.error(error)
+            cpu_temps = {}
 
         data = {}
-        for core, temp in cpu_temperatures(cpu_data).items():
+        for core, temp in cpu_temps.items():
             data[str(core)] = temp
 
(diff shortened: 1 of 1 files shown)


src/freenas/usr/lib/netdata/python.d/cputemp.chart.py
```diff
@@ -68,11 +68,13 @@ def get_data(self):
...
             self.error(error)
-            return None
 
-        data = {}
-        for core, temp in cpu_temperatures(cpu_data).items():
-            data[str(core)] = temp
+        try:
+            data = {}
+            for core, temp in cpu_temperatures(cpu_data).items():
+                data[str(core)] = temp
```
themylogin, line 75: If we have 8 cores and get an exception when processing, for example, 3rd, `data` will have only two values while it should have 8

Qubad786: @themylogin error if occurs would happen in `cpu_temperatures` implementation - i don't see it happening during iteration on some specific core..should we handle that too separately?

themylogin: In that case only `cpu_temperatures` should be inside the exception clause... broad exception clause especially.

themylogin (approved):

PR #12734, opened 2023-12-15
NAS-125632 / 24.04 / Update graphite exporter defaults

Qubad786:
## Problem

Default values of graphite exporter were pointing to a fixed scale release which does not sit well and needs to be updated on scale release changes.

## Solution

Update defaults to not point to a specific scale release to avoid changing/managing them each time.


src/middlewared/middlewared/alembic/versions/23.10/2023-11-10_23-19_add_exporting_table_for_reporting.py
```diff
@@ -33,12 +41,13 @@ def upgrade():
         sa.PrimaryKeyConstraint('id', name=op.f('pk_reporting_exports'))
     )
 
+    hostname = get_hostname()
```
themylogin, line 44: We should use the hostname specified in the networking configuration, not this

themylogin (changes requested):
Same for second migration

yocalebo:
@themylogin ping

themylogin (approved):
Thank you!

PR #12735, opened 2023-12-15
NAS-120363 / 24.04 / Add support for alias

bmeagherix:
Add `alias` to `scst.conf`

----
Successful CI run, plan to add a separate jira to enhance CI tests wrt alias.


src/middlewared/middlewared/etc_files/scst.conf.mako
```diff
@@ -307,6 +308,12 @@ TARGET_DRIVER iscsi {
 ## If no associated targets then disable
         enabled 0
 %   endif
+##
+## alias
+##
+%       if alias:
```
yocalebo, line 314: Spacing is different here.

yocalebo (approved):

PR #12739, opened 2023-12-15
Shift sysdataset setup to pool.create method

anodos325:
This can be called directly from pool.create job.


src/middlewared/middlewared/plugins/pool_/pool.py
```diff
@@ -649,7 +649,11 @@ async def do_create(self, job, data):
         self.middleware.create_task(self.middleware.call('disk.swaps_configure'))
         self.middleware.create_task(self.middleware.call('pool.restart_services'))
 
+        if (await middleware.call('systemdataset.config'))['pool'] == await middleware.call('boot.pool_name'):
```
yocalebo, line 652: Missing `self.middleware` 😄

PR #12741, opened 2023-12-16
NAS-125734 / 24.04 / add another data point to collections metrics

yocalebo:


william-gr (approved):

kmoore134:
Its mainly so we can get a basic number of users who are utilizing any particular feature / service. We always have to make resource decisions on where to invest in new feature work, so knowing which things are heavily used by the users helps us to prioritize things appropriately. Nothing more mysterious than that :)

PR #12755, opened 2023-12-19
NAS-125362 / 24.04 / Fix edge case for getting CPU temperature on AMD machines

Qubad786:
## Problem

For some AMD CPUs that have a single core, the temperature sensor sends a float value instead of a list. However, We expect a list, leading to a type error.

## Solution

This edge case is now handled in the AMD CPU temperature function.


yocalebo (approved):

yocalebo:
A single core CPU 🤮

PR #12757, opened 2023-12-19
NAS-125783 / 24.04 / Add query-filters for timestamps

anodos325:
Audit queries return as a field an ISO-8601 timestamp that's internally converted to a python datetime object and returned to external callers as a JSON object containing $date key. This commit adds special handling for filter-list to allow specifying ISO-8601 timestamp values for `timestamp.$date` key. In this case we will convert the specified string into datetime object during filters validation and use it for comparison with datetime object being handled in backend.

### DIFF
--- a/src/middlewared/middlewared/pytest/unit/utils/test_filter_list.py
+++ b/src/middlewared/middlewared/pytest/unit/utils/test_filter_list.py
@@ -474,29 +474,28 @@ def test__filter_list_select_as_validation():
         assert 'first item must be a string' in str(ve)
 
 
-def test__filter_list_timestamp():
-    # Test variations of not being ISO-8601 timestamp
-
-    # check for bogus string
+def test__filter_list_timestamp_invalid_string():
     with pytest.raises(ValueError) as ve:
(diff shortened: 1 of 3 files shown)


yocalebo (approved):

src/middlewared/middlewared/pytest/unit/utils/test_filter_list.py
```diff
@@ -440,3 +472,34 @@ def test__filter_list_select_as_validation():
...
+
+
+def test__filter_list_timestamp():
+    # Test variations of not being ISO-8601 timestamp
+
+    # check for bogus string
+    with pytest.raises(ValueError) as ve:
+        filter_list([], [["timestamp.$date", "=", "Canary"]])
+
+    assert 'must be an ISO-8601 formatted timestamp string' in str(ve)
```
themylogin, line 484: All these assertions and `with pytest.raises` must be separate test cases so that if one fails, others are still tested

themylogin (approved):
Thank you!

PR #12762, opened 2023-12-19
NAS-125786 / 24.04 / Clean before directive usage in ix-netif

sonicaj:
This PR is essentially a cosmetic change and no functionality is changed.


yocalebo (changes requested):
1. need a ticket 😄 
2. need an explanation on why this is needed

yocalebo (approved):
Thanks!

PR #12763, opened 2023-12-19
NAS-125626 / 24.04 / add webui.enclosure.dashboard endpoint

yocalebo:
This follows the same paradigm that I did with `pool.dataset.details` endpoint. The backend provides an endpoint that gathers up all the relevant information so the UI team can focus on implementing the UX side of it and not worrying about querying random endpoints on our side and then stitching them together on their side. This has numerous benefits but the primary one in this instance is that I make it as fast and efficient as possible (takes ~0.5 - 0.8 seconds on M60, with 12 JBODs, with ~1.2k HDDs).

The majority of the changes here are related to documenting returns objects as well as updating the unit tests.

The changes I've implemented are:
1. use `ControllerModels` enum in `nvme2.py` for the model checks (like we do in enclosure_class.py)
2. add a `zpool.status` endpoint that is designed to represent the equivalent of running `zpool status` from the CLI
3. add the new top-level keys to the `enclosure_class.py::Enclosure()` class (and in `nvme2.py`)
4. add `weui.enclosure.dashboard` which stitches all of the information together for the webUI team. (This endpoint negates essentially, the entirety of all the logic they had on their side for generating the enclosure webpage.) Note: the information returned in this endpoint is at the bequest of the UI team.
5. update the `test_enclosure2_query` unit tests to catch any regressions
6. remove the `test_enclosure_class.py` unit tests since these are unnecessary because the same functionality is tested in `test_enclosure2_query` tests


src/middlewared/middlewared/plugins/webui/enclosure.py
```diff
@@ -0,0 +1,294 @@
...
+            "pci": "17:0:19:0",
+            "rackmount": True,
+            "top_loaded": False,
+            "front_slots": 24,
+            "rear_slots": 4,
+            "internal_slots": 0,
+            "elements": {
+                "Array Device Slot": {
+                    "1": {
+                        "descriptor": "slot00",
```
themylogin, line 59: Do we need so many items in the example?

yocalebo: Nope, fixed.

themylogin (approved):

anodos325:
Do we maybe need to add STORAGE_READ and STORAGE_WRITE (or some appropriate name) roles for interacting with enclosures, disks, etc? Then grant READONLY role the STORAGE_READ role?

yocalebo:
> Do we maybe need to add STORAGE_READ and STORAGE_WRITE (or some appropriate name) roles for interacting with enclosures, disks, etc? Then grant READONLY role the STORAGE_READ role?

Yep, we should for sure. Trying to think of a proper name for the role 😓

PR #12770, opened 2023-12-19
NAS-125795 / 24.04 / Add FAILOVER roles

anodos325:
This commit adds FAILOVER_READ and FAILOVER_WRITE roles that can be used as webui hints on whether to expose options to end-user.


themylogin:
Why do we need to grant calling failover.call_remote to READONLY if Web UI is not using it?

anodos325:
> Why do we need to grant calling failover.call_remote to READONLY if Web UI is not using it?

Webui uses it.

themylogin:
@anodos325 for a very finite number of cases I suppose?

anodos325:
> @anodos325 for a very finite number of cases I suppose?

Do you want me to create a whitelist of methods?

themylogin:
@anodos325 I see WebUI using `failover.call_remote`  as sort of a hack. 

I searched their sources and the methods they call are `system.info` and `ipmi.lan.query` / `ipmi.lan.update`. First is understandable (however, we can easily create a wrapper); second looks strange (should we really be doing that on the UI side?)

@yocalebo what are your thoughts on WebUI calling methods directly on the standby node? Shouldn't the active node take care of that sort of communication/management?

yocalebo:
> @anodos325 I see WebUI using `failover.call_remote` as sort of a hack.
> 
> I searched their sources and the methods they call are `system.info` and `ipmi.lan.query` / `ipmi.lan.update`. First is understandable (however, we can easily create a wrapper); second looks strange (should we really be doing that on the UI side?)
> 
> @yocalebo what are your thoughts on WebUI calling methods directly on the standby node? Shouldn't the active node take care of that sort of communication/management?
> 
> I am fine with having just a whitelist for 24.04 if we have more important things to do though.

@themylogin yes, there is 0 reason webUI should be doing this. I've spoken to Evgeny and Andrew about this in slack. I'm going to provide a dedicated endpoint for them so we don't have to worry about adding perms to `call_remote` and leave it as accessible to root only.

yocalebo:
@themylogin I've updated this PR to remove the changes to `call_remote`. I want that method to stay accessible to root only and not be consumed by any 3rd parties. I'll work with webUI on adding dedicated endpoints/logic for when they need standby controller info.

yocalebo (approved):

PR #12771, opened 2023-12-19
NAS-125796 / 24.04 / Expand replication plugin access for readonly admins

anodos325:
Webui team requested access to the following endpoints:
* replication.list_datasets
* replication.list_naming_schemas
* replication.count_eligable_manual_snapshots


themylogin:
Will REPLICATION_READ role not work with the Web UI?

anodos325:
After some discussion, closing this in favor of having webui not call these endpoints when READONLY.

PR #12800, opened 2023-12-26
NAS-125621 / 24.04 / Handle missing users and groups in NFS share configuration

mgrimesix:
### Problem
On upgrade, if `maproot` or `mapall` is configured on an NFS share and the user or group is deleted as part of the upgrade, then all NFS shares end up disabled.

### Fix
- In nfs exports mako add check for required users or groups and disable NFS shares that use missing users or groups.
- Add oneshot alert when missing user or group is detected.   Remove oneshot alert when share configuration is repaired.
- Enhance reporting on attempt to configure NFS share with invalid user or group setting

### Other fixes/improvements
- Add CI test for missing user and group with NFS `maproot` and `mapall` configuration options
- Some refactoring of the NFS CI test module

### DIFF
--- a/src/middlewared/middlewared/etc_files/exports.mako
+++ b/src/middlewared/middlewared/etc_files/exports.mako
@@ -6,59 +6,48 @@
     from contextlib import suppress
     from middlewared.plugins.nfs_.utils import get_domain, leftmost_has_wildcards, get_wildcard_domain
 
-    def do_map(share, map_type, map_ids):
+    
+    def do_map(share, map_type, map_ids, alert_shares):
         output = []
-        raise_keyerror = False
+        invalid_name = []
 
         if usrname := share[f'{map_type}_user']:
-            uid = middleware.call_sync('user.username_to_uid', usrname)
-            if uid is not None:
+            try:
+                uid = middleware.call_sync('user.get_user_obj', {'username': usrname})['pw_uid']
                 map_ids[f'{map_type}_user'] = uid
                 output.append(f'anonuid={uid}')
-            else:
+            except KeyError:
                 # report the invalid name
+                middleware.logger.info(f"adding {usrname} to invalid_name")
                 map_ids[f'{map_type}_user'] = usrname
-                raise_keyerror = True
+                invalid_name.append(usrname)
 
         if grpname := share[f'{map_type}_group']:
-            gid = middleware.call_sync('group.groupname_to_gid', grpname)
-            if gid is not None:
+            try:
+                gid = middleware.call_sync('group.get_group_obj', {'groupname': grpname})['gr_gid']
                 map_ids[f'{map_type}_group'] = gid
                 output.append(f'anongid={gid}')
-            else:
-                # report the invalid group
+            except KeyError:
+                # report the invalid name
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/etc_files/exports.mako
```diff
@@ -8,24 +8,55 @@
...
+            if gid is not None:
+                map_ids[f'{map_type}_group'] = gid
+                output.append(f'anongid={gid}')
+            else:
+                # report the invalid group
+                map_ids[f'{map_type}_group'] = grpname
+                raise_keyerror = True
+
+        if raise_keyerror:
+            raise KeyError
```
yocalebo, line 34: We should set `__context__` here so this needs to look like
```
raise KeyError(f'Missing {key}') from None
```

src/middlewared/middlewared/etc_files/exports.mako
```diff
@@ -8,24 +8,55 @@
...
+                        v for k, v in map_ids.items() if type in k and not isinstance(v, int)
+                    ])
+                    middleware.logger.warning(
+                        f"{sharepath}: Disabling this NFS share. "
+                        f"Mapping invalid names: {invalid_names}"
+                    )
+                    middleware.call_sync("alert.oneshot_create", "NFSexportMappingInvalidNames",
+                        {'id': id, 'path': sharepath, 'names': invalid_names}
+                    )
+                    raise KeyError
```
yocalebo, line 55: Same review that I gave earlier.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/account_/internal_cache.py
```diff
@@ -20,6 +20,39 @@ async def get_builtin_user_id(self, username):
         except KeyError:
             raise CallError(f'{username!r} user not found', errno.ENOENT)
 
+    @private
+    async def username_to_uid(self, user_name):
```
anodos325, line 24: This basically duplicates `user.get_user_obj` We should just use the public method. It raises KeyError if user doesn't exist.

tests/api2/test_300_nfs.py
```diff
@@ -394,6 +380,32 @@ def test_11_perform_server_side_copy(request):
...
+    group_id = call('group.groupname_to_gid', data['name'])
+    assert group_id == data['gid'], f"Expected {data['name']} GID to be {data['gid']}, but found {group_id}"
+
+
+@pytest.mark.parametrize('data', [
+    ({'name': 'root', 'uid': 0, 'gid': 0}),
+    ({'name': 'ftp', 'uid': 121, 'gid': 14}),
+    ({'name': None, 'uid': 3456, 'gid': 3456}),  # Assumes 3456 is not attached to any name
+])
+def test_15_id_to_name(data):
```
anodos325, line 401: Duplicate test number.

anodos325: Can we remove these tests since methods were removed?

anodos325 (approved):

yocalebo (approved):
Thanks!

PR #12801, opened 2023-12-27
NAS-125689 / 24.04 / Make `core.download` no_authz_required, but validate access permissions to called method inside

themylogin:
UI can continue to use `core.download` as it was doing it before.

The only change wrt to `core.download` is that unprivileged user will not be able to call `filesystem.get`. I looked in the UI code and `filesystem.get` was used for the following cases:
* Download job logs (this will be replaced with `core.job_download_logs`, UI ticket https://ixsystems.atlassian.net/browse/NAS-125918)
* Download CA/CSR/certificates (these are not RBAC yet so no changes are needed)
* Download VM logs (same as above)

### DIFF
--- a/src/middlewared/middlewared/service/core_service.py
+++ b/src/middlewared/middlewared/service/core_service.py
@@ -208,6 +208,26 @@ class CoreService(Service):
         return jobs
 
     @no_authz_required
+    @accepts(Int('id'), Str('filename'), Bool('buffered', default=False))
+    @pass_app(rest=True)
+    async def job_download_logs(self, app, id_, filename, buffered):
+        """
+        Download logs of the job `id`.
+
+        Please see `core.download` method documentation for explanation on `filename` and `buffered` arguments,
+        and return value.
+        """
+        if app is None:
+            job = self.middleware.jobs[id_]
+        else:
+            job = self.__job_by_credential_and_id(app.authenticated_credentials, id_)
+
+        if job.logs_path is None:
+            raise CallError('This job has no logs')
+
+        return (await self._download(app, 'filesystem.get', [job.logs_path], filename, buffered))[1]
+
+    @no_authz_required
     @accepts(Int('id'))
     @job()
     async def job_wait(self, job, id_):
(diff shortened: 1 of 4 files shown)


themylogin:
time 4:00

src/middlewared/middlewared/service/core_service.py
```diff
@@ -207,6 +207,26 @@ def get_jobs(self, app, filters, options):
...
+        """
+        Download logs of the job `id`.
+
+        Please see `core.download` method documentation for explanation on `filename` and `buffered` arguments,
+        and return value.
+        """
+        if app is None:
+            job = self.middleware.jobs[id_]
+        else:
+            job = self.__job_by_credential_and_id(app.authenticated_credentials, id_)
```
yocalebo, line 223: If I'm understanding correctly, if a bogus id is given then `job` object will be `NoneType`?

yocalebo:
Seems like this isn't complete? The ticket states the UI team uses core.download for a bunch of other methods
```
error logs
job logs
pod logs
vm logs
audit export
certificates
CAs
CSRs
pool.dataset.export_keys_for_replication
pool.dataset.export_keys
pool.dataset.export_key
saving debug
saving config

src/middlewared/middlewared/service/core_service.py
themylogin: No, `__job_by_credential_and_id` returns a job or raises an exception

themylogin: If a bogus ID is given then a KeyError will be raised in that line

yocalebo: okay but `__job_by_credential_and_id` raises `CallError`. Since this is for the UI team, should we check with them to see if they want this endpoint to wrap these potential errors in `ValidationErrors`?

themylogin:
> Seems like this isn't complete? The ticket states the UI team uses core.download for a bunch of other methods
> 
> ```
> error logs
> job logs
> pod logs
> vm logs
> audit export
> certificates
> CAs
> CSRs
> pool.dataset.export_keys_for_replication
> pool.dataset.export_keys
> pool.dataset.export_key
> saving debug
> saving config
> ```

What else needs to be done with regards to that?

PR #12804, opened 2023-12-27
NAS-125829 / 24.04 / Do not upgrade boot pool automatically as users can't rollback to Bluefin then

themylogin:


themylogin:
time 0:05

yocalebo (approved):

PR #12813, opened 2023-12-28
NAS-125932 / 23.10.2 / Fix non-physical interface link address getting into database when it is updated

themylogin:


yocalebo (approved):

themylogin:
time 2:00

PR #12815, opened 2023-12-28
NAS-125938 / 24.04 / Remove glusterd support from middlewared

anodos325:
This commit removes the gluster plugin and majority of clustering-related backend and tests. Portions of the CTDB plugin are currently retained as it will be refactored to allow better state persistence with SMB on HA hardware.


src/middlewared/middlewared/plugins/cloud_backup/sync.py
```diff
@@ -22,13 +22,7 @@ async def restic(middleware, job, cloud_backup, dry_run):
     cmd = None
     try:
         if await middleware.call("filesystem.is_cluster_path", cloud_backup["path"]):
```
yocalebo, line 24: Forgot to remove or is there a reason to `pass`?

yocalebo (approved):
SHIP IT

PR #12822, opened 2024-01-02
NAS-125985 / 24.04 / Add SMB client tests for ZFS ACL compatibiltiy

anodos325:
This commit adds basic tests for the SMB client mount option that maps NT security descriptor to a ZFS ACL xattr. Tests are via loopback mount to ensure that all SIDs in the security descriptor will be properly mapped to Unix IDs.

### DIFF
--- a/src/middlewared/middlewared/test/integration/assets/smb.py
+++ b/src/middlewared/middlewared/test/integration/assets/smb.py
@@ -2,6 +2,7 @@
 import contextlib
 import logging
 import os
+import shlex
 import sys
 
 from middlewared.test.integration.utils import call, ssh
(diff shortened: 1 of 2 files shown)


tests/api2/test_smb_client.py
```diff
@@ -0,0 +1,185 @@
...
+                try:
+                    call('service.start', 'cifs')
+                    yield {'dataset': ds, 'share': s, 'user': u}
+                finally:
+                    call('service.stop', 'cifs')
+
+
+@pytest.fixture(scope='module')
+def mount_share(setup_smb_tests):
+    with smb_mount(setup_smb_tests['share']['name'], 'smbuser', 'Abcd1234$') as mp:
```
themylogin, line 64: I don't see `Abcd1234$` constant defined anywhere else in the project, should we make it a constant?

src/middlewared/middlewared/test/integration/assets/smb.py
```diff
@@ -23,3 +32,21 @@ def smb_share(path, name, options=None):
...
+def smb_mount(share, username, password, local_path='/mnt/cifs', options=None, ip=default_ip):
+    mount_options = [f'username={username}', f'password={password}'] + (options or [])
+    mount_cmd = [
+        'mount.cifs', f'//{ip}/{share}', local_path,
+        '-o', ','.join(mount_options)
+    ]
+
+    mount_string = ' '.join(mount_cmd)
+
+    ssh(f'mkdir {local_path}; {mount_string}')
```
themylogin, line 47: We should use `shlex.quote` in case `local_path` contains spaces. Or remove the ability to redefine it as it is not redefined by any caller.

anodos325: There will be more mounts used as I add more tests.

tests/api2/test_smb_client.py
```diff
@@ -0,0 +1,188 @@
...
```
themylogin, line 67:
```suggestion
    with smb_mount(setup_smb_tests['share']['name'], 'smbuser', TMP_SMB_USER_PASSWORD) as mp:
```

themylogin (approved):

PR #12828, opened 2024-01-02
NAS-125590 / 13.1 / Create nut user for ups service

Qubad786:
## Problem
In CORE 13.1, the user being used for upsmon service has been changed from `uccp` to `nut`. This alteration in the username is causing the Nut UPS service to crash.

## Solution
The previous user `uccp` has been replaced with the `nut` user to ensure the UPS service starts properly.


src/middlewared/middlewared/assets/account/builtin/freebsd/passwd
```diff
@@ -16,6 +16,7 @@ proxy:*:62:62:Packet Filter pseudo-user:/nonexistent:/usr/sbin/nologin
 _pflogd:*:64:64:pflogd privsep user:/var/empty:/usr/sbin/nologin
 _dhcp:*:65:65:dhcp programs:/var/empty:/usr/sbin/nologin
 uucp:*:66:66:UUCP pseudo-user:/var/spool/uucppublic:/usr/local/libexec/uucp/uucico
+nut:*:316:316:nut:/nonexistent:/bin/sh
```
anodos325, line 19: Perhaps `/var/empty` instead of `/nonexistent`.

anodos325: Why don't we change the runas user back to UUCP in 13.1? What's motivation for changing user?

yocalebo (changes requested):
We can change the port on our side to use `uucp` instead of adding another local user.

PR #12839, opened 2024-01-03
NAS-125877 / 24.04 / Make sure we don't exhaust /var/run with catalogs

sonicaj:
This PR adds changes to make sure we don't exhaust /var/run by cloning in catalogs. These are the changes we are introducing to prevent this scenario from happening:

1. Make sure we can't add more catalogs if apps pool is not configured
2. Do not sync non official catalogs if apps pool is not configured as it will result in tmpfs being exhausted
3. Make sure we don't sync non-official catalogs if apps pool's root dataset is locked

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/update.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/update.py
@@ -196,12 +196,11 @@ class CatalogService(CRUDService):
     @private
     async def dataset_mounted(self):
         if k8s_dataset := (await self.middleware.call('kubernetes.config'))['dataset']:
-            if catalog_ds := await self.middleware.call(
+            return bool(await self.middleware.call(
                 'filesystem.mount_info', [
                     ['mount_source', '=', os.path.join(k8s_dataset, 'catalogs')], ['fs_type', '=', 'zfs'],
                 ],
-            ):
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/catalogs_linux/update.py
```diff
@@ -193,6 +193,27 @@ async def common_validation(self, catalog, schema, data):
 
         verrors.check()
 
+    @private
+    async def dataset_mounted(self):
+        if k8s_dataset := (await self.middleware.call('kubernetes.config'))['dataset']:
+            if catalog_ds := await self.middleware.call(
+                'zfs.dataset.query', [['id', '=', os.path.join(k8s_dataset, 'catalogs')]], {
+                    'extra': {'properties': ['mounted']}
```
yocalebo, line 201: Why `zfs.dataset.query`? This is expensive even if you just do `properties: ['mounted']` since it retrieves all children by default.

Why not check `filesystem.mount_info`?

sonicaj: There are no children of this dataset but yup i can use that one to retrieve mount related data

anodos325 (approved):

yocalebo (approved):

PR #12842, opened 2024-01-04
NAS-125878 / 24.04 / Never run more than 8 disk temperature queries in parallel

themylogin:


themylogin:
time 0:15

yocalebo:
I have a POC for querying ATA drives and their temperatures using ctypes and ioctl. Need to add the equivalent for SCSI as well. Once I'm confident in the code, we should switch to this instead of doing subprocess.

themylogin:
@yocalebo so closing this and all related bugs?

yocalebo:
> @yocalebo so closing this and all related bugs?

no, sorry. I'd still like to use this as a fallback in case of error and I'm not comfortable pushing the code in its current state.

src/middlewared/middlewared/utils/asyncio_.py
```diff
@@ -1,10 +1,10 @@
 import asyncio
 
 
-async def asyncio_map(func, arguments, limit=None):
-    semaphore = None
-    if limit is not None:
-        semaphore = asyncio.BoundedSemaphore(limit)
+async def asyncio_map(func, arguments, limit=None, *, semaphore=None):
+    if limit is not None or semaphore is not None:
```
sonicaj, line 5: Should we raise an exception if both limit and semaphore are specified as `limit` would be redundant in that case ?

sonicaj (approved):

PR #12845, opened 2024-01-04
Add a webui crypto service for UI consuming crypto plugin endpoints

sonicaj:
This PR adds changes to add a service which webui can consume for certain crypto plugin endpoints which it requires to make sure certificate subsystem works as intended. Integration tests have also been added.

### DIFF
--- a/src/middlewared/middlewared/plugins/webui/crypto.py
+++ b/src/middlewared/middlewared/plugins/webui/crypto.py
@@ -1,0 +1,22 @@
+from middlewared.schema import accepts, Int
+from middlewared.service import Service
+
+
+class WebUICryptoService(Service):
+
+    class Config:
+        namespace = 'webui.crypto'
+        private = True
+        cli_private = True
+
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/webui/crypto.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class WebUICryptoService(Service):
+
+    class Config:
+        namespace = 'webui.crypto'
+        private = True
+        cli_private = True
+
+    @accepts(roles=['READONLY'])
```
yocalebo, line 12: This is great! However, let's add 2 new roles `CERT_READ` and `CERT_WRITE`. Then you can add `CERT_READ` as a member of `READONLY`. This follows same paradigm as the other plugins we've done.

yocalebo (changes requested):

yocalebo (approved):

PR #12850, opened 2024-01-04
Add roles for apps

sonicaj:
This PR adds changes to introduce roles for apps plugin. What has been essentially done is that 3 major roles have been added:

1. APPS_*
2. CATALOG_*
3. KUBERNETES_*

`APPS_*` based roles already have `CATALOG_*` roles so they can do CRUD operations on apps/catalogs. Users only having `CATALOG_*` roles can do CRUD operations on catalogs. `KUBERNETES_*` roles allows to configure/unconfigure k8s cluster and also allow/restrict taking/viewing backups of the kubernetes cluster.


src/middlewared/middlewared/plugins/catalogs_linux/apps.py
```diff
@@ -8,7 +8,7 @@ class AppService(Service):
     class Config:
         cli_namespace = 'app'
 
-    @filterable
+    @filterable(roles=['READONLY', 'CATALOG_READ'])
```
yocalebo, line 11: You don't need to add `READONLY` and `CATALOG_READ` here if you add `CATALOG_READ` as a member of `READONLY`.

src/middlewared/middlewared/role.py
```diff
@@ -48,27 +48,31 @@ class Role:
...
-                               'KEYCHAIN_CREDENTIAL_READ',
-                               'REPORTING_READ',
-                               'REPLICATION_TASK_CONFIG_READ',
-                               'REPLICATION_TASK_READ',
-                               'SERVICE_READ',
-                               'SNAPSHOT_TASK_READ',
-                               'SUPPORT_READ',
-                               'SYSTEM_AUDIT_READ'],
-                     builtin=False),
+    'READONLY': Role(includes=[
```
yocalebo, line 51: Am I blind? Seems like `APPS_READ` is missing from `READONLY`?

src/middlewared/middlewared/plugins/catalogs_linux/apps.py
```diff
@@ -8,7 +8,7 @@ class AppService(Service):
     class Config:
         cli_namespace = 'app'
 
-    @filterable
+    @filterable(roles=['READONLY'])
```
yocalebo, line 11: I'm sorry but I've been confusing on my review. I meant to add `APP_READ` to `READONLY` and then make these all `roles=['APP_READ']`

src/middlewared/middlewared/role.py
```diff
@@ -48,27 +48,32 @@ class Role:
...
```
themylogin, line 51: I wonder if we can do `ROLES['READONLY'] = Role(includes=[role for role in ROLES if role.endswith('_READ')]`)

yocalebo (approved):

PR #12854, opened 2024-01-05
NAS-126058 / 24.04 / Fix roles

themylogin:
* Credentials had all roles by default. This will be a huge security risk if we start using. roles to determine if users can execute certain things. We should default to `False`, credentials must have roles explicitly.
* As a consequence, token credentials (all WebUI sessions) had all roles, including `READONLY` role. This made the middleware redact all private variables for all return values for WebUI even if we were root.
* `has_role("READONLY")` is a weird way to determine whether we should redact private variables for return values. Should we just only show them to full admins?
* As for the original issue (SMART test schema being inconsistent with the actual return value and failing to redact private variables), let's fix this once we have SMART tests available to non-privileged users.

### DIFF
--- a/src/middlewared/middlewared/auth.py
+++ b/src/middlewared/middlewared/auth.py
@@ -25,7 +25,7 @@ class SessionManagerCredentials:
         return True
 
     def has_role(self, role):
-        return True
+        return False
 
     def notify_used(self):
         pass
--- a/src/middlewared/middlewared/auth.py
+++ b/src/middlewared/middlewared/auth.py
@@ -22,7 +22,7 @@ class SessionManagerCredentials:
         return True
 
     def authorize(self, method, resource):
-        return True
+        return False
 
     def has_role(self, role):
         return False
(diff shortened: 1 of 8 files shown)


themylogin:
https://ci.tn.ixsystems.net/jenkins/job/TrueNAS%20SCALE%20-%20Unstable/job/API%20Tests%20-%20TrueNAS%20SCALE%20with%20middleware%20reinstalled%20from%20a%20branch%20Incremental/1591/

themylogin:
@anodos325 @yocalebo what do you think about #3?

themylogin:
@anodos325 I agree with you, done.

themylogin:
@yocalebo @anodos325 sorry, one more thing. I also removed redacting out private fields for non-user access (e.g. API keys)

themylogin:
https://ci.tn.ixsystems.net/jenkins/job/TrueNAS%20SCALE%20-%20Unstable/job/API%20Tests%20-%20TrueNAS%20SCALE%20with%20middleware%20reinstalled%20from%20a%20branch%20Incremental/1592/

anodos325:
> @yocalebo @anodos325 sorry, one more thing. I also removed redacting out private fields for non-user access (e.g. API keys)

That's reasonable, otherwise we would break existing consumers.

anodos325 (approved):

yocalebo (approved):

themylogin:
time 0:30

PR #12857, opened 2024-01-07
NAS-125784 / 24.04 / Add VM and VM device roles

Qubad786:
## Context

PR adds roles for VM and VM device services endpoints.
Adds `READONLY` role for public endpoints in use by WebUI.


themylogin (changes requested):
VM display device has password, which is marked as private. This means that, in general case, it should be redacted when displaying to a read-only user. Please confirm if this is the case with @yocalebo and, if necessary, add the necessary logic and the test to `test_account_privilege_role_private_fields.py`.

yocalebo:
@Qubad786 friendly ping 👍

themylogin (approved):
Thank you!

PR #12874, opened 2024-01-09
NAS-126705 / 24.04 / Fix hactl returning UNKNOWN status

yocalebo:
On Cobia, `hactl` is showing a `Failover status: UNKNOWN` when the current node is processing a failover event. This is because a reason `LOC_FAILOVER_ONGOING` is missing in the enum in `hactl`. To fix this issue, I've created a `DisabledReasonsEnum` in `failover_/disabled_reasons.py` module. In `failover.disabled.reasons`, I've used this new enum when adding to the list of reasons of why failover can be non-functional. Finally, I've imported this enum into `hactl` script and used it so that, in theory, problems like this don't happen in the future.

### DIFF
--- a/src/middlewared/middlewared/plugins/failover_/disabled_reasons.py
+++ b/src/middlewared/middlewared/plugins/failover_/disabled_reasons.py
@@ -2,11 +2,31 @@
 #
 # Licensed under the terms of the TrueNAS Enterprise License Agreement
 # See the file LICENSE.IX for complete terms and conditions
+from enum import Enum
 
 from middlewared.schema import accepts, returns, List, Str
 from middlewared.service import Service, pass_app, no_auth_required, private
 from middlewared.plugins.interface.netif import netif
 from middlewared.utils.zfs import query_imported_fast_impl
+
+
+class DisabledReasonsEnum(str, Enum):
+    NO_CRITICAL_INTERFACES = 'No network interfaces are marked critical for failover.'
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/failover_/disabled_reasons.py
```diff
@@ -22,24 +41,8 @@ class Config:
...
-        NO_FENCED - Zpools are imported but fenced isn't running.
-        LOC_FAILOVER_ONGOING - This node is currently processing a failover event.
-        REM_FAILOVER_ONGOING - Other node is currently processing a failover event.
-        NO_HEARTBEAT_IFACE - Local heartbeat interface does not exist.
-        NO_CARRIER_ON_HEARTBEAT - Local heartbeat interface is down.
+        """Returns a list of reasons why failover is not enabled/functional.
+        See `DisabledReasonsEnum` for the reasons and their explanation.
         """
         reasons = self.middleware.call_sync('failover.disabled.get_reasons', app)
         if reasons != FailoverDisabledReasonsService.LAST_DISABLED_REASONS:
```
bmeagherix, line 48: Probably fine (and orthogonal to the changes in the PR), but Ill just observe that this will fire on the change from `None` to the empty set (or non-empty set, as the case may be).

bmeagherix (approved):

PR #12881, opened 2024-01-10
NAS-125921 / 24.04 / Make sure middleware process respects CA bundle

sonicaj:


yocalebo (approved):

src/middlewared/middlewared/plugins/crypto_/__init__.py
```diff
@@ -21,5 +21,10 @@ async def setup(middleware):
                 'Failed to set certificate for system.general plugin: %s', e, exc_info=True
             )
 
+    try:
+        await middleware.call('core.environ_update', {'REQUESTS_CA_BUNDLE': '/etc/ssl/certs/ca-certificates.crt'})
+    except Exception as e:
```
yocalebo, line 26: Why not `failure=True`?

sonicaj: It does not really disrupt certificate setup process as that reflects to HTTPS always being available for NAS

themylogin: Why can't this variable be put to the systemd unit file?

themylogin (approved):
Thank you!

PR #12884, opened 2024-01-11
NAS-126745 / 24.04 / add new endpoints for UI team

yocalebo:
This adds 2 new endpoints for the UI team to finalize the Roles work.
1. `system.advanced.sed_global_password_is_set`
2. `system.security.info.fips_available`

Finally, this moves the logic out of the `security/update.py` module to `security/info.py` where it makes sense. There should be no change in functionality.

### DIFF
--- a/src/middlewared/middlewared/plugins/security/info.py
+++ b/src/middlewared/middlewared/plugins/security/info.py
@@ -1,9 +1,7 @@
-import os
 from subprocess import run
 
 from middlewared.schema import accepts, returns, Bool
 from middlewared.service import CallError, Service
-from middlewared.plugins.system.product import LICENSE_FILE
 
 
 class SystemSecurityInfoService(Service):
@@ -19,10 +17,7 @@ class SystemSecurityInfoService(Service):
         mode made be toggled on this system"""
         # being able to toggle fips mode is hinged on whether
         # or not this is an iX licensed piece of hardware
-        # we also (at time of writing) don't put a specific
-        # feature flag in the license so we don't need to decode
-        # it, just need to check its existence
-        return os.path.exists(LICENSE_FILE)
+        return bool(self.middleware.call_sync('system.license'))
 
     @accepts(roles=['READONLY'])
     @returns(Bool('fips_available'))
--- a/src/middlewared/middlewared/plugins/security/info.py
+++ b/src/middlewared/middlewared/plugins/security/info.py
@@ -19,6 +19,9 @@ class SystemSecurityInfoService(Service):
         mode made be toggled on this system"""
         # being able to toggle fips mode is hinged on whether
         # or not this is an iX licensed piece of hardware
+        # we also (at time of writing) don't put a specific
+        # feature flag in the license so we don't need to decode
+        # it, just need to check its existence
         return os.path.exists(LICENSE_FILE)
 
     @accepts(roles=['READONLY'])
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/security/info.py
```diff
@@ -0,0 +1,31 @@
...
+
+    class Config:
+        namespace = 'system.security.info'
+        cli_namespace = 'system.security.info'
+
+    @accepts(roles=['READONLY'])
+    @returns(Bool('fips_available'))
+    def fips_available(self):
+        """Returns a boolean identifying whether or not FIPS
+        mode made be toggled on this system"""
```
sonicaj, line 17: mode made ?

src/middlewared/middlewared/plugins/security/update.py
```diff
@@ -73,8 +66,5 @@ async def do_update(self, data):
 
     @private
     def fips_enabled(self):
-        cp = subprocess.run(['openssl', 'list', '-providers'], capture_output=True)
-        if cp.returncode:
-            raise CallError(f'Failed to determine if fips is enabled: {cp.stderr.decode()}')
-
-        return b'OpenSSL FIPS Provider' in cp.stdout
+        # TODO: remove this method from this CRUD file
```
sonicaj, line 69: It should be safe to remove this method after removing it's usages as this is a private method

src/middlewared/middlewared/plugins/security/info.py
yocalebo: fixed, thanks

src/middlewared/middlewared/plugins/security/update.py
yocalebo: I wasn't sure if this was called anywhere in our installer code? I actually couldn't find a single occurrence of this endpoint being used anywhere?

sonicaj: @yocalebo it's being used in an integration test
```
./tests/api2/test_fips.py:17:    assert call('system.security.fips_enabled') is flag
```

src/middlewared/middlewared/plugins/security/info.py
```diff
@@ -0,0 +1,31 @@
...
+        cli_namespace = 'system.security.info'
+
+    @accepts(roles=['READONLY'])
+    @returns(Bool('fips_available'))
+    def fips_available(self):
+        """Returns a boolean identifying whether or not FIPS
+        mode may be toggled on this system"""
+        # being able to toggle fips mode is hinged on whether
+        # or not this is an iX licensed piece of hardware
+        return bool(self.middleware.call_sync('system.license'))
```
sonicaj, line 20: Also this method can be asynchronous ?

src/middlewared/middlewared/plugins/security/update.py
yocalebo: fixed, thanks

src/middlewared/middlewared/plugins/security/info.py
yocalebo: No real benefit to making it asynchronous

sonicaj: This will result in it being run in a thread when we don't really need it to be in a thread

sonicaj (approved):

src/middlewared/middlewared/plugins/security/info.py
yocalebo: it's calling a synchronous routine which runs in our thread pool no matter how we slice it.

sonicaj: Ah, makes sense

PR #12886, opened 2024-01-11
NAS-125367 / 24.10 / Disable learning of bridge interface members

sonicaj:
This commit fixes an issue where on certain hardware, the bridge interface takes some time to set up, causing a delay in network setup and disrupting other services when they attempt to use the network. To address this issue after syncing with Caleb we decided that we will disable learning of all bridge members allowing the bridge to become active without unnecessary delay

### DIFF
--- a/src/middlewared/middlewared/plugins/interface/netif_linux/bridge.py
+++ b/src/middlewared/middlewared/plugins/interface/netif_linux/bridge.py
@@ -18,6 +18,9 @@ class BridgeMixin:
     def add_member(self, name):
         run(["ip", "link", "set", name, "master", self.name])
 
+    def disable_learning(self, name):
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/interface/netif_linux/bridge.py
```diff
@@ -18,6 +18,9 @@ class BridgeMixin:
     def add_member(self, name):
         run(["ip", "link", "set", name, "master", self.name])
 
+    def disable_learning(self, name):
```
themylogin, line 21: Can it be `set_learning`?

sonicaj: Definitely

themylogin (approved):

yocalebo (approved):

PR #12892, opened 2024-01-15
NAS-126741 / 24.04 / Fix catalog sync tests

sonicaj:
Tests in`test_catalog_sync.py` are failing because the k8s cluster starts unconfiguring before it is fully set up. Consequently, the cluster is unable to stop properly, leading to a timeout error.

In this PR we make changes to make sure the k8s cluster is fully set up before initiating the unconfiguration process. This ensures that the cluster stops properly when needed and subsequently the tests run in a neat/clean manner.

### DIFF
--- a/tests/api2/test_catalog_sync.py
+++ b/tests/api2/test_catalog_sync.py
@@ -6,7 +6,7 @@ import time
 from middlewared.client.client import ClientException, ValidationErrors
 from middlewared.test.integration.assets.catalog import catalog
 from middlewared.test.integration.assets.pool import another_pool
-from middlewared.test.integration.utils import call, ssh
+from middlewared.test.integration.utils import call, fail, ssh
 
 from auto_config import pool_name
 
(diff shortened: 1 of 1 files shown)


tests/api2/test_catalog_sync.py
```diff
@@ -37,12 +38,30 @@ def kubernetes_pool():
...
+            timeout = 150
+            while True:
+                time.sleep(5)
+                kube_system_pods = call(
+                    'k8s.pod.query', [
+                        ['metadata.namespace', '=', 'kube-system']
+                    ], {'select': ['metadata.name', 'status.phase']}
+                )
+                if len([pod for pod in kube_system_pods if pod['status']['phase'] == 'Running']) >= 3 or timeout <= 0:
+                    break
```
themylogin, line 50: I don't understand much what's going on, but do we yield the same result if there are 3 pods running or there are less but a timeout occurred? Shouldn't we stop the tests if a timeout occurs?

themylogin (approved):

PR #12893, opened 2024-01-15
NAS-126788 / 24.04 / Fix VM device attributes schema

Qubad786:
### Problem

VM device attributes are generic so `private=True` was not working as desired for VM display devices' password making it visible in plain text for READONLY role.

### Fix

PR fixes the problem by using OROperator for dynamic inclusion of VM device attributes.


anodos325:
The following is slightly outside of scope of your fix, but I think for consistency sake, the password should use the explicit `Password` schema class rather than `Str`.

sonicaj (approved):

yocalebo (approved):

PR #12894, opened 2024-01-15
NAS-126791 / 24.04 / Improve handling for SIDS S-1-22-1 and S-1-22-2

anodos325:
S-1-22-1 and S-1-22-2 are special-use prefixes for denoting local unix accounts. They shouldn't ever be explicitly set in SMB share ACLs and so this commit adds validation to prevent users from doing this. This commit has practical consequence of requiring that users specify "SMB" accounts for the SMB share ACL.

Idmap-related functions are also refactored to directly look up the local user account rather than passing these special SIDs through libwbclient. The reason for this is that currently such lookups through libwbclient may replace the idmap cache entry for a normal SID with the Unix User / Unix Group SID resulting in unexpected SID resolution results. During the refactor, legacy private methods that convert only a single SID or unix ID were removed. This has the advantage of using a single libwbclient request to service many lookups, signicantly reducing the number of requests made to the idmap plugin.

Samba-related utilies cache various things in the gencache.tdb file. The ability to manipulate this file's contents have historically been somewhat limited (requiring to run commands in a subprocess). This commit adds handling for this file via Samba's python tdb bindings. Currently implementation is limited to fetching and removing idmap- related entries (such as SID <-> XID and SID <-> name mappings), and flushing the cache. This allows more targeted cache manipulation, such as removing a single UID2SID mapping when deleting a user (or flushing out any stale UID2SID entries when adding a new user). This gencache plugin will eventually be expanded to allow removal of the libads server affinity cache entries (and similar AD-related caches) to improve error handling during domain join.


anodos325:
Issue that raised this rather large refactor is covered by the newly-added tests `test_account_idmap.py`. Remaining areas touched already have coverage via other tests (most prominently test_040_ad_user_group_cache.py and existing user / group tests). Directory services privileges are validated in test_030_activedirectory.py.

yocalebo (approved):
I see no obvious issues and have to trust the CI catches anything else.

PR #12901, opened 2024-01-16
NAS-125606 / 24.04 / Add validation to ensure we don't allow configuring a pool for k8s which is being used by sharing services

sonicaj:


src/middlewared/middlewared/common/attachment/__init__.py
```diff
@@ -138,10 +139,10 @@ async def restart_reload_services(self, attachments):
     async def remove_alert(self, attachment):
         await self.middleware.call(f'{self.namespace}.remove_locked_alert', attachment['id'])
 
-    async def is_child_of_path(self, resource, path, check_parent):
+    async def is_child_of_path(self, resource, path, check_parent, exact_match):
```
themylogin, line 142: Can you please add a comment that explains all the parameters of this method and how it should work? And a set of unit tests? It's getting complicated.

PR #12927, opened 2024-01-18
NAS-125941 / 24.10 / Remove OSC and platform related logic

Qubad786:
### Context

PR removes OSC module and platform related logic from middleware code-base since CORE and SCALE are built from their own respective branches now.


src/middlewared/middlewared/alembic/versions/12.0/2020-05-18_14-29_nmbclusters-fix.py
```diff
@@ -84,7 +85,7 @@ def upgrade():
     #   1. this is a freeBSD system
     #   2. kern.ipc.nmbclusters is a "sysctl" type entry in the db
     #        AND is < the current OS value
-    if osc.IS_FREEBSD:
+    if platform.system().upper() == 'FREEBSD':
```
themylogin, line 88: This will never be true, these migrations are always executed on Linux

themylogin (changes requested):

src/middlewared/middlewared/alembic/versions/12.0/2020-05-18_14-29_nmbclusters-fix.py
Qubad786: @themylogin should we remove the complete logic of the migration then?

themylogin (approved):

PR #12936, opened 2024-01-21
NAS-126899 / 24.10 / Add roles for crypto plugin

sonicaj:


anodos325:
Is there a use-case that requires differentiating between certificate and certificate authority roles?

sonicaj:
Could be, when someone is setting up their public key infrastructure on TrueNAS - they would not want all users with cert role to be able to create/remove CAs i guess as that can potentially disrupt the complete chain.

src/middlewared/middlewared/plugins/crypto_/certificate_authorities.py
```diff
@@ -36,6 +36,7 @@ class Config:
         datastore_extend_context = 'certificateauthority.cert_extend_context'
         datastore_prefix = 'cert_'
         cli_namespace = 'system.certificate.authority'
+        role_prefix = 'CERTIFICATEAUTHORITY'
```
themylogin, line 39: We can use `_` to make reading this easy

themylogin (approved):

PR #12944, opened 2024-01-22
NAS-126922 / 24.10 / properly handle bond/vlan interfaces in virtual_ips.py

yocalebo:


anodos325 (approved):
We should probably add this to a list of items that need CI on our HA VMs.

yocalebo:
> We should probably add this to a list of items that need CI on our HA VMs.

agree

PR #12949, opened 2024-01-22
NAS-126842 / 24.10 / Fix UPS reporting plugin

Qubad786:
# Problem
If the UPS service is stopped, Netdata will not start the UPS plugin. Even if the UPS service is restarted, Netdata won't start the plugin until Netdata itself is restarted. Additionally, there are bugs related to UPS identifiers, which can change and affect chart names.

# Solution

Modify the Netdata UPS plugin to send default data if the UPS service is not running. This ensures that Netdata continues to send default stats even when the UPS service is stopped - we handle this gracefully when user tries to retrieve UPS stats when no UPS is running by making sure we are not sharing the default stats. When the UPS service is started, Netdata will begin sending the actual UPS stats. Properly obtain UPS identifiers for use in graphs to ensure the correctness of UPS plugin chart names.

### DIFF
--- a/src/freenas/usr/lib/netdata/charts.d/nut_ups.chart.sh
+++ b/src/freenas/usr/lib/netdata/charts.d/nut_ups.chart.sh
@@ -1,4 +1,4 @@
-. /usr/lib/netdata/charts.d/nut.chart.sh
+source /usr/lib/netdata/charts.d/nut.chart.sh
 
 nut_get_all() {
   run -t $nut_timeout upsc -l || echo "ix-dummy-ups"
@@ -54,5 +54,5 @@ nut_ups_update() {
   # remember: KEEP IT SIMPLE AND SHORT
   nut_ups_check
   nut_ups_create
-  nut_update
+  nut_update $@
 }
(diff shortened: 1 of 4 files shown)


src/freenas/usr/lib/netdata/charts.d/nut_ups.chart.sh
```diff
@@ -0,0 +1,58 @@
+. /usr/lib/netdata/charts.d/nut.chart.sh
```
mgrimesix, line 1: For readability it's nice to use 'source' instead of '.'

src/freenas/usr/lib/netdata/charts.d/nut_ups.chart.sh
```diff
@@ -0,0 +1,58 @@
...
+nut_ups_update() {
+  # the first argument to this function is the microseconds since last update
+  # pass this parameter to the BEGIN statement (see below).
+
+  # do all the work to collect / calculate the values
+  # for each dimension
+  # remember: KEEP IT SIMPLE AND SHORT
+  nut_ups_check
+  nut_ups_create
+  nut_update
```
mgrimesix, line 57: Maybe I'm missing something, but I believe nut_update expects the parameters described in the comment?  If so, then maybe use: `nut_update $@`

mgrimesix (commented):
I'll DM about another issue regarding nut driver syslog spam.

mgrimesix (approved):
Thanks!

PR #12952, opened 2024-01-23
NAS-125699 / 24.04-RC.1 / Remove system birthday

themylogin:


yocalebo (approved):

yocalebo:
Be sure and let Evgeny know that this field was removed from `system.info`. I don't believe they're using it but better be safe and let them know.

themylogin:
time 0:15

PR #12967, opened 2024-01-24
NAS-126731 / 24.10 / Improve failover handling of a large number of targets

bmeagherix:
### Background
SCST uses a script `scstadmin` to load the config file into the kernel.  This done serially, so when large configurations are present it can be slow.

Further, iSCSI ALUA targets use the linux kernel `dlm` (distributed lock manager).  Configuration requires the interaction of the kernel and middleware on both nodes.  This can be slow, especially when many targets are present. 

### Overview
A couple of changes to SCST have been made to improve configuration/lifecycle performance (PRs #[19](https://github.com/truenas/scst/pull/19) and #[23](https://github.com/truenas/scst/pull/23)).  Also allowed `vdisk_fileio` to use the same active parameter as `vdisk_blockio` (PR #[26](https://github.com/truenas/scst/pull/26)) - more below.

This PR contains substantial changes being made to middleware to further improve configuration/lifecycle performance:

- On boot do **not** start with `cluster_mode` enabled on target extents.
- When the STANDBY node boots, it will initiate a job that resettles the DLM (on both nodes) and then steps thru the extents (20 at a time), configuring `cluster_mode`
- Define the fileio/blockio based extents in _/etc/scst.conf_ for BOTH nodes - albeit with `active 0` parameter set on the STANDBY node.
- On failover on the ACTIVE node, _become_active_ iscsitarget rather than _restart_ it.  This avoids running `scstadmin`, instead will make _precise_ changes to the scst /sys interface.  We only do the bare minimum that we need (makes extents active, and replace the HA target based LUNs with real extent based ones).
- On failover on the STANDBY node, start iscsitarget but with minimal config.  This will be corrected by the `iscsi.alua.standby_after_start` job.
- For inter-node HA targets tweak the `recovery_tmo` parameter to 10 seconds.  Otherwise we can expect stalls on a newly elected ACTIVE node, before it logs out the HA targets it needed when it was STANDBY.

We can't achieve everything instantly, so the above is an attempt to get the targets on the ACTIVE node available ASAP, with the STANDBY node to follow.

On the newly elected ACTIVE node we split activities into three stages, as certain aspects can be kicked off early/in parallel with other activities:
1. `iscsi.alua.active_elected` job will run as soon as we know we've won the election (i.e. have started fenced)
2. `iscsi.alua.activate_extents` job will run once we have imported the pool and mounted volumes, etc.
3. `service.become_active iscsitarget` will run instead of a `service.restart`

### DIFF
--- a/src/middlewared/middlewared/etc_files/scst.conf.mako
+++ b/src/middlewared/middlewared/etc_files/scst.conf.mako
@@ -296,7 +296,8 @@ TARGET_DRIVER iscsi {
 %>\
 %   if target['id'] in associated_targets:
 ##
-## For ALUA rel_tgt_id is tied to controller, if not ALUA don't bother writing it
+## For ALUA rel_tgt_id is tied to controller, if not ALUA write it anyway
+## to avoid it changing when ALUA is toggled.
 ##
 %       if alua_enabled:
 %           if node == "A":
@@ -305,6 +306,8 @@ TARGET_DRIVER iscsi {
 %           if node == "B":
         rel_tgt_id ${idx + 32000}
 %           endif
+%       else:
+        rel_tgt_id ${idx}
 %       endif
 ##
 ## For ALUA target is enabled if MASTER, disabled for BACKUP
--- a/src/middlewared/middlewared/etc_files/scst.conf.mako
+++ b/src/middlewared/middlewared/etc_files/scst.conf.mako
@@ -106,7 +106,7 @@
                     (cluster_mode_targets, cluster_mode_luns) = middleware.call_sync('failover.call_remote', 'iscsi.target.cluster_mode_targets_luns')
                 except CallError as e:
                     if e.errno != CallError.ENOMETHOD:
-                        raise
+                        raise FileShouldNotExist
                     # We may have upgraded one node, but not yet the other
                     middleware.logger.warning('Failed to configure remote cluster_mode_targets_luns')
                 clustered_extents = set(middleware.call_sync("iscsi.target.clustered_extents"))
--- a/src/middlewared/middlewared/etc_files/scst.conf.mako
+++ b/src/middlewared/middlewared/etc_files/scst.conf.mako
@@ -231,11 +231,13 @@ TARGET_DRIVER copy_manager {
 }
 % endif
 ##
-## Do NOT write "HANDLER vdisk_fileio" and "HANDLER vdisk_blockio" sections if
-## we are BACKUP node.
-% if failover_status != "BACKUP":
+
 % for handler in extents_io:
 HANDLER ${handler} {
+## Do NOT write *contents* of "HANDLER vdisk_fileio" and "HANDLER vdisk_blockio" sections if
+## we are BACKUP node.  However, we still want to write the empty sections so that we will
+## load the handlers on scst startup.  This will facilitate a fast BACKUP -> MASTER reload on failover.
+% if failover_status != "BACKUP":
 %   for extent in extents_io[handler]:
     DEVICE ${extent['name']} {
         filename ${extent['extent_path']}
@@ -264,9 +266,9 @@ HANDLER ${handler} {
     }
 
 %   endfor
+% endif
 }
 % endfor
-% endif
 
 TARGET_DRIVER iscsi {
     enabled 1
--- a/src/middlewared/middlewared/etc_files/scst.conf.mako
+++ b/src/middlewared/middlewared/etc_files/scst.conf.mako
@@ -140,9 +140,7 @@
             extent['extent_path'] = extent['path']
             extents_io_key = 'vdisk_fileio'
 
-        if not os.path.exists(extent['extent_path']):
-            if alua_enabled and failover_status == "BACKUP":
-                continue
+        if not alua_enabled and not os.path.exists(extent['extent_path']):
             middleware.logger.debug(
                 'Skipping generation of extent %r as the underlying resource does not exist', extent['name']
             )
@@ -234,10 +232,6 @@ TARGET_DRIVER copy_manager {
 
 % for handler in extents_io:
 HANDLER ${handler} {
-## Do NOT write *contents* of "HANDLER vdisk_fileio" and "HANDLER vdisk_blockio" sections if
-## we are BACKUP node.  However, we still want to write the empty sections so that we will
-## load the handlers on scst startup.  This will facilitate a fast BACKUP -> MASTER reload on failover.
-% if failover_status != "BACKUP":
 %   for extent in extents_io[handler]:
     DEVICE ${extent['name']} {
         filename ${extent['extent_path']}
@@ -263,10 +257,12 @@ HANDLER ${handler} {
         cluster_mode 0
 %       endif
 %       endif
+%       if failover_status == "BACKUP" and alua_enabled:
+        active 0
+%       endif
     }
 
 %   endfor
-% endif
 }
 % endfor
 
--- a/src/middlewared/middlewared/etc_files/scst.conf.mako
+++ b/src/middlewared/middlewared/etc_files/scst.conf.mako
@@ -5,6 +5,8 @@
     from collections import defaultdict
     from pathlib import Path
 
+    from middlewared.service import CallError
+
     global_config = middleware.call_sync('iscsi.global.config')
 
     def existing_copy_manager_luns():
@@ -35,6 +37,21 @@
                     start_count += 1
                 cml[start_count] = device
         return cml
+
+    targets = middleware.call_sync('iscsi.target.query')
+    extents = {d['id']: d for d in middleware.call_sync('iscsi.extent.query', [['enabled', '=', True]], {'extra': {'use_cached_locked_datasets': False}})}
+    portals = {d['id']: d for d in middleware.call_sync('iscsi.portal.query')}
+    initiators = {d['id']: d for d in middleware.call_sync('iscsi.initiator.query')}
+    authenticators = defaultdict(list)
+    for auth in middleware.call_sync('iscsi.auth.query'):
+        authenticators[auth['tag']].append(auth)
+
+    associated_targets = defaultdict(list)
+    for a_tgt in filter(
+        lambda a: a['extent'] in extents and not extents[a['extent']]['locked'],
+        middleware.call_sync('iscsi.targetextent.query')
+    ):
+        associated_targets[a_tgt['target']].append(a_tgt)
 
     # There are several changes that must occur if ALUA is enabled,
     # and these are different depending on whether this is the
@@ -67,6 +84,8 @@
                 if 'address' in addr:
                     failover_virtual_aliases.append(addr['address'])
 
+    cluster_mode_targets = []
+    cluster_mode_luns = {}
     if failover_status == "MASTER":
         middleware.call_sync("iscsi.target.logout_ha_targets")
         local_ip = middleware.call_sync("failover.local_ip")
@@ -74,27 +93,23 @@
     elif failover_status == "BACKUP":
         if alua_enabled:
             logged_in_targets = middleware.call_sync("iscsi.target.login_ha_targets")
-            cluster_mode_targets = middleware.call_sync('failover.call_remote', 'iscsi.target.cluster_mode_targets')
+            try:
+                (cluster_mode_targets, cluster_mode_luns) = middleware.call_sync('failover.call_remote', 'iscsi.target.cluster_mode_targets_luns')
+            except CallError as e:
+                if e.errno != CallError.ENOMETHOD:
+                    raise
+                # We may have upgraded one node, but not yet the other
+                middleware.logger.warning('Failed to configure remote cluster_mode_targets_luns')
         else:
             middleware.call_sync("iscsi.target.logout_ha_targets")
+            targets = []
+            extents = {}
+            portals = {}
+            initiators = {}
+            associated_targets = defaultdict(list)
 
     nodes = {"A" : {"other" : "B", "group_id" : 101},
              "B" : {"other" : "A", "group_id" : 102}}
-
-    targets = middleware.call_sync('iscsi.target.query')
-    extents = {d['id']: d for d in middleware.call_sync('iscsi.extent.query', [['enabled', '=', True]], {'extra': {'use_cached_locked_datasets': False}})}
-    portals = {d['id']: d for d in middleware.call_sync('iscsi.portal.query')}
-    initiators = {d['id']: d for d in middleware.call_sync('iscsi.initiator.query')}
-    authenticators = defaultdict(list)
-    for auth in middleware.call_sync('iscsi.auth.query'):
-        authenticators[auth['tag']].append(auth)
-
-    associated_targets = defaultdict(list)
-    for a_tgt in filter(
-        lambda a: a['extent'] in extents and not extents[a['extent']]['locked'],
-        middleware.call_sync('iscsi.targetextent.query')
-    ):
-        associated_targets[a_tgt['target']].append(a_tgt)
 
     # Let's map extents to respective ios
     all_extent_names = []
@@ -158,10 +173,12 @@ HANDLER dev_disk {
 %                 for device in devices:
 
         DEVICE ${device} {
-## Sanity check to ensure the MASTER target is in cluster_mode, if so then so are we
+## Sanity check to ensure the MASTER target LUN is in cluster_mode, if so then so are we
 ## Note we use a similar check to determine whether the target will be enabled.
-%                 if name in cluster_mode_targets:
+%                 if name in cluster_mode_luns and int(device.split(':')[-1]) in cluster_mode_luns[name]:
             cluster_mode 1
+%                 else:
+            cluster_mode 0
 %                 endif
         }
 %                 endfor
--- a/src/middlewared/middlewared/etc_files/scst.conf.mako
+++ b/src/middlewared/middlewared/etc_files/scst.conf.mako
@@ -134,7 +134,7 @@
     hosts_iqns = middleware.call_sync('iscsi.host.get_hosts_iqns')
 
     if alua_enabled and failover_status == "BACKUP":
-        cml = calc_copy_manager_luns(list(itertools.chain.from_iterable(logged_in_targets.values())), True)
+        cml = calc_copy_manager_luns(list(itertools.chain.from_iterable([x for x in logged_in_targets.values() if x is not None])), True)
     else:
         cml = calc_copy_manager_luns(all_extent_names)
 %>\
--- a/src/middlewared/middlewared/etc_files/scst.conf.mako
+++ b/src/middlewared/middlewared/etc_files/scst.conf.mako
@@ -71,7 +71,9 @@
     # - Write a DEVICE_GROUP section with two TARGET_GROUPs
     # - TARGET GROUPs and rel_tgt_id are tied to the controller,
     #   *not* to whether it is currently the MASTER or BACKUP
-    #
+    # - clustered_extents is used to prevent cluster_mode from being
+    #   enabled on entents at startup.  We will have to explicitly
+    #   write 1 to cluster_mode elsewhere.
     is_ha = middleware.call_sync('failover.licensed')
     alua_enabled = middleware.call_sync("iscsi.global.alua_enabled")
     failover_status = middleware.call_sync("failover.status")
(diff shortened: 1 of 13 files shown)


src/middlewared/middlewared/etc_files/scst.conf.mako
```diff
@@ -67,35 +84,31 @@
...
+                (cluster_mode_targets, cluster_mode_luns) = middleware.call_sync('failover.call_remote', 'iscsi.target.cluster_mode_targets_luns')
+            except CallError as e:
+                if e.errno != CallError.ENOMETHOD:
+                    raise
+                # We may have upgraded one node, but not yet the other
+                middleware.logger.warning('Failed to configure remote cluster_mode_targets_luns')
         else:
             middleware.call_sync("iscsi.target.logout_ha_targets")
+            targets = []
+            extents = portals = initiators = {}
```
yocalebo, line 106: Just a note here. These all point to the same immutable dictionary object. This means if a change is made to `extents`, it'll also be reflected in `portals` and `initiators`
```
>>> a = b = c = {}
>>> a['key'] = 'value'
>>> 
>>> a
{'key': 'value'}
>>> b
{'key': 'value'}
>>> c
{'key': 'value'}
```
If this is expected, then ignore it 😄

bmeagherix: These values should not be modified after this point, but maybe it'll still be better to change just to avoid possible fragility.  Thanks.

bmeagherix:
@anodos325 and @yocalebo I've removed WIP label and requested you review, though there _is_ still more testing I'd like to do.

src/middlewared/middlewared/plugins/iscsi_/alua.py
```diff
@@ -0,0 +1,390 @@
...
+    To avoid this, SCST on the ACTIVE will start without cluster_mode being
+    set on extents.  Likewise on the STANDBY node, so the targets there will be
+    present but disabled.  However, the STANDBY will then initiate a job to
+    (gradually) enable cluster_mode on the ACTIVE and react.
+    """
+    class Config:
+        private = True
+        namespace = 'iscsi.alua'
+
+    # See HA_PROPAGATE in event.py.  Not only required when running command
```
yocalebo, line 33: Another comment that I can't quite grok

src/middlewared/middlewared/plugins/iscsi_/alua.py
```diff
@@ -0,0 +1,390 @@
...
+                await self.middleware.call('iscsi.alua.standby_after_start')
+
+    async def before_stop(self):
+        self._all_cluster_mode = False
+        self.standby_starting = False
+
+    async def standby_enable_devices(self, devices):
+        await self.middleware.call('iscsi.target.login_ha_targets')
+        extents = await self.middleware.call('iscsi.extent.logged_in_extents')
+        asked = set(devices)
```
yocalebo, line 71: The `asked` variable isn't used anywhere?

src/middlewared/middlewared/plugins/iscsi_/alua.py
```diff
@@ -0,0 +1,390 @@
...
+        job.set_progress(20, 'Read to activate')
+
+        if todo:
+            self.logger.debug('Activating extents')
+            retries = 10
+            while todo and retries:
+                do_again = []
+                for item in todo:
+                    # Mark them active
+                    # self.logger.debug(f'Activating extent {item}')
```
yocalebo, line 145: Remove this comment?

src/middlewared/middlewared/plugins/failover_/event.py
```diff
@@ -57,6 +56,10 @@ class Config:
     # before the other services during a failover event
     CRITICAL_SERVICES = ['iscsitarget', 'cifs', 'nfs']
 
+    # Any members of the FAILOVER_SERVICES set will be have
```
yocalebo, line 59: I can't grok this comment so I'm not sure why this new variable is needed?

src/middlewared/middlewared/plugins/service_/services/iscsitarget.py
```diff
@@ -6,19 +8,56 @@
...
             ["scstadmin", "-noprompt", "-force", "-config", "/etc/scst.conf"], check=False
         )).returncode == 0
+
+    async def failover(self):
+        if await self.middleware.call('iscsi.global.alua_enabled'):
+            if await self.middleware.call('iscsi.scst.is_kernel_module_loaded'):
+                if await self.middleware.call("failover.status") == "MASTER":
+                    try:
+                        return await self.middleware.call("iscsi.alua.failover_to_master")
+                    except Exception as e:
```
yocalebo, line 59: You don't need `e` and `exc_info=True`. Remove the `e` variable and leave `exc_info=True`

src/middlewared/middlewared/plugins/dlm.py
```diff
@@ -272,6 +279,46 @@ async def remote_down(self):
         """
         self.logger.info('Remote node %s down', self.peernodeID)
 
+    @private
+    async def local_remove_peer(self, lockspace_name):
```
anodos325, line 283: We should add some docstrings explaining new methods being added.

src/middlewared/middlewared/plugins/iscsi_/targets.py
```diff
@@ -622,3 +628,66 @@ async def cluster_mode_targets(self):
...
+                    cluster_mode_luns[target['name']].append(lunid)
+
+        return (cluster_mode_targets, cluster_mode_luns)
+
+    @private
+    async def active_targets(self):
+        """
+        Returns the names of all targets whose extents are neither disabled nor locked.
+        """
+        bad_extents = [extent['id'] for extent in await self.middleware.call('iscsi.extent.query',
```
yocalebo, line 678: Can we please fix the formatting of this line. It's too messy at this point. Something like
 ```
 bad_extents = []
 filters = [['OR', [['enabled', '=', False], ['locked', '=', True]]]]
 for extent in await self.middleware.call('iscsi.extent.query', filters):
    bad_extents.append(extent['id'])
```

File: src/middlewared/middlewared/plugins/iscsi_/alua.py [hunk 9]
```diff
@@ -0,0 +1,390 @@
...
+    between the two nodes in a HA pair.  This is performed in response to
+    cluster_mode being set on target extents.
+
+    However, when a LARGE number of extents(/targets) are present it becomes
+    impractical to leave/enter lockspaces on scst startup.
+
+    To avoid this, SCST on the ACTIVE will start without cluster_mode being
+    set on extents.  Likewise on the STANDBY node, so the targets there will be
+    present but disabled.  However, the STANDBY will then initiate a job to
+    (gradually) enable cluster_mode on the ACTIVE and react.
```
anodos325, line 27: We should probably have explicit checks within the methods to ensure that failover.status is the expected value when starting jobs that should only be on active or standby controller.

src/middlewared/middlewared/plugins/service.py
```diff
@@ -405,6 +405,16 @@ async def identify_process(self, procname):
                 if await service.identify(procname):
                     return service_name
 
+    @private
+    async def get_unit_state(self, service):
+        service_object = await self.middleware.call('service.object', service)
+        return await service_object.get_unit_state()
+
+    @private
+    async def failover(self, service):
```
anodos325, line 414: It would probably be less race-prone if we directly passed in argument for new failover status.

src/middlewared/middlewared/plugins/iscsi_/scst.py
```diff
@@ -0,0 +1,95 @@
...
+            return False
+
+    def is_kernel_module_loaded(self):
+        return pathlib.Path(SCST_BASE).exists()
+
+    def activate_extent(self, extent_name, extenttype, path):
+        p = pathlib.Path(path)
+        if p.exists():
+            if extenttype == 'DISK':
+                p = pathlib.Path(f'/sys/kernel/scst_tgt/handlers/vdisk_blockio/{extent_name}/active')
```
yocalebo, line 77: Why not define the path once and use `SCST_BASE`?
```
path = f'{SCST_BASE}/handlers/vdisk_blockio'
if extenttype == 'FILE':
    path = f'{SCST_BASE}/handlers/vdisk_fileio'
pathlib.Path(f'{path}/{extent_name}/active').write_text('1\n')

File: src/middlewared/middlewared/plugins/iscsi_/scst.py [hunk 12]
```diff
@@ -0,0 +1,95 @@
...
+                p = pathlib.Path(f'/sys/kernel/scst_tgt/handlers/vdisk_fileio/{extent_name}/active')
+            p.write_text('1\n')
+            return True
+        else:
+            return False
+
+    def activate_extents(self, extents):
+        for extent in extents:
+            if extent['type'] == 'DISK':
+                p = pathlib.Path(f'/sys/kernel/scst_tgt/handlers/vdisk_blockio/{extent["name"]}/active')
```
yocalebo, line 88: Same here in the other methods. Should use `SCST_BASE`

src/middlewared/middlewared/plugins/service_/services/iscsitarget.py
```diff
@@ -6,19 +8,56 @@
...
 
     async def reload(self):
         return (await run(
             ["scstadmin", "-noprompt", "-force", "-config", "/etc/scst.conf"], check=False
         )).returncode == 0
+
+    async def failover(self):
+        if await self.middleware.call('iscsi.global.alua_enabled'):
+            if await self.middleware.call('iscsi.scst.is_kernel_module_loaded'):
+                if await self.middleware.call("failover.status") == "MASTER":
```
yocalebo, line 56: I don't understand the need to call `failover.status` here? If we're in the `vrrp_master` or `vrrp_backup` methods, then we don't need to check `failover.status` since it's safe to assume we're either the master/backup controller appropriately.

src/middlewared/middlewared/plugins/iscsi_/scst.py
```diff
@@ -0,0 +1,95 @@
...
+        return True
+
+    async def set_devices_cluster_mode(self, devices, value):
+        text = f'{int(value)}\n'
+        paths = [f'{SCST_DEVICES}/{device}/cluster_mode' for device in devices]
+        if paths:
+            await asyncio.gather(*[self.middleware.call('iscsi.scst.path_write', path, text) for path in paths])
+
+    def disable(self):
+        p = pathlib.Path(SCST_TARGETS_ISCSI_ENABLED_PATH)
```
yocalebo, line 52: No reason to create a variable inside these functions. Can just do
```
pathlib.Path(SCST_TARGETS_ISCSI_ENABLED_PATH).write(text\n)
```

src/middlewared/middlewared/plugins/iscsi_/extents.py
```diff
@@ -459,3 +460,47 @@ async def remove_extent_file(self, data):
...
+            assoc
+        ):
+            target_id = a_tgt['target']
+            extent_name = extents[a_tgt['extent']]['name']
+            target_luns[target_id].add((a_tgt['lunid'], extent_name))
+
+        global_basename = (await self.middleware.call('iscsi.global.config'))['basename']
+        ha_basename = f'{global_basename}:HA:'
+
+        for iqn in iqns:
```
yocalebo, line 495:
```
for iqn in filter(lambda x: x.startswith(ha_basename):
    target_name = iqn.split(....
```

src/middlewared/middlewared/plugins/iscsi_/alua.py
```diff
@@ -0,0 +1,390 @@
...
+                self.logger.warning('standby_start job: %r', e, exc_info=True)
+                await asyncio.sleep(RETRY_SECONDS)
+        if not self.standby_starting:
+            job.set_progress(22, 'Abandoned job.')
+            return
+        else:
+            job.set_progress(22, 'Remote iscsitarget is active')
+
+        # Next login the HA targets.
+        reload_remote_retries = 5
```
anodos325, line 275: Should this be a constant?

src/middlewared/middlewared/plugins/iscsi_/alua.py
```diff
@@ -0,0 +1,390 @@
...
+                devices = list(itertools.chain.from_iterable([x for x in after_iqns.values() if x is not None]))
+                if await self.middleware.call('iscsi.scst.check_cluster_mode_paths_present', devices):
+                    break
+
+                self.logger.debug('Detected missing cluster_mode.  Retrying.')
+                await self.middleware.call('iscsi.target.logout_ha_targets')
+                await self.middleware.call('service.reload', 'iscsitarget')
+                job.set_progress(20, 'Logged out HA targets (local node)')
+            except Exception as e:
+                self.logger.warning('Failed to login and surface HA targets: %r', e, exc_info=True)
```
anodos325, line 321: We can probably just go with `exc_info=True` and not duplicate info.

src/middlewared/middlewared/plugins/iscsi_/alua.py
```diff
@@ -0,0 +1,390 @@
...
+                                remote_requires_reload = True
+                                await self.middleware.call('failover.call_remote', 'iscsi.scst.set_devices_cluster_mode', [list(todo_remote), 1])
+                            else:
+                                break
+                        except Exception:
+                            await asyncio.sleep(RETRY_SECONDS)
+
+                    # Enable on STANDBY.  If we fail here, we'll still go back around the main loop.
+                    ok = False
+                    enable_retries = 10
```
anodos325, line 349: Should this be a constant?

src/middlewared/middlewared/plugins/iscsi_/alua.py
```diff
@@ -0,0 +1,390 @@
...
+                    # Update progress
+                    self.enabled.update(to_enable)
+                    progress = 20 + (80 * (len(self.enabled) / len(extents)))
+                    job.set_progress(progress, f'Enabled {len(self.enabled)} extents')
+                    self.logger.info('Set cluster_mode on for %r extents', len(self.enabled))
+                else:
+                    break
+            except Exception as e:
+                # This is a fail-safe exception catch.  Should never occur.
+                self.logger.warning('standby_start job: %r', e, exc_info=True)
```
anodos325, line 369: Ditto about `exc_info=True`

src/middlewared/middlewared/etc_files/scst.conf.mako
```diff
@@ -67,35 +86,42 @@
...
-            cluster_mode_targets = middleware.call_sync('failover.call_remote', 'iscsi.target.cluster_mode_targets')
+            if middleware.call_sync("iscsi.alua.standby_write_empty_config"):
+                logged_in_targets = {}
+            else:
+                logged_in_targets = middleware.call_sync("iscsi.target.login_ha_targets")
+                try:
+                    (cluster_mode_targets, cluster_mode_luns) = middleware.call_sync('failover.call_remote', 'iscsi.target.cluster_mode_targets_luns')
+                except CallError as e:
+                    if e.errno != CallError.ENOMETHOD:
+                        raise
```
anodos325, line 109: Do we want to nuke the config file in this case (raise FileShouldNotExist)?

src/middlewared/middlewared/plugins/iscsi_/alua.py
bmeagherix: Used on next line.

src/middlewared/middlewared/plugins/iscsi_/scst.py
bmeagherix: I'll update somewhat, but the path in the parameter is **different**.  It's the path of the surfaced file/device.

src/middlewared/middlewared/plugins/service_/services/iscsitarget.py
bmeagherix: Per our discussion will replace with `become_active`.

src/middlewared/middlewared/plugins/iscsi_/alua.py
bmeagherix: Both `iscsi.alua.active_elected` and `iscsi.alua.activate_extents` are launched from `vrrp_master` before the `failover.status_refresh` call is made.  So, not really an option.

anodos325 (approved):
Looks good to me. Please wait for Caleb to also sign off before merge.

src/middlewared/middlewared/etc_files/scst.conf.mako
```diff
@@ -67,35 +86,42 @@
...
-            cluster_mode_targets = middleware.call_sync('failover.call_remote', 'iscsi.target.cluster_mode_targets')
+            if middleware.call_sync("iscsi.alua.standby_write_empty_config"):
+                logged_in_targets = {}
+            else:
+                logged_in_targets = middleware.call_sync("iscsi.target.login_ha_targets")
+                try:
+                    (cluster_mode_targets, cluster_mode_luns) = middleware.call_sync('failover.call_remote', 'iscsi.target.cluster_mode_targets_luns')
+                except CallError as e:
+                    if e.errno != CallError.ENOMETHOD:
+                        raise FileShouldNotExist
```
yocalebo, line 109: This means that if the other node is down for whatever reason, the entire scst.conf won't be generated since you're raising FileShouldNotExist. Is that desired?

yocalebo (approved):

yocalebo (approved):

bmeagherix:
@yocalebo I repushed commits to tip in an effort to allow the build to complete, but I did **not** modify them in _any_ way.  Unfortunately, the build is still currently failing wrt `cryptit`, but presumably this will just require an update of the image used - so orthogonal to this PR.

**Anyway**, FWIW the last 6 (six) commits were added in response to testing since you last approved this PR.  (I didn't want the above force-push to obscure this fact for you.)

src/middlewared/middlewared/etc_files/scst.conf.mako
bmeagherix: No, not desired.  Modified.

src/middlewared/middlewared/plugins/iscsi_/alua.py
```diff
@@ -535,3 +535,49 @@ async def removed_target_extent(self, target_id, extent_id):
         # etc, but (currently) these don't work.  Therefore we'll use a sledgehammer
         await self.middleware.call('failover.call_remote', 'iscsi.target.logout_ha_target', [targetname])
         await self.middleware.call('failover.call_remote', 'service.reload', ['iscsitarget'])
+
+    async def has_active_jobs(self):
```
yocalebo, line 539: This can be simplified by checking the `state` attribute of the job. i.e.
```
running_jobs = await self.middleware.call(
    'core.get_jobs', [
        ('method', 'in', [
            'iscsi.alua.active_elected',
            'iscsi.alua.activate_extents',
            'iscsi.alua.standby_after_start',
            'iscsi.alua.standby_reload',
            'iscsi.alua.standby_fix_cluster_mode',
        ]),
        ('state', '=', 'RUNNING'),
    ]
)
return bool(running_jobs)

File: src/middlewared/middlewared/plugins/iscsi_/alua.py [hunk 23]
```diff
@@ -399,5 +406,132 @@ async def standby_after_start(self, job):
 
         job.set_progress(100, 'All targets in cluster_mode')
         self.standby_starting = False
-        self._all_cluster_mode = True
         self.logger.debug('ALUA started on STANDBY')
+
+    @job(lock='standby_reload', transient=True)
+    async def standby_reload(self, job):
```
yocalebo, line 412: This is not a standby_reload function. If this gets called on the active controller, it will reload the service on this node.

Furthermore, the indiscriminate 30 second sleep causes me pause. Either way, the name of this method doesn't reflect reality unless I'm missing something obvious.

bmeagherix: This _only_ gets called when the `standby_node_requires_reload` flag gets set (because of a failure to communicate as expected with the other node).  And that flag only gets set in code protected by a check of `failover.status`.

That said it **certainly would** make sense to add a further check for `failover.status` BACKUP inside the function, immediately before we do the reload.

Wrt the 30 sec timer, well the two conditions we want to be able to recover from are:
1. Failure in `iscsi.target.login_ha_targets` (because the targets are not available yet from the ACTIVE node)
2. Failure in calling (`failover.call_remote`', '`iscsi.target.cluster_mode_targets_luns`')

Because this code is supposed to be a last-ditch fixup/recovery of an error path, I'm not sure if we can get more surgical in applying it.  Do you have a suggestion?

yocalebo (approved):

yocalebo (approved):

PR #12970, opened 2024-01-24
NAS-126080 / 24.10 / Fix pagination parameters

Qubad786:
## Problem

`offset` and `limit` parameters were being applied twice, one at the sql level and the then again at middleware level.

## Fix

PR fixes the problem by making sure this is not done at database level keeping in line with some other parameters we remove as well already.

### DIFF
--- a/src/middlewared/middlewared/service/crud_service.py
+++ b/src/middlewared/middlewared/service/crud_service.py
@@ -13,6 +13,9 @@ from .service import Service
 from .service_mixin import ServiceChangeMixin
 
 
+PAGINATION_OPTS = ('count', 'get', 'limit', 'offset', 'select')
+
+
 def get_datastore_primary_key_schema(klass):
     return convert_schema({
         'type': klass._config.datastore_primary_key_type,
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/service/crud_service.py
```diff
@@ -152,9 +152,8 @@ async def query(self, filters, options):
         # for filters for performance reasons.
         if not options['force_sql_filters'] and options['extend']:
             datastore_options = options.copy()
-            datastore_options.pop('count', None)
-            datastore_options.pop('get', None)
-            datastore_options.pop('select', None)
+            for option in ['count', 'get', 'limit', 'offset', 'select']:
```
yocalebo, line 155: Should we make a global object like
```
PAGINATION_OPTS = ('count', 'get', 'limit', 'offset', 'select')
```

And use it in this loop instead? Ideally the global variable might help us catch this in the future? Furthermore, we should probably add some tests for these.

themylogin (changes requested):
Agreed on the tests part

yocalebo:
@themylogin ping

themylogin (approved):

yocalebo (approved):

PR #12976, opened 2024-01-25
NAS-126980 / 24.10 / Fix `RuntimeError: Method ipmi.lan.query is public but has no @accept…

themylogin:
…s()`


themylogin:
time 0:30

yocalebo (approved):

PR #12980, opened 2024-01-25
NAS-126984 / 24.10 / Update the error message with a helpful hint.

mgrimesix:
When adding the server to Active Directory, all registered nameservers must also be part of the Active Directory.  This change adds this hint to the error message.


src/middlewared/middlewared/plugins/activedirectory_/dns.py
```diff
@@ -228,7 +228,8 @@ def get_host(srv_prefix):
                     raise CallError(
                         f'{name}: Nameserver {entry["nameserver"]} failed to resolve SRV '
                         f'record for domain {domain}. This may indicate a DNS misconfiguration '
-                        'on the TrueNAS server.',
+                        'on the TrueNAS server. NOTE: When configuring with Active Directory, all '
+                        'registered nameservers must be part of the Active Directory.',
```
anodos325, line 232: must be nameservers for the Active Directory domain.

anodos325 (approved):

PR #12985, opened 2024-01-25
NAS-126988 / 24.10 / Fix edge case in container pruning test

Qubad786:
## Problem
In the test `test_pruning_for_deleted_chart_release_images`, an assertion error still occurs sometimes even after waiting for a container to stop using its ID. After investigation, it appears that if a container of a different ID from the same image is still running, it can also cause this test to fail. This can happen if the container is still running from a previous test.

## Solution
Wait for the container to terminate in the previous test `test_pruning_for_deleted_chart_release_images`.

### DIFF
--- a/tests/api2/test_023_kubernetes.py
+++ b/tests/api2/test_023_kubernetes.py
@@ -4,7 +4,7 @@ import time
 from pytest_dependency import depends
 from functions import GET, PUT, wait_on_job
 from auto_config import ha, pool_name, interface, ip
-from middlewared.test.integration.utils import call, fail
+from middlewared.test.integration.utils import call
 
 
 # Read all the test below only on non-HA
@@ -82,7 +82,7 @@ if not ha:
                     break
 
             if timeout <= 0:
-                fail('Time to setup kubernetes exceeded 150 seconds')
+                pytest.fail('Time to setup kubernetes exceeded 150 seconds')
 
             timeout -= 5
 
--- a/tests/api2/test_024_container.py
+++ b/tests/api2/test_024_container.py
@@ -5,7 +5,7 @@ import os
 import pytest
 import sys
 import time
-from middlewared.test.integration.utils import call, fail
+from middlewared.test.integration.utils import call
 from pytest_dependency import depends
 # from pytest_dependency import depends
 apifolder = os.getcwd()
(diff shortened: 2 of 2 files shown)


yocalebo (approved):

tests/api2/test_024_container.py
```diff
@@ -49,6 +49,19 @@ def chart_release(app_name, values):
...
+    def wait_for_containerd(container_id):
+        timeout = 60
+        while True:
+            if container_id not in [container['id'] for container in call('docker.container.query')]:
+                break
+
+            time.sleep(10)
+            timeout -= 10
+            if timeout <= 0:
+                fail(f'{container_id!r} container was not deleted in 60 seconds')
```
yocalebo, line 62: Wait, we probably don't want to call `fail()` here since that will exit the _entire_ CI run. We just need to mark all tests that depend on this container being deleted to be skipped.

yocalebo (changes requested):

tests/api2/test_024_container.py
Qubad786: @yocalebo it won't fail the entire CI run and just the test in question https://github.com/truenas/middleware/blob/472ca363161b0c2c7d14065847d567027933e613/src/middlewared/middlewared/test/integration/utils/pytest.py#L6

themylogin: @Qubad786 it will fail the whole test suite exactly as the comment says

Qubad786: This was already being used and i assumed it was to fail the test in question..apologies did not go through the comment.
Fixing it and it's usages ✅

tests/api2/test_023_kubernetes.py
```diff
@@ -82,7 +82,7 @@ def test_07_kubernetes_pods_stats(request):
                     break
 
             if timeout <= 0:
-                fail('Time to setup kubernetes exceeded 150 seconds')
+                pytest.fail('Time to setup kubernetes exceeded 150 seconds')
```
themylogin, line 85: `pytest.fail` should not be used in our code as it can lead to effects described in `middlewared.test.integration.utils.fail` comment

We should fail the entire dependency (`setup_kubernetes` in this case) to prevent k3s tests from running

tests/api2/test_024_container.py
```diff
@@ -49,6 +49,19 @@ def chart_release(app_name, values):
...
+    def wait_for_containerd(container_id):
+        timeout = 60
+        while True:
+            if container_id not in [container['id'] for container in call('docker.container.query')]:
+                break
+
+            time.sleep(10)
+            timeout -= 10
+            if timeout <= 0:
+                pytest.fail(f'{container_id!r} container was not deleted in 60 seconds')
```
themylogin, line 62: Same here

themylogin (approved):
Sorry for misunderstanding, if you only want to fail one test then this is the way to go

PR #12992, opened 2024-01-25
NAS-126972 / 24.10 / fix low sel space alert

yocalebo:
Platform team has advised that these platforms have a FIFO for SEL (meaning it never "runs out of space").

### DIFF
--- a/src/middlewared/middlewared/alert/source/ipmi_sel.py
+++ b/src/middlewared/middlewared/alert/source/ipmi_sel.py
@@ -162,6 +162,14 @@ class IPMISELAlertSource(AlertSource):
 
         alerts = []
         alerts.extend(await self.produce_sel_elist_alerts())
+
+        platform = await self.middleware.call('truenas.get_chassis_hardware')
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/alert/source/ipmi_sel.py
```diff
@@ -162,6 +162,14 @@ async def check(self):
 
         alerts = []
         alerts.extend(await self.produce_sel_elist_alerts())
+
+        platform = await self.middleware.call('truenas.get_chassis_hardware')
+        if platform.startswith(('TRUENAS-F', 'TRUENAS-H', 'TRUENAS-R30'))
```
bmeagherix, line 167: Missing colon?

bmeagherix (changes requested):

themylogin (approved):

src/middlewared/middlewared/alert/source/ipmi_sel.py
yocalebo: thanks 🙁

bmeagherix (approved):

PR #13003, opened 2024-01-26
NAS-127008 / 24.10 / Do not use interface rename logic on systems that have NICs with dupl…

themylogin:
…icate hardware link addresses

### DIFF
--- a/src/middlewared/middlewared/plugins/interface/link_address.py
+++ b/src/middlewared/middlewared/plugins/interface/link_address.py
@@ -100,9 +100,16 @@ class DatabaseInterfaceCollection(InterfaceCollection):
 
 
 class RealInterfaceCollection(InterfaceCollection):
-    @property
-    def by_link_address(self):
-        return {i["state"]["hardware_link_address"]: i for i in self.interfaces}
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/interface/link_address.py
```diff
@@ -100,9 +100,16 @@ def get_name(self, i):
...
-    def by_link_address(self):
-        return {i["state"]["hardware_link_address"]: i for i in self.interfaces}
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.by_link_address = {}
+        for i in self.interfaces:
+            link_address = i["state"]["hardware_link_address"]
+            if link_address in self.by_link_address:
+                raise ValueError(f"Interfaces {self.by_link_address[link_address]['name']!r} and {i['name']!r} have "
```
yocalebo, line 110: Let's raise a custom exception class and catch it specifically and log a message that caters to this scenario.

yocalebo (changes requested):

themylogin:
backport

PR #13009, opened 2024-01-27
Disable AppArmor via grub

Jip-Hop:
Fixes https://ixsystems.atlassian.net/browse/NAS-127023


yocalebo:
As discussed in the ticket, we're not going to move forward with this change. Going to close this for now. Thanks as always!

PR #13014, opened 2024-01-29
NAS-126857 / 24.10 / Prevent combination of DISCARD aclmode and NFSv4 acltype

anodos325:
Setting a DISCARD aclmode when NFSv4 acltype is in use. The DISCARD aclmode causes ZFS to strip ACL whenever chmod is called on a file. This may lead to security incidents if administrator sets the parameter without actually knowing what it does.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/dataset.py
+++ b/src/middlewared/middlewared/plugins/pool_/dataset.py
@@ -289,18 +289,35 @@ class PoolDatasetService(CRUDService):
 
         dataset_pool_is_draid = await self.middleware.call('pool.is_draid_pool', parent['pool'])
         if data['type'] == 'FILESYSTEM':
-            if data.get('acltype', 'INHERIT') != 'INHERIT' or data.get('aclmode', 'INHERIT') != 'INHERIT':
-                to_check = data.copy()
-                check_ds = cur_dataset if mode == 'UPDATE' else parent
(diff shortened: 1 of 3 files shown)


sonicaj (approved):

src/middlewared/middlewared/plugins/pool_/dataset.py
```diff
@@ -302,6 +302,9 @@ async def __common_validation(self, verrors, schema, data, mode, parent=None, cu
                 if acltype in ['POSIX', 'OFF'] and to_check.get('aclmode', 'DISCARD') != 'DISCARD':
                     verrors.add(f'{schema}.aclmode', 'Must be set to DISCARD when acltype is POSIX or OFF')
 
+                elif acltype == 'NFSV4' and to_check.get('aclmode') == 'DISCARD':
```
sonicaj, line 305: Can `aclmode` be not specified ? And should we have a default like we do in 302 ?

sonicaj (approved):
Should we change the above `to_check.get` usage as well ?
Seems weird that at one place we check safely and at another we are certain it is there

anodos325:
> Should we change the above `to_check.get` usage as well ? Seems weird that at one place we check safely and at another we are certain it is there

Yeah, I'll do that.

yocalebo (approved):

PR #13015, opened 2024-01-29
NAS-127053 / 24.10 / Add max length validation to the support ticket title and description

themylogin:
QE team found this issue

### DIFF
--- a/tests/api2/test_charts_schema.py
+++ b/tests/api2/test_charts_schema.py
@@ -49,5 +49,5 @@ def test_text_schema_max_length(request):
                 'values': {'testtext': secrets.token_hex(2 * 1024 * 1024 // 2)},
             }):
                 assert ve.value.errors[0].errmsg == (
-                    'values.testtext: Value greater than 1048576 not allowed'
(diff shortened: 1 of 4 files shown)


tests/api2/test_charts_schema.py
```diff
@@ -49,5 +49,5 @@ def test_text_schema_max_length(request):
                 'values': {'testtext': secrets.token_hex(2 * 1024 * 1024 // 2)},
             }):
                 assert ve.value.errors[0].errmsg == (
-                    'values.testtext: Value greater than 1048576 not allowed'
+                    'values.testtext: The value may not be longer than 1048576 characters'
```
themylogin, line 52: @sonicaj this message here comes from the middleware, right? Thought it might as well come from catalog_validation, but wasn't able to find that code there.

sonicaj: @themylogin it indeed comes from middleware

sonicaj (approved):

themylogin:
time 0:10

PR #13018, opened 2024-01-29
NAS-125816 / 24.10 / Properly save nvram file for UEFI based guests

Qubad786:
### Problem
During installation, some operating systems like Debian modify the nvram file, enabling the VM to boot up. However, when we upgrade the host system, this file is removed. QEMU then recreates the file, restoring it to the default value. Consequently, after the upgrade, the VM is unable to boot up again.
### Solution
Ensure the persistence of nvram fd files after an upgrade by saving the under discussion files at `/data/subsystems/vm/nvram` and using them later after upgrade from the dir.

### DIFF
--- a/src/middlewared/middlewared/migration/0010_nvram_attr_vms.py
+++ b/src/middlewared/middlewared/migration/0010_nvram_attr_vms.py
@@ -1,0 +1,38 @@
+import os
+import shutil
+
+from middlewared.plugins.vm.utils import SYSTEM_NVRAM_FOLDER_PATH, get_vm_nvram_file_name
+
+
+DEFAULT_NVRAM_FOLDER_PATH = '/var/lib/libvirt/qemu/nvram'
+LIBVIRT_QEMU_UID = 64055
+LIBVIRT_QEMU_GID = 64055
+
+
+def migrate(middleware):
+    os.makedirs(SYSTEM_NVRAM_FOLDER_PATH, exist_ok=True)
+    os.chown(SYSTEM_NVRAM_FOLDER_PATH, LIBVIRT_QEMU_UID, LIBVIRT_QEMU_GID)
+
+    if middleware.call_sync('system.is_ha_capable'):
+        middleware.logger.debug('Skipping nvram migration as system is HA capable')
+        return
+
+    for vm in middleware.call_sync('vm.query', [['bootloader', '=', 'UEFI']]):
+        try:
+            migrate_vm_nvram_file(middleware, vm)
+        except Exception:
+            middleware.logger.error('Failed to migrate nvram file for VM %r(%r)', vm['name'], vm['id'], exc_info=True)
+
+
+def migrate_vm_nvram_file(middleware, vm):
+    file_name = get_vm_nvram_file_name(vm)
+    new_path = os.path.join(SYSTEM_NVRAM_FOLDER_PATH, file_name)
+    to_copy_path = os.path.join(DEFAULT_NVRAM_FOLDER_PATH, file_name)
+    if os.path.exists(to_copy_path):
+        shutil.copy2(to_copy_path, new_path)
+        os.chown(new_path, LIBVIRT_QEMU_UID, LIBVIRT_QEMU_GID)
+    else:
+        # File does not exist for us to copy, so we need to just log it
+        middleware.logger.debug(
+            'No nvram file found for VM %r(%r), hence setting it up with %r', vm['name'], vm['id'], new_path
+        )
--- a/src/middlewared/middlewared/plugins/vm/utils.py
+++ b/src/middlewared/middlewared/plugins/vm/utils.py
@@ -4,6 +4,7 @@ from xml.etree import ElementTree as etree
 
 
 ACTIVE_STATES = ['RUNNING', 'SUSPENDED']
+SYSTEM_NVRAM_FOLDER_PATH = '/data/subsystems/vm/nvram'
 LIBVIRT_URI = 'qemu+unix:///system?socket=/run/truenas_libvirt/libvirt-sock'
 LIBVIRT_USER = 'libvirt-qemu'
 NGINX_PREFIX = '/vm/display'
@@ -33,3 +34,7 @@ def get_pci_device_class(pci_path: str) -> str:
             return r.read().strip()
 
     return ''
+
+
+def get_vm_nvram_file_name(vm_data: dict) -> str:
+    return f'{vm_data["id"]}_{vm_data["name"]}_VARS.fd'
--- a/src/middlewared/middlewared/plugins/vm/supervisor/domain_xml.py
+++ b/src/middlewared/middlewared/plugins/vm/supervisor/domain_xml.py
@@ -1,7 +1,9 @@
+import os
 import shlex
 
 from middlewared.plugins.vm.devices import CDROM, DISK, PCI, RAW, DISPLAY, USB
 from middlewared.plugins.vm.numeric_set import parse_numeric_set
+from middlewared.plugins.vm.utils import SYSTEM_NVRAM_FOLDER_PATH, get_vm_nvram_file_name
 from middlewared.utils import Nid
 
 from .utils import create_element
@@ -256,10 +258,14 @@ def os_xml(vm_data):
         }
     )]
     if vm_data['bootloader'] == 'UEFI':
-        children.append(
+        children.extend([
             create_element(
                 'loader', attribute_dict={'text': f'/usr/share/OVMF/{vm_data["bootloader_ovmf"]}'},
                 readonly='yes', type='pflash',
-            )
-        )
+            ),
+            create_element('nvram', attribute_dict={
+                'text': os.path.join(SYSTEM_NVRAM_FOLDER_PATH, get_vm_nvram_file_name(vm_data)),
+            })
+        ])
+
     return create_element('os', attribute_dict={'children': children})


sonicaj (approved):

yocalebo (changes requested):
I'm still reviewing this so requesting changes so it doesn't get merged until after I can review. Will get to it shortly.

src/middlewared/middlewared/migration/0009_nvram_attr_vms.py
```diff
@@ -0,0 +1,67 @@
...
+        while True:
+            if not os.path.exists(new_path):
+                break
+            new_path = os.path.join('/mnt', root_ds_to_use, f'{file_name}_{i}')
+            i += 1
+
+        if file_name in existing_nvram_files:
+            to_copy_path = os.path.join(DEFAULT_NVRAM_FOLDER_PATH, file_name)
+            shutil.copy(to_copy_path, new_path)
+            shutil.copystat(to_copy_path, new_path)
```
themylogin, line 54: Why not just use `copy2`?

src/middlewared/middlewared/plugins/vm/vms.py
```diff
@@ -216,6 +221,24 @@ async def common_validation(self, verrors, schema_name, data, old=None):
                 'Invalid bootloader ovmf choice specified'
             )
 
+        if data['bootloader'] == 'UEFI' and not data['nvram_location']:
+            verrors.add(
+                f'{schema_name}.nvram_location',
+                'NVRAM location must be specified when booting with UEFI'
+            )
+
+        if data['nvram_location'] and await self.middleware.run_in_thread(
```
themylogin, line 230: What if the path does not exist but its parent is not writeable by libvirt user so automatic creation will fail?

src/middlewared/middlewared/migration/0009_nvram_attr_vms.py
```diff
@@ -0,0 +1,67 @@
...
+    root_ds_to_use = default_ds_to_use
+    for disk_device in middleware.call_sync('vm.device.query', [['vm', '=', vm['id']], ['dtype', '=', 'DISK']]):
+        if (root_ds := get_root_ds_from_disk_device(disk_device)) and root_ds in existing_root_datasets:
+            root_ds_to_use = root_ds
+            break
+
+    # We will copy the file here if it exists and update the nvram_location
+    if root_ds_to_use:
+        new_path = os.path.join('/mnt', root_ds_to_use, file_name)
+        i = 2
```
themylogin, line 44: Why 2?

src/middlewared/middlewared/migration/0009_nvram_attr_vms.py
```diff
@@ -0,0 +1,67 @@
...
+        while True:
+            if not os.path.exists(new_path):
+                break
+            new_path = os.path.join('/mnt', root_ds_to_use, f'{file_name}_{i}')
```
themylogin, line 48: Will the files be put directly to the root dataset? Can we consider putting them along the disk files/zvols?

Qubad786: @themylogin i thought there would be more edge cases which can arise like if zvol is `pool/dataset/zvol` and `pool/dataset` has readonly property set so it can cause problems with the migration.
Should I go ahead with trying this anyways or perhaps add a fallback to root dataset in that case?

src/middlewared/middlewared/plugins/vm/vms.py
Qubad786: I will add the validation ✅

src/middlewared/middlewared/migration/0009_nvram_attr_vms.py
themylogin: @yocalebo need your opinion here, personally I think we should keep similar files close, that's the strategy most users choose. Falling back to root dataset is a great idea.

yocalebo: As much as I _dont_ like saying it, I believe we should put these UEFI files on the system dataset. We already have all he logic in place to manage it.

anodos325:
It seems like we're probably going to get into situations where there are files on the boot volume that we want to persist across upgrades. Any thoughts about maybe modifying fhs.py in scale-build to add a `persistent-data` dataset, then you could create the qemu dir in there and we can clone it on upgrade. This also makes it easier to tar up for configuration backups.

themylogin:
> It seems like we're probably going to get into situations where there are files on the boot volume that we want to persist across upgrades. Any thoughts about maybe modifying fhs.py in scale-build to add a `persistent-data` dataset, then you could create the qemu dir in there and we can clone it on upgrade. This also makes it easier to tar up for configuration backups. This would also maybe give us a reasonable point in time to snapshot the data if needed.

@anodos325 why a separate dataset? We can store them perfectly fine in the `/data` dataset (e.g. in `/data/nvram`)

anodos325:
> > It seems like we're probably going to get into situations where there are files on the boot volume that we want to persist across upgrades. Any thoughts about maybe modifying fhs.py in scale-build to add a `persistent-data` dataset, then you could create the qemu dir in there and we can clone it on upgrade. This also makes it easier to tar up for configuration backups. This would also maybe give us a reasonable point in time to snapshot the data if needed.
> 
> @anodos325 why a separate dataset? We can store them perfectly fine in the `/data` dataset (e.g. in `/data/nvram`)

/data is _supposed_ to have permissions of 0o700 and the libvirt-qemu user needs to be able to read these.

Qubad786:
So should we look to add changes so we store this information in `/data` then?
We can just always use it then for these files and no need to have the ability to change that perhaps?

anodos325:
> So should we look to add changes so we store this information in `/data` then? We can just always use it then for these files and no need to have the ability to change that perhaps?

Precisely. This eliminates need to make it configurable and makes backporting trivial

src/middlewared/middlewared/migration/0009_nvram_attr_vms.py
```diff
@@ -0,0 +1,38 @@
+import os
+import shutil
+
+from middlewared.plugins.vm.utils import FINAL_NVRAM_FOLDER_PATH, get_vm_nvram_file_name
+
+
+DEFAULT_NVRAM_FOLDER_PATH = '/var/lib/libvirt/qemu/nvram'
+LIBVIRT_QEMU_UID = 64055
+LIBVIRT_QEMU_GID = 64055
```
anodos325, line 9: Is there any way we can shift libvirt to have lower IDs (below 1000)? Otherwise this runs the risk of colliding with LDAP users (with potential for an LDAP user to be able to unexpectedly edit things as the libvirt-qemu).

anodos325: I noticed the libvirt-qemu user also has its homedir set to `/var/lib/libvirt`, does this mean we're still facing potential for other settings not getting properly migrated between versions?

anodos325: NOTE: I am not familiar with libvirt internals, but this seems like something that we are probably able to control.

src/middlewared/middlewared/plugins/vm/utils.py
```diff
@@ -4,6 +4,7 @@
 
 
 ACTIVE_STATES = ['RUNNING', 'SUSPENDED']
+FINAL_NVRAM_FOLDER_PATH = '/data/subsystems/vm/nvram'
```
themylogin, line 7: Why do we call it `FINAL`? This name is confusing outside of data migration context.

src/middlewared/middlewared/migration/0009_nvram_attr_vms.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def migrate(middleware):
```
themylogin, line 12: Do we persist `/var/lib/libvirt/qemu/nvram` between upgrades somehow? I don't understand in what case will the files exist there when we first boot a TN release where this code will be included.

Qubad786: @themylogin yes we preserve them https://github.com/truenas/scale-build/pull/576

src/middlewared/middlewared/migration/0009_nvram_attr_vms.py
Qubad786: @anodos325 we do not want to preserve anything else in `/var/lib/libvirt` folder, it is just used by libvirt for running it's operations in our context. It has usecases but we are not using any of those so we should be safe.

I will work on changing the UID for libvirt user in a separate PR ✅

themylogin (approved):

src/middlewared/middlewared/migration/0010_nvram_attr_vms.py
```diff
@@ -0,0 +1,38 @@
...
+LIBVIRT_QEMU_UID = 64055
+LIBVIRT_QEMU_GID = 64055
+
+
+def migrate(middleware):
+    os.makedirs(SYSTEM_NVRAM_FOLDER_PATH, exist_ok=True)
+    os.chown(SYSTEM_NVRAM_FOLDER_PATH, LIBVIRT_QEMU_UID, LIBVIRT_QEMU_GID)
+
+    if middleware.call_sync('system.is_ha_capable'):
+        middleware.logger.error('Skipping nvram migration as system is HA capable')
```
yocalebo, line 17: Let's not make this an error. Can we do `.info`, please?

anodos325 (approved):

src/middlewared/middlewared/migration/0010_nvram_attr_vms.py
```diff
@@ -0,0 +1,38 @@
...
+def migrate_vm_nvram_file(middleware, vm):
+    file_name = get_vm_nvram_file_name(vm)
+    new_path = os.path.join(SYSTEM_NVRAM_FOLDER_PATH, file_name)
+    to_copy_path = os.path.join(DEFAULT_NVRAM_FOLDER_PATH, file_name)
+    if os.path.exists(to_copy_path):
+        shutil.copy2(to_copy_path, new_path)
+        os.chown(new_path, LIBVIRT_QEMU_UID, LIBVIRT_QEMU_GID)
+    else:
+        # File does not exist for us to copy, so we need to just log it
+        middleware.logger.error(
```
yocalebo, line 36: Same here, is this really an error? Maybe a `.debug` or `.info`

yocalebo (approved):

PR #13024, opened 2024-01-29
NAS-126979 / 24.10 / fix SCALE HA failover logic

yocalebo:
Take the following example:
4x total interfaces
2x marked critical for failover in the same failover group
the other 2x interfaces NOT marked critical for failover

take 1x critical interface down, no failover event is generated (by design since another interface in that group is functional)
take the other critical interface down, failover event is generated (by design since all other interfaces in that group are non-functional)

That's 2 out of the 4 interfaces taken down and a failover event is generated, however, the other controller that is becoming master receives 2 more events for the interfaces that are not marked critical for failover. By design, since these interfaces are not marked critical we ignore the failover event. This leaves HA in a broken state.

The changes being done here add filtering at the failover event thread level by simply ignoring interfaces that aren't marked critical for failover. The only events that will be forwarded down into middleware are critical interfaces.

### DIFF
--- a/src/middlewared/middlewared/plugins/device_/vrrp_events.py
+++ b/src/middlewared/middlewared/plugins/device_/vrrp_events.py
@@ -18,6 +18,7 @@ class VrrpObjs:
     fifo_thread = None
     event_thread = None
     event_queue = deque(maxlen=1)
+    non_crit_ifaces = tuple()
 
 
 class VrrpThreadService(Service):
@@ -37,6 +38,14 @@ class VrrpThreadService(Service):
         if VrrpObjs.event_thread is not None and VrrpObjs.event_thread.is_alive():
             VrrpObjs.event_thread.unpause()
 
+    def set_non_crit_ifaces(self):
+        if VrrpObjs.event_thread is not None and VrrpObjs.event_thread.is_alive():
+            VrrpObjs.event_thread.non_crit_ifaces = tuple(
+                i['int_interface'] for i in self.middleware.call_sync(
+                    'datastore.query', 'network.interfaces'
+                ) if not i['int_critical']
+            )
+
 
 class VrrpEventThread(Thread):
 
@@ -44,6 +53,7 @@ class VrrpEventThread(Thread):
         super(VrrpEventThread, self).__init__()
         self.middleware = kwargs.get('middleware')
         self.event_queue = VrrpObjs.event_queue
+        self.non_crit_ifaces = VrrpObjs.non_crit_ifaces
         self.shutdown_event = Event()
         self.pause_event = Event()
         self.grace_period = 0.5
@@ -124,6 +134,10 @@ class VrrpEventThread(Thread):
 
             if this_event is None:
                 # an event that we ignore
+                self.event_queue.pop()
+                continue
+            elif this_event['ifname'] in self.non_crit_ifaces:
+                LOGGER.debug('Received an event (%r) for a non-critical interface, ignoring.', this_event)
                 self.event_queue.pop()
                 continue
             elif last_event is None:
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -827,6 +827,9 @@ async def do_create(self, data):
             if interface_id:
                 await self.middleware.call('datastore.delete', 'network.interfaces', interface_id)
             raise
+        finally:
+            if await self.middleware.call('failover.licensed'):
+                await self.middleware.call('vrrpthread.set_non_crit_ifaces')
```
themylogin, line 832: Shouldn't we do this when the actual network configuration is applied instead?

src/middlewared/middlewared/plugins/device_/vrrp_events.py
```diff
@@ -289,6 +303,7 @@ async def _start_stop_vrrp_threads(middleware):
         if VrrpObjs.event_thread is None or not VrrpObjs.event_thread.is_alive():
             VrrpObjs.event_thread = VrrpEventThread(middleware=middleware, timeout=timeout)
             VrrpObjs.event_thread.start()
+            await middleware.call('vrrpthread.set_non_crit_ifaces')
```
themylogin, line 306: Isn't it better to have non-critical interfaces already before the thread is started to avoid race conditions?

src/middlewared/middlewared/plugins/device_/vrrp_events.py
```diff
@@ -18,6 +18,7 @@ class VrrpObjs:
     fifo_thread = None
     event_thread = None
     event_queue = deque(maxlen=1)
+    non_crit_ifaces = tuple()
```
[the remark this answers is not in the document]
yocalebo: Why would I use a set when I'm not making any alterations to this object and always overwriting it?

src/middlewared/middlewared/plugins/network.py
yocalebo: You mean during `interface.sync`?

src/middlewared/middlewared/plugins/device_/vrrp_events.py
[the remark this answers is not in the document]
themylogin: No practical reason, but `in` works faster for a set than for a tuple

src/middlewared/middlewared/plugins/network.py
themylogin: Yes (if that's what changes networking on HA)

src/middlewared/middlewared/plugins/device_/vrrp_events.py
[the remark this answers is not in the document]
themylogin: Because set is a hashtable (or a similar-purpose structure) while tuple only offers linear search.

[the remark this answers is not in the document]
yocalebo: derp, right

src/middlewared/middlewared/plugins/device_/vrrp_events.py
```diff
@@ -37,13 +38,22 @@ def unpause_events(self):
         if VrrpObjs.event_thread is not None and VrrpObjs.event_thread.is_alive():
             VrrpObjs.event_thread.unpause()
 
+    def set_non_crit_ifaces(self):
+        if VrrpObjs.event_thread is not None and VrrpObjs.event_thread.is_alive():
```
themylogin, line 42: We should remove is_alive check as currently we set the value on a thread that is not yet started

themylogin (approved):
Thank you!

PR #13028, opened 2024-01-30
NAS-127079 / 24.10 / Optimize role based integration tests

sonicaj:
This PR adds changes optimizing role based integration tests removing subsystem specific logic where we setup relevant resources needed for the subsystem calls to work. However earlier what we were missing was the fact that we don't really need to make sure those calls actually succeed, what we are testing is that does the consumer has relevant role to make that call in the first place - so just checking that the consumer can/cannot make the call is enough. This saves time/resources optimizing the integration tests for the specific case they are written for.

### DIFF
--- a/src/middlewared/middlewared/test/integration/assets/roles.py
+++ b/src/middlewared/middlewared/test/integration/assets/roles.py
@@ -13,8 +13,11 @@ def common_checks(
     with unprivileged_user_client(roles=[role]) as client:
         if valid_role:
             if valid_role_exception:
-                with pytest.raises(Exception):
+                with pytest.raises(Exception) as exc_info:
                     client.call(method, *method_args, **method_kwargs)
+
+                assert not isinstance(exc_info.value, ClientException)
+
             elif is_return_type_none:
                 assert client.call(method, *method_args, **method_kwargs) is None
             else:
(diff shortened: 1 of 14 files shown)


src/middlewared/middlewared/test/integration/assets/roles.py
```diff
@@ -0,0 +1,26 @@
...
+from middlewared.test.integration.assets.account import unprivileged_user_client
+
+
+def common_checks(
+    method, role, valid_role, valid_role_exception=True, method_args=None, method_kwargs=None, is_return_type_none=False
+):
+    method_args = method_args or []
+    method_kwargs = method_kwargs or {}
+    with unprivileged_user_client(roles=[role]) as client:
+        if valid_role:
```
mgrimesix, line 14: On the valid_role branch should we confirm `errno` is NOT `EACCES` and `error` is NOT 'Not authorized'?

sonicaj: @mgrimesix we can just check that the exception is not `ClientException` and that should be sufficient as other types of errors are raised when there is an issue with malformed parameters etc

mgrimesix (approved):

PR #13037, opened 2024-01-30
NAS-126993 / 24.04-RC.1 / Change graphite hostname attribute

Qubad786:
## Problem
The hostname attribute in the netdata graphite exporter is causing confusion among users who associate it with the actual hostname of the server. However, in reality, it acts more like a namespace where all the exported data is stored under the same name.

## Solution

Change the `hostname` attribute to `namespace` for better user understanding.


src/middlewared/middlewared/alembic/versions/24.04/2023-09-07_19-50_add_exporting_table_for_reporting.py
```diff
@@ -49,7 +49,7 @@ def upgrade():
             'destination_ip': graphite_ip[0],
             'destination_port': 2003,
             'prefix': 'scale',
-            'hostname': hostname,
+            'namespace': hostname,
```
anodos325, line 52: Why are we editing an existing migration?

Qubad786: Thank you @anodos325 for catching that - was a leftover of my debugging..

anodos325 (approved):

PR #13041, opened 2024-01-30
NAS-127251 / 24.10 / Add pytest markers for more intelligent runs

anodos325:
This commit adds basic framework for setting custom markers in our API tests. They can either be set globally for the module via
`pytestmark = pytest.mark.smb`

and multiple for module
`pytestmark = [pytest.mark.fs, pytest.mark.slow]`

or via test-specific decorator
`@pytest.mark.smb`

Long-running tests (more than a couple of minutes) should be marked as `slow` so that they can be skipped in incremental builds.

New markers can be added to the pytest.ini file. If someone feels particularly ambitious they can enhance in the future by adding a parser to assign markers based on test file names.

### DIFF
--- a/tests/runtest.py
+++ b/tests/runtest.py
@@ -37,6 +37,7 @@ Mandatory option
     --interface <interface>     - The interface that TrueNAS is run one
 
 Optional option
+    --markers "smb or nfs"      - Markers to run
     --test <test name>          - Test name (Network, ALL)
     --tests <test1>[,test2,...] - List of tests to be supplied to pytest
     --vm-name <VM_NAME>         - Name the the Bhyve VM
@@ -66,6 +67,7 @@ option_list = [
     "isns_ip=",
     "pool=",
     "tests=",
+    "markers=",
 ]
 
 # look if all the argument are there.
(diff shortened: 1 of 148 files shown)


tests/runtest.py
```diff
@@ -37,6 +37,7 @@
     --interface <interface>     - The interface that TrueNAS is run one
 
 Optional option
+    --markers "smb or nfs"      - Markers to run
```
mgrimesix, line 40: Not easy to indicate the format.  Maybe 
`--markers "<markers to run>"       - Quoted list separated by 'or', e.g. "smb or nfs".   See pytest.ini for list of markers.`

mgrimesix:
What would the runtest option look like if I wanted to run 'nfs' and tests without 'slow' markings only?

tests/runtest.py
```diff
@@ -123,6 +126,8 @@
         callargs.append('-s')
     elif output == '--tests':
         tests.extend(arg.split(','))
+    elif output == '--markers':
+        markers.extend(['-m', f'base or {arg}'])
```
themylogin, line 130: If `arg` can be a logic expression shouldn't we enclose it in parenthesis?

themylogin:
Can we fail the whole test suite if some test file does not have any of the markers?

PR #13047, opened 2024-01-31
NAS-127116 / 24.10 / Simplify filesystem plugin for ACLs

anodos325:
We no longer have unified ACL codebase with FreeBSD and so we no longer need to have separate acl_base and acl_linux files.


yocalebo:
@Qubad786 this change will break the `netdata_setup()` method. So this relates to the ticket I assigned you regarding that method. I won't approve this, yet but we should investigate it sooner rather than later 😄

src/middlewared/middlewared/plugins/filesystem_/utils.py
```diff
@@ -0,0 +1,174 @@
...
+                # File and this entry doesn't inherit on files
+                continue
+
+            if isdir:
+                if not flags.get('DIRECTORY_INHERIT', False):
+                    if flags['NO_PROPAGATE_INHERIT']:
+                        # doesn't apply to this dir and shouldn't apply to contents.
+                        continue
+
+                    # This is a directoy ACL and we have entry that only applies to files.
```
mgrimesix, line 140: Super small nit:  directoy

anodos325:
@yocalebo I think the relevant work has now been merged.

yocalebo (approved):

PR #13069, opened 2024-02-01
NAS-127087 / 24.10 / Support roles for event sources

themylogin:


sonicaj:
We should be backporting this to dragonfish as well ?

PR #13071, opened 2024-02-01
add readonly to truecommand.config events

yocalebo:
Needed by the UI team.

### DIFF
--- a/src/middlewared/middlewared/plugins/truecommand/__init__.py
+++ b/src/middlewared/middlewared/plugins/truecommand/__init__.py
@@ -10,7 +10,11 @@ async def _event_system_ready(middleware, event_type, args):
 
 async def setup(middleware):
     await middleware.call('truecommand.config')
-    middleware.event_register('truecommand.config', 'Sent on TrueCommand configuration changes.')
+    middleware.event_register(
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/truecommand/__init__.py
```diff
@@ -10,7 +10,11 @@ async def _event_system_ready(middleware, event_type, args):
 
 async def setup(middleware):
     await middleware.call('truecommand.config')
-    middleware.event_register('truecommand.config', 'Sent on TrueCommand configuration changes.')
+    middleware.event_register(
+        'truecommand.config',
+        'Sent on TrueCommand configuration changes.',
+        roles=['READONLY_ADMIN']
```
sonicaj, line 16: Should we perhaps add role for truecommand as well and then use it's read role here ?

sonicaj (approved):

src/middlewared/middlewared/plugins/truecommand/__init__.py
yocalebo: Yeah, I'm doing the roles based changes in a separate PR because we have to fix some other things.

PR #13091, opened 2024-02-05
NAS-127220 / 24.10 / Correct logic error in NFS share validation check

mgrimesix:
A secondary validation test was over riding an earlier validation check.  It is the earlier validation check which has wording the better matches the trapped misconfiguration resulting in a more sane error message.

Enhanced the CI testing to help catch this sort of error.


mgrimesix:
Mixed up the PR name.  Going to re-do this.

PR #13093, opened 2024-02-06
NAS-125640 / 24.10 / Improve checking if containerd socket is available

sonicaj:
This PR adds changes to consume containerd socket instead of just connecting to the socket to see if it is available. Problem we see is that sometimes it is possible the socket can be connected to but containerd is still initiating itself which leads to calls failing which involve containerd.

### DIFF
--- a/src/middlewared/middlewared/plugins/container_runtime_interface/containerd_utils.py
+++ b/src/middlewared/middlewared/plugins/container_runtime_interface/containerd_utils.py
@@ -4,6 +4,8 @@ from .client import ContainerdClient
 
 
 def check_containerd_connection() -> bool:
+    # Just to clarify that this does not make any internet requests but rather just checks if the socket is available
+    # and if we can connect to it. The image check returns nothing if the image is not present.
     with contextlib.suppress(Exception):
         with ContainerdClient('image') as client:
             client.get_image('alpine')
--- a/src/middlewared/middlewared/plugins/container_runtime_interface/containerd_utils.py
+++ b/src/middlewared/middlewared/plugins/container_runtime_interface/containerd_utils.py
@@ -1,0 +1,11 @@
+import contextlib
+
+from .client import ContainerdClient
+
+
+def check_containerd_connection() -> bool:
+    with contextlib.suppress(Exception):
+        with ContainerdClient('image') as client:
+            client.get_image('alpine')
+            return True
+    return False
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/container_runtime_interface/containerd_utils.py
```diff
@@ -0,0 +1,11 @@
+import contextlib
+
+from .client import ContainerdClient
+
+
+def check_containerd_connection() -> bool:
+    with contextlib.suppress(Exception):
+        with ContainerdClient('image') as client:
+            client.get_image('alpine')
```
themylogin, line 9: Is there a easier/faster operation to check containerd health? Or is this image always present even on just-initialized system?

stavros-k: We can use maybe https://hub.docker.com/r/rancher/pause 
This image is used by k3s anyway and its 290kb (compared to alpine ~3mb)

sonicaj: If the image is not present, it returns nothing - so it is faster that way in terms of performance rather then us listing containers etc. I think we should be good to go as is

stavros-k: Ah checks for presence. That should be fine yea!

yocalebo: this seems flawed. what happens when the network isn't up but containerd is functional? You're trading 1 problem for another.

sonicaj: Which network ?

yocalebo: The truenas network? If this check gets called before the truenas network is setup, then this will return False because it can't get out to the internet.

sonicaj: How this works is the following:

1. This request is definitely made after system is in ready state
2. This just connects to containerd socket and there is no outbound network call involved here - if the image exists, great - if not even then great as it is not pulling or anything, just checking it's existence on the filesystem

To conclude, that concern does not apply

yocalebo: what a strange name...`client.get_image` managed to confuse 3 different developers....are we sure this doesn't go out to the network at all?

themylogin (approved):
LGTM if you put a comment there explaining that no internet requests are made

sonicaj:
backport

PR #13094, opened 2024-02-06
NAS-127158 / 24.10 / change format of rsync push in module mode

stavros-k:
Changes the format of rsync push in module mode.
While the current format is valid based on the man pages (https://rsync.samba.org/ftp/rsync/rsync.html), it prompts for password, but the new proposed format does not.

Modified this on a test **Core** system and
image

And Rsync task worked, while previously it was failing with auth error (publickey,password).


themylogin (approved):

yocalebo:
backport

PR #13099, opened 2024-02-06
NAS-125924 / 24.10 / Make `run` use thread pool

themylogin:
Running `create_subprocess_exec` performs blocking IO in main loop. Let's delegate `fork+exec` to our `IoThreadPoolExecutor`.

Note that few `Popen` calls are left for cloudsync that reads process output line-by-line. There should not cause any damage as they are not called frequently.


themylogin:
time 10:00

yocalebo (approved):

PR #13100, opened 2024-02-06
Do not use ix temp dir when creating catalogs

sonicaj:
This PR fixes a problem where we were still using temp dir when a new catalog was being added so we could validate it before adding it to the system. With the changes, now we use the apps dataset to hold the catalog and validate it for any issues.

### DIFF
--- a/src/middlewared/middlewared/plugins/catalogs_linux/update.py
+++ b/src/middlewared/middlewared/plugins/catalogs_linux/update.py
@@ -265,9 +265,12 @@ class CatalogService(CRUDService):
         if not data.pop('force'):
             job.set_progress(40, f'Validating {data["label"]!r} catalog')
             # We will validate the catalog now to ensure it's valid wrt contents / format
+            k8s_dataset = (await self.middleware.call('kubernetes.config'))['dataset']
             path = os.path.join(
-                TMP_IX_APPS_DIR, 'validate_catalogs', convert_repository_to_path(data['repository'], data['branch'])
+                '/mnt', k8s_dataset, 'catalogs/validate_catalogs',
+                convert_repository_to_path(data['repository'], data['branch'])
             )
+            await self.middleware.run_in_thread(shutil.rmtree, path, ignore_errors=True)
             try:
                 await self.middleware.call('catalog.update_git_repository', {**data, 'location': path})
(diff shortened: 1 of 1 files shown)


anodos325:
Is there an earlier check to see whether user has configured a k8s dataset?

sonicaj:
@anodos325 yes we have a check before for that and we raise validation errors in case that is not configured

src/middlewared/middlewared/plugins/catalogs_linux/update.py
```diff
@@ -265,9 +265,12 @@ async def do_create(self, job, data):
         if not data.pop('force'):
             job.set_progress(40, f'Validating {data["label"]!r} catalog')
             # We will validate the catalog now to ensure it's valid wrt contents / format
+            k8s_dataset = (await self.middleware.call('kubernetes.config'))['dataset']
             path = os.path.join(
-                TMP_IX_APPS_DIR, 'validate_catalogs', convert_repository_to_path(data['repository'], data['branch'])
+                '/mnt', k8s_dataset, 'catalogs/validate_catalogs',
+                convert_repository_to_path(data['repository'], data['branch'])
             )
+            await self.middleware.run_in_thread(shutil.rmtree, path, ignore_errors=True)
```
yocalebo, line 273: Why are we doing this? This gets run in the `finally:` block down below?

yocalebo (changes requested):

src/middlewared/middlewared/plugins/catalogs_linux/update.py
sonicaj: This is to ensure that the directory does not exist before we try to clone into it

sonicaj: In ideal situations, almost always the directory won't be there and this is just a failsafe

sonicaj: This still gets consumed for official catalog when we sync catalog. Basically we still allow syncing official catalog to tmpfs to allow users to view available apps if they don't have apps configured (https://github.com/truenas/middleware/pull/12839).

yocalebo (approved):
thanks

PR #13114, opened 2024-02-11
NAS-127157 / 24.10 / Mark HA nodes as active/standby in debug

Qubad786:
## Context

There is no way currently to determine that which compressed file belongs to active or which belongs to standby when a debug for a HA system is taken. The changes made label the compressed folder to represent that they belong to active/standby appropriately.

### DIFF
--- a/src/middlewared/middlewared/plugins/system/debug.py
+++ b/src/middlewared/middlewared/plugins/system/debug.py
@@ -122,6 +122,7 @@ class SystemService(Service):
             node = self.middleware.call_sync('failover.node')
 
             tario = io.BytesIO()
+            host_status = self.middleware.call_sync('failover.status')
             with tarfile.open(fileobj=tario, mode='w') as tar:
 
                 if node == 'A':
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/system/debug.py
```diff
@@ -131,9 +132,9 @@ def debug(self, job):
                     my_hostname = network['hostname_b']
                     remote_hostname = network['hostname']
 
-                tar.add(debug_job.result, f'{my_hostname}.txz')
+                tar.add(debug_job.result, f'{my_hostname}_{"active" if host_status == "MASTER" else "standby"}.txz')
```
yocalebo, line 135: failover.status can return many different statuses. Including ERROR. We need to play it safe and be more explicit with the check. i.e. something like
```
if failover_status == 'MASTER':
    _active
else:
    don't_rename_archive

yocalebo (changes requested):

yocalebo (approved):

PR #13122, opened 2024-02-12
NAS-127319 / 24.10 / Make sure middleware socket is present

sonicaj:
This commit fixes an issue where checking file's existence we were running it whereas it required middlewared to be actually running - now this becomes a problem in our unit tests pipeline as middleware is not running in the docker container we create and these changes are not really required either for us torun the unit tests etc.

### DIFF
--- a/src/freenas/usr/bin/install-dev-tools
+++ b/src/freenas/usr/bin/install-dev-tools
@@ -27,7 +27,7 @@ PACKAGES=(
 )
 PIP_PACKAGES=()
 
-if [ -f /usr/local/libexec/disable-rootfs-protection ]; then
+if [[ -f /usr/local/libexec/disable-rootfs-protection && -S /var/run/middleware/middlewared.sock ]]; then
     # Running on SCALE
     /usr/local/libexec/disable-rootfs-protection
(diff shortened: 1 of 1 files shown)


src/freenas/usr/bin/install-dev-tools
```diff
@@ -27,7 +27,7 @@ PACKAGES=(
 )
 PIP_PACKAGES=()
 
-if [ -f /usr/local/libexec/disable-rootfs-protection ]; then
+if [ -f /usr/local/libexec/disable-rootfs-protection && -f /var/run/middleware/middlewared.sock ]; then
```
themylogin, line 30: It’s just `/run`. And what does “running on SCALE” comment below mean, where else can this be running?

themylogin (approved):

src/freenas/usr/bin/install-dev-tools
sonicaj: It can be running in a docker container and using the path we have defined in middleware client
```
MIDDLEWARE_RUN_DIR = '/var/run/middleware'
```
It would work well with both `/run` and `/var/run` :)

PR #13132, opened 2024-02-13
NAS-127347 / 24.10 / Add more error handling for filter_list

anodos325:
This adds handling for the case where API user specifies a filter that for a non-existent key during `in` operations.


themylogin (commented):
Let's add a unit test here

anodos325:
Addressed by fix in NAS-12480.

PR #13134, opened 2024-02-13
NAS-127350 / 24.10 / Add ability to filter by keys in list objects

anodos325:
There are some cases where API user may wish to query an endpoint based on nested values within a list. For example

[["local_groups.*.name", "=", "bad"]] passed to privilege.query to return all privileges where the group "bad" is present in the list of local groups to which the privilege applies.

### DIFF
--- a/src/middlewared/middlewared/apidocs/templates/websocket/query.md
+++ b/src/middlewared/middlewared/apidocs/templates/websocket/query.md
@@ -143,14 +143,13 @@ Javascript:
       ["local_groups.0.name", "=", "myuser"],
     ]
 
-Alternatively, the index may be omitted, which will result in the result returning if any array member has a key matching the
-value.
+Alternatively, an asterisk (`*`) may be substituted for the array index, which match any entry where an array member has a key matching the value. for example, the following query-filters if passed to the `privilege.query` endpoint will return entries where any member of the local groups array has a `name` key with the value of `myuser`.
 
 Javascript:
 
     :::javascript
     [
-      ["local_groups.name", "=", "myuser"],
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/apidocs/templates/websocket/query.md
```diff
@@ -132,6 +132,28 @@ Javascript:
...
+    ]
+
+Alternatively, the index may be omitted, which will result in the result returning if any array member has a key matching the
+value.
+
+Javascript:
+
+    :::javascript
+    [
+      ["local_groups.name", "=", "myuser"],
```
themylogin, line 153: Can we be writing this in a less ambiguous way? Like `local_groups.*.name`

anodos325: Yes. That's probably a better idea.

themylogin (approved):
Thank you!

src/middlewared/middlewared/utils/__init__.py
```diff
@@ -96,9 +100,40 @@ def partition(s):
...
+        elif isinstance(cur, (list, tuple)):
+            if not left.isdigit():
+                # return all members and the remaining portion of path
+                if left == '*':
+                    return FilterGetResult(result=cur, key=right, done=False)
+
+                raise ValueError(f'{left}: must be array index or wildcard character')
+
+            left = int(left)
+            cur = cur[left] if left < len(cur) else None
```
sonicaj, line 119: if `cur` is `None`, i don't think there is any point in continuing the loop ?

sonicaj (approved):

PR #13149, opened 2024-02-14
NAS-127112 / 24.10 / Improve netdata setup

Qubad786:
## Problem
The default path for Netdata database storage exists at `/var/cache/netdata`, but its data is stored in the system dataset. Symlinks were used to bind that dataset with the Netdata storage directory. However, the `netdata_setup` function, responsible for this process, runs on every netdata restart, although it should only be required once or when the system dataset is migrated.

## Solution
Simplify the Netdata setup logic by changing the Netdata database storage directory to `/var/db/system/netdata` and mounting the Netdata system dataset in that same directory. Other tasks performed by the `netdata_setup` function will also be handled in the system dataset in a clean/centric way.

### DIFF
--- a/src/middlewared/middlewared/etc_files/netdata/netdata.conf.mako
+++ b/src/middlewared/middlewared/etc_files/netdata/netdata.conf.mako
@@ -1,11 +1,13 @@
 <%
     from middlewared.plugins.reporting.netdata.utils import NETDATA_PORT, NETDATA_UPDATE_EVERY
+
     netdata_cache_dataset = middleware.call_sync('reporting.netdata_storage_location')
-    disk_space = middleware.call_sync('netdata.get_disk_space')
     if not netdata_cache_dataset:
         # Let's exit if netdata storage is not in place
         middleware.logger.error('Netdata configuration file could not be generated')
         raise FileShouldNotExist()
+
+    disk_space = middleware.call_sync('netdata.get_disk_space')
 %>\
 [global]
 	run as user = netdata
--- a/src/middlewared/middlewared/etc_files/netdata/netdata.conf.mako
+++ b/src/middlewared/middlewared/etc_files/netdata/netdata.conf.mako
@@ -1,13 +1,11 @@
 <%
-	from middlewared.plugins.reporting.netdata.utils import NETDATA_PORT, NETDATA_UPDATE_EVERY
-
-
-	if not middleware.call_sync('reporting.netdata_setup'):
-		# Let's exit if netdata storage is not in place
-		middleware.logger.error('Netdata configuration file could not be generated')
-		raise FileShouldNotExist()
-
-	disk_space = middleware.call_sync('netdata.get_disk_space')
+    from middlewared.plugins.reporting.netdata.utils import NETDATA_PORT, NETDATA_UPDATE_EVERY
+    netdata_cache_dataset = middleware.call_sync('reporting.netdata_storage_location')
+    disk_space = middleware.call_sync('netdata.get_disk_space')
+    if not netdata_cache_dataset:
+        # Let's exit if netdata storage is not in place
+        middleware.logger.error('Netdata configuration file could not be generated')
+        raise FileShouldNotExist()
 %>\
 [global]
 	run as user = netdata
@@ -22,6 +20,10 @@
 	mode = dbengine
 	storage tiers = 1
 	dbengine multihost disk space MB = ${disk_space}
+
+[directories]
+    cache = ${netdata_cache_dataset}
+    home = ${netdata_cache_dataset}
 
 [plugins]
 	proc = yes
--- a/src/middlewared/middlewared/plugins/sysdataset.py
+++ b/src/middlewared/middlewared/plugins/sysdataset.py
@@ -1,6 +1,5 @@
 import errno
 import json
-import middlewared.sqlalchemy as sa
 import os
 import shutil
 import subprocess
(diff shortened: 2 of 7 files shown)


src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -500,24 +501,36 @@ async def setup_datasets(self, pool, uuid):
...
+                }
                 if update_props_dict:
                     await self.middleware.call(
                         'zfs.dataset.update',
                         dataset,
                         {'properties': update_props_dict},
                     )
 
+            for create_path_config in config.get('create_paths', []):
+                os.makedirs(create_path_config['path'], exist_ok=True)
```
yocalebo, line 516: Lots of blocking IO in a coroutine.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -500,24 +501,39 @@ async def setup_datasets(self, pool, uuid):
...
                         dataset,
                         {'properties': update_props_dict},
                     )
 
+            await self.middleware.run_in_thread(self.__create_relevant_paths, config.get('create_paths', []))
+
+    def __create_relevant_paths(self, create_paths):
+        for create_path_config in create_paths:
+            os.makedirs(create_path_config['path'], exist_ok=True)
+            if all(create_path_config[k] for k in ('uid', 'gid')) and (
```
yocalebo, line 520: we need to simplify this `all(` if branch. Stat'ing files/paths is relatively cheap, but we should still do it as little as possible.
```
for cpath in create_paths:
    os.makedirs(cpath['path'], exist_ok=True)
    cpath_stat = os.stat(cpath['path'])
    if cpath_stat.st_uid != cpath['uid'] or cpath_stat.st_gid != path['gid']:
        os.chown(cpath['path'], cpath['uid'], cpath['gid'])
```
Finally, since this is in our system dataset plugin, it is imperative that we wrap all these file I/O operations in a `try: except` clause so that we don't crash here. If system dataset fails, then it will cause a cascading set of failures for other services.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -573,14 +595,37 @@ def __umount(self, pool, uuid):
...
+            (
+                f'netdata-{uuid}', {
+                    'chown_config': {'uid': 999, 'gid': 997},
+                    'mountpoint': f'{SYSDATASET_PATH}/netdata',
+                    'props': {'canmount': 'noauto'},
+                    'create_paths': [
+                        {'path': '/var/log/netdata', 'uid': 999, 'gid': 997},
+                    ],
+                }
+            ),
```
anodos325, line 615: I think behavior is getting complex enough that it's a good idea to have a spec file (like fhs.py in our installer) that defines our system dataset contents and configuration.

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -500,33 +502,57 @@ async def setup_datasets(self, pool, uuid):
...
+                mountpoint = f'{path}/{name}'
+
             if os.path.ismount(mountpoint):
                 continue
 
             with suppress(FileExistsError):
                 os.mkdir(mountpoint)
             subprocess.run(['mount', '-t', 'zfs', dataset, mountpoint], check=True)
+
+            if chown_config := ds_config.get('chown_config'):
```
anodos325, line 551: Let's default to `-1` for uid and gid in the chown config. Then we can skip the stat check and just chown. -1 here has special meaning to leave existing value.

Qubad786: @anodos325 to be sure i follow - for all datasets we have a default `chown_config` that where ever they get mounted we do `chown` but using `-1`?

Qubad786: In terms of operations - it would mean that we always do `chown` so another extra step..not sure if it has performance impacts as we are using `-1`

yocalebo: @Qubad786 -1 means the OS won't change the ownership values if they're the same so it actually reduces I/O.

src/middlewared/middlewared/etc_files/netdata/netdata.conf.mako
```diff
@@ -1,13 +1,11 @@
...
-
-	if not middleware.call_sync('reporting.netdata_setup'):
-		# Let's exit if netdata storage is not in place
-		middleware.logger.error('Netdata configuration file could not be generated')
-		raise FileShouldNotExist()
-
-	disk_space = middleware.call_sync('netdata.get_disk_space')
+    from middlewared.plugins.reporting.netdata.utils import NETDATA_PORT, NETDATA_UPDATE_EVERY
+    netdata_cache_dataset = middleware.call_sync('reporting.netdata_storage_location')
+    disk_space = middleware.call_sync('netdata.get_disk_space')
```
yocalebo, line 4: Is there a reason you're populating `disk_space` before the `if not netdata_cache_dataset` check with these changes? Seems like we should revert back to what you were doing before these changes. (i.e. only setting that variable after that check)

src/middlewared/middlewared/plugins/system_dataset/hierarchy.py
```diff
@@ -0,0 +1,178 @@
...
+                'required': ['mountpoint', 'readonly', 'snapdir'],
+            },
+            'chown_config': {
+                'type': 'object',
+                'properties': {
+                    'uid': {'type': 'integer'},
+                    'gid': {'type': 'integer'},
+                    'mode': {'type': 'integer'},
+                },
+                'required': ['uid', 'gid']
```
anodos325, line 42: Should we also require `mode` key?

anodos325 (approved):

anodos325:
Please run through CI before merge.

yocalebo (approved):

PR #13152, opened 2024-02-15
NAS-127344 / 24.10 / Add roles for truecommand plugin

sonicaj:


src/middlewared/middlewared/plugins/truecommand/wireguard.py
```diff
@@ -101,7 +101,7 @@ async def wireguard_connection_health(self):
 
     @no_auth_required
     @throttle(seconds=2, condition=throttle_condition)
-    @accepts()
+    @accepts(roles=['TRUECOMMAND_READ'])
```
yocalebo, line 104: How does this do anything? The endpoint is marked with `no_auth_required` so anyone can see this.

yocalebo (dismissed):

themylogin (approved):

PR #13185, opened 2024-02-20
NAS-127416 / 24.10 / Do not crash if DB already has a pool that is being imported

themylogin:
This is a pretty normal situation. E.g. if the DB has the pool named `tank`, but disks were not plugged to the system when it was booting (so boot-time pool import failed), but then the disks were hot-plugged and the pool is being imported explicitly using "import pool".

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/import_pool.py
+++ b/src/middlewared/middlewared/plugins/pool_/import_pool.py
@@ -118,7 +118,7 @@ class PoolService(Service):
         else:
             pool_name = new_name
 
-        # Let's umount any datasets if root dataset of the new pool is locked and it has unencrypted datasets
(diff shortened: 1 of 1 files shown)


themylogin:
time 0:15

src/middlewared/middlewared/plugins/pool_/import_pool.py
```diff
@@ -118,7 +118,7 @@ async def import_pool(self, job, data):
         else:
             pool_name = new_name
 
-        # Let's umount any datasets if root dataset of the new pool is locked and it has unencrypted datasets
```
yocalebo, line 121: Let's is the proper contraction. If you don't like the apostrophe, then use "let us"

yocalebo (approved):

PR #13199, opened 2024-02-20
NAS-127457 / 24.10 / Improve speed of jbof.create and jbof.delete APIs

bmeagherix:
Improve the speed of these APIs, generally by performing certain operations in parallel rather than in series.

Also add some additional logging.

---
The need for this became apparent when  we moved the JBOF to new hardware, and also fully populated the JBOF with drives.

### DIFF
--- a/src/middlewared/middlewared/plugins/jbof/crud.py
+++ b/src/middlewared/middlewared/plugins/jbof/crud.py
@@ -388,7 +388,6 @@ class JBOFService(CRUDService):
                 verrors.add(schema, 'Unable to communicate with the expansion shelf')
                 return
             self.logger.debug('Configured node: %r', connected_shelf_ips)
-        self.logger.debug('Configured host')
 
     @private
     async def hardwire_node(self, node, shelf_index, shelf_ip_to_mac, skip_ips=[]):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -372,18 +374,21 @@ async def hardwire_host(self, mgmt_ip, shelf_index, schema, verrors):
...
                     return
+                self.logger.debug('Configured node %r: %r', node, connected_shelf_ips)
         else:
             connected_shelf_ips = await self.hardwire_node('', shelf_index, shelf_ip_to_mac)
             if not connected_shelf_ips:
                 # Failed to connect any IPs => error
                 verrors.add(schema, 'Unable to communicate with the expansion shelf')
                 return
+            self.logger.debug('Configured node: %r', connected_shelf_ips)
+        self.logger.debug('Configured host')
```
mgrimesix, line 391: I like the messages, but with only 'Configured host' the context might be lost?

yocalebo: I agree with @mgrimesix. Seems that log message might be a bit superfluous.

mgrimesix (approved):

PR #13200, opened 2024-02-21
NAS-127460 / 24.10 / Improve disk partitioning

mgrimesix:
This PR does a few things to improve our disk format (partitioning) and clean handling.
### Add partition names/labels
This makes it easier to identify the partition function.   
Example output from `sgdisk -p /dev/sdb`, where `sdb` is a 4 GiB drive with 2 GiB swap partition:
```
Number  Start (sector)    End (sector)  Size       Code  Name
   1            2048         4196352   2.0 GiB     8200  swap
   2         4198400         8386560   2.0 GiB     BF01  data

```
### Clean up inconsistencies
Partitioning a disk without swap would result in a partition shortened by 1MiB
```
	Number  Start (sector)    End (sector)  Size       Code  Name
	   1            4096         4192256   2.0 GiB     BF01
```
The starting sector for that partition should be 2048
### Remove unused parameters
We now _always_ partition disks to the largest possible data partition.  As such, removed the parameter that specifies the size of the data partition. 
### Allow for partitioning smallest possible disk
This is mostly in support of TrueNAS VMs for development testing.  If the default 2 GiB swap partition is used the smallest 'supported' drive is 3 GiB.  If no swap is requested a drive as small as 1 GiB may be used.
### Improve user facing error messages
We provide error messages tailored to the condition
### Improve 'QUICK' disk wipe
The 'QUICK' disk wipe method would clear the MBR/GPT data that gets stored at both ends of the drive.  However,
it did not clean up 'partition' data.  This might, for example, include private data written in partition gaps or private data used by the filesystem.  This additional clean step is enacted when there is more than one partition.
### Improve disk format CI testing
The start, end and size of the partitions are now deterministic.  The CI tests have been improved to take advantage.
### Improve disk wipe CI testing
Add test for the added 'middle' partition cleaning

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/format.py
+++ b/src/middlewared/middlewared/plugins/disk_/format.py
@@ -6,7 +6,32 @@ from middlewared.service import CallError, private, Service
 class DiskService(Service):
 
     @private
-    def format(self, disk, data_size, swap_size_gb):
+    def format(self, disk, swap_size_gb):
+        """
+        Format a data drive with a maximized data partition
+        Rules:
+            - The min_data_size is 512MiB
+                i.e. the drive must be bigger than 512MiB + 2MiB (partition offsets)
+                NOTE: 512MiB is arbitrary, but allows for very small drives
+            - If swap_size_gb is not None, then
+                * The swap is sized in 1 GiB increments
+                * Drive partitioning will abort if requested swap cannot be accomodated
+                * A swap partition will be created only if the following is true:
+                    swap_size < drive_size - (data_size + partition_gaps)
+                * The data partition will be reduced by swap_size_gb
+            - The drive is left unchanged if the drive cannot be partitioned according to the rules
+
+        The current config default requested swap is 2 GiB
+        A typical drive partition diagram (assuming 1 MiB partition gaps):
+
+        | - unused - | - partition 1 - | - unused -| - partition 2 - | - unused - |
+        |------------|-----------------|-----------|-----------------|------------|
+        | 1 MiB gap  |   2 GiB swap    | 1 MiB gap |   N GiB data    | 1 MiB gap  |
+
+        """
+        if swap_size_gb is not None and (swap_size_gb < 0 or not isinstance(swap_size_gb, int)):
+            raise CallError('Requested swap must be a non-negative integer')
+
         dd = self.middleware.call_sync('device.get_disk', disk)
         if not dd:
             raise CallError(f'Unable to retrieve disk details for {disk!r}')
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/disk_/wipe.py
```diff
@@ -25,16 +25,28 @@ def _wipe_impl(self, job, dev, mode, event):
...
+            disk_parts = self.middleware.call_sync('disk.list_partitions', dev)
+            if len(disk_parts) > 1:
+                for partition in disk_parts[1:]:
+                    # Start 2 MiB back from the start and 'clean' 2 MiB past, 4 MiB total
+                    os.lseek(f.fileno(), partition['start'] - (2 * CHUNK), os.SEEK_SET)
+                    for i in range(4):
+                        os.write(f.fileno(), to_write)
+                        if event.is_set():
+                            return
+                        # This is quick. We can reasonably skip the progress update
```
bmeagherix, line 45: A couple of comments on this section
1. Does it only apply to 'QUICK' mode?  (i.e. non-quick mode is doing the whole disk, right?)
2. Should it be moved until after the HEAD and TAIL of the disk (though maybe read `disk_parts` first).  They _seem_ more important to scrub and I observe the `if event.is_set(): return`.  Hmm.

tests/api2/test_disk_wipe.py
```diff
@@ -24,6 +22,40 @@ def test_disk_wipe_exported_zpool_in_disk_get_unused():
         assert disk['exported_zpool'] is None
 
 
+def test_disk_wipe_partition_clean():
+    """
+    Confirm we clean up around the middle partitions
+    """
+    signal_msg = "ix private data"
+
+    disk = call("disk.get_unused")[0]["name"]
```
bmeagherix, line 31: Do we need to SKIP or FAIL the test if there are no unused disks available?

tests/api2/test_disk_wipe.py
```diff
@@ -24,6 +22,40 @@ def test_disk_wipe_exported_zpool_in_disk_get_unused():
...
+
+    # Create 1 GiB swap and a data partition
+    call('disk.format', disk, 1)
+    parts = call('disk.list_partitions', disk)
+    # This 'assumes' 512 byte sectors
```
bmeagherix, line 36: Why assume?  Does `int(parts[0]['start'] / parts[0]['start_sector'])` yield the sector size?

tests/api2/test_disk_wipe.py
```diff
@@ -24,6 +22,40 @@ def test_disk_wipe_exported_zpool_in_disk_get_unused():
...
+    assert signal_msg in readback_presence
+
+    # Clean the drive
+    call('disk.wipe', disk, 'QUICK', job=True)
+
+    # Confirm it's now clean
+    readback_clean = ssh(f"dd if=/dev/{disk} iseek={seek_blk} count=1").splitlines()[0]
+    assert signal_msg not in readback_clean
+
+
```
bmeagherix, line 58: Could we also try to `disk.list_partitions` again, and ensure we get nothing this time?

tests/api2/test_disk_format.py
```diff
@@ -1,118 +1,166 @@
...
-    assert len(partitions) == 2
-    # As much swap as we could afford
-    assert swap_size * 0.9 < partitions[0]['size'] <= swap_size
-    # Data of at least the requested size
-    assert partitions[1]['size'] >= data_size
+    # Hand wavey swap size test
+    assert int(partitions[0]['size'] / (1024 ** 3) + 0.5) == 2
 
+    # The data partition should start after a 'grain_size' gap
+    assert partitions[1]['start_sector'] == partitions[0]['end_sector'] + grain_size
```
bmeagherix, line 120: Maybe also
`assert partitions[1]['partition_type'] == DATA_TYPE_UUID
`

tests/api2/test_disk_format.py
```diff
@@ -1,118 +1,166 @@
...
+
+def test_disk_format_removes_existing_partition_table(unused_disk):
+    """
+    Confirm we can repartion
+    """
+    assert unused_disk[2] != 0, 'ERROR: Should not run this test without a non-zero grain_size'
+    disk = unused_disk[0]
+
+    call('disk.format', disk['name'], 2)
+    call('disk.format', disk['name'], 0)
```
bmeagherix, line 166: Maybe call `disk.list_partitions` after each of these and do a basic sanity check (e.g. partition count / type check).

src/middlewared/middlewared/plugins/disk_/format.py
```diff
@@ -6,41 +6,84 @@
...
+            raise CallError(f'Unable to format {disk!r}: grainSize = {device.optimumAlignment.grainSize}')
+
+        drive_size_s = parted.sizeToSectors(dd['size'], 'B', device.sectorSize)
+
+        # Allocate space for the requested swap size
+        leave_free_space = parted.sizeToSectors(swap_size_gb, 'GiB', device.sectorSize)
+
+        # Minimum data partition size of 512 MiB is arbitrary
+        min_data_size = parted.sizeToSectors(512, 'MiB', device.sectorSize)
+        max_data_size = drive_size_s - leave_free_space
```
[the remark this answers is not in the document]
mgrimesix: I'll refactor those lines and try to make it more sensible.

tests/api2/test_disk_wipe.py
themylogin: We don't need to skip tests if some resources are unavailable. VM environment misconfiguration will leave those tests to be not executed silently (at least with our current Jenkins).

tests/api2/test_disk_wipe.py
mgrimesix: Sure.  I was being lazy and avoiding the block size command to dd

bmeagherix (approved):
Thanks for the changes! LGTM.

bmeagherix (approved):

yocalebo (changes requested):

PR #13212, opened 2024-02-22
NAS-127499 / 24.10 / add supports_identify_light key to enclosure2 namespace

yocalebo:
The UI team is slowly starting to implement and use `webui.enclosure.dashboard` endpoint. However, one of the keys that they were missing was the `has_slot_status` key. After doing some research, this key represents whether or not the enclosure slot supports the "identify" related commands. (i.e. lighting up the LED on the disk bay). The UI uses this boolean value to either show (or hide) the `Identify Drive` This adds back the equivalent functionality but instead uses a different key name to (in theory) be a little bit more intuitive on what it actually does.

NOTE: to future developers, it is a mistake to assume you know whether or not enclosure slot supports identification. This must be provided by platform team.

### DIFF
--- a/src/middlewared/middlewared/pytest/unit/plugins/enclosure/test-cases/M60-WITH-12-ES102/expected.json
+++ b/src/middlewared/middlewared/pytest/unit/plugins/enclosure/test-cases/M60-WITH-12-ES102/expected.json
@@ -27,7 +27,7 @@
                     "value": null,
                     "value_raw": 285212672,
                     "dev": "sdasw",
-                    "supports_drive_identify": true,
+                    "supports_identify_light": true,
                     "original": {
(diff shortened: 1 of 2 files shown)


bmeagherix (approved):
I'll approve but with two (minor?) caveats:
1.  I can't speak to the values of `SUPPORTS_IDENTIFY_KEY` for each platform.
2. Tests would be good.

yocalebo:
> I'll approve but with two (minor?) caveats:
> 
> 1. I can't speak to the values of `SUPPORTS_IDENTIFY_KEY` for each platform.
> 2. Tests would be good.

1. don't expect you to and I'm sure some of these values are wrong (i've asked someone in platform team to give me an update "list")
2. updated the CI tests

yocalebo:
backport

PR #13214, opened 2024-02-22
NAS-127488 / 24.10 / Add lightweight replacement for disk.list_partitions

mgrimesix:
Recent changes to `disk.wipe` added a call to `disk.list_partitions`.  It was later determined `disk.list_partitions` might cause excessive delay when used in wipe on large systems.  This PR adds a lightweight alternative and eliminates the call to `disk.list_partitions`.  The new routine is called `get_partitions_quick`.

Also, during testing it was discovered we are failing to update the kernel partition table after a `disk.wipe` call.  This behavior was introduced in Electric Eel.  The fix is to use the block ioctl `BLKRRPATH`.  This was already conveniently implemented in `retaste.py` and only required a simple wrapper and small bug fix.

Finally, added a CI test to `test_disk_wipe.py` to detect regression of the kernel partition table updates.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/wipe.py
+++ b/src/middlewared/middlewared/plugins/disk_/wipe.py
@@ -62,7 +62,10 @@ class DiskService(Service):
 
             if mode == 'QUICK':
                 # Get partition info before it gets destroyed
-                disk_parts = self.get_partitions_quick(dev)
+                try:
+                    disk_parts = self.get_partitions_quick(dev)
+                except Exception:
+                    disk_parts = {}
 
                 _32 = 32
                 for i in range(_32):
@@ -87,13 +90,15 @@ class DiskService(Service):
                 # The middle partitions often contain old cruft.  Clean those.
                 if len(disk_parts) > 1:
                     for partnum, sector_start in disk_parts.items():
-                        if 1 != partnum:
-                            # Start 2 MiB back from the start and 'clean' 2 MiB past, 4 MiB total
-                            os.lseek(f.fileno(), sector_start - (2 * CHUNK), os.SEEK_SET)
-                            for i in range(4):
-                                os.write(f.fileno(), to_write)
-                                if event.is_set():
-                                    return
+                        if partnum == 1:
+                            continue
+
+                        # Start 2 MiB back from the start and 'clean' 2 MiB past, 4 MiB total
+                        os.lseek(f.fileno(), sector_start - (2 * CHUNK), os.SEEK_SET)
+                        for i in range(4):
+                            os.write(f.fileno(), to_write)
+                            if event.is_set():
+                                return
                     # This is quick. We can reasonably skip the progress update
 
             else:
--- a/src/middlewared/middlewared/plugins/disk_/wipe.py
+++ b/src/middlewared/middlewared/plugins/disk_/wipe.py
@@ -1,15 +1,32 @@
 import asyncio
 import threading
 import os
+from glob import glob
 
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/disk_/wipe.py
```diff
@@ -37,7 +54,7 @@ def _wipe_impl(self, job, dev, mode, event):
 
             if mode == 'QUICK':
                 # Get partition info before it gets destroyed
-                disk_parts = self.middleware.call_sync('disk.list_partitions', dev)
+                disk_parts = self.get_partitions_quick(dev)
```
yocalebo, line 57: We'll need to handle when this crashes (i.e. `FileNotFoundError` and `ValueError`)

yocalebo (changes requested):

src/middlewared/middlewared/plugins/disk_/wipe.py
mgrimesix: That was my original approach, but...
I used `glob` to find the files.  This doesn't generate an error or exception if nothing is found.  It just returns an empty dictionary.   To keep it lightweight I decided to not add a separate existence test.
If you think it's better, I can add a test and raise the exception when the path is missing.
Let me know.

yocalebo: Sorry, I meant when you do these 2 operations.
```
       with open(f"/sys/block/{dev_name}/queue/logical_block_size", "r") as fd:
             sectsize = int(fd.read().strip())
```
It would be crazy to have it fail there, but I've seen some crazier scenarios 😄

yocalebo: So something like
```
try:
    with open(blah) as f:
        sect size = int(f.bla)
except FileNotFoundError, ValueError:
    pass
except Exception:
    self.logger.error('Unexpected failure trying to blah', exc_info=True)

**Review by yocalebo — APPROVED** [review]

File: src/middlewared/middlewared/plugins/disk_/wipe.py [hunk 2]
```diff
@@ -61,14 +86,15 @@ def _wipe_impl(self, job, dev, mode, event):
...
-                    for partition in disk_parts[1:]:
-                        # Start 2 MiB back from the start and 'clean' 2 MiB past, 4 MiB total
-                        os.lseek(f.fileno(), partition['start'] - (2 * CHUNK), os.SEEK_SET)
-                        for i in range(4):
-                            os.write(f.fileno(), to_write)
-                            if event.is_set():
-                                return
-                            # This is quick. We can reasonably skip the progress update
+                    for partnum, sector_start in disk_parts.items():
+                        if 1 != partnum:
```
anodos325, line 90:
```
if partnum == 1:
    continue
```
would avoid indentation change and reduce diff.

src/middlewared/middlewared/plugins/disk_/wipe.py
```diff
@@ -78,12 +104,22 @@ def _wipe_impl(self, job, dev, mode, event):
...
 
-        with open(disk_path, 'wb'):
-            # we overwrote partition label information by the time
-            # we get here, so we need to close device and re-open
-            # it in write mode to trigger udev to rescan the
-            # device for new information
-            pass
+        # The call to update_partition_table_quick can require retries
+        error = {}
+        retries = 4
```
anodos325, line 109: Maybe define this as a constant somewhere.

PR #13221, opened 2024-02-23
NAS-127452 / 24.10 / Retry starting kerberos during DS initialization

anodos325:
If networking is not fully configured, kinit attempt may fail with EAGAIN. Retry several times before giving up. Since the actual kinit has a timeout set for 30 seconds, this loop may end up being quite long-running and so convert directoryservices.initialize into a middleware job. The only direct caller for this private endpoint is smb.configure which is itself a job and so we don't have to worry about any potential timeouts here.

### DIFF
--- a/src/middlewared/middlewared/plugins/directoryservices.py
+++ b/src/middlewared/middlewared/plugins/directoryservices.py
@@ -1,3 +1,4 @@
+import asyncio
 import enum
 import os
 import struct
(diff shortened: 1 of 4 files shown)


anodos325:
Needs testing on HA before review.

src/middlewared/middlewared/plugins/directoryservices.py
```diff
@@ -215,8 +215,24 @@ async def get_last_password_change(self, domain=None):
 
         return {"dbconfig": stored_ts, "secrets": passwd_ts}
 
+    async def __kerberos_start_retry(self, retries=10):
+        while retries:
```
mgrimesix, line 219: On the infinitesimal chance `retries` skips `0` and goes negative.  Maybe `while retries >= 0`

mgrimesix (approved):

PR #13241, opened 2024-02-27
NAS-127547 / 24.10 / Fix middle partition clean processing

mgrimesix:
### Problem
Previously, the check was to skip the first partition, which would start at 1MiB and process all others.
This did not account for drives that that had partitions within 1MiB of the start of the drive.
### Fix
Add check for partition start and exclude those that are less than 30MiB

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/wipe.py
+++ b/src/middlewared/middlewared/plugins/disk_/wipe.py
@@ -91,8 +91,10 @@ class DiskService(Service):
 
                 # The middle partitions often contain old cruft.  Clean those.
                 if len(disk_parts) > 1:
-                    for partnum, sector_start in disk_parts.items():
-                        if partnum == 1:
+                    _30MiB = 30 * CHUNK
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/disk_/wipe.py
```diff
@@ -91,8 +91,10 @@ def _wipe_impl(self, job, dev, mode, event):
 
                 # The middle partitions often contain old cruft.  Clean those.
                 if len(disk_parts) > 1:
-                    for partnum, sector_start in disk_parts.items():
-                        if partnum == 1:
+                    _30MiB = 30 * CHUNK
+                    for sector_start in disk_parts.values():
+                        # Skip any partitions that start under 30 MiB
+                        if sector_start < _30MiB:
```
bmeagherix, line 97: Should we also skip any partitions that are in the last 32MB of the drive?  (I know only those starting less than 4MB before the end would cause a problem.)

bmeagherix (approved):

PR #13250, opened 2024-02-27
NAS-127357 / 24.10 / Retrieve PCI slots in a new endpoint so that they are easily consumed by UI

Qubad786:
## Context

When GPU passthrough is desired, there are essentially 2 steps which need to be done:

1. Add GPU which is to be used for passthrough via `system.advanced` namespace
2. Create PCI devices for the GPU in question based on PCI devices it is consuming and any other which fall in the same iommu group (this happens during VM creation wizard)

For (2), we have an endpoint `vm.device.get_pci_ids_for_gpu_isolation` which UI should be consuming to find out PCI devices which should be created when user says he wants X GPU to be added to a VM.

For (1) however, UI was consuming the same endpoint and passing it's return argument to `system.advanced.update` which was bound to fail.
A new endpoint has been added `system.advanced.get_gpu_pci_choices` which basically does the parsing UI was already doing at the moment and gives values which UI should be passing on for each GPU which is selected to `system.advanced.update_gpu_pci_ids` along with unit tests to ensure consistent behavior.

### DIFF
--- a/src/middlewared/middlewared/plugins/system_advanced/gpu.py
+++ b/src/middlewared/middlewared/plugins/system_advanced/gpu.py
@@ -1,5 +1,6 @@
-from middlewared.schema import accepts, List, returns, Str
+from middlewared.schema import accepts, Dict, List, returns, Str
 from middlewared.service import private, Service, ValidationErrors
+from middlewared.utils.gpu import get_gpus
 
 
 class SystemAdvancedService(Service):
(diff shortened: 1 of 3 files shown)


sonicaj (approved):

src/middlewared/middlewared/plugins/system_advanced/gpu.py
```diff
@@ -8,6 +9,22 @@ class Config:
         namespace = 'system.advanced'
         cli_namespace = 'system.advanced'
 
+    @accepts()
+    @returns(Dict(additional_attrs=True))
+    def get_isolate_gpu_pci_choices(self):
```
yocalebo, line 14: Method name is a bit strange. Let's just call it `get_gpu_pci_choices`

yocalebo (changes requested):

Qubad786:
Only to be merged after https://github.com/truenas/middleware/pull/13301

yocalebo (approved):

PR #13253, opened 2024-02-28
NAS-123837 / 24.10 / add ES24N JBoF to enclosure management

yocalebo:
This adds the ES24N JBoF to enclosure2 namespace. This lays the framework for allowing me to also add this enclosure to the legacy enclosure namespace for DF (24.04.{1/2}).

Tested and confirmed this works on our in-house system. Furthermore, because this is added to enclosure2 namespace, the `webui.enclosure.dashboard` endpoint just "magically" picks it up and works.


src/middlewared/middlewared/plugins/enclosure_/jbof_enclosures.py
```diff
@@ -0,0 +1,159 @@
...
+    model = uri = None
+    try:
+        chassis = rclient.chassis()
+    except Exception:
+        LOGGER.error('Unexpected failure enumerating chassis info', exc_info=True)
+        return model, uri
+
+    try:
+        for _, uri in chassis.items():
+            info = rclient.get(uri)
```
bmeagherix, line 139: I take it you're iterating over things and grabbing the '`Model`" because you don't trust the name '`2U24`' & want to make the code more generally applicable - a noble goal.

One issue with this is that with the ES24n the 3 children of Chassis are `2U24`, `IOM1` & `IOM2` and they **ALL** have a `Model` component, so the current code is succeeding by luck of ordering - so not really generally applicable.
```
>>> rclient.get('/redfish/v1/Chassis/2U24').json()['Model']
'VDS2249R2'
>>> rclient.get('/redfish/v1/Chassis/IOM1').json()['Model']
'X'
>>> rclient.get('/redfish/v1/Chassis/IOM2').json()['Model']
'X'
```

I wonder whether using the Product from the [ServiceRoot schema](https://redfish.dmtf.org/schemas/v1/ServiceRoot.v1_16_1.json) might be easier & more robust.
```
>>> rclient.get('/redfish/v1').json()['Product']
'ves-vds2249r-RFlex'
```

At least it **is** in the documented Redfish schema.

src/middlewared/middlewared/plugins/enclosure_/slot_mappings.py
```diff
@@ -4,7 +4,26 @@
...
+    nvme drives that are mounted via NVMe over Fabrics"""
+    if model == JbofModels.ES24N.name:
+        return {
+            'any_version': True,
+            'versions': {
+                'DEFAULT': {
+                    'model': {
+                        model: {
+                            i: {SYSFS_SLOT_KEY: i, MAPPED_SLOT_KEY: i, SUPPORTS_IDENTIFY_KEY: True}
+                            for i in range(1, 25)
```
bmeagherix, line 21: Just an observation.  SUPPORTS_IDENTIFY_KEY is True, but currently I don't see any code present in this PR to facilitate that.

yocalebo: @bmeagherix right, I haven't added that part (yet). Will do so after this one.

src/middlewared/middlewared/plugins/enclosure_/jbof_enclosures.py
yocalebo: Bummer, great catch. I'll check up on this

yocalebo: Wait, no this is wrong. I account for this in the next few lines. The `IOM` entries also have a model but their model value isn't in our `JbofModels` enum.
```
            if info.ok:
                 try:
                     model = JbofModels(info.json().get('Model', '')).name
                     break
                 except KeyError:
                     continue

bmeagherix (approved):

PR #13259, opened 2024-02-28
NAS-127232 / 24.10 / Make kerberos.py functions async compatible

mgrimesix:
Most of the routines in kerberos.py are declared `async`, however the routines themselves often contain blocking IO calls.
The purpose of this PR is to cleanup the async routines with regards to blocking IO calls.

### DIFF
--- a/src/middlewared/middlewared/plugins/kerberos.py
+++ b/src/middlewared/middlewared/plugins/kerberos.py
@@ -1081,7 +1081,7 @@ class KerberosKeytabService(CRUDService):
     @private
     async def _ktutil_list(self, keytab_file=keytab['SYSTEM'].value):
         """
-        async wrapper for validate
+        async wrapper for ktutil_list
         """
         return await asyncio.to_thread(
             self._ktutil_list_impl, keytab_file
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/kerberos.py
```diff
@@ -1227,14 +1235,15 @@ async def check_updated_keytab(self):
 
         old_mtime = 0
         ad_state = await self.middleware.call('activedirectory.get_state')
-        if ad_state == 'DISABLED' or not os.path.exists(keytab['SYSTEM'].value):
+        ad_keytab_exists = await self.middleware.run_in_thread(os.path.exists, keytab['SYSTEM'].value)
+        if ad_state == 'DISABLED' or not ad_keytab_exists:
```
bmeagherix, line 1239: This **_used_** to only check the file existence if `ad_state` was != 'DISABLED', so maybe change to 
```
if ad_state == 'DISABLED' or not (await self.middleware.run_in_thread(os.path.exists, keytab['SYSTEM'].value)):
    return
```
... or otherwise avoid adding effort, e..g.
```
if ad_state == 'DISABLED':
    return
ad_keytab_exists = await self.middleware.run_in_thread(os.path.exists, keytab['SYSTEM'].value)
if not ad_keytab_exists:
    return
```

bmeagherix (changes requested):

bmeagherix (approved):

PR #13261, opened 2024-02-29
Add a new endpoint to reboot standby controller

sonicaj:
This PR adds a new endpoint which can be used to reboot standby controller so this can be used in both upgrading HA logic and when FIPS changes are made - both nodes can be rebooted easily.


sonicaj:
backport

PR #13271, opened 2024-02-29
NAS-127627 / 24.10 / Try to use sql filters in audit.query calls

anodos325:
Most queries to this endpoint can be passed directly to sqlalchemy which provides a very large performance boost.

Additionally, users may specify filters that can be used to exclude some of our audit databases from the query results.
This commit makes us more intelligent about how we handle this situation (avoiding unnecessarily querying
databases).

### DIFF
--- a/src/middlewared/middlewared/plugins/audit/audit.py
+++ b/src/middlewared/middlewared/plugins/audit/audit.py
@@ -17,6 +17,8 @@ from .utils import (
     AUDIT_DEFAULT_FILL_WARNING,
     AUDIT_REPORTS_DIR,
     AUDITED_SERVICES,
+    parse_query_filters,
+    may_use_sql_filters,
 )
 from .schema.middleware import AUDIT_EVENT_MIDDLEWARE_JSON_SCHEMAS, AUDIT_EVENT_MIDDLEWARE_PARAM_SET
 from .schema.smb import AUDIT_EVENT_SMB_JSON_SCHEMAS, AUDIT_EVENT_SMB_PARAM_SET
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/audit/audit.py
```diff
@@ -199,21 +200,41 @@ async def query(self, data):
...
                     )
 
+        services_to_check, filters = parse_query_filters(
+            data['services'], data['query-filters'], sql_filters
+        )
+        if not services_to_check:
+            verrors.add(
+                'audit.query.query-filters',
+                'The combination of filters and specified services would result '
+                'in no databases being queries.'
```
themylogin, line 210:
```suggestion
                'in no databases being queried.'
```
?

src/middlewared/middlewared/plugins/audit/utils.py
```diff
@@ -55,4 +66,100 @@ def generate_audit_table(svc, vers):
...
+
+def may_use_sql_filters(
+    services: list,
+    filters_in: list,
+    filters_for_sql: list,
+    options: dict
+) -> bool:
+    if filters_in != filters_for_sql:
+        # We will need to do additional filtering after retrieval
+        return False
```
themylogin, line 144: In this (and all other cases) we still can use SQL filters, can't we? We'll only pass some of the filters (that are SQL-compatible) to SQLite, it will use indexes as much as possible, then we'll run additional filtering/sorting with Python.

anodos325: Right. I probably should name this better. Filtering gets complicated when dealing with queries such as:
```
[["event_data.credentials.credentials_data.token", "=", "xltISV7M7eBKEw0kOody5nqlC6QZqniGhugX_ZePOOhT42tErGVHyZoRZz-FkgvI"]]
```
Where field in SQL is JSON. SQlite3 actually has JSON support, but conversion of this sort of query is non-trivial and probably not worth our time. This method basically determining whether or not we need to do additional python filtering.

themylogin (approved):

PR #13274, opened 2024-02-29
NAS-127622 / 24.10 / Do not change `readonly` property when we are running in the installe…

themylogin:
…r, and it was not set yet


themylogin:
time 0:20

anodos325 (approved):

PR #13285, opened 2024-03-05
add MemorySizeMismatch Alert

yocalebo:
The performance team had an internal system that had a memory module die. This resulted in a memory size mismatch between the controllers. There was no indication in the UI that this had happened so this adds functionality to check the physical memory size between the controllers on an HA system and ensure they match and raise an alert otherwise.

### DIFF
--- a/src/middlewared/middlewared/alert/source/memory_errors.py
+++ b/src/middlewared/middlewared/alert/source/memory_errors.py
@@ -1,4 +1,5 @@
 from middlewared.alert.base import Alert, AlertCategory, AlertClass, AlertLevel, AlertSource
+from middlewared.alert.schedule import CrontabSchedule
 
 
 class MemoryErrorsAlertClass(AlertClass):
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/alert/source/memory_errors.py
```diff
@@ -10,7 +11,16 @@ class MemoryErrorsAlertClass(AlertClass):
     proactive_support = True
 
 
+class MemorySizeMismatchAlertClass(AlertClass):
+    category = AlertCategory.HARDWARE
+    title = 'Memory Size Mismatch Detected'
+    text = 'Memory size on this controller %(r1)d doesn\'t match other controller %(r2)d'
```
themylogin, line 17: Should we format these numbers using `humanfriendly`?

yocalebo: Great question, I went back and forth in my head about this and still don't know what I want to do 😄 I guess i'll use your review as an excuse to format the numbers.

themylogin (approved):
Thank you!

PR #13289, opened 2024-03-06
NAS-127697 / 24.10 / Allow RDMA capable interfaces to be exposed for regular networking

bmeagherix:
When JBOF support was added, we added *all* RDMA interfaces to `rdma.interface.internal_interfaces`.  Instead only add those currently in use for JBOF purposes.

Likewise, in `rdma.get_link_choices` omit any interfaces that have been configured for general networking.


bmeagherix:
@yocalebo I added `rdma.get_configured_interfaces`, but there could be an argument to move this into `interface.get_configured_interfaces` instead.  Any comment?

yocalebo (changes requested):

yocalebo (approved):

PR #13297, opened 2024-03-07
NAS-127729 / 24.10 / Update nfs syslog filter test

mgrimesix:
### Problem
This test has been plagued with consistent failure because the system under test randomly stalls and delays the critical syslog messages.  The stall can be several minutes.
### Fix
Increase the timeout for this test.
Add more retries with increasing delay between them.  The expectation is to prevent unnecessary delay if the messages are output in a timely manner.


bmeagherix (changes requested):
33 is a magic number sprinkled in a few places.  Make it defined once.

mgrimesix:
This is going to require a very different approach.
Closing this PR.

PR #13317, opened 2024-03-12
NAS-127807 / 24.10 / Add vrrp_sync_group when both IPv4 and IPv6 virtual addresses are pre…

bmeagherix:
…sent

Without this, the IPv6 address was being presented concurrently on both the ACTIVE and STANDBY controllers.

Note this issue only occurs when **both** IPv4 and IPv6 addresses are added to the **same** interface.

* See also https://github.com/acassen/keepalived/issues/2146


bmeagherix:
Note this issue only occurs when **both** IPv4 and IPv6 addresses are added to the **same** interface.

* See also https://github.com/acassen/keepalived/issues/2146

PR #13320, opened 2024-03-12
Backport sysdataset fixes

anodos325:
Per request from automation team, backport fixes for:
NAS-127724 - If for some reason a process has an open handle on our old system dataset, log an error message and proceed with a lazy umount to avoid leaving system in broken state.

NAS-127671 - Fix error message for errors umounting sysdataset


yocalebo (approved):

anodos325:
NOTE: commit was added to NAS tickets referenced in commit message above.

PR #13327, opened 2024-03-13
NAS-127836 / 24.10 / Enable Netdata Web

william-gr:
Netdata provides a full featured web interface to show all graphs available, however it does not provide authentication. To workaround this we set up a proxy with nginx using Basic HTTP Auth. Passwords are generated and should be valid for around 8 hours.

WebUI should be able to automatically make a XML HTTP Request with generated password to "login" to netdata and then open a new tab.

Ticket:	NAS-127836


src/middlewared/middlewared/plugins/reporting/netdata_web.py
```diff
@@ -0,0 +1,83 @@
...
+            authenticated_user = app.authenticated_credentials.user['username']
+        else:
+            raise CallError('This method needs to be called from an authenticated user only.')
+
+        if not os.path.exists(BASIC_FILE):
+            with open(os.open(BASIC_FILE, flags=os.O_CREAT, mode=0o640)):
+                shutil.chown(BASIC_FILE, 'root', 'www-data')
+
+        ht = HtpasswdFile(BASIC_FILE, autosave=True, default_scheme='bcrypt')
+        password = ''.join(random.choice(
```
yocalebo, line 41: I've created a helper function that does this. It's in utils.crypto

william-gr: Sweet, changed

anodos325:
This doesn't look like a powerpoint presentation.

themylogin (commented):
We can easily add integration tests for both these functions.

william-gr:
@themylogin added, is that what you were thinking?

yocalebo (approved):

themylogin (approved):
We can also add the test for the expire function (you'll need to put mock value using `cache.get` before).

anodos325:
Is it worthwhile to more tightly associate the password with the authenticated middleware session (e.g. remove the randomized password on session logout)?

william-gr:
Added readonly access and test.

I dont see the point in associate with middleware session because then we will require our webui session to stay open, but I can do it if requested.

src/middlewared/middlewared/plugins/reporting/netdata_web.py
```diff
@@ -0,0 +1,82 @@
...
+    @returns(Str('password'))
+    @pass_app()
+    def netdataweb_generate_password(self, app):
+        """
+        Generate a password to access netdata web.
+        That password will be stored in htpasswd format for HTTP Basic access.
+        """
+        # Password schema is not used here because for READONLY_ADMIN
+        # will make it return "******" instead, breaking this method for that role.
+        if app.authenticated_credentials.is_user_session:
```
anodos325, line 32: `app` may be None if for some reason we ever call this internally.

src/middlewared/middlewared/plugins/reporting/netdata_web.py
```diff
@@ -0,0 +1,86 @@
...
+            raise CallError('This method needs to be called from an authenticated user only.')
+
+        if not os.path.exists(BASIC_FILE):
+            with open(os.open(BASIC_FILE, flags=os.O_CREAT, mode=0o640)):
+                shutil.chown(BASIC_FILE, 'root', 'www-data')
+
+        with HTPASSWD_LOCK:
+            ht = HtpasswdFile(BASIC_FILE, autosave=True, default_scheme='bcrypt')
+            password = generate_string(16, punctuation_chars=True)
+            ht.set_password(authenticated_user, password)
```
anodos325, line 46: I think at a minimum we should check whether the user already exists in the htpasswd file, and generate a debug message so that we have some insight into a situation that may break someone's netdata access.

src/middlewared/middlewared/plugins/reporting/netdata_web.py
```diff
@@ -0,0 +1,86 @@
...
+        return BASIC_FILE
+
+    @cli_private
+    @accepts(roles=['READONLY_ADMIN'])
+    @returns(Str('password'))
+    @pass_app()
+    def netdataweb_generate_password(self, app):
+        """
+        Generate a password to access netdata web.
+        That password will be stored in htpasswd format for HTTP Basic access.
```
anodos325, line 30: We should note that concurrent access to netdata UI for same credentials is not supported and may lead to undesired behavior.

src/middlewared/middlewared/plugins/reporting/netdata_web.py
william-gr: Added a warning but can change to a debug?

src/middlewared/middlewared/plugins/reporting/netdata_web.py
william-gr: Added!

anodos325 (approved):

PR #13329, opened 2024-03-13
NAS-127803 / 24.10 / Add optional force parameter to jbof.delete API

bmeagherix:
This will permit the JBOF to be removed even if all communication with the JBOF has been lost.

(Did a minor refactor of parameters to `ensure_redfish_client_cached` while replicating the call, but another upcoming PR will take things further.)

### DIFF
--- a/src/middlewared/middlewared/plugins/jbof/crud.py
+++ b/src/middlewared/middlewared/plugins/jbof/crud.py
@@ -5,10 +5,10 @@ import time
 import middlewared.sqlalchemy as sa
 from middlewared.plugins.jbof.redfish import (InvalidCredentialsError,
                                               RedfishClient)
-from middlewared.schema import (Dict, Int, IPAddr, Password, Patch, Str,
+from middlewared.schema import (Bool, Dict, Int, IPAddr, Password, Patch, Str,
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -208,18 +208,28 @@ async def do_update(self, id_, data):
...
         """
         # Will make a best-effort un tear down existing connections / wiring
         # To do that we first need to fetch the config.
         data = await self.get_instance(id_)
-        await self.middleware.run_in_thread(self.ensure_redfish_client_cached,
-                                            data['mgmt_ip1'],
-                                            data.get('mgmt_username'),
-                                            data.get('mgmt_password'))
+
+        if force:
```
yocalebo, line 220: Why not
```
try:
    self.middleware.run_in_thread(blah)
except Exception as e:
    if force:
        self.logger.blah
    else:
        raise e

bmeagherix: Agreed that looks better.  Thanks!

yocalebo (approved):

PR #13333, opened 2024-03-14
NAS-127851 / 24.10 / Update flush_chain_INPUT to flush both ip4 and ip6 tables

bmeagherix:
It is NOT obvious from the nfs man page, but it appears that if a family is not specified for a flush chain command then it defaults to ip.

(The man page does document similar behavior for TABLES.)


yocalebo (approved):
Gotta love abysmally bad man pages 😄

PR #13341, opened 2024-03-14
NAS-127856 / 24.10 / Detect and save mgmt_ip2 if not supplied during jbof.create

bmeagherix:
We want to save the `mgmt_ip2` in the database (even if not supplied during the create), so that we can add robustness later.  (Splitting this functionality out to backport to 24.04, but won't avail of it until 24.10)

### DIFF
--- a/src/middlewared/middlewared/plugins/jbof/crud.py
+++ b/src/middlewared/middlewared/plugins/jbof/crud.py
@@ -173,6 +173,26 @@ class JBOFService(CRUDService):
             verrors.add('jbof_create.mgmt_ip1', f'Failed to add JBOF: {e}')
         verrors.check()
 
+        # If the caller just supplied mgmt_ip1, let's fetch mgmt_ip2 to store in the DB
+        if data.get('mgmt_ip2') in ['', None]:
+            try:
+                # Grab the IOM redfish IPs
+                redfish = RedfishClient.cache_get(mgmt_ip)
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/jbof/crud.py
```diff
@@ -173,6 +173,26 @@ async def do_create(self, data):
             verrors.add('jbof_create.mgmt_ip1', f'Failed to add JBOF: {e}')
         verrors.check()
 
+        # If the caller just supplied mgmt_ip1, let's fetch mgmt_ip2 to store in the DB
+        if data.get('mgmt_ip2') in ['', None]:
+            try:
+                # Grab the IOM redfish IPs
+                redfish = RedfishClient.cache_get(mgmt_ip)
```
yocalebo, line 180: Since we're in a coroutine and we're spinning up a few single-use threads, let's add a private synchronous method to this class that does all the redfish checks and run it all in a single thread.

i.e.
```
@private
def get_mgmt_ips(blah):
    try:
        redfish = RedfishClient.cache_get(blah)
        blah


async def do_create
    if data.get('mgmt_ip2') in ('', None):
        await self.middleware.run_in_thread(self.get_mgmt_ips, blah)

yocalebo (changes requested):

src/middlewared/middlewared/plugins/jbof/crud.py
bmeagherix: Thanks, much better!

yocalebo (approved):

PR #13342, opened 2024-03-15
NAS-127771 / 24.10 / Fix `KeyError` in `rdma.get_link_choices`

themylogin:
```
  File "/usr/lib/python3/dist-packages/middlewared/plugins/rdma/rdma.py", line 69, in get_link_choices
    result.append({'rdma': link['ifname'], 'netdev': link['netdev']})
                                                     ~~~~^^^^^^^^^^
KeyError: 'netdev'
```

### DIFF
--- a/src/middlewared/middlewared/plugins/rdma/rdma.py
+++ b/src/middlewared/middlewared/plugins/rdma/rdma.py
@@ -78,7 +78,8 @@ class RDMAService(Service):
 
         result = []
         for link in json.loads(ret.stdout.decode()):
-            result.append({'rdma': link['ifname'], 'netdev': link['netdev']})
+            if 'netdev' in link:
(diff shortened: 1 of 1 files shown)


themylogin:
@bmeagherix however, I am not sure what does the absence of this key really mean, should this may be handled  differently?

themylogin:
time 0:20

bmeagherix (approved):
Thanks for the coded change.

I _would_ like to know more details about the device in question ... what type of RDMA capable interface is **not** providing a `netdev`.

Can you provide details?

It also would be good to see the output from `rdma link` and `rdma dev`, and grep output from _/var/log/kern.log_ pertaining to the device.

themylogin:
@bmeagherix can you please ask the user in the issue comments? He could not attach a debug because his system was not booting due to this issue.

PR #13353, opened 2024-03-15
NAS-127868 / 24.10 / NAS-127868 - Hide Passwords for SPICE Viewer

kmoore134:
Add a bit of JS to the SPICE VM Viewer to ensure that passwords are displayed with *** instead of plaintext.


kmoore134:
Updates from javascript prompt(); to a prompt that properly hides typed passwords. 

image

undsoft (commented):
So, looking at source code in the original repo https://gitlab.freedesktop.org/spice/spice-html5/-/blob/master/spice.html?ref_type=heads, I can't see `prompt` anywhere there.
Can't we update the package instead of overriding files?

kmoore134:
You're looking at the wrong file, its here:

https://gitlab.freedesktop.org/spice/spice-html5/-/blob/master/spice_auto.html?ref_type=heads#L73

I will see if I can send you a proper diff though, so you can see what was added.

undsoft (approved):
For what this PR does, it looks okay.
Overall code quality for `spice-html5` package is not great.

yocalebo:
retest this please

themylogin:
We should probably document where do we get these files from, so they can be updated appropriately in the future.

If we patch the files, it would be a good idea to also store the patch so we can re-apply it if we pull an updated version.

kmoore134:
@themylogin That is fair. Where would you recommend stashing the patch file, and do you want the comment in the file itself?

themylogin:
@kmoore134 I think we can store it along with these files, and we can provide original URLs in a comment in the patch file header

kmoore134:
> @kmoore134 I think we can store it along with these files, and we can provide original URLs in a comment in the patch file header

Done

kmoore134:
retest this please

themylogin (approved):
@kmoore134 LGTM, thank you!

kmoore134:
retest this please

PR #13362, opened 2024-03-15
NAS-127873 / 24.10 / Improve error messages for disable-rootfs-protection

anodos325:
Indicate to user that this requires being privileged user and account for possibility that /usr/sbin is not in user's PATH.

### DIFF
--- a/src/freenas/usr/local/libexec/disable-rootfs-protection
+++ b/src/freenas/usr/local/libexec/disable-rootfs-protection
@@ -10,6 +10,7 @@ from pathlib import Path
 from subprocess import run
 
 
+ZFS_CMD = '/usr/sbin/zfs'
 TO_CHMOD = ['apt', 'dpkg']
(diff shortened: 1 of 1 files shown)


src/freenas/usr/local/libexec/disable-rootfs-protection
```diff
@@ -75,7 +76,14 @@ if __name__ == '__main__':
     datasets = []
     args = process_args()
 
-    rv = run(['zfs', 'get', '-o', 'value', '-H', 'truenas:developer', '/'], capture_output=True)
+    if not os.getuid() == 0:
```
yocalebo, line 79: Looks weird. Why not os.getuid() != 0

anodos325: because Friday afternoon

yocalebo (approved):

PR #13363, opened 2024-03-15
Prevent ISCSIPortalIPAlert for HA STANDBY with ALUA

bmeagherix:
When `ISCSIPortalIPAlertSource.check` runs on the STANDBY node of a HA pair-while ALUA has been enabled, it will present the Alert with all of the used Virtual IPs.  Filter them out.

### DIFF
--- a/src/middlewared/middlewared/alert/source/iscsi.py
+++ b/src/middlewared/middlewared/alert/source/iscsi.py
@@ -37,5 +37,17 @@ class ISCSIPortalIPAlertSource(AlertSource):
                     )
                 )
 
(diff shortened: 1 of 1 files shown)


anodos325 (approved):

yocalebo (changes requested):
Why not `run_on_backup_node = False`?

bmeagherix:
> Why not `run_on_backup_node = False`?

Because if ALUA is enabled then (IMO) it _does_ make sense to check things on the STANDBY node.  In that situation, we _are_ serving iSCSI from the  STANDBY node, why would we **not** check the portals?

PR #13367, opened 2024-03-18
NAS-127150 / 24.10 / Add new endpoint to retrieve truecommand status

sonicaj:
This PR adds changes to add a new endpoint which can be used to retrieve if system is being managed by Truecommand.

### DIFF
--- a/src/middlewared/middlewared/plugins/truenas.py
+++ b/src/middlewared/middlewared/plugins/truenas.py
@@ -7,7 +7,7 @@ from ixhardware import TRUENAS_UNKNOWN, get_chassis_hardware
 
 from middlewared.plugins.truecommand.enums import Status as TrueCommandStatus
 from middlewared.schema import accepts, Bool, Dict, Patch, returns, Str
-from middlewared.service import cli_private, job, no_auth_required, pass_app, private, Service, throttle
+from middlewared.service import cli_private, job, no_auth_required, private, Service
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/truenas.py
```diff
@@ -41,7 +46,23 @@ class TruenasCustomerInformationModel(sa.Model):
 class TrueNASService(Service):
 
     class Config:
-        cli_namespace = "system.truenas"
+        cli_namespace = 'system.truenas'
+
+    @no_auth_required
+    @throttle(seconds=2, condition=throttle_condition)
```
yocalebo, line 52: Please remove `@throttle`, it is known to cause undefined behavior.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/truenas.py
```diff
@@ -41,7 +46,23 @@ class TruenasCustomerInformationModel(sa.Model):
...
+    @accepts()
+    @returns(Bool())
+    @pass_app()
+    async def managed_by_truecommand(self):
```
yocalebo, line 56: We don't need `@pass_app` here

yocalebo (changes requested):

yocalebo (approved):
Thanks! Don't forget to open up UI ticket

PR #13384, opened 2024-03-19
NAS-119224 / 24.10 / Properly manage system dataset UUIDs on HA

themylogin:


themylogin:
time 3:00

src/middlewared/middlewared/plugins/failover.py
```diff
@@ -359,6 +357,10 @@ def sync_to_peer(self, options):
         `reboot` as true will reboot the other controller after syncing.
         """
         standby = ' standby controller.'
+
+        self.logger.debug('Pulling system dataset UUID from' + standby)
```
yocalebo, line 361: This seems incorrect? Does this work with `+ standby`? Or should it be `UUID from %s', standby)`

themylogin: You wrote all the remaining code of this function using this style, I just copied it. It works perfectly.

src/middlewared/middlewared/plugins/sysdataset.py
```diff
@@ -122,36 +120,57 @@ def sysdataset_path(self, expected_datasetname=None):
...
+        remote_uuid_key = 'uuid_b'
+        if await self.middleware.call('failover.node') == 'B':
+            remote_uuid_key = 'uuid'
+
+        local_config = await self.middleware.call('datastore.config', 'system.systemdataset', {'prefix': 'sys_'})
+        if local_config[remote_uuid_key]:
+            self.logger.debug('We already know the standby controller system dataset UUID')
+            return
+
+        remote_config = await self.middleware.call(
```
yocalebo, line 158: This could crash since it's making a call to the remote system.

themylogin: Yes, but it's a part of `sync_to_peer` which will fail anyway if the remote node is down

anodos325:
Should we add some tests?

themylogin:
@anodos325 what do you propose to test?

yocalebo (approved):

PR #13406, opened 2024-03-23
NAS-127981 / 24.10 / Restrict changes to SMB aux parameters to FULL_ADMIN

anodos325:
There are various parameters in which the administrator can provide arbitrary commands and scripts to run via auxiliary parameters, some of which may be executed as the root user. This can present a risk of a limited administrator accidentally or intentionally performing actions on the NAS with privileges that are higher than intended.


yocalebo (commented):
Easy enough to add a check to one of the existing tests, right? Or even a 1-off test that ensures non-admins can't add aux params?

mgrimesix:
Is there a CI test for this?

mgrimesix (approved):
Please add a CI test.

PR #13407, opened 2024-03-25
NAS-127975 / 24.10 / Fix cloud sync pre-/post-scripts

themylogin:


themylogin:
time 4:00

src/middlewared/middlewared/plugins/cloud_sync.py
```diff
@@ -271,8 +271,10 @@ async def run_script(job, env, hook, script_name):
...
         return
 
-    if not hook.startswith("#!"):
-        hook = f"#!/bin/bash\n{hook}"
+    if hook.startswith("#!"):
+        shebang = shlex.split(hook.splitlines()[0][2:].strip())
+    else:
+        shebang = ["/bin/bash"]
 
     fd, name = tempfile.mkstemp()
```
yocalebo: I also asked that you fix up the entirety of this section by:
1. using the temfile context manager
2. and preventing all of this blocking I/O from being executed in a coroutine

yocalebo (changes requested):

src/middlewared/middlewared/plugins/cloud_sync.py
yocalebo: your comment doesn't matter. Python's builtin `open` and `os.chmod` don't release the GIL so it doesn't matter what the backing filesystem is.

themylogin: @yocalebo it does, ramfs operations will never block since there is nothing to block on. We take advantage of this all over the codebase.

yocalebo: @themylogin just for clarity. The backing filesystem doesn't matter. It matters whether or not python's builtin primitives release the GIL for the I/O based calls. To my surprise, `os.chown` `os.open` and at least most of the file I/O based primitives actually do release the GIL.....that's peculiar to me.

yocalebo (approved):

PR #13412, opened 2024-03-25
NAS-128021 / 13.3 / Fix collectd-daemon RC script

anodos325:
There are various combinations of running state for the daemon
process monitoring the collectd process and the process that it
is monitoring. In some pathological cases the original script
could hang for up to 60 seconds on 13.0 and indefinitely in 13.3

This commit addresses the problem by doing the following:

* consolidates the pre-stop and post-stop functions into a single
  stop function

* convert the rc script to use standard rc.subr functions to get pids

* validate that have pids we're trying to wait for

* Send a SIGKILL instead of SIGTERM to daemon process if the monitored
  process has gone away.

### DIFF
--- a/src/freenas/usr/local/etc/rc.d/collectd-daemon
+++ b/src/freenas/usr/local/etc/rc.d/collectd-daemon
@@ -27,8 +27,7 @@ pidfile=${collectd_daemon_pidfile:="/var/run/collectd-daemon.pid"}
 command="/usr/sbin/daemon"
 
 start_cmd=collectd_daemon_start
-stop_precmd=collectd_daemon_pre_stop
-stop_postcmd=collectd_daemon_post_stop
+stop_cmd=collectd_daemon_stop
 
 collectd_daemon_start()
 {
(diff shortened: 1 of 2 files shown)


src/freenas/usr/local/etc/rc.d/collectd-daemon
```diff
@@ -27,29 +27,51 @@ pidfile=${collectd_daemon_pidfile:="/var/run/collectd-daemon.pid"}
...
+	    kill -TERM "${collectd_pid}"
+	fi
+    elif [ -n "${d_pid}" ]; then
+	# We have a daemon process, but the process its monitoring has already died
+	# Send SIGKILL to prevent it from waiting long enough to try to restart the
+	# collectd process (in the best situation) or hanging forever (in worst case).
+        debug "killing: ${d_pid}"
+        kill -KILL "${d_pid}"
+    else
+	echo "collectd-daemon is already stopped"
```
yocalebo, line 69: Do you think this will be spammed in logs because of how we've implemented service mgmt on CORE?

anodos325: Good point. I'll switch to debug.

yocalebo (approved):

PR #13414, opened 2024-03-25
py-middlewared: use devel/py-bsd present in ports tree

rkojedzinszky:


rkojedzinszky:
py-bsd is present in ports tree, and also cloned using git during build. Some packages depend on devel/py-bsd, while middlewared depends on freenas/py-bsd. The later has different version string, howewer right now the very same commit is checked out. Thus, I propose to remove the confusing and redundant port.

anodos325:
py-bsd is a python package we internally have developed and so it's better to not have our changes collide with what the upstream port maintainer (not us) is doing. We're currently in-sync with them, but that's not guaranteed to be the case.

PR #13419, opened 2024-03-26
NAS-127978 / 24.10 / Allow NFS client to maintain file system processing across an HA failover

mgrimesix:
### The Problem:
NFS clients that were performing 'continuous' write operations on HA systems were losing connection on failover events.
The error received by the client was `permission denied`
### The Cause:
We consider NFS to be a critical service.  As such it remains active on the standby node, but network activity is blocked.  In failover processing, the standby node enables networking before the system is fully ready to resume normal processing.  It is during this time that the continuous writing client attempts file system activity and receives the `permission denied` error.
### The Fix:
Disabling mountd on the standby node prevents the NFS clients from attempting ill fated file system activity on failover events.  We currently restart all processes when the system is ready to resume full file system processing which safely restarts all NFS processing.


mgrimesix:
Checking on the possibility of a CI test before removing the WIP label.

bmeagherix (approved):
I'll approve, but I wonder whether it would be possible to write a unit test to test this corrected behavior.

PR #13420, opened 2024-03-26
NAS-128040 / 24.10 / Fix aux parameter comment parsing in services->SMB

anodos325:
Since the initial angelfish release in order to facilitate SMB clustering we have stored Samba's global section (services->SMB) in the samba registry via libsmbconf. 

Libsmbconf is more strict at validation of what gets written in the configuration. Some core users have copious smb_options with many comments in them. This shifts the function that was used for stripping comments from auxiliary parameters from sharing.smb to a utility function for consumption by both sharing.smb and services.smb.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb_/utils.py
+++ b/src/middlewared/middlewared/plugins/smb_/utils.py
@@ -1,0 +1,12 @@
+def smb_strip_comments(auxparam_in):
+    parsed_config = ""
+    for entry in auxparam_in.splitlines():
+        if entry == "" or entry.startswith(('#', ';')):
+            continue
+
+        # For some reason user may have added more comments after the value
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/smb_/utils.py
```diff
@@ -0,0 +1,12 @@
+def smb_strip_comments(auxparam_in):
+        parsed_config = ""
```
mgrimesix, line 2: Is there a reason for the extra large indentation?

anodos325: Yes, I forgot to shift it back by four after moving it to new file :)

mgrimesix (approved):

PR #13423, opened 2024-03-27
NAS-127943 / 24.10 / Cloud backup API

themylogin:


themylogin:
time 6:00

src/middlewared/middlewared/plugins/cloud_backup/snapshot.py
```diff
@@ -0,0 +1,101 @@
...
+                Str("name"),
+                Str("path"),
+                Str("type", enum=["dir", "file"]),
+                Int("size"),
+                Datetime("mtime"),
+                additional_attrs=True,
+            ),
+        ]),
+    )
+    def list_snapshot_directory(self, id_, snapshot_id, path):
```
anodos325, line 70: Have we tested what performance is like when dir contains 100K files? 1 million files? Does it need to be a job?

src/middlewared/middlewared/plugins/cloud_backup/restore.py
```diff
@@ -0,0 +1,44 @@
...
+
+    class Config:
+        cli_namespace = "task.cloud_backup"
+        namespace = "cloud_backup"
+
+    @accepts(
+        Int("id"),
+        Str("snapshot_id", validators=[NotMatch(r"^-")]),
+        Str("subfolder"),
+        Str("destination_path", validators=[NotMatch(r"^-")]),
```
anodos325, line 17: Are these local filesystem paths? Do we need to prevent people from doing silly things like overwriting `/etc`?

src/middlewared/middlewared/plugins/cloud_backup/snapshot.py
```diff
@@ -0,0 +1,101 @@
...
+                "cloud_backup_snapshot",
+                Str("id"),
+                Str("hostname"),
+                Datetime("time"),
+                List("paths", items=[Str("path")]),
+                additional_attrs=True,
+            ),
+        ]),
+    )
+    def list_snapshots(self, id_):
```
anodos325, line 30: Have we tested under large snapshot counts to see whether this needs to be a job?

src/middlewared/middlewared/plugins/cloud_backup/snapshot.py
themylogin: It's not a job in the original cloud sync plugin and it never caused any issues. People don't use this to list directories with 100k files (and we can't offer them to do so since no pagination is available).

src/middlewared/middlewared/plugins/cloud_backup/snapshot.py
themylogin: 10000 snapshots take 30 seconds to list, I doubt any reasonable use will go above that.

yocalebo: Are these 10k "restic" snapshots or zfs snapshots?

themylogin: Restic snapshots, ZFS snapshots are not involved here

yocalebo (approved):
I'm good with it but wait for @anodos325 final approval before merge, please

PR #13437, opened 2024-03-29
NAS-128100 / 24.10 / Switch iSCSI tests from REST to websocket

bmeagherix:
- Better to use websocket as this is what the UI uses
- the REST test functions (`GET`, `POST`, `PUT`, `DELETE`) never use the VIP in a HA config (and are therefore susceptible to breakage when Node A is the STANDBY node).


yocalebo (approved):
makes me happy seeing this 😄

PR #13440, opened 2024-03-29
NAS-129269 / 24.10 / Make rsyslog support IPv6 destinations

Leseratte10:
I noticed that the current rsyslog-ng implementation was buggy and didn't support IPv6 destinations. 

Entering an IPv6 IP as the syslog server failed due to input validation (both with the [] around the IP and without), and entering a hostname still only made it connect using IPv4. 

This PR should fix that issue. 

The regex for the input validation in the syslog server textbox has been changed to allow colons, percent signs and brackets [] as well which are needed for IPv6 adresses, and the template that creates the syslog-ng config has been modified to add the "ip-protocol(6) parameter. 

According to the documentation this parameter defaults to "4" making it only send data to IPv4 destinations, but setting it to "6" makes it support both IPv4 and IPv6 destinations, under the condition that the OS kernel supports IPv6 (i.e., a "ping6 ::1" must work, which is probably the case for all Truenas installations). 

My tests seem to confirm that, so with that change syslog-ng can send to both IPv4 and IPv6 destinations (both using a direct IP or using a hostname).

### DIFF
--- a/src/middlewared/middlewared/etc_files/syslog-ng/syslog-ng.conf.mako
+++ b/src/middlewared/middlewared/etc_files/syslog-ng/syslog-ng.conf.mako
@@ -3,10 +3,13 @@ logger = middleware.logger
 
 def generate_syslog_remote_destination(advanced_config):
     result = ""
-    if ":" in advanced_config["syslogserver"]:
-        host, port = advanced_config["syslogserver"].rsplit(":", 1)
+    syslog_server = advanced_config["syslogserver"]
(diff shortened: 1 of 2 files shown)


Leseratte10:
Any update here yet?

yocalebo:
@Leseratte10 sorry for the delay. I've assigned a ticket to this and have put it into the queue for review. Hoping to get to this soon!

themylogin (approved):
Thank you!

PR #13441, opened 2024-03-30
VMs: Consider ::1 a localhost IP as well

Leseratte10:
Currently, the dropdown for the VMs Display bind address allows you to choose between 0.0.0.0 (all IPv4), :: (all IPv6) and 127.0.0.1 (localhost IPv4). ::1 (localhost IPv6) is not an option, so it's not possible to have a VM whose display is reachable over IPv6 but only from the local machine. 

This PR changes that, by making the code add "::1" to the list of loopback IPs in network.py as well, instead of just 127.0.0.1. 

Related to #13439


Qubad786 (approved):
Thank you @Leseratte10 for your contribution

PR #13443, opened 2024-03-31
NAS-123519 / 24.10 / Add globally unique system ID

Qubad786:
### Context
We needed a way to uniquely identify ours systems with some IDs generated and store one time in the configuration database

### Solution
Added a new table in database which holds uuid4 value saved in it on initial install of the system or upgrade of the system.

### DIFF
--- a/src/middlewared/middlewared/migration/0009_system_global_id.py
+++ b/src/middlewared/middlewared/migration/0009_system_global_id.py
@@ -1,0 +1,10 @@
+import uuid
+
+
+async def migrate(middleware):
+    await middleware.call(
+        'datastore.insert',
+        'system.globalid', {
+            'system_uuid': str(uuid.uuid4())
+        }
+    )
--- a/src/middlewared/middlewared/alembic/versions/24.10/2024-03-21_12-15_add_global_system_id.py
+++ b/src/middlewared/middlewared/alembic/versions/24.10/2024-03-21_12-15_add_global_system_id.py
@@ -1,0 +1,28 @@
+"""Add global system ID
+
+Revision ID: 14974a858948
+Revises: 7836261b2f64
+Create Date: 2024-03-21 12:15:50.886820+00:00
+
+"""
+from alembic import op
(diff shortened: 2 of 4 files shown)


src/middlewared/middlewared/migration/0009_system_global_id.py
```diff
@@ -0,0 +1,10 @@
+import uuid
+
+
+async def migrate(middleware):
+    await middleware.call(
```
yocalebo, line 5: Is this guaranteed this will only ever be called once for the lifetime of the install? Either way, we should probably check if there is a UUID before inserting here always.

yocalebo (changes requested):

themylogin:
Do we need something more complex to prevent issues like people getting duplicate IDs when importing existing config on a different machine?

Qubad786:
> Do we need something more complex to prevent issues like people getting duplicate IDs when importing existing config on a different machine?

@themylogin if an existing config is imported on a different machine, that machine is technically the same now or is that not the case?

themylogin:
> > Do we need something more complex to prevent issues like people getting duplicate IDs when importing existing config on a different machine?
> 
> @themylogin if an existing config is imported on a different machine, that machine is technically the same now or is that not the case?

@yocalebo your feedback is needed here

yocalebo:
> > > Do we need something more complex to prevent issues like people getting duplicate IDs when importing existing config on a different machine?
> > 
> > 
> > @themylogin if an existing config is imported on a different machine, that machine is technically the same now or is that not the case?
> 
> @yocalebo your feedback is needed here

That's correct, this id is _not_ hardware dependent so this is the intended behavior.

src/middlewared/middlewared/migration/0009_system_global_id.py
yocalebo: okay

yocalebo (approved):
Please wait for @themylogin final approval before merge, thanks.

PR #13450, opened 2024-04-02
NAS-123721 / 24.10 / make runtest.py work with new jenkins

yocalebo:
This makes the necessary changes to allow the `runtest.py` script work on our new jenkins infrastructure. There were an innumerable amount of issues (I'm exaggerating obviously) that were found while testing these changes so I'll list the most important ones:
1. our websocket client function `call` was never using the VIP on an HA system
2. because of this fact, the `log_test_name_to_middlewared_log` was causing the entire HA CI run to fail on the iSCSI tests. The reason why this was failing is because there is a test that fails over from the A to B node. This function was making a direct connection to the B node as well as the A node and so the connection was raising socket.timeout errors (as expected). This resolves that underlying issue by using the VIP (when appropriate)
3. The entirety of our REST helper functions (POST, GET, PUT, DELETE, etc) DO NOT use the VIP at all. To make matters worse, if a test fails over from the A to B nodes, the REST functions just don't even work. This is _NOT_ fixed in this PR and will be done so in another, however, we do fix the websocket issues
4. Add the bare minimum amount of new arguments to `runtest.py` so it can be called against an HA "pair" so that it can further configure the HA system and run tests
5. Simplify the logic in some of the early tests so that it's very clear what IP address(es) are being used

There are more issues to resolve but this needs to be merged as it is so that the PR doesn't grow too unwieldy.

### DIFF
--- a/src/middlewared/middlewared/test/integration/utils/client.py
+++ b/src/middlewared/middlewared/test/integration/utils/client.py
@@ -10,6 +10,7 @@ from middlewared.client import Client
 from middlewared.client.utils import undefined
 
 __all__ = ["client", "host", "host_websocket_uri", "password", "session", "url", "websocket_url"]
+IS_HA = os.environ.get('SERVER_TYPE') == 'ENTERPRISE_HA'
 
 
 @contextlib.contextmanager
@@ -17,20 +18,31 @@ def client(*, auth=undefined, auth_required=True, py_exceptions=True, log_py_exc
     if auth is undefined:
         auth = ("root", password())
 
+    uri = host_websocket_uri(host_ip)
     try:
-        with Client(host_websocket_uri(host_ip), py_exceptions=py_exceptions, log_py_exceptions=log_py_exceptions) as c:
+        with Client(uri, py_exceptions=py_exceptions, log_py_exceptions=log_py_exceptions) as c:
             if auth is not None:
                 logged_in = c.call("auth.login", *auth)
                 if auth_required:
                     assert logged_in
             yield c
-    except socket.timeout:
-        fail('socket timeout')
+    except socket.timeout as e:
+        if IS_HA:
+            # don't fail the entire run on HA systems since
+            # we perform failovers to and from each controller
+            # so timeouts could happen (which are expected)
+            raise e from None
+        else:
+            fail(f'socket timeout on URI: {uri!r} HOST_IP: {host_ip!r}')
 
 
 def host():
     if "NODE_A_IP" in os.environ:
+        # this is not to be confused with HA systems
+        # as it should only be set on NON-HA environments
         return os.environ["NODE_A_IP"]
+    elif IS_HA:
+        return os.environ["virtual_ip"]
     else:
         return os.environ["MIDDLEWARE_TEST_IP"]
 
--- a/src/middlewared/middlewared/test/integration/utils/client.py
+++ b/src/middlewared/middlewared/test/integration/utils/client.py
@@ -27,13 +27,7 @@ def client(*, auth=undefined, auth_required=True, py_exceptions=True, log_py_exc
                     assert logged_in
             yield c
     except socket.timeout as e:
-        if IS_HA:
-            # don't fail the entire run on HA systems since
-            # we perform failovers to and from each controller
-            # so timeouts could happen (which are expected)
-            raise e from None
-        else:
-            fail(f'socket timeout on URI: {uri!r} HOST_IP: {host_ip!r}')
+        fail(f'socket timeout on URI: {uri!r} HOST_IP: {host_ip!r}')
 
 
 def host():
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,10 +3,11 @@ import os
 
 import pytest
 
-pytest.register_assert_rewrite("middlewared.test")
-
-from middlewared.test.integration.utils import call
+from auto_config import ha, ip, vip
+from middlewared.test.integration.utils.client import client
 from middlewared.test.integration.utils.pytest import failed
+
+pytest.register_assert_rewrite("middlewared.test")
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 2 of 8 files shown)


src/middlewared/middlewared/test/integration/utils/client.py
```diff
@@ -10,27 +10,39 @@
...
                 if auth_required:
                     assert logged_in
             yield c
-    except socket.timeout:
-        fail('socket timeout')
+    except socket.timeout as e:
+        if IS_HA:
+            # don't fail the entire run on HA systems since
+            # we perform failovers to and from each controller
+            # so timeouts could happen (which are expected)
```
themylogin, line 33: Are they expected on certain tests or random tests?

src/middlewared/middlewared/test/integration/utils/client.py
```diff
@@ -10,27 +10,39 @@
...
+            raise e from None
+        else:
+            fail(f'socket timeout on URI: {uri!r} HOST_IP: {host_ip!r}')
 
 
 def host():
     if "NODE_A_IP" in os.environ:
+        # this is not to be confused with HA systems
+        # as it should only be set on NON-HA environments
```
themylogin, line 42: Why is it named `NODE_A` then?

src/middlewared/middlewared/test/integration/utils/client.py
```diff
@@ -10,27 +10,39 @@
...
         return os.environ["NODE_A_IP"]
+    elif IS_HA:
+        return os.environ["virtual_ip"]
     else:
         return os.environ["MIDDLEWARE_TEST_IP"]
```
themylogin, line 47: How is this case different from `NODE_A_IP`?

src/middlewared/middlewared/test/integration/utils/client.py
```diff
@@ -10,27 +10,39 @@
...
+            fail(f'socket timeout on URI: {uri!r} HOST_IP: {host_ip!r}')
 
 
 def host():
     if "NODE_A_IP" in os.environ:
+        # this is not to be confused with HA systems
+        # as it should only be set on NON-HA environments
         return os.environ["NODE_A_IP"]
+    elif IS_HA:
+        return os.environ["virtual_ip"]
```
themylogin, line 45: Lowercase environment variable name?

tests/conftest.py
```diff
@@ -17,20 +18,40 @@ def fail_fixture():
...
-            call("test.notify_test_end", request.node.name, client_kwargs=client_kwargs)
+    # That's why we also notify test ends. What happens between a test end
+    # and the next test start is caused by session/package/module/class
+    # fixtures setup code.
+    with client(host_ip=ip_to_use) as c:
+        c.call("test.notify_test_end", test_name)
+        if ha:
+            # the CI suite does all kinds of things to the standby controller
+            # on HA systems, so we need to suppress connection errors here
+            with contextlib.suppress(Exception):
```
themylogin, line 52: We should not do this if the test start caused an exception.

themylogin (changes requested):

src/middlewared/middlewared/test/integration/utils/client.py
yocalebo: Don't know and don't care. Happened long before us and I'm not changing it right now.

src/middlewared/middlewared/test/integration/utils/client.py
yocalebo: Don't know and, again, don't care :) happened long before us and I'm not interested in fixing it in this PR

tests/runtest.py
```diff
@@ -143,20 +169,81 @@
...
+os.environ['USE_VIP'] = 'NO'
+if ha:
+    if vip:
+        os.environ['virtual_ip'] = vip
+    elif os.environ.get('virtual_ip'):
+        vip = os.environ['virtual_ip']
+    else:
+        for i in ip_interface(f'{ip}/{netmask}').network:
+            last_octet = int(i.compressed.split('.')[-1])
+            if last_octet < 15 or last_octet >= 250:
```
mgrimesix, line 212: Might need to update the DNS?
My HA VM in the new KVM uses 10.220.38.[6.7.8]

src/middlewared/middlewared/test/integration/utils/client.py
yocalebo: Yes? Many other environment variables are set as lower-case in `runtest.py` I didn't change this part

src/middlewared/middlewared/test/integration/utils/client.py
yocalebo: That's hard to predict and is why I don't even like this functionality in the first place. I'd much rather just log to the active node on an HA machine then try and send duplicate messages across the heartbeat for no real gain. If anything, it's confusing because the standby node actually isn't doing _anything_ wrt to these tests.

tests/runtest.py
yocalebo: Your VM is using those as DNS servers?

src/middlewared/middlewared/test/integration/utils/client.py
themylogin: WHich tests? I don't understand. This fails the whole test suite if the connection to another node times out. There is no much sense to try to run each and every test against the dead node. Is HA VIP node expected to be down, but just for a while, and we assume it's fine?

yocalebo: Oh I see, my branch is missing a commit.... one moment

tests/runtest.py
yocalebo: Oh I see what you're saying, this method takes into consideration the subnet mask so it it'll iterate over every address. There won't be any collisions.

tests/runtest.py
```diff
@@ -143,20 +169,81 @@
...
+        ns1 = net_config.get('nameserver1')
+        ns2 = net_config.get('nameserver2')
+        _ip_to_use = socket.gethostbyname(ip_to_use)
+        for i in c.call('interface.query'):
+            for j in i['state']['aliases']:
+                if j.get('address') == _ip_to_use:
+                    iface = i['id']
+                    net = j['netmask']
+                    for k in c.call('route.system_routes'):
+                        if k.get('network') == '0.0.0.0' and k.get('interface') == i['id']:
```
bmeagherix, line 192: When I'm using `runtest.py` locally against VMs their default gateway is on the NATed interface, but I'm communicating with them (VM-to-VM) on another (host-only) network interface.

Therefore, I added an additional block here:
```
                    for k in c.call('route.system_routes'):
                        if k.get('network') == '0.0.0.0' and k.get('interface') == i['id']:
                            if k['gateway']:
                                return iface, net, k['gateway'], ns1, ns2
                    for k in c.call('route.system_routes'):
                        if k.get('network') == '0.0.0.0':
                            if k['gateway']:
                                return iface, net, k['gateway'], ns1, ns2
```

Acceptable to include, or will I just maintain it out-of-tree for my own purposes?

themylogin (approved):
Thank you!

yocalebo:
NOTE: this is also backwards compatible with our old jenkins infrastructure.

bmeagherix (approved):
LGTM, but one possible addition below...

mgrimesix (approved):

tests/runtest.py
yocalebo: Nope, it's safe to include. I'll push it here, thanks!

yocalebo: Fixed 👍

yocalebo:
backport

PR #13453, opened 2024-04-03
NAS-128164 / 24.10 / remove notify_test* calls on HA

yocalebo:
This is breaking all of the CI tests and I'm tired of dealing with it. It doesn't provide any benefit to log the test name to the remote controller. Especially since the remote controller doesn't even run the test. If we were to maintain this functionality, it means we're going to have to figure out which exact tests expect a failure here and which ones do not. That's too much maintenance overhead and doesn't provide any benefit.

NOTE: this still means we log this functionality on HA systems, it just means we log it on the controller that's actually running the test.


anodos325 (approved):

yocalebo:
backport

PR #13459, opened 2024-04-03
NAS-127788 / 24.10 / Fix netdata boot time order wrt UPS

Qubad786:
## Problem
Some users are experiencing issues with Netdata not displaying UPS (Uninterruptible Power Supply) statistics. Upon investigation, it was found that the problem lies in the systemd service boot order. Netdata starts before the nut services, leading to the issue where Netdata attempts to report UPS stats before the UPS service is started during boot time.

## Solution
Adjust the Netdata boot order by adding the nut service as a dependency in the Netdata service configuration.
(This was already fixed when we added `smartmontools.service` dependency but just adding `nut.target` for explicitness now)


yocalebo:
backport

PR #13476, opened 2024-04-04
NAS-128099 / 24.04.1 / Add rel_tgt_id to iSCSI target database

bmeagherix:
We **need** to provide `rel_tgt_id` in the _scst.conf_ in HA ALUA configurations, so that we can coordinate what is being written to _scst.conf_ on each controller.

Prior to PR #12967 we were not writing `rel_tgt_id` to _scst.conf_ for non-HA configs, but a flaw was noticed where toggling ALUA on could cause `rel_tgt_id` movement/clashes.  Therefore, decided to **always** write `rel_tgt_id`.

Unfortunately, since we _**still**_ were not using a proper value, but instead the target index from the list returned by `iscsi.target.query`, this meant that changes to the list returned (depending on the order `scstadmin` processed things) could result in scst errors, e.g.
```
scst: ***ERROR***: Relative target id 9 is not unique
```
Adding a unique `rel_tgt_id` to the database & then using it when generating _scst.conf_ appears to be the correct way to handle things.


bmeagherix:
Closing.  Will generate against 24.04.0 instead.

PR #13480, opened 2024-04-05
Add rel_tgt_id to iSCSI target database

bmeagherix:
We **need** to provide `rel_tgt_id` in the _scst.conf_ in HA ALUA configurations, so that we can coordinate what is being written to _scst.conf_ on each controller.

Prior to PR #12967 we were not writing `rel_tgt_id` to _scst.conf_ for non-HA configs, but a flaw was noticed where toggling ALUA on could cause `rel_tgt_id` movement/clashes.  Therefore, decided to **always** write `rel_tgt_id`.

Unfortunately, since we _**still**_ were not using a proper value, but instead the target index from the list returned by `iscsi.target.query`, this meant that changes to the list returned (depending on the order `scstadmin` processed things) could result in scst errors, e.g.
```
scst: ***ERROR***: Relative target id 9 is not unique
```
Adding a unique `rel_tgt_id` to the database & then using it when generating _scst.conf_ appears to be the correct way to handle things.


src/middlewared/middlewared/plugins/iscsi_/targets.py
```diff
@@ -25,6 +25,7 @@ class iSCSITargetModel(sa.Model):
     iscsi_target_alias = sa.Column(sa.String(120), nullable=True, unique=True)
     iscsi_target_mode = sa.Column(sa.String(20), default='iscsi')
     iscsi_target_auth_networks = sa.Column(sa.JSON(list))
+    iscsi_target_rel_tgt_id = sa.Column(sa.Integer(), nullable=False, unique=True)
```
themylogin, line 28:
```suggestion
    iscsi_target_rel_tgt_id = sa.Column(sa.Integer(), unique=True)
```
`nullable=False` is the default

themylogin (approved):

PR #13486, opened 2024-04-05
Add USE_VIP to allow tests when HA/VIP already configured

bmeagherix:
If using `runtest.py` against a **previously configured** HA system, and we are skipping `test_001_configure_interface`, then we may want to force the `truenas_server.ip` to the VIP.


yocalebo (approved):

william-gr:
backport

PR #13495, opened 2024-04-08
NAS-128261 / 24.10 / replace test-400 with proper test

yocalebo:
test_400 makes REST requests in the module's global scope. This is bad. It also uses REST instead of WS. Replace it with a proper equivalent test module.


tests/api2/test_enable_disable_services.py
```diff
@@ -0,0 +1,10 @@
+from middlewared.test.integration.utils import call
+
+def enable_services():
+    for svc in filter(lambda x: not x['enable'], call('service.query'):
+        call('service.update', svc['id'], {'enable': True})
+
+
+def disable_services():
+    for svc in filter(lambda x: x['enable'], call('service.query'):
+        call('service.update', svc['id'], {'enable': True})
```
bmeagherix, line 10: Presumably `disable_services` should be:
```
call('service.update', svc['id'], {'enable': False})
```

bmeagherix (changes requested):

tests/api2/test_enable_disable_services.py
yocalebo: fixed, thanks!

bmeagherix (approved):

PR #13506, opened 2024-04-09
NAS-128277 / 24.04.1 / Restart winbindd while initializing directory services

anodos325:
During QE testing it was observed that in-memory idmap backend settings of the idmap worker process was stale after reboot.

SIGHUP was insufficient to prompt reload and so this adds a winbind restart to ensure we always have correct configuration loaded.


anodos325:
This area is being refactored in EE and so does not apply there.

bmeagherix (approved):

PR #13513, opened 2024-04-09
NAS-128253 / 24.04.1 / Move to tier 1 storage for netdata

Qubad786:
## Context

It was requested that we have 2 storage tiers for netdata now and a configurable endpoint which allows users to toggle on how may days of data they would like to store.
Changes have been added to accommodate this where now number of days is configurable and the interval chosen for tier1 is 5 minutes for now but that too is configurable and can be changed as requested.


src/middlewared/middlewared/alembic/versions/24.04/2024-04-08_21-43_netdata_update_storage_tier.py
```diff
@@ -0,0 +1,31 @@
+"""netdata update storage tier 2
```
yocalebo, line 1:
```suggestion
"""netdata update storage tier 1
```

yocalebo (approved):
Fantastic job!

PR #13534, opened 2024-04-11
NAS-127770 / 24.10 / Add auditing of NFS and FTP

mgrimesix:
Add auditing of NFS and FTP.

NFS auditing includes configuration and share create, update and delete.
FTP auditing includes configuration only.

### DIFF
--- a/tests/api2/test_audit_ftp.py
+++ b/tests/api2/test_audit_ftp.py
@@ -48,7 +48,7 @@ def test_ftp_config_audit(api):
         # Restore initial state
         restore_payload = {
             'clients': initial_ftp_config['clients'],
-            'onlyanonymous': initial_ftp_config['onlyanonymous'],
+            'rootlogin': initial_ftp_config['rootlogin'],
             'banner': initial_ftp_config['banner']
         }
         if api == 'ws':
--- a/tests/api2/test_audit_ftp.py
+++ b/tests/api2/test_audit_ftp.py
@@ -1,0 +1,60 @@
+import os
+import sys
+
+import pytest
+from middlewared.test.integration.assets.pool import dataset
+from middlewared.test.integration.utils import call
+from middlewared.test.integration.utils.audit import expect_audit_method_calls
+
+sys.path.append(os.getcwd())
+from functions import PUT
+
+
+@pytest.fixture(scope='module')
+def nfs_audit_dataset(request):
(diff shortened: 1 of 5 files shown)


tests/api2/test_audit_ftp.py
```diff
@@ -0,0 +1,60 @@
...
+            elif api == 'rest':
+                result = PUT('/ftp/', payload)
+                assert result.status_code == 200, result.text
+            else:
+                raise ValueError(api)
+    finally:
+        # Restore initial state
+        restore_payload = {
+            'clients': initial_ftp_config['clients'],
+            'onlyanonymous': initial_ftp_config['onlyanonymous'],
```
bmeagherix, line 51: It appears that above you did **_not_** change `onlyanonymous`, but **_did_** change `rootlogin`.  Therefore this "restore initial state" isn't clear.  If there's some behind-the-scenes stuff that means this _**is**_ correct, please add a comment explaining why (or at least noting it).

tests/api2/test_audit_nfs.py
```diff
@@ -0,0 +1,153 @@
...
+                assert result.status_code == 200, result.text
+            else:
+                raise ValueError(api)
+    finally:
+        # Restore initial state
+        restore_payload = {
+            'mountd_log': initial_nfs_config['mountd_log'],
+            'mountd_port': initial_nfs_config['mountd_port'],
+            'protocols': initial_nfs_config['protocols']
+        }
```
bmeagherix, line 82: Does this undo the `nfs.add_principal` ?  If so, maybe add a comment to that effect to clarify.

bmeagherix (changes requested):

tests/api2/test_audit_nfs.py
mgrimesix: I added a comment that the `add_principal` would call would fail.  So , no setting to undo there.

tests/api2/test_audit_ftp.py
mgrimesix: Yep, should be rootlogin that gets restored.

tests/api2/test_audit_nfs.py
bmeagherix: OK

bmeagherix (approved):

PR #13553, opened 2024-04-15
NAS-127702 / 24.10 / Recursively mount root dataset during boot if it is not encrypted

usaleem-ix:
Import the pools during boot without mounting any of the datasets. Check on 'feature@encryption' for imported pool. If feature is not active, recursively mount the root datasets, which effectively mounts all datasets present on the pool.

If the encryption feature is active, continue and unlock the datasets.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/import_pool.py
+++ b/src/middlewared/middlewared/plugins/pool_/import_pool.py
@@ -213,12 +213,66 @@ class PoolService(Service):
         return True
 
     @private
+    def recursive_mount(self, name):
+        cmd = [
+            'zfs', 'mount',
+            '-R',  # recursive flag
+            name,  # name of the zpool / root dataset
+        ]
+        try:
+            self.logger.debug('Going to mount root dataset recusively: %r', name)
+            cp = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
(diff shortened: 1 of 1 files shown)


yocalebo:
@sonicaj upstream merged changes where we can do `zfs mount -R <root-dataset>` to recursively mount from the parent dataset (including zpool). I've asked Umer to make the changes here at boot. Can you review for me please? Also, I do believe we have some quite complex handling in this area of our code related to encrypted zpools. We can remove that with these changes but that should probably be done in a separate PR.

src/middlewared/middlewared/plugins/pool_/import_pool.py
```diff
@@ -212,13 +212,67 @@ async def import_pool(self, job, data):
 
         return True
 
+    @private
+    def recursive_mount(self, name):
+        cmd = [
+            'zfs', 'mount',
+            '-R',  # recusrive flag
```
sonicaj, line 219: Comment typo

sonicaj (approved):
LGTM apart from the typo!

src/middlewared/middlewared/plugins/pool_/import_pool.py
usaleem-ix: Thanks, typo is fixed.

sonicaj (approved):
Thanks!

PR #13554, opened 2024-04-15
NAS-128337 / 24.10 / Speed up ACL execute permissions check

anodos325:
The validation that ACL entries have ability to chdir to specified path had relatively terrible performance because it was forking a worker process for each change_to_user operation. This commit refactors so that we only fork once for validating the entire ACL.

This required refactoring our user context code to leave the saved UID and saved GID in-tact so that we can repeatedly setresuid and setresgid from same child process.


yocalebo (approved):
Cool!

PR #13556, opened 2024-04-15
NAS-128339 / 24.10 / account for disk['State'] == 'Absent' on ES24N

yocalebo:
When a disk is removed from ES24N, the REST API doesn't report a `SerialNumber` key. The fix is to check for the `State` key as being `Absent`. This fixes this particular issue but also adds a bit more guard-rails when mapping the disks for this platform. If we crash inside this function, the entirety of the enclosure section in the UI ceases to function which isn't desirable.

### DIFF
--- a/src/middlewared/middlewared/plugins/enclosure_/jbof_enclosures.py
+++ b/src/middlewared/middlewared/plugins/enclosure_/jbof_enclosures.py
@@ -101,8 +101,22 @@ def map_es24n(model, rclient, uri):
     }
     mapped = dict()
     for disk in all_disks['Members']:
-        slot = int(disk['Id'])
-        if found := mounted_disks.get(disk['SerialNumber']):
+        if disk['State'] == 'Absent':
+            mapped[slot] = None
+            continue
+
+        sn = disk.get('SerialNumber')
+        if not sn:
+            mapped[slot] = None
+            continue
+
+        slot = disk.get('Id', '')
+        if not slot or not slot.isdigit():
+            # shouldn't happen but need to catch edge-case
+            mapped[slot] = None
+
+        slot = int(slot)
+        if found := mounted_disks.get(sn):
             try:
                 # we expect namespace 1 for the device (i.e. nvme1n1)
                 idx = found[1]['namespaces'].index(f'{found[0]}n1')
--- a/src/middlewared/middlewared/plugins/enclosure_/jbof_enclosures.py
+++ b/src/middlewared/middlewared/plugins/enclosure_/jbof_enclosures.py
@@ -114,6 +114,7 @@ def map_es24n(model, rclient, uri):
         if not slot or not slot.isdigit():
             # shouldn't happen but need to catch edge-case
             mapped[slot] = None
+            continue
 
         slot = int(slot)
         if found := mounted_disks.get(sn):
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/enclosure_/jbof_enclosures.py
```diff
@@ -101,8 +101,22 @@ def map_es24n(model, rclient, uri):
...
+
+        sn = disk.get('SerialNumber')
+        if not sn:
+            mapped[slot] = None
+            continue
+
+        slot = disk.get('Id', '')
+        if not slot or not slot.isdigit():
+            # shouldn't happen but need to catch edge-case
+            mapped[slot] = None
```
themylogin, line 116: Do we need a `continue` here? Subsequent `int(slot)` will fail

yocalebo: oops, I didn't push that change 👍

src/middlewared/middlewared/plugins/enclosure_/jbof_enclosures.py
```diff
@@ -101,8 +101,23 @@ def map_es24n(model, rclient, uri):
...
+        if disk['State'] == 'Absent':
+            mapped[slot] = None
+            continue
+
+        sn = disk.get('SerialNumber')
+        if not sn:
+            mapped[slot] = None
+            continue
+
+        slot = disk.get('Id', '')
```
themylogin, line 113: Should this be the first thing to be done in the loop body? We use `slot` above and it might be undefined

yocalebo: Juggling too many things. Good catch, I've fixed.

themylogin:
@yocalebo I see no changes :(

yocalebo:
> @yocalebo I see no changes :(

pushed it again just now

src/middlewared/middlewared/plugins/enclosure_/jbof_enclosures.py
```diff
@@ -101,8 +101,23 @@ def map_es24n(model, rclient, uri):
     }
     mapped = dict()
     for disk in all_disks['Members']:
-        slot = int(disk['Id'])
-        if found := mounted_disks.get(disk['SerialNumber']):
+        if disk['State'] == 'Absent':
+            mapped[slot] = None
```
themylogin, line 105: `slot` is still unassigned here

yocalebo: I think it's time for me to stop computer'ing today. 😄

src/middlewared/middlewared/plugins/enclosure_/jbof_enclosures.py
```diff
@@ -101,8 +101,22 @@ def map_es24n(model, rclient, uri):
...
     for disk in all_disks['Members']:
-        slot = int(disk['Id'])
-        if found := mounted_disks.get(disk['SerialNumber']):
+        slot = dis.get('Id', '')
+        if not slot or not slot.isdigit():
+            # shouldn't happen but need to catch edge-case
+            continue
+
+        if disk['State'] == 'Absent':
+            mapped[slot] = None
```
themylogin, line 112: Sorry, but do we care if `slot` is `16` or `"16"` at this point?

themylogin (approved):
Looks good to me now, thank you!

PR #13567, opened 2024-04-16
NAS-128359 / 24.10 / Apply settings from `/data/post_install.json`

themylogin:
This file will be written by the installer API server https://ixsystems.atlassian.net/browse/NAS-127428 to set up the network configuration (as discussed in NEP-045)

### DIFF
--- a/src/middlewared/middlewared/plugins/post_install.py
+++ b/src/middlewared/middlewared/plugins/post_install.py
@@ -1,0 +1,79 @@
+import json
+import os
+
+import jsonschema
+
+from middlewared.service import Service
+
+PATH = "/data/post_install.json"
+
+SCHEMA = {
+    "type": "object",
+    "additionalProperties": False,
+    "properties": {
+        "network_interfaces": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "required": ["name"],
+                "additionalProperties": False,
+                "properties": {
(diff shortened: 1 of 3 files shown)


themylogin:
time 1:00

src/middlewared/middlewared/plugins/post_install.py
```diff
@@ -0,0 +1,79 @@
...
+
+from middlewared.service import Service
+
+PATH = "/data/post_install.json"
+
+SCHEMA = {
+    "type": "object",
+    "additionalProperties": False,
+    "properties": {
+        "network_interface": {
```
sonicaj, line 14: We can have this as a required property to make sure we don't have an empty dict ?

sonicaj: Also maybe this should be `network_interfaces` ?

themylogin: What's the problem with having an empty dict, why does network interface configuration have to be required?

sonicaj: There shouldn't be any, idea was consistency and being strict about values where this can be an empty list but at least the key is there so we don't have to always be safe about it when retrieving values.

sonicaj (approved):

PR #13577, opened 2024-04-17
NAS-128161 / 24.10 / Get all interface names for reporting

Qubad786:
## Problem
In a fresh install, physical interfaces are not stored in the database. When the `reporting.realtime` event tries to retrieve the interface names from the database, it returns an empty list, resulting in network stats not being displayed.

## Solution

Retrieve all interface names from `interface.query` for reporting events, which obtains interface information directly from the system if interfaces are not stored in the database.

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -189,6 +189,7 @@ class InterfaceService(CRUDService):
         """
         Query Interfaces with `query-filters` and `query-options`
         """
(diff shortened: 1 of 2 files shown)


yocalebo (commented):
Why do we want to report network stats for interfaces that haven't been configured?

Qubad786:
@yocalebo when user first sets up the machine, we don't report stats for interfaces even though they are working atm and we are deviating from old behavior where in bluefin it used to work but with the transition to netdata - we broke this

yocalebo (approved):

PR #13584, opened 2024-04-17
NAS-128373 / 13.3 / Update ZFS plugin for py-libzfs changes

anodos325:
* We deprecated and removed an unused private API parameter in SCALE. Said parameter is also unused in 13.3 by both middleware and webui, and since it is a private API there is no guarantee of stability between versions.

* Recent versions of python-libzfs removed __getstate__() methods for python 3.10 compatibilty and replaced with asdict().


bmeagherix (approved):
Seems reasonable

yocalebo (approved):

PR #13586, opened 2024-04-17
NAS-128374 / 24.10 / Add options to failover.send_file

bmeagherix:
Add options to `failover.send_file` and restrict the `mode` when calling.

### DIFF
--- a/src/middlewared/middlewared/plugins/failover_/remote.py
+++ b/src/middlewared/middlewared/plugins/failover_/remote.py
@@ -161,12 +161,13 @@ class RemoteClient:
             except Exception:
                 logger.warning('Failed to run callback for %s', name, exc_info=True)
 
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/failover_/remote.py
```diff
@@ -161,7 +161,7 @@ def _sub_callback(self, name, type_, **message):
             except Exception:
                 logger.warning('Failed to run callback for %s', name, exc_info=True)
 
-    def send_file(self, token, local_path, remote_path):
+    def send_file(self, token, local_path, remote_path, options={}):
```
yocalebo, line 164: It's not common to specify `options={}` in the function's kwarg section. Let's set it to `options=None` and then inside the function do something like
```
options = options or {}

yocalebo (approved):

PR #13587, opened 2024-04-18
NAS-128382 / 24.10 / fix failover.send_small_file

yocalebo:
Investigating an unrelated issue, I noticed a few issues with this method.

- prone to race conditions
- unoptimized because it was calling `os.stat` on the same path 3 different times
- creating a new dictionary object inside the `while True` loop
- no validation on the `path` (i.e. we expect and limit this to a regular file)

This fixes all of the above issues. There should be no functional change in behavior.


bmeagherix (approved):
I'll approve, but maybe add a test?

PR #13598, opened 2024-04-18
NAS-128399 / 24.10 / More nfs failover fixes

mgrimesix:
### Problem
NFS was not configured to properly support HA function.  NFS holds client state information in a database that, by default, is located at /var/lib/nfs.  To properly support failover we need that to be a shared directory.  For TrueNAS that is the system dataset.

### Fix
Create and NFS entry in the system dataset.  Generate the necessary directory structure and to support upgrade we also initialize it with data from /var/lib/nfs.
In systemd we bind the NFS client tracking daemon to nfs-server to allow the transfer of the client tracking database.
Update the nfs configuration file with the path to the nfs data on the system dataset: `/var/db/system/nfs`

### Other related items in the PR

1. Updated the CI tests
2. Minor flake8 cleanup and refactoring

### DIFF
--- a/src/middlewared/middlewared/plugins/nfs.py
+++ b/src/middlewared/middlewared/plugins/nfs.py
@@ -1,12 +1,10 @@
 import enum
 import errno
-import grp
 import ipaddress
 import itertools
 import os
 import pathlib
-import pwd
-import subprocess
+import shutil
 
 from middlewared.common.attachment import LockableFSAttachmentDelegate
 from middlewared.common.listen import SystemServiceListenMultipleDelegate
@@ -26,14 +24,9 @@ class NFSPath(enum.Enum):
     STATEDIR = (os.path.join(SYSDATASET_PATH, 'nfs'), 0o755, True, {'uid': 0, 'gid': 0})
     CLDDIR = (os.path.join(SYSDATASET_PATH, 'nfs', 'nfsdcld'), 0o700, True, {'uid': 0, 'gid': 0})
     CLDTRKDIR = (os.path.join(SYSDATASET_PATH, 'nfs', 'nfsdcltrack'), 0o700, True, {'uid': 0, 'gid': 0})
-    SMDIR = (os.path.join(SYSDATASET_PATH, 'nfs', 'sm'), 0o755, True, {
-        'uid': pwd.getpwnam("statd").pw_uid,
-        'gid': grp.getgrnam("nogroup").gr_gid
-    })
-    SMBAKDIR = (os.path.join(SYSDATASET_PATH, 'nfs', 'sm.bak'), 0o755, True, {
-        'uid': pwd.getpwnam("statd").pw_uid,
-        'gid': grp.getgrnam("nogroup").gr_gid
-    })
+    # Fix up the uid and gid in setup_directories
+    SMDIR = (os.path.join(SYSDATASET_PATH, 'nfs', 'sm'), 0o755, True, {'uid': 'statd', 'gid': 'nogroup'})
+    SMBAKDIR = (os.path.join(SYSDATASET_PATH, 'nfs', 'sm.bak'), 0o755, True, {'uid': 'statd', 'gid': 'nogroup'})
     V4RECOVERYDIR = (os.path.join(SYSDATASET_PATH, 'nfs', 'v4recovery'), 0o755, True, {'uid': 0, 'gid': 0})
 
     def platform(self):
@@ -110,7 +103,7 @@ class NFSService(SystemServiceService):
     )
 
     @private
-    async def setup_directories(self):
+    def setup_directories(self):
         '''
         We are moving the NFS state directory from /var/lib/nfs to
         the system dataset: /var/db/system/nfs.
@@ -121,33 +114,47 @@ class NFSService(SystemServiceService):
 
         We always make sure the expected directories are present
         '''
-        def create_dirs(spec, path):
-            try:
-                os.chmod(path, spec.mode())
-                stat = os.stat(path)
-                if (stat.st_uid != spec.owner().get('uid')) or (stat.st_gid != spec.owner().get('gid')):
-                    self.logger.warning("%s: invalid owner for path. Correcting.", path)
-                    os.chown(path, spec.owner().get('uid'), spec.owner().get('gid'))
-            except FileNotFoundError:
-                if spec.is_dir():
-                    os.mkdir(path, spec.mode())
-                    os.chown(path, spec.owner().get('uid'), spec.owner().get('gid'))
+        def name_to_id_conversion(usrgrp):
+            ''' Accept [usr,grp], return [usr,grp] as IDs '''
+            if any(isinstance(owner, str) for owner in usrgrp):
+                # Convert from name to ID, then apply
+                try:
+                    usrgrp[0] = self.middleware.call_sync('user.get_builtin_user_id', usrgrp[0])
+                    usrgrp[1] = self.middleware.call_sync('group.get_builtin_group_id', usrgrp[1])
+                except Exception as e:
+                    self.logger.error(
+                        "%s: Failed to apply ownership. usr=%r, grp=%r %r", usrgrp[0], usrgrp[1], e
+                    )
+            return usrgrp
 
-        if not os.listdir(NFSPath.STATEDIR.platform()):
-            # System db is empty, populate it with contents of /var/lib/nfs
-            # Going forward, the system dataset should hold the NFS state data
-            if not os.path.isdir('/var/lib/nfs'):
-                res = subprocess.run(
-                    ['cp', '-a', '/var/lib/nfs/*', NFSPath.STATEDIR.platform()], check=False, capture_output=True,
-                )
-                if res.returncode != 0:
-                    self.logger.error('Failed to initialize NFS state from /var/lib/nfs: %r', res.stderr.decode())
-                # Continue anyway.
+        # Initialize the system dataset NFS state directory
+        try:
+            if not any(os.scandir(NFSPath.STATEDIR.platform())):
+                # System db is empty, populate it with contents of /var/lib/nfs
+                # Going forward, the system dataset should hold the NFS state data
+                if not os.path.isdir('/var/lib/nfs'):
+                    try:
+                        shutil.copytree('/var/lib/nfs', NFSPath.STATEDIR.platform())
+                    except Exception as e:
+                        self.logger.error('Failed to initialize NFS state from /var/lib/nfs: %r', e)
+                    # Continue anyway.
+        except Exception as e:
+            self.logger.error("Could not find required path: %r", e)
 
         # Make sure we have the necessary directories
         for p in NFSPath:
             path = p.platform()
-            await self.middleware.run_in_thread(create_dirs, p, path)
+            usrgrp = name_to_id_conversion([p.owner().get('uid'), p.owner().get('gid')])
+            try:
+                os.chmod(path, p.mode())
+                stat = os.stat(path)
+                if (stat.st_uid != usrgrp[0]) or (stat.st_gid != usrgrp[1]):
+                    self.logger.info("%s: Fix up owner for path", path)
+                    os.chown(path, usrgrp[0], usrgrp[1])
+            except FileNotFoundError:
+                if p.is_dir():
+                    os.mkdir(path, p.mode())
+                    os.chown(path, usrgrp[0], usrgrp[1])
 
         if os.path.exists('/proc/fs/nfsd/nfsv4recoverydir'):
             try:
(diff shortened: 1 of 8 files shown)


src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -77,6 +109,53 @@ class Config:
         Bool('managed_nfsd', default=True),
     )
 
+    @private
+    async def setup_directories(self):
```
anodos325, line 113: I don't think this should be async.

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -15,6 +18,35 @@
...
+from middlewared.plugins.system_dataset.utils import SYSDATASET_PATH
+
+
+class NFSPath(enum.Enum):
+    # nfs conf sections that use STATEDIR: exportd, mountd, statd
+    STATEDIR = (os.path.join(SYSDATASET_PATH, 'nfs'), 0o755, True, {'uid': 0, 'gid': 0})
+    CLDDIR = (os.path.join(SYSDATASET_PATH, 'nfs', 'nfsdcld'), 0o700, True, {'uid': 0, 'gid': 0})
+    CLDTRKDIR = (os.path.join(SYSDATASET_PATH, 'nfs', 'nfsdcltrack'), 0o700, True, {'uid': 0, 'gid': 0})
+    SMDIR = (os.path.join(SYSDATASET_PATH, 'nfs', 'sm'), 0o755, True, {
+        'uid': pwd.getpwnam("statd").pw_uid,
```
anodos325, line 30: Rather than using pwd and grp here, you should inside method call `user.get_builtin_user_id`. See `set_etc_file_perms()` in etc.py for example.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -77,6 +109,53 @@ class Config:
...
+                if res.returncode != 0:
+                    self.logger.error('Failed to initialize NFS state from /var/lib/nfs: %r', res.stderr.decode())
+                # Continue anyway.
+
+        # Make sure we have the necessary directories
+        for p in NFSPath:
+            path = p.platform()
+            await self.middleware.run_in_thread(create_dirs, p, path)
+
+        if os.path.exists('/proc/fs/nfsd/nfsv4recoverydir'):
```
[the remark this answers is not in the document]
mgrimesix: This setting is mostly done for completeness.  
It's a proc file in support of _legacy_ handling of NFSv4 client state.  It does not exist if `nfs-server` is not running.

At 'boot' this is hard-coded to `/var/lib/nfs/v4recovery`
`static char user_recovery_dirname[PATH_MAX] = "/var/lib/nfs/v4recovery";`

The routine `setup_directories` is called once at nfs-server start, which does introduce a window for TOCTOU.

The 'x' mode is for creation.  I'm not going to be creating this file.  If it doesn't exist, then no harm.
It's the same for 'w'.  I think 'r+' _is_ the proper choice here.
https://docs.python.org/3/library/functions.html#open

src/middlewared/middlewared/plugins/nfs.py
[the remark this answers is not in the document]
mgrimesix: Made suggested change

src/middlewared/middlewared/plugins/nfs.py
mgrimesix: Implemented usage of get_builtin_user_id as a post-create fixup.

src/middlewared/middlewared/plugins/nfs.py
mgrimesix: Made suggested change.

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -77,6 +109,53 @@ class Config:
...
+                stat = os.stat(path)
+                if (stat.st_uid != spec.owner().get('uid')) or (stat.st_gid != spec.owner().get('gid')):
+                    self.logger.warning("%s: invalid owner for path. Correcting.", path)
+                    os.chown(path, spec.owner().get('uid'), spec.owner().get('gid'))
+            except FileNotFoundError:
+                if spec.is_dir():
+                    os.mkdir(path, spec.mode())
+                    os.chown(path, spec.owner().get('uid'), spec.owner().get('gid'))
+
+        if not os.listdir(NFSPath.STATEDIR.platform()):
```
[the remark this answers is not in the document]
mgrimesix: Converted to a scandir solution.

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -15,6 +16,30 @@
...
+    # nfs conf sections that use STATEDIR: exportd, mountd, statd
+    STATEDIR = (os.path.join(SYSDATASET_PATH, 'nfs'), 0o755, True, {'uid': 0, 'gid': 0})
+    CLDDIR = (os.path.join(SYSDATASET_PATH, 'nfs', 'nfsdcld'), 0o700, True, {'uid': 0, 'gid': 0})
+    CLDTRKDIR = (os.path.join(SYSDATASET_PATH, 'nfs', 'nfsdcltrack'), 0o700, True, {'uid': 0, 'gid': 0})
+    # Fix up the uid and gid in setup_directories
+    SMDIR = (os.path.join(SYSDATASET_PATH, 'nfs', 'sm'), 0o755, True, {'uid': 'statd', 'gid': 'nogroup'})
+    SMBAKDIR = (os.path.join(SYSDATASET_PATH, 'nfs', 'sm.bak'), 0o755, True, {'uid': 'statd', 'gid': 'nogroup'})
+    V4RECOVERYDIR = (os.path.join(SYSDATASET_PATH, 'nfs', 'v4recovery'), 0o755, True, {'uid': 0, 'gid': 0})
+
+    def platform(self):
```
anodos325, line 32: This should just be a path, right? `platform` was something we were doing when there was mixed FreeBSD / Linux consumers of the source.

mgrimesix: path is a more sensible name.

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -77,6 +102,67 @@ class Config:
...
+            except FileNotFoundError:
+                if p.is_dir():
+                    os.mkdir(path, p.mode())
+                    os.chown(path, usrgrp[0], usrgrp[1])
+
+        procfs_path = '/proc/fs/nfsd/nfsv4recoverydir'
+        try:
+            with open(procfs_path, 'r+') as fp:
+                fp.write(f'{NFSPath.V4RECOVERYDIR.path()}\n')
+        except FileNotFoundError:
```
yocalebo, line 163: Sorry, I most be missing something obvious so bear with me 😄  I'm still not following why you're only catching `FileNotFoundError`? Previously, you checked for files existence and then wrote to it catching any exception that may come up during the write I/O.

Now you're only catching if the file doesn't exist and then logging an error but you will crash here if writing to it fails.

What I'm getting at is, does it matter if this file exists or not?

mgrimesix: I don't think it matters if the file exists.   
This file was used in support of an early solution for maintaining client state.  It was supplanted with the database managed by `nfsdcld`.  However, the code supporting the old mechanism was not removed.  Since it's not yet been removed, I figure it's best to 'manage' it.
Side note:  More recently, `nfsdcld` itself has been replaced with `nfsdcltrk`, but Debian still uses `nfsdcld`.   Currently, all three are in Debian, but to my knowledge only `nfsdcld` is really 'active'.
I'll change to trap on any error.  It's more safe.

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -77,6 +102,72 @@ class Config:
...
+        We are moving the NFS state directory from /var/lib/nfs to
+        the system dataset: /var/db/system/nfs.
+        When setup_directories is called /var/db/system/nfs is expected to exist.
+
+        If STATEDIR is empty, then this might be an initialization
+        and there might be current info in /var/lib/nfs.
+
+        We always make sure the expected directories are present
+        '''
+        def name_to_id_conversion(usrgrp):
```
yocalebo, line 117: let's move the `name_to_id_conversion` inner method to its own `@private` method in this class. Using inner methods actually incurs a non-trivial amount of overhead. You can keep the same name if you want. I also see that if `usergrp[0]` fails to be resolved, then we won't try and resolve the group (usergrp[1]). I'm not sure if that's by design but I believe we should make the method a bit more robust and explicit. I also see that the log message is talking about "applying ownership" but it's doing none of the sorts. Looks like it's just resolving the name to a uid/gid 😄 I picture the method to look something like:
```
@private
def name_to_id_converstion(name, name_type='user'):
    if any((not isinstance(name, str), isinstance(name, int))):
        # it's not a string (NoneType, float, w/e) or it's an int
        # so there is nothing to do
        return name

    if name_type = 'user':
        method = 'user.get_builtin_user_id'
    elif name_type == 'group':
        method = 'group.get_builtin_group_id'
    else:
        self.logger.error('Unexpected name_type (%r)', name_type)
        return name
    try:
        return self.middleware.call_sync(method, name)
    except Exception as e:
        if hasattr(e, 'errno') and e.errno == errno.ENOENT:
            self.logger.error('Failed to resolve builtin %s %r', name_type, name)
        else:
             self.logger.error('Unexpected error resolving builtin %s %r, name_type, name, exc_info=True)
        return name

File: src/middlewared/middlewared/plugins/nfs.py [hunk 8]
```diff
@@ -77,6 +102,72 @@ class Config:
...
+                except FileExistsError:
+                    pass
+                except Exception as e:
+                    self.logger.error('Failed to initialize NFS state from /var/lib/nfs: %r', e)
+                # Continue anyway.
+        except Exception as e:
+            self.logger.error("Could not find required path: %r", e)
+
+        # Make sure we have the necessary directories
+        for p in NFSPath:
```
yocalebo, line 147: This for loop is prone to a few different race conditions. It's also using the "file handle" instead of opening the file/directory and then using the fd based syscalls. I think we need to simplify this to something simple.
     
Something like:
```
for i in NFSServicePathInfo:
    uid = self.name_to_id_conversion(i.owner()['uid'], name_type='user')
    gid = self.name_to_id_conversion(i.owner()['gid'], name_type='group')
    path = i.path()
    if i.is_dir():
        open_flags = os.O_RDWR | os.O_DIRECTORY
        os.makedirs(path, exist_ok=True)
    else:
        open_flags = os.O_RDWR

    try:
        with open(os.open(path, flags=open_flags)) as f:
            os.fchmod(f.fileno(), i.mode())
            os.fchown(f.fileno(), uid, gid)
    except IsADirectoryError:
        if not i.is_dir():
            # the path is expected to be a file but it's a directory
            self.logger.error('Expected %r to be a file but instead found a directory', path)
        else:
            self.logger.error('Unexpected failure updating file %r', path, exc_info=True)
    except NotADirectoryError:
        if i.is_dir():
            # the path is expected to be a directory but it's a file
            self.logger.error('Expected %r to be a directory but instead found a file', path)
        else:
            self.logger.error('Unexpected failure updating directory %r', path, exc_info=True)
    except Exception:
        self.logger.error('Unexpected failure initializing %r', path, exc_info=True)

**Review by yocalebo — CHANGES_REQUESTED** [review]
Just a few remaining items 👍

File: src/middlewared/middlewared/plugins/nfs.py [hunk 9]
```diff
@@ -77,6 +102,72 @@ class Config:
...
+                    usrgrp[0] = self.middleware.call_sync('user.get_builtin_user_id', usrgrp[0])
+                    usrgrp[1] = self.middleware.call_sync('group.get_builtin_group_id', usrgrp[1])
+                except Exception as e:
+                    self.logger.error(
+                        "%s: Failed to apply ownership. usr=%r, grp=%r %r", usrgrp[0], usrgrp[1], e
+                    )
+            return usrgrp
+
+        # Initialize the system dataset NFS state directory
+        try:
```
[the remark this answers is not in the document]
mgrimesix: Wait... the point of the scandir check was to allow the copy operation _**only**_ if /var/db/system/nfs is empty.

yocalebo (approved):
Fantastic, great job! Glad to see this general issue get resolved, finally. Be sure and wait for @anodos325 before merge 👍

src/middlewared/middlewared/plugins/nfs.py
```diff
@@ -77,6 +102,93 @@ class Config:
         Bool('managed_nfsd', default=True),
     )
 
+    @private
+    def name_to_id_converstion(self, name, name_type='user'):
```
yocalebo, line 106: woops, a typo in the method name 😄 
```suggestion
    def name_to_id_converstion(self, name, name_type='user'):
```

src/middlewared/middlewared/plugins/nfs.py
mgrimesix: Implemented the simplified solution.

yocalebo (approved):
nice!

anodos325 (approved):

PR #13600, opened 2024-04-19
NAS-128404 / 24.10 / Improve file receive

yocalebo:
`file_receive` was susceptible to the TOCTOU race conditions when checking for a paths existence (or any path component) before opening said path. This should get rid of theoretical conditions and also switches to using the `fchmod`/`fchown` syscalls. I've also fixed a few flake8 issues while I'm here. There should be no functional change.


src/middlewared/middlewared/plugins/filesystem.py
```diff
@@ -476,21 +484,17 @@ def file_receive(self, path, content, options):
...
-        if mode := options.get('mode'):
-            os.chmod(path, mode)
-        if (uid := options.get('uid')) is not None and (gid := options.get('gid')) is not None:
-            os.chown(path, uid, gid)
+            if mode := options.get('mode'):
+                os.fchmod(f.fileno(), mode)
+
+            os.fchown(f.fileno(), options.get('uid', -1), options.get('gid', -1))
+
         if path == PWENC_FILE_SECRET:
```
anodos325, line 495: I wonder if in case of PWENC_FILE_SECRET we should always force mode to be 0o600 (avoid potential for someone writing bug where they make it world-readable).

yocalebo: Agree, I actually kind of assumed that the api call below this check did so. I'll make the change.

mgrimesix (dismissed):

mgrimesix:
I approved too quickly.  Got spoofed by the latest simple update.

anodos325 (approved):

PR #13603, opened 2024-04-22
NAS-123677 / 24.10 / Improve error handling for already in use PCI devices

Qubad786:
**Problem:**

An issue arises when the same PCI device is added to two virtual machines (VMs), and one of the VMs is already running. When the user attempts to start the second VM, QEMU raises an error indicating that the PCI device is already in use by another VM.

**Solution:**

To address this problem, validation for the PCI device has been updated. It now also considers whether the same PCI device is in use by another VM. If the PCI device is not in use by any other VM, it is marked as available for use or otherwise error out on middleware side that the device is already being used.

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/devices/pci.py
+++ b/src/middlewared/middlewared/plugins/vm/devices/pci.py
@@ -2,20 +2,41 @@ import subprocess
 
 from middlewared.service import CallError
 from middlewared.schema import Dict, Str
+from middlewared.utils import filter_list
 
 from .device import Device
 from .utils import ACTIVE_STATES, create_element, LIBVIRT_URI
 
 
-class PCI(Device):
+class PCIBase(Device):
+
+    def is_available(self):
+        return self.get_details()['available']
+
+    def in_use_by_vm(self, vms, vm_devices):
+        return any(vm['status']['state'] in ACTIVE_STATES for vm in self.get_vms_using_device(vms, vm_devices))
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/vm/devices/pci.py
```diff
@@ -37,6 +37,10 @@ def reattach_device(self):
 
     def pre_start_vm_device_setup(self, *args, **kwargs):
         device = self.get_details()
+
+        if not self.not_in_use_by_vm():
```
yocalebo, line 41: This is a double-negative which makes it pretty hard to follow what's actually being checked. I also think the name `not_in_use_by_vm()` is actually the inverse of what it's checking? It's actually checking if the device is IN USE by a vm? Either way, this is very confusing so we should rename/fix it.

yocalebo (changes requested):

Migsi:
Wouldn't it be better to just highlight in use devices instead of removing them from the list alltogether? I'm running a setup where I pass through the same GPU to a Windows and a Linux VM, which are running mutually exclusive. While obviously not done every day, having to stop either VM before being able to setup (not run) a second, thrid, fourth, etc. still seems quite a bit annoying to me.

Qubad786:
@Migsi yes that is accurate so your case is still good to go ✅

yocalebo (approved):

PR #13606, opened 2024-04-23
NAS-128442 / 24.10 / Avoid restarting SMB / NFS in critical group if DS enabled

anodos325:
SMB and NFS services will be restarted once we have initialized kerberos tickets and directory services if they are enabled.


anodos325:
I'm approaching this in differently in dragonfish.

PR #13607, opened 2024-04-23
Allow configuring ha_propagate default per-service

anodos325:
There are many services on TrueNAS for which we should not propogate service changes to standby controller. Historically the default has been to always propagate these changes to the standby controller. This commit allows changing the default on a per-service basis, which reduces risk of introducing issues by forgetting to specify not to propagate to other controller.


anodos325:
This was handled in a different way recently

PR #13610, opened 2024-04-23
NAS-128466 / 24.10 / Make robustize_clean_type_and_path more tolerant

bmeagherix:
A recent change to `iscsi_extent_locked` in PR #13572 wrt space handling meant `clean_type_and_path` could return an error.  Rectify.

Note:
- Currently the WebUI does **_not_** do this, so at the moment the space breakage only impacts CI tests.
- There is **another** issue with this test, that will probably be handled by a zfs change.  (TBD.)


yocalebo:
Can we not use the helper methods somewhere in here? (zvol_name_to_path and friends?)

bmeagherix:
> Can we not use the helper methods somewhere in here? (zvol_name_to_path and friends?)

Ok, done.

bmeagherix:
Instead or pursuing this, we decided to revert the original PR that introduced the issue.

PR #13613, opened 2024-04-24
NAS-128155 / 24.10 / Validate GRUB config syntax validity when passing extra kernel options

themylogin:


themylogin:
time 0:30

yocalebo (approved):
Awesome!

PR #13615, opened 2024-04-24
Fix crash when using SCALE client to talk to 13.0

anodos325:
13.0 middleware jobs fail with KeyError. Minimally, we shouldn't crash when trying to initiate jobs there.

### DIFF
--- a/src/middlewared/middlewared/client/client.py
+++ b/src/middlewared/middlewared/client/client.py
@@ -198,7 +198,7 @@ class Job:
                 trace={
                     'class': job['exc_info']['type'],
                     'formatted': job['exception'],
-                    'repr': job['exc_info']['repr'],
+                    'repr': job['exc_info'].get('repr'),
                 },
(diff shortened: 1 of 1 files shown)


mgrimesix (approved):

src/middlewared/middlewared/client/client.py
```diff
@@ -198,7 +198,7 @@ def result(self):
                 trace={
                     'class': job['exc_info']['type'],
                     'formatted': job['exception'],
-                    'repr': job['exc_info']['repr'],
+                    'repr': job['exc_info'].get('repr'),
```
themylogin, line 201: Let's use the last line of the traceback when this is not available then
```suggestion
                    'repr': job['exc_info'].get('repr', job['exception'].splitlines()[-1]),
```

anodos325: Great idea.

themylogin (approved):
Thank you!

PR #13618, opened 2024-04-24
NAS-128506 / 24.10 / Fix test iscsi auth network

bmeagherix:
Fix a bunch of tests that are failing in CI
- When running `iscsiadm` in tests precede with a `sudo` if we are not running as root.
- Improve _test_iscsi_auth_network.py_ in several ways
  - Make `my_ip4` a fixture so it only needs to run once
  - Extend `my_ip4` so that it can even handle NAT situations (`ssh` / `SSH_CLIENT`)
  - `flake8` fixes

Also required the ZFS fix in PR #[230 ](https://github.com/truenas/zfs/pull/230)

### DIFF
--- a/tests/api2/test_iscsi_auth_network.py
+++ b/tests/api2/test_iscsi_auth_network.py
@@ -1,32 +1,40 @@
+import contextlib
 import ipaddress
-import os
-import pytest
 import socket
-import sys
 
-apifolder = os.getcwd()
-sys.path.append(apifolder)
+import pytest
+from auto_config import ip
 
 from middlewared.test.integration.assets.iscsi import target_login_test
 from middlewared.test.integration.assets.pool import dataset
 from middlewared.test.integration.utils import call, ssh
 
-import contextlib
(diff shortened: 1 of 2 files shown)


tests/api2/test_iscsi_auth_network.py
```diff
@@ -1,32 +1,40 @@
...
+        # Fall back
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sock.settimeout(2)
+        result = sock.connect_ex((ip, 80))
+        assert result == 0
+        myip = sock.getsockname()[0]
+        sock.close()
+        # Check that we have an IPv4 address
+        socket.inet_pton(socket.AF_INET, myip)
+        print("BRIAN myip", myip)
```
yocalebo, line 34: Leftover?

bmeagherix: D'oh!

yocalebo (approved):

PR #13624, opened 2024-04-26
NAS-128574 / 24.10 / Allow HeaderDigest and DataDigest to be configured for iSCSI targets

bmeagherix:
Enhance iSCSI APIs to allow some target options to be specified

### DIFF
--- a/src/middlewared/middlewared/plugins/iscsi_/targets.py
+++ b/src/middlewared/middlewared/plugins/iscsi_/targets.py
@@ -364,8 +364,22 @@ class iSCSITargetService(CRUDService):
         await self._service_change('iscsitarget', 'reload', options={'ha_propagate': False})
 
         # Then process the BACKUP config if we are HA and ALUA is enabled.
-        if await self.middleware.call("iscsi.global.alua_enabled") and await self.middleware.call('failover.remote_connected'):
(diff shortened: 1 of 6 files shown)


src/middlewared/middlewared/plugins/iscsi_/targets.py
```diff
@@ -97,6 +127,18 @@ async def extend(self, data):
             ),
         ]),
         List('auth_networks', items=[IPAddr('ip', network=True)]),
+        Dict(
+            'options',
+            Str('DataDigest', enum=DigestEnum.choices()),
```
anodos325, line 132: I don't think we typically have camelcase parameters. Convention is all lower-case with underscores as-needed.

bmeagherix: I understand, but we're exposing **raw** iSCSI settings, so IMO this is clearer.

e.g. https://www.rfc-editor.org/rfc/rfc7143#section-13.1

anodos325: @yocalebo do you have strong opinion on this one?

bmeagherix:
Closing for now.  Need to perform performance analysis before proceeding.

PR #13625, opened 2024-04-28
NAS-128613 / 24.10 / Do not start netdata on boot in HA

Qubad786:
## Problem

We don't have netdata running on standby node but when passive boots there is a brief window when netdata starts because by default it is enabled to do so and then it leaves a spam in systemd logs.

## Solution

Make sure netdata does not run on passive node and it only runs on the active node.

### DIFF
--- a/debian/debian/ix-reporting.service
+++ b/debian/debian/ix-reporting.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=Setup TrueNAS Reporting
+DefaultDependencies=no
+
+# The after here reflects the after of netdata.service
+After=middlewared.service network-online.target httpd.service squid.service nfs-server.service mysqld.service named.service postfix.service smartmontools.service nut.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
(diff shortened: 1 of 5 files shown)


sonicaj (approved):

debian/debian/ix-reporting.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Setup TrueNAS Reporting
+DefaultDependencies=no
+
+After=middlewared.service network-online.target httpd.service squid.service nfs-server.service mysqld.service named.service postfix.service smartmontools.service nut.target
```
yocalebo, line 5: What is going on here? This is a ton of lines in `After`?

yocalebo (changes requested):

debian/debian/ix-reporting.service
Qubad786: @yocalebo we are replicating `netdata` service's `After` here

yocalebo: Okay, please put a comment in this file describing this fact.

yocalebo (approved):

PR #13627, opened 2024-04-29
NAS-128593 / 24.10 / Fix generating smartd.conf for items that run on Sundays

themylogin:
`0` is a valid reference to Sunday in `crontab`, but is not a valid reference to Sunday in `smartd.conf`


themylogin:
time 0:30

sonicaj (approved):

PR #13628, opened 2024-04-29
NAS-128636 / 24.10 / add back identifier key to disk.get_unused

yocalebo:
https://github.com/truenas/middleware/pull/13622 introduced a subtle regression where-by the `identifier` key wasn't included in the response. I noticed that the function that does this can be moved into its own utility function so that I don't have to go through `self.middleware` to get the information. (self.middleware does incur some overhead and we want this endpoint to be as fast as possible)

I did the same for `get_valid_zfs_partition_type_uuids`.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/utils.py
+++ b/src/middlewared/middlewared/plugins/disk_/utils.py
@@ -1,0 +1,26 @@
+def valid_zfs_partition_uuids():
+    # https://salsa.debian.org/debian/gdisk/blob/master/parttypes.cc for valid zfs types
+    # 516e7cba was being used by freebsd and 6a898cc3 is being used by linux
+    return (
+        '6a898cc3-1dd2-11b2-99a6-080020736631',
+        '516e7cba-6ecf-11d6-8ff8-00022d09712b',
+    )
+
+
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/disk_/utils.py
```diff
@@ -0,0 +1,26 @@
...
+    )
+
+
+def dev_to_ident(name, sys_disks, uuids):
+    """Map a disk device (i.e. sda5) to its respective "identifier"
+    (i.e. "{serial_lunid}AAAA_012345")"""
+    try:
+        dev = sys_disks[name]
+    except KeyError:
+        return ''
```
bmeagherix, line 16: Presumably we shouldn't ever hit this return.  Log something?

bmeagherix (approved):

src/middlewared/middlewared/plugins/disk_/utils.py
yocalebo: On the contrary. Many systems in the wild will run esoteric hardware or import disks/zpools from foreign systems that can (and often do) cause all kinds of "fun". 😄

PR #13634, opened 2024-04-30
NAS-128661 / 24.10 / Don't allow disk_resize behind a vds224r2 shelf as resizes namespace …

darkfiberiru:
…but can't reattach controller to opposite IOM

### DIFF
--- a/src/freenas/usr/local/sbin/disk_resize
+++ b/src/freenas/usr/local/sbin/disk_resize
@@ -182,6 +182,12 @@ nvme)
 		exit 1
 	fi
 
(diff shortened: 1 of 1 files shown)


yocalebo (changes requested):
This can change from vendor at any moments notice. Need to see if the nvme device has a transport address

darkfiberiru:
@cale

> This can change from vendor at any moments notice. Need to see if the nvme device has a transport address

It's a specific quark to this shelf family though? Yes, it can change but we qualify all firmware updates.

yocalebo:
Handled in PR: https://github.com/truenas/middleware/pull/13637

PR #13636, opened 2024-04-30
NAS-128673 / 24.10 / Fix SSSD startup

anodos325:
SSSD does not automatically create its cache directories on startup.

### DIFF
--- a/src/middlewared/middlewared/plugins/ldap.py
+++ b/src/middlewared/middlewared/plugins/ldap.py
@@ -2,6 +2,7 @@ import enum
 import errno
 import ipaddress
 import ldap as pyldap
+import os
 import socket
 import struct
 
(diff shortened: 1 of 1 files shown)


mgrimesix:
Do we also need to update debian/truenas-sssd.dirs?
https://github.com/truenas/sssd/blob/SCALE-v2.9-stable/debian/truenas-sssd.dirs

anodos325:
> Do we also need to update debian/truenas-sssd.dirs? https://github.com/truenas/sssd/blob/SCALE-v2.9-stable/debian/truenas-sssd.dirs

Not really because this dir is on tmpfs and would need to be recreated on reboot anyway.

bmeagherix (approved):

PR #13638, opened 2024-04-30
NAS-128677 / 24.10 / Loosen restrictions for LDAP idmap validation

anodos325:
When generating synthetic accounts (ones used for user.query and group.query results) stop requiring the LDAP UID and GID to fit within the allocated idmap ranges in winbindd. We can do this safely now that we return whether group is `local` in nss module results. This is first step toward removal of legacy samba NT domain support as well.


bmeagherix (approved):
It would also be _nice_ to take this opportunity to add a general comment for each of the three modified functions.

mgrimesix (approved):

PR #13645, opened 2024-05-01
NAS-128712 / 24.10 / Simplify user.get_user_obj and group.get_group_obj

anodos325:
For historical reasons this methods were originally a thin wrapper around dscache.get_uncached_user and dscache.get_uncached_group methods. There are no longer any direct callers of the private methods and so we can elmininate the wrapper and save an internal middleware call.


mgrimesix (approved):
Nice improvement!

PR #13649, opened 2024-05-02
NAS-128631 / 24.10 / Make sure we redirect logs to relevant log files instead of copying them

Qubad786:
## Problem

We have different log filters specified where we want to make sure that different sources i.e k3s/scst etc do not pollute other log files with their messages but unfortunately what in reality is happening is that we have their logs being logged to separate files but they are still getting logged in other log files i.e daemon.log etc which indicates that logs are being copied rather then redirected completely.

## Solution

Add changes to make sure we don't process other log filters if we have a match for a filter earlier and once we have gone through our custom log file solutions we can finally have rest of the filters for other log files being processed.

### DIFF
--- a/src/freenas/etc/logrotate.d/netdata_override
+++ b/src/freenas/etc/logrotate.d/netdata_override
@@ -1,0 +1,8 @@
+/var/log/netdata/*.log {
+    size 10M
+    rotate 7
+    compress
+    missingok
+    notifempty
+    copytruncate
+}
--- a/src/middlewared/middlewared/etc_files/syslog-ng/syslog-ng.conf.mako
+++ b/src/middlewared/middlewared/etc_files/syslog-ng/syslog-ng.conf.mako
@@ -113,6 +113,38 @@ source tn_remote_src_files {
 #######################
 # Log paths
 ########################
+log {
+  source(s_src);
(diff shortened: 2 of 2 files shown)


anodos325:
Have you verified that the logs still get forwarded correctly to remote syslog server if you configure one?

Qubad786:
> Have you verified that the logs still get forwarded correctly to remote syslog server if you configure one?

@anodos325 the logs are still being forwarded as they were before so we are good on that end

src/freenas/etc/logrotate.d/netdata_override
```diff
@@ -0,0 +1,8 @@
+/var/log/netdata/*.log {
```
sonicaj, line 1: Would this override only specific logging configuration netdata package itself provides or is like a complete override of what's specified here and original provided config is retained if that config options are not specified here ?

Qubad786: Yes so for example if `abcd` is specified in original config and is not overridden here, that will be retained when the configuration is applied as a whole ✅

sonicaj (approved):

anodos325 (approved):

PR #13650, opened 2024-05-02
Switch namespace of directory services cache

anodos325:
The directory services cache pre-dates when we had a dedicated generic plugin for directory services. Move it to the proper namespace and clean up.


src/middlewared/middlewared/plugins/directoryservices.py
```diff
@@ -330,7 +330,7 @@ def setup(self, job):
             self.middleware.call_sync('ldap.started')
 
         job.set_progress(10, 'Refreshing cache'),
-        cache_refresh = self.middleware.call_sync('dscache.refresh')
+        cache_refresh = self.middleware.call_sync('directoryservices.cache.refresh')
```
mgrimesix, line 333: Missed one at line 141?
`return await job.wrap(await self.middleware.call('dscache.refresh'))`

mgrimesix (changes requested):

PR #13653, opened 2024-05-03
Skip `Update Available` errors

themylogin:
There is no reason to have this alert if TrueNAS does not have an internet connection.


themylogin:
time 0:05

yocalebo (approved):

PR #13661, opened 2024-05-04
NAS-128762 / 24.10 / bump max_tasks_per_child to 20

yocalebo:
Back in 23.10 I changed this setting to 5. (c.f. ced399a41ca). This has proven to be fine, however, we have had quite a few users from the community express "higher CPU usage" after upgrading to 24.04.0. (c.f. https://github.com/truenas/middleware/pull/13660 for details).

During the investigation of this problem, I've realized that I can further reduce the CPU usage by bumping this setting a bit. This is not a _solution_ but is a band-aid. There are tradeoffs with changing this. Lowering the value here means, less memory consumption for the total uptime of middlewared but increase CPU usage. Bumping this will lessen the CPU usage, but will increase the memory consumption. This value is based off anecdotal investigation on a dual-core CPU system that reproduced the "higher CPU usage" issue.

Keep in mind, that before this was set to `5`, the process pool grew to very large memory sizes on high uptime systems. This is, what I would consider, "fine tuning" the setting.

### DIFF
--- a/src/middlewared/middlewared/main.py
+++ b/src/middlewared/middlewared/main.py
@@ -1337,7 +1337,7 @@ class Middleware(LoadPluginsMixin, ServiceCallMixin):
     def __init_procpool(self):
         self.__procpool = concurrent.futures.ProcessPoolExecutor(
             max_workers=5,
-            max_tasks_per_child=5,
+            max_tasks_per_child=20,
             initializer=functools.partial(worker_init, self.debug_level, self.log_handler)
         )
 


themylogin (approved):
I would even set this to 50, does it leak so fast? 5 was also really bad for performance as even loading one UI page sometimes causes worker process restart

yocalebo:
> I would even set this to 50, does it leak so fast? 5 was also really bad for performance as even loading one UI page sometimes causes worker process restart

I don't disagree but if we have a UI page that is making > 5 calls to endpoints that require our process pool, then we should fix that (or investigate to make sure it's not by mistake) :smile:

PR #13662, opened 2024-05-04
NAS-128772 / 24.10 / cache boot disks

yocalebo:
Another area for improvement that I found while investigating (c.f. https://github.com/truenas/middleware/pull/13660) is that I noticed 4 duplicate calls to our process pool were being made on middlewared startup. The duplicate calls were determining the disks of the boot pool. The particular system I investigated was an HA machine. The amount of these calls on non-HA would be half (so 2).

Either way, the boot disk information can be easily cached. This information rarely changes and when it does, we can reset the cache. On the system I tested these changes on, it reduced the number of calls to 1 (instead of 4) for determining the disks in the boot pool.

NOTE: Since this is a globally cached value, I've converted the object type to be immutable (tuple) from mutable (list). I've updated the callers accordingly.

### DIFF
--- a/src/middlewared/middlewared/plugins/boot.py
+++ b/src/middlewared/middlewared/plugins/boot.py
@@ -8,7 +8,7 @@ from middlewared.utils import run
 from middlewared.validators import Range
 
 BOOT_ATTACH_REPLACE_LOCK = 'boot_attach_replace'
-BOOT_POOL_NAME = None
+BOOT_POOL_NAME = BOOT_POOL_DISKS = None
 BOOT_POOL_NAME_VALID = ['freenas-boot', 'boot-pool']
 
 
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/boot.py
```diff
@@ -112,8 +135,8 @@ async def attach(self, job, dev, options):
         # If the user is upgrading his disks, let's set expand to True to make sure that we
         # register the new disks capacity which increase the size of the pool
         await self.middleware.call('zfs.pool.online', BOOT_POOL_NAME, zfs_dev_part['name'], True)
-
         await self.update_initramfs()
+        await self.clear_disks_cache()
```
themylogin, line 139: If update-initramfs fails then the cache will not be cleared, even though pool already has other disks

themylogin (commented):
Should we clear the cache on ZFS events instead? That way it will reflect the correct pool state even if it was modified outside of middleware.

yocalebo:
> Should we clear the cache on ZFS events instead? That way it will reflect the correct pool state even if it was modified outside of middleware.

That's a better idea. Updated it to do just that.

src/middlewared/middlewared/plugins/boot.py
```diff
@@ -34,13 +34,36 @@ async def get_state(self):
...
+        very often and we have a ton of callers (especially on HA)
+        that need to determine this information. By caching this,
+        it reduces the amount of times we have to use our ProcessPool
+        """
+        global BOOT_POOL_DISKS
+        if BOOT_POOL_DISKS is None:
+            # Using an immutable object is very important since this is
+            # a globally cached value
+            BOOT_POOL_DISKS = tuple(await self.middleware.call('zfs.pool.get_disks', BOOT_POOL_NAME))
+        return BOOT_POOL_DISKS
```
themylogin, line 52: `return list(BOOT_POOL_DISKS)` here and you will never have to make a copy in callers.

themylogin (changes requested):

src/middlewared/middlewared/plugins/boot.py
yocalebo: ahh yes, much cleaner that way!

PR #13669, opened 2024-05-06
NAS-128802 / 24.10 / Improve configuration file generation

anodos325:
1. use os.rename to overwrite config files. Avoid possibility of having applications read partially written configuration files by first writing the changed config to a temporary file, then renaming over existing file.

2. have write_if_changed allow specifying the desired permissions and ownership of the file. This allows us to ensure that ownership and permissions changes happen before we close and rename the file.

3. alter return value for write_if_changed from boolean indicating whether changes were written to a bitmask of changes made. This allows us to flag when permissions and ownership had to be altered (logging potentially unexpected CLI changes).

4. add optional non-default parameter to raise an exception if we had to change ownership and permissions. This exception can be used in future to generate audit messages / alerts if we had unexpected config file changes.

5. add tests for functionality

### DIFF
--- a/src/middlewared/middlewared/plugins/etc.py
+++ b/src/middlewared/middlewared/plugins/etc.py
@@ -6,10 +6,11 @@ import stat
 
 from mako import exceptions
 from middlewared.service import CallError, Service
-from middlewared.utils.io import write_if_changed
+from middlewared.utils.io import write_if_changed, UnexpectedFileChange
 from middlewared.utils.mako import get_template
 
 DEFAULT_ETC_PERMS = 0o644
+DEFAULT_ETC_XID = 0
 
 
 class FileShouldNotExist(Exception):
@@ -325,36 +326,15 @@ class EtcService(Service):
 
         return rv
 
-    def set_etc_file_perms(self, fd, entry):
-        perm_changed = False
+    def get_perms_and_ownership(self, entry):
         user_name = entry.get('owner')
         group_name = entry.get('group')
         mode = entry.get('mode', DEFAULT_ETC_PERMS)
 
-        if all(i is None for i in (user_name, group_name, mode)):
-            return perm_changed
+        uid = self.middleware.call_sync('user.get_builtin_user_id', user_name) if user_name else DEFAULT_ETC_XID
+        gid = self.middleware.call_sync('group.get_builtin_group_id', group_name) if group_name else DEFAULT_ETC_XID
 
-        uid = self.middleware.call_sync('user.get_builtin_user_id', user_name) if user_name else -1
-        gid = self.middleware.call_sync('group.get_builtin_group_id', group_name) if group_name else -1
-        st = os.fstat(fd)
-        uid_to_set = -1
-        gid_to_set = -1
-
-        if uid != -1 and st.st_uid != uid:
-            uid_to_set = uid
-
-        if gid != -1 and st.st_gid != gid:
-            gid_to_set = gid
-
-        if gid_to_set != -1 or uid_to_set != -1:
-            os.fchown(fd, uid_to_set, gid_to_set)
-            perm_changed = True
-
-        if mode and stat.S_IMODE(st.st_mode) != mode:
-            os.fchmod(fd, mode)
-            perm_changed = True
-
-        return perm_changed
+        return {'uid': uid, 'gid': gid, 'perms': mode}
 
     def make_changes(self, full_path, entry, rendered):
         mode = entry.get('mode', DEFAULT_ETC_PERMS)
@@ -366,11 +346,14 @@ class EtcService(Service):
         if outfile_dirname != '/etc':
             os.makedirs(outfile_dirname, exist_ok=True)
 
-        with open(full_path, "w", opener=opener) as f:
-            perms_changed = self.set_etc_file_perms(f.fileno(), entry)
-            contents_changed = write_if_changed(f.fileno(), rendered)
+        payload = self.get_perms_and_ownership(entry)
+        try:
+            changes = write_if_changed(full_path, rendered, raise_error=True, **payload)
+        except UnexpectedFileChange as e:
+            self.logger.error(str(e))
+            changes = e.changes
 
-        return perms_changed or contents_changed
+        return changes
 
     async def generate(self, name, checkpoint=None):
         group = self.GROUPS.get(name)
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/etc.py
```diff
@@ -366,11 +346,14 @@ def opener(path, flags):
...
             os.makedirs(outfile_dirname, exist_ok=True)
 
-        with open(full_path, "w", opener=opener) as f:
-            perms_changed = self.set_etc_file_perms(f.fileno(), entry)
-            contents_changed = write_if_changed(f.fileno(), rendered)
+        payload = self.get_perms_and_ownership(entry)
+        try:
+            changes = write_if_changed(full_path, rendered, raise_error=True, **payload)
+        except UnexpectedFileChange as e:
+            self.logger.error(str(e))
```
yocalebo, line 353: Do we not just want to do `exc_info=True` here?

yocalebo: We should also probably catch all exceptions here actually.

anodos325: Yeah, that's a good idea, I got side-tracked by writing tests.

yocalebo (approved):
nice!

src/middlewared/middlewared/utils/io.py
```diff
@@ -1,24 +1,124 @@
...
     if isinstance(data, str):
         data = data.encode()
 
-    if isinstance(path, int):
-        path = f'/proc/self/fd/{path}'
+    if not isinstance(perms, int):
+        raise ValueError('perms must be an integer')
+
+    if perms & ~(stat.S_IRWXU | stat.S_IRWXG | stat.S_IRWXO):
+        raise ValueError(f'{perms}: invalid mode')
```
themylogin, line 51: Why is it invalid?

src/middlewared/middlewared/plugins/etc.py
```diff
@@ -366,11 +346,20 @@ def opener(path, flags):
         if outfile_dirname != '/etc':
             os.makedirs(outfile_dirname, exist_ok=True)
 
-        with open(full_path, "w", opener=opener) as f:
-            perms_changed = self.set_etc_file_perms(f.fileno(), entry)
-            contents_changed = write_if_changed(f.fileno(), rendered)
+        payload = self.get_perms_and_ownership(entry)
+        try:
+            changes = write_if_changed(full_path, rendered, raise_error=True, **payload)
+        except UnexpectedFileChange as e:
```
themylogin, line 352: Should we just log this in the `write_if_changed` instead of handling this as an exception? I feel like this try/except will be reused a lot. Also raising an exception has a connotation that the changes were not written.

src/middlewared/middlewared/utils/io.py
```diff
@@ -1,24 +1,124 @@
...
+
+    `data` - expected file contents. May be bytes or string
+
+    `uid` - expected numeric UID for file
+
+    `gid` - expected numeric GID for file
+
+    `perms` - numeric permissions that file should have
+
+    `raise_error` - raise an exception if file ownership or permissions have unexpectedly
```
mgrimesix, line 40: Looks like raise_error covers data changes too?

src/middlewared/middlewared/plugins/etc.py
anodos325: That's reasonable.

src/middlewared/middlewared/utils/io.py
```diff
@@ -1,24 +1,124 @@
...
+            os.fchmod(tmp.file.fileno(), perms)
+            os.fchown(tmp.file.fileno(), uid, gid)
+            source_path = tmp.name
+
+        if dirfd is not None:
+            os.rename(source_path, path, src_dir_fd=dirfd, dst_dir_fd=dirfd)
 
-    changed = False
+        else:
+            os.rename(source_path, path)
```
mgrimesix, line 119: With regards to write, flush, fsync, fchmod, fchown,  how are exceptions or failures handled or passed to the caller?

src/middlewared/middlewared/utils/io.py
```diff
@@ -1,24 +1,124 @@
...
+                changes |= FileChanges.UID
+
+            if st.st_gid != gid:
+                changes |= FileChanges.GID
+
+            if changes & (FileChanges.UID | FileChanges.GID):
+                os.fchown(f.fileno(), uid, gid)
+
+            if changes & FileChanges.PERMS:
+                os.fchmod(f.fileno(), perms)
```
bmeagherix, line 99: Could put these under an if to skip if contents have changed (to avoid doing the `fchown` or `fchmod` twice - on the file itself, and then on the temp file).

src/middlewared/middlewared/utils/io.py
anodos325: Okay. I'll specify that we only allow setting RWX on user, group, other.

src/middlewared/middlewared/utils/io.py
anodos325: Passed back up to caller (keeping existing behavior).

src/middlewared/middlewared/utils/io.py
```diff
@@ -1,24 +1,124 @@
...
+            raise ValueError(f'{name} must be between 0 and {ID_MAX}')
+
+    if dirfd is not None:
+        if not isinstance(dirfd, int):
+            raise ValueError('dirfd must be a valid file descriptor')
+
+        if not os.path.exists(f'/proc/self/fd/{dirfd}'):
+            raise ValueError(f'{dirfd}: file descriptor not found')
+
+    if dirfd is not None:
```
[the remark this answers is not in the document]
anodos325: os.open() will fail if it's not a directory.

src/middlewared/middlewared/utils/io.py
anodos325: I was thinking that if for some reason the tempfile write fails or rename fails we still want to correct the permissions and ownership on the original file.

bmeagherix (approved):
LGTM!

src/middlewared/middlewared/pytest/unit/utils/test_write_if_changed.py
```diff
@@ -0,0 +1,192 @@
...
+    os.unlink(target)
+
+
+def test__write_file_exceptions(create_etc_dir):
+    target = os.path.join(create_etc_dir, 'testfile6')
+
+    changes = write_if_changed(target, "canary")
+    assert changes == FileChanges.CONTENTS
+
+
```
mgrimesix, line 129: flake8 will probably complain about this extra line.

mgrimesix (approved):

PR #13670, opened 2024-05-07
NAS-128814 / 24.10 / Remove legacy samba domain support

anodos325:
DragonFish was last release where we support this legacy feature. This commit removes most of middleware code, but leaves database untouched. Next related PR for FreeIPA support will finalize schema changes.


src/middlewared/middlewared/plugins/smb_/util_smbconf.py
```diff
@@ -262,17 +234,6 @@ def generate_smb_conf_dict(
                 else:
                     domain = smb_service_config['workgroup']
             case 'DS_TYPE_LDAP':
-                if not ldap_enabled or not lc['has_samba_schema']:
```
mgrimesix, line 265: Maybe add a comment why this is an empty case.
We _could_ add a log message if something is unexpectedly calling it.

mgrimesix (approved):
99% removal.

src/middlewared/middlewared/plugins/ldap.py
```diff
@@ -263,7 +263,6 @@ class Config:
         Int('dns_timeout', default=5),
         Int('kerberos_realm', null=True),
         Str('kerberos_principal'),
-        Bool('has_samba_schema', default=False),
```
bmeagherix, line 266: Just a question.  I believe removing this equates to a public API change, and can see the change in docs for the `do_update`.  Is this all that needs to occur for a public API change?

anodos325: Yes. That's why one needs to be somewhat careful about changing a public accepts decorator.

PR #13672, opened 2024-05-07
NAS-128822 / 24.10 / Log container registry exceptions in separate file

sonicaj:
Changes have been made to add image's registry errors to separate file `/var/log/container_images.log` to avoid unnecessary noise in middleware logs.


anodos325:
Do you think these are something a user would want to get forwarded to remote syslog server?

sonicaj:
@anodos325 not really, lots of container registries do not conform to standards which means this is just an awful lot of noise with no really good reason to keep i.e does not really help in debugging why an app is not running or anything like that - this information is retrieved just to see if an update image is available for a particular image. I think we are good to go without sending it to a remote syslog server

Qubad786 (approved):

PR #13674, opened 2024-05-07
NAS-128828 / 24.10 / Fix off-by-one in getpwall and getgrall

anodos325:
We were omitting first NSS entry. Add basic CI tests to ensure
that our pwd and grp output matches what is expected from base
python module.


anodos325:
Needs tests.

bmeagherix (approved):

PR #13680, opened 2024-05-08
NAS-127870 / 24.10 / fix starting apps with a bridge interface

yocalebo:
Since Cobia, users that use apps with a bridge interface have been broken because `kubernetes.validate_interfaces` was too strict on checking whether or not the interface is LINK_STATE_UP. It would only check once and then bail out.

This makes it so that instead of bailing out immediately when the interface isn't up, we actually wait for a bit for it to come online. This was reproduced on a system 100% of the time and after these changes, apps successfully started after a reboot.

I took the liberty to improve this method while we're here because it was calling `interface.query` too many times for no reason.

Finally, and arguably the most frustrating part of this endeavor is that, I fixed the fact that `kubernetes.validate_k8s_fs_setup` was swallowing errors from `validate_interfaces`. It did more than swallow the actual error, it showed a _VERY_ misleading and incorrect error message. I thought, for a long time, that the `br*` interface DIDN'T EXIST by the time this method was started but that's not the case. The issue is that it wasn't up. This makes it so that the caller where this was failing shows the proper error message.

The error message received before these changes was
```
  File "/usr/lib/python3/dist-packages/middlewared/plugins/kubernetes_linux/lifecycle.py", line 231, in validate_k8s_fs_setup
    raise CallError(f'Unable to lookup configured interfaces: {", ".join([v[1] for v in iface_errors])}')
middlewared.service_exception.CallError: [EFAULT] Unable to lookup configured interfaces: br0
```
when the ACTUAL error message was
```
[2024/05/08 08:26:12] (DEBUG) KubernetesService.validate_interfaces():392 - errors: [('route_v4_interface', 'br0', "Please specify a valid interface which is active (i.e '').")]
```

NOTE: I don't believe this will solve ALL issues surrounding this area since all hardware/software behaves wildly differently but this is an easy stop-gap solution.

### DIFF
--- a/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
+++ b/src/middlewared/middlewared/plugins/kubernetes_linux/update.py
@@ -1,6 +1,8 @@
+import contextlib
 import ipaddress
 import itertools
 import json
+import time
 import os.path
 
 import middlewared.sqlalchemy as sa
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/kubernetes_linux/update.py
```diff
@@ -369,23 +371,46 @@ async def validate_data(self, data, schema, old_data):
 
         verrors.check()
 
+    @private
+    def wait_on_interface_link_state_up(self, interface):
+        max_wait, sleep_time, time_waited = 60, 1, 0
```
anodos325, line 376: Would be good to at least have `max_wait` defined as a constant somewhere.

anodos325 (approved):

src/middlewared/middlewared/plugins/kubernetes_linux/update.py
```diff
@@ -369,23 +373,46 @@ async def validate_data(self, data, schema, old_data):
 
         verrors.check()
 
+    @private
+    def wait_on_interface_link_state_up(self, interface):
+        sleep_time, time_waited = 1, 0
+        is_up, error_msg = False, ''
+        while time_waited < IFACE_LINK_STATE_MAX_WAIT:
+            time.sleep(sleep_time)
+            time_waited += 1
```
aiden3d, line 382: This may be pedantic, but wouldn't we want `time_waited += sleep_time`? In this case since it's just 1 it doesn't really matter.

PR #13682, opened 2024-05-08
NAS-128839 / 24.10 / Fix modernize ad tests

bmeagherix:
CI test runs against HA were failing some tests in _test_030_activedirectory.py_ (e.g. `test_07_enable_leave_activedirectory`).  Modernize and tweak so that they work against both single-node and HA.

Note: had to add `dnspython` to the test requirements.txt in order to be able to trap `dns.resolver.NXDOMAIN` exceptions.

### DIFF
--- a/tests/api2/test_030_activedirectory.py
+++ b/tests/api2/test_030_activedirectory.py
@@ -1,26 +1,24 @@
 import ipaddress
-import errno
 import os
-import sys
 from time import sleep
 
+import dns.resolver
 import pytest
-from pytest_dependency import depends
-
-apifolder = os.getcwd()
-sys.path.append(apifolder)
-from auto_config import ip, ha
-from functions import GET, POST, make_ws_request
-from protocols import smb_connection, smb_share
-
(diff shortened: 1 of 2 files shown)


tests/api2/test_030_activedirectory.py
```diff
@@ -1,26 +1,24 @@
...
-from middlewared.test.integration.assets.directory_service import active_directory, override_nameservers
-from middlewared.test.integration.utils import call, ssh, client
 from middlewared.test.integration.assets.product import product_type
+from middlewared.test.integration.utils import call, client, ssh
+from pytest_dependency import depends
+
+from auto_config import ha, ip, vip
+from protocols import smb_connection, smb_share
 
+public_ip = vip if ha else ip
```
anodos325, line 21: You can:
```
from middlewared.test.integration.utils.client import client, truenas_server
```
and just use `truenas_server.ip` where tests need it.

anodos325 (approved):

PR #13684, opened 2024-05-08
NAS-128471 / 24.10 / Fix app stats when they are using host network

Qubad786:
## Problem
The network statistics for apps appear unusually high when the app is utilizing the host network. Upon investigation, it was discovered that for apps using the host network, Kubernetes employs the kube-bridge interface. Consequently, traffic passing from kube-bridge to the physical interface is duplicated, as traffic from all interfaces used by the pod is aggregated.

## Solution
When a pod is utilizing the kube-bridge interface, only consider the traffic associated with the kube-bridge interface and disregard traffic from other interfaces. This approach acknowledges that all pod traffic traverses through the kube-bridge interface.


sonicaj (commented):
Nice catch!

sonicaj (approved):

PR #13692, opened 2024-05-09
NAS-128882 / 24.10 / Improve mDNS configuration generation

anodos325:
* create general-purpose utility functions to generate valid mDNS service records since there has been some discussion of providing ability to create custom advertisements for services provided by apps.

* ensure that service definitions are written separately so that we can properly log errors regarding permissions and ownership.

* use standard function to write configuration files in etc plugin so that we can avoid unnecessary service toggles during config file writes.

* Add tests for formatting of files.


src/middlewared/middlewared/etc_files/local/avahi/services/ADISK.service.py
```diff
@@ -0,0 +1,42 @@
+from middlewared.plugins.etc import FileShouldNotExist
+from middlewared.utils import filter_list
```
mgrimesix, line 2: Nit: filter_list is not used.
Here and in the other service files.

mgrimesix (approved):

src/middlewared/middlewared/etc_files/local/avahi/services/ADISK.service.py
anodos325: Good catch.

bmeagherix (approved):

src/middlewared/middlewared/etc_files/local/avahi/services/ADISK.service.py
```diff
@@ -0,0 +1,42 @@
...
+from middlewared.utils import mdns
+
+
+def render(service, middleware, render_ctx):
+
```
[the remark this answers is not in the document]
anodos325: Oh, right, I'll throw that in the adisk python script.

PR #13701, opened 2024-05-10
API call for vm.import_disk_image

kmoore134:
Bar the gates. Hide the children. Cower in fear. Kris has gone and tried his hand at Python. Yes, you should be that afraid, here be :dragon: 

This new disk_utils.py extension to our vm plugin provides an endpoint which can be used to convert and import a VM disk image (VMDK, QCOW2, VDI, etc) to a specified ZVOL for usage by a VM. 

It takes two arguments, the local location of the VM disk image, and then the target zvol. In the background it runs `qemu-img` to do the conversion / copy. It reports progress along the way since the import may take a little while depending on image size, pool speeds, etc.

Special Thank you to @william-gr  for a review of my noob Python code.


yocalebo:
@themylogin need you to spot-check me on the `lock=` arguments in this PR. IIRC, we had weirdness when 2 methods share the same lock name and `lock_queue_size=1`.

themylogin (approved):
We have a problem where we can't share the lock name between synchronous and asynchronous methods because they use different lock implementations. but here both methods are synchronous so it shouldn't be a problem. LGTM.

yocalebo (approved):

PR #13703, opened 2024-05-10
NAS-128912 / 24.10 / Skip service announcements for uninteresting services

bmeagherix:
In `AvahiBrowserCollector` skip service announcements for services we don't care about

Was encountering
```
>                       service_type = AvahiBrowserCollector.name_to_service[items[4]]
E                       KeyError: '_nut._tcp'
```

### DIFF
--- a/tests/api2/test_310_service_announcement.py
+++ b/tests/api2/test_310_service_announcement.py
@@ -1,31 +1,26 @@
 import contextlib
-import os
 import random
 import re
 import socket
 import string
-import sys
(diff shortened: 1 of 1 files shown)


anodos325:
This is actually a regression due to mDNS changes (we used to auto-delete unexpected files there).

anodos325:
I think it would be better to remove all SRV records when making ISO / update files (if possible). The SRV nonsense is coming from the nut debian package.

PR #13704, opened 2024-05-13
NAS-128961 / 24.10 / Improve CDROM device validation

Qubad786:
### Context

Changes have been added to enhance validation for CDROM devices and avoid adding path which are not a file or are not under `/mnt`.

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/devices/cdrom.py
+++ b/src/middlewared/middlewared/plugins/vm/devices/cdrom.py
@@ -1,8 +1,10 @@
 import os
 
-from middlewared.schema import Dict, Str
+from middlewared.plugins.boot import BOOT_POOL_NAME
+from middlewared.schema import Dict, File
 from middlewared.service import CallError
-from middlewared.validators import Match
+from middlewared.utils.zfs import query_imported_fast_impl
+from middlewared.validators import check_path_resides_within_volume_sync, Match
 
 from .device import Device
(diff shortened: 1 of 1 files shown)


anodos325 (changes requested):
We should use standard path validator for middleware unless there is specific reason not to.

anodos325 (commented):
I was referring to our async path validator

anodos325:
https://github.com/truenas/middleware/pull/13741/files
I added a sync version of path validator.

src/middlewared/middlewared/plugins/vm/devices/cdrom.py
```diff
@@ -40,9 +43,10 @@ def xml(self, *args, **kwargs):
 
     def _validate(self, device, verrors, old=None, vm_instance=None, update=True):
         path = device['attributes']['path']
-        if not os.path.exists(path):
-            verrors.add('attributes.path', f'Unable to locate CDROM device at {path!r}')
-        elif not self.middleware.call_sync('vm.device.disk_uniqueness_integrity_check', device, vm_instance):
+        check_path_resides_within_volume_sync(
+            verrors, 'attributes.path', path, [p['name'] for p in self.middleware.call_sync('pool.query')]
```
anodos325, line 47: Since we only care about imported pools you can avoid expensive call by using `query_imported_fast_impl` from `middlewared.utils.zfs`.

anodos325 (approved):

PR #13705, opened 2024-05-13
NAS-128590 / 24.10 / Fix Range validator args

sonicaj:
This commit fixes an issue where when we change max to max_ in middleware api for validators, this change was not addressed.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/schema.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/schema.py
@@ -114,10 +114,10 @@ def get_schema(variable_details, update, existing=NOT_PROVIDED):
             obj.enum = [v['value'] for v in schema_details['enum']]
 
         if schema_class == Str:
-            if range_args.get('max'):
+            if range_args.get('max_'):
                 # This needs to be done as string schema has built in support for max length as
                 # well apart from the range validator we add
-                obj.max_length = range_args['max']
+                obj.max_length = range_args['max_']
             if 'valid_chars' in schema_details:
                 obj.validators.append(Match(
                     schema_details['valid_chars'], explanation=schema_details.get('valid_chars_error')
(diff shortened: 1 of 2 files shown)


yocalebo:
@sonicaj why is this a draft?

yocalebo (changes requested):
This seems fundamentally broken. I think I see other areas in this file where it's using `max` for other schema classes. I don't understand why we're writing another abstraction layer on top of our schema classes because errors like these should have been caught at CI time on github. Not after the fact.

sonicaj:
@yocalebo that is the `max/min` app devs specify in `questions.yaml` - we internally changed our API to use `max_` instead. So it seems very weird to use `max_` in yaml file to outline maximum length, we internally changed it because it collided with python builtins. Also it would be very messy to have app devs switch to `max_` if we wanted to (i don't think we should), as older apps which app devs have not updated will no longer work and would be a decent change depending on the size of catalog.
I think we should have tests in place to catch this sooner, i'll write some and push them if that sounds good ?

yocalebo:
> @yocalebo that is the `max/min` app devs specify in `questions.yaml` - we internally changed our API to use `max_` instead. So it seems very weird to use `max_` in yaml file to outline maximum length, we internally changed it because it collided with python builtins. Also it would be very messy to have app devs switch to `max_` if we wanted to (i don't think we should), as older apps which app devs have not updated will no longer work and would be a decent change depending on the size of catalog. I think we should have tests in place to catch this sooner, i'll write some and push them if that sounds good ?

Yep agree, tests would be a good start!

yocalebo:
Are you sure this fixes all the failures? On lines 108, 109 I still see references to `min` and `max`

sonicaj:
> Are you sure this fixes all the failures? On lines 108, 109 I still see references to `min` and `max`

@yocalebo yes I am and i have added a unit test as well to outline that. Those `min/max` references are to the values in `questions.yaml` and not our schema so we are safe on that end

yocalebo (approved):

PR #13707, opened 2024-05-13
NAS-128851 / 24.10 / Enhance permissions check on data directory

Qubad786:
### Context

Newly added integration test checks the permission of `/data` dir and its contents to ensure that perms are like below: 

1. `u=rwX,g=,o=` on `/data` recursively ensuring every dir inside has `0o700` perms and every file has `0o600`
2. `0o755` on just `/data` itself
3. `0o755` recursively on `/data/subsystems`

Scale build change was added in the [PR](https://github.com/truenas/scale-build/pull/635).

### DIFF
--- a/tests/api2/test_007_early_settings.py
+++ b/tests/api2/test_007_early_settings.py
@@ -1,3 +1,5 @@
+import stat
+
 from auto_config import ha
 from middlewared.test.integration.utils import call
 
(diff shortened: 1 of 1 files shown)


tests/api2/test_007_early_settings.py
```diff
@@ -14,13 +14,40 @@ def test_sysctl_arc_max_is_set():
...
+def test_data_dir_perms():
+    default_dir_mode = 0o700
+    default_file_mode = 0o600
+    data_subsystems_mode = 0o755
+
+    def check_permissions(directory):
+        contents = call('filesystem.listdir', directory)
+        for i in contents:
+            i_path = i['path']
+            i_mode = call('filesystem.stat', i_path)['mode']
```
anodos325, line 26: Why are we calling stat for these contents. Don't we already have `mode` in listdir output?

tests/api2/test_007_early_settings.py
```diff
@@ -14,13 +14,40 @@ def test_sysctl_arc_max_is_set():
...
+            if i['type'] == 'DIRECTORY':
```
anodos325, line 27: If we have mode in listdir output then you can simply import `stat` module and call `stat.S_ISDIR(i['mode'])`

tests/api2/test_007_early_settings.py
```diff
@@ -14,13 +14,40 @@ def test_sysctl_arc_max_is_set():
...
+                assert i_mode & 0o777 == data_subsystems_mode, \
```
anodos325, line 28: `stat.S_IMODE(i_mode)  == data_subsystems_mode` perhaps

anodos325 (approved):

PR #13710, opened 2024-05-14
NAS-128988 / 24.04.1 / disable swap by default

yocalebo:
This is a safety belt for 24.04.1. A more involved and proper fix will follow but this solves the immediate issue.

NOTE: if a user wants to re-enable the swap partition(s), they will need to run `swapon -a` manually or use a post-init script (which is what the overwhelming majority of users are doing on the forums, except they're doing `swapoff -a` :smile: )


kmoore134 (approved):

yocalebo:
backport

PR #13711, opened 2024-05-14
NAS-128993 / 24.10 / Enhance WebSocket Handling with Dynamic Connection Header for VM Display devices

Qubad786:
## Problem
WebSocket connections require a long-running connection between the client and server. When using WebSocket connections with a reverse proxy in Nginx, a problem arises because reverse proxy connections follow the hop-by-hop protocol, meaning they open once and then close. To upgrade this connection to support WebSocket, the Upgrade header is used, which keeps the connection open on every hop. The main issue in the code is that we are not closing the connection ever. It remains open because this header is hard-coded. As a result, when a user opens their VM display more than 1000 times without rebooting, they start experiencing problems with the display.

## Solution
Close the WebSocket connection when `http_upgrade` is none.

## Reference
https://stackoverflow.com/questions/49488091/getting-connection-reset-by-peer-while-proxying-upgraded-connection-connectio

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
+++ b/src/middlewared/middlewared/etc_files/local/nginx/nginx.conf.mako
@@ -127,6 +127,11 @@ http {
         keepalive 64;
     }
 
+    map $http_upgrade $connection_upgrade {
+        default upgrade;
+        '' close;
+    }
+
     server {
         server_name  localhost;
 % if ssl_configuration:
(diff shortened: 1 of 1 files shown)


sonicaj (approved):
Does this solve the case as well where if display device is left idle for sometime, the page gets unresponsive ?

Qubad786:
@sonicaj that unfortunately is a different issue and not resolved by this

PR #13713, opened 2024-05-14
NAS-128634 / 24.10 / Show IPv6 autoconf default, also reflect preference with sysctl

aiden3d:
When creating a network interface, the system automatically configures IPv6 despite the default setting being False. In the UI, this creates an issue where the "Autoconfigure IPv6" checkbox is unchecked but the value under `/proc/sys/net/ipv6/conf/[interface]/autoconf` is 1. 

This is also true of DHCP, but was fixed in the UI by setting the value to True specifically if we haven't created/configured the interface yet. I applied the same fix to `ipv6_auto`. I've also made it set the proper `/proc/.../autoconf` value when an existing interface is updated so the user's settings are respected.


yocalebo (changes requested):
Looks great! However, we should add some integration tests. You can find them in `tests/api2` directory.

sonicaj (commented):
I think we have git merge issues here, can we please fix them so the diff can be better reviewed ? Thanks!

PR #13726, opened 2024-05-15
NAS-128889 / 24.10 / Remove ability to configure swap from the API

Qubad786:
## Context

We now don't try to configure or manage swap and finally now we are removing it from our API as well so consumers won't be able to configure this anymore via our API.


sonicaj (approved):

yocalebo (changes requested):
I think this is the perfect opportunity to remove the `create_swap` references through the API. Can you please go ahead and do that in this PR since you're making changes to the db.

yocalebo (approved):

PR #13727, opened 2024-05-15
NAS-128998 / 24.04.2 / prevent middleware service from starting when config is reset

yocalebo:
If someone resets the configuration database to defaults, middlewared service will not start-up on config upload. The reason why is because the netdata plugin isn't handling this specific case. This does 2 things:

1. be extra safe and catch all exceptions here so that we don't prevent the middlewared service from starting
2. use the proper methods to return the builtin user `netdata`

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/netdata_configure.py
+++ b/src/middlewared/middlewared/plugins/reporting/netdata_configure.py
@@ -13,8 +13,20 @@ class ReportingService(Service):
 
     @private
     async def cache_netdata_uid_gid(self):
-        self.NETDATA_UID = await self.middleware.call('user.get_builtin_user_id', NETDATA_USERNAME)
-        self.NETDATA_GID = await self.middleware.call('group.get_builtin_user_id', NETDATA_GROUPNAME)
+        # this method runs in middleware init phase so it's important
+        # to handle errors and log a message instead of crashing. If
+        # we crash here, middleware service won't start
+        try:
+            self.NETDATA_UID = await self.middleware.call('user.get_builtin_user_id', NETDATA_USERNAME)
+        except Exception:
+            self.logger.error('Unexpected failure resolving username %r to uid', NETDATA_USERNAME, exc_info=True)
+            return
+
+        try:
+            self.NETDATA_GID = await self.middleware.call('group.get_builtin_user_id', NETDATA_GROUPNAME)
+        except Exception:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/reporting/netdata_configure.py
```diff
@@ -11,9 +13,20 @@ class ReportingService(Service):
...
-        user_obj = await self.middleware.call('user.get_user_obj', {'username': 'netdata'})
-        self.NETDATA_UID = user_obj['pw_uid']
-        self.NETDATA_GID = user_obj['pw_gid']
+        # this method runs in middleware init phase so it's important
+        # to handle errors and log a message instead of crashing. If
+        # we crash here, middleware service won't start
+        try:
+            self.NETDATA_UID = await self.middleware.call('user.get_builtin_user_id', NETDATA_USERNAME)
+        except Exception:
+            self.logger.error('Unexpected failure resolving username %r to uid', NETDATA_USERNAME, exc_info=True)
```
anodos325, line 22: Maybe also set `self.NETDATA_UID = -1` (and same for GID) on error because this will be passed to chown.

anodos325: So that `netdata_setup()` doesn't break.

themylogin:
@yocalebo is `netdata` user not being created by `account_/sync_builtin.py` as it should? This should run prior to `reporting` plugin setup as indicated in `__plugins_setup`

anodos325:
> @yocalebo is `netdata` user not being created by `account_/sync_builtin.py` as it should? This should run prior to `reporting` plugin setup as indicated in `__plugins_setup`

The problem is that the method is running during middleware setup and so crashing with KeyError (user doesn't exist) before we sync builtins.

themylogin:
@yocalebo it should not run before reporting setup

themylogin:
@yocalebo `run_migration` key is not set on config upload, that's why `netdata` user is not present.

themylogin:
@yocalebo https://github.com/truenas/middleware/pull/13728

src/middlewared/middlewared/plugins/reporting/netdata_configure.py
yocalebo: done

yocalebo:
> @yocalebo #13728

okay thanks, the change I've introduced here is still necessary nonetheless.

themylogin (approved):
I don't see this as necessary since netdata user should always exist, but ok

PR #13736, opened 2024-05-15
NAS-129019 / 24.10 / More statx improvements

anodos325:
Bitwise comparisons with IntFlag enum types can incur significant cost if performing it repeatedly since it requires creating a new enum instance. Avoid this where possible. This also improves API for statx to make it more consistent with syscall interface.

### DIFF
--- a/src/middlewared/middlewared/pytest/unit/utils/test_statx.py
+++ b/src/middlewared/middlewared/pytest/unit/utils/test_statx.py
@@ -96,7 +96,7 @@ def test__check_statx_empty_path(tmpdir):
     stx1 = sx.statx(testfile)
     try:
         fd = os.open(testfile, os.O_PATH)
-        stx2 = sx.statx('', dir_fd=fd, flags=sx.ATFlags.EMPTY_PATH)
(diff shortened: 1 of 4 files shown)


yocalebo (approved):

src/middlewared/middlewared/pytest/unit/utils/test_statx.py
```diff
@@ -96,7 +96,7 @@ def test__check_statx_empty_path(tmpdir):
     stx1 = sx.statx(testfile)
     try:
         fd = os.open(testfile, os.O_PATH)
-        stx2 = sx.statx('', {'dir_fd': fd, 'flags': sx.ATFlags.EMPTY_PATH})
+        stx2 = sx.statx('', dir_fd=fd, flags=sx.ATFlags.EMPTY_PATH)
```
bmeagherix, line 99: Should this be `sx.ATFlags.EMPTY_PATH.value`

anodos325: Either works since it's IntFlag enum and not performance critical.

bmeagherix (approved):

PR #13695, opened 2024-05-09
Remove the creation of swap partitions

yocalebo:
The remnants of this legacy behavior lurks in eldritch obscurity. Those who dare to comprehend its incomprehensible logic fall into a deep state of despair. As I delved into the tangled syntax where logic intertwines with madness, trying to remove this forbidden code seemed to resist, as if it were a living, breathing entity, guarding its existence through screens flickering with cryptic errors and warnings that chill the spine. Alas, the victor of this battle goes to sanity. The war, however, is not over. For this code is a labyrinthine whose tentacles stretch deep into the very fabric of TrueNAS. More changes will be needed but one must start somewhere.

PR #13738, opened 2024-05-16
NAS-129028 / 24.10 / Reintroduce device.settle_udev_events to avoid race

bmeagherix:
Found that CI test `test_002_create_permanent_zpool` started (intermittently) failing recently (following PR #13695 I believe).

The headline was
```
AssertionError: Unable to create test pool: ClientException("[Errno 2] No such file or directory: '/sys/block/sdf/sdf1/start'"). Aborting tests.
```

Reintroducing the `device.settle_udev_events` call removed in the above PR seems to rectify the issue.

----
Tested in CI runs #[399 ](http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/399/) and #[400](http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/400/)

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/format.py
+++ b/src/middlewared/middlewared/plugins/disk_/format.py
@@ -46,6 +46,10 @@ class DiskService(Service):
         part.name = 'data'  # give a human readable name to the label
         parted_disk.addPartition(part, constraint=dev.optimalAlignedConstraint)
         parted_disk.commit()
+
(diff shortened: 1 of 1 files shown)


anodos325 (approved):

yocalebo:
The reason why I removed this was because that function settles _ALL_ udev events at the time it's called. This means it can block up to 120 seconds (default timeout).

bmeagherix:
After discussion merged for now, but a better fix may be forthcoming.

PR #13741, opened 2024-05-16
NAS-129038 / 24.10 / Add synchronous version of standard path validator

anodos325:
Convert `check_path_resides_within_volume` to a wrapper around a synchronous validator `check_path_resides_within_volume_sync` since basically all of logic there relies on stat-related calls.

### DIFF
--- a/src/middlewared/middlewared/async_validators.py
+++ b/src/middlewared/middlewared/async_validators.py
@@ -1,10 +1,6 @@
 import asyncio
-import os
 import socket
-from pathlib import Path
 
-from middlewared.plugins.zfs_.utils import ZFSCTL
-from middlewared.utils.path import path_location
 from middlewared.validators import IpAddress, check_path_resides_within_volume_sync
 
 
--- a/src/middlewared/middlewared/async_validators.py
+++ b/src/middlewared/middlewared/async_validators.py
@@ -5,66 +5,15 @@ from pathlib import Path
 
 from middlewared.plugins.zfs_.utils import ZFSCTL
 from middlewared.utils.path import path_location
-from middlewared.validators import IpAddress
+from middlewared.validators import IpAddress, check_path_resides_within_volume_sync
 
 
 async def check_path_resides_within_volume(verrors, middleware, name, path):
-
-    # we need to make sure the sharing service is configured within the zpool
-    def get_file_info(path):
-        """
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/async_validators.py
```diff
@@ -1,70 +1,15 @@
...
 import socket
-from pathlib import Path
 
-from middlewared.plugins.zfs_.utils import ZFSCTL
-from middlewared.utils.path import path_location
-from middlewared.validators import IpAddress
+from middlewared.validators import IpAddress, check_path_resides_within_volume_sync
 
 
 async def check_path_resides_within_volume(verrors, middleware, name, path):
```
bmeagherix, line 7: Maybe rename `name` to `schema_name`, just to be clearer.

src/middlewared/middlewared/validators.py
```diff
@@ -371,3 +375,63 @@ def __call__(self, value):
...
+
+    `vol_names` - list of expected pool names
+
+    It checks the following:
+    * path is within /mnt
+    * path is located within one of the specified `vol_names`
+    * path is not explicitly a `.zfs` or `.zfs/snapshot` directory
+    * path is not ix-applications dataset
+    """
+    if path_location(path) == 'EXTERNAL':
```
bmeagherix, line 399: `if path_location(path) is FSLocation.EXTERNAL:`
?

src/middlewared/middlewared/pytest/unit/test_validators.py
```diff
@@ -37,3 +40,32 @@ def test__email_schema(email, should_raise):
...
+        shutil.rmtree('/mnt/pool')
+
+
+@pytest.mark.parametrize("path,should_raise", [
+    ('/tmp', True),
+    ('EXTERNAL://smb_server.local/SHARE', True),
+    ('/mnt/does_not_exist', True),
+    ('/mnt/pool/foo', False),
+    ('/mnt/pool', False),
+    ('/mnt/pool//symlink', True)
```
bmeagherix, line 61: Are the double slash intended here? (`//`)  Suspect not.

Also maybe add another test where the symlink /tmpdir and points under /mnt/foo

bmeagherix (changes requested):

bmeagherix (approved):

PR #13743, opened 2024-05-16
NAS-129043 / 24.10 / HTOP Default Configuration

kmoore134:
Add a nicer set of out of box defaults for htop which shows more information to the user about:

* ZFS ARC
* ZFS Compressed ARC
* Network
* Disk
* CPU Freq
* CPU Temp
* Kernel Version


kmoore134:
Example Default View:

image

src/freenas/etc/htoprc
```diff
@@ -0,0 +1,13 @@
+# Beware! This file is rewritten by htop when settings are changed in the interface.
+# The parser is also very primitive, and not human-friendly.
+htop_version=3.2.2
```
anodos325, line 3: We can probably drop this line

anodos325 (approved):

kmoore134:
backport

PR #13744, opened 2024-05-16
NAS-129045 / 24.10 / Improve disk.get_partitions_quick

yocalebo:
As pointed out in recent PR: https://github.com/truenas/middleware/pull/13738, `device.settle_udev_events` is needed (sporadically) after partitioning disks. The reason why is that certain operations will generate udev events. We try to access the information before those events have been processed and so sysfs information will not be complete.

The problem by calling `settle_udev_events` is that it will tell udev to drain the entire event queue. The queue holds information for ALL hardware and related events. This means, depending on what the system is doing at the time this is called, it can block and wait for 120 seconds (default timeout).

This PR changes it so that `disk.get_partitions_quick` actually tries 10 times (maximum) for the sysfs files to be generated. While I'm here, I noticed that if a disk had > 9 partitions, this would not return correct information. To be fair, however, we don't have any logic to handle disks with that many partitions but I went ahead and fixed this while I was here.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/wipe.py
+++ b/src/middlewared/middlewared/plugins/disk_/wipe.py
@@ -41,7 +41,7 @@ class DiskService(Service):
                 # so we'll break out early
                 return startsect
             else:
-                time.sleep(0.2)
+                time.sleep(0.5)
 
             try:
                 sectsize = int((path_obj / 'queue/logical_block_size').read_text().strip())
--- a/src/middlewared/middlewared/plugins/disk_/wipe.py
+++ b/src/middlewared/middlewared/plugins/disk_/wipe.py
@@ -53,7 +53,7 @@ class DiskService(Service):
             except (FileNotFoundError, ValueError):
                 continue
             except Exception:
-                if _try == retries:
+                if _try == tries:
                     self.logger.error('Unexpected failure gathering partition info', exc_info=True)
 
         return startsect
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/disk_/wipe.py
```diff
@@ -16,25 +16,45 @@
...
+                sectsize = int((path_obj / 'queue/logical_block_size').read_text().strip())
+                with os.scandir(path_obj) as dir_contents:
+                    for partdir in filter(lambda x: x.is_dir() and x.name.startswith(dev_name), dir_contents):
+                        part_num = int((partdir / 'partition').read_text().strip())
+                        part_start = int((partdir / 'start').read_text().strip()) * sectsize
+                        startsect[part_num] = part_start
+            except (FileNotFoundError, ValueError):
+                continue
+            except Exception:
+                if _try == tries:
```
bmeagherix, line 56: Shouldn't this be
`if _try == tries - 1:`
to catch the last time round the loop?

bmeagherix (changes requested):

src/middlewared/middlewared/plugins/disk_/wipe.py
yocalebo: `range()` being half-open is always fun :smile: (It should be `tries + 1`)

bmeagherix: You say "potato".

yocalebo: Fixed it nonetheless, thanks!

bmeagherix (approved):

PR #13756, opened 2024-05-19
Fix truecommand issues on HA in CORE 13

sonicaj:
This PR adds changes to backport some of the fixes which were made to SCALE and not backported to CORE and also fix an issue where wireguard service started on standby whenever the service was started on active. Reason behind that was that we propagate service actions to standby automatically by default and that is not desirable in this case as wireguard should only be running on 1 node at a time. To address that `truecommand` has been added to list of blacklisted services which shouldn't be touched when any of the service verb is called for it.

An edge case for nginx config has also been handled where we were not adding wireguard interface ip to listen directive on failover and a subsequent nginx config reload was warranted because of that.


sonicaj:
backport

yocalebo (approved):

PR #13758, opened 2024-05-20
NAS-128634 / 24.10 / Show IPv6 autoconf default, also reflect preference with sysctl

aiden3d:
When creating a network interface, the system automatically configures IPv6 despite the default setting being False. In the UI, this creates an issue where the "Autoconfigure IPv6" checkbox is unchecked but the value under /proc/sys/net/ipv6/conf/[interface]/autoconf is 1.

This is also true of DHCP, but was fixed in the UI by setting the value to True specifically if we haven't created/configured the interface yet. I applied the same fix to ipv6_auto. I've also made it set the proper /proc/.../autoconf value when an existing interface is updated so the user's settings are respected.

### DIFF
--- a/tests/api2/test_005_interface.py
+++ b/tests/api2/test_005_interface.py
@@ -8,7 +8,7 @@ import pytest
 
 from auto_config import interface, ha, netmask
 from middlewared.test.integration.utils.client import client, truenas_server
-from middlewared.test.integration.utils import fail, call
+from middlewared.test.integration.utils import call
 
 @pytest.fixture(scope='module')
 def ws_client():
@@ -65,7 +65,7 @@ def get_payload(ws_client):
 def test_001_check_ipvx(request):
     # Verify that dhclient is running
     running, _ = call('interface.dhclient_status', interface)
-    assert running
+    assert running is True
 
     # Check that our proc entry is set to its default 1.
     assert int(call('tunable.get_sysctl', f'net.ipv6.conf.{interface}.autoconf')) == 1
--- a/tests/api2/test_005_interface.py
+++ b/tests/api2/test_005_interface.py
@@ -8,6 +8,7 @@ import pytest
 
 from auto_config import interface, ha, netmask
 from middlewared.test.integration.utils.client import client, truenas_server
+from middlewared.test.integration.utils import fail
 
 
 @pytest.fixture(scope='module')
@@ -21,6 +22,7 @@ def get_payload(ws_client):
     if ha:
         payload = {
             'ipv4_dhcp': False,
+            'ipv6_auto': False,
             'failover_critical': True,
             'failover_group': 1,
             'aliases': [
@@ -49,7 +51,7 @@ def get_payload(ws_client):
         # that the machine has been handed an IPv4 address
         # from a DHCP server. That's why we're getting this information.
         ans = ws_client.call('interface.query', [['name', '=', interface]], {'get': True})
-        payload = {'ipv4_dhcp': False, 'aliases': []}
+        payload = {'ipv4_dhcp': False, 'ipv6_auto': False, 'aliases': []}
         to_validate = []
         ip = truenas_server.ip
         for info in filter(lambda x: x['address'] == ip, ans['state']['aliases']):
@@ -60,8 +62,17 @@ def get_payload(ws_client):
 
     return payload, to_validate
 
+# Make sure that our initial conditions are met
+def test_001_check_ipvx(request, ws_client, get_payload):
+    # Verify that dhclient is running
+    ps_count = int(SSH_TEST('ps -aux | grep dhclient | wc -l', user, password, truenas_server.ip)['stdout'])
+    assert ps_count > 1 # account for the grep
 
-def test_001_configure_interface(request, ws_client, get_payload):
+    # Check that our proc entry is set to its default 1.
+    autoconf = int(SSH_TEST(f'cat /proc/sys/net/ipv6/conf/{interface}/autoconf', user, password, truenas_server.ip)['stdout'])
+    assert autoconf == 1
+
+def test_002_configure_interface(request, ws_client, get_payload):
     if ha:
         # can not make network changes on an HA system unless failover has
         # been explicitly disabled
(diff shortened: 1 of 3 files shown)


aiden3d:
Here's a link to my [passing test](http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/459). Look for `check_ipvx`!

tests/api2/test_system_dataset.py
```diff
@@ -67,7 +67,7 @@ def test_lock_passphrase_encrypted_pool_with_system_dataset(passphrase_encrypted
 def test_system_dataset_mountpoints():
     system_config = call("systemdataset.config")
     for system_dataset_spec in call(
-        "systemdataset.get_system_dataset_spec", system_config["pool"], system_config["uuid_a"]
+        "systemdataset.get_system_dataset_spec", system_config["pool"], system_config["uuid"]
```
sonicaj, line 70: I think this is here because the changes were cherry-picked ?
We can workaround this by doing `git rebase master` in your branch if you would like to update your version of the branch from master

tests/api2/test_005_interface.py
```diff
@@ -60,8 +63,17 @@ def get_payload(ws_client):
 
     return payload, to_validate
 
+# Make sure that our initial conditions are met
+def test_001_check_ipvx(request, ws_client, get_payload):
+    # Verify that dhclient is running
+    ps_count = int(SSH_TEST('ps -aux | grep dhclient | wc -l', user, password, truenas_server.ip)['stdout'])
```
sonicaj, line 69: I wonder if it might be better to query interface API and see if we have dhcp activated for any interface there ?

tests/api2/test_005_interface.py
```diff
@@ -60,8 +63,17 @@ def get_payload(ws_client):
 
     return payload, to_validate
 
+# Make sure that our initial conditions are met
+def test_001_check_ipvx(request, ws_client, get_payload):
```
sonicaj, line 67: Do we want `get_payload/ws_client` here ?

sonicaj: @yocalebo i think using `ws_client` should not be required anymore as we have refined `call` function now which we consume for communicating - do you agree ?

tests/api2/test_005_interface.py
```diff
@@ -60,8 +63,17 @@ def get_payload(ws_client):
...
+    # Verify that dhclient is running
+    ps_count = int(SSH_TEST('ps -aux | grep dhclient | wc -l', user, password, truenas_server.ip)['stdout'])
+    assert ps_count > 1 # account for the grep
 
-def test_001_configure_interface(request, ws_client, get_payload):
+    # Check that our proc entry is set to its default 1.
+    autoconf = int(SSH_TEST(f'cat /proc/sys/net/ipv6/conf/{interface}/autoconf', user, password, truenas_server.ip)['stdout'])
```
sonicaj, line 73: I don't think there is any need to use `SSH_TEST` here, we can get the same effect with
```
midclt call tunable.get_sysctl 'net.ipv6.conf.br0.autoconf'
```

tests/api2/test_005_interface.py
```diff
@@ -6,8 +6,10 @@
 
 import pytest
 
-from auto_config import interface, ha, netmask
+from auto_config import interface, ha, netmask, user, password
 from middlewared.test.integration.utils.client import client, truenas_server
+from middlewared.test.integration.utils import fail
+from functions import SSH_TEST
```
anodos325, line 12: preference is to use `ssh()` asset from `middlewared.test.integration.utils`

tests/api2/test_005_interface.py
```diff
@@ -6,8 +6,10 @@
 
 import pytest
 
-from auto_config import interface, ha, netmask
+from auto_config import interface, ha, netmask, user, password
```
anodos325, line 9: If you use `ssh` asset, then you don't need to import `user` and `password`

tests/api2/test_system_dataset.py
aiden3d: Yeah that was totally my bad, shouldn't happen in the future!

tests/api2/test_005_interface.py
yocalebo: We could make the test a bit more robust by adding both checks. Ensure our API reflects that dhcp is started (using interface API) but also checking that dhclient is actually running is worthwhile as well. We should do both to cover all our bases.

tests/api2/test_system_dataset.py
sonicaj: So if we run
```
git rebase --onto 3d4112a3ad475a55bbb5926a134fc0b4281ba526 b86ab7685456b234a7ecc1ec85a2e4e8572b4e60
```
And then force push to this branch, that will effectively remove this commit from your branch and we should be clean on this end ;) 
(the other way would be to do `git rebase -i HEAD~9` and then drop the problematic commit)

tests/api2/test_005_interface.py
```diff
@@ -8,7 +8,7 @@
 
 from auto_config import interface, ha, netmask
 from middlewared.test.integration.utils.client import client, truenas_server
-
+from middlewared.test.integration.utils import fail, call
```
sonicaj, line 11: `fail` is not being used anywhere ?

tests/api2/test_005_interface.py
```diff
@@ -60,8 +61,16 @@ def get_payload(ws_client):
 
     return payload, to_validate
 
+# Make sure that our initial conditions are met
+def test_001_check_ipvx(request):
+    # Verify that dhclient is running
+    running, _ = call('interface.dhclient_status', interface)
+    assert running
```
sonicaj, line 68: Also better to be explicit about the assert i.e `assert running is True` (assuming `running` should be a boolean)

sonicaj (commented):
Just another couple of minor nits

sonicaj (approved):
Thanks!

aiden3d:
Of course! Thank you a ton for your suggestions.

PR #13761, opened 2024-05-20
NAS-129118 / 24.10 / Fix log spam during AD leave

anodos325:
Remove some legacy behavior from before the secrets plugin was refactored to use common APIs for tdb plugin. When we leave AD our secrets entries are automatically deleted from secrets.tdb. Simply performing backup will also force them to be removed from our config file as well.


bmeagherix (approved):
Approving, but

> When we leave AD our secrets entries are automatically deleted from secrets.tdb

Do we have a test to that effect, or can we add one?

PR #13770, opened 2024-05-21
NAS-129065 / 24.10 / Handle mDNS SRV record for NUT service

bmeagherix:
Use our own mako infrastructure to handle the mDNS SRV record for the NUT service.  This will override the file installed by the upstream debian package, and will only be present when UPS is configured/running.

Also include make several small fixes to allow unit tests pass.

### DIFF
--- a/src/middlewared/middlewared/plugins/service_/services/ups.py
+++ b/src/middlewared/middlewared/plugins/service_/services/ups.py
@@ -21,6 +21,10 @@ class UPSService(SimpleService):
             await self._systemd_unit("nut-driver-enumerator", "start")
         await self._unit_action("Start")
 
+    async def after_start(self):
+        # Reconfigure mdns (add nut service)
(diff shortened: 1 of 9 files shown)


src/middlewared/middlewared/plugins/service_/services/ups.py
```diff
@@ -21,6 +21,10 @@ async def start(self):
             await self._systemd_unit("nut-driver-enumerator", "start")
         await self._unit_action("Start")
 
+    async def after_start(self):
+        # Reconfigure mdns (add nut service)
+        await self.middleware.call('service.reload', 'mdns')
```
yocalebo, line 26: `service.*` related endpoints propagate the service action to the other controller on HA systems by default. Do we need to do that with this service?

bmeagherix: No

anodos325 (approved):

PR #13771, opened 2024-05-21
NAS-128480 / 24.10 / Make changes in netdata nut plugin to support SLAVE UPS

Qubad786:
## Problem
There were two issues with the existing netdata nut plugin:
1. It doesn't support UPS SLAVE mode because it requires a remote address.
2. When UPS is not configured, the failed `upsc -l` command returns an error, which keeps spamming the netdata logs and causes the log file size to increase continuously.

## Solution
1. Store the UPS configuration in netdata's storage dataset in JSON format, including information related to UPS mode and its address. The Netdata plugin will use this info to run the upsc command accordingly.
2. Also, store information about whether the service has started or not. Modify the netdata plugin to avoid executing the `upsc` command when the UPS configuration is not stored or the UPS service has not started successfully.

### DIFF
--- a/src/freenas/usr/lib/netdata/charts.d/nut_ups.chart.sh
+++ b/src/freenas/usr/lib/netdata/charts.d/nut_ups.chart.sh
@@ -1,8 +1,51 @@
 source /usr/lib/netdata/charts.d/nut.chart.sh
 
-nut_get_all() {
-  run -t $nut_timeout upsc -l || echo "ix-dummy-ups"
+
+ups_config=''
+
+get_ups_config(){
+  if [ -e /var/db/system/netdata/ups_info.json ]; then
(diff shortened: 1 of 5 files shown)


sonicaj (approved):

theofficialgman:
@Qubad786 when will this fix be included? this is still an issue in the latest release

https://ixsystems.atlassian.net/browse/NAS-132924
https://forums.truenas.com/t/scale-remote-ups-data-not-showing-on-reports/11521/7

mmattel:
I can back this request. Still an issue with 25.04.2.6 configuring UPS in slave mode.
No UPS reporting and spamming the log in 1s steps:
```
sudo tail /var/log/netdata/error.log | grep ups
2025-11-01 20:23:04: charts.d: : nut_ups: command 'upsc -l ' failed with code 1:
```

PR #13772, opened 2024-05-21
NAS-129152 / 24.10 / Fix /data/license perms

bmeagherix:
Noticed that a CI test (`test_007_early_settings.test_data_dir_perms`, #13707) was failing because of the mode of _/data/license_ in a HA system.

Fix that particular issue.  (Other issues remain.)

### DIFF
--- a/src/middlewared/middlewared/plugins/system/product.py
+++ b/src/middlewared/middlewared/plugins/system/product.py
@@ -16,6 +16,7 @@ from middlewared.utils.license import LICENSE_ADDHW_MAPPING
 
 
 LICENSE_FILE = '/data/license'
+LICENSE_FILE_MODE = 0o600
 
 
 class SystemService(Service):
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/system/product.py
```diff
@@ -209,7 +210,7 @@ def license_update(self, license_):
                     raise ValidationError('system.license', 'This is not an HA capable system.')
 
         prev_product_type = self.middleware.call_sync('system.product_type')
-        with open(LICENSE_FILE, 'w+') as f:
+        with open(os.open(LICENSE_FILE, os.O_CREAT | os.O_WRONLY | os.O_TRUNC, mode=LICENSE_FILE_MODE), 'w+') as f:
```
yocalebo, line 213: Should we just do os.fchown on f.fileno()?

yocalebo: And by os.fchown....I mean os.fchmod 😐

yocalebo (approved):

PR #13776, opened 2024-05-22
NAS-128941 / 24.10 / Allow VM deletion if virtualization is not supported

Qubad786:
### Context

A user switched from virtualization supported TN system to virtualization not supported and encountered this issue where he couldn't query and  then remove the VMs from the TN system. 

### Fix

PR adds the change to enable deletion of VMs even if virtualization is not supported. After the changes vm query will show the existing VMs but in error state like below. 

`"status": {
      "state": "ERROR",
      "pid": null,
      "domain_state": "ERROR"
    }`

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/vms.py
+++ b/src/middlewared/middlewared/plugins/vm/vms.py
@@ -15,7 +15,7 @@ from middlewared.service import CallError, CRUDService, item_method, job, privat
 from middlewared.validators import Range, UUID
 from middlewared.plugins.vm.numeric_set import parse_numeric_set, NumericSet
 
-from .utils import ACTIVE_STATES
+from .utils import ACTIVE_STATES, get_default_status
 from .vm_supervisor import VMSupervisorMixin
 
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/vm/vms.py
```diff
@@ -99,10 +99,10 @@ def bootloader_ovmf_choices(self):
     @private
     def extend_context(self, rows, extra):
         status = {}
-        if rows:
+        if rows and self._is_kvm_supported():
```
yocalebo, line 102: `self._is_kvm_supported()` is called potentially N times + 1 (N being the number of `rows`). Please call this once and store it in a variable.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/vm/vms.py
Qubad786: @yocalebo the implementation of this function already has a cached variable so no extra processing is done..should i still save it as a var before?

yocalebo: Yes, because it's not immediately obvious that the function returns a cached variable.

yocalebo: I also have no idea how that cache value is updated so you could, in theory, have a condition where you call it once and it returns true and then the very next time you call it, it returns false.

Qubad786: I am not sure how that would be possible?
I mean we check if kvm is supported on the machine - that won't change on a running machine?

I'll make changes to use the variable but clarifying your last comment

gcs8:
FYI, this also happens in the datasets view, iirc after unlocking the encrypted datasets.

yocalebo (approved):

Qubad786:
backport

PR #13780, opened 2024-05-22
NAS-128368 / 24.10 / nvme-of: Remove/Add disk on discovery change event

ixhamza:
Jira Ticket: [NAS-128368](https://ixsystems.atlassian.net/browse/NAS-128368)

[NAS-128368](https://ixsystems.atlassian.net/browse/NAS-128368) reports three different issues related to NVMe-oF hotplugging:

### Problem:
NAS-128368 reports three different issues related to NVMe-oF hotplugging:
1. When a drive is physically removed while part of the pool, the Linux/zfs still sees it as ONLINE.
2. A drive is not visible to the OS upon insertion until the next drive is inserted, i.e., drive addition by OS is lagged by one event. For example, if a drive is inserted into Slot-1, it won't be added by Linux until another drive is inserted into a different slot. However, the addition of the next drive will then be lagged by the drive inserted after that.
3. If fenced is holding a drive, the new drive still points to the previous NVMe subsystem, showing the previous serial number of the last drive inserted.

### Solution:
1. The discovery controller receives the udev event, and by default, the NVMe-oF service runs the connect-all command, which is insufficient to remove the drive. To address this, we check the drives that are part of the enclosure and disconnect those not matched with the discovery log page.
2. It takes up to 3 seconds after receiving the discovery event for the drives to be ready for connection on the client side. To accommodate this, we added a timeout of five seconds to ensure the drive connects properly.
3. This issue is fixed by a Linux kernel patch: https://github.com/truenas/linux/pull/178. We now create a new subsystem if the previous one does not have any valid controllers.

Additionally, by default, systemd does not queue the events and ignores the service start if the service is already running. To resolve this, we added a simple queue to handle at most one event and created a wrapper to unblock systemd for the next udev events.

[NAS-128368]: https://ixsystems.atlassian.net/browse/NAS-128368?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NAS-128368]: https://ixsystems.atlassian.net/browse/NAS-128368?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/freenas/etc/systemd/system/nvmf-connect@.service.d/override.conf
+++ b/src/freenas/etc/systemd/system/nvmf-connect@.service.d/override.conf
@@ -1,0 +1,3 @@
+[Service]
+ExecStart=
+ExecStart=/bin/sh -c "/usr/local/bin/nvmf-wrapper.sh `/bin/echo -e '${CONNECT_ARGS}'`"
--- a/src/freenas/usr/local/bin/nvmf-connect.sh
+++ b/src/freenas/usr/local/bin/nvmf-connect.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+
+RECONNECT_TIMEOUT_SEC=5
+discovery_dev=$(echo "$@" | grep -oP '(?<=--device=)\w+')
+discovery_addr=$(</sys/class/nvme/${discovery_dev}/address)
+queue="/var/run/${discovery_dev}-fabric-queue"
+
+while [ -s $queue ]; do
+  truncate -s 0 $queue
+  local_nqns=""
+  disconnected=0
+  discovery_nqns=$(nvme discover $@ | grep subnqn: | awk '{gsub(/^subnqn: */,"")}1')
+
+  for nqn_f in /sys/class/nvme-fabrics/ctl/nvme*/subsysnqn; do
+    dev=$(basename "$(dirname "$nqn_f")")
+    nqn=$(<"$nqn_f")
+    addr=$(</sys/class/nvme/${dev}/address)
+
+    if [[ "$dev" == "$discovery_dev" || "$addr" != "$discovery_addr" ]]; then
+      continue
+    fi
+
+    if [[ "$discovery_nqns" != *"$nqn"* ]]; then
+      nvme disconnect -d $dev
+      disconnected=1
+    else
+      local_nqns+="$nqn"$'\n'
+    fi
+  done
+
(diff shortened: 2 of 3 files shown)


yocalebo (changes requested):
Please explain what this does. The ticket explains what the issue is but this PR has no context on how it fixes the problem.

ixhamza:
@yocalebo - I have added a detailed explanation.

src/freenas/usr/local/bin/nvmf-connect.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+RECONNECT_TIMEOUT_SEC=5
+discovery_dev=$(echo "$@" | grep -oP '(?<=--device=)\w+')
+discovery_addr=$(</sys/class/nvme/${discovery_dev}/address)
+queue="/var/tmp/${discovery_dev}-fabric-queue"
```
anodos325, line 6: Maybe use a path under `/var/run`. `/var/run/middleware` might be an option. Sometimes files can get cleaned out of /tmp at unfortunate times.

src/freenas/etc/systemd/system/nvmf-connect@.service.d/override.conf
```diff
@@ -0,0 +1,3 @@
+[Service]
+ExecStart=
```
bmeagherix, line 2: Duplicate (blank) `ExecStart` line?

ixhamza: @bmeagherix - It's to clear the previous entry.

src/freenas/usr/local/bin/nvmf-connect.sh
ixhamza: @anodos325 - I can change it as you suggested but `/var/tmp` is not mounted on `tmpfs` like `/tmp`.

usaleem-ix (approved):

ixhamza:
@yocalebo - would be appreciated if you could review this, please.

yocalebo (approved):

PR #13787, opened 2024-05-23
NAS-129188 / 24.10 / Refactored dhparam generation

aiden3c:
The previous implementation of generating dhparam.pem did unnecessary things like print to console (this was purely to give visual feedback on Core) and make middleware calls to itself for a string path.

### DIFF
--- a/src/middlewared/middlewared/plugins/crypto_/dhparams.py
+++ b/src/middlewared/middlewared/plugins/crypto_/dhparams.py
@@ -21,8 +21,7 @@ class CertificateService(Service):
     def dhparam_setup(self, job):
         """Generate dhparam.pem if it doesn't exist, or has no data in it"""
         with open(DHPARAM_PEM_PATH, 'a+') as f:
-            if os.fstat(f.fileno()).st_size > 0:
-                return
-            subprocess.run(
-                ['openssl', 'dhparam', '-rand', '/dev/urandom', '2048'], stdout=f, check=True
-            )
+            if os.fstat(f.fileno()).st_size == 0:
+                subprocess.run(
+                    ['openssl', 'dhparam', '-rand', '/dev/urandom', '2048'], stdout=f, check=True
+                )
--- a/src/middlewared/middlewared/plugins/crypto_/dhparams.py
+++ b/src/middlewared/middlewared/plugins/crypto_/dhparams.py
@@ -21,7 +21,7 @@ class CertificateService(Service):
     def dhparam_setup(self, job):
         """Generate dhparam.pem if it doesn't exist, or has no data in it"""
         with open(DHPARAM_PEM_PATH, 'a+') as f:
-            if os.stat(DHPARAM_PEM_PATH).st_size > 0:
+            if os.fstat(f.fileno()).st_size > 0:
                 return
             subprocess.run(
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/crypto_/dhparams.py
```diff
@@ -11,15 +13,16 @@ class Config:
...
-        dhparam_path = self.middleware.call_sync('certificate.dhparam')
-        if not os.path.exists(dhparam_path) or os.stat(dhparam_path).st_size == 0:
-            with open('/dev/console', 'wb') as console:
-                with open(dhparam_path, 'wb') as f:
-                    subprocess.run(
-                        ['openssl', 'dhparam', '-rand', '/dev/urandom', '2048'], stdout=f, stderr=console, check=True
-                    )
+        """Generate dhparam.pem if it doesn't exist, or has no data in it"""
+        with open(DHPARAM_PEM_PATH, 'a+') as f:
+            if os.fstat(f.fileno()).st_size > 0:
```
anodos325, line 24: I think it would be clearer to subprocess if `st_size == 0`

src/middlewared/middlewared/plugins/crypto_/dhparams.py
```diff
@@ -11,15 +13,15 @@ class Config:
...
-        dhparam_path = self.middleware.call_sync('certificate.dhparam')
-        if not os.path.exists(dhparam_path) or os.stat(dhparam_path).st_size == 0:
-            with open('/dev/console', 'wb') as console:
-                with open(dhparam_path, 'wb') as f:
-                    subprocess.run(
-                        ['openssl', 'dhparam', '-rand', '/dev/urandom', '2048'], stdout=f, stderr=console, check=True
-                    )
+        """Generate dhparam.pem if it doesn't exist, or has no data in it"""
+        with open(DHPARAM_PEM_PATH, 'a+') as f:
+            if os.fstat(f.fileno()).st_size == 0:
```
anodos325, line 24: Should we validate / set appropriate permissions on file?

Qubad786: Yes please..i am already pushing a PR which fixes perms stuff in `/data` and this is one of those files. It should have `0o600`

aiden3c: I didn't do perm validation on my end because of the PR from @Qubad786 . Currently the perms are `0o644` in the tests, once that gets merged it should be the proper `0o600`!

anodos325: I believe they need to be looser than 0o600 to facility key exchange.

yocalebo: I was getting ready to say that we should probably validate whether or not 0600 is too restrictive

Qubad786: There aren't any issues when i use `0o600` and it being owned by root..so nginx master process which runs as root reads this which is why we don't have any problems. I have tested https and certs creation etc and we should be good ✅ 
ref: https://superuser.com/a/1250447

yocalebo (approved):

PR #13792, opened 2024-05-23
NAS-129054 / 24.10 / Fix permissions on files under data directory

Qubad786:
## Problem

The `test_data_dir_perms` of the early_settings integration test was failing because some files, including `user-services.json`, `truenas-eula-pending`, and `dhparam.pem`, do not have `0o600` permissions when created by middlewared.

## Solution

Specify the permissions of these files to be `0o600` when they are created by middlewared.

P.S permission for dhparam are not fixed as they should be handled in another PR we have open

### DIFF
--- a/src/middlewared/middlewared/etc_files/systemd.py
+++ b/src/middlewared/middlewared/etc_files/systemd.py
@@ -1,5 +1,6 @@
 import json
 import re
+import os
 
 from middlewared.service import CallError
 from middlewared.utils import run
@@ -38,4 +39,5 @@ async def render(service, middleware):
 
     # Write out a user enabled services to json file which shows which services user has enabled/disabled
     with open('/data/user-services.json', 'w') as f:
+        os.fchmod(f.fileno(), 0o600)
         f.write(json.dumps(services_enabled))
--- a/src/middlewared/middlewared/plugins/system/product.py
+++ b/src/middlewared/middlewared/plugins/system/product.py
@@ -9,6 +9,7 @@ from datetime import date
 from licenselib.license import ContractType, Features, License
 from pathlib import Path
 
+from middlewared.plugins.truenas import EULA_PENDING_PATH
 from middlewared.schema import accepts, Bool, returns, Str
 from middlewared.service import no_auth_required, no_authz_required, private, Service, ValidationError
 from middlewared.utils import sw_info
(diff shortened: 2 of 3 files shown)


src/middlewared/middlewared/plugins/truenas.py
```diff
@@ -114,6 +114,7 @@ async def accept_eula(self):
     async def unaccept_eula(self):
         with open(EULA_PENDING_PATH, "w"):
             pass
+        os.chmod(EULA_PENDING_PATH, 0o600)
```
anodos325, line 117: I don't think this should be an async coroutine since it's only doing file IO.

Qubad786: Some other methods were async as well in this file - i fixed those too ✅

src/middlewared/middlewared/etc_files/systemd.py
```diff
@@ -39,3 +40,4 @@ async def render(service, middleware):
     # Write out a user enabled services to json file which shows which services user has enabled/disabled
     with open('/data/user-services.json', 'w') as f:
         f.write(json.dumps(services_enabled))
+    os.chmod('/data/user-services.json', 0o600)
```
yocalebo, line 43: os.fchmod(f.fileno(), ...

src/middlewared/middlewared/plugins/truenas.py
```diff
@@ -111,9 +111,10 @@ async def accept_eula(self):
                 raise
 
     @private
-    async def unaccept_eula(self):
+    def unaccept_eula(self):
         with open(EULA_PENDING_PATH, "w"):
             pass
+        os.chmod(EULA_PENDING_PATH, 0o600)
```
yocalebo, line 117: os.fchmod(f.fileno(), ...

yocalebo (changes requested):

src/middlewared/middlewared/plugins/system/product.py
```diff
@@ -218,6 +218,8 @@ def license_update(self, license_):
         SystemService.PRODUCT_TYPE = None
         if self.middleware.call_sync('system.is_enterprise'):
             Path('/data/truenas-eula-pending').touch(exist_ok=True)
```
yocalebo, line 220: Instead of using `Path` to create the file, we should just do `with open(..., 'a+') as f:` Finally, we need to use `EULA_PENDING_PATH` var in this file.

yocalebo (approved):

PR #13794, opened 2024-05-24
NAS-129193 / 24.10 / Make test_disk_wipe_abort HA aware

bmeagherix:
On HA systems `fenced` will have a reference to the disk, so it will never have no references.  Change the test to accommodate.

### DIFF
--- a/tests/api2/test_disk_wipe.py
+++ b/tests/api2/test_disk_wipe.py
@@ -95,7 +95,9 @@ def test_disk_wipe_abort():
     if ha:
         # In HA systems fenced may be using the disk.  Obtain the PID
         # so that we can ignore it.
-        expected_pids.add(ssh('cat /run/middleware/fenced.pid'))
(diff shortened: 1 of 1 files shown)


tests/api2/test_disk_wipe.py
```diff
@@ -88,6 +90,14 @@ def test_disk_get_partitions_quick(dev_name):
 
 
 def test_disk_wipe_abort():
+    """Test that we can sucessfully abort a disk.wipe job"""
+    expected_pids = set()
+    if ha:
+        # In HA systems fenced may be using the disk.  Obtain the PID
+        # so that we can ignore it.
+        expected_pids.add(ssh('cat /run/middleware/fenced.pid'))
```
yocalebo, line 98: I have a method for fenced to return the pid and whether or not it's running. failover.fenced.run_info

yocalebo (approved):

PR #13797, opened 2024-05-24
NAS-129041 / 24.10 / OOM score adjustment

aiden3c:
Setting our `oom_score_adj` to -1000 makes our middleware process the lowest priority when the kernel goes to kill processes due to low memory. 
This also set all worker processes' `oom_score_adj` to -1000 as well.


aiden3c:
Newest [passing test](http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/544/console)! Check for `oom_check`.

yocalebo (approved):

PR #13804, opened 2024-05-29
NAS-129260 / 24.10 / Replace shareuser with UserAssets.ShareUser01 in depends

bmeagherix:
During PR #13491 the following:
```
@pytest.mark.dependency(name="shareuser")
def test_27_creating_shareuser_to_test_sharing(request):
```
was replaced with
```
@pytest.mark.dependency(name=UserAssets.ShareUser01['depends_name'])
def test_020_create_and_verify_shareuser():
```

However, dependent changes were only made in _test_011_user.py_, e.g.
```
def test_038_change_homedir_to_existing_path(request):
    depends(request, [UserAssets.ShareUser01['depends_name'], UserAssets.TestUser01['depends_name']])
    ....
```
(Note that the actual string referred to by `depends_name` has changed to "share_user_01")

This PR updates a couple of **other** tests in different modules that also depend on "shareuser".  Also did some tidy/`flake8` fixes.

Before this PR they were being skipped.   (PR tested in CI run [550](http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/550/))

Question: worth backporting to 24.04.2 ?

### DIFF
--- a/tests/api2/test_810_delete_user.py
+++ b/tests/api2/test_810_delete_user.py
@@ -4,17 +4,14 @@
 # License: BSD
 # Location for tests into REST API of FreeNAS
 
-import pytest
-import sys
-import os
 from pytest_dependency import depends
-apifolder = os.getcwd()
-sys.path.append(apifolder)
-from functions import GET, DELETE
+from test_011_user import UserAssets
+
+from functions import DELETE, GET
 
 
 def test_01_deleting_user_shareuser(request):
-    depends(request, ["shareuser"], scope="session")
+    depends(request, [UserAssets.ShareUser01['depends_name']], scope="session")
     userid = GET('/user?username=shareuser').json()[0]['id']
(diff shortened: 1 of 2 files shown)


tests/api2/test_810_delete_user.py
yocalebo, line 1: This file is worthless. Should just be removed

bmeagherix: > This file is worthless. Should just be removed

Perhaps, but what inspired the PR is that _another_ test was complaining about left-over users (one of which was _shareuser_, because this test was being skipped).  So, meh.

anodos325: Maybe we should eliminate the share_user entirely and only use our existing user fixtures scoped to `module` or `function`.

yocalebo:
Handled in https://github.com/truenas/middleware/pull/13987

PR #13815, opened 2024-05-29
NAS-128426 / 13.3 / Generate ed25519 by default when creating ssh key pairs

sonicaj:
This commit adds changes to generate ed25519 keys by default when creating ssh key pairs.


aiden3c (approved):

anodos325:
Are you sure we actually need to do this? IIRC there was an issue with certain situations with outdated 3rd-party SSH libraries (e.g. paramiko), not RSA keys themselves.

sonicaj:
@anodos325 contrary to the ticket description, it was working for me however i think switching to ed25519 would be fine as well (https://linux-audit.com/using-ed25519-openssh-keys-instead-of-dsa-rsa-ecdsa/)

sonicaj:
Closing this after discussion with @anodos325

PR #13816, opened 2024-05-29
properly detect nvme devices in smart plugin

yocalebo:
This is required for the h-series platform. Disks sit behind a tri-mode HBA and so nvme devices appear to OS as `sd` devices.

### DIFF
--- a/src/middlewared/middlewared/common/smart/smartctl.py
+++ b/src/middlewared/middlewared/common/smart/smartctl.py
@@ -27,6 +27,9 @@ async def get_smartctl_args(context, disk, smartoptions):
     if device is None:
         return
 
+    if device["vendor"].lower().strip() == "nvme":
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/common/smart/smartctl.py
```diff
@@ -27,6 +27,9 @@ async def get_smartctl_args(context, disk, smartoptions):
     if device is None:
         return
 
+    if device["vendor"].lower().strip() == "nvme":
```
mgrimesix, line 30: Why not combine this with the nvme trap on line 23?

yocalebo: Because the `device` object is the dictionary with the `disk` details. The `disk` is just the name of it (`nvme`, `sd`, etc).

themylogin (approved):

mgrimesix (approved):

PR #13820, opened 2024-05-30
NAS-126750 / 24.10 / Add mismatched NICs to failover reasons

aiden3c:
In the `FailoverDisabledReasonsService`, we now check for NICs that exist on one node and not the other. 
In the scenario where node a has `get_nics`, but node b doesn't (such as during an update), we skip the check entirely.


src/middlewared/middlewared/plugins/failover_/disabled_reasons.py
```diff
@@ -160,6 +161,16 @@ def get_remote_reasons(self, app, ifaces, reasons):
             mismatch_disks = self.middleware.call_sync('failover.mismatch_disks')
             if mismatch_disks['missing_local'] or mismatch_disks['missing_remote']:
                 reasons.add(DisabledReasonsEnum.MISMATCH_DISKS.name)
+
+            local_nics = self.middleware.call_sync('interface.get_nics')
+            try:
+                remote_nics = self.middleware.call_sync('failover.call_remote', 'interface.get_nics')
```
bmeagherix, line 167: IIRC these will be converted to lists during the API calls, and I **don't** think we are _guaranteed_ ordering.

Might be safer to make them sets again locally before doing the comparison.  e.g.
```
local_nics = set(self.middleware.call_sync('interface.get_nics'))
```

bmeagherix (dismissed):

src/middlewared/middlewared/plugins/failover.py
```diff
@@ -435,6 +435,22 @@ async def get_disks_local(self):
         except Exception:
             self.logger.error('Unhandled exception in get_disks_local', exc_info=True)
 
+    @private
+    def mismatch_nics(self):
+        """Determine if NICs match between both controllers."""
+        result = False
+        try:
+            local_nics = self.middleware.call_sync('interface.get_nic_names')
+            remote_nics = self.middleware.call_sync('failover.call_remote', 'interface.get_nic_names')
```
yocalebo, line 444: This remote_nics call can block for awhile.

yocalebo, line 444: You need to review the `raise_connect_error` argument to `failover.call_remote` instead of catching CallError here.

src/middlewared/middlewared/plugins/failover.py
```diff
@@ -435,6 +435,22 @@ async def get_disks_local(self):
         except Exception:
             self.logger.error('Unhandled exception in get_disks_local', exc_info=True)
 
+    @private
+    def mismatch_nics(self):
+        """Determine if NICs match between both controllers."""
+        result = False
+        try:
+            local_nics = self.middleware.call_sync('interface.get_nic_names')
```
yocalebo, line 443: putting `local_nics =` call inside the `try: except` means that your error message "Unable to call interface.get_nic_names on remote" could be wrong.

src/middlewared/middlewared/plugins/network.py
```diff
@@ -15,6 +15,8 @@
 from .interface.interface_types import InterfaceType
 from .interface.lag_options import XmitHashChoices, LacpduRateChoices
 
+from os import scandir
```
yocalebo, line 18: Since this is a builtin module, we need to put the import at the top of the file above the `from .inter*` import lines.

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1936,6 +1938,16 @@ def ip_in_use(self, choices):
                         list_of_ip.append(alias_dict)
 
         return list_of_ip
+    
+    @private
+    def get_nic_names(self) -> set:
+        """
```
yocalebo, line 1944:
```suggestion
        """Get network interface names excluding internal interfaces"""
```

src/middlewared/middlewared/plugins/network.py
```diff
@@ -1936,6 +1938,16 @@ def ip_in_use(self, choices):
...
+        Get NICs without including internal_interfaces
+        """
+        with scandir("/sys/class/net/") as nics:
```
yocalebo, line 1947: Mixture of using double-quotes and single-quotes

src/middlewared/middlewared/plugins/failover.py
```diff
@@ -435,6 +435,22 @@ async def get_disks_local(self):
         except Exception:
             self.logger.error('Unhandled exception in get_disks_local', exc_info=True)
 
+    @private
+    def mismatch_nics(self):
```
yocalebo, line 439: We're going to have to add an alert for this new check. Furthermore, this method doesn't really match what `failover.mismatch_disks` does.

The mismatch_disks method actually tracks what's missing on the local node and what's missing on the remote node. We need to do the same here with this method.

Finally,  you're going to need to add an alert as well.

src/middlewared/middlewared/plugins/failover_/disabled_reasons.py
```diff
@@ -14,6 +14,7 @@ class DisabledReasonsEnum(str, Enum):
     NO_CRITICAL_INTERFACES = 'No network interfaces are marked critical for failover.'
     MISMATCH_DISKS = 'The quantity of disks do not match between the nodes.'
     MISMATCH_VERSIONS = 'TrueNAS software versions do not match between storage controllers.'
+    MISMATCH_NICS = 'NIC hardware does not match between nodes.'
```
yocalebo, line 17: Since this is public facing verbiage, we need to get with our documentation team on some guidance of what to use. For example, "NIC" isn't very proper since acronyms are ambiguous at best. @DjP-iX or @Mrt134 any suggestions on what we should use here?

I had the idea of using similar verbiage for other messages. i.e. `Network interfaces do not match between storage controllers.`

yocalebo (changes requested):

src/middlewared/middlewared/plugins/failover_/disabled_reasons.py
bmeagherix: Mea culpa.

yocalebo (approved):
great job!

PR #13823, opened 2024-05-31
NAS-129661 / 24.10 / Add basic FreeIPA connection control script

anodos325:
This commit adds a basic helper script to manage different aspects of a FreeIPA join such as:

* Joining a freeipa domain
* Leaving a freeipa domain
* Setting up services (NFS / SMB)
* deleting services (NFS / SMB)
* Querying SMB domain information

### DIFF
--- a/src/freenas/usr/local/libexec/ipa_ctl.py
+++ b/src/freenas/usr/local/libexec/ipa_ctl.py
@@ -1,0 +1,574 @@
+#!/usr/bin/python3
+
+# IPA control script for TrueNAS IPA Client. This provides support for some
+# basic IPA-related operations that the TrueNAS middleware performs.
+#
+# Although it is written as a standalone script, it is not intended for use
+# outside of the scope of TrueNAS developers. Ad-hoc usage of this script may
+# result in undefined sever behavior and is not supported in any way.
+
+import argparse
+import base64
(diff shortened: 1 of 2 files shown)


src/freenas/usr/local/libexec/ipa_ctl.py
```diff
@@ -0,0 +1,574 @@
...
+
+    def __get_errmsg(self, error_code_map):
+        if self.rpc_response:
+            return json.dumps(self.rpc_response)
+
+        if self.text:
+            return self.text
+
+        if error_code_map:
+            return error_code_map.get(self.error_code)
```
bmeagherix, line 125: Do we want to handle a `KeyError` or let crash?  (Methinks better to handle, even if we never expect it.)

anodos325: Since it's going through dictionary get, it will default to None if not present, Which I think is OK (though it shouldn't happen).

anodos325: We can fall through though and maybe give more useful error.

mgrimesix (approved):
Comments only, so I'll approve.

bmeagherix (approved):

PR #13826, opened 2024-06-02
NAS-128919 / 24.04.2 / Show only pods which are running when retrieving choices for pod console access

Qubad786:
## Problem

We were showing pods which were no longer running for console access choices which would obviously not work as the pod is no longer active.

## Solution

Make sure only those pods are shown as valid for pod console access choices which are actually running at that time.

### DIFF
--- a/src/middlewared/middlewared/plugins/chart_releases_linux/resources.py
+++ b/src/middlewared/middlewared/plugins/chart_releases_linux/resources.py
@@ -15,18 +15,23 @@ class ChartReleaseService(Service):
         namespace = 'chart.release'
 
     @private
-    async def retrieve_pod_with_containers(self, release_name):
+    async def retrieve_pod_with_containers(self, release_name, retrieve_active_pods=False):
         await self.middleware.call('kubernetes.validate_k8s_setup')
         release = await self.middleware.call(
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/chart_releases_linux/resources.py
```diff
@@ -15,18 +15,23 @@ class Config:
...
+    async def retrieve_pod_with_containers(self, release_name, retrieve_active_pods=False):
         await self.middleware.call('kubernetes.validate_k8s_setup')
         release = await self.middleware.call(
             'chart.release.query', [['id', '=', release_name]], {'get': True, 'extra': {'retrieve_resources': True}}
         )
         choices = {}
-        for pod in release['resources']['pods']:
-            choices[pod['metadata']['name']] = []
+        for pod in filter(
+            lambda p: retrieve_active_pods and p['status']['phase'] == 'Running' or not retrieve_active_pods,
```
sonicaj, line 25: K8s api can be inconsistent sometimes, are we certain these keys will always be there ?

sonicaj (approved):

yocalebo (approved):

PR #13836, opened 2024-06-04
NAS-129261 / 24.10 / Fix validation for httpproxy

Qubad786:
## Problem
Users are entering incorrect proxy URL formats in the network configuration.

## Solution
Add validation in the network global configuration to prevent users from adding incorrectly formatted proxy URLs.

Ref:
https://requests.readthedocs.io/en/latest/api/#requests.Session.proxies
https://docs.aiohttp.org/en/stable/client_advanced.html#proxy-support

### DIFF
--- a/src/middlewared/middlewared/plugins/network_/global_config.py
+++ b/src/middlewared/middlewared/plugins/network_/global_config.py
@@ -5,7 +5,7 @@ import signal
 
 import middlewared.sqlalchemy as sa
 from middlewared.service import ConfigService, private
-from middlewared.schema import accepts, Patch, List, Dict, Int, Str, Bool, IPAddr, Ref, ValidationErrors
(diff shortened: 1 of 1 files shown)


ramondsq:
As you say the URL formats are "incorrect", why would it work for the Apps(k3s)?

yocalebo (approved):

Qubad786:
@ramondsq that is because it is special in that sense and it can accept it in both formats (https://stackoverflow.com/questions/58841014/set-proxy-on-docker / https://docs.docker.com/network/proxy/#configure-the-docker-client).

ramondsq:
> @ramondsq that is because it is special in that sense and it can accept it in both formats (https://stackoverflow.com/questions/58841014/set-proxy-on-docker / https://docs.docker.com/network/proxy/#configure-the-docker-client).

Thank you for your answer

PR #13837, opened 2024-06-04
NAS-128090 / 24.10 / Track system dataset migration state for failover disabled reasons

Qubad786:
## Context

Changes have been added to track system dataset migration state in HA machines so we can correctly report that is in place when UI consumes `failover.disabled.reasons` to see if we have any reason to mark HA has disabled for now.

### DIFF
--- a/src/middlewared/middlewared/plugins/failover_/disabled_reasons.py
+++ b/src/middlewared/middlewared/plugins/failover_/disabled_reasons.py
@@ -29,6 +29,8 @@ class DisabledReasonsEnum(str, Enum):
     NO_CARRIER_ON_HEARTBEAT = 'Local heartbeat interface is down.'
     LOC_FIPS_REBOOT_REQ = 'This node needs to be rebooted to apply FIPS configuration'
     REM_FIPS_REBOOT_REQ = 'Other node needs to be rebooted to apply FIPS configuration'
+    LOC_SYSTEM_DATASET_MIGRATION_IN_PROGRESS = 'This node is currently configuring the system dataset'
+    REM_SYSTEM_DATASET_MIGRATION_IN_PROGRESS = 'Other node is currently configuring the system dataset'
 
 
 class FailoverDisabledReasonsService(Service):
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/failover_/disabled_reasons.py
```diff
@@ -165,6 +172,13 @@ def get_remote_reasons(self, app, ifaces, reasons):
             mismatch_nics = self.middleware.call_sync('failover.mismatch_nics')
             if mismatch_nics['missing_local'] or mismatch_nics['missing_remote']:
                 reasons.add(DisabledReasonsEnum.MISMATCH_NICS.name)
+
+            with contextlib.suppress(Exception):
+                # We do this safely because other node might not even have this method implemented
+                # if on different versions
+                if self.middleware.call_sync('failover.call_remote', 'failover.disabled.get_systemdataset_state'):
```
yocalebo, line 179: This can block for 60 seconds or more. Please review mismatch_disks method or mismatch_nics method for how to work around this. No need for contextlib either.

src/middlewared/middlewared/plugins/failover_/disabled_reasons.py
```diff
@@ -29,6 +30,8 @@ class DisabledReasonsEnum(str, Enum):
     NO_CARRIER_ON_HEARTBEAT = 'Local heartbeat interface is down.'
     LOC_FIPS_REBOOT_REQ = 'This node needs to be rebooted to apply FIPS configuration'
     REM_FIPS_REBOOT_REQ = 'Other node needs to be rebooted to apply FIPS configuration'
+    LOC_SYSTEM_DATASET_MIGRATION_IN_PROGRESS = 'This node is currently setting up system dataset'
```
yocalebo, line 33: "current configuring the system dataset"

yocalebo (changes requested):

yocalebo (approved):
Looks good! The UI also has a disabled reasons enum that matches what we have on our side. You'll need to update their side too or open a ticket with them to let them know you've added a new HA disabled reason.

PR #13838, opened 2024-06-04
fix long boot times when >10k snapshots

yocalebo:
We have had a total of 2 users complain about increase pool import times after upgrading from Cobia to Dragonfish. Fortunately, one of those users granted me remote access to their system and I was able to identify where the issue is.

Some notes:
1. both reporters have systems with > 10k snapshots (12k, and 89k respectively)
2. when `sharenfs` or `sharesmb` is set, something is iterating over every single snapshot of all child datasets on the pool

To make matters worse, the pool that I looked at was 5.4k RPM 6-wide RAIDz2 so waiting on spindle accesses play into the "speed" of this as well.

To recap, there are 2 issues:
1. there is a proper upstream zfs regression (this will be fixed (hopefully) in EE+)
2. the middleware can be a bit more intelligent by NOT updating properties unless the properties are set to values we don't expect

This PR does the latter.

Finally, I added a bit more logging. The logging _before_ these changes were _very_ misleading and didn't easily allow us to understand what exactly what taking so long at boot time. This should (in theory) remedy.

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/import_pool.py
+++ b/src/middlewared/middlewared/plugins/pool_/import_pool.py
@@ -287,40 +287,49 @@ class PoolService(Service):
         except Exception:
             self.logger.error('Unhandled exception importing %r', vol_name, exc_info=True)
             return False
+        else:
+            self.logger.debug('Done importing %r with guid %r', vol_name, vol_guid)
 
         # normalize ZFS dataset properties on boot. Pool may be foreign to SCALE
         # (including those created on CORE)
         try:
+            self.logger.debug('Calling zfs.dataset.query on %r with guid %r', vol_name, vol_guid)
             ds = self.middleware.call_sync(
-                'pool.dataset.query',
+                'zfs.dataset.query',
                 [['id', '=', vol_name]],
                 {'get': True, 'extra': {'retrieve_children': False}}
-            )
+            )['properties']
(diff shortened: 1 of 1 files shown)


aiden3c (approved):

src/middlewared/middlewared/plugins/pool_/import_pool.py
```diff
@@ -287,40 +287,49 @@ def import_on_boot_impl(self, vol_name, vol_guid, set_cachefile=False):
...
+            if ds['aclmode']['value'] != 'discard':
+                opts['properties'].update({'aclmode': {'value': 'discard'}})
 
-        opts['properties'].update({
-            'sharenfs': {'value': 'off'}, 'sharesmb': {'value': 'off'},
-        })
+        if ds['sharenfs']['value'] != 'off':
+            opts['properties'].update({'sharenfs': {'value': 'discard'}})
+        if ds['sharesmb']['value'] != 'off':
+            opts['properties'].update({'sharesmb': {'value': 'discard'}})
```
anodos325, line 323: Incorrect values

yocalebo: fixed and added a method that performs this operations since we were performing the same logic in 2 different methods.

anodos325 (approved):

PR #13840, opened 2024-06-04
NAS-129273 / 24.10 / fix enclosure2.query logic on R20/MINI3XP/MINI3E

yocalebo:
We were including the 2nd virtual AHCI enclosure object in `enclosure2.query` which was causing `webui.enclosure.dashboard` to crash with a `KeyError`.

This fixes that logic and prevents the key error.

### DIFF
--- a/src/middlewared/middlewared/plugins/enclosure_/enclosure_class.py
+++ b/src/middlewared/middlewared/plugins/enclosure_/enclosure_class.py
@@ -72,8 +72,11 @@ class Enclosure:
                 ControllerModels.MINI3E.value,
             )),
         )):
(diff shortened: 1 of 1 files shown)


bmeagherix (approved):
I'll approve, but it might be a good idea to extend the "These platforms" comment to summarize the platforms (in English) for clarity.

It looks like the olde code also had a bug because the `all` could **_never_** take place (can't be both r20 series and Mini3XP | Mini3E) -right?  If you disagree, then my review/approve may not be valid.

yocalebo:
Okay, I think I did what you asked.

PR #13841, opened 2024-06-04
NAS-129400 / 24.10 / Remove kubernetes implementation

sonicaj:
This PR adds changes removing kubernetes integration and adding basic integration for new catalog management.
(Docker is not configurable with these changes atm and neither are we creating it's state atm (mounting apps dataset changes will be reflected in the next PR as well), those changes will be introduced in another PR)


src/middlewared/middlewared/plugins/catalog/app_version.py
```diff
@@ -45,32 +45,31 @@ class Config:
...
         """
-        catalog = self.middleware.call_sync('catalog.get_instance', options['catalog'])
-        item_location = os.path.join(catalog['location'], options['train'], item_name)
-        if not os.path.exists(item_location):
-            raise CallError(f'Unable to locate {item_name!r} at {item_location!r}', errno=errno.ENOENT)
-        elif not os.path.isdir(item_location):
-            raise CallError(f'{item_location!r} must be a directory')
+        catalog = self.middleware.call_sync('catalog.config')
+        app_location = os.path.join(get_train_path(catalog['location']), options['train'], app_name)
+        if not os.path.exists(app_location):
```
yocalebo, line 54: We need to clean this up while you're here. os.path.exists and os.path.isdir stat the underlying object. We can reduce syscalls by donig this all at once.
```
try
    if not stat.S_ISDIR(os.stat(app_location).st_mode):
        raise CallError(f'{app_location!r} must be a directory')
except FileNotFoundError:
    raise CallError(f'Unable to locate {app_name!r} at {app_location!r}', errno=errno.ENOENT)

File: src/middlewared/middlewared/plugins/catalog/sync.py [hunk 2]
```diff
@@ -0,0 +1,65 @@
...
+    @private
+    def update_git_repository(self, catalog):
+        self.middleware.call_sync('network.general.will_perform_activity', 'catalog')
+        return pull_clone_repository(
+            catalog['repository'], os.path.dirname(catalog['location']), catalog['branch'],
+        )
+
+    @private
+    async def initiate_first_time_sync(self):
+        await (await self.middleware.call('catalog.sync')).wait()
```
yocalebo, line 64: What happens if this fails? We still set `self.SYNCED = True`?

src/middlewared/middlewared/plugins/catalog/update.py
```diff
@@ -0,0 +1,145 @@
...
+        return await self.config()
+
+    @private
+    async def update_train_for_enterprise(self):
+        catalog = await self.middleware.call('catalog.get_instance', OFFICIAL_LABEL)
+        if await self.middleware.call('system.product_type') == 'SCALE_ENTERPRISE':
+            preferred_trains = []
+            # Logic coming from here
+            # https://github.com/truenas/middleware/blob/e7f2b29b6ff8fadcc9fdd8d7f104cbbf5172fc5a/src/middlewared
+            # /middlewared/plugins/catalogs_linux/update.py#L341
```
yocalebo, line 123: This comment makes a reference to a path that is being removed in this PR?? `catalogs_linux` is going away or am I misunderstanding something?

src/middlewared/middlewared/plugins/docker/state_setup.py
```diff
@@ -0,0 +1,69 @@
...
+                    'extra': {
+                        'properties': list(update_props),
+                        'retrieve_children': False,
+                        'user_properties': False,
+                    }
+                }
+            )
+            if not dataset:
+                test_path = os.path.join('/mnt', dataset_name)
+                if await self.middleware.run_in_thread(os.path.exists, test_path):
```
yocalebo, line 52: You're creating 2 threads everytime this method is called to perform a `shutil.move` which is unnecessary.

This is prone to TOCTOU issues as well so you can simply do `shutil.move` and catch `FileNotFoundError` since it'll fail if `test_path` doesn't exist.

src/middlewared/middlewared/plugins/docker/state_setup.py
```diff
@@ -0,0 +1,69 @@
...
+            return
+
+        await self.pools(config)
+
+    @private
+    async def pools(self, config):
+        return await self.create_update_docker_datasets(applications_ds_name(config['pool']))
+
+    @private
+    async def create_update_docker_datasets(self, docker_ds):
```
yocalebo, line 32: Are we sure we need to do this? We're iterating in a for loop and running `zfs get -o blah` for each dataset.

src/middlewared/middlewared/plugins/docker/update.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        Update Docker service configuration.
+        """
+        raise CallError('Configuring docker is disabled for now')
+        old_config = await self.config()
+        old_config.pop('dataset')
+        config = old_config.copy()
+        config.update(data)
+
+        if len(set(old_config.items()) ^ set(config.items())) > 0:
```
yocalebo, line 55: Why is this check so complicated? It also seems wrong since it's calling "set" on a dictionaries keys but dictionaries keys ALWAYS have to be unique so the set is a NO-OP? I also don't know why we consider this different if the dictionaries keys don't match? I feel like this can be simplified to `if old_config != config`?

yocalebo (changes requested):

src/middlewared/middlewared/plugins/docker/state_setup.py
sonicaj: This code will not be used as we have changed how we want to handle actual datasets reality on the filesystem, so this will be addressed in my subsequent PR

src/middlewared/middlewared/plugins/catalog/update.py
sonicaj: The link will still remain active and can be accessed later on as well to see why we have this logic in place

yocalebo: Oh derp, I see the URL has the old git commit hash :smile:

src/middlewared/middlewared/plugins/docker/state_setup.py
yocalebo: Fantastic!

src/middlewared/middlewared/plugins/catalog/sync.py
sonicaj: Nope, but having synced flag below is redundant though

yocalebo (approved):

Qubad786 (approved):
LGTM!

PR #13850, opened 2024-06-06
NAS-128439 / 24.10 / Send interface events when commit/rollback for network is performed

Qubad786:
## Context

It was requested that we send events when commit/rollback for network is performed. Also it was requested that we provide not just the id of the interface in this case but the complete interface details so UI would not have to make a subsequent call.

## Proposed Solution

Requested cases have been addressed in following ways:

1. For added/removed/changed events, we try to send the complete payload in events if it is there for CRUD services
2. On commit/rollback of network changes, we isolate interfaces which were changed/removed/added and send their appropriate events in that case

### DIFF
--- a/src/middlewared/middlewared/plugins/network_/utils.py
+++ b/src/middlewared/middlewared/plugins/network_/utils.py
@@ -1,0 +1,69 @@
+import typing
+
+
+def get_interface_name(item: dict, key: str, interface_dict: dict) -> typing.Optional[str]:
+
+    def get_iface_name(iface_key: str, iface: dict) -> typing.Optional[str]:
+        iface_id = iface[iface_key]
+        if iface_id in interface_dict:
+            return interface_dict[iface_id]['int_interface']
+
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/network_/utils.py
```diff
@@ -0,0 +1,72 @@
...
+        return item['vlan_vint']
+    elif key == 'lagg':
+        return get_iface_name('lagg_interface', item)
+    elif key == 'laggmembers' and 'lagg_interface' in item:
+        return item['lagg_interface']
+    elif key == 'alias' and 'alias_interface' in item:
+        return get_iface_name('alias_interface', item)
+
+
+def find_interface_changes(original_datastores: dict, current_datastores: dict) -> typing.List[str]:
```
yocalebo, line 25: I'm sorry but no. This function is insanely complex. I don't want to make our network API any more complex than what it already is. This function is unmanageable, frankly. If this is the complexity that is required to send changed events, then we need to not send changed events.

yocalebo (changes requested):

PR #13857, opened 2024-06-07
NAS-129466 / 24.10 / Fix failing tests because of apps removal

Qubad786:
## Context

We had some leftover usages of `chart.release` namespace which have been covered here and were also causing some tests to fail.


src/middlewared/middlewared/plugins/pool_/dataset_details.py
```diff
@@ -325,19 +325,7 @@ def build_details(self, mntinfo):
...
-            for path in filter(
-                lambda x: x.startswith('/mnt/') and 'ix-applications/' not in x,
-                paths
-            ):
-                results['app'].append({
-                    'name': app_name,
-                    'path': path,
-                    'mount_info': self.get_mount_info(path, mntinfo),
-                })
+        # FIXME: Add app details here, currently it would be an empty list
```
yocalebo, line 328: Let's still return an "apps" key with an empty list so it doesn't break UI on the datasets page.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/pool_/dataset_details.py
Qubad786: We are doing that, it's already defined as empty above

Qubad786: @yocalebo here https://github.com/truenas/middleware/blob/fd8e3e12ff05adf5b4d7743696e9ee8a22bd8fa3/src/middlewared/middlewared/plugins/pool_/dataset_details.py#L271-L271

yocalebo (approved):
Thanks!

aiden3c (approved):

PR #13860, opened 2024-06-07
NAS-129384 / 24.10 / Fix netdata spamming logs

Qubad786:
## Problem
We were seeing logs being continuously written to, after investigation, we found that only the UPS plugin of netdata is causing issues. When UPS is not configured, the failed `upsc -l` command returns an error, spamming the Netdata logs and causing the log file size to increase continuously.

## Solution
Modify the Netdata plugin to avoid executing the `upsc` command when `upsmon` process is not running. Additionally, disable access logs since only the `errors.log` is used for debugging, thereby preventing unnecessary log file growth.

### DIFF
--- a/src/middlewared/middlewared/etc_files/netdata/netdata.conf.mako
+++ b/src/middlewared/middlewared/etc_files/netdata/netdata.conf.mako
@@ -36,7 +36,7 @@
     lib = ${netdata_state_location}
 
 [logs]
-    access = off
+	access = off
 
 [plugins]
 	proc = yes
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/etc_files/netdata/netdata.conf.mako
```diff
@@ -35,6 +35,9 @@
     home = ${netdata_cache_dataset}
     lib = ${netdata_state_location}
 
+[logs]
+    access = off
```
anodos325, line 39: tabs vs spaces?

yocalebo (changes requested):
Again sorry, but no. We're shell'ing out every 1/2 seconds for this stuff. We can't do this on our enterprise systems so we're going to have to come up with a better solution. If that means we have to disable UPS reporting for now, then so be it.

anodos325 (approved):

PR #13863, opened 2024-06-10
NAS-128893 / 24.10 / Remove md/swap usage

aiden3c:
We're removing swap entirely from Middleware, refer to this parent [Jira ticket](https://ixsystems.atlassian.net/browse/NAS-128887) for more information. 

This PR has removed all instances of swap minus the four following areas:
`10-truenas.conf` sets kernel swappiness. Unsure the correct value to set here, or if the line should entirely be removed. I'm leaning remove the line.

`htoprc` sets swap as a statistic to include. Modifying this doesn't seem to make any change on a VM.

`replace_disk.py` within `replace`. I was going to just remove the entire section, but we use `from_disk` at a later point. Wanted to make sure that someone reviews if both that section and the other usage of `from_disk` (in the try/except/else) can be safely deleted. 

`boot.py` within `legacy_schema`. I'm not entirely sure what is being done here (compared to get_boot_type alone).


sonicaj (commented):
Do we recall if we used to create persistent md mirrors because in that case they would still be around and we might probably still need to remove them and then format disks ?

src/middlewared/middlewared/plugins/pool_/export.py
```diff
@@ -128,13 +128,9 @@ async def export(self, job, oid, options):
 
         await self.middleware.call('iscsi.global.terminate_luns_for_pool', pool['name'])
 
-        job.set_progress(30, 'Removing pool disks from swap')
+        job.set_progress(30, 'Getting disks')
```
themylogin, line 131: I would suggest labeling this step as `running pre-export actions`

src/middlewared/middlewared/plugins/pool_/replace_disk.py
```diff
@@ -66,8 +66,6 @@ async def replace(self, job, oid, options):
                 # If the disk we are replacing is still available, remove it from swap as well
                 swap_disks.append(from_disk)
```
themylogin, line 67: Remove this line (and `swap_disks` variable altogether), but keep `from_disk` as it is being wiped later

themylogin:
> I'm not entirely sure what is being done here (compared to get_boot_type alone).

We've used to format boot device to use with only BIOS/EFI boot. We don't do this anymore, but we have to keep supporting formatting boot devices like this in order to be able to create mirrors for existing boot pools. Your changes are correct.

aiden3c:
> Do we recall if we used to create persistent md mirrors because in that case they would still be around and we might probably still need to remove them and then format disks ?

I found [this](https://github.com/truenas/middleware/pull/9676) PR, which seems to be making those persistent swap mirrors. I'm still going through it all to figure out how I'd check for old persistent mirrors and do said removing of them (if I need to).

sonicaj:
@aiden3c you can use the same command as in the PR to create persistent mirror and then try formatting disks to see if it works with your changes. If it does not, then that means we still need to break/destroy that mirror before actually trying to format the disks involved as that mirror would come up.

themylogin (approved):
LGTM!

aiden3c:
> @aiden3c you can use the same command as in the PR to create persistent mirror 

The command to create swap mirrors hasn't existed since [this PR](https://github.com/truenas/middleware/pull/13695). Should I go to a commit where it exists, create the mirror, then go back to my changes?

sonicaj:
@aiden3c so the path of least resistance might be to run it manually yourself without going to a scale version which actually had those changes. If you run into issues with that, then yes - the next best thing would be to go to a scale version where we were creating these mirrors.

Motivation behind this is that users who were on this scale version and had these mirrors setup, when they come to a version with your changes - they would still have those mirrors and format would likely fail in that case which we want to be sure doesn't happen.

yocalebo (approved):

PR #13864, opened 2024-06-10
NAS-129505 / 24.10 / Beginning of type hinting for middleware

creatorcary:
Type hinting is an internal documentation technique that does not affect the execution of code during runtime; type hinting is not actually enforced by the interpreter. Instead, it integrates with most IDE's and linters to make the development process cleaner by improving readability and allowing certain bugs to be noticed earlier.

### DIFF
--- a/src/middlewared/middlewared/event.py
+++ b/src/middlewared/middlewared/event.py
@@ -2,7 +2,7 @@ import asyncio
 import contextlib
 import json
 import threading
-from typing import Iterable
+import typing
 
 from middlewared.role import RoleManager
 from middlewared.schema import Any, clean_and_validate_arg, ValidationErrors
@@ -10,15 +10,12 @@ from middlewared.settings import conf
 
 
 class Events:
-    _events: dict[str, dict[str]]
-    __events_private: set[str]
-
-    def __init__(self, role_manager:RoleManager):
+    def __init__(self, role_manager: RoleManager):
         self.role_manager = role_manager
-        self._events = {}
-        self.__events_private = set()
+        self._events: typing.Dict[str, dict[str]] = {}
+        self.__events_private: typing.Set[str] = set()
 
-    def register(self, name:str, description:str, private:bool, returns:bool, no_auth_required, no_authz_required, roles:Iterable[str]):
+    def register(self, name: str, description: str, private: bool, returns, no_auth_required, no_authz_required, roles: typing.Iterable[str]):
         if name in self._events:
             raise ValueError(f'Event {name!r} already registered.')
         self.role_manager.register_event(name, roles)
@@ -33,7 +30,7 @@ class Events:
         if private:
             self.__events_private.add(name)
 
-    def get_event(self, name:str) -> dict[str]:
+    def get_event(self, name: str) -> typing.Optional[dict[str]]:
         return self._events.get(name)
 
     def __contains__(self, name):
@@ -80,7 +77,7 @@ class EventSource(metaclass=EventSourceMetabase):
         self._cancel = asyncio.Event()
         self._cancel_sync = threading.Event()
 
-    def send_event(self, event_type:str, **kwargs):
+    def send_event(self, event_type: str, **kwargs):
         if conf.debug_mode and event_type in ('ADDED', 'CHANGED'):
             verrors = ValidationErrors()
             clean_and_validate_arg(verrors, self.RETURNS[0], kwargs.get('fields'))
--- a/src/middlewared/middlewared/event.py
+++ b/src/middlewared/middlewared/event.py
@@ -12,7 +12,7 @@ from middlewared.settings import conf
 class Events:
     def __init__(self, role_manager: RoleManager):
         self.role_manager = role_manager
-        self._events: typing.Dict[str, dict[str]] = {}
+        self._events: typing.Dict[str, dict[str, typing.Any]] = {}
         self.__events_private: typing.Set[str] = set()
 
     def register(self, name: str, description: str, private: bool, returns, no_auth_required, no_authz_required, roles: typing.Iterable[str]):
@@ -30,7 +30,7 @@ class Events:
         if private:
             self.__events_private.add(name)
 
-    def get_event(self, name: str) -> typing.Optional[dict[str]]:
+    def get_event(self, name: str) -> typing.Optional[dict[str, typing.Any]]:
         return self._events.get(name)
 
     def __contains__(self, name):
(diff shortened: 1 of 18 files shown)


src/middlewared/middlewared/event.py
```diff
@@ -2,18 +2,23 @@
...
+from typing import Iterable
 
+from middlewared.role import RoleManager
 from middlewared.schema import Any, clean_and_validate_arg, ValidationErrors
 from middlewared.settings import conf
 
 
 class Events:
-    def __init__(self, role_manager):
+    _events: dict[str, dict[str]]
```
sonicaj, line 13: Why move them to class variables ? We can specify their types in `__init__` as well ?

src/middlewared/middlewared/event.py
```diff
@@ -2,18 +2,23 @@
...
+    __events_private: set[str]
+
+    def __init__(self, role_manager:RoleManager):
         self.role_manager = role_manager
         self._events = {}
         self.__events_private = set()
 
-    def register(self, name, description, private, returns, no_auth_required, no_authz_required, roles):
+    def register(self, name:str, description:str, private:bool, returns:bool, no_auth_required, no_authz_required, roles:Iterable[str]):
```
sonicaj, line 21: There should be a space after `:` i.e `name: str`

src/middlewared/middlewared/event.py
```diff
@@ -28,7 +33,7 @@ def register(self, name, description, private, returns, no_auth_required, no_aut
         if private:
             self.__events_private.add(name)
 
-    def get_event(self, name):
+    def get_event(self, name:str) -> dict[str]:
```
sonicaj, line 36: This can be null as well so we should mark it as such i.e `typing.Optional`

src/middlewared/middlewared/schema/attribute.py
```diff
@@ -80,7 +80,7 @@ def to_json_schema(self, parent=None):
         """
         raise NotImplementedError("Attribute must implement to_json_schema method")
 
-    def _to_json_schema_common(self, parent):
+    def _to_json_schema_common(self, parent) -> dict[str]:
```
sonicaj, line 83: `typing.Dict[type_here, value_type_here]`

src/middlewared/middlewared/schema/attribute.py
```diff
@@ -12,7 +12,7 @@ class Attribute:
 
     def __init__(
         self, name='', title=None, description=None, required=False, null=False, empty=True, private=False,
-        validators=None, register=False, hidden=False, editable=True, example=None, **kwargs
+        validators:list[function]=None, register=False, hidden=False, editable=True, example=None, **kwargs
```
sonicaj, line 15: This would be syntax error i.e `function` is not defined anywhere so this won't execute and the correct way to do this is with `typing.List`

sonicaj: Also i think we should have a base validators class with `__call__` method raising `NotImplementedError` and all validators inheriting from it and we can specify that validator here like this
`typing.List[Validator]`

creatorcary: I wasn't able to find a general Validator type, so I left it as List[Callable] for now.

src/middlewared/middlewared/schema/attribute.py
creatorcary: As of Python 3.9 you can use just regular list and dict, but I'll change to the preferred List and Dict

src/middlewared/middlewared/event.py
sonicaj: Also this would be `typing.Dict[key_type, value_type]`

sonicaj: Please see different modules in middleware which already have type hinting in place, that would be a nice point to see how we have been doing it so far

src/middlewared/middlewared/schema/attribute.py
creatorcary: I'm also learning the repo as I go, so Dict[str] evaluates to Dict[str, Any] since I wasn't sure

src/middlewared/middlewared/schema/attribute.py
sonicaj: Okay so to clarify, can we add something like this
```
class ValidatorBase:
    def __call__(self, *args, **kwargs):
        raise NotImplementedError()


class Exact(ValidatorBase):
    def __init__(self, value):
        self.value = value

    def __call__(self, value):
        if value != self.value:
            raise ValueError(f"Should be {self.value!r}")
```
We can import the validator base here and then use it in type hints

aiden3c (approved):

src/middlewared/middlewared/event.py
```diff
@@ -2,18 +2,20 @@
...
-    def __init__(self, role_manager):
+    def __init__(self, role_manager: RoleManager):
         self.role_manager = role_manager
-        self._events = {}
-        self.__events_private = set()
+        self._events: typing.Dict[str, dict[str, typing.Any]] = {}
+        self.__events_private: typing.Set[str] = set()
 
-    def register(self, name, description, private, returns, no_auth_required, no_authz_required, roles):
+    def register(self, name: str, description: str, private: bool, returns, no_auth_required, no_authz_required, roles: typing.Iterable[str]):
```
sonicaj, line 18: `no_auth_required/no_authz_required` have been left out ?

src/middlewared/middlewared/main.py
```diff
@@ -1645,7 +1648,7 @@ def event_register(self, name, description, *, private=False, returns=None, no_a
         roles = roles or []
         self.events.register(name, description, private, returns, no_auth_required, no_authz_required, roles)
 
-    def send_event(self, name, event_type, **kwargs):
+    def send_event(self, name, event_type: str, **kwargs):
```
sonicaj, line 1651: Missing type hint for `name` ?

src/middlewared/middlewared/schema/attribute.py
creatorcary: Thanks for the thorough review, I've updated the PR with everything you pointed out.

src/middlewared/middlewared/schema/attribute.py
```diff
@@ -1,8 +1,10 @@
 import copy
 import json
 import textwrap
+import typing
```
sonicaj, line 4: To be sure, we are not right now completing type hints here and would be done at a later stage ?

sonicaj:
Oops, got a bit late :D

src/middlewared/middlewared/event.py
creatorcary: If I came across anything that I wasn't sure about, I just skipped it. Figured it's better to be less specific than wrong, plus I plan on spending the whole week working on type hints/docstrings so I can go back to it.

PR #13865, opened 2024-06-10
NAS-129515 / 24.04.3 / Robustize ALUA

bmeagherix:
This PR contains a few components.

Summary:
- Better recovery when an service failover takes > 15 seconds
- Avoid service failover taking > 15 seconds (session teardown)
- Avoid STANDBY causing the ACTIVE node to reboot in certain circumstances

#### Robustize ALUA STANDBY against empty targets

Add `logout_empty_ha_targets` and call from `standby_after_start`.  This will prevent targets that were suspended when the STANDBY node logged into them from remaining in place, preventing further logins once they have been unsuspended.

They will not recover without a fresh login.

#### Better iSCSI suspend / resume
Suspend iSCSI activity earlier in `vrrp_master` and force close all current sessions.

Also use `clear_suspend` to resume activity, either in `become_active` if it runs to conclusion, or later in `vrrp_master` if not.

#### Remove failover.status check from active_elected
It appears that during `vrrp_master`, and **_after_** we have become MASTER `failover.status` was **not** returning "MASTER" 100% of the time.  (Presumably some race condition.)  Since we _know_ active_elected is **only** called from vrrp_master after we become MASTER, just eliminate the check.

#### dlm state robustness
Add some checks to `standby_after_start` when bringing up the STANDBY node.  Remove the `local_reset` shortcut calls, and instead cleanup `cluster_mode` one extent at a time.

This avoids an issue whereby the ACTIVE node could reboot when STANDBY was coming up, leading to an APD (All Paths Down) event.

### DIFF
--- a/src/middlewared/middlewared/plugins/dlm.py
+++ b/src/middlewared/middlewared/plugins/dlm.py
@@ -324,8 +324,9 @@ class DistributedLockManagerService(Service):
         retries = 10
         while retries and await self.middleware.call('dlm.peer_lockspaces'):
             await asyncio.sleep(1)
-            self.logger.info('Waited for lockspace to settle')
             retries -= 1
+        if retries != 10:
+            self.logger.info('Waited %d seconds for lockspace to settle', 10 - retries)
 
         # Finally turn off cluster mode locally on all extents
         try:
--- a/src/middlewared/middlewared/plugins/dlm.py
+++ b/src/middlewared/middlewared/plugins/dlm.py
@@ -312,6 +312,7 @@ class DistributedLockManagerService(Service):
     async def local_reset(self, disable_iscsi=True):
         """Locally remove the PEER node from all lockspaces and reset cluster_mode to
         zero, WITHOUT talking to the peer node."""
+        self.logger.info('local_reset starting: %r', disable_iscsi)
         # First turn off all access to targets from outside.
         if disable_iscsi:
             await self.middleware.call('iscsi.scst.disable')
@@ -319,12 +320,26 @@ class DistributedLockManagerService(Service):
         # Locally eject the peer.  Will prevent remote comms below.
         await self.eject_peer()
 
+        # Wait for up to 10 seconds for things to settle
+        retries = 10
+        while retries and await self.middleware.call('dlm.peer_lockspaces'):
+            await asyncio.sleep(1)
+            self.logger.info('Waited for lockspace to settle')
+            retries -= 1
+
         # Finally turn off cluster mode locally on all extents
         try:
             DistributedLockManagerService.resetting = True
             await self.middleware.call('iscsi.scst.set_all_cluster_mode', 0)
         finally:
             DistributedLockManagerService.resetting = False
+        self.logger.info('local_reset done')
+
+    @private
+    async def is_local_reset_complete(self):
+        if await self.middleware.call('dlm.peer_lockspaces'):
+            return False
+        return await self.middleware.call('iscsi.scst.check_cluster_modes_clear')
 
 
 async def udev_dlm_hook(middleware, data):
--- a/src/middlewared/middlewared/plugins/iscsi_/alua.py
+++ b/src/middlewared/middlewared/plugins/iscsi_/alua.py
@@ -170,10 +170,6 @@ class iSCSITargetAluaService(Service):
 
     async def become_active(self):
         self.logger.debug('Becoming active upon failover event starting')
-        # If we have not completed standby_after_start then we cannot just
-        # become ready, instead we will need to restart iscsitarget
-        if not self.standby_alua_ready:
-            raise ValueError('STANDBY node was not yet ready')
         iqn_basename = (await self.middleware.call('iscsi.global.config'))['basename']
         thisnode = await self.middleware.call('failover.node')
 
@@ -192,6 +188,14 @@ class iSCSITargetAluaService(Service):
             await self.activate_extents_job.wait()
             self.logger.debug('Waited for activate to complete')
             self.activate_extents_job = None
+
+        # If we have NOT completed standby_after_start then we cannot just
+        # become ready, instead we will need to restart iscsitarget
+        if not self.standby_alua_ready:
+            self.logger.debug('STANDBY node was not yet ready, skip become_active shortcut')
+            await self.middleware.call('service.restart', 'iscsitarget')
+            self.logger.debug('iscsitarget restarted')
+            return
 
         self.logger.debug('Updating LUNs')
         await self.middleware.call('iscsi.scst.suspend', 10)
--- a/src/middlewared/middlewared/plugins/iscsi_/alua.py
+++ b/src/middlewared/middlewared/plugins/iscsi_/alua.py
@@ -2,6 +2,7 @@ import asyncio
 import itertools
 
 from middlewared.service import Service, job
+from middlewared.utils import run
 
 CHUNK_SIZE = 20
 RETRY_SECONDS = 5
@@ -190,6 +191,7 @@ class iSCSITargetAluaService(Service):
 
         self.logger.debug('Updating LUNs')
         await self.middleware.call('iscsi.scst.suspend', 10)
+        self.logger.debug('iSCSI suspended')
         for assoc in assocs:
             extent_id = assoc['extent']
             if extent_id in extents:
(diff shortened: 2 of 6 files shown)


aiden3c (approved):

src/middlewared/middlewared/plugins/iscsi_/alua.py
```diff
@@ -103,16 +104,15 @@ async def active_elected(self, job):
...
-
-                job.set_progress(100, 'ACTIVE node ALUA reset completed')
-                self.logger.debug('ACTIVE node ALUA reset completed')
-                return
+            # Just do the bare minimum here.  This API will only be called
+            # on the new MASTER.
+            try:
+                await self.middleware.call('dlm.eject_peer')
+            except Exception as e:
+                self.logger.warning('active_elected job: %r', e)
```
yocalebo, line 112: This log statement doesn't make sense. Furthermore, no need for `e` here. Just do `self.logger.warning('blah', exc_info=True)`

src/middlewared/middlewared/plugins/iscsi_/alua.py
```diff
@@ -281,6 +251,27 @@ async def standby_after_start(self, job):
...
         else:
             job.set_progress(22, 'Remote iscsitarget is active')
+            self.logger.debug('Remote iscsitarget is active')
+
+        # Next turn off cluster_mode for all the extents.
+        # this will avoid "ignore dlm msg because seq mismatch" errors when we reconnect
+        # Rather than try to execute in parallel, we will take our time
+        while self.standby_starting:
+            try:
+                devices = await self.middleware.call('failover.call_remote', 'iscsi.scst.cluster_mode_devices_set')
```
yocalebo, line 261: When making a `call_remote` call and the TCP session is still open for the remote side but the remote side was abruptly terminated or died (panic'ed), this can hang for awhile. Best to add some arguments to the `call_remote` call here. Please see `mismatch_disks` method for the arguments that I use to be passed to `call_remote`.

src/middlewared/middlewared/plugins/iscsi_/alua.py
```diff
@@ -281,6 +251,27 @@ async def standby_after_start(self, job):
...
+                if not devices:
+                    break
+                for device in devices:
+                    await self.middleware.call('failover.call_remote', 'iscsi.scst.set_device_cluster_mode', [device, 0])
+            except Exception as e:
+                # This is a fail-safe exception catch.  Should never occur.
+                self.logger.warning('standby_start job: %r', e, exc_info=True)
```
yocalebo, line 268: No reason to have `e` here. `exc_info=True` is enough.

src/middlewared/middlewared/plugins/iscsi_/alua.py
```diff
@@ -611,3 +613,49 @@ async def wait_for_alua_settled(self, sleep_interval=1, retries=10):
             await asyncio.sleep(sleep_interval)
             retries -= 1
         self.logger.warning('Gave up waiting for ALUA to settle')
+
+    @job(lock='force_close_sessions', transient=True, lock_queue_size=1)
+    async def force_close_sessions(self, job):
+        job.set_progress(0, 'Start force-close of iSCSI sessions')
+        self.logger.debug('Start force-close of iSCSI sessions')
+
+        await run('scst_util.sh', 'force-close')
```
yocalebo, line 622: Is there a reason why we're fork'ing exec'ing in the main event loop to run some commands that we can do in native python?

src/middlewared/middlewared/plugins/iscsi_/scst.py
```diff
@@ -62,9 +82,20 @@ def disable(self):
...
         pathlib.Path(SCST_TARGETS_ISCSI_ENABLED_PATH).write_text('1\n')
 
-    def suspend(self, value):
+    def suspend(self, value=10):
         pathlib.Path(SCST_SUSPEND).write_text(f'{value}\n')
 
+    def clear_suspend(self):
+        """suspend could have been called several times, and will need to be decremented
+        several times to clean"""
+        p = pathlib.Path(SCST_SUSPEND)
```
yocalebo, line 91: Should we just do something like
```
try:
    for i in range(30):
        if pathlib.Path(SCST_SUSPEND).read_text().strip() == '0':
            return True
        else:
            p.write_text('-1\n')
            sleep(0.2)
except FileNotFoundError:
    pass

return False

**Review by yocalebo — CHANGES_REQUESTED** [review]

File: src/middlewared/middlewared/plugins/iscsi_/alua.py (hunk 4 above)
Comment (bmeagherix) [reply src/middlewared/middlewared/plugins/iscsi_/alua.py:622]: Could easily be writing to a **large** number of paths, and want to do it in parallel.

File: src/middlewared/middlewared/plugins/iscsi_/alua.py [hunk 6]
```diff
@@ -465,9 +459,17 @@ async def standby_fix_cluster_mode(self, job, devices):
...
-                await self.middleware.call('iscsi.scst.set_devices_cluster_mode', lextents, 1)
+                # Then ensure cluster_mode is enabled on the STANDBY node.  Retry if necessary.
+                retries = 10
+                while retries:
+                    try:
+                        self.logger.debug(f'Setting cluster_mode on STANDBY node for {lextents}')
+                        await self.middleware.call('iscsi.scst.set_devices_cluster_mode', lextents, 1)
+                        break
+                    except Exception as e:
+                        self.logger.warning(f'Failed to set cluster_mode on STANDBY node for {lextents}: %r', e, exc_info=True)
```
yocalebo, line 470: No need for `e` here as well.

src/middlewared/middlewared/plugins/iscsi_/alua.py
yocalebo: Derp, I completely glanced over the trailing ampersand in the script :slightly_smiling_face:

src/middlewared/middlewared/plugins/iscsi_/alua.py
```diff
@@ -611,3 +613,49 @@ async def wait_for_alua_settled(self, sleep_interval=1, retries=10):
...
+        # The main difference is these are logged out in series, to allow e.g. cluster_mode settle
+        # This is also why it is a job. it may take longer to run.
+        iqns = await self.middleware.call('iscsi.target.active_ha_iqns')
+
+        # Check what's already logged in
+        existing = await self.middleware.call('iscsi.target.logged_in_iqns')
+
+        # Generate the set of things we want to logout (don't assume every IQN, just the HA ones)
+        todo = set()
+        for iqn in iqns.values():
```
yocalebo, line 643: This isn't needed but these 4 lines can be easily converted to a single line using a comprehension
```
todo = set(i for i in iqn.values() if i in existing)
```

src/middlewared/middlewared/plugins/iscsi_/alua.py
```diff
@@ -281,6 +251,29 @@ async def standby_after_start(self, job):
...
+                devices = await self.middleware.call('failover.call_remote',
+                                                     'iscsi.scst.cluster_mode_devices_set', [],
+                                                     {'raise_connect_error': False, 'timeout': 10, 'connect_timeout': 10})
+                if not devices:
+                    break
+                for device in devices:
+                    await self.middleware.call('failover.call_remote', 'iscsi.scst.set_device_cluster_mode', [device, 0])
+            except Exception:
+                # This is a fail-safe exception catch.  Should never occur.
+                self.logger.warning('standby_start job', exc_info=True)
```
yocalebo, line 270:
```suggestion
                self.logger.warning('Unexpected failure....', exc_info=True)
```

src/middlewared/middlewared/plugins/dlm.py
```diff
@@ -312,19 +312,34 @@ async def eject_peer(self):
...
             await self.middleware.call('iscsi.scst.disable')
 
         # Locally eject the peer.  Will prevent remote comms below.
         await self.eject_peer()
 
+        # Wait for up to 10 seconds for things to settle
+        retries = 10
+        while retries and await self.middleware.call('dlm.peer_lockspaces'):
+            await asyncio.sleep(1)
+            self.logger.info('Waited for lockspace to settle')
```
yocalebo, line 327: We'll have this message get logged 10 times (potentially). I'd rather say something like `Waited %d seconds for lockspace to settle, retries`

yocalebo (approved):

PR #13877, opened 2024-06-12
NAS-129576 / 24.10 / Add ability to manage docker state

sonicaj:
This PR adds changes to bring in docker to scale image and adds ability to manager docker service lifecycle including it's state.


src/middlewared/middlewared/plugins/docker/state_setup.py
```diff
@@ -49,7 +81,7 @@ async def create_update_docker_datasets(self, docker_ds):
             )
             if not dataset:
                 test_path = os.path.join('/mnt', dataset_name)
```
Qubad786, line 83: We are not mounting to the app's dataset's natural mountpoint so this logic would not hold - we should account for `IX_APPS_DIR_NAME ` when doing this

Qubad786 (commented):
LGTM 🚀 Just one minor nit

Qubad786 (approved):

PR #13881, opened 2024-06-13
NAS-128335 / 24.10 / Fix management of SNMPv3 user

mgrimesix:
Our handling of the creation and deletion of the SNMPv3 user was incorrect for modern usage.  The result was a non-functional SNMPv3 user and the authentication and privacy phrases were stored in clear text in a file.

What was done:
- Added a 'system' SNMP user to perform actions to manage the SNMPv3 user creation and deletion
-- The 'system' user is regenerated on every SNMP restart with a new auth password
- Updated the management of the creation and deletion of the SNMPv3 user
-- The fields associated with SNMPv3 user can be independently updated, with applicable rules.
- Added the snmp set of client tools necessary for management of SNMPv3 users
- Updated the mako to eliminate SNMPv3 user private information and added the 'system' user as a permanent user

Additional items:
- The process of managing the SNMPv3 user can result in several SNMP restarts.  The restart process is quick.
- To avoid the systemd restart rate count, the limit for snmpd and snmp-agent was raise from 5 to 15.
- Added several CI tests.

Opportunistic Fix:
- The snmp-agent was using an obsolete python function: `utcnow()`.  Update this to it's modern replacement.


src/freenas/usr/local/bin/snmp-agent.py
```diff
@@ -441,11 +441,11 @@ def get_list_of_zvols():
     agent.start()
 
     prev_zpool_info = {}
-    last_update_at = datetime.min
+    last_update_at = datetime.now(timezone.utc)
```
themylogin, line 444: We should be using monotonic clock for this, not system clock

tests/api2/test_440_snmp.py
```diff
@@ -63,65 +132,254 @@ def validate_snmp_get_sysname_uses_same_ip(hostip):
...
+    if add_cmd:
+        cmd += add_cmd
+    cmd += "localhost iso.3.6.1.6.3.15.1.2.2.1.3"
+
+    # This call will timeout if SNMP is not running
+    res = ssh(cmd)
+    return [x.split()[-1].strip('\"') for x in res.splitlines()]
+
+
+@pytest.mark.dependency(name='SNMP_CONFIGURED', scope="module")
```
themylogin, line 156: What's the benefit of using this and not just fixtures?

tests/api2/test_440_snmp.py
```diff
@@ -63,65 +132,254 @@ def validate_snmp_get_sysname_uses_same_ip(hostip):
...
+def test_01_Configure_SNMP(initialize_for_snmp_tests):
```
themylogin, line 157: Why do we need these `test_01`, they are executed in the order they are defined anyway

themylogin (changes requested):

src/freenas/usr/local/bin/snmp-agent.py
anodos325: +1

src/middlewared/middlewared/plugins/snmp.py
```diff
@@ -1,10 +1,126 @@
...
+
+
+def _set_authuser_secret(secret):
+    """
+    Save the auth user secret.
+    Internal helper function for use by this module.
+    INPUT: ascii string (not encoded)
+    """
+    secret_c = Fernet(_SNMP_SYSTEM_USER['key']).encrypt(secret.encode())
+    with open(SNMP_PRIV_KEY, 'wb') as fd:
```
anodos325, line 52: Do we need to set permissions on this?

anodos325: os.fchmod(fd.fileno(), mode)

anodos325: I think it would be better to have this in a python script in `etc_files` and call into it via `etc.generate`. That way we can use standardized methods of generating config files.

src/middlewared/middlewared/plugins/snmp.py
```diff
@@ -1,10 +1,126 @@
...
+            '-l', 'authNoPriv', '-a', f'{_SNMP_SYSTEM_USER["auth_type"]}', '-A', f'{pwd}',
+            'localhost', 'delete', user
+        ]
+        # This call will timeout if SNMP is not running
+        subprocess.run(cmd, capture_output=True)
+    else:
+        raise CallError
+
+
+def _add_system_user():
```
anodos325, line 104: It would probably be better to create a directory `snmp_` and put these into a `utils.py` file within it.

src/middlewared/middlewared/plugins/snmp.py
```diff
@@ -1,10 +1,126 @@
...
+    create_v3_user += f'{snmp["v3_authtype"]} "{user_pwd}" '
+
+    if snmp.get('v3_privproto'):
+        user_phrase = snmp['v3_privpassphrase']
+        create_v3_user += f'{snmp["v3_privproto"]} "{user_phrase}" '
+
+    create_v3_user += '\n'
+
+    # Example: createUser newPrivUser MD5 "abcd1234" DES "abcd1234"
+    with open(SNMP_PRIV_CONF, 'a') as f:
```
anodos325, line 81: What permissions need to be set on this file?

tests/api2/test_440_snmp.py
mgrimesix: 'This' being the dependency mark or?
I'm not sure I understand your comment.  Could you give an example?

src/middlewared/middlewared/plugins/snmp.py
mgrimesix: agreed

src/freenas/usr/local/bin/snmp-agent.py
mgrimesix: agreed

tests/api2/test_440_snmp.py
themylogin: Yes, "this" being the dependency mark. Why do we need test dependencies?

If we want to skip executing a test if something fails to initialize, fixtures will do this perfectly.

src/freenas/usr/local/bin/snmp-agent.py
```diff
@@ -441,11 +441,11 @@ def get_list_of_zvols():
...
     while True:
         agent.check_and_process()
 
-        if datetime.utcnow() - last_update_at > timedelta(seconds=1):
+        if datetime.now(timezone.utc) - last_update_at > timedelta(seconds=1):
```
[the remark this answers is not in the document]
mgrimesix: I was thinking the same, but changing that was not part of this ticket.

src/middlewared/middlewared/etc_files/local/snmpd.conf.mako
```diff
@@ -17,17 +18,10 @@ sysObjectID 1.3.6.1.4.1.50536.3.${"1" if not middleware.call_sync("system.is_ent
...
-    % if config["v3_username"] and config["v3_password"]:
-createUser ${config["v3_username"]} ${config["v3_authtype"]} "${config["v3_password"]}" \
-        % if config["v3_privproto"] and config["v3_privpassphrase"]:
-${config["v3_privproto"]} "${config["v3_privpassphrase"]}"
-        % else:
-
-        % endif
+rwuser ${_SNMP_SYSTEM_USER['name']}
 
+% if config["v3"]:
```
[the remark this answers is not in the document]
mgrimesix: I went back and forth on this.  
We do not always need this user set in the 'public' SNMP config file, but it slightly simplifies management of the private user and leaving it reduces thrashing on this file.
The user is not functional without authentication, so there should be no harm

tests/api2/test_440_snmp.py
mgrimesix: OK, I see.  Yes, that's a nice improvement!  Thanks

anodos325 (approved):

tests/api2/test_440_snmp.py
```diff
@@ -63,65 +132,254 @@ def validate_snmp_get_sysname_uses_same_ip(hostip):
...
+        assert get_systemctl_status('snmp-agent') == "RUNNING"
+
+        # Confirm user function after change
+        user_config = call('snmp.config')
+        res = user_list_users(user_config)
+        assert user_config['v3_username'] in res
+
+
+def test_40_validate_SNMPv3_user_delete(request):
+    depends(request, ["SNMPv3_USER_ADD"], scope="module")
```
themylogin, line 367: This the same couple of issues. Numbers in the test names that serve for nothing (tests are really executed in their definition order) and dependencies that should really be fixtures.

tests/api2/test_440_snmp.py
```diff
@@ -63,65 +132,254 @@ def validate_snmp_get_sysname_uses_same_ip(hostip):
...
-
+        'location': LOCATION
+    })
 
-def test_02_Enable_SNMP_service_at_boot():
-    results = PUT('/service/id/snmp/', {'enable': True})
-    assert results.status_code == 200, results.text
 
+@pytest.mark.dependency(name='SNMP_ENABLED', scope="module")
+def test_02_Enable_SNMP_service_at_boot(request):
```
anodos325, line 175: Enabling SNMP service should be done in a pytest fixture scoped for the module.

tests/api2/test_440_snmp.py
```diff
@@ -63,65 +132,254 @@ def validate_snmp_get_sysname_uses_same_ip(hostip):
...
+    assert isinstance(id, int)
 
-def test_03_verify_snmp_do_not_leak_password_in_middleware_log(request):
-    cmd = f"""grep -R "{PASSWORD}" /var/log/middlewared.log"""
-    results = SSH_TEST(cmd, user, password)
-    assert results['result'] is False, str(results['output'])
 
+def test_03_verify_snmp_does_not_leak_password_in_logs():
+    with pytest.raises(AssertionError):
+        ssh(f'grep -R "{PASSWORD}" /var/log/middlewared.log')
```
anodos325, line 183: I don't think this is a meaningful test (I can't see an avenue for it to leak and we have separate tests to validate auto-redaction of fields marked as `private=True`

tests/api2/test_440_snmp.py
```diff
@@ -63,65 +132,254 @@ def validate_snmp_get_sysname_uses_same_ip(hostip):
...
+    depends(request, ["SNMP_ENABLED"], scope="module")
+    res = call('service.query', [['service', '=', 'snmp']])
+    assert res[0]['enable'] is True
 
 
-def test_06_checking_to_see_if_snmp_service_is_running():
-    results = GET("/service?service=snmp")
-    assert results.json()[0]["state"] == "RUNNING", results.text
+@pytest.mark.dependency(name='SNMP_STARTED', scope="module")
+def test_05_starting_snmp_service(request):
```
anodos325, line 196: starting snmp service can also be done within fixture.

tests/api2/test_440_snmp.py
```diff
@@ -63,65 +138,235 @@ def validate_snmp_get_sysname_uses_same_ip(hostip):
...
 
-def test_03_verify_snmp_do_not_leak_password_in_middleware_log(request):
-    cmd = f"""grep -R "{PASSWORD}" /var/log/middlewared.log"""
-    results = SSH_TEST(cmd, user, password)
-    assert results['result'] is False, str(results['output'])
 
+# =====================================================================
+#                           Tests
+# =====================================================================
+@pytest.mark.dependency(name='SNMP_STARTED')
```
themylogin, line 165: Started SNMP service can also be modelled as a fixture.

tests/api2/test_440_snmp.py
```diff
@@ -63,65 +138,235 @@ def validate_snmp_get_sysname_uses_same_ip(hostip):
...
+    """
+    with pytest.raises(ValidationErrors) as ve:
+        call('snmp.update', payload)
+    if attrib:
+        assert f"{attrib}" in ve.value.errors[0].attribute
+    if errmsg:
+        assert f"{errmsg}" in ve.value.errors[0].errmsg
+
+
+@pytest.mark.dependency(name='SNMPv3_USER_ADD', depends=["SNMP_STARTED"])
```
themylogin, line 270: Configured SNMP user can also be modelled as a dependency

mgrimesix: With regards to dependency, what change do you think should be made?

themylogin (approved):
Thank you, LGTM!

PR #13892, opened 2024-06-17
NAS-127183 / 24.10 / Customer designated login banner

aiden3c:
Government systems require a login banner prior to being allowed to log in. In the ticket it was determined to reuse the motd as our banner. 
This PR has the middleware portion of this ticket. On top of  giving a public endpoint for motd, we also have our motd_before_login option, which updates our sshd configuration to show our MOTD prior to login. Our database string length has also been upped from 1024 to 4096 (DoD requires 1300 minimum to fit the bare minimum message).


src/middlewared/middlewared/plugins/system_advanced/config.py
```diff
@@ -310,3 +316,8 @@ async def sed_global_password(self):
             'datastore.config', 'system.advanced', {'prefix': self._config.datastore_prefix}
         ))['sed_passwd']
         return passwd if passwd else await self.middleware.call('kmip.sed_global_password')
+
+    @accepts()
+    @no_auth_required
```
anodos325, line 321: Needs `@returns()`

anodos325: With proper schema

anodos325:
Is there any reason to not just always display motd before login?

aiden3c:
> Is there any reason to not just always display motd before login?

On the WebUI it could get annoying to have a popup modal every single time you log in with the motd. For the SSH part, we could do that but might as well keep parity with the WebUI for that logic.

src/middlewared/middlewared/schema/string_schema.py
```diff
@@ -34,7 +34,7 @@ class Str(EnumMixin, Attribute):
 
     def __init__(self, *args, **kwargs):
         # Sqlite limits ( (2 ** 31) - 1 ) for storing text - https://www.sqlite.org/limits.html
-        self.max_length = kwargs.pop('max_length', 1024) or (2 ** 31) - 1
+        self.max_length = kwargs.pop('max_length', 4096) or (2 ** 31) - 1
```
anodos325, line 37: Why are we increasing this?

aiden3c: A motd longer than 1024 characters can't be saved as it gives us a validation error, DoD needs 1300 minimum so I upped it for that. Shouldn't change the size stored in the DB, sqlite3 makes all string/varchars/etc into Text which as a limit of some really large number.

themylogin: We should pass `max_length` to the motd field, not change the default value for all fields.

anodos325: You can simply set `max_length=None` to remove the max length limitation in the Str schema for motd.

aiden3c: Didn't realize that was an option lol, fixing that right now

src/middlewared/middlewared/plugins/system_advanced/config.py
```diff
@@ -310,3 +316,9 @@ async def sed_global_password(self):
...
+    @returns(Str())
```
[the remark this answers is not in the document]
aiden3c: Good catch, works for RO now and has been working for length > 1024!

src/middlewared/middlewared/plugins/system_advanced/config.py
```diff
@@ -317,7 +317,7 @@ async def sed_global_password(self):
         ))['sed_passwd']
         return passwd if passwd else await self.middleware.call('kmip.sed_global_password')
 
-    @accepts()
+    @accepts(roles=['SYSTEM_ADVANCED_READ'])
```
anodos325, line 320: You don't need the roles. since you're setting `@no_auth_required`. I was just making sure we'd validated it worked properly when not authenticated.

Perhaps add an API test as an unauthenticated session and verify you can call it properly.

aiden3c: When making this call as a RO admin, even if `@no_authz_required` is included, without the roles a RO will still get `Not authorized`. Adding the role and just having `@no_auth_required` seems to be sufficient. 

Unless you're saying that a RO admin not being allowed is intended

aiden3c:
Instead of parsing the SSH session, would verifying that the SSHd config contains the banner line be sufficient?

src/middlewared/middlewared/plugins/system_advanced/config.py
anodos325: no_auth_required and no_authz_required should bypass privilege checks always. If that is not happening then a NAS ticket should be opened.

bmeagherix (commented):
It's usually a good idea to add a CI test.  I see there are some MOTD tests already in _test_480_system_advanced.py_

aiden3c:
Have a [passing test](http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/864)!

aiden3c:
Still have to go through and redo the DB migration so its revision hashes are correct

yocalebo (changes requested):
Hmm, I don't think adding a new db column is necessary. I'd rather keep this as simple as possible. We should just generate SSH to show the `motd` if one is set in the database. I don't want to have another boolean option that enables this functionality.

To summarize:

1. if `adv_motd` is set, then generate ssh accordingly.
2. let's not call it `get_motd`. This is actually being used as a login banner. So we can name it `login_banner`.
3. `return datastore.query, system.advanced, [], {'get': True})['adv_motd']`

aiden3c:
I added the checkbox as it was suggested in the ticket! The reasoning (from what I assume) is to prevent having a popup every single time you log in (although we could just have it show up on the login page in general, not as a modal, but then there's UI flow issues with large banners). 
This is also to not suddenly start showing the MOTD for every user with it set (which would be just about everyone since we have a default "Welcome to TrueNAS).

src/middlewared/middlewared/etc_files/local/ssh/sshd_config.mako
```diff
@@ -48,6 +48,8 @@
 	users = middleware.call_sync('user.query')
 	root_user = filter_list(users, [['username', '=', 'root']], {'get': True})
 
+	login_banner = middleware.call_sync('datastore.query', 'system.advanced')[0]['adv_login_banner']
```
sonicaj, line 51: You can use the endpoint directly you have in place for getting login banner only

src/middlewared/middlewared/plugins/system_advanced/config.py
```diff
@@ -259,6 +261,13 @@ async def do_update(self, data):
             if original_data['motd'] != config_data['motd']:
                 await self.middleware.call('etc.generate', 'motd')
 
+            if original_data['login_banner'] != config_data['login_banner']:
+                with open('/etc/login_banner', 'w', encoding='utf-8') as f:
```
sonicaj, line 265: This is an async function so we should avoid file handling in these

src/middlewared/middlewared/plugins/system_advanced/config.py
```diff
@@ -259,6 +261,13 @@ async def do_update(self, data):
...
+                    f.write(config_data['login_banner']+'\n')
+                    os.fchmod(f.fileno(), 0o600)
+                await self.middleware.call('etc.generate', 'ssh')
+                await self.middleware.call('service.restart', 'ssh')
```
sonicaj, line 269: Would reload have the same effect ?

src/middlewared/middlewared/plugins/system_advanced/config.py
sonicaj: I think you can move this to etc plugin and specify another mako file where you do this and add the perms there too - it will handle applying them nicely as well

src/middlewared/middlewared/plugins/system_advanced/config.py
```diff
@@ -310,3 +319,10 @@ async def sed_global_password(self):
             'datastore.config', 'system.advanced', {'prefix': self._config.datastore_prefix}
         ))['sed_passwd']
         return passwd if passwd else await self.middleware.call('kmip.sed_global_password')
+
+    @accepts(roles=['SYSTEM_ADVANCED_READ'])
+    @no_auth_required
+    @returns(Str())
+    def login_banner(self):
+        """Returns user set login banner"""
+        return self.middleware.call_sync('datastore.query', 'system.advanced')[0]['adv_login_banner']
```
sonicaj, line 328: We also have `datastore.config` to simplify this somewhat

src/middlewared/middlewared/plugins/system_advanced/config.py
aiden3c: What if we put the file generation in our initial `sshd_config.mako` instead? I feel like having a whole new .mako just to write one string is a little much. We already do some file IO in this config too

sonicaj: That should work

src/middlewared/middlewared/etc_files/local/ssh/sshd_config.mako
```diff
@@ -48,6 +48,12 @@
 	users = middleware.call_sync('user.query')
 	root_user = filter_list(users, [['username', '=', 'root']], {'get': True})
 
+	login_banner = middleware.call_sync('system.advanced.login_banner')
+	if login_banner != '':
+		with open('/etc/login_banner', 'w', encoding='utf-8') as f:
```
sonicaj, line 53: I think we should remove this file when `login_banner` is empty string ?

sonicaj (commented):
LGTM, just one minor nit

src/middlewared/middlewared/plugins/system_advanced/config.py
```diff
@@ -259,6 +261,10 @@ async def do_update(self, data):
             if original_data['motd'] != config_data['motd']:
                 await self.middleware.call('etc.generate', 'motd')
 
+            if original_data['login_banner'] != config_data['login_banner']:
+                await self.middleware.call('etc.generate', 'ssh')
```
sonicaj, line 265: Also you probably don't need to do this anymore because on `service.action` we generate the service's related etc group

src/middlewared/middlewared/etc_files/local/ssh/sshd_config.mako
aiden3c: What's the preferred way to delete a file? Pathlib unlink, or os.remove after checking for it existing?

sonicaj: You can do any i guess and no need to check first if the file exists or not as that can potentially lead to TOCTOU issue, instead you can catch file not found exception with `contextlib.suppress` or try/catch

aiden3c: Gotcha, I'll use os just because it's already being imported

sonicaj (approved):
Thanks! (Do sync with Andrew please about the roles thing if not already done)

src/middlewared/middlewared/etc_files/local/ssh/sshd_config.mako
```diff
@@ -48,6 +48,17 @@
...
```
anodos325, line 53: We should not be creating multiple files in a single mako file. Create a separate mako for `/etc/login_banner`

anodos325 (changes requested):
Login banner needs to be a separate etc file with entry in etc plugin defining permissions required. no_auth_required + roles still needs to be resolved.

aiden3c:
@anodos325 how would we go about deleting the old /etc/login_banner if it's in its own mako? If exiting early stops the file from being written, then I could probably do a delete then exit early.

anodos325:
> @anodos325 how would we go about deleting the old /etc/login_banner if it's in its own mako? If exiting early stops the file from being written, then I could probably do a delete then exit early.

raise FileShouldNotExist within mako file

aiden3c:
@anodos325 I've implemented those changes! Is there any way to specify that when we're reloading the ssh service that we should etc generate our login_banner as well? Would save us a line in the `config.py`

aiden3c:
> Login banner needs to be a separate etc file with entry in etc plugin defining permissions required. no_auth_required + roles still needs to be resolved.

I'm looking into this right now

anodos325:
> @anodos325 I've implemented those changes! Is there any way to specify that when we're reloading the ssh service that we should etc generate our login_banner as well? Would save us a line in the `config.py`

If you put the login banner in the etc group for `ssh` then it will be generated every time `etc.generate ssh` is called.

aiden3c:
The no_authz_required issue was me putting the accepts decorator above the no_authz_required, moving it below fixes it.

Should be good for final review now!

tests/api2/test_480_system_advanced.py
```diff
@@ -93,12 +93,26 @@ def test_10_system_advanced_check_motd_using_ssh(request):
...
+    results = PUT('/system/advanced/', {
+        'login_banner': 'TrueNAS login banner.'
+    })
+    assert results.status_code == 200, results.text
+    results = GET('/system/advanced')
+    assert results.status_code == 200, results.text
+    data = results.json()
+    assert data['login_banner'] is True
+
+    results = SSH_TEST('cat /etc/ssh/sshd_config | grep Banner', user, password)
```
anodos325, line 106: You can just `grep Banner /etc/ssh/sshd_config`

anodos325: We can probably opportunistically clean up this test file while you're here.

1. remove test numbers
2. convert REST to WS calls

aiden3c: I was following the paradigm from earlier in the tests, I'll just directly grep lol
I'll also update that earlier one too

src/middlewared/middlewared/plugins/etc.py
```diff
@@ -297,6 +297,7 @@ class EtcService(Service):
                 {'type': 'mako', 'path': 'pam.d/sshd', 'local_path': 'pam.d/sshd_linux'},
                 {'type': 'mako', 'path': 'local/users.oath', 'mode': 0o0600, 'checkpoint': 'pool_import'},
                 {'type': 'py', 'path': 'local/ssh/config'},
+                {'type': 'mako', 'path': 'login_banner', 'mode': 0o600},
```
anodos325, line 300: Also need to add `system.advanced.login_banner` to `ctx` above so that we only query once and provide to all scripts in this group.

tests/api2/test_480_system_advanced.py
aiden3c: In regards to converting REST to WS calls, should we still be checking that a dict is being returned or is that guaranteed with our WS calls?

tests/api2/test_480_system_advanced.py
```diff
@@ -19,87 +18,95 @@ def sysadv_dict():
...
 
 
-def test_11_Setting_sysloglevel():
-    results = PUT("/system/advanced/", {"sysloglevel": SYSLOGLEVEL})
-    assert results.status_code == 200, results.text
+def test_11_system_advanced_login_banner():
+    results = call('system.advanced.update', {
+        'login_banner': 'TrueNAS login banner.'
+    })
+    assert results
```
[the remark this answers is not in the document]
aiden3c: Awesome, thank you! I thought so

tests/api2/test_480_system_advanced.py
```diff
@@ -18,88 +17,84 @@ def sysadv_dict():
     return {}
 
 
-def test_01_system_advanced_get():
-    results = GET('/system/advanced/')
-    assert results.status_code == 200, results.text
-    assert isinstance(results.json(), dict)
+def test_system_advanced_get():
+    results = call('system.advanced.config')
```
anodos325, line 21: I think this test can be totally removed. It is pointless and other tests will fail spectacularly if we stop returning dict.

anodos325 (approved):

tests/api2/test_system_advanced_login_banner.py
```diff
@@ -0,0 +1,15 @@
+import os
+import pytest
+import sys
+apifolder = os.getcwd()
```
yocalebo, line 4: You don't need to do this. Just need to import the relevant modules.

yocalebo (changes requested):

yocalebo (approved):
Awesome, thanks!

PR #13899, opened 2024-06-19
NAS-125959 / 24.10 / New-style API definition

themylogin:
This PR provides an example on how to declare our API methods signatures and return values using new-style API definition. Once this is merged, our whole team will be able to start migrating from `@accepts()` and `@returns()` to new `@api_method()` decorator while I'll start working on the endpoints for the backwards-compatible API.


src/middlewared/middlewared/api/v25_04_0/common.py
```diff
@@ -0,0 +1,25 @@
+from typing import Any, List, Optional
+
+from middlewared.api.base import *
+
+__all__ = ["QueryOptions", "QueryArgs"]
+
+
+class QueryOptions(BaseModel):
+    relationships: bool = True
+    extend: Optional[str] = None
```
yocalebo, line 10: We've created a NEP for defining rules about how we do type annotations. I do not like using the `types` or `typing` module unless absolutely necessary. Instead of doing `Optional[str]`, we can replace it by doing `extend: str | None = None`

src/middlewared/middlewared/api/v25_04_0/common.py
```diff
@@ -0,0 +1,25 @@
...
+    extend_context: Optional[str] = None
+    prefix: Optional[str] = None
+    extra: dict = {}
+    order_by: List[str] = []
```
yocalebo, line 14: Same here with `List`. Can be replaced with `order_by: list[str] = []`

src/middlewared/middlewared/api/v25_04_0/user.py
```diff
@@ -0,0 +1,107 @@
+from typing import List, Optional
+
+from annotated_types import Ge, Le
+from pydantic import EmailStr
+from typing_extensions import Annotated
+
+from middlewared.api.base import *
```
yocalebo, line 7: Doing this type of import is discouraged for numerous reasons. It can be expensive since this module now has to import everything in `api.base`. It also runs risk of causing namespace collisions. Overall it's just bad practice.

src/middlewared/debian/control
```diff
@@ -23,6 +23,7 @@ Build-Depends: alembic,
                python3-cryptit,
                python3-dbus,
                python3-dnspython,
+               python3-email-validator,
```
yocalebo, line 26: Why is this being pulled in? I don't see where it's being used.

src/middlewared/middlewared/api/base/types/user.py
```diff
@@ -0,0 +1,25 @@
+import string
+
+from pydantic.functional_validators import AfterValidator
+from typing_extensions import Annotated
+
+__all__ = ["LocalUsername"]
+
+
+def validate_local_username(val):
```
yocalebo, line 9: I'm a bit torn on how this is structured. This function is in a file called `api/base/types/user.py`. That's not very intuitive to me. I wonder if we need to have a directory structure where we put functions in their own files and keep the "types" directory only filled with files that have type definitions/annotations?

yocalebo (changes requested):
Making just a few comments on my first round of review.

src/middlewared/middlewared/api/v25_04_0/user.py
```diff
@@ -0,0 +1,107 @@
...
+    local: bool
+    immutable: bool
+    twofactor_auth_configured: bool
+    nt_name: Optional[str]
+    sid: Optional[str]
+    roles: List[str]
+
+
+class UserCreate(UserEntry):
+    id: excluded() = excluded_field()
```
mgrimesix, line 51: Flake8 complains about these types of statements:  "Call expression not allowed in type expression"

mgrimesix (commented):
All my comments are regarding Flake8 complaints.
One nit (too many spaces).
The biggest Flake8 issue is the usage of `import *` which, if necessary, could be quieted with judicious use of `# noqa`.

src/middlewared/debian/control
themylogin: `pydantic` will crash on attempt to use `EmailStr` if this module is not present

src/middlewared/middlewared/api/base/types/user.py
themylogin: Is it important where this function is? It's not meant to be used publicly (`__all__` only contains `LocalUsername`)

creatorcary (approved):

yocalebo (approved):

themylogin:
time 80:00

PR #13903, opened 2024-06-20
NAS-129677 / 24.10 / Remove unnecessary dumping of args

Qubad786:
## Context

We removed websocket logging in 2982c2cfa39921917b646b8e254a050765526e38 but we left out a usage which resulted un us dumping args for method calls and it is no longer required/needed to do that anymore.


anodos325 (approved):
Good catch

PR #13906, opened 2024-06-21
NAS-129674 / 24.10 / Give core.ping no_authz_required status

creatorcary:
As per [NAS-129674](https://ixsystems.atlassian.net/browse/NAS-129674), `midclt call core.ping` would output `Not authorized` for users with the read-only admin role.


sonicaj (approved):

anodos325 (approved):

anodos325:
It probably shouldn't be too hard to add this to list of endpoints we test with RO cred.

PR #13908, opened 2024-06-21
NAS-129689 / 24.10 / Properly manage readonly state when executing initrd script

Qubad786:
## Problem
The `truenas-initrd.py` script is not correctly assessing the system state. As a result, it fails to handle specific situations like enabling dev tools. Furthermore, proper exception handling is not in place while restoring the system state.

Basically if for whatever reason this script errors out we were potentially left in an erroneous state and secondly when `install-dev-tools` had been executed and this got executed, it would change system state to readonly as well.

Also we were repeatedly unnecessarily doing `set_readonly(False)` on different function executions we have which was redundant if it had already executed once.

## Solution
Correctly assess the system state, including dev tools, and restore the system state accordingly. Additionally, add exception handling and ensure the system state is restored in the final block.

### DIFF
--- a/src/freenas/usr/local/bin/truenas-initrd.py
+++ b/src/freenas/usr/local/bin/truenas-initrd.py
@@ -58,7 +58,7 @@ def initialize_readonly_state(root: str):
     for mountpoint, key in mountpoints.items():
         try:
             st_mnt_id = stat_x.statx(mountpoint).stx_mnt_id
-            readonly_val = 'RO' in getmntinfo(st_mnt_id)[st_mnt_id]['super_opts']
+            readonly_val = 'RO' in getmntinfo(mnt_id=st_mnt_id)[st_mnt_id]['super_opts']
             for prefix in ("original_", "current_"):
                 setattr(readonly_state, f"{prefix}{key}", readonly_val)
         except (FileNotFoundError, KeyError) as e:
@@ -99,7 +99,8 @@ def set_readonly(root, readonly_desired_value=None, restore_state=None):
             # In the event this is called first with readonly set
             os.rename(os.path.join(root, "usr/local/bin/dpkg.bak"), os.path.join(root, "usr/local/bin/dpkg"))
 
-    for device_key, ds_name in filter(lambda i: i[0] in mountpoints_mapping, get_ds_conf().items()):
+    for device_key, ds_name in filter(lambda i: i[0] in mountpoints_mapping, get_ds_conf(root).items()):
+        # Do not change `readonly` property when we're running in the installer, and it was not set yet
         if (
             subprocess.run(
                 ["zfs", "get", "-H", "-o", "source", "readonly", ds_name],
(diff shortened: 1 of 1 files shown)


src/freenas/usr/local/bin/truenas-initrd.py
```diff
@@ -243,24 +297,28 @@ def update_zfs_module_config(root, database):
...
 
+        initialize_readonly_state(root)
+
         from middlewared.service_exception import CallError
         from middlewared.utils.db import FREENAS_DATABASE, query_config_table, query_table
         from middlewared.utils.gpu import get_gpus
 
         database = args.database or os.path.join(root, FREENAS_DATABASE[1:])
 
+        set_readonly(root, readonly_desired_value=False)
```
themylogin, line 308: Why change readonly state if update is not required?

Qubad786: I can specify this in different functions so this gets executed when we actually want to just turn it off, the function should handle this nicely now

themylogin (approved):

src/freenas/usr/local/bin/truenas-initrd.py
```diff
@@ -14,38 +15,95 @@
...
+
+def initialize_readonly_state(root: str):
+    global readonly_state
+
+    mountpoints = {
+        root: "root",
+        os.path.join(root, "usr"): "usr",
+    }
+
+    for partition in psutil.disk_partitions():
```
yocalebo, line 38: Do we really need to iterate over every mounted filesystem just to query a very select few datasets? I'm wondering if we just query the datasets recursively?

Finally, if we were to use `getmntinfo` function, we can parse the mount options for `RO` instead of having to subprocess out inside a loop.

yocalebo: @anodos325 any input on this loop? I believe we can get around having to subprocess entirely but maybe I'm missing something obvious.

anodos325: When TrueNAS image is created we create a file `/conf/truenas_root_ds.json` that includes all the dataset names as well as very initial options that we set. Perhaps we can just read that file and select the datasets we care about.

src/freenas/usr/local/bin/truenas-initrd.py
```diff
@@ -12,40 +13,109 @@
...
+        ):
+            readonly_value = mountpoints_mapping[device_key]
+            subprocess.run(["zfs", "set", f"readonly={'on' if readonly_value else 'off'}", ds_name])
+            setattr(readonly_state, f"current_{device_key}", readonly_value)
 
-    readonly_state = readonly
+    if mountpoints_mapping.get("usr") is False:
+        os.chmod(os.path.join(root, "usr/bin/dpkg"), 0o755)
+        with contextlib.suppress(FileNotFoundError):
+            # In the event this is called twice with readonly unset
```
anodos325, line 117: I think we should expand the comment explaining what `/usr/local/bin/dpkg` is and why we are renaming it.

yocalebo (approved):
Thanks!

PR #13912, opened 2024-06-21
NAS-129692 / 24.10 / rate limit methods that have @no_auth_required

yocalebo:
For obvious reasons, we _MUST_ protect public endpoints that are marked with `@no_auth_required` by rate limiting them. This implementation will rate limit _all_ methods that are marked with `@no_auth_required`. These rate limits are implemented by hashing them by method name AND the origin IP address that is sending the request to us. This isn't a fool-proof way and won't cover every single scenario but it's significantly better than what we currently have.

Implementation details:
1. ignore the rate limit if the method is being called across the unix socket (internal)
2. ignore the rate limit if the method is being called via the heartbeat interface on HA systems
3. store the method name and ip address of the remote client making the request
4. if a method that doesn't require authentication has been called by the same origin IP address more than 20 times within a 1min period, the rate limit will kick in
5. if the rate limits are being enforced, we will sleep a random delay to provide some form of backpressure to the offending client
6. if the global cache limit has been reached (100 total entries), requests made by _ANY_ remote caller to methods that don't require authentication will be blocked
7. the global rate limit cache will be cleared every 10mins

After adding this functionality and running it through our CI, I saw test failures for modules that use `make_ws_request`. This is for numerous reasons but the most important one is that it logged in every time it was called. This kicked in our rate limiting mechanism and so tests (specifically the FTP test) started to fail. After kicking the proverbial can down the road, I decided to go ahead and remove all uses of `make_ws_request` function since it was no longer needed.


src/middlewared/middlewared/main.py
```diff
@@ -1328,6 +1334,36 @@ def pipe(self, buffered=False):
...
+        with the `@no_auth_required` decorator. This is needed, for example, to
+        prevent anyone that knows the IP address of the TrueNAS from hammering
+        these endpoints endlessly causing the SoftSemaphoreLimit to be hit by
+        a single unauthenticated user.
+        """
+        if ip is None:
+            return
+
+        try:
+            info = self._rate_limit_info[f'{name}_{ip}']
```
anodos325, line 1351: How do we shrink this?

src/middlewared/middlewared/main.py
```diff
@@ -1328,6 +1334,36 @@ def pipe(self, buffered=False):
...
+                now = time.monotonic()
+                if info['max_period'] - (now - info['last_reset']) <= 0:
+                    # time window elapsed, so time to reset
+                    info['num_times_called'] = 0
+                    info['last_reset'] = now
+
+                # always increment
+                info['num_times_called'] += 1
+                if info['num_times_called'] > info['max_calls']:
+                    raise CallError('rate limit exceeded')
```
anodos325, line 1365: Maybe set an errno like EBUSY.

anodos325: Would it be a good idea to put a sleep in before raising the CallError to make a client back off a bit if there's a bug where application is in tight loop calling the endpoint? I've seen UI do this before.

src/middlewared/middlewared/main.py
```diff
@@ -1339,12 +1375,38 @@ def _call_prepare(
...
+                raise CallError(
+                    f'{serviceobj.__class__.__name__}.{methodobj.__name__} requires @pass_app decorator'
+                )
+            elif app is not None and isinstance(app.origin, TCPIPOrigin):
+                # it's important to rate limit based on remote addr/port
+                # combination. If we were to just apply this to the method
+                # then _ALL_ calls would be rate limited (including internal).
+                ip, port = app.origin.addr, app.origin.port
+                if not is_ha_connection(ip, port):
+                    key = '_'.join([name, ip])
```
anodos325, line 1397: Should we also include port number in key?

anodos325: You could maybe just `key = '_'.join([name, str(app.origin)])` since the str method concatenates them already.

william-gr:
We used to have a @throttle decorator for somethings.
Dunno if the intent was to handle that later or you could remove that as part of the PR, e.g. system.build_time

tests/api2/test_032_ad_kerberos.py
```diff
@@ -1,31 +1,49 @@
...
+    pytestmark = pytest.mark.skip(
+        reason='1 or all of AD_DOMAIN, ADPASSWORD, ADUSERNAME, AD_COMPUTER_OU are missing in config.py"'
+    )
+
+SAMPLE_KEYTAB = (
```
anodos325, line 21: Can we just specify noqa on the keytab file instead please

anodos325: I'm in process of refactoring these tests actually for IPA / DS changes. Unless they are functional changes (not simply flake8) please revert so that I can deal with them in my PR.

anodos325 (approved):

PR #13913, opened 2024-06-21
NAS-127804 / 24.10 / Add AsyncRedfishClient

bmeagherix:
This (`aiohttp`-based) async client is capable of transparently utilizing multiple redfish interfaces to the same JBOF.

It will be used by the new JBOF enclosure services for performing redfish operations.  (Will be introduced in a separate PR.)  When multiple JBOFs are present, this client will be used to fetch the data from them in parallel.

(The existing synchronous `requests`-based client is also being retained for initial setup/config operations.  With that client we _deliberately_ do **not** transparently handle switching between redfish interfaces.)

### DIFF
--- a/src/middlewared/middlewared/plugins/jbof/redfish/client.py
+++ b/src/middlewared/middlewared/plugins/jbof/redfish/client.py
@@ -18,6 +18,7 @@ HEADER = {'Content-Type': 'application/json', 'Vary': 'accept'}
 REDFISH_ROOT_PATH = '/redfish/v1'
 ODATA_ID = '@odata.id'
 LOGGER = logging.getLogger(__name__)
+REDFISH_SESSIONS = '/redfish/v1/SessionService/Sessions'
 
 
 class InvalidCredentialsError(Exception):
@@ -89,7 +90,7 @@ class RedfishClient(AbstractRedfishClient):
         try:
             self.login_url = self.root['Links']['Sessions']['@odata.id']
         except KeyError:
-            self.login_url = '/redfish/v1/SessionService/Sessions'
+            self.login_url = REDFISH_SESSIONS
 
         if username and password:
             self.login()
@@ -385,6 +386,7 @@ class AsyncRedfishClient(AbstractRedfishClient):
         self._sessions = {}
         # Implement a little value cache
         self._attributes = {}
+        self.logged = {}
 
     def _add_session(self, base_url, session):
         """Add a session corresponding to the base_url"""
@@ -430,7 +432,7 @@ class AsyncRedfishClient(AbstractRedfishClient):
         try:
             self.login_url = self.root['Links']['Sessions']['@odata.id']
         except KeyError:
-            self.login_url = '/redfish/v1/SessionService/Sessions'
+            self.login_url = REDFISH_SESSIONS
 
         return self
 
@@ -494,6 +496,11 @@ class AsyncRedfishClient(AbstractRedfishClient):
                 raise ValueError('Invalid auth supplied:', authtype)
         # Save the session for reuse
         self._add_session(base_url, newsession)
+        # Clear any silenced exception logging
+        try:
+            self.logged.remove(base_url)
+        except KeyError:
+            pass
         return newsession
 
     async def _logout(self, base_url, session):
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/jbof/redfish/client.py
```diff
@@ -341,3 +346,240 @@ def link_status(self, uri):
...
+            try:
+                if not session:
+                    session = await self._login(base_url)
+                async with session.post(uri, ssl=False, json=payload) as response:
+                    if response.ok:
+                        return await response.json()
+            except asyncio.TimeoutError:
+                await self._del_session(base_url, session)
+                continue
+            except Exception:
```
anodos325, line 561: Would it be helpful in these situations to log error?

src/middlewared/middlewared/plugins/jbof/redfish/client.py
```diff
@@ -341,3 +346,240 @@ def link_status(self, uri):
...
+                     default_prefix=REDFISH_ROOT_PATH,
+                     timeout=DEFAULT_REDFISH_TIMEOUT_SECS):
+        """Async factory to create an AsyncRedfishClient object."""
+        self = cls(base_urls, username, password, authtype, default_prefix, timeout)
+        self.root = await self.get_root_object()
+
+        try:
+            self.login_url = self.root['Links']['Sessions']['@odata.id']
+        except KeyError:
+            self.login_url = '/redfish/v1/SessionService/Sessions'
```
anodos325, line 433: Would it make sense to have this be a constant?

src/middlewared/middlewared/plugins/jbof/redfish/client.py
```diff
@@ -341,3 +346,240 @@ def link_status(self, uri):
...
+        return newsession
+
+    async def _logout(self, base_url, session):
+        if self.authtype == AuthMethod.BASIC:
+            self.auth = None
+        elif self.authtype == AuthMethod.SESSION:
+            if location := self.session_location.get(base_url):
+                async with session.delete(location, ssl=False) as response:
+                    if response.ok:
+                        del self.session_location[base_url]
```
anodos325, line 506: Is it possible to fail with KeyError here?

src/middlewared/middlewared/plugins/jbof/redfish/client.py
anodos325: (and above)

src/middlewared/middlewared/plugins/jbof/redfish/client.py
bmeagherix: I don't believe so, as we just did the dict.get, but can protect anyway.

anodos325 (approved):

src/middlewared/middlewared/plugins/jbof/redfish/client.py
```diff
@@ -341,3 +347,258 @@ def link_status(self, uri):
...
+                await self._del_session(base_url, session)
+                continue
+        raise CallError(f'Failed to GET {uri}:', errno.EBADMSG)
+
+    async def post(self, uri, **kwargs):
+        if not uri.startswith(self.prefix):
+            uri = f'{self.prefix}{uri}'
+        payload = kwargs.get('data', {})
+        # Iterate over the available paths
+        for base_url, session in self.sessions():
```
yocalebo, line 566: Maybe it's because there isn't any consumer of this new class but I'm failing to understand how the `post` and `get` methods work. If I'm reading it correctly, the methods are given a URI and then we iterate over all cached sessions and on the first session for which we get a valid response, we short-circuit and return the information. So this means we return the information on the 1st successful cached session?

bmeagherix: Correct.  The idea is that each `base_url` represents a path by which an operation may be accomplished (GET, POST, etc) - we don't care which path gets used, but we want to be resilient in case one path goes down.

In the `cache_get` classmethod you can see how `base_urls` gets set to `['https://<mgmt_ip1>', 'https://<mgmt_ip2>']` and (in `_login`) we construct `aiohttp.ClientSession` corresponding to each (well, we're lazy so we may never actually construct the 2nd if the 1st works all the time).

FWIW, the `chassis` and `managers` methods indirectly call the `get`.

PR #13917, opened 2024-06-21
NAS-129695 / 24.10 / Change download method from call to call_with_audit.

mgrimesix:
The method calls that are part of `core.download` are not being audited.

The parameters for `core.download` include 
When adding auditing of the audit subsystem it was discovered that method calls via `core.download` were not being audited.
This PR converts `core.download` to include auditing.

### DIFF
--- a/src/middlewared/middlewared/service/core_service.py
+++ b/src/middlewared/middlewared/service/core_service.py
@@ -694,7 +694,11 @@ class CoreService(Service):
         return await self._download(app, method, args, filename, buffered)
 
     async def _download(self, app, method, args, filename, buffered):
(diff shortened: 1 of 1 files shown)


themylogin (changes requested):
Can you add a test for this please?

mgrimesix:
CI tests are going to follow.  
We need a download that includes a call to an audited subsystem.
This issue was discovered in the new CI tests of the changes for adding auditing of the audit subsystem.  This particular change _was_ going to be part of that PR, but it was requested I separate it out.

themylogin (approved):

PR #13921, opened 2024-06-24
Fix NoneType - set() regression

aiden3c:
To fix this NoneType regression, we null coalesce our NIC sets so they return an empty set if our `get_nic_names`.
This was chosen over updating the `get_nic_names` function as in the scenario that caused the regression was there not being a `get_nic_names` function on NodeB. This caused our remote call to return nothing.


aiden3c:
Retest this please

creatorcary (approved):

PR #13925, opened 2024-06-25
NAS-129686 / 24.10 / Force directory service cache insertion on share ACL read

anodos325:
When user / group enumeration are disabled for a directory service we rely on user.query and group.query to perform cache insertions as we have successes on lookups so that UI can present users with more useful options. This commit slightly refactors the getacl method for SMB shares to contribute to the lazy caching effort.


anodos325:
Closing PR and will make new ticket since this one was reassigned to UI team.

PR #13926, opened 2024-06-25
NAS-129735 / 24.10 / Force directory service cache insertion on share ACL read

anodos325:
When user / group enumeration are disabled for a directory service we rely on user.query and group.query to perform cache insertions as we have successes on lookups so that UI can present users with more useful options. This commit slightly refactors the getacl method for SMB shares to contribute to the lazy caching effort.

### DIFF
--- a/src/middlewared/middlewared/plugins/idmap.py
+++ b/src/middlewared/middlewared/plugins/idmap.py
@@ -1081,35 +1081,37 @@ class IdmapDomainService(CRUDService):
         return filter_list(out, filters, options)
 
     @private
-    async def id_to_name(self, id_, id_type):
+    async def id_to_name(self, xid, id_type):
         idtype = IDType[id_type]
         idmap_timeout = 5.0
+        options = {'extra': {'additional_information': ['DS']}, 'get': True}
 
-        if idtype == IDType.GROUP or idtype == IDType.BOTH:
-            method = "group.get_group_obj"
-            to_check = {"gid": id_}
-            key = 'gr_name'
-        elif idtype == IDType.USER:
-            method = "user.get_user_obj"
-            to_check = {"uid": id_}
-            key = 'pw_name'
-        else:
-            raise CallError(f"Unsupported id_type: [{idtype.name}]")
+        match idtype:
+            case IDType.GROUP | IDType.BOTH:
+                method = 'group.query'
+                filters = [['gid', '=', xid]]
+                key = 'group'
+            case IDType.USER:
+                method = 'user.query'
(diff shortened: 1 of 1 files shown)


anodos325:
Test coverage is already provided by AD cache tests.

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -1081,35 +1081,37 @@ async def builtins(self, filters, options):
...
             self.logger.debug(
-                "timeout encountered while trying to convert %s id %s "
+                "timeout encountered while trying to convert %s id %d "
                 "to name. This may indicate significant networking issue.",
-                id_type.lower(), id_
+                id_type.lower(), xid
             )
             name = None
-        except KeyError:
+        except MatchNotFound:
```
mgrimesix, line 1114: Need to import `MatchNotFound`

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -1081,35 +1081,37 @@ async def builtins(self, filters, options):
...
-        elif idtype == IDType.USER:
-            method = "user.get_user_obj"
-            to_check = {"uid": id_}
-            key = 'pw_name'
-        else:
-            raise CallError(f"Unsupported id_type: [{idtype.name}]")
+        options = {'extra': {'additional_information': ['DS']}, 'get': True}
+
+        match idtype:
+            case IDType.GROUP | IDType.BOTH:
```
mgrimesix, line 1090: While not directly part of this PR, could you add some comments in `idmap_constants.py` with a simple explanation of what `BOTH` means?  I _assume_ it means 'either', but... 🤷‍♂️

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -1081,35 +1081,37 @@ async def builtins(self, filters, options):
...
-            method = "group.get_group_obj"
-            to_check = {"gid": id_}
-            key = 'gr_name'
-        elif idtype == IDType.USER:
-            method = "user.get_user_obj"
-            to_check = {"uid": id_}
-            key = 'pw_name'
-        else:
-            raise CallError(f"Unsupported id_type: [{idtype.name}]")
+        options = {'extra': {'additional_information': ['DS']}, 'get': True}
```
mgrimesix, line 1087: Is this the DS cache insertion?  Maybe add a comment to that effect.

mgrimesix (changes requested):
Missing import.

mgrimesix (approved):

PR #13929, opened 2024-06-26
NAS-129453 / 24.04.3 / Allow VM deletion if virtualization is not supported

Qubad786:
## Context

This change was made to EE but not backported (https://github.com/truenas/middleware/pull/13776).
Essentially `vm.query` was failing on an unrelated operation as it was being called by filesystem attachment delegates in the case when virtualization was not supported.


yocalebo:
Please add the `backport-24.04.3` label to the original PR and then make a comment to say "backport" so that it gets tracked under same ticket. Close this ticket out in jira.

PR #13936, opened 2024-06-26
NAS-127613 / 24.10 / Add ES24N sensor support

bmeagherix:
Make `map_jbof` async and process JBOFs in parallel.

Refactor code and in an async `map_jbof` now call `map_es24n` in parallel.  This will use the new `AsyncRedfishClient` (being added in PR #13913) to obtain various information from the JBOF redfish interface, which will then be mapped to a similar `dict` as is done for other enclosure types.

This commit adds "Power Supply", "Cooling", "Temperature Sensors" and "Voltage Sensor" to the elements returned by `map_es24n` (was already returning "Array Device Slot").

Abstracted out some strings from `ElementStatus` and `ElementType` to enums for reuse.


src/middlewared/middlewared/plugins/enclosure_/jbof/es24n.py
```diff
@@ -0,0 +1,140 @@
...
+    which is unfortunate but there isn't much we can do about it
+    at the moment. We know what the URI _should_ be for this
+    platform and we _thought_ we knew what the model should be so
+    we'll hard-code these values and check for the specific URI
+    and then check if the model at the URI at least has some
+    semblance of an ES24N"""
+    # FIXME: This function shouldn't exist and the OEM should fix
+    # this at some point. When they do (hopefully) fix the model,
+    # remove this function
+    expected_uri = '/redfish/v1/Chassis/2U24'
```
anodos325, line 116: Should this be defined as a constant?

yocalebo:
Something I've noticed and should be considered. This area of our code is called quite a bit so the more logging statements we add, the more our log files will be spammed. When I wrote `map_es24n` function, I explicitly didn't add a bunch of failures there so we didn't spam. I tried to only log when it an exception occurred. Seems like we log a bit more with these changes. It's not necessarily wrong, but it's something to keep in mind.

yocalebo (approved):

PR #13937, opened 2024-06-26
NAS-129763 / 24.10 / Expose ability for users to configure SMB encryption behavior

anodos325:
This commit adds a new SMB server global parameter for SMB transport encryption. It exposes 4 options:

DEFAULT -- follow upstream / truenas default (reflected as a null in db) -- currently NEGOTIATE
NEGOTIATE -- only encrypt transport if explicitly requested by the SMB client
DESIRED -- encrypt transport if supported by client during session negotiation.
REQUIRED -- always encrypt transport (rejecting access if client does not support encryption -- incompatible with SMB1 server).

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -70,6 +70,7 @@ class SMBModel(sa.Model):
     cifs_srv_next_rid = sa.Column(sa.Integer(), nullable=False)
     cifs_srv_secrets = sa.Column(sa.EncryptedText(), nullable=True)
     cifs_srv_multichannel = sa.Column(sa.Boolean, default=False)
+    cifs_srv_encryption = sa.Column(sa.String(120), nullable=True)
 
 
 class SMBService(ConfigService):
@@ -98,6 +99,7 @@ class SMBService(ConfigService):
         smb['netbiosname_local'] = smb['netbiosname']
         smb['netbiosalias'] = (smb['netbiosalias'] or '').split()
         smb['loglevel'] = LOGLEVEL_MAP.get(smb['loglevel'])
+        smb['encryption'] = smb['encryption'] or 'DEFAULT'
         smb.pop('secrets', None)
         return smb
 
@@ -499,6 +501,12 @@ class SMBService(ConfigService):
                 'Please provide a valid value for unixcharset'
             )
 
+        if new['enable_smb1'] and new['encryption'] == 'REQUIRED':
+            verrors.add(
+                'smb_update.encryption',
+                'Encryption may not be set to REQUIRED while SMB1 support is enabled.'
+            )
+
         for i in ('workgroup', 'netbiosname', 'netbiosalias'):
             """
             There are two cases where NetBIOS names must be rejected:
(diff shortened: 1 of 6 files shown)


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -598,6 +606,7 @@ async def validate_smb(self, new, verrors):
         Str('dirmask'),
         Bool('ntlmv1_auth'),
         Bool('multichannel', default=False),
+        Str('encryption', enum=['DEFAULT', 'NEGOTIATE', 'DESIRED', 'REQUIRED']),
```
bmeagherix, line 609: So, this parameter **_must_** be supplied?  (i.e. no default of DEFAULT?)

mgrimesix (approved):

tests/api2/test_smb_encryption.py
```diff
@@ -94,3 +105,27 @@ def test__smb_client_encrypt_required(smb_setup):
...
+@pytest.mark.parametrize('enc_param', ('DESIRED', 'REQUIRED'))
+def test__smb_client_server_encrypt(smb_setup, enc_param):
+    with server_encryption(enc_param):
+        with smb_connection(
+            share=smb_setup['share']['name'],
+            username=SHAREUSER,
+            password=PASSWD,
+            encryption='DEFAULT'
+        ) as c:
+            # perform basic op to fully initialize SMB session
```
bmeagherix, line 119: Hmm.  Should the next line be the `c.ls('/')` ?

tests/api2/test_smb_encryption.py
```diff
@@ -94,3 +105,27 @@ def test__smb_client_encrypt_required(smb_setup):
...
+            assert c.get_smb_encryption() == 'DEFAULT'
```
bmeagherix, line 120: Is this line just showing what the client requested?  Maybe add a comment.

src/middlewared/middlewared/pytest/unit/plugins/test_smb.py
```diff
@@ -43,6 +44,10 @@
 SMB_SMB1 = BASE_SMB_CONFIG | {'enable_smb1': True}
 SMB_MULTICHANNEL = BASE_SMB_CONFIG | {'multichannel': True}
 SMB_OPTIONS = BASE_SMB_CONFIG | {'smb_options': 'canary = bob\n canary2 = bob2 \n #comment\n ;othercomment'}
+SMB_ENCRYPTION_NEGOTIATE = BASE_SMB_CONFIG | {'encryption': 'NEGOTIATE'}
```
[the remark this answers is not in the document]
anodos325: OK

src/middlewared/middlewared/plugins/smb.py
anodos325: No. If it is omitted then current value is preserved. I left it as this to prevent accidental changes to transport encryption settings via incomplete payload.

tests/api2/test_smb_encryption.py
anodos325: Yes. Adding comment.

src/middlewared/middlewared/pytest/unit/plugins/test_smb.py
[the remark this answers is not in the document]
anodos325: Actually it's set to `DEFAULT` in BASE_SMB_CONFIG, so we're good on this count.

tests/api2/test_smb_encryption.py
anodos325: Yes

bmeagherix (approved):

PR #13938, opened 2024-06-27
NAS-129756 / 24.04.3 / Do not spam logs with alert formatting errors

themylogin:


src/middlewared/middlewared/alert/base.py
```diff
@@ -271,15 +271,6 @@ def formatted(self):
         try:
             return self.klass.format(self.args)
         except Exception:
-            logger.debug("Alert class %r was unable to format args %r, falling back to default formatter",
-                         self.klass, self.args, exc_info=True)
-
-            if self.args:
-                try:
-                    return self.text % (tuple(self.args) if isinstance(self.args, list) else self.args)
```
themylogin, line 279: This can be removed because default  `self.klass.format` implementation does the same already (and it is not overridden by any existing alert class).

themylogin:
time 0:10

yocalebo (approved):

PR #13942, opened 2024-06-27
NAS-129780 / 24.10 / Simplify API for querying directory services users and groups

anodos325:
Cache query responses are fast enough that we should include by default. Directory services users and groups can be reasonably omitted if filters explicitly call for local or builtin users.


src/middlewared/middlewared/plugins/account.py
```diff
@@ -289,16 +307,23 @@ async def query(self, filters, options):
                 # broken
                 self.logger.error('Failed to retrieve passdb information', exc_info=True)
 
-        if dssearch:
-            ds_state = await self.middleware.call('directoryservices.get_state')
-            if ds_state['activedirectory'] == 'HEALTHY' or ds_state['ldap'] == 'HEALTHY':
-                ds_users = await self.middleware.call('directoryservices.cache.query', 'USER', filters, options.copy())
-                # For AD users, we will not have 2FA attribute normalized so let's do that
```
bmeagherix, line 296: Lost this comment .  Does it still apply to the updated code?

bmeagherix (changes requested):
sshd_config.mako contains:
```
	users = middleware.call_sync('user.query')
	root_user = filter_list(users, [['username', '=', 'root']], {'get': True})
```
Presumably this should be updated.  (And if not, maybe add a comment saying why not.)

----
plugins/smb.py `share_precheck `contains
```
            local_smb_user_cnt = await self.middleware.call(
                'user.query',
                [['smb', '=', True]],
                {'count': True}
            )
```
should this also be updated?   (And if not, maybe add a comment saying why not.)

bmeagherix (approved):

PR #13943, opened 2024-06-27
NAS-129782 / 24.10 / Add auditing of audit ops

mgrimesix:
Added auditing of audit ops.  Also includes:
- Returning to `refquota`
- Fixing a typo and converting the unit conversion divisions to `//`

Very minor cleanup of audit utils:  Fix Flake8 complaint about extra lines

Refactored test_audit_basic:
- Enhanced testing
-- More checks with ZFS dataset values
- Converted to using refquota.

Added CI tests for the new auditing in test_audit_audit.py.

### DIFF
--- a/tests/api2/test_audit_basic.py
+++ b/tests/api2/test_audit_basic.py
@@ -3,7 +3,6 @@ from middlewared.test.integration.assets.pool import dataset
 from middlewared.test.integration.assets.smb import smb_share
 from middlewared.test.integration.utils import call, url
 from protocols import smb_connection
-from pytest_dependency import depends
 from time import sleep
 
 import os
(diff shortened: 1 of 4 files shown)


tests/api2/test_audit_basic.py
```diff
@@ -15,10 +14,60 @@
...
+            'used_by_reservation': zfs_config['properties']['usedbyrefreservation']['parsed']
+        }
+    }
+    # return zfs[key]
+    return types[type][key]
 
 
-@pytest.fixture(scope='module')
-def initialize_for_smb_tests(request):
+@pytest.fixture(scope='class')
```
anodos325, line 69: Why are we changing scope?

mgrimesix: Only need the smb stuff for the operations section.

tests/api2/test_audit_basic.py
```diff
@@ -35,147 +84,156 @@ def initialize_for_smb_tests(request):
...
-        'quota_fill_warning': 75,
-        'quota_fill_critical': 95
-    }
-
-
-@pytest.fixture(scope='module')
-def audit_config(request):
+@pytest.fixture(scope='class')
+def init_audit():
+    """ Provides the audit and dataset configs and cleans up afterwared """
```
bmeagherix, line 89: Typo: afterwared

tests/api2/test_audit_basic.py
```diff
@@ -35,147 +84,156 @@ def initialize_for_smb_tests(request):
...
+        for service in ['MIDDLEWARE', 'SMB', 'SUDO']:
+            assert service in config['enabled_services']
+
+        # Confirm audit dataset settings
+        for key in [k for k in AUDIT_DATASET_CONFIG if AUDIT_DATASET_CONFIG[k] == 'zfs']:
+            assert get_zfs('zfs', key, dataset) == config[key], f"config[{key}] = {config[key]}"
+
+    def test_audit_config_dataset_defaults(self):
+        """ Confirm Audit dataset uses Audit default settings """
+        ds_config = call('audit.get_audit_dataset')
```
bmeagherix, line 128: Any reason not to use the dataset value from `init_audit` here?

tests/api2/test_audit_basic.py
```diff
@@ -15,10 +14,60 @@
...
+        'retention': 7,
+        'quota': 0,
+        'reservation': 0,
+        'quota_fill_warning': 75,
+        'quota_fill_critical': 95
+    }
+
+
+# def get_zfs(key, zfs_config):
+def get_zfs(type, key, zfs_config):
```
bmeagherix, line 45: `type` is a python builtin (class).  Maybe better to use a different name instead.

tests/api2/test_audit_basic.py
```diff
@@ -35,147 +84,156 @@ def initialize_for_smb_tests(request):
...
+
+        # Confirm audit dataset settings
+        for key in [k for k in AUDIT_DATASET_CONFIG if AUDIT_DATASET_CONFIG[k] == 'zfs']:
+            assert get_zfs('zfs', key, dataset) == config[key], f"config[{key}] = {config[key]}"
+
+    def test_audit_config_dataset_defaults(self):
+        """ Confirm Audit dataset uses Audit default settings """
+        ds_config = call('audit.get_audit_dataset')
+        assert ds_config['org.freenas:quota_warning'] == AUDIT_CONFIG.defaults['quota_fill_warning']
+        assert ds_config['org.freenas:quota_critical'] == AUDIT_CONFIG.defaults['quota_fill_critical']
```
bmeagherix, line 130: Huh.  I see 4 entries ...
- `org.freenas:quota_warning`
- `org.freenas:quota_critical`
- `org.freenas:refquota_warning`
- `org.freenas:refquota_critical`
From the PR description you mention `refquota`, so shouldn't we be testing that (instead?)

tests/api2/test_audit_basic.py
```diff
@@ -35,147 +84,156 @@ def initialize_for_smb_tests(request):
...
+        # Check that we're actually setting the quota by evaluating available space
+        # Change the the quota to something more interesting
+        new_config = call('audit.update', {'quota': 2})
+        assert new_config['quota'] == 2
+
+        audit_dataset = call('audit.get_audit_dataset')
+        assert get_zfs('zfs', 'refquota', audit_dataset) == 2*GiB  # noqa (allow 2*GiB)
+
+        used_in_dataset = get_zfs('space', 'used_by_dataset', audit_dataset)
+        assert abs(new_config['space']['available'] - 2*GiB) == used_in_dataset  # noqa (allow 2*GiB)
```
bmeagherix, line 160: Subtraction seems backward (albeit inside an `abs`), but maybe add a comment to explain this test.

tests/api2/test_audit_basic.py
```diff
@@ -35,147 +84,156 @@ def initialize_for_smb_tests(request):
...
+                c.read(fd, 0, 3)
+            c.close(fd, True)
+
+        sleep(10)
+        ops = call('audit.query', {
+            'services': ['SMB'],
+            'query-filters': [['username', '=', SMBUSER]],
+            'query-options': {'count': True}
+        })
+        assert ops > 0
```
bmeagherix, line 203: Hmmm.
If we ran this section of code before the `with smb_connection`, would we get `ops` == 0?

If not, then we're not testing anything.  Would it make sense to run before and after and ensure the count increased?

tests/api2/test_audit_basic.py
mgrimesix: I was re-using the original, but yeah, we can eliminate the `abs` with reversal and actually it makes for a better test since if the value goes negative that indicates the quota is not working!

tests/api2/test_audit_basic.py
mgrimesix: This is not a new test from me, but the test creates a temporary user, then calls `audit.query` _filtering for that user_.
So if the query and filters are working we should get non-zero only after creating the user.
Now, in general it's also good to do negative tests, so I'll add the same query before the user add and confirm `0`.
Thanks!

bmeagherix (approved):

PR #13944, opened 2024-06-28
NAS-129791 / 24.10 / Fix validation for homedir mountinfo

anodos325:
Users may receive spurious validation error messages if they bind mount their home dataset to another filesystem path readonly because we are looking up filesystem by dev_t and not the stx_mnt_id.

This commit converts a filesystem.mount_info call for path or parent path into a filesystem.statfs call (which performs correct looup) for the same.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -320,18 +320,13 @@ class UserService(CRUDService):
         )
 
     @private
-    def validate_homedir_mountinfo(self, verrors, schema, dev):
-        mntinfo = self.middleware.call_sync(
-            'filesystem.mount_info',
-            [['device_id.dev_t', '=', dev]],
-            {'get': True}
-        )
-
-        if 'RO' in mntinfo['mount_opts']:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/account.py
```diff
@@ -320,18 +320,13 @@ async def query(self, filters, options):
...
     @private
-    def validate_homedir_mountinfo(self, verrors, schema, dev):
-        mntinfo = self.middleware.call_sync(
-            'filesystem.mount_info',
-            [['device_id.dev_t', '=', dev]],
-            {'get': True}
-        )
-
-        if 'RO' in mntinfo['mount_opts']:
+    def validate_homedir_mountinfo(self, verrors, schema, home_path):
```
yocalebo, line 323: `dev` and `home_path` are the same?? Seems strange that we were passing a `pathlib.Path` object into the filters of `filesystem.mount_info`? How did this work at all?

anodos325: Originally st_dev was being passed, not a pathlib.Path.

yocalebo (approved):

PR #13945, opened 2024-06-28
NAS-129750 / 24.10 / fix scrambing start sector of data partitions in pool.expand

yocalebo:
A fix was done here recently to ensure we maintain the start sector offset of the data partition for `pool.expand`. However, it uncovered the fact that we're not calculating the start _sector_ properly because the kernel has 0 documentation on these sysfs files (which is where we enumerate this information) and so we made an incorrect assumption of how it works. This manifests itself on 4kn drives. Almost all userspace tools for manipulating partitions will default to expecting the information is provided in sectors. Without these changes, we're not doing that properly. Reproduced in-house trivially and was able to test these changes afterwards.

NOTE: I will be pushing another proper PR with tests after this is merged. However, I need to add 4kn emulated drives to our VM environment for the tests to be valid. This needs to be merged in time for 24.04.2 so this is why I'm opening it up without tests right now.

NOTE 2: I've found that we have a bunch of code duplication related to enumerating partition info. I will push, yet another, PR to master that consolidates this code and makes this situation much harder to hit in the future.

### DIFF
--- a/src/middlewared/middlewared/plugins/disk_/disk_info.py
+++ b/src/middlewared/middlewared/plugins/disk_/disk_info.py
@@ -1,3 +1,4 @@
+import collections
 import contextlib
 import glob
 import os
@@ -8,6 +9,32 @@ import pyudev
 from middlewared.service import CallError, private, Service
 
 
+# The basic unit of a block I/O is a sector. A sector is
+# 512 (2 ** 9) bytes. In sysfs, the files (sector_t type)
+# `<disk>/<part>/start` and `<disk>/<part>/size` are
+# shown as a multiple of 512 bytes. Most user-space
+# tools (fdisk, parted, sfdisk, etc) treat the partition
+# offsets in sectors.
+BYTES_512 = 512
+PART_INFO_FIELDS = (
+    # queue/logical_block_size (reported as a multiple of BYTES_512)
+    'lbs',
+    # starting offset of partition in sectors
+    'start_sector',
+    # ending offset of partition in sectors
+    'end_sector',
+    # total partition size in sectors
+    'total_sectors',
+    # starting offset of partition in bytes
+    'start_byte',
+    # ending offset of partition in bytes
+    'end_byte',
+    # total size of partition in bytes
+    'total_bytes',
+)
+PART_INFO = collections.namedtuple('part_info', PART_INFO_FIELDS, defaults=(0,) * len(PART_INFO_FIELDS))
+
+
 def get_partition_size_info(disk_name, s_offset, s_size):
     """Kernel sysfs reports most disk files related to "size" in 512 bytes.
     To properly calculate the starting SECTOR of partitions, you must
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/disk_/disk_info.py
```diff
@@ -7,6 +8,31 @@
...
+    do some calculations. It is _very_ important to do this properly
+    since almost all userspace tools that format disks expect partition
+    positions to be in sectors."""
+    lbs = start_sect = end_sect = sect_size = 0
+    with contextlib.suppress(FileNotFoundError, ValueError):
+        with open(f'/sys/block/{disk_name}/queue/logical_block_size') as f:
+            lbs = int(f.read().strip())
+
+    if not lbs:
+        # this should never happen
```
anodos325, line 24: Do we need to raise an exception or log an error message?

yocalebo: I'm afraid to since this is landing in 24.04.2 and this function is being called in critical paths. I'd rather just return without crashing since I'm not sure what other regressions may occur if I raise an exception. I'll push a more involved fix for EE+ to clean all of this up.

src/middlewared/middlewared/plugins/disk_/disk_info.py
```diff
@@ -7,6 +8,31 @@
...
+        return lbs, start_sect, end_sect, sect_size
+
+    # important when dealing with 4kn drives
+    divisor = lbs // 512
+    start_sect = s_offset // divisor
+    sect_size = s_size // divisor
+    end_sect = sect_size + start_sect - 1
+
+    return lbs, start_sect, end_sect, sect_size
```
themylogin, line 33: It would be great to return a namedtuple instead

themylogin (approved):

src/middlewared/middlewared/plugins/disk_/disk_info.py
yocalebo: good idea, I'll add that

anodos325 (approved):

src/middlewared/middlewared/plugins/disk_/disk_info.py
```diff
@@ -7,6 +9,65 @@
...
 
+# The basic unit of a block I/O is a sector. A sector is
+# 512 (2 ** 9) bytes. In sysfs, the files (sector_t type)
+# `<disk>/<part>/start` and `<disk>/<part>/size` are
+# shown as a multiple of 512 bytes. Most user-space
+# tools (fdisk, parted, sfdisk, etc) treat the partition
+# offsets in sectors.
+BYTES_512 = 512
+PART_INFO_FIELDS = (
+    # queue/logical_block_size (reported as a multiple of BYTES_512)
```
themylogin, line 20: If you want default values and documentation, you’ll be better off using a dataclass

themylogin (approved):

PR #13948, opened 2024-06-28
Use pre-defined constant for UID validator

anodos325:
We shouldn't have second definition of TRUENAS_IDMAP_DEFAULT_LOW.

### DIFF
--- a/src/middlewared/middlewared/api/base/types/user.py
+++ b/src/middlewared/middlewared/api/base/types/user.py
@@ -2,11 +2,10 @@ import string
 
 from annotated_types import Ge, Le
 from pydantic.functional_validators import AfterValidator
+from middlewared.plugins.idmap_.idmap_constants import TRUENAS_IDMAP_DEFAULT_LOW
 from typing_extensions import Annotated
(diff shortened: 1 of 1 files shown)


themylogin (changes requested):
Changing it there will retroactively change API definition for already released APIs, that should not be the case.

If it is not a part of API definition, let's validate it in the code, not in the API definitions,

anodos325:
> Changing it there will retroactively change API definition for already released APIs, that should not be the case.
> 
> If it is not a part of API definition, let's validate it in the code, not in the API definitions,

Obviously if someone is changing constants they'll need to be aware of its impact on other areas of the product. I don't see any reason to legitimately allow different ID ranges based on which endpoint an API consumer chooses to use.

themylogin:
> if someone is changing constants they'll need to be aware of its impact on other areas of the product.

We took an obligation to have backwards-compatible API. One of the methods we decided to use in order to comply with it is to never import anything in the API definition.

This value should probably be validated in the create/update method itself, not in the API schema.

PR #13955, opened 2024-07-01
NAS-129286 / 24.10 / Do not try to dump result when it is a job

Qubad786:
## Problem
Currently, `call_with_audit` is not working for APIs that have a return type defined but are jobs. This happens because `call_with_audit` tries to call `schema.dump` for the result, but instead of getting a dict, it gets a job object, which causes an error.

## Solution
Dump the result only if the API has a return type and it is not a job.

### DIFF
--- a/src/middlewared/middlewared/main.py
+++ b/src/middlewared/middlewared/main.py
@@ -1438,6 +1438,9 @@ class Middleware(LoadPluginsMixin, ServiceCallMixin):
                 for i, arg in enumerate(args)]
 
     def dump_result(self, method, result, expose_secrets):
+        if isinstance(result, Job):
+            return result
+
         if hasattr(method, "new_style_returns"):
             return serialize_result(method.new_style_returns, result, expose_secrets)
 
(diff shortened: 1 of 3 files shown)


anodos325:
We dump to auto-redact private information for READONLY users. What is impact on READONLY admins?

src/middlewared/middlewared/main.py
```diff
@@ -1441,7 +1441,7 @@ def dump_result(self, method, result, expose_secrets):
         if hasattr(method, "new_style_returns"):
             return serialize_result(method.new_style_returns, result, expose_secrets)
 
-        if not expose_secrets and hasattr(method, "returns") and method.returns:
+        if not expose_secrets and not isinstance(result, Job) and hasattr(method, "returns") and method.returns:
```
themylogin, line 1444: This will continue failing for new-style returns, please move this check above

themylogin (changes requested):

themylogin (approved):

PR #13956, opened 2024-07-01
NAS-129790 / 24.10 / Fix `cloudsync.credentials.verify`

themylogin:


themylogin:
time 1:00

src/middlewared/middlewared/api/base/decorator.py
```diff
@@ -2,12 +2,20 @@
...
 
 __all__ = ["api_method"]
 
 
-def api_method(accepts, returns, audit=None, audit_callback=False, audit_extended=None, roles=None):
+def api_method(
+    accepts: type[BaseModel],
+    returns: type[BaseModel],
+    audit: str | None = None,
+    audit_callback=False,
```
sonicaj, line 15: Let's add type hints here as well as args before/after have them

tests/api2/test_130_cloudsync.py
```diff
@@ -1,11 +1,9 @@
-#!/usr/bin/env python3
-
 import pytest
 import sys
 import os
 import time
 import urllib.parse
-from pytest_dependency import depends
+
```
yocalebo, line 6: We should opportunistically clean this test file up and convert to websocket.

themylogin: RIght now?

yocalebo: Yeah, I don't see why we shouldn't. We're doing this in other PRs as well.

src/middlewared/middlewared/api/base/decorator.py
themylogin: `audit` has a boolean default which implies its type, and `audit_callback` can't have a type hint, because it's a callable with non-specific arguments

yocalebo (approved):
Awesome, thanks :+1:

PR #13957, opened 2024-07-01
NAS-129754 / 24.10 / Update `api_key.py` to use the new API style

creatorcary:
The goal of this PR is to continue replacing the old `@accepts` and `@returns` decorators with the new `@api_method` decorator which enforces types using Pydantic. The eventual end goal is to entirely replace the old style.

Please note that the type hints added will work with the Pydantic validation and with the `@api_method` decorator, but they are not entirely truthful. This is because I have preserved our use of dictionaries for now until further discussion can be had about whether to completely move to using Pydantic models instead of dictionaries. This will enforce greater type control and allow for nicer syntax (e.g. `data.name` instead of `data["name"]`, but it may come at the cost of sometimes needing to convert between dictionaries and Pydantic models.

### DIFF
--- a/src/middlewared/middlewared/api/base/excluded.py
+++ b/src/middlewared/middlewared/api/base/excluded.py
@@ -9,7 +9,7 @@ from middlewared.utils.lang import undefined
 __all__ = ["Excluded", "excluded_field"]
 
 
-class ExcludedField(Any):
+class ExcludedField:
     @classmethod
     def __get_pydantic_core_schema__(
         cls, source_type: Any, handler: GetCoreSchemaHandler
--- a/src/middlewared/middlewared/api/base/excluded.py
+++ b/src/middlewared/middlewared/api/base/excluded.py
@@ -9,7 +9,7 @@ from middlewared.utils.lang import undefined
 __all__ = ["Excluded", "excluded_field"]
 
 
-class ExcludedField:
+class ExcludedField(Any):
     @classmethod
     def __get_pydantic_core_schema__(
         cls, source_type: Any, handler: GetCoreSchemaHandler
--- a/src/middlewared/middlewared/api/v25_04_0/api_key.py
+++ b/src/middlewared/middlewared/api/v25_04_0/api_key.py
@@ -1,43 +1,48 @@
 from datetime import datetime
-from typing import Literal
+from typing import Literal, TypeAlias
+from typing_extensions import Annotated
 
-from middlewared.api.base import BaseModel, NonEmptyString
+from pydantic import ConfigDict, StringConstraints
 
-__all__ = [
-    "ApiKeyCreateArgs",
-    "ApiKeyCreateResult",
-    "ApiKeyUpdateArgs",
-    "ApiKeyUpdateResult",
-    "ApiKeyDeleteArgs",
-    "ApiKeyDeleteResult",
-]
+from middlewared.api.base import BaseModel, Excluded, excluded_field, NonEmptyString, Private
+
+
+HttpVerb: TypeAlias = Literal["GET", "POST", "PUT", "DELETE", "CALL", "SUBSCRIBE", "*"]
 
 
 class AllowListItem(BaseModel):
-    method: Literal["GET", "POST", "PUT", "DELETE", "CALL", "SUBSCRIBE", "*"]
+    method: HttpVerb
     resource: NonEmptyString
 
 
-class ApiKeyCreate(BaseModel):
-    name: NonEmptyString
+class ApiKeyEntry(BaseModel):
+    """Represents a record in the account.api_key table."""
+    #: This allows the model to be created from an instance of plugins/api_key.APIKeyModel
+    model_config = ConfigDict(from_attributes=True)
+
+    id: int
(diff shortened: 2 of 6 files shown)


src/middlewared/middlewared/api/base/excluded.py
```diff
@@ -9,7 +9,7 @@
 __all__ = ["Excluded", "excluded_field"]
 
 
-class ExcludedField(Any):
+class ExcludedField:
```
creatorcary, line 12: Attempting to subclass `typing.Any` throws a `TypeError`. I do not believe this needs to derive from anything in order to do its job, but correct me if I'm wrong.

src/middlewared/middlewared/plugins/api_key.py
```diff
@@ -183,12 +163,12 @@ async def _validate(self, schema_name, data, id_=None):
     def _generate(self):
         return "".join([random.SystemRandom().choice(string.ascii_letters + string.digits) for _ in range(64)])
 
-    def _serve(self, data, key):
+    def _serve(self, data: ApiKeyEntry, key: str | None) -> ApiKeyEntry:
         if key is None:
             return data
 
         return dict(data, key=f"{data['id']}-{key}")
```
themylogin, line 170: Is `data` a `dict` or an `ApiKeyEntry`?

themylogin: I would propose sticking to dictionaries at the current stage and only changing API declarations. Thoughts, @yocalebo ?

No need to revert already made changes in any case; however, let's make type hints reflect the reality. If the function accepts a dict, let it be a dict.

src/middlewared/middlewared/api/base/excluded.py
themylogin: Unit tests pass, it seems to be working fine.

src/middlewared/middlewared/plugins/api_key.py
creatorcary: Sounds good, I will fix this.

src/middlewared/middlewared/api/base/excluded.py
creatorcary: Just double checked and you're right, I was using Python 3.10. Python 3.11 does not raise a TypeError.

src/middlewared/middlewared/utils/allowlist.py
```diff
@@ -19,10 +21,10 @@ def __init__(self, allowlist):
                 self.exact.setdefault(method, set())
                 self.exact[method].add(resource)
 
-    def authorize(self, method, resource):
+    def authorize(self, method: HttpVerb, resource: NonEmptyString):
```
themylogin, line 24: Do we need to use annotated types here? Won't `str` be enough (for both)? I doubt this will work good with many IDEs and other tools

src/middlewared/middlewared/api/v25_04_0/api_key.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class AllowListItem(BaseModel):
+    method: HttpVerb
+    resource: NonEmptyString
+
+
+class ApiKeyEntry(BaseModel):
+    """Represents a record in the account.api_key table."""
+    #: This allows the model to be created from an instance of plugins/api_key.APIKeyModel
```
themylogin, line 20: Sorry, what does this do, exactly? We don't have instances of `APIKeyModel` anywhere, we don't use SQLAlchemy ORM

themylogin (changes requested):

src/middlewared/middlewared/api/v25_04_0/api_key.py
creatorcary: It just allows you to validate and create an `ApiKeyEntry` from an `APIKeyModel`. I only added this as an option because I noticed it matches the structure of the ORM model we already have. I can remove if it's unnecessary.

creatorcary: https://docs.pydantic.dev/latest/concepts/models/#arbitrary-class-instances

themylogin: We don't have any instances of SQLalchemy ORM models and we never will, so this feature won't be used and can be removed.

src/middlewared/middlewared/utils/allowlist.py
creatorcary: Never _need_ to use type hints, but I think here we should. I would argue that type hints are most helpful when they're as specific as possible. IDEs/linters resolve type aliases the same way they would any other user-defined type. The default type checker in VS Code flags anything that isn't listed in the definition of `HttpVerb` if I try to annotate as being an `HttpVerb`.

themylogin: @creatorcary agreed with HttpVerb if it works, but NonEmptyString seems excessive

creatorcary: Yeah that's fair.

themylogin (approved):
Perfect, thank you!

PR #13970, opened 2024-07-03
NAS-129798 / 24.10 / New plugin that checks for the vendor sentinel file

creatorcary:
This new vendor plugin will be used in the soon upcoming [NAS-129798](https://ixsystems.atlassian.net/browse/NAS-129797).

Its job is simply to check for the existence of the vendor sentinel file /data/.vendor and return its contents, the vendor name.

I am also proposing two new common API models, `NoArgs` and `NoneReturn`.

### DIFF
--- a/src/middlewared/middlewared/plugins/system/vendor/vendor.py
+++ b/src/middlewared/middlewared/plugins/system/vendor/vendor.py
@@ -1,6 +1,8 @@
 import os
 
-from middlewared.service import private, Service
+from middlewared.api import api_method
+from middlewared.api.current import NoArgs, NoneReturn, VendorNameResult
+from middlewared.service import Service
 
 
 SENTINEL_FILE_PATH = '/data/.vendor'
@@ -8,7 +10,7 @@ SENTINEL_FILE_PATH = '/data/.vendor'
 
 class VendorService(Service):
 
-    @private
+    @api_method(NoArgs, VendorNameResult, private=True)
     def name(self) -> str | None:
         try:
             with open(SENTINEL_FILE_PATH, 'r') as file:
@@ -17,7 +19,7 @@ class VendorService(Service):
         except FileNotFoundError:
             return None
 
-    @private
+    @api_method(NoArgs, NoneReturn, private=True)
     def unvendor(self):
         try:
             os.remove(SENTINEL_FILE_PATH)
--- a/src/middlewared/middlewared/plugins/system/vendor/vendor.py
+++ b/src/middlewared/middlewared/plugins/system/vendor/vendor.py
@@ -1,3 +1,4 @@
+import json
 import os
 
 from middlewared.api import api_method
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/system/vendor/vendor.py
```diff
@@ -0,0 +1,27 @@
...
+
+SENTINEL_FILE_PATH = '/data/.vendor'
+
+
+class VendorService(Service):
+
+    @api_method(NoArgs, VendorNameResult, private=True)
+    def name(self) -> str | None:
+        try:
+            with open(SENTINEL_FILE_PATH, 'r') as file:
```
yocalebo, line 16: Get with @aiden3c because this file will be json formatted so we can't just read it outright. We'll need to de-serialize it.

themylogin: Is this a mutable file? Should we cache this method?

src/middlewared/middlewared/plugins/system/vendor/vendor.py
```diff
@@ -0,0 +1,27 @@
...
+                if contents := file.read():
+                    return contents
+        except FileNotFoundError:
+            return None
+
+    @api_method(NoArgs, NoneReturn, private=True)
```
themylogin, line 22: We can’t use this approach. If we decide to add arguments or return value later, changing model names here will have a retroactive effect on already released API. Unfortunately, we’ll have to be very repetitive and declare empty arguments and empty result for each method that accepts/returns them separately.

themylogin (changes requested):

src/middlewared/middlewared/plugins/system/vendor/vendor.py
creatorcary: That is unfortunate but I get your point.

src/middlewared/middlewared/plugins/system/vendor/vendor.py
creatorcary: > Is this a mutable file? Should we cache this method?

I'm going to guess that it is mutable and should not be cached given the `unvendor` method removes the file.

src/middlewared/middlewared/plugins/system/vendor/vendor.py
```diff
@@ -0,0 +1,28 @@
...
+
+class VendorService(Service):
+
+    @api_method(VendorNameArgs, VendorNameResult, private=True)
+    def name(self) -> str | None:
+        try:
+            with open(SENTINEL_FILE_PATH, 'r') as file:
+                if contents := file.read():
+                    return json.loads(contents).get('name')
+        except FileNotFoundError:
```
yocalebo, line 20: You beat me to it but let's catch the `JSONDecodeError` as well.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/system/vendor/vendor.py
yocalebo: You can do `except FileNotFoundError json.JSONDecodeError)`

creatorcary: Wasn't sure if it would be desirable to.

yocalebo (approved):

aiden3c (approved):

PR #13972, opened 2024-07-03
NAS-129889 / 24.10 / Fix schema for directory services users

anodos325:
We impose more restrictive validation on our local usernames than NSS on Linux imposes. nss_winbind for instance uses \ as a separator for domain name and username to prevent collisions with local users.

This commit fixes our validator for UserEntry but maintains for UserCreate and fixes various minor issues with schema returned by directory services cache.


src/middlewared/middlewared/api/base/types/user.py
```diff
@@ -4,26 +4,47 @@
...
+        assert val_len <= max_length, f'Username cannot exceed {max_length} charaters in length'
+    if valid_start_chars is not None:
+        assert val[0] in valid_start_chars, 'Username must start with a letter or an underscore'
+
     assert '$' not in val or val[-1] == '$', 'Username must end with a dollar sign character'
     assert all(char in valid_chars for char in val), f'Valid characters for a username are: {", ".join(valid_chars)!r}'
     return val
 
 
+def validate_local_username(val):
```
yocalebo, line 34: might as well type annotate these 2 functions

yocalebo (approved):
Really nice

PR #13973, opened 2024-07-03
NAS-129230 / 24.10 / Fix progress reporting in formatting disks and it's related usages

Qubad786:
## Problem

`pool.format_disks` was being used where it expected job as input and it updated progress without discriminating from where it was being called which would result in incorrect job progress reporting.

## Solution

Make sure we properly report job progress reporting for endpoints which consume `pool.format_disks`

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/format_disks.py
+++ b/src/middlewared/middlewared/plugins/pool_/format_disks.py
@@ -1,5 +1,3 @@
-import asyncio
-
 from middlewared.service import private, Service
 from middlewared.utils.asyncio_ import asyncio_map
 
@@ -15,7 +13,6 @@ class PoolService(Service):
 
         formatted = 0
         len_disks = len(disks)
-        percentage_lock = asyncio.Lock()
         current_percentage = base_percentage
         single_disk_percentage = (upper_percentage - base_percentage) / len_disks
 
@@ -23,10 +20,9 @@ class PoolService(Service):
             nonlocal formatted, current_percentage
             disk, config = arg
             await self.middleware.call('disk.format', disk)
-            async with percentage_lock:
-                formatted += 1
-                current_percentage += single_disk_percentage
-                job.set_progress(current_percentage, f'Formatting disks ({formatted}/{len_disks})')
+            formatted += 1
+            current_percentage += single_disk_percentage
+            job.set_progress(current_percentage, f'Formatting disks ({formatted}/{len_disks})')
 
         await asyncio_map(format_disk, disks.items(), limit=16)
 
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/pool_/format_disks.py
```diff
@@ -1,29 +1,37 @@
...
+        single_disk_percentage = (upper_percentage - base_percentage) / len_disks
+
         async def format_disk(arg):
-            nonlocal formatted
+            nonlocal formatted, current_percentage
             disk, config = arg
             await self.middleware.call('disk.format', disk)
-            formatted += 1
-            job.set_progress(15, f'Formatting disks ({formatted}/{len_disks})')
+            async with percentage_lock:
```
yocalebo, line 26: Why do we need to grab a lock? Have you investigated the performance impact on a system when a zpool is created with 100's of disks?

yocalebo (changes requested):

src/middlewared/middlewared/plugins/pool_/format_disks.py
Qubad786: We wouldn't want this to be executed asynchronously ?
I have not tested it on a pool with 100s of disks but the penalty should be quite low considering no cpu intensive task is being done - should i go ahead and test it there too?

yocalebo: The penalty is probably more than you expect. It means all async tasks that are formatting a disk now share a global lock.

Furthermore, the `nonlocal formatted` variable was being updated without a lock before your changes so I'm not convinced a lock is needed for percentage.

Qubad786: All tasks won't share the lock i.e like if a pool is being created and another is being extended, they would have different locks - however i am trying to simulate failure scenario - i think it would be hard to hit it, i'll update the PR after running some short tests locally and remove the lock ✅

yocalebo: That's not what I'm saying. I'm saying that all disks being formatted or extended will now share the same lock. So if you create a pool with 250 disks, the format job will spawn 16 tasks at a time that share the same lock. Then the next 16 run sharing the same lock, etc.

themylogin: This lock is unnecessary, `formatted` variable will never be accessed concurrently as it is incremented within an event loop, and we have only one event loop

Qubad786: Makes sense, making that update, thanks.

yocalebo (approved):

PR #13989, opened 2024-07-08
NAS-129807 / 24.10 / Fix nut reporting in netdata

Qubad786:
## Problem

Currently, the Netdata user cannot access the `/run/nut/upsmon.pid` file, causing it to return dummy UPS data even after UPS is started. Furthermore, the dummy UPS data is causing problems in the Netdata UI because users can see the dummy UPS when using netdata webui, which doesn't actually exist. Additionally, there is an issue with the UPS graphs showing spikes instead of a straight line. This is due to the UPS chart.d plugin sleeping for 60 seconds, and the Netdata natural points flag not working correctly.

## Solution

Refactor the UPS plugin to remove the need for dummy UPS data by not sending any metrics if the UPS is not running. This will keep the UPS plugin running without errors and it will start working normally once the UPS starts. Also, fix `/run/nut` permissions to make it accessible to the Netdata user. Restore the default timeout because checking `/run/nut/upsmon.pid` file existence is not an expensive operation.

### DIFF
--- a/src/middlewared/middlewared/etc_files/local/nut/ups_config.py
+++ b/src/middlewared/middlewared/etc_files/local/nut/ups_config.py
@@ -16,7 +16,7 @@ def generate_ups_config(middleware):
 
     ups_group = middleware.call_sync('group.query', [['group', '=', UPS_USER]], {'get': True})
     os.chown(UPS_VARPATH, 0, ups_group['gid'])
-    os.chmod(UPS_VARPATH, 0o770)
+    os.chmod(UPS_VARPATH, 0o775)
 
 
 def render(service, middleware):
--- a/src/middlewared/middlewared/etc_files/local/nut/ups_config.py
+++ b/src/middlewared/middlewared/etc_files/local/nut/ups_config.py
@@ -5,6 +5,7 @@ import shutil
 UPS_CONFPATH = '/etc/nut'
 UPS_USER = 'nut'
 UPS_VARPATH = '/var/run/nut'
+UPSSCHED_VARPATH = '/var/run/nut/private'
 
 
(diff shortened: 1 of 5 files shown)


yocalebo (approved):

aiden3c (approved):

src/middlewared/middlewared/etc_files/local/nut/ups_config.py
```diff
@@ -16,7 +16,7 @@ def generate_ups_config(middleware):
 
     ups_group = middleware.call_sync('group.query', [['group', '=', UPS_USER]], {'get': True})
     os.chown(UPS_VARPATH, 0, ups_group['gid'])
-    os.chmod(UPS_VARPATH, 0o770)
+    os.chmod(UPS_VARPATH, 0o775)
```
anodos325, line 19: What is impact of making this world-readable? Have you checked for what other files nut may write there?

Qubad786: @anodos325 we have `upssched.pipe` there apart from the pid file, i have adjusted its perms to be 660 and we should be good hopefully..testing the changes to make sure there are no regressions from this change

Qubad786: @anodos325 my latest changes are good, confirmed after testing ✅

anodos325: Maybe we should configure a separate rundir for upssched? It looks configurable via PIPEFN / LOCKFN.

anodos325 (approved):

PR #13995, opened 2024-07-09
NAS-129973 / 24.10 / More verbosely indicate unexpected test state in SMB simple share test

anodos325:
This SMB test may fail if other tests fail to properly clean up temporary users. We should report the usernames so that it's easier to track down offending tests.


yocalebo (approved):
Should be resolved in https://github.com/truenas/middleware/pull/13994 but this will help in the unexpected event something else dies :+1:

PR #13999, opened 2024-07-09
NAS-129776 / 24.10 / Fix iSCSI ALUA LUN removal issue

bmeagherix:
When a LUN is removed from an ALUA target, we were already calling `iscsi.alua.removed_target_extent` to perform some fixup.  However, because this was being called on the STANDBY **before** the MASTER node was updated the LUN was still available and logged into by the STANDBY node.  Simply reordering _avoids_ the issue.

(For robustness, this PR also adds a check/delay to ensure that the LUN is removed from the MASTER before `removed_target_extent` is called on the STANDBY.)

Add CI tests to check proper operation wrt adding and removing LUNs from an existing ALUA target.  Separated these tests out into a new file (_test_262_iscsi_alua.py_) and structured to ease adding additional ALUA tests (e.g. to `TestFixtureConfiguredALUA`)

Passing CI test [here](http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/1327/).

### DIFF
--- a/src/middlewared/middlewared/plugins/iscsi_/target_to_extent.py
+++ b/src/middlewared/middlewared/plugins/iscsi_/target_to_extent.py
@@ -159,12 +159,16 @@ class iSCSITargetToExtentService(CRUDService):
             # Check that the HA target is no longer offering the LUN that we just deleted.  Wait a short period
             # if necessary (though this should not be required).
             retries = 5
+            lun_removed = False
             iqn = await self.middleware.call('iscsi.target.ha_iqn', target_name)
             while retries:
                 if associated_target['lunid'] not in await self.middleware.call('iscsi.target.iqn_ha_luns', iqn):
+                    lun_removed = True
                     break
                 retries -= 1
                 await asyncio.sleep(1)
+            if not lun_removed:
+                self.logger.warning('Failed to remove lun %r of target %r from internal target', associated_target['lunid'], target_name, exc_info=True)
 
             try:
                 # iscsi.alua.removed_target_extent includes a local service reload
--- a/src/middlewared/middlewared/plugins/iscsi_/target_to_extent.py
+++ b/src/middlewared/middlewared/plugins/iscsi_/target_to_extent.py
@@ -168,7 +168,7 @@ class iSCSITargetToExtentService(CRUDService):
                 retries -= 1
                 await asyncio.sleep(1)
             if not lun_removed:
-                self.logger.warning('Failed to remove lun %r of target %r from internal target', associated_target['lunid'], target_name, exc_info=True)
+                self.logger.warning('Failed to remove lun %r from internal target %r', associated_target['lunid'], iqn, exc_info=True)
 
             try:
                 # iscsi.alua.removed_target_extent includes a local service reload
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/iscsi_/target_to_extent.py
```diff
@@ -141,13 +143,29 @@ async def do_delete(self, audit_callback, id_, force):
...
                                                       {'select': ['name'], 'get': True}))['name']
             extent_name = (await self.middleware.call('iscsi.extent.query',
                                                       [['id', '=', associated_target['extent']]],
                                                       {'select': ['name'], 'get': True}))['name']
+
+            # Check that the HA target is no longer offering the LUN that we just deleted.  Wait a short period
+            # if necessary (though this should not be required).
+            retries = 5
+            iqn = await self.middleware.call('iscsi.target.ha_iqn', target_name)
+            while retries:
```
mgrimesix, line 163: Should we log the case where we've exhausted the retries?

mgrimesix (approved):
Approve with comment.

src/middlewared/middlewared/plugins/iscsi_/target_to_extent.py
```diff
@@ -1,7 +1,9 @@
-import middlewared.sqlalchemy as sa
+import asyncio
 
-from middlewared.schema import accepts, Bool, Dict, Int, Patch
-from middlewared.service import CallError, CRUDService, private, ValidationErrors
+import middlewared.sqlalchemy as sa
+from middlewared.schema import Bool, Dict, Int, Patch, accepts
+from middlewared.service import (CallError, CRUDService, ValidationErrors,
```
yocalebo, line 5: Can you fix the import? I'm not sure if it's your IDE, but our flake8 checks are 120 characters on a single line which also includes imports. No reason to wrap this.

tests/api2/test_262_iscsi_alua.py
```diff
@@ -0,0 +1,149 @@
...
+                                    verify_luns)
+from assets.websocket.service import ensure_service_enabled
+from middlewared.test.integration.utils import call
+from middlewared.test.integration.utils.client import truenas_server
+
+from auto_config import ha, pool_name
+from protocols import iscsi_scsi_connection
+
+if not ha:
+    pytest.skip("skipping ALUA tests", allow_module_level=True)
```
yocalebo, line 22: Instead of checking `if not ha:` you can do `pytestmark = pytest.mark.skipif(not ha, reason='reason here')`

tests/api2/test_262_iscsi_alua.py
```diff
@@ -0,0 +1,149 @@
+#!/usr/bin/env python3
```
yocalebo, line 1: Is there a reason to name this test module `test_262`? Does this test module have dependencies on `test_261`?

bmeagherix: No reason really.  I just liked the idea of it running after 261 (which does more basic tests).

yocalebo: okay, I just wanted to make sure that we aren't adding dependencies between test modules :+1:

bmeagherix:
backport

PR #14009, opened 2024-07-11
NAS-129799 / 24.10 / Add service to start websocat on HexOS builds

creatorcary:
**The websocat binary should be merged first** (needs approval): https://github.com/truenas/binaries/pull/7

The `vendor_service` script checks if the vendor is "HexOS" and starts a transient websocat service using the arguments specified by the vendor.

Passing test: http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/1403/testReport/api2/test_system_vendor/


src/middlewared/middlewared/scripts/vendor_service.py
```diff
@@ -0,0 +1,62 @@
...
+    except:
+        pass
+
+
+def main():
+    with Client() as client:
+        vendor_name = client.call("system.vendor.name")
+        if vendor_name == "HexOS":
+            url = "wss://api.hexos.com"
+            systemd_opts = (
```
Qubad786, line 21: Should there be a `Restart` parameter as well so that it is restarted if it fails for some reason?

debian/debian/ix-vendor.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+After=middlewared.service
+
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+TimeoutStartSec=1200
```
Qubad786, line 8: I think this should be smaller so it does not stop the boot process if something goes south and also probably it will never take this long?

debian/debian/ix-vendor.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+After=middlewared.service
```
Qubad786, line 2: @yocalebo do we want to run this after we consider the system to be in ready state?

tests/api2/test_system_vendor.py
```diff
@@ -0,0 +1,14 @@
+from middlewared.test.integration.utils import call, ssh
+
+
+SENTINEL_FILE_PATH = "/data/.vendor"
+
+
+def test_name_is_none():
+    # /data/.vendor should not exist
+    output = ssh(f"test -e {SENTINEL_FILE_PATH} && echo {SENTINEL_FILE_PATH}", check=False)
```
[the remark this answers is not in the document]
creatorcary: Thanks, will look into it.

src/middlewared/middlewared/plugins/system_vendor/vendor.py
```diff
@@ -11,13 +11,16 @@
...
+        namespace = 'system.vendor'
+        cli_private = True
+
     @api_method(VendorNameArgs, VendorNameResult, private=True)
     def name(self) -> str | None:
         try:
             with open(SENTINEL_FILE_PATH, 'r') as file:
-                if contents := file.read():
-                    return json.loads(contents).get('name')
-        except (FileNotFoundError, json.JSONDecodeError):
```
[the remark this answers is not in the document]
creatorcary: This was left over from my previous pr. It was originally there because there was going to be some logic if the file was empty, but it was decided to never raise any exceptions in `name`. My change cuts out the extra `read()` step.

debian/debian/ix-vendor.service
creatorcary: This was just copied from another unrelated service file so I'll look into this too.

src/middlewared/middlewared/plugins/system_vendor/vendor.py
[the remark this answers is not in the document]
Qubad786: So it is always better that we have explicit exceptions mentioned in the except clause and not leave it broad imho, so whichever exceptions you think can be raised i think can be listed down here

src/middlewared/middlewared/scripts/vendor_service.py
```diff
@@ -0,0 +1,75 @@
+import hashlib
+import os
+import subprocess
+import sys
+
+from truenas_api_client import Client
+
+
+def load_envvars(envvars_file: str):
```
Qubad786, line 9: We already have `dotenv` available in scale, so you can do something like this
```
>>> from dotenv import load_dotenv
>>> load_dotenv('/root/.env')
True
>>>
```

Qubad786: Also if the transient service should have this env, then you can also specify it to systemd to load environment from specific file

creatorcary: Unfortunately I couldn't find a `systemd-run` option equivalent to setting the `EnvironmentFile` property in a service file, and I don't think `--property=EnvironmentFile=` works. I'll look into `dotenv`.

src/middlewared/middlewared/scripts/vendor_service.py
```diff
@@ -0,0 +1,69 @@
...
+                "-c",
+                " ".join([wsocat_path, *wsocat_opts, local_server, remote_server])
+            ])
+
+
+if __name__ == "__main__":
+    try:
+        main()
+    finally:
+        sys.exit(0)  # Never fail
```
aiden3c, line 69: Why do we sys.exit here?

creatorcary: I am copying the `wait_on_disks` script since both should never fail. https://github.com/truenas/middleware/blob/master/src/middlewared/middlewared/scripts/wait_on_disks.py

aiden3c (approved):
LGTM!

src/middlewared/middlewared/scripts/vendor_service.py
```diff
@@ -0,0 +1,76 @@
...
+        vendor_name = client.call("system.vendor.name")
+
+    if vendor_name == "HexOS":
+        start_hexos_websocat()
+
+
+if __name__ == "__main__":
+    try:
+        main()
+    finally:
```
Qubad786, line 75: Probably better to catch any exception which might be and log it and just safely exit which would ensure zero status code rather then doing an explicit sys.exit?

src/middlewared/middlewared/plugins/system_vendor/vendor.py
```diff
@@ -11,18 +11,25 @@
...
         try:
             with open(SENTINEL_FILE_PATH, 'r') as file:
-                if contents := file.read():
-                    return json.loads(contents).get('name')
+                return json.load(file).get('name') or None  # Don't return an empty string.
         except (FileNotFoundError, json.JSONDecodeError):
-            return None
+            pass
+        except Exception:
+            self.logger.error('Unexpected error while reading %r', SENTINEL_FILE_PATH, exc_info=True)
```
Qubad786, line 26: If we don't want to log anything in the case the file is not there, L23 is good but if the file exists - i think we should log if the json inside is malformed?
Also what other exceptions are we looking at here? (idea is to avoid using broad except clause where possible)

src/middlewared/middlewared/plugins/system_vendor/vendor.py
```diff
@@ -11,18 +11,25 @@
...
 
     @api_method(UnvendorArgs, UnvendorResult, private=True)
     def unvendor(self):
         try:
             os.remove(SENTINEL_FILE_PATH)
         except FileNotFoundError:
-            return None
+            pass
+        except Exception:
```
Qubad786, line 34: Maybe better to catch `OSError` along with the one we are already handling and just leave it at that as highly unlikely something else will come here?

creatorcary: The theory behind this unsatisfying exception handling came from a discussion with @yocalebo . The plugin should _never_ raise any exceptions. The safest way to guarantee this is to catch everything. However it was decided to separate the "expected" exceptions (e.g. `FileNotFoundError`) from the highly unlikely catch-all.

src/middlewared/middlewared/scripts/vendor_service.py
creatorcary: I don't like it either, just trying to follow precedent. But since @aiden3c already flagged this I'll change it.

src/middlewared/middlewared/plugins/system_vendor/vendor.py
Qubad786: Sure, if @yocalebo agrees we are good 👍 My perception is that in such cases we should always try to catch all the possible exceptions which can arise and mark them appropriately - so if by any chance something out of the blue arises and it actually needs to be fixed, it would never be raised because we were catching everything

Qubad786: Also if we were doing something really complex which could result in many exceptions being raised, then probably yes i would prefer using base exception but still log it in that case so we know something went south and if it needs to be fixed we can have it fixed

src/middlewared/middlewared/plugins/system_vendor/vendor.py
yocalebo: That's a good point. Yes, you're correct

src/middlewared/middlewared/scripts/vendor_service.py
```diff
@@ -0,0 +1,93 @@
...
+
+from middlewared.plugins.system_vendor.vendor import SENTINEL_FILE_PATH
+from truenas_api_client import Client
+
+
+def get_hostid() -> str | None:
+    try:
+        with open("/etc/hostid", "rb") as f:
+            return hashlib.file_digest(f, "sha256").hexdigest()
+    except FileNotFoundError:
```
Qubad786, line 17: This is redundant and just `except Exception` would suffice ?

debian/debian/ix-vendor.service
Qubad786: Considering we are waiting below for the system to be in ready state - the dependency here should change as well to `ix-postinit`

Qubad786 (commented):
Also `ix-vendor` entry should be made to `debian/debian/rules` file so it actually gets pulled in

yocalebo (approved):

PR #14015, opened 2024-07-12
NAS-130015 / 24.10 / Fix failing SMB CI tests

anodos325:
This commit cleans up some of the SMB CI tests and fixes a few issues:

* Rapid restarts of SMB service was hitting systemd service restart limit causing spurious test failures

* Dynamic generation of nsswitch.conf means that Owner Rights group will resolve via NSS and so the ACL inheritance test needs to use gid instead of name

* Convert a lot of test cases to use pytest fixtures to properly manage test state.

* Fix a few tests related to auxiliary parameters.

### DIFF
--- a/tests/api2/test_420_smb.py
+++ b/tests/api2/test_420_smb.py
@@ -1,7 +1,6 @@
 import pytest
 import sys
 import os
-from pytest_dependency import depends
 import secrets
 import string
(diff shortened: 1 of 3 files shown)


mgrimesix (approved):
Looks good to me.

PR #14016, opened 2024-07-12
NAS-130002 / 24.10 / Inject vendor in main grub.cfg

aiden3c:
This configures the main grub entry to include the vendor name if supplied

### DIFF
--- a/src/freenas/usr/local/bin/truenas-grub.py
+++ b/src/freenas/usr/local/bin/truenas-grub.py
@@ -2,6 +2,7 @@
 import math
 import psutil
 import os
+import json
 
 from middlewared.utils.serial import serial_port_choices
 from middlewared.utils.db import query_config_table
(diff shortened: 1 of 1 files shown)


src/freenas/usr/local/bin/truenas-grub.py
```diff
@@ -15,6 +16,13 @@ def get_serial_ports():
     advanced = query_config_table("system_advanced", prefix="adv_")
     kernel_extra_options = advanced.get("kernel_extra_options") or ""
 
+    # check for /data/.vendor
+    try:
+        with open("/data/.vendor", "r") as f:
+            vendor = json.loads(f.read()).get("name", "TrueNAS Scale")
+    except FileNotFoundError:
+        vendor = "TrueNAS Scale"
+
```
creatorcary, line 25: Using the vendor plugin this becomes
```
with Client() as c:
    vendor = c.call("system.vendor.name") or "TrueNAS Scale"
```
Not sure if this is the preferred way to do it.

aiden3c: I think that since that would require extra imports, it'd be safe to keep it the way it is reading the JSON. We also run truenas-grub from scale-build, I'm not entirely sure what the context looks like there, so just to be safe there too.

themylogin: We can't use middleware client here, middleware is not running during ISO install

creatorcary (approved):

yocalebo:
please wait for @themylogin final review before merge, thanks!

themylogin (approved):

PR #14020, opened 2024-07-13
NAS-130042 / 24.10 / Add custom exit code for fail() in test runs

anodos325:
There are certain conditions under which we exit pytest early via a `fail` test asset. These usually represent serious errors that require attention, and should have a special indication within the jenkins pipeline (but still gather test results assets). This commit adds a custom return code for cases where we fail tests (EX_SOFTWARE / 70) for the situation. The value was chosen because it is significantly higher than the maximum pytest.ExitCode enum value and should not collide with future returncodes from upstream project.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import os
 import pytest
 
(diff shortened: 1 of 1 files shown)


yocalebo (approved):
That's neat

PR #14021, opened 2024-07-15
NAS-130078 / 24.10 / Add features support for apps in app schema

sonicaj:
This PR restores features support in apps schema so app developer can have a way of retrieving information from the running SCALE instance.

We would be adding more refs but after coordinating with Stavros on what needs to be added so that we are enforcing it on apps side as well (that should be relatively straight forward now considering the framework is in place to support those).

### DIFF
--- a/src/middlewared/middlewared/plugins/apps/schema_validation.py
+++ b/src/middlewared/middlewared/plugins/apps/schema_validation.py
@@ -117,7 +117,7 @@ class AppSchemaService(Service):
             return
 
         if not filter_list(
-                await self.middleware.call('app.certificate_authority_choices'), [['id', '=', value]]
+            await self.middleware.call('app.certificate_authority_choices'), [['id', '=', value]]
         ):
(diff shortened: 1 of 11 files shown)


src/middlewared/middlewared/plugins/apps/schema_validation.py
```diff
@@ -96,3 +104,38 @@ async def validate_question(
...
+
+        if not filter_list(await self.middleware.call('app.certificate_choices'), [['id', '=', value]]):
+            verrors.add(schema_name, 'Unable to locate certificate.')
+
+    async def validate_certificate_authority(self, verrors, value, question, schema_name, app_data):
+        if not value:
+            return
+
+        if not filter_list(
+                await self.middleware.call('app.certificate_authority_choices'), [['id', '=', value]]
```
Qubad786, line 120: Extra indent

Qubad786 (approved):
LGTM, one formatting nit

PR #14023, opened 2024-07-16
NAS-127735 / 24.10 / Add alerts for critical audit service failures

mgrimesix:
### Alerts for Audit service failures.
* Add one-shot alert for failure to start audit backends.
* Add one-shot alert for failure to run audit setup configuration
* Add an audit 'health' check that performs a simple query every 20 mins and generate an alert if the query fails.

Some Flake8 cleanup.

Add CI tests for all three.
The CI test results are [here](http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/1456/testReport/api2/test_audit_alerts/).

### DIFF
--- a/src/middlewared/middlewared/alert/source/audit.py
+++ b/src/middlewared/middlewared/alert/source/audit.py
@@ -40,8 +40,7 @@ class AuditServiceHealthAlertSource(AlertSource):
         try:
             await self.middleware.call(
                 'audit.query', {
-                    "query-filters": [["event", "=", "AUTHENTICATION"]],
-                    "query-options": {"offset": -1}
+                    "query-options": {"count": True}
                 }
             )
         except Exception as e:
--- a/src/middlewared/middlewared/alert/source/audit.py
+++ b/src/middlewared/middlewared/alert/source/audit.py
@@ -1,0 +1,52 @@
+from datetime import timedelta
+import logging
+from middlewared.alert.base import AlertClass, AlertCategory, Alert, AlertLevel, AlertSource, SimpleOneShotAlertClass
+from middlewared.alert.schedule import IntervalSchedule
+
+log = logging.getLogger("audit_check_alertmod")
+
+
+# -------------- OneShot Alerts ------------------
+class AuditBackendSetupAlertClass(AlertClass, SimpleOneShotAlertClass):
+    category = AlertCategory.AUDIT
+    level = AlertLevel.ERROR
+    title = "Audit Service Backend Failed"
+    text = "Audit service failed backend setup: %(service)s. See /var/log/middlewared.log"
+
+
+class AuditSetupAlertClass(AlertClass, SimpleOneShotAlertClass):
+    category = AlertCategory.AUDIT
+    level = AlertLevel.ERROR
+    title = "Audit Service Setup Failed"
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/alert/source/audit.py
```diff
@@ -0,0 +1,52 @@
...
+    '''
+    schedule = IntervalSchedule(timedelta(minutes=20))
+    run_on_backup_node = False
+
+    async def check(self):
+        try:
+            await self.middleware.call(
+                'audit.query', {
+                    "query-filters": [["event", "=", "AUTHENTICATION"]],
+                    "query-options": {"offset": -1}
```
anodos325, line 44: Did you test how expensive this is when we have a couple of hundred thousand entries?

anodos325: I think it would be better to short-circuit query with `{"get": true}` rather than specifying offsets (pending tests).

src/middlewared/middlewared/plugins/audit/audit.py
```diff
@@ -558,6 +558,7 @@ async def setup(self):
         try:
             await self.middleware.call('audit.update_audit_dataset', audit_config)
         except Exception:
+            await self.middleware.call('alert.oneshot_create', 'AuditSetup', None)
```
themylogin, line 561: When is this alert removed?

src/middlewared/middlewared/alert/source/audit.py
themylogin: What does negative offset even mean?

tests/api2/test_audit_alerts.py
```diff
@@ -0,0 +1,127 @@
+import pytest
+
+from middlewared.test.integration.utils import call, ssh, mock
+from time import sleep
+
+
+def setup_test(alert_class, alert_key, path):
```
themylogin, line 7: You can turn this into a contextmanager to handle setup and teardown in the same function

src/middlewared/middlewared/plugins/audit/audit.py
mgrimesix: I purposefully left it to be manually dismissed by the customer.

src/middlewared/middlewared/alert/source/audit.py
mgrimesix: Negative offset means: Get the last entry.

src/middlewared/middlewared/plugins/audit/backend.py
```diff
@@ -126,9 +126,12 @@ def setup(self):
         - after audit database deletion or rename
         """
         for svc, conn in self.connections.items():
+            # Dismiss any existing AuditSetup one-shot alerts
+            self.middleware.call_sync('alert.oneshot_delete', 'AuditBackendSetup', {"service": svc})
             try:
                 conn.setup()
             except Exception:
+                self.middleware.call_sync('alert.oneshot_create', 'AuditBackendSetup', {"service": svc})
```
anodos325, line 134: I think maybe we should build list of failures during setup (because there can be more than one) and generate / delete the alert before returning.

anodos325 (approved):

PR #14027, opened 2024-07-16
Prevent permission denied error for non-root csh/tcsh over ssh

danmack:
The .warning file is in root's home directory and thus cannot be read by non-root users on login.

This solution moves it in the earlier block that checks for user root.

Another solution would be to move the warning file out of root's home directory.


yocalebo:
Hi @danmack thanks for the PR but this was made against the wrong branch. `master` is for SCALE only. Pleas re-submit this PR against the `truenas/13.3-stable` repo. Thanks!

PR #14030, opened 2024-07-17
Fix API key API

themylogin:


themylogin:
time 1:30

src/middlewared/middlewared/api/v25_04_0/api_key.py
```diff
@@ -16,15 +16,16 @@ class AllowListItem(BaseModel):
...
     allowlist: list[AllowListItem]
 
 
+class ApiKeyEntryWithKey(ApiKeyEntry):
+    key: Private[str]
+
+
 class ApiKeyCreate(ApiKeyEntry):
     id: Excluded = excluded_field()
     key: Excluded = excluded_field()
```
creatorcary, line 31: Since `ApiKeyEntry` no longer has a `key`, it seems like this line can be removed.

tests/api2/test_api_key_crud.py
```diff
@@ -0,0 +1,49 @@
...
+def api_key(allowlist):
+    key = call("api_key.create", {"name": "Test API Key", "allowlist": allowlist})
+    try:
+        yield key
+    finally:
+        call("api_key.delete", key["id"])
+
+
+def test_has_key_after_creation_but_not_read():
+    api_key = call("api_key.create", {"name": "Test", "allowlist": []})
```
creatorcary, line 16: It's a little confusing to use the name of a function as the name of a local variable imo

anodos325 (approved):

PR #14038, opened 2024-07-18
NAS-130098 / 24.10 / Fix initial cache fill for LDAP and IPA

anodos325:
SSSD will report a domain as online while it is still building internal caches and unable to properly handle NSS requests. This commit adds checks / sleeps in the directory services cache fill method until a getpwent call to the SSS NSS module returns an entry.

### DIFF
--- a/src/middlewared/middlewared/plugins/directoryservices_/cache.py
+++ b/src/middlewared/middlewared/plugins/directoryservices_/cache.py
@@ -178,8 +178,11 @@ class DSCache(Service):
             if client.domain_info()['online']:
                 return
 
-            job.set_progress(10, 'Waiting for domain to come online')
-            self.logger.debug('Waiting for domain to come online')
+            # only log every 10th iteration
+            if waited % 10 == 0:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/directoryservices_/cache.py
```diff
@@ -183,6 +185,26 @@ def idmap_online_check_wait_wbclient(self, job):
...
+        then we can use getpwent call to determine whether the domain is in a state
+        where we can actually fill our caches.
+        """
+        waited = 0
+        while waited <= 60:
+            for pwd in iterpw(module=NssModule.SSS.name):
+                # We have an entry provided by sssd
+                return
+
+            job.set_progress(10, 'Waiting for domain to come online')
```
yocalebo, line 201: Maybe only log every 10 seconds so we don't pollute logs with 60 duplicate log messages?

yocalebo (approved):

bmeagherix (approved):

PR #14049, opened 2024-07-19
NAS-129207 / 24.10 / Utilize NUT dummy-ups to facilitate UPS event CI testing

bmeagherix:
Add the ability to specify `dummy-ups` driver, and add some tests that utilize it.

(Also, first updated the existing tests to use WS client, but didn't otherwise modify them.)

---
Passing CI test is [here](http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/1547/).

### DIFF
--- a/tests/api2/test_530_ups.py
+++ b/tests/api2/test_530_ups.py
@@ -1,9 +1,20 @@
 #!/usr/bin/env python3
 # License: BSD
 
-import pytest
+import os
+from tempfile import NamedTemporaryFile
+from time import sleep
 
-from middlewared.test.integration.utils import call
+import pytest
+from assets.websocket.service import ensure_service_enabled, ensure_service_started
+from auto_config import password, user
+from functions import send_file
+
+from middlewared.test.integration.utils import call, ssh
+from middlewared.test.integration.utils.client import truenas_server
+
+DUMMY_FAKEDATA_DEV = '/tmp/fakedata.dev'
+SHUTDOWN_MARKER_FILE = '/tmp/is_shutdown'
 
 first_ups_list = [
     'rmonitor',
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/ups.py
```diff
@@ -168,7 +169,7 @@ async def validate_data(self, data, schema):
 
         driver = data.get('driver')
         if driver:
-            if driver not in (await self.middleware.call('ups.driver_choices')).keys():
+            if driver not in (await self.middleware.call('ups.driver_choices')).keys() and driver != NUT_DUMMY_UPS:
```
anodos325, line 172: Would it be possible to instead just mock over `ups.driver_choices` in our jenkins CI runs?

tests/api2/test_530_ups.py
```diff
@@ -1,12 +1,20 @@
 #!/usr/bin/env python3
```
yocalebo, line 1: Remove both of these top 2 lines in the test file. They're superfluous.

tests/api2/test_530_ups.py
```diff
@@ -1,12 +1,20 @@
...
-import pytest
 import os
-import sys
-apifolder = os.getcwd()
-sys.path.append(apifolder)
-from functions import GET, PUT, POST
+from tempfile import NamedTemporaryFile
+from time import sleep
+
+import pytest
```
yocalebo, line 8: Space between `import pytest` and `from assets.websocket...`

tests/api2/test_530_ups.py
```diff
@@ -61,38 +136,31 @@ def test_04_Set_UPS_options():
...
     }
-    results = PUT('/ups/', payload)
-    assert results.status_code == 200, results.text
+    results = call('ups.update', payload)
 
 
 @pytest.mark.parametrize('data', first_ups_list)
 def test_05_look_at_UPS_options_output_of_(data):
-    assert payload[data] == results.json()[data], results.text
+    assert payload[data] == results[data], results.text
```
yocalebo, line 144: Still referencing `results.text`

tests/api2/test_530_ups.py
```diff
@@ -28,25 +36,92 @@
...
+
+def remove_file(filepath):
+    ssh(f'rm {filepath}', check=False)
+
+
+def did_shutdown():
+    return ssh(f'cat {SHUTDOWN_MARKER_FILE}', check=False) == "done\n"
+
+
+def write_fake_data(data={}):
```
yocalebo, line 60: data=None and then inside function do `all_data = default_dummy_data | data or {}`

tests/api2/test_530_ups.py
```diff
@@ -28,25 +36,92 @@
...
+            f.write(f'{k}: {v}\n'.encode('utf-8'))
+        f.flush()
+        os.fchmod(f.fileno(), 0o644)
+        results = send_file(f.name, DUMMY_FAKEDATA_DEV, user, password, truenas_server.ip)
+        assert results['result'], str(results['output'])
+
+
+def wait_for_alert(klass, retries=10):
+    assert retries > 0
+    while retries:
```
yocalebo, line 73: Endless loop. `retries` is never decremented.

tests/api2/test_530_ups.py
```diff
@@ -28,25 +36,92 @@
...
+    call('ups.update', payload)
+    try:
+        yield
+    finally:
+        call('ups.update', old_config)
+        remove_file(SHUTDOWN_MARKER_FILE)
+        remove_file(DUMMY_FAKEDATA_DEV)
+
 
 def test_01_get_ups_service_id():
```
yocalebo, line 110: I know you didn't write this test but we need to improve it while you're here. `global` is a keyword that should be avoided at all costs, even in tests. For example, this function just literally queries the service by name and gets the database id. However, the only other functions that use thisid is in `test_26` and `test_28`. There is no reason for that. We can simply query by service name.

In another test file I added a function that queries the service state instead of writing `service.query [["...` a bunch of times.
```
def get_service_state():
    return call(service.query, [['service', '=', 'ups']], {'get': True})
```

tests/api2/test_530_ups.py
```diff
@@ -28,25 +36,92 @@
...
-    assert results.json()[0]['state'] == 'STOPPED', results.text
-    assert results.json()[0]['enable'] is False, results.text
-    ups_id = results.json()[0]['id']
+    results = call('service.query', [['service', '=', 'ups']], {'get': True})
+    assert results['state'] == 'STOPPED', results
+    assert results['enable'] is False, results
+    ups_id = results['id']
 
 
 def test_02_Enabling_UPS_Service_at_boot():
```
yocalebo, line 118: `test_02_` and `test_03` aren't valuable imo. We need to fix tests like this (there are a bunch in this file). Even more so because you've imported context managers `ensure_service_enabled` and `ensure_service_started`. These types of "tests" can be removed entirely because you've got other areas in the code that will fail if the service isn't enabled/started. If you disagree, then we should combine tests like this into 1 test function.
```
def test_enable_service():
    call('service.update', blah...)
    rv = get_service_state()
    assert rv['enable'], rv

File: tests/api2/test_530_ups.py [hunk 9]
```diff
@@ -28,25 +36,92 @@
...
 
 def test_03_look_if_UPS_service_is_enable():
-    results = GET(f'/service/id/{ups_id}/')
-    assert results.status_code == 200, results.text
-    assert results.json()['enable'] is True, results.text
+    results = call('service.query', [['service', '=', 'ups']], {'get': True})
+    assert results['enable'] is True, results
 
 
 def test_04_Set_UPS_options():
```
yocalebo, line 127: Camel case function names is a paradigm that we don't follow. Please convert them while you're here to all lower case.

tests/api2/test_530_ups.py
```diff
@@ -61,38 +136,31 @@ def test_04_Set_UPS_options():
...
 
 def test_07_look_UPS_service_status_is_running():
-    results = GET(f'/service/id/{ups_id}/')
-    assert results.status_code == 200, results.text
-    assert results.json()['state'] == 'RUNNING', results.text
+    results = call('service.query', [['service', '=', 'ups']], {'get': True})
+    assert results['state'] == 'RUNNING', results
 
 
 def test_08_get_API_reports_UPS_configuration_as_saved():
```
yocalebo, line 156: Another example of a function that uses `global` keyword for no apparent reason. Furthermore, the next few tests that test the `results` object can be condensed into 1 function. We're making multiple `ups.update` calls (which is fine) but we can test all the changes in 1 function.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/ups.py
Qubad786: This should probably do the job https://github.com/truenas/middleware/blob/bf02e80889e1d491023c6e9de58643c149f849e8/src/middlewared/middlewared/test/integration/utils/mock.py#L11

yocalebo (approved):
Brilliant, thanks!

PR #14052, opened 2024-07-20
NAS-130120 / 24.10 / Introduce a migration to allow migrating old apps to new apps

sonicaj:
This PR adds changes to allow migrating old apps to new apps.


yocalebo (approved):
Making a note here so we don't forget but we'll need to add the ability to migrate apps after upgrade as well. This can be done in a separate PR.

PR #14056, opened 2024-07-23
NAS-130170 / 24.10 / Add available node bind ips to questions context

Qubad786:
## Context

`apps_validation` requires available node bind ips available in questions context so it can properly render available bind ips which user can consume with an application.

### DIFF
--- a/src/middlewared/middlewared/plugins/apps/resources.py
+++ b/src/middlewared/middlewared/plugins/apps/resources.py
@@ -1,4 +1,4 @@
-from middlewared.schema import accepts, Int, List, Ref, returns
+from middlewared.schema import accepts, Dict, Int, List, Ref, returns, Str
 from middlewared.service import Service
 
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/apps/resources.py
```diff
@@ -41,3 +41,14 @@ async def used_ports(self):
...
+
+    @accepts()
+    @returns(Dict(Str('ip_choice')))
+    async def ip_choices(self):
+        """
+        Returns IP choices which can be used by applications.
+        """
+        return {
+            ip['address']: ip['address']
+            for ip in await self.middleware.call('interface.ip_in_use', {'static': True, 'any': True})
```
stavros-k, line 53: Is "any": True, the 0.0.0.0?

stavros-k (approved):

src/middlewared/middlewared/plugins/apps/resources.py
Qubad786: Yes, that includes the wildcard ips

PR #14057, opened 2024-07-23
NAS-130172 / 24.10 / Fix portals when querying apps

Qubad786:
## Problem

Portals or app install notes were not being shown when apps were being queried.

## Solution

Portals / app install notes have been added to app metadata which is being saved to optimize querying them and it will be included in query output when querying apps.

### DIFF
--- a/src/middlewared/middlewared/plugins/apps/ix_apps/portals.py
+++ b/src/middlewared/middlewared/plugins/apps/ix_apps/portals.py
@@ -1,0 +1,44 @@
+import contextlib
+
+import yaml
+
+from apps_validation.portals import IX_NOTES_KEY, IX_PORTAL_KEY, validate_portals_and_notes, ValidationErrors
+
+from .lifecycle import get_rendered_templates_of_app
+
+
+def normalized_port_value(scheme: str, port: int) -> str:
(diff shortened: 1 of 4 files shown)


stavros-k (approved):

src/middlewared/middlewared/plugins/apps/ix_apps/portals.py
```diff
@@ -0,0 +1,44 @@
...
+    if portal_and_notes_config:
+        try:
+            validate_portals_and_notes('portal', portal_and_notes_config)
+        except ValidationErrors:
+            return config
+
+    portals = {}
+    for portal in portal_and_notes_config.get(IX_PORTAL_KEY, []):
+        port_value = normalized_port_value(portal['scheme'], portal['port'])
+        portals[portal['name']] = f'{portal["scheme"]}://{portal["host"]}{port_value}{portal.get("path", "")}'
```
sonicaj, line 39: Considering ip choices, for wildcard this could be `0.0.0.0` which wouldn't work obviously. I wonder if we should try to get current request IP from app object and use that in the event of a wildcard being specified

sonicaj (approved):

PR #14061, opened 2024-07-23
Remove `dotenv`

creatorcary:
Continuation of https://github.com/truenas/middleware/pull/14009

1. `dotenv` is not included in SCALE. Replace it.
2. The context manager `get_client()` caused a runtime error since it yielded more than once. Replace the context manager with a function that reliably retrieves the vendor name.

### DIFF
--- a/src/middlewared/middlewared/scripts/vendor_service.py
+++ b/src/middlewared/middlewared/scripts/vendor_service.py
@@ -4,9 +4,20 @@ import subprocess
 import sys
 import time
 
-from dotenv import dotenv_values
-
 from truenas_api_client import Client
+
+
+def load_envvars(envvars_file: str):
+    try:
+        envvars = []
+        with open(envvars_file, "r") as f:
+            for line in f:
+                l = line.strip()
+                if l and not l.startswith("#"):
+                    envvars.append(l.split("=", 1))
+        return dict(envvars)
+    except (OSError, ValueError):
+        return dict()
 
 
 def get_hostid() -> str | None:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/scripts/vendor_service.py
```diff
@@ -61,25 +71,34 @@ def start_hexos_websocat():
...
 
 
 def main():
-    vendor_name = None
-    with get_client() as client:
-        vendor_name = client.call("system.vendor.name")
+    vendor_name = get_vendor_name()
 
-    if vendor_name == "HexOS":
+    if vendor_name.upper() == "HEXOS":
```
anodos325, line 101: Do we have these defined in an enum somewhere? In theory they may not be the only vendor we deal with.

creatorcary: Not in middleware, and I don't think anywhere else yet. Where would be a good place to define one?

anodos325: We  could define an enum in `src/middlewared/middlewared/utils/vendor.py` unless @yocalebo has better idea of where he wants it.

src/middlewared/middlewared/scripts/vendor_service.py
```diff
@@ -61,25 +72,34 @@ def start_hexos_websocat():
...
+    Wait one second after each failed attempt. Raise an exception after failing the maximum allowed number of attempts.
+
+    """
+    tries = max_tries
+    while tries > 0:
         try:
             with Client() as c:
-                if c.call("system.ready"):
-                    yield c
+                for _ in range(tries):
```
anodos325, line 85: Why do we need this inner loop?

creatorcary: In case the client successfully opens but the system is not ready, instead of closing the client and opening a new one every second it will hold onto the client it successfully opened.

anodos325 (approved):

yocalebo (approved):

PR #14062, opened 2024-07-23
NAS-130160 / 24.10 / Prevent bug tickets from vendored install

aiden3c:
On a vendored system, we want to prevent tickets made via the API from going through. On top of returning false in `is_available`, we also raise an error if the ticket creation API method is called directly.

I also updated the grub generation to match the paradigm in [this PR](https://github.com/truenas/truenas-installer/pull/87).

If there are other areas that need to have these safeguards, let me know!

### DIFF
--- a/src/freenas/usr/local/bin/truenas-grub.py
+++ b/src/freenas/usr/local/bin/truenas-grub.py
@@ -6,6 +6,7 @@ import json
 
 from middlewared.utils.serial import serial_port_choices
 from middlewared.utils.db import query_config_table
+from middlewared.utils.vendor import Vendors
 
 
 def get_serial_ports():
@@ -16,10 +17,10 @@ if __name__ == "__main__":
     advanced = query_config_table("system_advanced", prefix="adv_")
     kernel_extra_options = advanced.get("kernel_extra_options") or ""
 
-    vendor = "TrueNAS Scale"
+    vendor = Vendors.TRUENAS_SCALE
     try:
         with open("/data/.vendor", "r") as f:
-            vendor = json.loads(f.read()).get("name", "TrueNAS Scale")
+            vendor = json.loads(f.read()).get("name", Vendors.TRUENAS_SCALE)
     except Exception:
         pass
 
--- a/src/freenas/usr/local/bin/truenas-grub.py
+++ b/src/freenas/usr/local/bin/truenas-grub.py
@@ -3,11 +3,13 @@ import math
 import psutil
 import os
 import json
+import logging
 
 from middlewared.utils.serial import serial_port_choices
 from middlewared.utils.db import query_config_table
 from middlewared.utils.vendor import Vendors
 
+logger = logging.getLogger(__name__)
 
 def get_serial_ports():
     return {e['start']: e['name'].replace('uart', 'ttyS') for e in serial_port_choices()}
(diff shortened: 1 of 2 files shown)


src/freenas/usr/local/bin/truenas-grub.py
```diff
@@ -16,12 +16,12 @@ def get_serial_ports():
     advanced = query_config_table("system_advanced", prefix="adv_")
     kernel_extra_options = advanced.get("kernel_extra_options") or ""
 
-    # check for /data/.vendor
+    vendor = "TrueNAS Scale"
```
anodos325, line 19: Let's define these in enum. See conversation and coordinate with https://github.com/truenas/middleware/pull/14061

aiden3c: Sounds good! I'll keep tabs on it

aiden3c: I'll update this branch once that PR gets merged!

src/freenas/usr/local/bin/truenas-grub.py
```diff
@@ -16,12 +17,12 @@ def get_serial_ports():
...
     kernel_extra_options = advanced.get("kernel_extra_options") or ""
 
-    # check for /data/.vendor
+    vendor = Vendors.TRUENAS_SCALE
     try:
         with open("/data/.vendor", "r") as f:
-            vendor = json.loads(f.read()).get("name", "TrueNAS Scale")
-    except FileNotFoundError:
-        vendor = "TrueNAS Scale"
+            vendor = json.loads(f.read()).get("name", Vendors.TRUENAS_SCALE)
```
anodos325, line 23: Should we have a check that the name read from the vendor file matches an value in our enum?

anodos325: At a minimum we should probably log an error message.

yocalebo: This will crash (by design) on proper TrueNAS machines so don't want to pollute logs unnecessarily. However, we could narrow the exception scope.
```
try:
    with open(
except FileNotFoundError:
    pass
except Exception:
    log.here

Comment (yocalebo) [reply src/freenas/usr/local/bin/truenas-grub.py:23]: This will crash (by design) on proper TrueNAS machines so don't want to pollute logs unnecessarily. However, we could narrow the exception scope.
```
try:
    with open(
except FileNotFoundError:
    pass
except Exception:
    log.here

yocalebo: This will crash (by design) on proper TrueNAS machines so don't want to pollute logs unnecessarily. However, we could narrow the exception scope.
```
try:
    with open(
except FileNotFoundError:
    pass
except Exception:
    log.here

File: src/middlewared/middlewared/plugins/support.py [hunk 3]
```diff
@@ -224,6 +227,9 @@ async def new_ticket(self, job, data):
         For SCALE Enterprise `token` and `type` attributes are not required.
         """
 
+        if await self.middleware.call('system.vendor.name'):
+            raise CallError('Support is not available for this product', errno.EINVAL)
```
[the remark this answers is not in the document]
aiden3c: Do we have a preference between `Support is not available for [name]` or `Support is not available for this product ([name])`?

src/freenas/usr/local/bin/truenas-grub.py
aiden3c: > Should we have a check that the name read from the vendor file matches an value in our enum?

I don't think we need to do this, as an invalid vendor will result in default TrueNAS functionality (just with a different Grub entry name). A user would have to really go out of their way to add a vendor during the install process.



> This will crash (by design) on proper TrueNAS machines so don't want to pollute logs unnecessarily. However, we could narrow the exception scope.

Gotcha, will do!

yocalebo (approved):
Wait for @anodos325 final review before merge please!

PR #14066, opened 2024-07-23
NAS-130193 / 24.10 / Correctly reference used ports key when validating app ports

sonicaj:


src/middlewared/middlewared/plugins/apps/schema_validation.py
```diff
@@ -131,7 +131,11 @@ def validate_acl_entries(self, verrors, value, question, schema_name, app_data):
             verrors.add(schema_name, f'{value["path"]}: path does not exist')
 
     async def validate_port_available_on_node(self, verrors, value, question, schema_name, app_data):
-        if app_data and value in [p['port'] for p in app_data['used_ports']]:
+        if app_data and value in [
```
yocalebo, line 134: These nested comprehensions are kind of brutal. Is there any possibility to simply iterate over the array and return early?
```
if app_data:
    for port in app_data['active_workloads']['used_ports']['host_ports']:
        if value == port['host_port']:
            return

yocalebo (approved):

PR #14068, opened 2024-07-23
Remove check for whether localhost connection is root

anodos325:
We don't use this functionality internally and it's a potential security liability if someone decides to set up their own internal proxy to middlewared socket that's running as root, or if an application allows root account and has access to host networking.


bmeagherix (approved):

yocalebo (approved):
I'm okay with this, but please wait for @themylogin to give it a once over before merge.

themylogin (changes requested):
We can remove `get_peer_process` definition and `RootTcpSocketSessionManagerCredentials` class, these are not used by anything else

tests/api2/test_ip_auth.py
```diff
@@ -8,17 +8,13 @@
```
themylogin, line 8: This test file can also be removed, it tests nothing now

themylogin (changes requested):

themylogin (approved):

PR #14069, opened 2024-07-23
NAS-130194 / 24.10 / Add additional ARC stats for UI

Qubad786:
## Problem

The UI requires new files that are currently missing in the `realtime.reporting` event.

## Solution

The necessary fields have been added to the `realtime.reporting` event. The UI can now retrieve the relevant information from the following keys in ZFS ARC reporting:

- Current size: `arc_size`
- Maximum size: `arc_max_size`
- Demand data hits: `arc_demand_data_hits`
- Demand metadata hits: `arc_demand_metadata_hits`
- Prefetch data hits: `arc_prefetch_data_hits`
- Prefetch metadata hits: `arc_prefetch_metadata_hits`

---
## Reference
https://learn.netdata.cloud/docs/collecting-metrics/freebsd/zfs
https://github.com/netdata/netdata/blob/master/src/collectors/proc.plugin/zfs_common.c#L19C24-L19C28

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/events.py
+++ b/src/middlewared/middlewared/plugins/reporting/events.py
@@ -47,6 +47,10 @@ class RealtimeEventSource(EventSource):
             'zfs',
             Int('arc_max_size'),
             Int('arc_size'),
+            Int('arc_demand_data_hits'),
+            Int('arc_prefetch_data_hits'),
+            Int('arc_demand_metadata_hits'),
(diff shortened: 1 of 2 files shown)


yocalebo (changes requested):
It's unfortunate but I've been working on this ticket all day. Netdata doesn't calculate the values properly. We need to go back to doing this ourselves unfortunately.

PR #14074, opened 2024-07-24
NAS-130214 / 24.10 / Make `Job.wait_sync` signature match `Job.wait`

themylogin:
Parameters order was changed, but no callers use positional args so we're good.


themylogin:
time 0:10

anodos325 (approved):

PR #14075, opened 2024-07-24
NAS-130225 / 24.10 / Add changes to allow retrieving app version details and config when querying apps

sonicaj:
This PR adds changes to allow UI to retrieve app version details and config of apps when querying apps so it can actually allow edits.

### DIFF
--- a/src/middlewared/middlewared/plugins/apps/crud.py
+++ b/src/middlewared/middlewared/plugins/apps/crud.py
@@ -71,11 +71,13 @@ class AppService(CRUDService):
         """
         Query all apps with `query-filters` and `query-options`.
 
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/apps/crud.py
```diff
@@ -72,20 +72,37 @@ def query(self, app, filters, options):
         Query all apps with `query-filters` and `query-options`.
 
         `query-options.extra.host_ip` can be provided to override portal IP address if it is a wildcard.
+
+        `query-options.extra.include_app_schema` can be provided to include app schema in the response.
```
Qubad786, line 76: We should indicate it's a boolean and it can be set to include relevant details

Qubad786 (approved):

PR #14076, opened 2024-07-24
NAS-128003 / 24.10 / Update zvol collection method

mgrimesix:
### Problem
The TrueNAS snmp agent was reporting zvols that resided directly on a pool.   However, zvols can be created in any dataset which means we don't report all possible zvols.
### Fix
Change the method to find zvols by doing a full depth search on each pool found in `/dev/zvol`

-----------------------------
Created a CI test for this condition.

**Included in this PR:**
* Fixed a few typos in TRUENAS-MIB.txt.  This required an update to TRUENAS-MIB.py (an auto-generated file).
* Updated the filesystem asset with adding a 'file' context manager.  It creates simple files.


src/freenas/usr/local/bin/snmp-agent.py
```diff
@@ -413,9 +414,9 @@ def get_list_of_zvols():
     zvols = set()
     root_dir = '/dev/zvol/'
     with contextlib.suppress(FileNotFoundError):  # no zvols
-        for zpool in pathlib.Path(root_dir).iterdir():
-            for zvol in filter(lambda x: '@' not in x.name, zpool.iterdir()):
-                zvol_normalized = zvol.as_posix().removeprefix(root_dir)
+        for zvol in iglob(root_dir + '**', recursive=True):
```
yocalebo, line 417: iglob is significantly slower than doing `os.scandir`, for example. It's why I used `Path.iterdir()` here. In python3.12, Path.iterdir() will begin to use os.scandir underneath the hood.

src/middlewared/middlewared/test/integration/assets/filesystem.py
```diff
@@ -11,3 +11,25 @@ def directory(path, options=None):
         yield path
     finally:
         ssh(f'rm -rf {path}')
+
+
+@contextlib.contextmanager
+def file(path, size=None):
```
yocalebo, line 17: It's a nit but can we rename this to `mkfile`

src/freenas/usr/local/bin/snmp-agent.py
yocalebo: Could even simplify to using `os.walk` since it uses scandir as well. (NOTE: I haven't tested this :sweat_smile: )
```
for r, d, f in os.walk(root_dir):
    for _f in filter(lambda x: '@' not in x, f):
        zvols.add(os.path.join(r, _f).removeprefix(root_dir).replace('+', ' '))

**Review by yocalebo — APPROVED** [review]

File: tests/api2/test_440_snmp.py [hunk 3]
```diff
@@ -97,6 +100,66 @@ def add_SNMPv3_user():
     yield
 
 
+@pytest.fixture(scope='function')
+def create_nested_structure():
+    """
+    Create the following structure:
+        tank -+-> dataset_1 -+-> dataset2 -+-> dataset_3
```
bmeagherix, line 107: Presumably this is a documentation bug.  Should `dataset2` be `dataset_2` ?

tests/api2/test_440_snmp.py
```diff
@@ -97,6 +100,66 @@ def add_SNMPv3_user():
...
+    dir_list = []
+    file_list = []
+    with ExitStack() as es:
+
+        for i in range(1, 4):
+            preamble = f"{ds_path + '/' if i > 1 else ''}"
+            vol_path = f"{preamble}zvol_{i}"
+
+            # Create zvols
+            for c in crange('a', 'b'):
```
bmeagherix, line 126: Would it make sense to also create a zol that contains a space character (to exercise the `replace('+', ' ')` )?

bmeagherix (approved):

PR #14078, opened 2024-07-25
NAS-130240 / 24.10 / skip known failing test

yocalebo:
We have 1 failing API test and its this one. This is a known problem and the OS team has been made aware. They're working on a fix but it's complicated so we'll skip it for now. We  have a high priority ticket internally.


anodos325:
Instead of 100% skipping this test, can we just comment out the part that is failing?

yocalebo:
> Instead of 100% skipping this test, can we just comment out the part that is failing?

yep, fixed it.

anodos325 (approved):

PR #14083, opened 2024-07-25
NAS-130230 / 24.10 / Prevent removing failover group from critical interfaces

aiden3c:
This prevents clearing `failover_group` from a NIC marked as `failover_critical`

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -1026,6 +1026,12 @@ class InterfaceService(CRUDService):
                     'Failover needs to be disabled to perform network configuration changes.'
                 )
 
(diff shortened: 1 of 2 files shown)


aiden3c:
[Passing test](http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/1657/console) (look for `test_004_remove_critical_failover_group`)

yocalebo (changes requested):
This changes the behavior of how an error is returned. Before this change, on interface create, we raised a CallError with specific verbiage. Now this will return a ValidationError with different error message on create.

On line 1046, we validate these during a create operation. We should probably add an `else` branch to that check (so we know it's an update operation) and raise your validation error.

yocalebo (approved):

PR #14088, opened 2024-07-26
NAS-129603 / 24.10 / Do not load alerts for resources that do not exist

themylogin:
Dangling cloud sync task alerts should have been fixed in 22.12, but the issue presented itself again. Let's put an additional layer of protection.


themylogin:
time 4:00

yocalebo (approved):

PR #14090, opened 2024-07-26
NAS-129685 / 24.10 / Ensure correct user homedir ownership

themylogin:
If user homedir is not owner by user, it won't be able to log in via SSH using key auth

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -842,7 +842,14 @@ class UserService(CRUDService):
     def recreate_homedir_if_not_exists(self, user, group, mode):
         # sigh, nothing is stopping someone from removing the homedir
         # from the CLI so recreate the original directory in this case
-        if not os.path.isdir(user['home']):
+        if os.path.isdir(user['home']):
+            if user['home'].startswith('/mnt/'):
+                self.middleware.call_sync('filesystem.chown', {
+                    'path': user['home'],
+                    'uid': user['uid'],
+                    'gid': group['bsdgrp_gid'],
+                }).wait_sync(raise_error=True)
+        else:
             if os.path.exists(user['home']):
                 raise CallError(f'{user["home"]!r} already exists and is not a directory')
 
--- a/src/middlewared/middlewared/plugins/keychain.py
+++ b/src/middlewared/middlewared/plugins/keychain.py
@@ -704,6 +704,8 @@ class KeychainCredentialService(CRUDService):
 
         # Write public key in user authorized_keys for SSH
         with open(f"{dotsshdir}/authorized_keys", "a+") as f:
+            os.fchmod(f.fileno(), 0o600)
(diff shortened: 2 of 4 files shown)


src/middlewared/middlewared/plugins/keychain.py
```diff
@@ -703,7 +703,10 @@ def ssh_pair(self, data):
         os.chown(dotsshdir, user["uid"], user["group"]["bsdgrp_gid"])
 
         # Write public key in user authorized_keys for SSH
-        with open(f"{dotsshdir}/authorized_keys", "a+") as f:
+        authorized_keys_path = f"{dotsshdir}/authorized_keys"
```
yocalebo, line 706: Any reason why you moved the path into its own variable? Seems like that change is superfluous.

src/middlewared/middlewared/plugins/account.py
```diff
@@ -842,7 +842,14 @@ def do_update(self, app, audit_callback, pk, data):
     def recreate_homedir_if_not_exists(self, user, group, mode):
         # sigh, nothing is stopping someone from removing the homedir
         # from the CLI so recreate the original directory in this case
-        if not os.path.isdir(user['home']):
+        if os.path.isdir(user['home']):
+            if user['home'].startswith('/mnt/'):
+                self.middleware.call_sync('filesystem.setperm', {
```
anodos325, line 847: Can't we `filesystem.chown` since we're not actually changing permissions?

src/middlewared/middlewared/plugins/account.py
```diff
@@ -842,7 +842,14 @@ def do_update(self, app, audit_callback, pk, data):
     def recreate_homedir_if_not_exists(self, user, group, mode):
         # sigh, nothing is stopping someone from removing the homedir
         # from the CLI so recreate the original directory in this case
-        if not os.path.isdir(user['home']):
+        if os.path.isdir(user['home']):
```
yocalebo, line 845: I think we can simplify this a bit more. We're calling `os.path.isdir` and potentially `os.path.exists` so we could remove the double stat call by doing something like this:
```
try:
    st = self.middleware.call_sync('filesystem.stat', user['home'])
except CallError as e:
    if e.errno == errno.ENOENT:
        pass
    else:
        # maybe log something here? or do we raise?
else:
    if st['type'] == 'DIRECTORY':
        if st['realpath'].startswith('/mnt'):
            ...
    else:
        raise CallError(f'{user["home"]!r} already exists and is not a directory')

yocalebo: Wonder what @anodos325 thinks about this. I could be wrong :smile:

anodos325: Yeah, that's probably better. If we have stat output then we can also use it to determine whether we _actually_ need to chown the directory.

yocalebo:
Handled in https://github.com/truenas/middleware/pull/14126

PR #14096, opened 2024-07-29
NAS-130295 / 24.10 / fix test_pool_replace_disk.py api failures

yocalebo:
This one is still consistently failing but it's failing inside the `disk.get_instance` calls. This is because the VMs can be a bit unpredictable wrt to populating udev/disk/symlink info. To try and work-around this, I've wrapped these calls in a simple retry loop. Passing tests are [here](http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/1723/testReport/api2/test_pool_replace_disk/)

### DIFF
--- a/tests/api2/test_pool_replace_disk.py
+++ b/tests/api2/test_pool_replace_disk.py
@@ -43,5 +43,25 @@ def test_pool_replace_disk(topology, i):
         assert len(disks(pool["topology"])) == count
         assert disks(pool["topology"])[i]["disk"] == new_disk["devname"]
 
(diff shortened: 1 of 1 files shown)


themylogin:
Should we just wait for `disk.sync_all` to complete? I can be wrong, but that seems to be the root cause of the issue.

themylogin:
The root cause of this issue is `sysevent.fs.zfs.config_sync` being triggered asynchronously. Since we have no easy way to wait for it, this solution is ok.

themylogin (approved):

PR #14111, opened 2024-07-30
NAS-130278 / 13.3 / NAS-130278 - Community Patch - Switch to bhyve TCP mode to fix VNC connections

kmoore134:
Switch to tcp mode for bhyve VNC sessions with community provided patch. Breaks some RGB color pallet in the WebVNC, but otherwise is reported as functional and restores basic VNC functionality. 

Note: Bhyve is unsupported in 13.3, we will not be doing any additional testing of this.


yocalebo (approved):

yocalebo:
Cool, thanks!

PR #14112, opened 2024-07-30
NAS-130338 / 24.10 / Allow overriding execute check in setacl in some cases

anodos325:
When TrueNAS is joined to active directory it's possible that the AD administrator has created nested security groups in such a way that it becomes non-trivial to validate whether a user can gain access to a path by virtue of being a member of a particular group.

This is because the nested security groups are flattened only when resolving the groups for a particular user via getgrouplist(3).

Since nested groups only exist if directory services are enabled, this bypass raises a ValidationError if the server is in a standalone configuration.

We still default to checking access (previous behavior) because using nested security groups in this way is a security anti-pattern as it renders the effective permissions on filesystem paths very difficult or impossible to easily audit.

### DIFF
--- a/src/middlewared/middlewared/plugins/filesystem_/acl.py
+++ b/src/middlewared/middlewared/plugins/filesystem_/acl.py
@@ -162,7 +162,7 @@ class FilesystemService(Service):
             verrors.add("filesystem.chown.uid",
                         "Please specify either user or group to change.")
 
-        loc = self._common_perm_path_validate("filesystem.chown", data, verrors)
+        self._common_perm_path_validate("filesystem.chown", data, verrors)
         verrors.check()
 
         if not options['recursive']:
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/filesystem_/acl.py
```diff
@@ -85,6 +85,14 @@ def _common_perm_path_validate(self, schema, data, verrors):
                 'Path component for is currently encrypted and locked'
             )
 
+        if data['options']['skip_execute_check']:
+            if self.middleware.call('directoryservices.status')['type'] is None:
```
bmeagherix, line 89: Should be `call_sync`, I believe.

bmeagherix (changes requested):

bmeagherix (approved):

PR #14113, opened 2024-07-30
NAS-130339 / 24.10 / Apps filesystem attachment handler

sonicaj:


src/middlewared/middlewared/plugins/apps/fs_attachments.py
```diff
@@ -0,0 +1,53 @@
...
+
+class AppFSAttachmentDelegate(FSAttachmentDelegate):
+    name = 'apps'
+    title = 'Apps'
+
+    async def query(self, path, enabled, options=None):
+        apps_attached = []
+        for app in await self.middleware.call('app.query'):
+            if not app['active_workloads']['volumes'] or (
+                app['state'] == 'STOPPED' if enabled else app['state'] != 'STOPPED'
```
yocalebo, line 12: I have looked at this `or` conditional so many times, my eyes hurt and I _still_ can't quite wrap my brain about why we're doing things like `app['state'] == 'STOPPED' if enabled`. It just doesn't make sense to me.

Can we put these checks into their own `if` `elif` branches with a comment on what they actually represent? I'm thinking something like this
```
if not app['active_workloads']['volumes']:
    # app isn't running or has no volume??
    continue
elif app['state'] == 'STOPPED' and enabled:
    # ???
    continue
elif app['state'] != 'STOPPED':
   # ???
   continue

yocalebo (approved):

PR #14143, opened 2024-08-06
NAS-130450 / 24.10 / Add changes to allow executing apps migration more then once

sonicaj:
This PR adds changes so that consumers can now execute app migration from k8s to docker even if it has already been applied. Motivation for this change is to allow users to migrate apps which were not migrated before because of perhaps the app not existing in the new catalog or it's migration script not being available on the apps side.
It also adds `migrated` flag which shows if the app was migrated over from k8s or is a newly installed app.


Qubad786 (approved):

Qubad786:
@stavros-k we have an issue with plex app's hostpaths - in my case it is getting migrated again and again because my original app had host path's involved and each time it fails because current/new plex's app's host path schema is broken

PR #14148, opened 2024-08-06
NAS-130459 / 24.10 / Add audit trails for snapshot and replication tasks

anodos325:
Generate audit trail when users configure snapshots and replication.


anodos325:
@themylogin feel free to make recommendations about what else to audit.

themylogin (approved):

PR #14151, opened 2024-08-06
NAS-130440 / 24.10 / fix pool.replace regression (introduced in 026691d39e4711107f40c7d0e8b48bbfc4788e9e)

yocalebo:
In https://github.com/truenas/middleware/commit/026691d39e4711107f40c7d0e8b48bbfc4788e9e, a commit was made to also wipe the disk that was being replaced when `pool.replace` was being called. I looked at that ticket but it's still not clear to me why this logic was added. In every possible scenario, it's entirely unnecessary to wipe the disk that is being replaced. Remove that logic and update `pool_replace_disk` api test accordingly. (also improve the test by documenting it as well).

### DIFF
--- a/src/middlewared/middlewared/plugins/pool_/replace_disk.py
+++ b/src/middlewared/middlewared/plugins/pool_/replace_disk.py
@@ -1,5 +1,4 @@
 import errno
-import os
 
 from middlewared.schema import accepts, Bool, Dict, Int, returns, Str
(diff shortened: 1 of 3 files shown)


yocalebo:
Passing tests are here: http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/1835/

anodos325 (approved):

PR #14154, opened 2024-08-07
NAS-128845 / 24.10 / Do not render all IPv4 addresses in Web UI URLs list if :: is in the listen address list, but `0.0.0.0` is not

themylogin:


themylogin:
time 1:30

Qubad786 (approved):

yocalebo (approved):

PR #14158, opened 2024-08-07
NAS-130485 / 24.10 / Improve auditing for middleware methods

anodos325:
Populate `username` and `address` fields in middleware audit entries based on information from the middleware session.

The UI prominently displays the `username` field which is currently hard-coded to root. This commit adds a helper function to generate a username string based on the current authenticated credentials for the session generating the audit entry.

### DIFF
--- a/src/middlewared/middlewared/plugins/auth.py
+++ b/src/middlewared/middlewared/plugins/auth.py
@@ -5,19 +5,18 @@ import errno
 import time
 import warnings
 
-import psutil
-
-from middlewared.auth import (SessionManagerCredentials, UserSessionManagerCredentials,
-                              UnixSocketSessionManagerCredentials, LoginPasswordSessionManagerCredentials,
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/auth.py
```diff
@@ -166,44 +160,6 @@ def dump(self):
         }
 
 
-class TokenSessionManagerCredentials(SessionManagerCredentials):
```
bmeagherix, line 169: Now `flake8` says that the following imports are not used in this file:
- `SessionManagerCredentials`
- `List`
- `Ref`
- `TCPIPOrigin`

(I suppose it's _possible_ that we reexport them, but unlikely.)

Also, extra space on line 609:
```
                user_info =  (await middleware.call(
```

bmeagherix (approved):

PR #14160, opened 2024-08-07
NAS-130390 / 25.04 / Move discovery auth from per-portal to system-wide

bmeagherix:
In CORE _discovery authentication_ was tied to portals.  This is not the case for SCALE, where it is system-wide.

This PR covers the first step in modifying SCALE to reflect reality.  We will continue to re-use the Authorized Access groups which are also used in _target auth_.  Multiple groups can be specified for discovery auth & will be combined, with one caveat ... only a single `peeruser/peersecret` can be in use (for Mutual CHAP).  (Actually, this caveat is not new.)

Alerts will be generated both on migration and subsequently if multiple mutual CHAP peers are defined.  A sentinel file is used to communicate the various alerts generated on DB migration (as `middlewared` is not running at that point).

- Perform database migration
- Generate alerts from migration if appropriate
- Add new `iscsi.discoveryauth.*` APIs
- Stub out auth aspect of `iscsi.portal.*` APIs, for later removal

----
Passing CI runs [1848](http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/1848/) (initial commit) and [1850](http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/1850/) (after 2 subsequent commits).

### DIFF
--- a/src/middlewared/middlewared/alembic/versions/24.10/2024-08-02_15-57_discoveryauth.py
+++ b/src/middlewared/middlewared/alembic/versions/24.10/2024-08-02_15-57_discoveryauth.py
@@ -12,8 +12,6 @@ import json
 import sqlalchemy as sa
 from alembic import op
 
-CHAP_TYPES = ['CHAP', 'CHAP Mutual']
-
 # revision identifiers, used by Alembic.
 revision = '208a066c65f7'
 down_revision = '81b8bae8fb11'
--- a/src/middlewared/middlewared/alembic/versions/24.10/2024-08-02_15-57_discoveryauth.py
+++ b/src/middlewared/middlewared/alembic/versions/24.10/2024-08-02_15-57_discoveryauth.py
@@ -1,0 +1,86 @@
+"""Add iSCSI discoverauth
+
+Flatten the per-portal discovery auth to a system-wide discovery auth.
+
+Revision ID: 208a066c65f7
+Revises: 81b8bae8fb11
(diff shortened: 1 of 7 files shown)


src/middlewared/middlewared/alembic/versions/24.10/2024-08-02_15-57_discoveryauth.py
```diff
@@ -0,0 +1,86 @@
...
+            alerts['ISCSIDiscoveryAuthMultipleMutualCHAP'] = {'peeruser': active_peeruser}
+
+    # Remove the obsolete columns
+    with op.batch_alter_table('services_iscsitargetportal', schema=None) as batch_op:
+        batch_op.drop_column('iscsi_target_portal_discoveryauthgroup')
+        batch_op.drop_column('iscsi_target_portal_discoveryauthmethod')
+
+    # Save any alerts
+    if alerts:
+        with open(DISCOVERY_AUTH_UPGRADE_COMPLETE_SENTINEL, 'w') as f:
```
anodos325, line 81: We can use `keyvalue` plugin in lieu of sentinel file (should be examples in other migrations).

bmeagherix: ok, thanks

bmeagherix:
Closing PR, as will be generating another one against a different branch first.

PR #14161, opened 2024-08-07
NAS-130271 / 24.10 / Check SnapshotCountAlert on SMB shares only

creatorcary:
This alert should only be checked for relevant datasets: those that are shared over SMB. The limit of 512 snapshots only applies due to Windows File Explorer limiting the number of snapshots shown in the Previous Versions tab.

I have also changed the alert description to more precisely describe the problem.

Relevant documentation
- [Snapshots](https://www.truenas.com/docs/scale/scaletutorials/datasets/managesnapshotsscale/#:~:text=File%20Explorer%20limits,functionality%20might%20degrade)
- [Compound `with`](https://docs.python.org/3.11/reference/compound_stmts.html#with:~:text=With%20more%20than,in%20multiple%20lines.)

API tests
- [All](http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/1845/)
- [Relevant](http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/1875/)

nas130271-alerts


anodos325:
Do we still have some sort of alert when users are doing absurd things with snapshots?

creatorcary:
> Do we still have some sort of alert when users are doing absurd things with snapshots?

Yeah the alert that triggers when there's over 10,000 total snapshots is still there.

src/middlewared/middlewared/alert/source/snapshot_count.py
```diff
@@ -43,11 +42,28 @@ async def check(self):
...
-                return Alert(
-                    SnapshotCountAlertClass,
-                    {"dataset": dataset, "count": count, "max": max_},
-                    key=None,
-                )
+    async def _check_smb(self) -> list[Alert]:
+        """Return an `Alert` for every dataset shared over smb whose number of snapshots exceeds the limit."""
+        max_ = await self.middleware.call("pool.snapshottask.max_count")
+        datasets = await self.middleware.call("zfs.snapshot.count")
+        smb_shares = await self.middleware.call("sharing.smb.query")
```
yocalebo, line 49: We're creating a bunch of unnecessary variables here which is memory inefficient. You're also now making duplicate calls to `zfs.snapshot.count` unnecessarily. The caller should initialize that variable and pass it around to the other methods. We can simplify this entire method quite a bit. I'd do something like this:
```
async def _check_smb(self, datasets: dict) -> list[Alert]:
    max_ = await self.middleware.call("pool.snapshottask.max_count")
    to_alert = list()
    for i in await self.middleware.call('sharing.smb.query'):
        count = datasets.get(i['path'].removeprefix('/mt/'), -1)
        if count > max_:
            to_alert.append(...

File: src/middlewared/middlewared/alert/source/snapshot_count.py [hunk 2]
```diff
@@ -43,11 +42,28 @@ async def check(self):
...
+                if count > max_:
+                    to_alert.append(Alert(
+                        SnapshotCountAlertClass,
+                        {"dataset": path, "count": count, "max": max_},
+                        key=path,
+                    ))
+
+        return to_alert
+
+    async def check(self):
```
yocalebo, line 65: You need to call `zfs.snapshot.count` in this method and then pass it around to the other methods.

yocalebo (changes requested):

src/middlewared/middlewared/alert/source/snapshot_count.py
```diff
@@ -43,11 +42,28 @@ async def check(self):
...
+        alerts = await self._check_smb()
+        if total_alert := await self._check_total():
```
yocalebo, line 67: There is no reason to check for `total_alert` and then append it to `alerts`. Make the methods always return a list and then inside `async def check` the method gets simplified to this
```
alerts = await self._check_smb()
alerts.extend(await self._check_total())
return alerts

Comment (yocalebo) [reply src/middlewared/middlewared/alert/source/snapshot_count.py:67]: If you want to get real fancy, you could simplify it to just
```
return await self.check_smb() + await self._check_total()

src/middlewared/middlewared/alert/source/snapshot_count.py
creatorcary: I do like this better.

yocalebo (approved):
Excellent! Please wait for @anodos325 approval before merge.

src/middlewared/middlewared/alert/source/snapshot_count.py
```diff
@@ -17,37 +17,48 @@ class SnapshotCountAlertClass(AlertClass):
...
 
-        for dataset in sorted(datasets.keys()):
-            count = datasets[dataset]
+    async def _check_smb(self, snapshot_counts: dict[str, int]) -> list[Alert]:
+        """Return an `Alert` for every dataset shared over smb whose number of snapshots exceeds the limit."""
+        max_ = await self.middleware.call("pool.snapshottask.max_count")
+        to_alert = list()
+
+        for share in await self.middleware.call("sharing.smb.query"):
+            path = share["path"].removeprefix("/mnt/")
```
anodos325, line 49: Share path may begin with `EXTERNAL:` in which case we should _not_ get snapshots.

anodos325: You can import `path_location` method from `middlewared.utils.path` and then check the location type and if it is not FSLocation.LOCAL skip the check.

anodos325 (changes requested):
Check for path location to ensure it's local to NAS before checking its snapshot count.

src/middlewared/middlewared/alert/source/snapshot_count.py
creatorcary: Should we also exclude external snapshot counts from `_check_total`?

anodos325 (approved):

PR #14162, opened 2024-08-07
NAS-130490 / 24.10 / Fix order_by when using raw sql queries on audit backend

anodos325:
Order_by directives were not being applied to the generated SQL queryset that was being passed to the individual audit backend services. This wasn't caught because often filtering for generic audit queries happens via filter_list because the filters cannot be fully duplicated using native SQL features.


mgrimesix (approved):
Approve after discussion with Andrew.

PR #14163, opened 2024-08-07
NAS-130498 / 24.10 / Omit SMB audit results from audit.query output by default

anodos325:
This is an optimization to improve memory consumption by default queries used by the webui. Initially we included SMB audit results by default with the expectation that user would specify the exact audited service they wanted to query. Since SMB audit logs can span upwards to hundreds of thousands of entries in some extreme cases we need to ensure that UI / API consumers use optimized SQL queryfilters with appropriate pagination.

### DIFF
--- a/src/middlewared/middlewared/plugins/audit/audit.py
+++ b/src/middlewared/middlewared/plugins/audit/audit.py
@@ -129,7 +129,7 @@ class AuditService(ConfigService):
 
     @accepts(Dict(
         'audit_query',
-        List('services', items=[Str('db_name', enum=ALL_AUDITED)], default=ALL_AUDITED),
+        List('services', items=[Str('db_name', enum=ALL_AUDITED)], default=['MIDDLEWARE', 'SUDO']),
         Ref('query-filters'),
         Ref('query-options'),
         register=True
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/audit/audit.py
```diff
@@ -129,7 +129,7 @@ async def compress(self, data):
 
     @accepts(Dict(
         'audit_query',
-        List('services', items=[Str('db_name', enum=ALL_AUDITED)], default=ALL_AUDITED),
+        List('services', items=[Str('db_name', enum=ALL_AUDITED)], default=['MIDDLEWARE', 'SUDO']),
```
bmeagherix, line 132: Nothing _wrong_ with this, but maybe could make a little more future-proof (so that it won't break when additional services are added).  For example...
```
ALL_AUDITED = [svc[0] for svc in AUDITED_SERVICES]
BULK_AUDIT = ['SMB']
NON_BULK_AUDIT = [svc for svc in ALL_AUDITED if svc not in BULK_AUDIT]
...
 List('services', items=[Str('db_name', enum=ALL_AUDITED)], default=NON_BULK_AUDIT),
```

Meh.

bmeagherix (approved):

PR #14167, opened 2024-08-08
NAS-130495 / 24.10 / Add support for deploying custom apps

sonicaj:
This PR adds support for a custom app which essentially will be a custom provided docker compose file from the user.


yocalebo (approved):
This is actually a pretty clean change. Awesome!

PR #14173, opened 2024-08-09
NAS-130513 / 24.10 / Add app events

sonicaj:
This PR adds changes to listen for docker events and based off those trigger app events so UI can get app status when the actual app's status changes and there is no need to poll for it.

### DIFF
--- a/src/middlewared/middlewared/plugins/apps/events.py
+++ b/src/middlewared/middlewared/plugins/apps/events.py
@@ -1,11 +1,8 @@
-import asyncio
-
 from middlewared.service import Service
 
 from .ix_apps.utils import get_app_name_from_project_name
 
 
-EVENT_LOCK = asyncio.Lock()
 PROCESSING_APP_EVENT = set()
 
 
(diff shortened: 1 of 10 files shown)


src/middlewared/middlewared/plugins/docker/events.py
```diff
@@ -0,0 +1,47 @@
...
+        ):
+            if not isinstance(container_event, dict):
+                continue
+
+            if project := container_event.get('Actor', {}).get('Attributes', {}).get(PROJECT_KEY):
+                self.middleware.send_event('docker.events', 'ADDED', id=project, fields=container_event)
+
+
+async def setup(middleware):
+    middleware.event_register('docker.events', 'Docker container events')
```
Qubad786, line 45: We are missing roles here

src/middlewared/middlewared/plugins/apps/events.py
```diff
@@ -0,0 +1,43 @@
+import asyncio
+
+from middlewared.service import Service
+
+from .ix_apps.utils import get_app_name_from_project_name
+
+
+EVENT_LOCK = asyncio.Lock()
```
yocalebo, line 8: There is no reason for this lock

yocalebo (approved):

PR #14178, opened 2024-08-10
NAS-130536 / 24.10 / Factor in container images as well when doing app upgrades

Qubad786:
## Context

Logic has been added which allows UI/consumers to see if an app has images which are outdated and similarly pull images for that app if that is the case. For custom apps, modification has been done so that we mark app as ready for upgrade if it has container images which need to be updated and similar `app.upgrade` has been modified to account for that change.

### DIFF
--- a/src/middlewared/middlewared/plugins/apps/pull_images.py
+++ b/src/middlewared/middlewared/plugins/apps/pull_images.py
@@ -1,0 +1,55 @@
+from middlewared.plugins.apps_images.utils import normalize_reference
+from middlewared.schema import accepts, Bool, Dict, List, returns, Str
+from middlewared.service import job, private, Service
+
+from .compose_utils import compose_action
+
+
+class AppService(Service):
+
+    class Config:
+        namespace = 'app'
+        cli_namespace = 'app'
+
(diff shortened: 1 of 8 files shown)


sonicaj (approved):

src/middlewared/middlewared/plugins/apps/pull_images.py
```diff
@@ -0,0 +1,55 @@
...
+        Str('name'),
+        Dict(
+            'options',
+            Bool('redeploy', default=True),
+        ),
+        roles=['APPS_WRITE']
+    )
+    @returns()
+    @job(lock=lambda args: f'pull_images_{args[0]}')
+    def pull_images(self, job, app_name, options):
```
anodos325, line 39: Should there be auditing for this action?

sonicaj: There should be, but apps don't have auditing changes in place atm. I think we can do that for RC for complete app plugin ?

anodos325: Sure. We can do that in RC.

PR #14194, opened 2024-08-12
NAS-129836 / 25.04 / replace `datetime.utcnow()`

creatorcary:
Deprecated in a future version of Python:
https://docs.python.org/3/whatsnew/3.12.html#deprecated:~:text=datetime%3A%20datetime,gh%2D103857.)

Tests: http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/1968/


themylogin (commented):
`utcnow` returns a naive datetime, `now(datetime.UTC)` returns a timezone-aware datetime. I am very surprised just replacing one with another just works everywhere.

aiden3c (approved):

yocalebo (changes requested):
Please implement what we discussed a bit ago :+1: Thanks!

creatorcary:
The `datetime` attribute of the `Alert` class should probably be named something other than its type. My type checker does not interpret its type annotation as intended and is flagging the `datetime` module as "unused". I won't try to refactor that in this ticket though because it would probably be too involved considering how many times `Alert` is referenced in middleware.

src/middlewared/middlewared/alert/base.py
```diff
@@ -245,8 +248,8 @@ class Alert:
     klass: type[AlertClass]
     args: typing.Union[dict[str, typing.Any], list]
     key: typing.Any
-    datetime: datetime
-    last_occurrence: datetime
+    datetime: DateTimeType
+    last_occurrence: DateTimeType
```
creatorcary, line 252: Type annotation was being resolved as the member variable `datetime` rather than the imported module `datetime`.

themylogin (changes requested):
The name `time_now` tells us nothing about the `utc` component. IMO it would be better to name it `utcnow` as it was named before

src/middlewared/middlewared/plugins/idmap.py
```diff
@@ -322,7 +322,7 @@ async def get_next_idmap_range(self):
     @private
     async def snapshot_samba4_dataset(self):
         sysdataset = (await self.middleware.call('systemdataset.config'))['basename']
-        ts = str(datetime.datetime.now(datetime.timezone.utc).timestamp())[:10]
```
themylogin, line 325: What's wrong with using the original method?

src/middlewared/middlewared/utils/time_utils.py
```diff
@@ -0,0 +1,6 @@
+from datetime import datetime, UTC
+
+
+def time_now(naive=True):
+    """Wrapper for `datetime.now()`. Exclude timezone if `naive=True`."""
+    return datetime.now(None if naive else UTC)
```
themylogin, line 6: `datetime.now(None)` and `datetime.utcnow()` are different

themylogin (changes requested):

themylogin:
> The `datetime` attribute of the `Alert` class should probably be named something other than its type. My type checker does not interpret its type annotation as intended and is flagging the `datetime` module as "unused". I won't try to refactor that in this ticket though because it would probably be too involved considering how many times `Alert` is referenced in middleware.

`datetime` is the correct type hint. What type checker are you using and what is the error message?

src/middlewared/middlewared/utils/time_utils.py
creatorcary: You're right, I'll have to go with [@aiden3c 's fix](https://github.com/truenas/middleware/pull/14194#issuecomment-2284094947).

src/middlewared/middlewared/plugins/idmap.py
creatorcary: I believe the rationale behind @yocalebo 's idea to write a time util function was to consolidate all of our `datetime.now` calls into one place since they're used so frequently in middleware. This makes it easier in the future to control how we get the current time.

creatorcary:
> `datetime` is the correct type hint. What type checker are you using and what is the error message?

`datetime` is the correct type hint, but the problem is that the `datetime` member variable is defined in a smaller scope and is taking precedence over the imported `datetime` class. A better solution is to change the name of this variable, but this solution is much simpler and brings less risk of breaking things.

Pyright flags the first type hint with `Type of "datetime" could not be determined because it refers to itself` and both type hints with `Variable not allowed in type expression`.

creatorcary:
> The name `time_now` tells us nothing about the `utc` component. IMO it would be better to name it `utcnow` as it was named before

The idea was to not copy the names from any standard libraries, hence why I went with `time_utils` instead of `time` and `time_now` instead of just `now`. I think the compromise here might be `utc_now` if that's not too confusing...

src/middlewared/middlewared/plugins/idmap.py
creatorcary: After realizing the nuances of working with `datetime` objects, I'm going to heed your warning and not touch the existing `datetime.now` calls...

themylogin (approved):
LGTM

creatorcary:
retest this please

yocalebo (approved):

PR #14195, opened 2024-08-12
NAS-130342 / 25.04 / Install Nvidia driver

themylogin:


src/middlewared/middlewared/plugins/nvidia.py
```diff
@@ -0,0 +1,130 @@
...
+            if line.split()[0] == "nvidia":
+                return True
+
+        return False
+
+    @job(lock="nvidia", description=lambda *args: "Installing NVIDIA drivers")
+    def install(self, job, start_docker=None):
+        self.middleware.call_sync("network.general.will_perform_activity", "catalog")
+
+        with ReadonlyRootfsManager() as rrm:
```
yocalebo, line 46: Before we enter this context manager, should we first check `self.installed` and return early?

src/middlewared/middlewared/alembic/versions/24.10/2024-08-12_08-09_nvidia.py
```diff
@@ -0,0 +1,57 @@
...
+revision = '5654da8713d1'
+down_revision = '4b0b7ba46e63'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_table('services_container')
+    op.drop_table('services_kubernetes')
```
sonicaj, line 22: We don't want to remove this for now, it's used for migration

src/middlewared/middlewared/utils/rootfs.py
```diff
@@ -0,0 +1,124 @@
...
+        binaries = (
+            # Used in `nvidia.install`
+            "apt",
+            "apt-config",
+            "apt-key",
+            # Some initramfs scripts use `dpkg --print-architecture` or similar calls
+            "dpkg",
+        )
+        if readonly:
+            for binary in binaries:
```
yocalebo, line 114: We can simplify this and remove duplicate code:
```
mode = 0o644 if readonly else 0o755
for binary in binaries:
    os.chmod(..., mode)
    ...

File: src/middlewared/middlewared/plugins/nvidia.py (hunk 1 above)
Comment (yocalebo) [reply src/middlewared/middlewared/plugins/nvidia.py:49]: Oh, I didn't see a few lines up. Derp.

File: src/middlewared/middlewared/utils/rootfs.py (hunk 3 above)
Comment (themylogin) [reply src/middlewared/middlewared/utils/rootfs.py:114]: How would you handle file renames then?

Comment (yocalebo) [reply src/middlewared/middlewared/utils/rootfs.py:114]: I'd do something similar inside the loop. Maybe something like
```
for binary in binaries:
    to_, from_ = f'blah/{binary}.bak', f'blah/{binary}'
    if not readonly:
        to_, from_ = from_, to_
```
Now that I write it out, I'm not sure I like mine either :slightly_smiling_face:. You can ignore this suggestion, I'm fine with whatever.

File: src/middlewared/middlewared/plugins/docker/state_management.py [hunk 4]
```diff
@@ -107,8 +107,19 @@ async def _event_system_ready(middleware, event_type, args):
...
-    if (await middleware.call('docker.config'))['pool']:
-        middleware.create_task(middleware.call('docker.state.start_service'))
+    docker_pool_configured = bool((await middleware.call('docker.config'))['pool'])
+    start_docker = True
+    if (
+        (await middleware.call('docker.config'))['nvidia'] and
+        await middleware.call('nvidia.present') and
+        not await middleware.call('nvidia.installed')
+    ):
+        await middleware.call('nvidia.install', docker_pool_configured)
```
yocalebo, line 117: Should we wait for this to complete since it's a job?

yocalebo: Hmm, so it seems you'll start a background task in this event and not start the docker service here but instead start it after the installation has completed. I guess that makes sense.

yocalebo (approved):
Approving on the condition that we put a temporary fix for ixdiagnose and mtree results for BETA.1 but we do another investigation to see if we can do something with overlayfs (systemd-sysext) by the time we release 24.10.0

themylogin:
backport

PR #14197, opened 2024-08-12
NAS-127040 / 25.04 / Add alert for local admin account use

anodos325:
Since it's trivial for users to create dedicated service accounts for administrative purposes via RBAC we should add an alert per GPOS STIG.

### DIFF
--- a/src/middlewared/middlewared/alert/source/auth.py
+++ b/src/middlewared/middlewared/alert/source/auth.py
@@ -1,0 +1,14 @@
+from middlewared.alert.base import AlertCategory, AlertClass, AlertLevel, SimpleOneShotAlertClass
+
+
+class AdminSessionActiveAlertClass(AlertClass, SimpleOneShotAlertClass):
+    category = AlertCategory.SYSTEM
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/alert/source/auth.py
```diff
@@ -0,0 +1,14 @@
+from middlewared.alert.base import AlertCategory, AlertClass, AlertLevel, SimpleOneShotAlertClass
+
+
+class AdminSessionActiveAlertClass(AlertClass, SimpleOneShotAlertClass):
+    category = AlertCategory.SYSTEM
+    level = AlertLevel.WARNING
+    title = "Active administrator session on TrueNAS server"
+    text = (
+        "System administrator (root or truenas_administrator) has one or more "
```
yocalebo, line 9: Name of the account is `truenas_admin`.

yocalebo (approved):

bmeagherix (approved):

william-gr:
backport

PR #14199, opened 2024-08-12
NAS-130573 / 25.04 / Update REST tests to use Websocket

aiden3c:
I've updated the following tests to use our websocket calls instead of REST methods

- test_020_account.py
- test_090_boot.py
- test_110_certificate.py
- test_490_system_general.py


aiden3c:
[Passing tests](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/75/console)

tests/api2/test_020_account.py
```diff
@@ -6,45 +6,39 @@
 
 import sys
 import os
+from middlewared.test.integration.utils import call, ssh
```
bmeagherix, line 9: `ssh` no longer needed.

tests/api2/test_090_boot.py
```diff
@@ -8,51 +8,43 @@
...
-    assert isinstance(results.json(), dict) is True, results.text
+    depends(request, ["BOOT_DISKS"])
     global boot_state
-    boot_state = results.json()
+    results = call("boot.get_state")
+    boot_state = results
 
 
-@pytest.mark.dependency(name='BOOT_SCRUB')
+@pytest.mark.dependency(name="BOOT_SCRUB")
```
anodos325, line 29: Can we maybe opportunistically remove the `depends` usage and any tests that are somewhat pointless.

tests/api2/test_020_account.py
```diff
@@ -6,45 +6,39 @@
...
 
     return user_id, group_id
 
 
 def test_01_delete_group_delete_users():
     user_id, group_id = delete_group_delete_users(True)
 
-    results = GET(f"/user/id/{user_id}")
-    assert results.status_code == 404, results.text
+    results = call(f"user.query", [["id", "=", user_id]])
```
bmeagherix, line 34: f-string not needed here

tests/api2/test_090_boot.py
aiden3c: Gotcha
The depends isn't necessary because pytest does the tests in order, correct?

tests/api2/test_490_system_general.py
```diff
@@ -10,59 +10,38 @@
 from pytest_dependency import depends
```
bmeagherix, line 10: Don't need to import `depends` here.

tests/api2/test_490_system_general.py
```diff
@@ -10,59 +10,38 @@
...
-from auto_config import user, password
 TIMEZONE = "America/New_York"
 
 
 def test_01_get_system_general():
-    results = GET("/system/general/")
-    assert results.status_code == 200, results.text
-    assert isinstance(results.json(), dict)
-
+    call("system.general.config")
```
bmeagherix, line 18: Another newline to silence `flake8`.

tests/api2/test_090_boot.py
anodos325: It's fixing a long-running anti-pattern in test development that we've been incrementally addressing. Generally setup or shared information between tests should be done through pytest fixtures.

tests/api2/test_110_certificate.py
```diff
@@ -9,9 +9,9 @@
 import os
 
 from time import sleep
+from middlewared.test.integration.utils import call
 apifolder = os.getcwd()
 sys.path.append(apifolder)
```
anodos325, line 14: We might be able to eliminate this as well

tests/api2/test_020_account.py
```diff
@@ -6,45 +6,39 @@
 
 import sys
 import os
+from middlewared.test.integration.utils import call, ssh
 apifolder = os.getcwd()
 sys.path.append(apifolder)
```
anodos325, line 11: Check if we still need to do this.

tests/api2/test_020_account.py
sonicaj: We should probably finally amend github action to factor in tests as well for linting

aiden3c:
New [passing tests](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/81/consoleFull)

tests/api2/test_020_account.py
```diff
@@ -1,50 +1,33 @@
-#!/usr/bin/env python3
+import pytest
```
yocalebo, line 1: Thanks! One last thing about this `test_20_account.py` file. We already have a `test_account.py` file that tests quite a few other things. We should remove this file and copy and paste the tests over to that file and re-use existing boilerplate over there if we can.

yocalebo (changes requested):

aiden3c:
I ended up removing `test_020_account.py` as every method it tests is done (either directly or inadvertently) in `test_account.py`.

yocalebo (approved):

PR #14220, opened 2024-08-13
NAS-130598 / 25.04 / Fix handling of call to session logout

anodos325:
Since we are toggling alerts potentially when performing logout the session logout method needs to be converted properly into an async coroutine like session longin, and we need to make callers to the function properly await it.

### DIFF
--- a/tests/api2/test_root_session_alert.py
+++ b/tests/api2/test_root_session_alert.py
@@ -1,0 +1,46 @@
+import pytest
+
+from middlewared.test.integration.utils.client import client, truenas_server
+from middlewared.test.integration.utils import call
+
+
+def get_session_alert(call_fn):
+    alerts = call_fn('alert.list')
+    alert_msg = None
+
+    for alert in alerts:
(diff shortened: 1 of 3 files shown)


tests/api2/test_root_session_alert.py
```diff
@@ -0,0 +1,46 @@
...
+    for alert in alerts:
+        if alert['klass'] == 'AdminSessionActive':
+            alert_msg = alert['formatted']
+            break
+
+    assert alert_msg is not None, str(alerts)
+    return alert_msg
+
+
+def check_session_alert(call_fn):
```
yocalebo, line 20: What is call_fn?

anodos325: `call` or `c.call` e.g. client call methods.

yocalebo: derp, I haven't had coffee yet

yocalebo (approved):

PR #14223, opened 2024-08-14
NAS-130617 / 24.10-RC.1 / Undo discoveryauth

bmeagherix:
The decision to add this to EE has been reverted.  Revert the previous commits.


yocalebo:
@themylogin need your input on this one since we're removing a migration. NOTE: this did not get merged into BETA.1

anodos325 (approved):

themylogin (approved):
If we're fine with some Electriceel Nightly release being non-upgradable due to this, then it's ok to remove the last migration.

PR #14225, opened 2024-08-14
NAS-130618 / 25.04 / Set maximum size of message we will parse

anodos325:
RFC5424 specifies that maximum size of SDATA for a syslog message as 64 KiB. Since these messages will in some circumstances be converted into audit entries and sent to remote syslog server we should set an upper bound on size of message we process (to prevent dropping audited messages). In principle there is no reason to ever be receiving giant messages.

### DIFF
--- a/src/middlewared/middlewared/api/base/server/ws_handler/rpc.py
+++ b/src/middlewared/middlewared/api/base/server/ws_handler/rpc.py
@@ -19,6 +19,7 @@ from middlewared.schema import Error
 from middlewared.service_exception import (CallException, CallError, ValidationError, ValidationErrors, adapt_exception,
                                            get_errname)
 from middlewared.utils.debug import get_frame_details
+from middlewared.utils.limits import MsgSizeError, MsgSizeLimit, parse_message
 from middlewared.utils.lock import SoftHardSemaphore, SoftHardSemaphoreLimit
 from middlewared.utils.origin import Origin
 from .base import BaseWebSocketHandler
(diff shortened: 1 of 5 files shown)


bmeagherix (approved):

themylogin (changes requested):
We can receive even larger messages (e.g. calls to `filesystem.file_receive`)

tests/api2/test_large_message.py
```diff
@@ -0,0 +1,33 @@
...
+    assert MSG_TOO_BIG_ERR in ce.value.error
+
+
+def test_large_message_extended():
+    LARGE_PAYLOAD_1 = 'x' * 65537
+    LARGE_PAYLOAD_2 = 'x' * 2097153
+
+    with pytest.raises(ValidationErrors):
+        with client() as c:
+            c.call('filesystem.file_receive', LARGE_PAYLOAD_1)
```
bmeagherix, line 27: Is this not a valid payload size for this method?

src/middlewared/middlewared/utils/limits.py
```diff
@@ -0,0 +1,73 @@
...
+from truenas_api_client import json as ejson
+
+
+class MsgSizeLimit(enum.IntEnum):
+    UNAUTHENTICATED = 8192  # maximum size of message processed from unauthenticated session
+    AUTHENTICATED = 65536  # maximum size of message processed from authentication session
+    EXTENDED = 2097152  # maximum size of message that sends a file
+
+
+class MsgSizeExtended(enum.Enum):
```
themylogin, line 12: This can be just `set`

themylogin (changes requested):

src/middlewared/middlewared/main.py
```diff
@@ -1855,15 +1856,31 @@ async def ws_handler(self, request):
                     )
                     break
 
-                if not connection.authenticated and len(msg.data) > 8192:
+                datalen = len(msg.data)
+
+                try:
+                    message = msg_size_check(connection.authenticated, datalen, msg.data)
```
themylogin, line 1862: It parses JSON and then checks the message size. I think we can call it `parse_message` or something like this.

src/middlewared/middlewared/api/base/server/ws_handler/rpc.py
```diff
@@ -216,15 +217,29 @@ async def process(self, origin: Origin, ws: WebSocketResponse):
...
+                    self.middleware.logger.error(
+                        'Client using credentials [%s] at [%s] sent message with payload size [%d bytes] '
+                        'exceeding limit of %d for method %s',
+                        creds, origin, err.datalen, err.limit, err.method_name
+                    )
                     await ws.close(
-                        code=WSCloseCode.INVALID_TEXT,
-                        message=b"Anonymous connection max message length is 8 kB",
+                        code=WSCloseCode.MESSAGE_TOO_BIG,
+                        message='Max message length is 64 kB'.encode('utf-8'),
```
themylogin, line 240: We can raise `MsgSizeError` with proper `message`. ANother advantage of this is that we won't be hard-coding duplicate message size constants here.

src/middlewared/middlewared/main.py
anodos325: OK. I'll rename.

themylogin (approved):

PR #14230, opened 2024-08-15
NAS-130523 / 25.10 / Add incus as a dependency

william-gr:
Adding incus as a dependency to facilitate testing/development using nightly images.


yocalebo:
Adding packages here in this file pulls in about every suggestion and recommended package. Does this include a bunch of other stuff?

william-gr:
Recommends: apparmor, dnsmasq-base, incus-agent, minio-client

Prefer to do it in scale-build and do not install these?

yocalebo:
> Recommends: apparmor, dnsmasq-base, incus-agent, minio-client
> 
> Prefer to do it in scale-build and do not install these?

Yeah, let's do it there. I already had to do it with some other package.

PR #14231, opened 2024-08-15
NAS-130624 / 25.04 / Fix system.ready and ix-vendor.service

yocalebo:
A few issues fixed here:

1. ix-vendor.service created a cyclical dependency which was causing a cascading set of failures. Most notably, `system.ready` endpoint would get stuck returning `False` because `system.state` was stuck in `BOOTING`.
2. the `vendor_service.py` script wasn't getting installed, so after fixing the service file, it failed because it couldn't find the executable.
3. remove the use of `Client()` in `vendor_service.py` by adding a `get_vendor()` function to `system_vendor/vendor.py`. This simplifies the script and makes it more resilient.
4. Add a `-` to the `ExecStart` stanza in `ix-vendor.service`. This is required because we use some weird legacy entrypoints stuff to create executable files for these scripts defined in `middlewared/setup.py`. Because of this, the actual executable (`start_vendor_service`) will exit uncleanly in the event of any failure, even though the actual script (`vendor_service.py`) has `try: finally: exit(0)`

### DIFF
--- a/debian/debian/ix-vendor.service
+++ b/debian/debian/ix-vendor.service
@@ -7,5 +7,5 @@ After=ix-netif.service
 Type=oneshot
 RemainAfterExit=yes
 TimeoutStartSec=30
-ExecStart=/usr/bin/vendor_service
+ExecStart=-/usr/bin/vendor_service
 StandardOutput=null
--- a/debian/debian/ix-vendor.service
+++ b/debian/debian/ix-vendor.service
@@ -7,5 +7,5 @@ After=ix-netif.service
 Type=oneshot
 RemainAfterExit=yes
 TimeoutStartSec=30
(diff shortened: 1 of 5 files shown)


debian/debian/ix-vendor.service
```diff
@@ -1,14 +1,11 @@
 [Unit]
 # vendor_service.py must run after network routes are set up.
-After=ix-postinit.service
+After=ix-netif.service
```
sonicaj, line 3: I think we should keep this to `ix-postinit` as that means the system is ready at this point and then we are calling some vendor script which can be anything and can try to do anything which might/might not work based on how far along middleware is ready at that point.
We should change `WantedBy` to `truenas.target` instead - that could probably be causing the dependency issue

PR #14247, opened 2024-08-16
add more default zfs props for apps internal dataset

yocalebo:
1. add more default zfs properties to the internal apps dataset
2. convert the global mutable dictionary to an immutable dataclass to prevent the possibility of scrambling this object during runtime (the dataclass is also frozen so it's more memory efficient)


anodos325 (approved):

src/middlewared/middlewared/plugins/kubernetes_to_docker/migrate.py
```diff
@@ -182,7 +182,7 @@ def migrate(self, job, kubernetes_pool, options):
                         'snapshot': snapshot,
                         'dataset_dst': destination_ds,
                         'dataset_properties': {
-                            k: v for k, v in DATASET_DEFAULTS.items() if k not in ['casesensitivity']
+                            k: v for k, v in DATASET_DEFAULTS.to_dict().items() if k != 'casesensitivity'
```
anodos325, line 185: We probably need to expand this to include `normalization` as well

yocalebo: Working on simplifying this a bit more. We're setting `encyprtion=off in another function instead of keeping everything in one place.

sonicaj (approved):

yocalebo:
Thanks @sonicaj !

PR #14260, opened 2024-08-16
NAS-130011 / 25.04 / Add drive bay light status

bmeagherix:
This PR adds initial support for `drive_bay_light_status` in `enclosure2.query` and `webui.enclosure.dashboard` APIs.

Support is added for JBOF and SES based enclosures.  Generally, being able to read light status follows whether it is possible to
set it (as indicated by `supports_identify_light` in slot mappings).  However, this can be overridden by adding a specific
`supports_identify_light_status` key/value to slot mappings - e.g. for platforms where it is possible to set the light status, but **not** to read it back again.

Other changes:
- As requested, change the `enclosure2.set_slot_status` API `status` parameter.  It will now accept the strings **"ON"**, **"OFF"** or **"CLEAR"**.
- For SES based enclosures the `original slot` value was incorrect, resulting in an off-by-one wrt which disk had its IDENT lit when requested.  Rectify by actually extracting the `slot` number from the `descriptor` string returned by the status method in `libsg3`.

### DIFF
--- a/src/middlewared/middlewared/plugins/enclosure_/enclosure_class.py
+++ b/src/middlewared/middlewared/plugins/enclosure_/enclosure_class.py
@@ -4,7 +4,6 @@
 # See the file LICENSE.IX for complete terms and conditions
 
 import logging
-import re
 
 from middlewared.utils.scsi_generic import inquiry
 
@@ -27,9 +26,6 @@ from .sysfs_disks import map_disks_to_enclosure_slots
 from .slot_mappings import get_slot_info
 
 logger = logging.getLogger(__name__)
-
-SLOTNNN_RE = re.compile('^slot[0-9]{2,3}$')
-SLOT_NNN_COMMA_RE = re.compile('^SLOT [0-9]{2,3},')
 
 
 class Enclosure:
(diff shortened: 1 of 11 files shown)


src/middlewared/middlewared/plugins/enclosure_/enclosure_class.py
```diff
@@ -306,16 +312,44 @@ def _parse_elements(self, elements):
...
+                else:
+                    parsed[DRIVE_BAY_LIGHT_STATUS] = None
+
                 mapped_slot = self.disks_map[slot][MAPPED_SLOT_KEY]
                 # is this a front, rear or internal slot?
                 parsed.update(disk_position_mapping.get(mapped_slot, dict()))
 
+                # Pick out the original slot number from the descriptor,
+                # for example 'slot00', 'slot01', ... etc
+                try:
```
yocalebo, line 335: Sorry but this won't work. I'll go investigate why we're off by 1 for SES proper devices.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/enclosure_/enclosure_class.py
yocalebo: We're off by one in py-sg3 in the `status()` method here
```
                elements = {}
                for i in range(num_dsc):
                    type = tc_buff[i]
                    dsc = self.decode_response(desc_buff[i])
                    stts = [int(x, 16) for x in self.decode_response(stat_buff[i]).split()]
                    elements[i] = { "type": type, "descriptor": dsc, "status": stts }

yocalebo (approved):

PR #14262, opened 2024-08-19
NAS-127511 / 25.04 / Fix zpool status for reporting used spare disks correctly

Qubad786:
## Problem

`zpool.status` was not properly reporting path for disk properly when a spare had replaced a data disk.

## Solution

We are now using `zpool status` for retrieving serialized information about the zpool so we can properly account for disk path being reported for spare based disks and that is much better optimized for a system with large number of disks as compared to py-libzfs.

### DIFF
--- a/src/middlewared/middlewared/plugins/webui/enclosure.py
+++ b/src/middlewared/middlewared/plugins/webui/enclosure.py
@@ -37,7 +37,6 @@ class WebUIEnclosureService(Service):
             pool_info['vdev_disks'].append(info)
 
     def dashboard_impl(self):
-        disks_to_pools = dict()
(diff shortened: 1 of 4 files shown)


themylogin:
Can we have integration tests for this?

yocalebo (changes requested):
`zpool status` will resolve the vdev devices so we need to remove the `resolve_*` methods accordingly.

If `real_paths` is true, we can use `-PL` otherwise, we can just use `-P`.

`-P` will give full path to device (i.e. `/dev/disk/by-partuuid/**`)
`-PL` will give full path to real device (i.e. `/dev/sda1`)

yocalebo (approved):
No, actually, we'll leave it as is for now since this is going into RC.1

PR #14281, opened 2024-08-20
Fix attribute error for endpoints consuming pass_app

Qubad786:
## Problem

An `AttributeError` was being raised when trying to evaluate conditions in the any() function, specifically when accessing the origin attribute of a NoneType object.

## Solution

The error occurred because Python evaluates all elements in a tuple before passing them to `any()`. The fix involved using a logical and to ensure that the condition `not isinstance(app.origin, TCPIPOrigin)` is only evaluated if app is not None, preventing the AttributeError.


Qubad786:
Was keeping the earlier changes but yes - no need to have `any` here ✅

PR #14290, opened 2024-08-20
NAS-130641 / 25.04 / Fix rsync tasks with IPv6 hostnames

themylogin:


themylogin:
time 0:30

yocalebo (approved):

src/middlewared/middlewared/plugins/rsync.py
```diff
@@ -615,6 +615,11 @@ def commandline(self, id_):
                     port = rsync['remoteport']
                     extra_args = ''
 
+                remote_username, remote_host = remote.rsplit('@', 1)
+                if ':' in remote_host:
```
bmeagherix, line 619: Do we know that the `remote_host` doesn't already have enclosing brackets?

themylogin: Yes, it's a valid hostname

PR #14297, opened 2024-08-20
NAS-130708 / 25.04 / Clear pool status alert cache when I/O errors occur on the pool

themylogin:
These errors, if corrected, do not degrade vdev and don't cause `sysevent.fs.zfs.config_sync`.

Not able to add an integration test since causing these errors with `zinject` irreversibly makes the system barely usable.


themylogin:
time 2:00

yocalebo (approved):

PR #14299, opened 2024-08-21
Add test coverage for MacOS-style SMB renames

anodos325:
Newer MacOS clients may rename files on SMB shares via names that are not returned in the directory listing it requested. For example if source of rename is foo and target is Foo, it may try to rename Foo to Foo resulting in hitting an optimization in samba that turns this rename into a no-op. This commit expands tests specifically to cover this case.


tests/api2/test_425_smb_protocol.py
```diff
@@ -459,14 +459,25 @@ def test_068_case_insensitive_rename(request):
...
+        share=SMB_NAME,
+        username=SMB_USER,
+        password=SMB_PWD,
+        smb1=False
+    ) as c:
+        fd = c.create_file("to_rename", "w")
+        c.close(fd)
+        c.rename("to_rename", "To_rename")
+        files = [x['name'] for x in c.ls('\\')]
+        assert "To_rename" in files
```
bmeagherix, line 472: Maybe add a
```
assert "to_rename" not in files
```

tests/api2/test_425_smb_protocol.py
```diff
@@ -459,14 +459,25 @@ def test_068_case_insensitive_rename(request):
...
+
+        # MacOS Sonoma currently (Aug 2024) gets SMB handle on file to be renamed
+        # potentially via the target of the rename which potentially hits optimization
+        # in samba. This validates that rename in this way also works on case-insensitve
+        # filesystems.
+        c.rename("to_rename", "to_rename")
+        files = [x['name'] for x in c.ls('\\')]
+        assert "to_rename" in files
```
bmeagherix, line 480: Maybe add a
```
assert "To_rename" not in files
```

bmeagherix (approved):
Made a couple of possibly redundant suggestions (but they also clarify, IMO).

Regardless, approving.

PR #14301, opened 2024-08-21
NAS-130653 / 25.04 / Updated debug generate lock queue size

aiden3c:
Adding this limit stops people from spamming debug generation, as most of the time just one job is required.

### DIFF
--- a/src/middlewared/middlewared/job.py
+++ b/src/middlewared/middlewared/job.py
@@ -50,7 +50,7 @@ class State(enum.Enum):
     ABORTED = 5
 
 
-class JobSharedLock(object):
+class JobSharedLock:
     """
     Shared lock for jobs.
     Each job method can specify a lock which will be shared
@@ -83,8 +83,7 @@ class JobSharedLock(object):
         return self.lock.release()
 
 
-class JobsQueue(object):
-
+class JobsQueue:
     def __init__(self, middleware):
         self.middleware = middleware
         self.deque = JobsDeque()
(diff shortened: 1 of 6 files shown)


themylogin:
What will happen with the subsequent request in the UI?

aiden3c:
> What will happen with the subsequent request in the UI?

If we're at the limit of the queue, every subsequent result in the UI takes the same job. One issue with this actually is that the download url given is "one time use only" so those extra calls fail. This is if the `debug` job specifically is limited, if `debug_generate` is limited then it effectively does nothing as `debug` is what's called from the UI.

Thinking about the best way to go about this, maybe just reject subsequent calls from the UI entirely?

themylogin:
> > What will happen with the subsequent request in the UI?
> 
> If we're at the limit of the queue, every subsequent result in the UI takes the same job. One issue with this actually is that the download url given is "one time use only" so those extra calls fail. This is if the `debug` job specifically is limited, if `debug_generate` is limited then it effectively does nothing as `debug` is what's called from the UI.
> 
> Thinking about the best way to go about this, maybe just reject subsequent calls from the UI entirely?

So how will the UI behave if we merge this changes? If I request a debug while a debug is already being gathered, will both UI tabs just download me the same file?

aiden3c:
It'll try to, but after the file is downloaded the link expires. The other UI tab that tries to use that link returns with
image

themylogin:
@aiden3c does the double click lead to the same behavior? We need to resolve this somehow. I would say, we refuse this type of request on our side (need modifications in `core.download`), and UI fix their double request issue.

themylogin:
time 1:30

themylogin:
http://jenkins.eng.ixsystems.net:8080/job/master/job/custom/473/

anodos325 (approved):

PR #14304, opened 2024-08-21
NAS-130751 / 25.04 / Make sure pool specified by user actually exists for apps

Qubad786:
## Context

We were not validating if pool specified by user actually exists for apps.


sonicaj (approved):

yocalebo (changes requested):
This is too expensive. Please use the `query_imported_fast_impl` function from `middlewared.utils.zfs`

PR #14307, opened 2024-08-21
NAS-130753 / 25.04 / fix AttributeError crash in app.query

yocalebo:


bmeagherix (approved):
Oh, if `info` is not None, but doesn't have a `local_ip` attribute?

yocalebo:
> Oh, if `info` is not None, but doesn't have a `local_ip` attribute?

The return definition for `read_proc_net` is an empty list or a list full of InetInfoEntry objects or a InetInfoEntry object. My check was too strict and I didn't consider the fact that it can return an emply list. A list, obviously. doesn't have a `local_ip` attribute.

Hindsight being 20-20, I could have made the return less complex in `read_proc_net` but I've also discovered that we shouldn't be passing the `app` object around like this and we can easily get access to this information but that will come in a separate PR to keep churn as low as possible in BETA.1.

PR #14313, opened 2024-08-21
NAS-130755 / 25.04 / Extend search for ip in websocket_interface

bmeagherix:
Recent addition of `test_websocket_interface` (in PR #14276) was failing frequently.

Apparently the VIP was not being handled well for HA systems.

Add search of `iface['state']['aliases']` if search of `iface['aliases']` does not yield a result.

### DIFF
--- a/src/middlewared/middlewared/plugins/network.py
+++ b/src/middlewared/middlewared/plugins/network.py
@@ -1474,15 +1474,9 @@ class InterfaceService(CRUDService):
         if local_ip is None:
             return
 
-        interfaces = await self.middleware.call('interface.query')
-        for iface in interfaces:
-            for alias in iface['aliases']:
-                if alias['address'] == local_ip:
-                    return iface
-        for iface in interfaces:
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/network.py
```diff
@@ -1475,10 +1474,15 @@ async def websocket_interface(self, app):
         if local_ip is None:
             return
 
-        for iface in await self.middleware.call('interface.query'):
+        interfaces = await self.middleware.call('interface.query')
+        for iface in interfaces:
             for alias in iface['aliases']:
                 if alias['address'] == local_ip:
                     return iface
+        for iface in interfaces:
```
yocalebo, line 1482: Why another loop? Why not simplify to just:
```
for iface in await self.middleware...:
    for alias in filter(lambda x: x['address'] == local_ip, iface['aliases'] + iface['state']['aliases']):
        return iface

yocalebo (approved):

PR #14314, opened 2024-08-21
NAS-130197 / 25.04 / Add NFS and FTP statistics to usage reporting

mgrimesix:
Add NFS and FTP usage statistics

- NFS
-- Enabled protocols, krb5 enabled, number of clients
- FTP
-- Number of connections

Added a simple CI test for the usage reporting
Added `connection_count` as an NFS service function.
Added `nfs_server` test asset and made the `config` parameter optional in the `ftp_server` test asset.

Test results are [here](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/269/).

### DIFF
--- a/src/middlewared/middlewared/plugins/ftp_/status.py
+++ b/src/middlewared/middlewared/plugins/ftp_/status.py
@@ -13,7 +13,7 @@ class FTPService(Service):
         ftp = 21
 
         try:
-            proc_data = self.middleware.call(read_proc_net)
+            proc_data = read_proc_net()
             ftp_proclist = list(filter(lambda x: x.local_port == ftp and x.remote_port != 0, proc_data))
         except Exception:
             num_conn = 0
--- a/src/middlewared/middlewared/plugins/ftp_/status.py
+++ b/src/middlewared/middlewared/plugins/ftp_/status.py
@@ -1,0 +1,25 @@
+from middlewared.service import Service, private
+from middlewared.schema import Int, returns
+from middlewared.utils.network_.procfs import read_proc_net
+
+
+class FTPService(Service):
+
+    @private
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/usage.py
```diff
@@ -329,6 +329,37 @@ async def gather_pools(self, context):
...
+                'kerberos': nfs_config['v4_krb_enabled'],
+                'num_clients': num_clients,
+            }
+        }
+
+    async def gather_ftp(self, context_unused):
+        ftp_config = await self.middleware.call('ftp.config')
+
+        try:
+            res = subprocess.run(['netstat', '-4', '-6'], capture_output=True, text=True)
```
yocalebo, line 347: A couple things:
1. `gather_ftp` is a coroutine but you're blocking main event loop with a synchronous fork+exec (subprocess).
2. I _really_ don't like this being called because busy systems could potentially have lots of entries and this is resolving ip addresses to dns names by default (which is also expensive).

yocalebo (changes requested):

src/middlewared/middlewared/plugins/ftp_/status.py
```diff
@@ -0,0 +1,25 @@
...
+
+    @private
+    @returns(Int('number_of_connections'))
+    def connection_count(self):
+        ''' Return the number of active connections '''
+        # FTP listening port is 21
+        ftp = 21
+
+        try:
+            proc_data = self.middleware.call(read_proc_net)
```
yocalebo, line 16: This won't work. You can just call read_proc_net() since this method is synchronous.

yocalebo (approved):

PR #14316, opened 2024-08-22
NAS-130765 / 25.04 / Fix read_proc_net.

mgrimesix:
The `read_proc_net` function does not generate correct addresses and does not support both big-endian and little-endian systems.

Updated routine generates correct ipv6 and ipv4 addresses from the hex values in /proc/net/tcp[6]


yocalebo (approved):
I have tested these changes on real system with proper ipv6/v4 and it works. Thanks!

PR #14322, opened 2024-08-22
NAS-130768 / 25.04 / Improve passdb handling

anodos325:
This commit adds utility functions to pack / unpack samba's passdb.tdb file. This allows us to remove subprocess calls to manipulate our SMB entries and provides a clean mechanism for maintaining more robust account details in the future when we expand our user database objects. In ad-hoc testing this resulted in about a 2x speed improvement for user account CRUD operations.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb_/passdb.py
+++ b/src/middlewared/middlewared/plugins/smb_/passdb.py
@@ -1,12 +1,17 @@
-from middlewared.service import filterable, Service, job, private
-from middlewared.service_exception import CallError, MatchNotFound
-from middlewared.utils import run, filter_list
-from middlewared.utils.sid import db_id_to_rid
-from middlewared.plugins.idmap_.idmap_constants import IDType
-from middlewared.plugins.smb import SMBCmd, SMBPath
-
 import os
-import time
+
+from middlewared.api.current import UserEntry
+from middlewared.service import filterable, Service, job, private
+from middlewared.utils.sid import get_domain_rid
+from .util_passdb import (
+    delete_passdb_entry,
+    insert_passdb_entries,
+    query_passdb_entries,
+    update_passdb_entry,
+    user_entry_to_passdb_entry,
+    PassdbMustReinit,
+    PASSDB_PATH
+)
 
 
 class SMBService(Service):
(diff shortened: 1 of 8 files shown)


src/middlewared/middlewared/pytest/unit/utils/test_passdb.py
```diff
@@ -0,0 +1,217 @@
...
+    assert util_passdb.user_smbhash_to_nt_pw('smbuser', smb_hash_legacy) == nt_hash
+    assert util_passdb.user_smbhash_to_nt_pw('smbuser', nt_hash) == nt_hash
+
+
+@pytest.mark.parametrize('user_data,expected', [
+    ({'locked': False, 'password_disabled': False}, util_passdb.UserAccountControl.NORMAL_ACCOUNT),
+    ({'locked': True, 'password_disabled': False}, LOCKED_ACCOUNT),
+    ({'locked': False, 'password_disabled': True}, DISABLED_ACCOUNT),
+    ({'locked': True, 'password_disabled': True}, DISABLED_ACCOUNT | LOCKED_ACCOUNT),
+])
```
bmeagherix, line 187: Nothing wrong with this, but wondering why `user_entry_to_uac_flags` doesn't also handle `DONT_EXPIRE_PASSWORD`.  If it's not applicable maybe add a comment to that effect.

src/middlewared/middlewared/plugins/smb_/passdb.py
```diff
@@ -15,229 +20,71 @@ class Config:
...
-
-            # Remove any potential negative cache entries
-            try:
-                await self.middleware.call('idmap.gencache.del_idmap_cache_entry', {
-                    'entry_type': 'UID2SID',
-                    'entry': user['uid']
-                })
-            except MatchNotFound:
-                pass
+    def update_passdb_user(self, user: UserEntry):
```
bmeagherix, line 34: Maybe add a comment to clarify usage.

`update_passdb_entry` called below states "Update an existing passdb entry", but this code appears to support an `existing_entry` of None, which suggests adding a **new** entry.  Is this correct?

src/middlewared/middlewared/plugins/smb_/util_passdb.py
```diff
@@ -0,0 +1,543 @@
...
+    """
+    Retrieve SAMU data based on passdb RID entry and parse bytes into a PDBEntry
+    object.
+
+
+    """
+    try:
+        if (pdb_bytes := hdl.get(f'{USER_PREFIX}{b64decode(tdb_val)[:-1].decode()}')) is None:
+            # malformed passdb entry. Shouldn't happen we'll return None to force
+            # rewrite
```
bmeagherix, line 274: If this shouldn't happen, would it make sense to log so that we'll know that it _**has**_ happened,  (Either here or where ever we handle the `PassdbMustReinit`.)

src/middlewared/middlewared/plugins/smb_/util_passdb.py
```diff
@@ -0,0 +1,543 @@
...
+    return _unpack_pdb_bytes(b64decode(pdb_bytes))
+
+
+def passdb_entries(as_dict: bool = False) -> Iterable[PDBEntry, dict]:
+    """ Iterate the passdb.tdb file
+
+    Each SMB user contains two TDB entries. One that maps a RID value to the username
+    and the other maps the username to a samu buffer. These are both written simultaneously
+    under a transaction lock and so we should never be in a situation where they are
+    incosistent; however if we are unable to look up a USER entry based on the username
```
bmeagherix, line 288: Typo: incosistent

src/middlewared/middlewared/plugins/smb_/util_passdb.py
```diff
@@ -0,0 +1,543 @@
...
+                    key=f'{RID_PREFIX}{rid:08x}'
+                )
+            ])
+        except RuntimeError:
+            # entries do not exist
+            pass
+
+
+def update_passdb_entry(entry: PDBEntry) -> None:
+    """ Update an existing passdb entry
```
bmeagherix, line 410: Hmmm.  Again are we _only_ doing an update (and not an insert)?  That seems to be at odds with handling a missing `PASSDB_PATH`.  Seems like its insert or update.

bmeagherix (approved):

src/middlewared/middlewared/pytest/unit/utils/test_passdb.py
anodos325: We don't have a corresponding user field _yet_.

PR #14180, opened 2024-08-10
NAS-130530 / 25.04 / Do not allow VMs to consume zvols on boot-pool

Qubad786:
## Problem

There have been users who have created zvols with `zfs create` and then consumed them in VMs where the zvols resided in boot-pool.

## Solution

Enhance validation to not allow users to specify zvols which reside on boot-pool.

PR #14337, opened 2024-08-23
NAS-130796 / 25.04 / Do not allow iSCSI extents to use zvols on boot-pool

bmeagherix:
This PR was 100% inspired by PR #14180

Add private API `boot.is_boot_pool_path` and call to prevent an iSCSI extent from being based on a ZVOL on the boot-pool.

### DIFF
--- a/src/middlewared/middlewared/plugins/boot.py
+++ b/src/middlewared/middlewared/plugins/boot.py
@@ -329,6 +329,10 @@ class BootService(Service):
         if properties:
             await self.middleware.call('zfs.pool.update', BOOT_POOL_NAME, {'properties': properties})
 
+    @private
+    async def is_boot_pool_path(self, path):
+        return path.startswith(f'/dev/zvol/{await self.middleware.call("boot.pool_name")}/')
+
 
 async def on_config_upload(middleware, path):
     await middleware.call('boot.update_initramfs', {'database': path})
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/boot.py
```diff
@@ -329,6 +329,10 @@ async def check_update_ashift_property(self):
         if properties:
             await self.middleware.call('zfs.pool.update', BOOT_POOL_NAME, {'properties': properties})
 
+    @private
+    async def is_boot_pool_path(self, path):
+        return path.startswith(f'/dev/zvol/{await self.middleware.call("boot.pool_name")}/')
```
yocalebo, line 334: No reason to go through self.middleware.call. Just call await self.pool_name()

src/middlewared/middlewared/plugins/boot.py
```diff
@@ -331,7 +331,7 @@ async def check_update_ashift_property(self):
 
     @private
     async def is_boot_pool_path(self, path):
-        return path.startswith(f'/dev/zvol/{await self.middleware.call("boot.pool_name")}/')
+        return path.startswith(f'/dev/zvol/{await self.pool_name()}/')
```
mgrimesix, line 334: Do you still need the `await`?

mgrimesix (approved):

anodos325 (approved):

PR #14338, opened 2024-08-23
NAS-130769 / 25.04 / Default to accept policy for forward chain

Qubad786:
## Context

Docker after starting changes the default policy of forward chain to `DROP` which results in networking for VMs getting broken, we switch it back to `ACCEPT` after docker has started to restore connectivity to VMs.


yocalebo:
This is what we have in `docker/daemon.json.py`. Seems we already knew about this issue?
```
    base = {
        'data-root': data_root,
        'exec-opts': ['native.cgroupdriver=cgroupfs'],
        'iptables': True,  # FIXME: VMs connectivity would be broken
        'storage-driver': 'overlay2',
    }

sonicaj (approved):

PR #14339, opened 2024-08-24
Add hybrid mode to UPS

FoxtrotCZ:
Implements this feature request: https://forums.truenas.com/t/ups-server-additional-shutdown-mode/11257

Hybrid mode:
When UPS is on battery Shutdown Timer is started. If low battery is reached shutdown is executed immediately. 

WebUI pull request: https://github.com/truenas/webui/pull/10549


FoxtrotCZ:
I just found out that shutdown mode "UPS goes on battery" actually acts the same even if not explicitly stated.
If it detects low battery (when on battery) it executes immediate shutdown.
So hybrid mode is not needed.
I am closing this.

PR #14348, opened 2024-08-27
NAS-130820 / 25.04 / Invalidate pool status alerts when pool states change

themylogin:
It seems that we can't rely solely on ZFS events for such an important matter.


src/middlewared/middlewared/alert/source/volume_status.py
```diff
@@ -81,3 +96,14 @@ async def enabled(self):
             return status in ("MASTER", "SINGLE")
 
         return True
+
+    def pools_states(self):
```
yocalebo, line 100: We already have a function that does this. It's in `utils/zfs.py` named `query_fast_imported_impl`

yocalebo (changes requested):

yocalebo (approved):

PR #14359, opened 2024-08-28
NAS-130471 / 25.04 / Add an event source for app statistics

sonicaj:
This PR adds an event source for retrieving app statistics from docker.

### DIFF
--- a/src/middlewared/middlewared/plugins/apps/stats.py
+++ b/src/middlewared/middlewared/plugins/apps/stats.py
@@ -1,4 +1,3 @@
-import copy
 import time
 
 from middlewared.event import EventSource
@@ -26,7 +25,7 @@ class AppStatsEventSource(EventSource):
                 'stats',
                 Str('app_name'),
                 Int('cpu_usage', description='Percentage of cpu used by an app'),
-                Int('memory', description='Current memory( in bytes) used by an app'),
+                Int('memory', description='Current memory(in bytes) used by an app'),
                 List(
                     'networks',
                     items=[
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/apps/stats.py
```diff
@@ -0,0 +1,77 @@
...
+
+        old_projects_stats = list_resources_stats_by_project()
+        interval = self.arg['interval']
+        time.sleep(interval)
+
+        while not self._cancel_sync.is_set():
+            try:
+                project_stats = list_resources_stats_by_project()
+                self.send_event(
+                    'ADDED', fields=normalize_projects_stats(copy.deepcopy(project_stats), old_projects_stats, interval)
```
yocalebo, line 62: Why do we need to `deepcopy(project_stats)`?

src/middlewared/middlewared/plugins/apps/stats.py
```diff
@@ -0,0 +1,77 @@
...
+                )
+                old_projects_stats = project_stats
+                time.sleep(interval)
+            except Exception:
+                if self.middleware.call_sync('docker.config')['pool'] is None:
+                    return
+                if self.middleware.call('service.started', 'docker') is False:
```
yocalebo, line 69:
```suggestion
                if self.middleware.call_sync('service.started', 'docker') is False:
```

src/middlewared/middlewared/plugins/apps/stats.py
```diff
@@ -0,0 +1,77 @@
...
+                    self.middleware.logger.error('Unable to retrieve app stats as docker service has been stopped')
```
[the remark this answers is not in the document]
sonicaj: This would be quite an edge case, docker service would only be stopped when apps are unconfigured. So this would only happen when someone does it manually

yocalebo (approved):
Nice!

PR #14370, opened 2024-08-29
NAS-130861 / 25.04 / Define custom address pool capacity for docker

sonicaj:
This commit defines custom address pools to optimize Docker network creation capacity. The pools are set to 172.30.0.0/16 and 172.31.0.0/16 with a /27 subnet size, allowing for up to 4096 Docker networks, each with 32 IP addresses. This change significantly increases the number of available Docker networks, which is crucial for adding large number of apps which are meant to run simultaneously because by default behaviour the capacity would roughly be at 32 networks meaning roughly 32 apps.


stavros-k (approved):

aiden3c (approved):

sonicaj:
backport

PR #14372, opened 2024-08-29
NAS-130867 / 25.04 / Simplify {TCPIP/UnixSock}Origin to ConnectionOrigin

yocalebo:
I broke `app.query` recently because we're passing `@pass_app` objects around. After investigating this for 2 days, I realized we can simplify all of this and get rid of public methods like `websocket_local_ip` and `websocket_interface` and get the same information directly from the `app.origin` object.

This removes all the Origin classes (and subclass) and creates a singular `ConnectionOrigin` object. I moved over the logic from the other classes to this one but added a few convenient properties (`is_ha_connection`) so it's easier to use. This removes the necessity for importing these origin classes everywhere and doing `isinstance` calls. I also get rid of the `read_proc_net` and `get_remote_addr` functions. The former is actually deprecated and the latter is unnecessarily expensive.

CI runs for both HA and Single node passed with no obvious regressions: ([here](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/345/) and [here](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/346/#showFailuresLink))


src/middlewared/middlewared/plugins/auth.py
```diff
@@ -161,13 +160,14 @@ def dump(self):
...
 def is_internal_session(session):
-    if isinstance(session.app.origin, UnixSocketOrigin) and session.app.origin.uid == 0:
-        return True
-
-    if isinstance(session.app.authenticated_credentials, TrueNasNodeSessionManagerCredentials):
-        return True
-
-    return False
+    return any((
+        all((
```
themylogin, line 164: Why `any` and `all`?

yocalebo: the first `all` requires 2 conditionals to be True, which is the first boolean in the tuple to `any`. It's the same as:
```
if (true and true) or (true)

Comment (themylogin) [reply src/middlewared/middlewared/plugins/auth.py:164]: `all((x, y))` will evaluate `y` even if `x` is false while `x and y` will only evaluate `x`. Similarly, `any((x, y))` will evaluate `y` even if `x` is true, while `x or y` will only evaluate `x`. These differences can lead to bugs, e.g. https://github.com/truenas/middleware/pull/14281

Comment (yocalebo) [reply src/middlewared/middlewared/plugins/auth.py:164]: Very well, I'll fix

Comment (yocalebo) [reply src/middlewared/middlewared/plugins/auth.py]: fixed

File: src/middlewared/middlewared/utils/rate_limit/cache.py [hunk 2]
```diff
@@ -65,25 +64,25 @@ def rate_limit_exceeded(self, method_name: str, ip: str) -> bool:
...
-
-        ip, port = origin.addr, origin.port
-        if any((ip is None, port is None)) or is_ha_connection(ip, port):
-            # Short-circuit if:
-            # 1. if the IP address is None
-            # 2. OR the port is None
-            # 3. OR the origin of the request is from our HA P2P heartbeat
-            #   connection
+        try:
+            if any((
```
themylogin, line 71: One `any(` left that can be converted to a boolean expression

themylogin (approved):
Thank you!

PR #14385, opened 2024-08-30
NAS-130886 / 25.04 / stop calling websocket_local_ip

yocalebo:
With the changes introduced here: https://github.com/truenas/middleware/pull/14372, we no longer need to pass the `app` object around and call `websocket_local_ip`.

### DIFF
--- a/src/middlewared/middlewared/plugins/apps/crud.py
+++ b/src/middlewared/middlewared/plugins/apps/crud.py
@@ -93,9 +93,17 @@ class AppService(CRUDService):
             return filter_list([], filters, options)
 
         extra = options.get('extra', {})
+        host_ip = extra.get('host_ip')
+        if not host_ip:
+            try:
+                if app.origin.is_tcp_ip_family:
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/apps/crud.py
```diff
@@ -93,9 +93,17 @@ def query(self, app, filters, options):
             return filter_list([], filters, options)
 
         extra = options.get('extra', {})
+        host_ip = extra.get('host_ip')
+        if not host_ip:
+            try:
+                if app.origin.is_tcp_ip_family:
+                    host_ip = app.origin.loc_addr
```
bmeagherix, line 100: Oh, below you handled IPv6.  Do we need to do so here too?

yocalebo: @sonicaj do we need to put ipv6 address in brackets here for host_ip key in apps?

PR #14390, opened 2024-08-30
NAS-130696 / 25.04 / Fix issues with user.update endpoint

anodos325:
User may have nologin shell causing the attempt to run cp in a
subprocess to fail due to failure to su to the user in question (due to
error in pam_open_session()).

After review of history for why we were subprocessing out to perform
this task, I failed to find a compelling reason why we could not simply
copy them without switching the user context.

This commit adds a new `copytree` filesystem utility that provides the
following features that are relevant to copying home directories.

1. ability to control behavior regarding whether the copy operation
will walk into child datasets. Generally this is not desired because
we have no design or provisions for creating child datasets within a
user home directory. This situation is most likely to arise when users
mistakenly set an incorrect path for a home directory, e.g.
/mnt/tank/share rather than /mnt/tank/share/wilbur

2. awareness of various ACL types on TrueNAS. We need to rigidly preserve
the permissions from the source when writing to a new destination in
case the administrator made a configuration mistake like in (1) in order
to avoid unintentional data disclosure.

3. awareness of the special .zfs directory. Some users enable snapdir
visibility. This means that conventional copy methods can end up
attempting to copy the entirety of the .zfs directory to a new path
resulting in excessive time spent copying and space usage.

4. ability to use block cloning. This speeds up the actual data copy and
reduces the amount of space used on the filesystem when the sysadmin
changes a user home directory path.

5. add symlink race protection by using *_at variants of system calls.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -10,6 +10,7 @@ import wbclient
 from pathlib import Path
 from contextlib import suppress
 
+from dataclasses import asdict
 from middlewared.api import api_method
 from middlewared.api.current import *
 from middlewared.schema import accepts, Bool, Dict, Int, List, Password, Patch, returns, SID, Str
@@ -21,6 +22,7 @@ import middlewared.sqlalchemy as sa
 from middlewared.utils import run, filter_list
 from middlewared.utils.crypto import generate_nt_hash, sha512_crypt
 from middlewared.utils.directoryservices.constants import DSType, DSStatus
+from middlewared.utils.filesystem.copy import copytree, CopyTreeConfig
 from middlewared.utils.nss import pwd, grp
 from middlewared.utils.nss.nss_common import NssModule
 from middlewared.utils.privilege import credential_has_full_admin, privileges_group_mapping
@@ -753,17 +755,13 @@ class UserService(CRUDService):
         # Copy the home directory if it changed
         home_copy = False
         home_old = None
-        if (
-            has_home and
-            'home' in data and
-            data['home'] != user['home'] and
-            not data['home'].startswith(f'{user["home"]}/')
-        ):
-            if had_home:
-                home_copy = True
-                home_old = user['home']
+        if has_home and 'home' in data:
             if data.get('home_create', False):
                 data['home'] = os.path.join(data['home'], data.get('username') or user['username'])
+
+            if had_home and user['home'] != data['home']:
+                home_copy = True
+                home_old = user['home']
 
         # After this point user dict has values from data
         user.update(data)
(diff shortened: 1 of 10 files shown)


bmeagherix (approved):

src/middlewared/middlewared/plugins/account.py
```diff
@@ -1305,7 +1305,7 @@ async def do_home_copy(self, job, home_old, home_new, username, new_mode, uid):
         await perm_job.wait()
 
         command = f"/bin/cp -a {shlex.quote(home_old) + '/' + '.'} {shlex.quote(home_new + '/')}"
-        do_copy = await run(["/usr/bin/su", "-", username, "-c", command], check=False)
+        do_copy = await run(["/usr/bin/su", "-s", "/usr/sbin/bash", "-", username, "-c", command], check=False)
```
bmeagherix, line 1308: Should be /usr/bin/bash

bmeagherix (changes requested):

src/middlewared/middlewared/plugins/account.py
anodos325: Leave it to me to typo I one-line commit. :D

bmeagherix (approved):

src/middlewared/middlewared/plugins/account.py
```diff
@@ -1303,11 +1299,7 @@ async def do_home_copy(self, job, home_old, home_new, username, new_mode, uid):
             })
 
         await perm_job.wait()
-
-        command = f"/bin/cp -a {shlex.quote(home_old) + '/' + '.'} {shlex.quote(home_new + '/')}"
-        do_copy = await run(["/usr/bin/su", "-", username, "-c", command], check=False)
-        if do_copy.returncode != 0:
-            raise CallError(f"Failed to copy homedir [{home_old}] to [{home_new}]: {do_copy.stderr.decode()}")
+        shutil.copytree(home_old, home_new, dirs_exist_ok=True)
```
yocalebo, line 1302: Blocking main event loop with this call here

anodos325: Yeah, I should have switched back to WIP once I realized there were additional issues. Was testing this specifically with CI. I also need to switch user context when performing the action.

src/middlewared/middlewared/pytest/unit/utils/test_statx.py
```diff
@@ -3,7 +3,7 @@
 import stat
```
mgrimesix, line 3: flake8:  stat not used

src/middlewared/middlewared/utils/filesystem/directory.py
```diff
@@ -275,10 +283,15 @@ def dir_fd(self):
...
     @property
-    def request_mask(self):
+    def request_mask(self) -> DirectoryRequestMask:
         return self.__request_mask
 
-    def close(self, force=False):
+    @property
+    def stat(self) -> StructStatx:
+        return self.__stat
+
```
mgrimesix, line 292: flake8: one too many blank lines

mgrimesix (approved):
flake8 complaints.

yocalebo (approved):
pretty cool

PR #14399, opened 2024-09-03
NAS-130907 / 25.04 / Removed REST from test_150_cronjob

aiden3c:
Also updated the file name to remove the numbering

### DIFF
--- a/tests/api2/test_cronjob.py
+++ b/tests/api2/test_cronjob.py
@@ -1,17 +1,16 @@
-import pytest
 from time import sleep
+
+import pytest
 
 from middlewared.test.integration.utils import call, ssh
 
 TESTFILE = '/tmp/.testFileCreatedViaCronjob'
 
 
-@pytest.fixture(scope='module')
-def cronjob_dict():
-    return {}
+def test_create_check_update_verify_cron_job():
+    cronjob_dict = {}
 
-
-def creating_new_cron_job_which_will_run_every_minute(cronjob_dict):
+    # create job
     results = call("cronjob.create", {
         'user': 'root',
         'command': f'touch "{TESTFILE}"',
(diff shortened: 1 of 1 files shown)


tests/api2/test_cronjob.py
```diff
@@ -0,0 +1,52 @@
+import pytest
+from time import sleep
+
+from middlewared.test.integration.utils import call, ssh
+
+TESTFILE = '/tmp/.testFileCreatedViaCronjob'
```
yocalebo, line 6: TESTFILE is unused it seems.

tests/api2/test_cronjob.py
```diff
@@ -0,0 +1,52 @@
...
+def creating_new_cron_job_which_will_run_every_minute(cronjob_dict):
+    results = call("cronjob.create", {
+        'user': 'root',
+        'command': f'touch "{TESTFILE}"',
+        'schedule': {'minute': '*/1'}
+    })
+    cronjob_dict.update(results)
+
+
+def checking_to_see_if_cronjob_was_created_and_enabled(cronjob_dict):
```
yocalebo, line 23: None of these functions will run because they aren't named properly (i.e. pytest dictates test functions need to be named `test_*`)

tests/api2/test_cronjob.py
```diff
@@ -0,0 +1,52 @@
...
+def wait_a_minute():
+    sleep(65)
+
+
+def updating_cronjob_status_to_disabled_updating_command(cronjob_dict):
+    id = cronjob_dict['id']
+    call('cronjob.update', id, {'enabled': False})
+
+
+def checking_that_API_reports_the_cronjob_as_updated(cronjob_dict):
```
yocalebo, line 38: Be sure and name the functions all lower-case.

tests/api2/test_cronjob.py
```diff
@@ -0,0 +1,52 @@
+import pytest
```
yocalebo, line 1: Imports are wrong.

Should be
```
from time import sleep

import pytest

from middlewared.test.integration.utils import call, ssh
```

tests/api2/test_cronjob.py
yocalebo: nope, i'm wrong.

yocalebo (changes requested):
Looking at what this test module does, we can combine all these steps into 2 functions:
1. create cronjob and wait for it to run, verify it etc
2. delete cronjob and cleanup TESTFILE

aiden3c:
> Looking at what this test module does, we can combine all these steps into 2 functions:

I think that I might just make it one so it can use the same `cronjob_dict` object, but I'll make sure to annotate with comments

yocalebo (approved):

PR #14400, opened 2024-09-03
NAS-130898 / 25.04 / Remove REST from test_auth_token.py

creatorcary:


tests/api2/test_auth_token.py
```diff
@@ -1,14 +1,8 @@
 import io
 import json
-
```
yocalebo, line 3: Keep the space here between imports.

yocalebo (approved):

PR #14401, opened 2024-09-03
NAS-130900 / 25.04 / Remove REST from test_350_pool_dataset_quota_alert.py

creatorcary:


tests/api2/test_pool_dataset_quota_alert.py
```diff
@@ -1,17 +1,13 @@
-#!/usr/bin/env python3
-
 # License: BSD
```
yocalebo, line 1: Remove all the comments please.

tests/api2/test_pool_dataset_quota_alert.py
```diff
@@ -1,17 +1,13 @@
...
 
-import sys
-import os
 import re
 import pytest
 from pytest_dependency import depends
```
yocalebo, line 5: Let's put a single space in between builtin imports, 3rd party imports and finally local imports

tests/api2/test_008_pool_dataset_quota_alert.py
```diff
@@ -78,10 +68,7 @@ def test_dataset_quota_alert(request, datasets, expected_alerts):
         results = SSH_TEST("midclt call --job core.bulk alert.process_alerts '[[]]'", user, password)
```
yocalebo, line 68: There is no reason to do `SSH_TEST` and do `midclt call` here. These can be done using `call(`

yocalebo (changes requested):

yocalebo (approved):

PR #14408, opened 2024-09-03
NAS-130981 / 24.10-RC.1 / Allow specifying custom address pools for docker service

sonicaj:
This PR adds changes to make address pools of docker service configurable so user can choose whatever address pools he/she wishes to consume for docker containers.

### DIFF
--- a/src/middlewared/middlewared/plugins/docker/update.py
+++ b/src/middlewared/middlewared/plugins/docker/update.py
@@ -7,6 +7,7 @@ from middlewared.validators import Range
 
 from .state_utils import Status
 from .utils import applications_ds_name
+from .validation_utils import validate_address_pools
 
 
 class DockerModel(sa.Model):
@@ -75,6 +76,11 @@ class DockerService(ConfigService):
             verrors.add('docker_update.pool', 'Pool not found.')
 
         verrors.check()
+
+        if config['address_pools'] != old_config['address_pools']:
+            validate_address_pools(
+                await self.middleware.call('interface.ip_in_use', {'static': True}), config['address_pools']
+            )
 
         if old_config != config:
             if config['pool'] != old_config['pool']:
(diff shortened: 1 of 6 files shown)


src/middlewared/middlewared/plugins/docker/validation_utils.py
```diff
@@ -0,0 +1,45 @@
...
+    verrors.check()
+
+    network_cidrs = set([
+        ipaddress.ip_network(f'{ip["address"]}/{ip["netmask"]}', False)
+        for ip in system_ips
+    ])
+    seen_networks = set()
+    for index, user_network in enumerate(user_specified_networks):
+        base_network = ipaddress.ip_network(user_network['base'], False)
+        subnet_prefix = int(user_network['base'].split('/')[-1])
```
yocalebo, line 19: This is the subnet suffix based on taking last entry after `split()`. Furthermore, you shouldn't need to do this since `base_network` should have the information?

src/middlewared/middlewared/plugins/docker/update.py
```diff
@@ -32,6 +38,13 @@ class Config:
         Str('dataset', required=True),
         Str('pool', required=True, null=True),
         Bool('nvidia', required=True),
+        List('address_pools', items=[
+             Dict(
+                 'address_pool',
+                 IPAddr('base', cidr=True),
+                 Int('size', validators=[Range(min_=1, max_=32)])
```
yocalebo, line 45: Sorry but what is `size` here?

sonicaj: `size` is essentially the number of IP addresses the pool can contain starting from base address. So if it s 24, then each network created can accommodate 256 interfaces

yocalebo: That's incredibly confusing :smile:. So much so that I had to go look it up (shocker, the documentation isn't very forthcoming). The `size` parameter sub-divides the base prefix (using our defaults `/16`) into `/27`'s. A /27 is 30 usable IPs, so each time more than 30 ips are used within a subnet, it'll start to use the next subnet from the base pool.

yocalebo (approved):

PR #14415, opened 2024-09-04
NAS-130905 / 25.04 / Remove REST from test_050_alert.py

creatorcary:
Test: http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/621/


tests/api2/test_alert.py
```diff
@@ -0,0 +1,86 @@
+from time import sleep
+
+import pytest
+from pytest_dependency import depends
+
+from auto_config import pool_name
+from middlewared.test.integration.utils import call, ssh
+
+
+def get_alert_by_id(alert_id):
```
yocalebo, line 10: Can simplify this to:
```
for alert in filter(lambda x: x['id'] == alert_id, call("alert.list")):
    return alert

File: tests/api2/test_alert.py [hunk 2]
```diff
@@ -0,0 +1,86 @@
...
+    results = call("alert.list")
+    for alert in results:
+        if alert['id'] == alert_id:
+            return alert
+
+
+def alert_exists(alert_id):
+    return isinstance(get_alert_by_id(alert_id), dict)
```
yocalebo, line 18: `get_alert_by_id` returns None if the alert isn't found, no need to have this extra `alert_exists` function.

The caller can very simply do:
```
while get_by_alert_id(alert_id) is None:
    sleep(1)

File: tests/api2/test_alert.py [hunk 3]
```diff
@@ -0,0 +1,86 @@
...
+
+
+@pytest.mark.dependency(name='degrade_pool')
+def test_degrading_a_pool_to_create_an_alert(request):
+    get_pool = call("pool.query", [["name", "=", pool_name]])[0]
```
yocalebo, line 23: You can pass `[["name", "=", pool_name]], {"get": True}` and `get_pool` will be returned so you don't have to do `[0]`

tests/api2/test_alert.py
```diff
@@ -0,0 +1,86 @@
...
+
+def test_clear_the_pool_degradation(request):
+    depends(request, ["degrade_pool"], scope="session")
+    ssh(f'zpool clear {pool_name}')
+
+
+def test_verify_the_pool_is_not_degraded(request):
+    depends(request, ["degrade_pool"], scope="session")
+    gptid = request.config.cache.get("alert/gptid", "Not a valid id")
+    stdout = ssh(f'zpool status {pool_name} | grep {gptid}')
```
yocalebo, line 77: Let's not use `ssh` to check zpool status. We can have different methods to return the zpool status using our API.

yocalebo (changes requested):

tests/api2/test_alert.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+ID_PATH = "/dev/disk/by-partuuid/"
+
+
+def get_alert_by_id(alert_id):
+    return next(filter(lambda alert: alert["id"] == alert_id, call("alert.list")), None)
+
+
+@pytest.mark.dependency(name="degrade_pool")
```
themylogin, line 17: Let's not use dependencies where we can use fixtures

tests/api2/test_alert.py
```diff
@@ -0,0 +1,82 @@
...
+    status = call("zpool.status", {"name": pool_name})[pool_name][ID_PATH + gptid]["disk_status"]
+    assert status == "DEGRADED"
+
+
+@pytest.mark.timeout(120)
+@pytest.mark.dependency(name="set_alert_id")
+def test_wait_for_the_alert_and_get_the_id(request):
+    depends(request, ["degrade_pool"], scope="session")
+    call("alert.process_alerts")
+    while True:
```
themylogin, line 37: You can just run `alert.run_source` to test what specific alert source returns. Please see other tests for examples.

themylogin (changes requested):

tests/api2/test_alert.py
creatorcary: Just to be clear, do you want me to turn both tests that are marked as dependencies into fixtures and then call for those fixtures in the other tests? In the case that the fixtures fail, instead of skipping the tests that depend on them they would all fail. The fixtures would also be run every time they're called for in a test.

tests/api2/test_alert.py
creatorcary: Will change

creatorcary: Had to undo this change because `alert.run_source` does not return `id`.

tests/api2/test_alert.py
```diff
@@ -0,0 +1,73 @@
...
+def test_clear_the_pool_degradation(degraded_pool_gptid):
+    ssh(f"zpool clear {pool_name}")
+    status = call("zpool.status", {"name": pool_name})
+    disk_status = status[pool_name]["data"][ID_PATH + degraded_pool_gptid]["disk_status"]
+    assert disk_status != "DEGRADED"
+
+
+@pytest.mark.timeout(120)
+def test_wait_for_the_alert_to_disappear(alert_id):
+    while get_alert_by_id(alert_id) is not None:
```
themylogin, line 72: You can add `call("alert.process_alerts")` here to accelerate the process

themylogin (approved):

tests/api2/test_alert.py
```diff
@@ -0,0 +1,75 @@
...
+        for alert in call("alert.list"):
+            if (
+                alert["source"] == "VolumeStatus" and
+                alert["args"]["volume"] == pool_name and
+                alert["args"]["state"] == "DEGRADED"
+            ):
+                return alert["id"]
+        sleep(1)
+
+
```
themylogin, line 27: Why `autouse` if then used explicitly?

creatorcary: Just signifies clearly that it is a dependency for the whole module, even for hypothetical tests added in the future.

tests/api2/test_alert.py
```diff
@@ -0,0 +1,75 @@
...
+    get_pool = call("pool.query", [["name", "=", pool_name]], {"get": True})
+    gptid = get_pool["topology"]["data"][0]["path"].replace(ID_PATH, "")
+    ssh(f"zinject -d {gptid} -A fault {pool_name}")
+    return gptid
+
+
+@pytest.fixture(scope="module")
+def alert_id(request, degraded_pool_gptid):
+    call("alert.process_alerts")
+    timeout = request.param if hasattr(request, "param") else 120
```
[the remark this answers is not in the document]
creatorcary: This allows a timeout to be passed to the fixture since `wait_for_alert` accepts the optional argument. So currently, nothing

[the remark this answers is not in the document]
themylogin: What's the use case for this?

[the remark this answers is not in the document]
creatorcary: To be honest, [this commit](https://github.com/truenas/middleware/pull/14415/commits/8b7ffe067919cf654484f30851fcf0fbc13bf175) was experimental. Wanted to figure out how to do something like this and it worked so I left it in. Will remove :)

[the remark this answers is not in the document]
creatorcary: But to answer your question
```python
@pytest.mark.parametrize('alert_id', [60], indirect=True)
def test_some_alert_functionality(alert_id): ...
```
in case 120 was too long for whatever reason

tests/api2/test_alert.py
creatorcary: Fair point

creatorcary:
One failed test, don't think it's related though: http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/621/

yocalebo (approved):

PR #14419, opened 2024-09-04
NAS-130995 / 25.04 / Allow retrieving logs of apps in crashed state

Qubad786:
## Context

Currently only those app's logs can be viewed which have containers which are either all running or at least one of them is in running or starting state. However this leads to the problem when an app has containers where all of them have crashed for any reason i.e bad configuration / app issues, unfortunately it is not possible to view logs for those apps yet.

A `CRASHED` state has been introduced which covers the case mentioned above where all the containers in the app if have crashed, that app is marked as `CRASHED` and we would still be able to view logs of such apps.

### DIFF
--- a/src/middlewared/middlewared/plugins/apps/ix_apps/query.py
+++ b/src/middlewared/middlewared/plugins/apps/ix_apps/query.py
@@ -6,7 +6,7 @@ from .docker.query import list_resources_by_project
 from .metadata import get_collective_config, get_collective_metadata
 from .lifecycle import get_current_app_config
 from .path import get_app_parent_config_path
-from .utils import get_app_name_from_project_name, normalize_reference, PROJECT_PREFIX
+from .utils import AppState, ContainerState, get_app_name_from_project_name, normalize_reference, PROJECT_PREFIX
 
 
 COMPOSE_SERVICE_KEY: str = 'com.docker.compose.service'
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/apps/ix_apps/query.py
```diff
@@ -86,12 +86,18 @@ def list_apps(
...
+        exited_containers = 0
         for container in workloads['container_details']:
             if container['state'] == 'starting':
-                state = 'DEPLOYING'
+                state = AppState.DEPLOYING.value
                 break
             elif container['state'] == 'running':
-                state = 'RUNNING'
+                state = AppState.RUNNING.value
+            elif container['state'] == 'exited':
```
yocalebo, line 96: We should add this to the Enum as well. No reason to treat it differently.

src/middlewared/middlewared/plugins/apps/ix_apps/query.py
```diff
@@ -86,12 +86,19 @@ def list_apps(
         # When we stop docker service and start it again - the containers can be in exited
         # state which means we need to account for this.
         state = 'STOPPED'
+        exited_containers = 0
         for container in workloads['container_details']:
-            if container['state'] == 'starting':
-                state = 'DEPLOYING'
+            container_state = ContainerState(container['state'])
+            if container_state == ContainerState.STARTING:
```
yocalebo, line 92: Each one of these if branches instantiates a new instance of the ContainerState class. You iterate the enum and see if the container_state matches the value

src/middlewared/middlewared/plugins/apps/logs.py
```diff
@@ -38,8 +39,8 @@ def __init__(self, *args, **kwargs):
 
     def validate_log_args(self, app_name, container_id):
         app = self.middleware.call_sync('app.get_instance', app_name)
-        if app['state'] not in ('RUNNING', 'DEPLOYING'):
-            raise CallError(f'App "{app_name}" is not running')
+        if AppState(app['state']) not in (AppState.CRASHED, AppState.RUNNING, AppState.DEPLOYING):
```
yocalebo, line 42: Same here, 3 separate instances of the AppState class

Qubad786: Fixed it ✅ (also it would just be 1 instance of the class as the other are declared values)

yocalebo (approved):

PR #14422, opened 2024-09-04
NAS-130937 / 25.04 / Remove REST from test_290_mail.py

creatorcary:
Test: http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/436/


tests/api2/test_mail.py
```diff
@@ -0,0 +1,17 @@
...
+    payload = {
+        "fromemail": "william.spam@ixsystems.com",
+        "outgoingserver": "mail.ixsystems.com",
+        "pass": "changeme",
+        "port": 25,
+        "security": "PLAIN",
+        "smtp": True,
+        "user": "william.spam@ixsystems.com"
+    }
+    call("mail.update", payload)
```
mgrimesix, line 14: Could this be done with context call or try so that we maintain the same configuration exiting the test as we had entering?

creatorcary: The tests we have now do not do this, but I can add it. The problem is if the test fails, restoring the state of the configuration will probably also fail since it would use the same endpoints (`mail.config` and `mail.update`) that the test uses.

mgrimesix: Many of the old tests did not attempt to restore state.  Many (most?) of the updated and new tests do proper state restore.  Yes, if the endpoint is failing, then restoring the state is mute because you weren't able to effect any change in the first place. 
It's not critical, but good practice for a CI test environment.
I'll approve and let you decide which way to go.

mgrimesix (approved):

tests/api2/test_mail.py
creatorcary: Changed it, will keep in mind as I'm going through the "rest" of the tests

mgrimesix: I approve the pun. 😄

PR #14427, opened 2024-09-05
NAS-131017 / 25.04 / Fix zpool.status regression

Qubad786:
## Context

Integration test for `zpool.status` was failing earlier because we were not using `filesystem.listdir` to list disks and hence the relevant mapping could not be generated which eventually led to failure of the test.


yocalebo (changes requested):
This doesn't fix anything. Please see the numerous test failures here: http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/443/testReport/junit/api2/. All of those are failing because `disk.get_unused` is crashing with a `KeyError`.

mgrimesix (approved):

mgrimesix:
Removed accidental approval.

PR #14431, opened 2024-09-05
NAS-130863 / 25.04 / Properly try waiting for default interface to come up before consuming in apps

sonicaj:
This PR adds changes to try and wait for default interface to come up before trying to consume networking in apps, this is especially problematic for bridge interfaces as they can take some time to come in ready state.

### DIFF
--- a/src/middlewared/middlewared/plugins/docker/state_setup.py
+++ b/src/middlewared/middlewared/plugins/docker/state_setup.py
@@ -58,7 +58,7 @@ class DockerSetupService(Service):
     async def validate_interfaces(self):
         default_iface, success = wait_for_default_interface_link_state_up()
         if default_iface is None:
-            raise CallError('Unable to determine default interface.')
+            raise CallError('Unable to determine default interface')
         elif not success:
             raise CallError(f'Default interface {default_iface!r} is not in active state')
 
--- a/src/middlewared/middlewared/plugins/docker/state_setup.py
+++ b/src/middlewared/middlewared/plugins/docker/state_setup.py
@@ -6,10 +6,10 @@ import uuid
 from datetime import datetime
 
 from middlewared.service import CallError, private, Service
+from middlewared.utils.interface import wait_for_default_interface_link_state_up
 
 from .state_utils import (
-    DatasetDefaults, DOCKER_DATASET_NAME, docker_datasets, IX_APPS_MOUNT_PATH, IX_APPS_DIR_NAME,
-    missing_required_datasets,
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/docker/state_setup.py
```diff
@@ -50,6 +50,18 @@ async def validate_fs(self):
                 errno=CallError.EDATASETISLOCKED,
             )
 
+        # What we want to validate now is that the interface on default route is up and running
+        # This is problematic for bridge interfaces which can or cannot come up in time
+        await self.validate_interfaces()
+
+    @private
+    async def validate_interfaces(self):
+        default_iface, success = wait_for_default_interface_link_state_up()
```
yocalebo, line 59: You're blocking the main event loop.

yocalebo (changes requested):

src/middlewared/middlewared/utils/interface.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+IFACE_LINK_STATE_MAX_WAIT: int = 60
+
+
+def get_default_interface() -> str | None:
+    data = []
+    with contextlib.suppress(FileNotFoundError):
+        with open('/proc/net/route', 'r') as f:
+            data = [line.split() for line in f.readlines()]
```
yocalebo, line 13: This is a bit careless and looking at route.h in linux sources, we can determine the default gateway a bit more "properly".
```
RT_UP = 0x0001
RT_GATEWAY = 0x0002
with open('/proc/net/route') as f:
    for lineno, line in filter(lambda x: x[0] > 1, f):
        if (columns := line.strip().split()) and columns[3] == RTF_UP | RTF_GATEWAY:
                return columns[0].strip()

File: src/middlewared/middlewared/utils/interface.py [hunk 3]
```diff
@@ -0,0 +1,43 @@
...
+IFACE_LINK_STATE_MAX_WAIT: int = 60
+RTF_GATEWAY: int = 0x0002
+RTF_UP: int = 0x0001
+
+
+def get_default_interface() -> str | None:
+    data = []
+    with contextlib.suppress(FileNotFoundError):
+        with open('/proc/net/route', 'r') as f:
+            data = [line.split() for line in f.readlines()][1:]
```
yocalebo, line 15: There is no reason to do `data = []` and then do `readlines()[1:]`. You're still reading the entire file into memory and then iterating it again. I'd rather iterate over the file object and return early.

yocalebo (approved):
Thanks!

PR #14441, opened 2024-09-06
NAS-131044 / 25.04 / Add an endpoint for UI to retrieve available space for ix-apps

sonicaj:
This commit adds changes to add an endpoint which UI can consume to retrieve available space in ix-apps dataset so it can show to a user when he wants to install an app that by default X space is available in ix-apps dataset.

### DIFF
--- a/src/middlewared/middlewared/plugins/apps/resources.py
+++ b/src/middlewared/middlewared/plugins/apps/resources.py
@@ -120,7 +120,7 @@ class AppService(Service):
         Returns space available in bytes in the configured apps pool which apps can consume
         """
         await self.middleware.call('docker.state.validate')
-        return (await self.middleware.call('filesystem.statfs', IX_APPS_MOUNT_PATH))['free_bytes']
+        return (await self.middleware.call('filesystem.statfs', IX_APPS_MOUNT_PATH))['avail_bytes']
 
     @accepts(roles=['APPS_READ'])
     @returns(Dict('gpu_choices', additional_attrs=True))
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/apps/resources.py
```diff
@@ -112,6 +112,29 @@ async def ip_choices(self):
...
+            'source': 'NONE',
+            'source_info': None,
+        }
+    ))
+    async def available_space(self):
+        """
+        Returns space available in the configured apps pool which apps can consume
+        """
+        await self.middleware.call('docker.state.validate')
+        return (await self.middleware.call(
```
yocalebo, line 132: We need to get this data using a less expensive operation. We're using the process pool here when I fear the UI team is going to call this endpoint  _a lot_. I'd rather do `filesystem.statfs`

src/middlewared/middlewared/plugins/apps/resources.py
```diff
@@ -112,6 +113,15 @@ async def ip_choices(self):
...
         }
 
+    @accepts(roles=['CATALOG_READ'])
+    @returns(Int())
+    async def available_space(self):
+        """
+        Returns space available in bytes in the configured apps pool which apps can consume
+        """
+        await self.middleware.call('docker.state.validate')
+        return (await self.middleware.call('filesystem.statfs', IX_APPS_MOUNT_PATH))['free_bytes']
```
yocalebo, line 123: This is a nit but we should use the `avail_bytes` key since this an `available_space` endpoint.

yocalebo (approved):
Thanks!

PR #14447, opened 2024-09-06
NAS-131061 / 25.04 / enable SA-based xattrs by default on new pools

anodos325:
These have been used pretty successfully for a number of years in Core and SCALE. We can just enable by default these days and get the corresponding performance boost.


yocalebo (commented):
Can we add to an existing API test that verifies this property is set at pool creation? Probably in test_006 would be sufficient.

yocalebo (approved):

PR #14449, opened 2024-09-07
NAS-131063 / 25.04 / Make sure docker networking is not broken after interface sync

sonicaj:
This PR adds changes to introduce a docker network crud from where we can retrieve configured docker networks and then we make sure that these interfaces are marked as internal because currently `interface.sync` is unconfiguring docker based interfaces which results in apps networking loss each time interfaces are synced.

### DIFF
--- a/src/middlewared/middlewared/plugins/docker/network.py
+++ b/src/middlewared/middlewared/plugins/docker/network.py
@@ -1,6 +1,6 @@
 from middlewared.plugins.apps.ix_apps.docker.networks import list_networks
 from middlewared.schema import Dict, Str
-from middlewared.service import CRUDService, filterable
+from middlewared.service import CRUDService, filterable, private
 from middlewared.utils import filter_list
 
 
@@ -42,3 +42,7 @@ class DockerNetworkService(CRUDService):
             })
 
         return filter_list(networks, filters, options)
+
+    @private
+    def interfaces_mapping(self):
+        return [f'br-{network["short_id"]}' for network in self.query()]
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/docker/network.py
```diff
@@ -0,0 +1,53 @@
...
+        additional_attrs=True,
+    )
+
+    @filterable
+    def query(self, filters, options):
+        """
+        Query all docker networks
+        """
+        if not self.middleware.call_sync('docker.state.validate', False):
+            return filter_list([], filters, options)
```
yocalebo, line 34: No reason to call filter_list here. Just return an empty array.

src/middlewared/middlewared/plugins/docker/network.py
```diff
@@ -0,0 +1,53 @@
...
+                    'ipam', 'labels', 'created', 'driver', 'id', 'name', 'scope', 'short_id',
+                )
+            })
+
+        return filter_list(networks, filters, options)
+
+    @private
+    def interfaces_mapping(self):
+        try:
+            return [f'br-{network["short_id"]}' for network in self.query()]
```
yocalebo, line 49: Can there ever be non bridge interfaces? This will prefix "br-" for all networks even if there isn't a bridge on the host.

src/middlewared/middlewared/plugins/docker/network.py
yocalebo: Ahh right

yocalebo (approved):

PR #14455, opened 2024-09-09
NAS-130903 / 25.04 / Remove REST calls in test_140_core

aiden3c:
[Passing test](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/510/consoleFull) (look for test_140_core)

Just to be pedantic, I have another test running after updating the file/test name. That's [here](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/556/console).


tests/api2/test_core_download.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def test_get_download_for_config_dot_save():
+    # ping core for sanity
+    assert call('core.ping') == 'pong'
+
+    # set up core download
+    results = call('core.download', 'config.save', [], 'freenas.db')
+    url = results[1]
+    job_id = results[0]
```
creatorcary, line 15: This can be unpacked as `job_id, url = call(...)`

tests/api2/test_core_download.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+    # is the job running?
+    results = call('core.get_jobs', [['id', '=', job_id]], {"get": True})
+    assert results['state'] == 'RUNNING', results.text
```
creatorcary: `results` is a dictionary now so it won't have a `text` attribute (same on line 30)

tests/api2/test_core_download.py
```diff
@@ -0,0 +1,29 @@
+from os import stat
+
+from urllib.request import urlretrieve
+from middlewared.test.integration.utils.client import truenas_server
+from middlewared.test.integration.utils import call
+
+
+def test_get_download_for_config_dot_save():
+    # ping core for sanity
+    assert call('core.ping') == 'pong'
```
yocalebo, line 10: No reason to call core.ping here.

creatorcary (changes requested):

tests/api2/test_core_download.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+    # set up core download
+    job_id, url = call('core.download', 'config.save', [], 'freenas.db')
+
+
+    # is the job running?
+    results = call('core.get_jobs', [['id', '=', job_id]], {"get": True})
```
yocalebo, line 18: This is a race. Starting the job and then asserting that's running is prone to failures.

tests/api2/test_core_download.py
```diff
@@ -0,0 +1,29 @@
...
+    assert results['state'] == 'RUNNING'
+
+
+    # download from URL
+    rv = urlretrieve(f'http://{truenas_server.ip}{url}')
```
yocalebo, line 23: Why would you stat the size before the job returns `SUCCESS`?

tests/api2/test_core_download.py
```diff
@@ -0,0 +1,29 @@
...
+    assert stat(rv[0]).st_size > 0
+
+
+    # verify success
+    results = call('core.get_jobs', [['id', '=', job_id]], {"get": True})
```
yocalebo, line 28: We can simplify this by doing `job=True` and then stating the file afterwards.

yocalebo (changes requested):

creatorcary (approved):

tests/api2/test_core_download.py
aiden3c: This (and the ping) were there and structured that way from the old test, I had assumed there was some specific reason for it. For this I was assuming that the download job is set to "running" while there's a pending URL to download from. Will update accordingly.

tests/api2/test_core_download.py
aiden3c: Honestly since this is checking the success of the job that's now set up to wait until it's finished, we can remove this then just check for success on the job after it runs. Then we can stat the file.

themylogin (changes requested):
Let's be consistent across our tests. Please see https://github.com/truenas/middleware/blob/master/tests/api2/test_rest_api_download.py#L54 as an example on how to download from our API

aiden3c:
New [passing tests](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/598/console)!

tests/api2/test_core_download.py
```diff
@@ -0,0 +1,13 @@
...
+from middlewared.test.integration.utils import call
+
+
+def test_get_download_for_config_dot_save():
+    # set up core download
+    job_id, url = call('core.download', 'config.save', [], 'freenas.db')
+
+    # download from URL
+    rv = requests.get(f'http://{truenas_server.ip}{url}')
+    assert len(rv.content) > 0
```
themylogin, line 13: What if response body is non-empty, but status is 500?

themylogin (changes requested):

tests/api2/test_core_download.py
aiden3c: Gotcha, made sure to check for 200

yocalebo:
@themylogin ping

themylogin (approved):

PR #14459, opened 2024-09-09
NAS-131078 / 25.04 / Separate logging for app migration

sonicaj:
This PR adds changes to have a separate log file for app migrations and issue an alert as well if there are some apps which fail to migrate.

### DIFF
--- a/src/middlewared/middlewared/alert/source/applications.py
+++ b/src/middlewared/middlewared/alert/source/applications.py
@@ -6,8 +6,8 @@ class FailuresInAppMigrationAlertClass(AlertClass, OneShotAlertClass):
 
     category = AlertCategory.APPLICATIONS
     level = AlertLevel.ERROR
-    title = 'Application(s) failed to migrate from kubernetes'
-    text = 'App(s) failed to migrate from dragonfish, please check /var/log/app_migrations.log for more details.'
+    title = 'App(s) failed to migrate'
+    text = 'App(s) failed to migrate, please check /var/log/app_migrations.log for more details.'
 
     async def create(self, args):
         return Alert(FailuresInAppMigrationAlertClass, args)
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/alert/source/applications.py
```diff
@@ -1,6 +1,21 @@
 from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, Alert, OneShotAlertClass
 
 
+class FailuresInAppMigrationAlertClass(AlertClass, OneShotAlertClass):
+    deleted_automatically = False
+
+    category = AlertCategory.APPLICATIONS
+    level = AlertLevel.ERROR
+    title = 'Application(s) failed to migrate from kubernetes'
+    text = 'App(s) failed to migrate from dragonfish, please check /var/log/app_migrations.log for more details.'
```
yocalebo, line 10: Remove the `dragonfish` verbiage, please. Just need to state that migration failed.

src/middlewared/middlewared/alert/source/applications.py
```diff
@@ -1,6 +1,21 @@
 from middlewared.alert.base import AlertClass, AlertCategory, AlertLevel, Alert, OneShotAlertClass
 
 
+class FailuresInAppMigrationAlertClass(AlertClass, OneShotAlertClass):
+    deleted_automatically = False
+
+    category = AlertCategory.APPLICATIONS
+    level = AlertLevel.ERROR
+    title = 'Application(s) failed to migrate from kubernetes'
```
yocalebo, line 9: Don't think we need to list kubernetes here either. It's an implementation detail.

src/middlewared/middlewared/alert/source/applications.py
yocalebo, line 10: You have `App(s)` and `Application(s)`. We should standardize on 1 or the other but not both.

yocalebo (changes requested):

yocalebo (approved):

PR #14460, opened 2024-09-09
NAS-131010 / 25.04 / Fix errors with test_bootenv

aiden3c:
After investigating why the bootenv tests are failing, I found that the target bootenv is set up to activate on next boot. The `active` key becomes `R` to indicate this reboot requirement. Since there's no way to actually test bootenv activation (without rebooting the test VM which opens a whole can of worms), the test has instead been rewritten to check creating/updating/deleting bootenvs via API, then dataset promotion using the raw ZFS commands.

As always, [passing test](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/561/consoleFull) (look for `test_bootenv`)


Qubad786 (commented):
Do you think it perhaps make sense to still activate a different BE and validate it has been activated and then re-activate current BE again and again ensure it is activated?
Motivation would be to validate that activation logic works as intended

aiden3c:
> Do you think it perhaps make sense to still activate a different BE and validate it has been activated and then re-activate current BE again and again ensure it is activated? Motivation would be to validate that activation logic works as intended

Yeah, that's probably a good idea. I initially hadn't as we can't "guarantee" that on reboot the pending activations will work, but we can see that it's pending an activation for the next reboot.

aiden3c:
Running a [new test](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/571/console)

aiden3c:
Test is passing!

Qubad786 (approved):

PR #14467, opened 2024-09-09
NAS-130931 / 25.04 / Replace REST calls with websocket calls

bmeagherix:
Replace REST calls with websocket calls

In addition, made **extensive** use of (class scoped) fixtures, for better cleanup and also to facilitate reuse.

---
CI run with passing tests from this file [here](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/581/).

### DIFF
--- a/tests/api2/test_260_iscsi.py
+++ b/tests/api2/test_260_iscsi.py
@@ -1,26 +1,24 @@
-#!/usr/bin/env python3
-
-# License: BSD
-
-import os
-import pytest
 import random
 import string
-import sys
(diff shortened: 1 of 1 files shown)


yocalebo (approved):
Nice! Should we rename? That's the paradigm we've been taking with the other tests.

bmeagherix:
> Nice! Should we rename? That's the paradigm we've been taking with the other tests.

I'll do the rename of all iSCSI tests together / later.  (Want to test such.)

PR #14475, opened 2024-09-10
NAS-130929 / 25.04 / Removed REST from test_audit_nfs

aiden3c:
[Passing tests](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/602/console)!


yocalebo:
Please note to reviewers: I will be adding back the REST related audit tests but doing so in their own sub-directory. This is to ease the removal of said tests when REST goes away in some distant future.

PR #14476, opened 2024-09-10
NAS-131087 / 13.3 / Also account for pylibzfs-initialed scrubs when checking last pool scrub datetime

themylogin:


src/middlewared/middlewared/plugins/pool.py
```diff
@@ -43,7 +43,7 @@
 
 GELI_KEYPATH = '/data/geli'
 
-RE_HISTORY_ZPOOL_SCRUB_CREATE = re.compile(r'^([0-9\.\:\-]{19})\s+zpool (scrub|create)', re.MULTILINE)
```
yocalebo, line 46: The name of the regex changes which breaks this.

yocalebo (changes requested):

yocalebo (approved):

themylogin:
time 0:10

PR #14479, opened 2024-09-10
NAS-130745 / 25.04 / Update smartctl usage to use JSON

aiden3c:
I've updated our smartctl usage throughout middleware to use the JSON flag to get serialized data. Unfortunately I haven't been able to run this through a full gauntlet of SMART test types for different drives. Going through the smartmontools source code and doing what I can with different drive types I had available, I've gotten the JSON parsing as accurate as I could.

This is something that might end up breaking, but getting tickets will let us fix any assumed/implicit keys we're accessing.

[Passing tests](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/694/), but the majority are our `CI build` tests, which are passing.


src/middlewared/middlewared/plugins/disk.py
```diff
@@ -143,9 +143,7 @@ async def disk_extend(self, disk, context):
             if await self.middleware.call('truenas.is_ix_hardware'):
                 disk['supports_smart'] = True
             else:
-                disk['supports_smart'] = disk['name'].startswith('nvme') or bool(RE_SMART_AVAILABLE.search(
-                    await self.middleware.call('disk.smartctl', disk['name'], ['-a'], {'silent': True}) or ''
-                ))
+                disk['supports_smart'] = disk['name'].startswith('nvme') or await self.middleware.call('disk.smartctl', disk['name'], ['-a', '--json=c'], {'silent': True})['smart_support']['available']
```
themylogin, line 146: Looks like this will fail, you're awaiting on `self.middleware.call('disk.smartctl', disk['name'], ['-a', '--json=c'], {'silent': True})['smart_support']['available']`.

src/middlewared/middlewared/plugins/smart.py
```diff
@@ -21,159 +22,142 @@
...
-            test = {
-                "num": int(line[1:3].strip()),
-                "description": line[5:24].strip(),
-                "status_verbose": line[25:54].strip(),
-                "remaining": int(remaining.strip()) / 100,
-                "lifetime": int(lifetime.strip()),
-                "lba_of_first_error": line[77:].strip(),
-            }
+                # remaining_percent is in the dict only if the test is in progress (status value & 0x0f)
+                if remaining := entry["status"]["value"] & 0x0f:
```
themylogin, line 49: Can we have this as a constant that you would copy from smartmontools source? Or is this a hex literal there also?

src/middlewared/middlewared/plugins/smart.py
```diff
@@ -21,159 +22,142 @@
...
-                    "sct": sct,
-                    "code": code,
+                    "num": index,
+                    "description": entry["self_test_code"]["string"],
+                    "status_verbose": entry["self_test_result"]["string"],
+                    "power_on_hours": entry["power_on_hours"],
+                    "failing_lba": lba,
+                    "nsid": entry.get("nsid"),
+                    "seg": entry.get("segment"),
+                    "sct": entry.get("status_code_type") or 0x0,
```
themylogin, line 90: `sct` was a string before, but `0x0` is an integer

Maybe we can take advantage of this, define type classes and make this method return `list[AtaSelfTestLog | NvmeSelfTestLog | ScsiSelfTestLog]`?

src/middlewared/middlewared/pytest/unit/plugins/test_smart.py
```diff
@@ -38,45 +67,64 @@ def test__parse_smart_selftest_results__ataprint__1():
 
 
 def test__parse_smart_selftest_results__ataprint__2():
-    assert parse_smart_selftest_results(textwrap.dedent("""\
-        Num  Test_Description    Status                  Remaining  LifeTime(hours)  LBA_of_first_error
-        # 1  Offline             Self-test routine in progress 100%         0         -
-    """)) == [
+    data = json.loads("""\
```
themylogin, line 70: Why `json.loads`? This is a perfectly fine Python literal itself

src/middlewared/middlewared/pytest/unit/plugins/test_smart.py
```diff
@@ -1,23 +1,52 @@
...
+            "error_count_total": 0,
+            "error_count_outdated": 0
+            }
+        }
+}
+        """))
+    assert parse_smart_selftest_results(data) == [
         {
-            "num": 1,
+            "num": 0,
```
themylogin, line 49: Why do results change?

src/middlewared/middlewared/pytest/unit/plugins/test_smart.py
```diff
@@ -38,45 +67,64 @@ def test__parse_smart_selftest_results__ataprint__1():
...
+        """)
+    assert parse_smart_selftest_results(data) == [
         {
-            "num": 1,
+            "num": 0,
             "description": "Offline",
             "status": "RUNNING",
             "status_verbose": "Self-test routine in progress",
-            "remaining": 1.0,
+            "remaining": 100,
```
themylogin, line 102: Let's not change this please unless absolutely necessary

src/middlewared/middlewared/utils/disks.py
```diff
@@ -17,40 +17,8 @@ class Disk:
...
-    reg = re.search(r'Temperature Sensor [0-9]+:\s+([0-9]+) Celsius', stdout, re.M)
-    if reg:
-        return int(reg.group(1))
-
-    # scsiprint.cpp
-
-    reg = re.search(r'Current Drive Temperature:\s+([0-9]+) C', stdout, re.M)
-    if reg:
-        return int(reg.group(1))
+def parse_smartctl_for_temperature_output(json) -> Optional[int]:
```
themylogin, line 20: Does this account for all the possible cases listed above?

themylogin (changes requested):
Good job! There are some details I want to pay attention too.

src/middlewared/middlewared/utils/disks.py
aiden3c: It does! The JSON output for the three drive types all have ["temperature"]["current"] at the root of their JSON output.

src/middlewared/middlewared/pytest/unit/plugins/test_smart.py
aiden3c: Between ata, nvme, and iscsi, there are inconsistencies between the test number being 0 indexed (or included at all in the JSON). Sometimes even in the same drive type (iirc iscsi's normal output and JSON output are 0 and 1 indexed respectively lol).
This was to standardize it all to zero indexed

src/middlewared/middlewared/pytest/unit/plugins/test_smart.py
aiden3c: That's actually a really good point, I'll update those tests to just use a dict

src/middlewared/middlewared/plugins/smart.py
aiden3c: It's also a hex literal in the source, the last byte has the progress encoded in it.

src/middlewared/middlewared/pytest/unit/plugins/test_smart.py
aiden3c: I swear there was some valid reason I did this, I'll re-evaluate it again lol

src/middlewared/middlewared/plugins/disk_/smart_attributes.py
```diff
@@ -38,7 +38,7 @@ async def smart_attributes(self, name):
         """
         Returns S.M.A.R.T. attributes values for specified disk name.
         """
-        output = json.loads(await self.middleware.call('disk.smartctl', name, ['-A', '-j']))
+        output = json.loads(await self.middleware.call('disk.smartctl', name, ['-A', '--json=c']))
 
         if 'ata_smart_attributes' in output:
```
yocalebo, line 43: This method is only returning ATA based smart attributes but we allow returning smart attributes for SAS and NVMe. While you're here, you need to go ahead and add these drives types to this method.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/smart.py
aiden3c: Unfortunately there's some implicit keys that are sometimes there or sometimes don't have a value. It'd be useful for the areas that consume those test logs, but I think it would get weirdly complicated to have type classes when the schema isn't consistent. 
It actually might be worth putting into another ticket though to look into further.

themylogin: The schema is consistent, the old code was returning the dictionaries with consistent sets of keys.

I propose to further formalize these dictionaries into dataclasses because I see that there is a confusion on which type `sct` field should have.

src/middlewared/middlewared/plugins/disk_/smart_attributes.py
aiden3c: It doesn't look like SAS or NVMe drives report smart attribute information from smartctl sadly : (

aiden3c:
New [passing tests](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/788/console)!

src/middlewared/middlewared/api/v25_04_0/smartctl.py
```diff
@@ -0,0 +1,35 @@
+from typing import Any
+
+from middlewared.api.base import BaseModel
+
+__all__ = ["AtaSelfTest", "NvmeSelfTest", "ScsiSelfTest"]
+
+class AtaSelfTest(BaseModel):
```
themylogin, line 7: flake8 does not complain here, because new error count is less than previous error count, but every class declaration should be prefixed by two newlines.

src/middlewared/middlewared/plugins/disk.py
```diff
@@ -143,9 +142,9 @@ async def disk_extend(self, disk, context):
             if await self.middleware.call('truenas.is_ix_hardware'):
                 disk['supports_smart'] = True
             else:
-                disk['supports_smart'] = disk['name'].startswith('nvme') or bool(RE_SMART_AVAILABLE.search(
-                    await self.middleware.call('disk.smartctl', disk['name'], ['-a'], {'silent': True}) or ''
-                ))
+                disk_query = await self.middleware.call('disk.smartctl', disk['name'], ['-a', '--json=c'], {'silent': True})
```
themylogin, line 145: This is performed even if `disk['name'].startswith('nvme')` is `True`, which is unnecessary.

src/middlewared/middlewared/pytest/unit/plugins/test_smart.py
```diff
@@ -1,163 +1,215 @@
...
-        {
-            "num": 2,
-            "description": "Short offline",
-            "status": "SUCCESS",
-            "status_verbose": "Completed without error",
-            "remaining": 0.0,
-            "lifetime": 16589,
-            "lba_of_first_error": None,
+    data = {
+    "ata_smart_self_test_log": {
```
themylogin, line 14: Bad indentation

src/middlewared/middlewared/api/v25_04_0/smartctl.py
```diff
@@ -0,0 +1,35 @@
...
+    num: int
+    description: str
+    status: str
+    status_verbose: str
+    remaining: float
+    lifetime: int
+    lba_of_first_error: int | None = None
```
themylogin, line 14: WHy default value, where is it used?

src/middlewared/middlewared/api/v25_04_0/smartctl.py
```diff
@@ -0,0 +1,35 @@
...
+class NvmeSelfTest(BaseModel):
+    num: int
+    description: str
+    status: str
+    status_verbose: str
+    power_on_hours: int
+    failing_lba: int | None = None
+    nsid: int | None = None
+    seg: int | None = None
+    sct: int | None = 0x0
```
themylogin, line 25: Why is it nullable?

themylogin (changes requested):
Good, only minor issues are left

src/middlewared/middlewared/api/v25_04_0/smartctl.py
aiden3c: For these instances where these fields in the schema are nullable, it's because those keys only show up in specific circumstances in the JSON output. Usually related to if a test is running, and more usually if a test has failed.

themylogin (approved):

yocalebo (approved):

PR #14480, opened 2024-09-10
NAS-130709 / 25.04 / Fix LDAP status on failover

anodos325:
This fix addresses a few minor issues with directory services state recovery.

1. Rather than simply failing directoryservices.setup if unhealthy, we should take recovery steps

2. If for some reason SSSD fails to start triggering a health check failure we should go through etc.generate steps to make sure we have all required files written before trying to start it again.

3. Always generate the nsswitch.conf after setting up directory services. This avoids possibilty of having SSSD / winbind running without them being present in nsswitch.conf on failover.

### DIFF
--- a/src/middlewared/middlewared/plugins/smb.py
+++ b/src/middlewared/middlewared/plugins/smb.py
@@ -382,21 +382,12 @@ class SMBService(ConfigService):
         job.set_progress(30, 'Setting up server SID.')
         await self.middleware.call('smb.set_system_sid')
 
-        """
-        If the ldap passdb backend is being used, then the remote LDAP server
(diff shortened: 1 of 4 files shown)


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -389,14 +389,12 @@ async def configure(self, job, create_paths=True):
         LDAP server.
```
mgrimesix, line 389: Comment needs update:  The comment talks about skipping these steps if ldap passdb backend is used.

mgrimesix (approved):
I see only a comment update.  Approving anyway.

PR #14483, opened 2024-09-10
NAS-130700 / 25.04 / Add support for remote controller audit query and download

mgrimesix:
### Problem
On HA systems each controller maintains an independent audit database.  Actions on a controller are reported only in the audit DB associated with that controller.

### Solution
Provide the capability to query and download the audit DB from the 'other' controller.
If no controller is specified the audit action is performed on the 'current' controller.
If the controller is specified the audit action will be performed on the requested controller.
The controller can be specified by any of the following names:  'BACKUP', 'Standby', 'MASTER', 'Active'.

Passing CI tests: http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/631/

### DIFF
--- a/src/middlewared/middlewared/plugins/audit/audit.py
+++ b/src/middlewared/middlewared/plugins/audit/audit.py
@@ -12,7 +12,6 @@ import yaml
 from truenas_api_client import json as ejson
 
 from .utils import (
-    AUDIT_CONTROLLER_SELECTIONS,
     AUDIT_DATASET_PATH,
     AUDIT_LIFETIME,
     AUDIT_DEFAULT_RESERVATION,
@@ -32,7 +31,7 @@ from middlewared.schema import (
     accepts, Bool, Datetime, Dict, Int, List, Patch, Ref, returns, Str, UUID
 )
 from middlewared.service import filterable, filterable_returns, job, private, ConfigService
-from middlewared.service_exception import CallError, ValidationErrors
+from middlewared.service_exception import CallError, ValidationErrors, ValidationError
 from middlewared.utils import filter_list
 from middlewared.utils.mount import getmntinfo
 from middlewared.utils.functools_ import cache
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/audit/utils.py
```diff
@@ -24,6 +24,8 @@
     AuditEventParam.EVENT.value,
     AuditEventParam.SUCCESS.value,
 )
+AUDIT_CONTROLLER_SELECTIONS = ['MASTER', 'Active', 'BACKUP', 'Standby']
```
anodos325, line 27: Do you need all these options?

mgrimesix: Yes and no. 

The middleware routine `failover.status` reports the controllers as `MASTER` and `BACKUP`
_However..._
The standalone python script `hactl` translates those and reports the controller names as: `Active` and `Standby`.
The extra names are included to support both.

src/middlewared/middlewared/plugins/audit/audit.py
```diff
@@ -193,9 +199,27 @@ async def query(self, data):
...
+            'MASTER' if data['controller'] in ['MASTER', 'Active'] else 'BACKUP'
+        )
+        if ctrlr_state != 'SINGLE':
+            if ctrlr_state not in ['MASTER', 'BACKUP']:
+                verrors.add('audit.query.controller', f'controller status: {ctrlr_state}')
+                verrors.check()
+            else:
+                if want_ctrlr != ctrlr_state:
+                    # Get the data from the other controller
+                    return await self.middleware.call('failover.call_remote', 'audit.query', [data])
```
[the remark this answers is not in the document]
mgrimesix: Excellent!

src/middlewared/middlewared/plugins/audit/audit.py
```diff
@@ -193,9 +197,27 @@ async def query(self, data):
         `success` - boolean value indicating whether the action generating the
         event message succeeded.
         """
-        sql_filters = data['query-options']['force_sql_filters']
 
         verrors = ValidationErrors()
+
+        # If HA, handle the possibility of remote controller requests
+        ctrlr_state = await self.middleware.call('failover.status')
```
yocalebo, line 204: There's no reason to query `failover.status`. You can call `if failover.licensed and data['remote_controller']` and then just try to download the remote logs since you're covering the failure conditions down below.

yocalebo (approved):

PR #14488, opened 2024-09-11
Fix regression in getting sid_info for local users

anodos325:
While fixing an unrelated bug that affected LDAP failover, a restart of the winbind service was removed from the directory services setup method. Changes to directory services health checks in electric eel improved error recovery and eliminated the need for this step.

Unfortunately, this removal exposed that user.get_user_obj and group.get_group_obj were still relying on winbind to perform SID resolution (which failed because winbind had not started yet on the server).

As of electric eel, the SID values for local accounts are derived from the `id` value of the user / group entry in their respective database table. This commit robustizes SID lookups for local accounts through get_obj methods by retrieving via user.query and group.query.


src/middlewared/middlewared/plugins/smb.py
```diff
@@ -415,6 +415,10 @@ async def configure(self, job, create_paths=True):
         )
         await ds_job.wait()
 
+        # Ensure that winbind is running once we configure SMB service
+        if not await self.middleware.call('service.started', 'idmap'):
+            await self.middleware.call('service.start', 'idmap')
```
yocalebo, line 420: Should this be propagated to other controller on HA systems? Same for the `service.restart cifs` line a few lines below this?

src/middlewared/middlewared/plugins/account.py
```diff
@@ -1022,9 +1020,23 @@ def get_user_obj(self, data):
...
+                    idmap_ctx = None
+                    db_entry = self.middleware.call_sync('user.query', [[
+                        'username', '=', user_obj['pw_name']
+                    ]], {'select': ['sid']})
+                    if not db_entry:
+                        self.logger.error(
+                            '%s: local user exists on server but does not exist in the '
+                            'the user account table.', user_obj['pw_name']
+                        )
+                        sid = None
```
bmeagherix, line 1035: `sid` is now already set to `None` above, so this line is unnecessary.  (But meh.)

src/middlewared/middlewared/plugins/account.py
```diff
@@ -1933,9 +1934,24 @@ def get_group_obj(self, data):
...
+                    idmap_ctx = None
+                    db_entry = self.middleware.call_sync('group.query', [[
+                        'group', '=', grp_obj['gr_name']
+                    ]], {'select': ['sid']})
+                    if not db_entry:
+                        self.logger.error(
+                            '%s: local group exists on server but does not exist in the '
+                            'the group account table.', grp_obj['gr_name']
+                        )
+                        sid = None
```
bmeagherix, line 1950: Same comment: `sid` already `None`.

tests/api2/test_011_user.py
```diff
@@ -212,7 +212,7 @@ def test_002_verify_user_exists_in_pwd(request):
     assert pw['pw_dir'] == VAR_EMPTY
 
     # At this point, we're not an SMB user
```
bmeagherix, line 214: We expect `sid` to be `None` because we created the user with `smb` = `False`, right?

Perhaps update `test_005_convert_non_smbuser_to_smbuser` to rerun this `user.get_user_obj` and validate the `sid` is **not** `None`?

bmeagherix (approved):

PR #14510, opened 2024-09-13
NAS-131168 / 25.04 / harden ix-apps metadata handling

anodos325:
This commit makes two minor changes to harden our ix-apps metadata to avoid crashes due to yaml_loads returning None instead of the expected dict. One reason this can happen is that the file in question is empty when read.

1. Switch to using the middleware io util `write_if_changed` This makes writes to the metadata yaml files de-facto atomic as we write to temporary file, flush it, then rename over existing file. Since the files are only used by middlewared I am proactively setting permissions on them to 0o600.

2. Add a wrapper around yaml.safe_load to return an empty dict instead of None type if file is empty. Consumers of these methods expect to receive a dict and so we must handle this edge case.

### DIFF
--- a/src/middlewared/middlewared/plugins/apps/ix_apps/lifecycle.py
+++ b/src/middlewared/middlewared/plugins/apps/ix_apps/lifecycle.py
@@ -5,6 +5,7 @@ import typing
 import yaml
 
 from middlewared.service_exception import CallError
+from middlewared.utils.io import write_if_changed
 
 from .path import (
     get_installed_app_config_path, get_installed_app_rendered_dir_path, get_installed_app_version_path,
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/apps/ix_apps/lifecycle.py
```diff
@@ -18,7 +19,8 @@ def get_rendered_template_config_of_app(app_name: str, version: str) -> dict:
     for rendered_file in get_rendered_templates_of_app(app_name, version):
         with contextlib.suppress(FileNotFoundError, yaml.YAMLError):
             with open(rendered_file, 'r') as f:
-                rendered_config.update(yaml.safe_load(f.read()))
+                if (data := yaml.safe_load(f.read())) is not None:
+                    rendered_config.update(data)
```
bmeagherix, line 23: Both the original and updated code here uses `yaml.safe_load(f.read())`.  OTOH, there's a bunch of code that does in this PR that does `yaml.safe_load(f)`.  Any reason to not be consistent?

yocalebo: Agree. My preference is to pass the file object instead of doing `f.read()`

Qubad786 (approved):

yocalebo (approved):

bmeagherix (approved):

PR #14515, opened 2024-09-13
NAS-131161 / 25.04 / Properly filter for local admins emails

anodos325:
Originally the mail.local_administrators only returned users who were members of the default builtin local administrators privilege by virtue of auxiliary group membership. This commit makes the logic more generic. Since the method was originally written we expanded user.query output to include information about the user's roles and so we can select all users with FULL_ADMIN privileges (which may be granted by non-default privilege sets).


bmeagherix (approved):
I'll approve, but it might also be an idea to also add a **non**-admin user, and ensure their email is **not** in the returned list.

PR #14526, opened 2024-09-16
NAS-131224 / 25.04 / Fix long-standing alert issue on HA

yocalebo:
Back in 2021 ([vide](https://github.com/truenas/middleware/pull/8015)), the PM team decided that `SCALE_ENTERPRISE` no longer meant HA capable _ONLY_ hardware. This product type was extended to all iX branded appliances. This was a subtle, but profoundly impacting change. Specifically with how we handle things like alerts, for example. A recent alert was added (at the bequest of PM) to create proactive alerts when an HA system's other controller doesn't respond after a certain threshold of time ([vide](https://github.com/truenas/middleware/pull/14356)). Unfortunately, the incorrect (but non-intuitive) assumption was made that `failover_related = True` would run on _HA ONLY_ systems. This _was_ true back in 2021, however, it hasn't been true since. Further investigation showed that every single _other_ failover related alert had an explicit check for `failover.licensed`. This is the only reason why we haven't seen this issue sooner.

To remedy the above situation, I have restored the 2021 behavior so that alerts marked with `failover_related = True` will run on HA systems _only_.

While I'm here, I fixed quite a few issues with `__run_alerts`:
1. remove a total of 4 inner for-loops
2. break up the method into smaller methods so that it's easier to maintain, review, optimize, etc
3. update the other failover related alerts to remove an unnecessary call to `failover.licensed`

Full API CI run [here](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/764)

### DIFF
--- a/src/middlewared/middlewared/plugins/alert.py
+++ b/src/middlewared/middlewared/plugins/alert.py
@@ -2,6 +2,7 @@ from collections import defaultdict, namedtuple
 import copy
 from datetime import datetime, timezone
 import errno
+from itertools import zip_longest
 import os
 import textwrap
 import time
@@ -38,7 +39,7 @@ from middlewared.utils import bisect
 from middlewared.utils.plugins import load_modules, load_classes
 from middlewared.utils.python import get_middlewared_dir
 from middlewared.utils.time_utils import utc_now
-
+from middlewared.plugins.failover_.remote import NETWORK_ERRORS
 
 POLICIES = ["IMMEDIATELY", "HOURLY", "DAILY", "NEVER"]
 DEFAULT_POLICY = "IMMEDIATELY"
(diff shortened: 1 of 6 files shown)


src/middlewared/middlewared/plugins/alert.py
```diff
@@ -655,31 +656,75 @@ async def __should_run_or_send_alerts(self):
...
-                    if remote_version == await self.middleware.call("system.version"):
-                        if remote_system_state == "READY" and remote_failover_status == "BACKUP":
-                            run_on_backup_node = True
+                    run_on_backup_node = all((
+                        await self.middleware.call("system.version") == rem_ver,
+                        rem_state == "READY",
+                        rem_fstat == "BACKUP",
+                    ))
+
+        return (this_node, other_node, run_on_backup_node, run_failover_related)
```
themylogin, line 683: Too much variables, let's return a namedtuple/dataclass here

src/middlewared/middlewared/plugins/alert.py
```diff
@@ -696,69 +741,29 @@ async def __run_alerts(self):
...
-                    ]
 
-            for alert in alerts_b:
-                alert.node = backup_node
+                if run_on_backup_node and alert_source.run_on_backup_node:
+                    other_node_alerts = await self.__run_other_node_alert_source(alert_source.name)
 
-            for alert in alerts_a + alerts_b:
-                self.__handle_alert(alert)
+            for talert, oalert in zip_longest(this_node_alerts, other_node_alerts, fillvalue=None):
```
themylogin, line 757: Why do we need to zip them?

yocalebo: 2 reasons:
1. i'm lazy
2. ergonomics (1 for loop to assign the `alert.node` attribute, and call `__handle_alert(`)

themylogin: Sorry to unresolve this, but I really had a difficulty understanding this code. Seeing `zip` makes me think we do something with pairs of items from `this_node_alerts` and `other_node_alerts`. And it is very confusing, because these alerts are not related.

If you want to do everything in one loop, I would suggest:
```
for alert, node in (
        [(this_node, alert) for alert in this_node_alerts] +
        [(other_node, alert) for alert in other_node_alerts]
):
    alert.node = node
    self.__handle_alert(alert)
```

yocalebo: I appreciate the review but I'm not doing as you suggested. I'm trying to remove unnecessary list comprehensions since this is done inside a for loop already and you're adding 2 more. The zip_longest function is perfectly fine here. It's aggregating 2 different iterables (of potentially uneven length).

themylogin (approved):

PR #14530, opened 2024-09-17
NAS-128203 / 25.04 / `reboot` plugin

themylogin:


william-gr:
FWIW, in the original ticket we thought we need a reason for generic shutdown as well, not only reboot.

themylogin:
@william-gr this PR is for displaying "this system must be rebooted, here is why", not for requiring a reason to reboot/shutdown, or do I misunderstand something?

william-gr:
We had a ticket for generic shutdown/reboot reason, if someone goes to the UI and do it manually. I wasn't sure if you were going to address everything in one go, if not, apologies.

themylogin:
@william-gr that's no problem, I was going to address shutdown/reboot audit in a different PR

src/middlewared/middlewared/plugins/failover_/reboot.py
```diff
@@ -18,72 +28,74 @@ class Config:
...
-    @returns()
+        changed = False
+
+        try:
+            other_node = await self.middleware.call('failover.call_remote', 'system.reboot.info', [], {
+                'raise_connect_error': False,
+                'timeout': 2,
+                'connect_timeout': 2,
+            })
+        except Exception:
```
yocalebo, line 66: You'll need to handle ENOMETHOD a bit more gracefully here since this will always fail on HA upgrade.

src/middlewared/middlewared/plugins/failover_/reboot.py
```diff
@@ -18,72 +28,74 @@ class Config:
...
+            self.logger.warning('Unexpected error querying remote reboot info', exc_info=True)
+            other_node = None
+
+        if other_node is not None:
+            for remote_reboot_reason_code, remote_reboot_reason in list(self.remote_reboot_reasons.items()):
+                if remote_reboot_reason.boot_id is None:
+                    # This reboot reason was added while the remote node was not functional.
```
yocalebo, line 73: The reason why I used the keyvalue plugin was to avoid double-reboot scenarios like this so I would interpret this new behavior as a regression. Was there a reason why you stopped using keyvalue?

src/middlewared/middlewared/plugins/security/update.py
```diff
@@ -1,10 +1,12 @@
 import middlewared.sqlalchemy as sa
 
 from middlewared.plugins.failover_.disabled_reasons import DisabledReasonsEnum
-from middlewared.plugins.failover_.reboot import FIPS_KEY
 from middlewared.schema import accepts, Bool, Dict, Int, Patch
 from middlewared.service import ConfigService, ValidationError, job, private
 
+FIPS_REBOOT_CODE = 'FIPS'
```
yocalebo, line 7: I'd rather create an enum class in the `system/reboot_reasons.py` file (or similar) and use that instead of having random global variables in different plugins. This will become important for tools like `hactl`.

src/middlewared/middlewared/plugins/failover_/reboot.py
themylogin: You are correct. I added the persistence for `self.remote_reboot_reasons` via `keyvalue` plugin.

src/middlewared/middlewared/plugins/failover_/reboot.py
```diff
@@ -132,3 +152,43 @@ async def other_node(self, job):
...
+            info = await self.info()
+
+        self.middleware.send_event('failover.reboot.info', 'CHANGED', id=None, fields=info)
+
+    @private
+    async def load_remote_reboot_reasons(self):
+        self.remote_reboot_reasons_key = f'remote_reboot_reasons_{await self.middleware.call("failover.node")}'
+        self.remote_reboot_reasons = {
+            k: RemoteRebootReason(**v)
+            for k, v in (await self.middleware.call('keyvalue.get',self.remote_reboot_reasons_key, {})).items()
```
yocalebo, line 168:
```suggestion
            for k, v in (await self.middleware.call('keyvalue.get',self.remote_reboot_reasons_key, {})).items()
```
```suggestion
            for k, v in (await self.middleware.call('keyvalue.get', self.remote_reboot_reasons_key, {})).items()
```

yocalebo (approved):
Be sure and open UI ticket with the new behavior since HA upgrades will be broken in nightlies with these changes.

themylogin:
https://ixsystems.atlassian.net/browse/NAS-131325

themylogin:
time 20:00

PR #14533, opened 2024-09-18
Fix typo in cache refresh calls

anodos325:
This commit fixes some errors introduced while renaming private directory services cache methods.


TheJulianJES:
Had already created PR https://github.com/truenas/middleware/pull/14532 with the exact same change, but I guess this works too 😄. Thanks for picking it up so fast.

yocalebo (approved):

PR #14542, opened 2024-09-19
NAS-131143 / 25.04 / Always use python serializer and preserve types

themylogin:
…htly different results

### DIFF
--- a/src/middlewared/middlewared/api/base/handler/accept.py
+++ b/src/middlewared/middlewared/api/base/handler/accept.py
@@ -72,9 +72,4 @@ def validate_model(model: type[BaseModel], data: dict, *, exclude_unset=False, e
 
         raise verrors from None
 
-    if expose_secrets:
-        mode = "python"
-    else:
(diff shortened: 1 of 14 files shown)


anodos325:
Won't this change the type for `roles` in user.query and `group.query` response?

src/middlewared/middlewared/plugins/account.py
```diff
@@ -1612,11 +1612,13 @@ async def group_extend(self, group, ctx):
         group['name'] = group['group']
         group['users'] = ctx['memberships'].get(group['id'], [])
 
-        privilege_mappings = privileges_group_mapping(ctx['privileges'], [group['gid']], 'local_groups')
+        privilege_mappings = await self.middleware.call(
```
anodos325, line 1615: I'd rather not add an extra middleware call per-group here.

anodos325 (approved):
Looks good.

themylogin:
time 4:00

PR #14544, opened 2024-09-19
NAS-130966 / 25.04 / Make sure we delete any app volume mount datasets created on app install failure

sonicaj:
This PR adds changes to remove any app volume mount datasets created during installation in the event installation process failed for whatever reason. We are picky on this end and make sure we only remove them if the app volume dataset didn't exist before as when a user deletes an app, it's optional for him/her to remove the app volume mounts datasets and we don't want them nuked if a user installs an app with the same name and app install fails for whatever reason.

### DIFF
--- a/src/middlewared/middlewared/plugins/apps/crud.py
+++ b/src/middlewared/middlewared/plugins/apps/crud.py
@@ -248,23 +248,28 @@ class AppService(CRUDService):
                 compose_action(app_name, version, 'up', force_recreate=True, remove_orphans=True)
         except Exception as e:
             job.set_progress(80, f'Failure occurred while installing {app_name!r}, cleaning up')
-            apps_volume_ds = self.get_app_volume_ds(app_name)
-            for method, args, kwargs in (
-                (compose_action, (app_name, version, 'down'), {'remove_orphans': True}),
-                (shutil.rmtree, (get_installed_app_path(app_name),), {}),
-            ) + (
-                self.middleware.call_sync, ('zfs.dataset.delete', apps_volume_ds, {'recursive': True})
-            ) if apps_volume_ds else ():
-                with contextlib.suppress(Exception):
-                    method(*args, **kwargs)
-
-            self.middleware.call_sync('app.metadata.generate').wait_sync(raise_error=True)
+            self.remove_failed_resources(app_name, version)
             self.middleware.send_event('app.query', 'REMOVED', id=app_name)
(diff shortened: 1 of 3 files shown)


Dremor:
I have a question : Will we be able to remove ti ix-volumes if we remove an app without choosing to remove the ix-volume at that time ? I'm currently stuck with multiple ix-volumes I cannot remove, and would like to be able to remove them.

src/middlewared/middlewared/plugins/apps/crud.py
```diff
@@ -248,21 +249,31 @@ def create_internal(
...
+            self.remove_failed_resources(app_name, version, app_volume_ds_exists is False)
             self.middleware.send_event('app.query', 'REMOVED', id=app_name)
             raise e from None
         else:
             if dry_run is False:
                 job.set_progress(100, f'{app_name!r} installed successfully')
                 return self.get_instance__sync(app_name)
 
+    @private
+    def remove_failed_resources(self, app_name, version, remove_ds=False):
```
yocalebo, line 263: This for loop is wicked complicated for no real reason. Please simplify this so it's easier for future developers to read/maintain.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/apps/crud.py
```diff
@@ -248,21 +249,32 @@ def create_internal(
...
+        apps_volume_ds = self.get_app_volume_ds(app_name) if remove_ds else None
+
+        with contextlib.suppress(Exception):
+            compose_action(app_name, version, 'down', remove_orphans=True)
+
+        shutil.rmtree(get_installed_app_path(app_name), ignore_errors=True)
+
+        if apps_volume_ds and remove_ds:
+            with contextlib.suppress(Exception):
```
Qubad786, line 272: Maybe at least log this so we know why it failed if it did as the dataset should exist based on above func call

Qubad786 (approved):

yocalebo (approved):

sonicaj:
> I have a question : Will we be able to remove ti ix-volumes if we remove an app without choosing to remove the ix-volume at that time ? I'm currently stuck with multiple ix-volumes I cannot remove, and would like to be able to remove them.

@Dremor we would be adding some changes for FangTooth on this end so users can remove any ix-volumes which are not associated with any installed app.

PR #14545, opened 2024-09-19
NAS-129059 / 25.04 / hide dimm2 and dimm3 temps on x10

yocalebo:
This platform has 4x dimm slots but only 2 are populated on X10. The dimms that are not populated report a status of "OK" but the temperature is reported as `-1C`. This has been happening since CORE days. Fix it by not reporting those dimm sensors.

### DIFF
--- a/src/middlewared/middlewared/plugins/enclosure_/enclosure_class.py
+++ b/src/middlewared/middlewared/plugins/enclosure_/enclosure_class.py
@@ -21,7 +21,7 @@ from .constants import (
     DRIVE_BAY_LIGHT_STATUS,
 )
 from .element_types import ELEMENT_TYPES, ELEMENT_DESC
-from .enums import ControllerModels, ElementDescriptorsToIgnore, ElementStatusesToIgnore, JbodModels
+from .enums import ControllerModels, ElementDescriptorsToIgnore, ElementStatusesToIgnore, ElementType, JbodModels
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/enclosure_/enclosure_class.py
```diff
@@ -205,23 +205,32 @@ def _get_model_and_controller(self):
...
-            )),
-        ))
+        if parsed_element_status.lower() == ElementStatusesToIgnore.UNSUPPORTED.value:
+            return True
+        elif self.is_xseries:
+            if element_type == ElementType.TEMPERATURE_SENSORS.value:
+                if self.model == ControllerModels.X10.value and desc.startswith(('dimm2', 'dimm3')):
+                    # x10 platform only has 2 DIMM slots (dimm0/1) populated so the others report
+                    # -1C temperature but the descriptor status reports as "OK".
+                    return True
```
mgrimesix, line 222: The x10 can _never_ and will _never_ support the dimm2 and dimm3 slots?
It would be nice if there was a better way to handle these platform specific things.

yocalebo: That's what I alluded to in my comment. The dimm module isn't populated but the descriptor still reports as "OK". So there isn't a concrete way of knowing that the module isn't populated. We basically have to make some form of assumption here.

PR #14546, opened 2024-09-19
NAS-131141 / 25.04 / Shutdown/reboot audit

themylogin:


src/middlewared/middlewared/plugins/config.py
```diff
@@ -105,7 +106,10 @@ def upload(self, job):
             is_tar = tarfile.is_tarfile(stf.name)
             self.upload_impl(stf.name, is_tar_file=is_tar)
 
-        self.middleware.run_coroutine(self.middleware.call('system.reboot', {'delay': 10}), wait=False)
+        self.middleware.run_coroutine(
+            self.middleware.call('system.reboot', 'Configuration upload', {'delay': 10}, app=app),
```
yocalebo, line 110: The `Configuration upload` string is typed many times and is prone to typos. This is why I asked for a `RebootReasons` enum in a previous PR. Let's put all these statically typed strings in that class and re-use them everywhere.

themylogin: @yocalebo it's a different class of strings. The `RebootReasons` class we added is why the system must be rebooted. This is the string that will be logged to the audit log when the system is rebooted.

Do you want another centralized storage for all the reasons we reboot the system ourselves, or will a per-file string constant(s) suffice? I see no reason to centralize this, the reboot/shutdown reason supplied by user can be arbitrary.

yocalebo: Per-file string constants are fine for this. We just need to focus on the details here since this will show up in audit logs and we need to be consistent.

src/middlewared/middlewared/plugins/failover.py
```diff
@@ -388,7 +388,7 @@ def sync_to_peer(self, options):
         )
 
         if options['reboot']:
-            self.middleware.call_sync('failover.call_remote', 'system.reboot', [{'delay': 2}])
+            self.middleware.call_sync('failover.call_remote', 'Failover sync to peer', 'system.reboot', [{'delay': 2}])
```
william-gr, line 391: Typo? Should come after system.boot, no?

william-gr:
Just to make sure its not forgotten, it will need midcli changes too, right?

src/middlewared/middlewared/plugins/config.py
```diff
@@ -105,7 +108,10 @@ def upload(self, job):
...
```
yocalebo, line 112:
```suggestion
            self.middleware.call('system.reboot', CONFIGURATION_UPLOAD_REBOOT_REASON, {'delay': 10}, app=app),
```

src/middlewared/middlewared/plugins/failover.py
```diff
@@ -388,7 +389,7 @@ def sync_to_peer(self, options):
         )
 
         if options['reboot']:
-            self.middleware.call_sync('failover.call_remote', 'system.reboot', [{'delay': 2}])
+            self.middleware.call_sync('failover.call_remote', 'system.reboot', 'Failover sync to peer', [{'delay': 2}])
```
yocalebo, line 392:
```suggestion
            self.middleware.call_sync('failover.call_remote', 'system.reboot', SYSTEM_UPGRADE_REBOOT_REASON, [{'delay': 2}])
```

themylogin:
> Just to make sure its not forgotten, it will need midcli changes too, right?

https://github.com/truenas/midcli/pull/110

src/middlewared/middlewared/plugins/failover.py
themylogin: Please confirm that these two should not differ

tests/api2/test_system_lifetime.py
```diff
@@ -0,0 +1,30 @@
...
+
+    call("system.reboot", "Integration test")
+
+    for i in range(180):
+        try:
+            new_boot_id = call("system.boot_id")
+        except Exception:
+            pass
+        else:
+            if new_boot_id == boot_id:
```
yocalebo, line 17: How does this work? `new_boot_id` will never match the `boot_id` since the system will have been rebooted?

themylogin: Yes, this was a mistake. The test was not testing the fact that the system was indeed rebooted. Updated.

themylogin: http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/842/

src/middlewared/middlewared/plugins/update.py
```diff
@@ -288,7 +291,7 @@ async def update(self, job, attrs):
         await self.middleware.call_hook('update.post_update')
 
         if attrs['reboot']:
-            await self.middleware.call('system.reboot', {'delay': 10})
+            await self.middleware.call('system.reboot', 'System upgrade', {'delay': 10}, app=app)
```
yocalebo, line 294:
```suggestion
            await self.middleware.call('system.reboot', SYSTEM_UPGRADE_REBOOT_REASON, {'delay': 10}, app=app)
```

tests/api2/test_system_lifetime.py
```diff
@@ -0,0 +1,35 @@
...
+
+from middlewared.test.integration.utils import call
+
+from auto_config import ha
+
+
+if not ha:
+    # This cannot be tested on a HA system since rebooting this node will just fail over to another node
+
+    def test_system_reboot():
```
yocalebo, line 11: Please use `@pytest.mark.skipif(not ha, reason='HA only test')`

william-gr (commented):
lgtm, thanks!

yocalebo (approved):
Thanks!

themylogin:
time 30:00

PR #14557, opened 2024-09-21
NAS-131350 / 25.04 / Fix failing build

sonicaj:
This commit adds changes to move IANA numbers to trash as outlined by https://github.com/truenas/middleware/pull/13868/files because they are now coming from ipmitool we have in debian, however they have a decent diff from the URI we are pulling from and the comment there outlines it needs to be a version after 1.18.9-4 but we are still at the same version. So with this change, we remove the enterprise-numbers.txt file coming from ipmitool and keep on pulling our own instead.


sonicaj:
Closing this in favor of https://github.com/truenas/middleware/pull/14558

PR #14562, opened 2024-09-23
NAS-131318 / 25.04 / Get rid of `unable to open cache: unable to locate cache directory: neither $XDG_CACHE_HOME nor $HOME are defined`

themylogin:
This cache was neither used previously nor necessary (we don't retrieve same files from the backup constantly)


themylogin:
time 0:20

yocalebo (approved):

PR #14564, opened 2024-09-23
NAS-131365 / 25.04 / Better error message for bad storj bucket name

creatorcary:
Currently the returned error message is "An error occurred (InvalidBucketName) when calling the CreateBucket operation: The specified bucket is not valid." To save the user some research, we can clarify this message to "Bucket name can only contain lowercase letters, numbers, and hyphens" which is verbatim in the [storj documentation](https://storj.dev/support/object-browser#:~:text=The%20bucket%20name%20can%20only%20contain%20lowercase%20letters%2C%20numbers%2C%20and%20hyphens.).

bad_bucket_name_alert

### DIFF
--- a/src/middlewared/middlewared/rclone/remote/storjix.py
+++ b/src/middlewared/middlewared/rclone/remote/storjix.py
@@ -45,6 +45,8 @@ class StorjIxRcloneRemote(BaseRcloneRemote):
                 s3_client.create_bucket(Bucket=name)
             except s3_client.exceptions.BucketAlreadyExists as e:
                 raise CallError(str(e), errno=errno.EEXIST)
+            except Exception as e:
+                raise CallError("Bucket name can only contain lowercase letters, numbers, and hyphens", errno.EINVAL, str(e))
 
         return await self.middleware.run_in_thread(create_bucket_sync)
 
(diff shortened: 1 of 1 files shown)


william-gr:
Should we improve validation error too? IIRC it accepts uppercase there now
I dont know if Storj issue is starting with uppercase or any uppercase

creatorcary:
> IIRC it accepts uppercase there now I dont know if Storj issue is starting with uppercase or any uppercase
I tested on a vm and it doesn't accept any caps. Here's the exact [restrictions on bucket names](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html#:~:text=Bucket%20names%20can%20consist%20only%20of%20lowercase%20letters%2C%20numbers%2C%20dots%20(.)%2C%20and%20hyphens%20(%2D).) that I found.

creatorcary:
IMHO I don't know about hard-coding the business logic of a third party like this. It might be better to leave the current error message and only change the webui validation.

william-gr:
Well, if there are clear restriction on bucket names, we dont we validate it just like we do anything else, like dataset/pool names?

Right now all we have is: Str("bucket", required=True, empty=False) ?

themylogin (approved):
I agree with the current approach (only rewording the message).

However, what if the bucket already exists? Did you test that it would not be an InvalidName error?

yocalebo:
I spoke directly with Logan and we can actually make this a bit more proper, going to close this and he'll push another PR.

PR #14567, opened 2024-09-24
NAS-131370 / 25.04 / Fix post mount actions execution on system dataset setup

Qubad786:
## Problem

When system dataset is being migrated, if we have a pool from where it i being migrated - we are temporarily mounting it to another location before it is mounted at it's rightful place so we can copy over data.
However during this transition, we have post mount actions being executed which is not alright and can trigger different failures which in the case i ended up finding this was an issue related to netdata setup not being properly executed.

## Solution

Ensure all post mount actions only when the system dataset is mounted at it's rightful place i.e `/var/db/system`. Additionally, permissions for ix-state directory in netdata are corrected if any exception occurs on the first time system tries to setup netdata.

### DIFF
--- a/src/middlewared/middlewared/plugins/reporting/netdata_configure.py
+++ b/src/middlewared/middlewared/plugins/reporting/netdata_configure.py
@@ -2,6 +2,7 @@ import os
 import subprocess
 
 from middlewared.service import private, Service
+from middlewared.utils.filesystem.copy import copytree, CopyTreeConfig
 
 from .utils import get_netdata_state_path
 
@@ -24,17 +25,20 @@ class ReportingService(Service):
 
     @private
     def post_dataset_mount_action(self):
-        if os.path.exists(get_netdata_state_path()):
+        netdata_state_path = get_netdata_state_path()
+        # We want to make sure this path exists always regardless of an error so that
+        # at least netdata can start itself gracefully
+        try:
+            os.makedirs(netdata_state_path, exist_ok=False)
+        except FileExistsError:
             return
 
-        cp = subprocess.run(
-            ['cp', '-a', '/var/lib/netdata', get_netdata_state_path()], check=False, capture_output=True,
-        )
-        if cp.returncode != 0:
-            self.logger.error('Failed to copy netdata state over from /var/lib/netdata: %r', cp.stderr.decode())
-            # We want to make sure this path exists always regardless of an error so that
-            # at least netdata can start itself gracefully
-            os.makedirs(get_netdata_state_path(), exist_ok=True)
+        try:
+            copytree('/var/lib/netdata', netdata_state_path, config=CopyTreeConfig())
+        except Exception:
+            self.logger.error('Failed to copy netdata state over from /var/lib/netdata', exc_info=True)
+            os.chown(netdata_state_path, uid=999, gid=997)
+            os.chmod(netdata_state_path, mode=0o755)
 
     @private
     async def start_service(self):
--- a/tests/api2/test_system_dataset.py
+++ b/tests/api2/test_system_dataset.py
@@ -77,3 +77,11 @@ def test_system_dataset_mountpoints():
         assert ds_stats["uid"] == system_dataset_spec["chown_config"]["uid"]
         assert ds_stats["gid"] == system_dataset_spec["chown_config"]["gid"]
         assert ds_stats["mode"] & 0o777 == system_dataset_spec["chown_config"]["mode"]
+
+
+def test_netdata_post_mount_action():
+    # We rely on this to make sure system dataset post mount actions are working as intended
+    ds_stats = call("filesystem.stat", "/var/db/system/netdata/ix_state")
+    assert ds_stats["uid"] == 999, ds_stats
+    assert ds_stats["gid"] == 997, ds_stats
+    assert ds_stats["mode"] & 0o777 == 0o755, ds_stats
--- a/src/middlewared/middlewared/plugins/sysdataset.py
+++ b/src/middlewared/middlewared/plugins/sysdataset.py
@@ -579,7 +579,8 @@ class SystemDatasetService(ConfigService):
                 os.chmod(mountpoint, mode_perms)
 
             mounted = True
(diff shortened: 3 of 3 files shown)


anodos325:
Are there any concrete steps we can take to improve our CI for these behaviors?

yocalebo:
Changes like this make me squirm. This can break failover on our HA. Have you run an entire CI run against an HA system?

src/middlewared/middlewared/plugins/reporting/netdata_configure.py
```diff
@@ -24,17 +24,20 @@ def netdata_state_location(self):
...
     @private
     def post_dataset_mount_action(self):
-        if os.path.exists(get_netdata_state_path()):
+        netdata_state_path = get_netdata_state_path()
+        if os.path.exists(netdata_state_path):
             return
 
         cp = subprocess.run(
-            ['cp', '-a', '/var/lib/netdata', get_netdata_state_path()], check=False, capture_output=True,
+            ['cp', '-a', '/var/lib/netdata', netdata_state_path], check=False, capture_output=True,
```
yocalebo, line 32: Why are we doing fork+exec'ing here? Why not use `shutil.rmtree`?

yocalebo: Woops, not `shutil.rmtree`. I meant to say `shutil.copy2` :smile:

anodos325: We can use `middleware.utils.filesystem.copy.copytree` as well.

Qubad786:
> Are there any concrete steps we can take to improve our CI for these behaviors?

@anodos325 i'll push a new integration test for this ✅

src/middlewared/middlewared/plugins/reporting/netdata_configure.py
```diff
@@ -24,17 +25,21 @@ def netdata_state_location(self):
 
     @private
     def post_dataset_mount_action(self):
-        if os.path.exists(get_netdata_state_path()):
+        netdata_state_path = get_netdata_state_path()
+        if os.path.exists(netdata_state_path):
```
yocalebo, line 29: This is prone to races and is a bit confusing. Why do we check for the existence of this path and return but down below we do `os.makedirs(..exist_ok=True)`? I think we should just remove the existence check, right? @anodos325 thoughts?

src/middlewared/middlewared/plugins/reporting/netdata_configure.py
```diff
@@ -24,17 +25,21 @@ def netdata_state_location(self):
...
-            os.makedirs(get_netdata_state_path(), exist_ok=True)
+        # We want to make sure this path exists always regardless of an error so that
+        # at least netdata can start itself gracefully
+        # Also copytree will not apply correct perms on parent dir, so we will apply them as well
+        os.makedirs(netdata_state_path, exist_ok=True)
+        os.chown(netdata_state_path, uid=999, gid=997)
+        os.chmod(netdata_state_path, mode=0o755)
+
+        try:
+            copytree('/var/lib/netdata', netdata_state_path, config=CopyTreeConfig(traverse=True))
```
anodos325, line 40: Do we need traverse here? Do we expect to have other filesystems mounted in /var/lib/netdata?

Qubad786: I think we should be able to get along without it, testing that bit

Qubad786: Yes works as expected without traverse

src/middlewared/middlewared/plugins/reporting/netdata_configure.py
yocalebo: Okay then change os.makedirs logic to have exist_ok=False and catch the DirectoryExists exception

anodos325: Sounds like a bug. I will work on a fix shortly.

src/middlewared/middlewared/plugins/reporting/netdata_configure.py
```diff
@@ -24,17 +25,21 @@ def netdata_state_location(self):
...
-            ['cp', '-a', '/var/lib/netdata', get_netdata_state_path()], check=False, capture_output=True,
-        )
-        if cp.returncode != 0:
-            self.logger.error('Failed to copy netdata state over from /var/lib/netdata: %r', cp.stderr.decode())
-            # We want to make sure this path exists always regardless of an error so that
-            # at least netdata can start itself gracefully
-            os.makedirs(get_netdata_state_path(), exist_ok=True)
+        # We want to make sure this path exists always regardless of an error so that
+        # at least netdata can start itself gracefully
+        # Also copytree will not apply correct perms on parent dir, so we will apply them as well
```
anodos325, line 34: C.f. https://github.com/truenas/middleware/pull/14580

anodos325: Once that's merged we can remove the chown / chmod.

tests/api2/test_system_dataset.py
```diff
@@ -77,3 +77,11 @@ def test_system_dataset_mountpoints():
         assert ds_stats["uid"] == system_dataset_spec["chown_config"]["uid"]
         assert ds_stats["gid"] == system_dataset_spec["chown_config"]["gid"]
         assert ds_stats["mode"] & 0o777 == system_dataset_spec["chown_config"]["mode"]
+
+
+def test_netdata_post_mount_action():
+    # We rely on this to make sure system dataset post mount actions are working as intended
```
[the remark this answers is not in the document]
Qubad786: It should not make a difference, because each time after it runs we expect this to be in place so doing this here will have the same effect.
Should i make the changes?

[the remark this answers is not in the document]
anodos325: Nope. Just clarifying

anodos325 (approved):

PR #14568, opened 2024-09-24
NAS-131373 / 25.04 / Fix serialization error

aiden3c:
On `eng-gpu-dev106.dc1`, when running the recent SMART changes, we get errors related to JSON serialization. Updating the return type to convert our SelfTest classes to dicts was an effective solution. Putting these changes onto `eng-gpu-dev106.dc1` fixes the serialization error and everything shows up as it should.

### DIFF
--- a/src/middlewared/middlewared/plugins/smart.py
+++ b/src/middlewared/middlewared/plugins/smart.py
@@ -113,7 +113,7 @@ def parse_smart_selftest_results(data) -> list[dict[str, Any]] | None:
     if "scsi_self_test_0" in data: # 0 is most recent test
         for index in range(0, 20): # only 20 tests can ever return
             test_key = f"scsi_self_test_{index}"
-            if not test_key in data:
+            if test_key not in data:
                 break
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/api/v25_04_0/smartctl.py
```diff
@@ -1,3 +1,4 @@
+import json
```
themylogin, line 1: Where is it used in this file?

aiden3c:
Actually, getting some CI errors, putting WIP tag on

src/middlewared/middlewared/plugins/smart.py
```diff
@@ -4,6 +4,7 @@
 import re
```
yocalebo, line 4: Something I've noticed in this file. On line 115 you have this
```
            if not test_key in data:
                break
```
This should be written as `if test_key not in data`

themylogin (approved):

PR #14569, opened 2024-09-24
NAS-131330 / 25.04 / Better error handling for bad storj bucket name

creatorcary:
Currently the returned `CallError` is "An error occurred (InvalidBucketName) when calling the CreateBucket operation: The specified bucket is not valid." To save the user some research, we can clarify this message to the `ValidationError` "Bucket name can only contain lowercase letters, numbers, and hyphens" which is verbatim in the [storj documentation](https://storj.dev/support/object-browser#:~:text=The%20bucket%20name%20can%20only%20contain%20lowercase%20letters%2C%20numbers%2C%20and%20hyphens.) or "The bucket name must be between 3-63 characters in length and cannot contain uppercase."

[s3 bucket naming rules](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html) (we do not enforce them all!)

We should ask UI to render these errors as red validation text as per the submitted ticket.

validation_bucke+
validation_BUCKET

### DIFF
--- a/src/middlewared/middlewared/plugins/cloud_sync.py
+++ b/src/middlewared/middlewared/plugins/cloud_sync.py
@@ -10,9 +10,10 @@ from middlewared.plugins.cloud.crud import CloudTaskServiceMixin
 from middlewared.plugins.cloud.model import CloudTaskModelMixin, cloud_task_schema
 from middlewared.plugins.cloud.path import get_remote_path, check_local_path
 from middlewared.plugins.cloud.remotes import REMOTES, remote_classes
+from middlewared.rclone.remote.storjix import StorjIxError
 from middlewared.schema import accepts, Bool, Cron, Dict, Int, Password, Patch, Str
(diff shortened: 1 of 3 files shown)


william-gr:
I still dont quite understand why we dont validate the bucket name prior to actually trying to make the network request?

yocalebo:
> I still dont quite understand why we dont validate the bucket name prior to actually trying to make the network request?

The idea is that we don't maintain a list of requirements on our side that the s3 api already enforces. The moment we enforce these, s3 will change their requirements and then we'll have to update ours to match theirs etc.

creatorcary:
I should note that some of the hyper-specific naming rules are not enforced by the storj api. For example, naming a bucket "amzn-s3-demo-bucket" is allowed.

aiden3c (approved):
LGTM!

themylogin (approved):

PR #14572, opened 2024-09-24
NAS-131115 / 25.04 / Fix null data points in UPS stats

Qubad786:
### Problem
When the UPS service is started after a long period, null values distort the mean calculation because they are included in the average. Furthermore, for datasets larger than 3000 points, Netdata reduces the number of points by creating 3000 groups and calculating the average for each group to get single datapoint. Null values affect the accuracy of these group calculations as well.

### Solution
Skip all zero values when calculating the mean and use the median to represent a single data point in a group if Netdata has more than 3000 data points. This will ensure more accurate aggregation.


yocalebo (approved):

yocalebo:
backport

PR #14578, opened 2024-09-25
NAS-131396 / 25.04 / Add support for user-linked API keys

anodos325:
The primary motivation for converting API keys so that they are explicitly linked to user accounts is to provide some method for TrueNAS Connect to use tie in to local or directory services accounts to a credential that is persistently stored in the TrueNAS Connect keychain.

The legacy API key mechanism was insufficient for this purpose because the mismatch between user accounts and API keys could lead to administrators retaining NAS access after account deletion, expiration, or locking. For more in-depth context and reasoning please refer to the internal design document NEP-053.

For this purpose, this commit makes the following changes:
1. Legacy API keys are migrated to a user_identifier LEGACY_API_KEY which is automatically linked to the root, admin, or truenas_admin account depending on server configuration.
2. If the legacy API key granted less than FULL_CONTROL, then the API key is migrated with a `revoked` state so that the system administrator has an opportunity to review and generate a new key and/or service account that provides the correct level of access.
3. API key authentication now passes through libpam.
4. user.query results now include an `api_keys` key that contains a list of IDs for API keys that exist for the user.
5. Authenticated users with READONLY_ADMIN privilege or greater are able to create and manage their own API keys.

During development of this new feature, it was determined that the original API keys were written with insufficient hashing rounds leading to the following changes:

6. On successful authentication the stored hash of any LEGACY_API_KEY is automatically upgraded to a newer sha512-based standard with significantly increased hashing rounds.
7. Dependency on the passlib library was removed.

Since libpam is now used for all non-token authentication methods, a new auth login endpoint (auth.login_ex) was added to middleware to facilitate future enhancements for challenge-response authentication mechanisms. Authentication via username + password + OTP token has been converted to this new standard. This new endpoint is more closely aligned with standards and requirements in NIST SP 800-63B. In summary the following changes were made:

8. An AuthentationContext dataclass was created to house a middleware session's PAM context and associated information. An instance of this is stored in the session's App object.
9. A new endpoint auth.login_ex was added that is expandable and is aware of the server's configured "authentication assurance level".
10. Existing login endpoints (auth.login, auth.login_with_api_key, auth.login_with_token) were converted to wrappers around auth.login_ex.
11. Initial implementation of session lifetime and inactivity guidelines for different authentication assurance levels was added.
12. Challenge-response workflow for username + password + OTP token implemented with associated tests.


src/middlewared/middlewared/plugins/api_key.py
```diff
@@ -1,17 +1,22 @@
...
 
+from datetime import datetime, UTC
 from middlewared.api import api_method
 from middlewared.api.current import (
     ApiKeyEntry, ApiKeyCreateArgs, ApiKeyCreateResult, ApiKeyUpdateArgs, ApiKeyUpdateResult,
-    ApiKeyDeleteArgs, ApiKeyDeleteResult, HttpVerb,
+    ApiKeyDeleteArgs, ApiKeyDeleteResult, ApiKeyMyKeysArgs, ApiKeyMyKeysResult,
 )
-from middlewared.service import CRUDService, private, ValidationErrors
+from middlewared.service import CRUDService, pass_app, private, ValidationError, ValidationErrors
```
mgrimesix, line 12: ValidationError is not used.

src/middlewared/middlewared/plugins/api_key.py
```diff
@@ -1,17 +1,22 @@
...
 import middlewared.sqlalchemy as sa
-from middlewared.utils.allowlist import Allowlist
+from middlewared.utils import filter_list
+from middlewared.utils.auth import LEGACY_API_KEY_USERNAME
+from middlewared.utils.crypto import generate_pbkdf2_512, generate_string
+from middlewared.utils.privilege import credential_has_full_admin
+from middlewared.utils.sid import sid_is_valid
 from middlewared.utils.time_utils import utc_now
 if TYPE_CHECKING:
     from middlewared.main import Middleware
```
mgrimesix, line 22: Middleware is unused, but this might be unavoidable.  The flake8 error is `F401`. 
Maybe add `# noqa: F401`

tests/api2/test_api_key.py
```diff
@@ -1,106 +1,237 @@
...
-            # verify that token has same access rights
-            c.call("system.info")
-            with pytest.raises(Exception):
-                c.call("system.version")
-            with pytest.raises(Exception):
-                c.call("service.start", "cifs")
-            with pytest.raises(Exception):
-                c.call("service.update", "cifs", {"enable": True})
+            resp = c.call('auth.login_ex', {
+               'mechanism': 'API_KEY_PLAIN',
```
mgrimesix, line 108: flake8: Cut-n-paste issue.  Every usage of these are under indented
There are seven other of these in this file.

mgrimesix (commented):
This is a superficial flake8 issue review.
Second pass will be real review.

tests/api2/test_api_key.py
anodos325: I may not be right, but at least I'm consistently wrong.

src/middlewared/middlewared/alembic/versions/25.04/2024-09-19_18-48_user_api_key.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def upgrade():
+    conn = op.get_bind()
+    to_revoke = []
+    for row in conn.execute("SELECT id, allowlist FROM account_api_key").fetchall():
+        try:
+            if json.loads(row['allowlist']) != DEFAULT_ALLOW_LIST:
+                to_revoke.append(str(row['id']))
+        except Exception:
```
sonicaj, line 30: Should we log it somewhere ?

sonicaj (commented):
Had an initial pass, awesome work 🚀

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -409,103 +402,494 @@ async def login(self, app, username, password, otp_token):
...
+            twofactor_auth = await self.middleware.call('auth.twofactor.config')
+            if twofactor_auth['enabled'] and '2FA' in resp['user_data']['account_attributes']:
+                otp_required = True
+
+        return resp | {'otp_required': otp_required}
 
     @cli_private
     @no_auth_required
-    @accepts(Password('token'))
+    @accepts(Password('api_key'))
```
yocalebo, line 842: While you're here, we should change the old `@accepts` decorator to the new one (also get rid of `@returns`

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -409,103 +402,494 @@ async def login(self, app, username, password, otp_token):
...
 
-        await self.session_manager.login(app, TokenSessionManagerCredentials(self.token_manager, token))
-        token.session_ids.add(app.session_id)
-        return True
+        return resp['response_type'] == AuthResp.SUCCESS
 
-    @private
+    @cli_private
     @no_auth_required
+    @accepts(Password('token'))
```
yocalebo, line 880: Same here, we should use the new `@api_method` decorator

src/middlewared/middlewared/alembic/versions/25.04/2024-09-19_18-48_user_api_key.py
anodos325: We don't do that in other migrations, but I think I should add an alert for revoked API keys.

src/middlewared/middlewared/alert/source/api_key.py
```diff
@@ -0,0 +1,27 @@
...
+        "use the renwed API key."
+    )
+
+    async def create(self, args):
+        return Alert(ApiKeyRevokedAlertClass, args, key=args['key_name'])
+
+    async def delete(self, alerts, key_name_set):
+        remaining = []
+        for alert in alerts:
+            if (alert_name := json.loads(alert.key)) not in key_name_set:
```
mgrimesix, line 22: Is the `alert_name` assignment required?  It's not really used.

anodos325: Come to think of it, no. I was using that assignment for debugging purposes.

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -388,124 +383,511 @@ def get_token_for_shell_application(self, token_id, origin):
...
+                    # being expired.
+                    resp['pam_response']['reason'] = 'Api key is expired.'
+
+                if resp['pam_response']['code'] == pam.PAM_SUCCESS:
+                    if thehash.startswith('$pbkdf2-sha256'):
+                        # Legacy API key with insufficient iterations. Since we
+                        # know that the plain-text we have here is correct, we can
+                        # use it to update the hash in backend.
+                        await self.middleware.call('api_key.update_hash', data['api_key'])
+
```
bmeagherix, line 680: IIRC the plan was to auto-revoke an API_KEY that was used on a non-private transport.  I didn't notice this being done anywhere.  Would here be a good location?

src/middlewared/middlewared/auth.py
```diff
@@ -38,12 +44,54 @@ def dump(self):
...
+        self.inactivity_timeout = None
+        self.last_used_at = now
+
+        if assurance:
+            self.expiry = now + self.assurance.max_session_age
+            self.inactivity_timeout = self.assurance.max_inactivity
+
+    def notify_used(self):
+        if self.inactivity_timeout:
+            self.last_used_at = monotonic()
```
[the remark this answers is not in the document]
anodos325: You're right. It would be safer to skip updating the ts if we're already expired.

src/middlewared/middlewared/etc_files/pam_tdb.py
```diff
@@ -0,0 +1,43 @@
...
+)
+
+
+def convert_keys(username, keys) -> PamTdbEntry:
+    user_api_keys = []
+
+    for key in keys:
+        if key['expires_at'] is None:
+            expiry = 0
+        else:
```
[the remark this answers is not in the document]
anodos325: We actually filter out `revoked` when we `etc.generate`. That said I'll add error handling here in case someone helpfully edits `plugins/etc.py` without understanding context.

src/middlewared/middlewared/plugins/api_key.py
```diff
@@ -22,161 +27,356 @@ class APIKeyModel(sa.Model):
...
+            # prevent keys we can't resolve from being written
+            item['revoked'] = True
+
+        match expiry:
+            case -1:
+                # key has been forcibly revoked
+                item['revoked'] = True
+            case 0:
+                # zero value indicates never expires
+                pass
```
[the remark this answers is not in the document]
anodos325: Sure.

src/middlewared/middlewared/plugins/api_key.py
```diff
@@ -22,161 +27,356 @@ class APIKeyModel(sa.Model):
...
-    async def load_keys(self):
-        self.keys = {
-            key["id"]: key
-            for key in await self.middleware.call("datastore.query", "account.api_key")
-        }
+    def _validate(self, schema_name: str, data: dict, verrors: ValidationErrors, id_: int = None):
+        if self.middleware.call_sync('datastore.query', self._config.datastore, [
+            ['name', '=', data['name']], ['id', '!=', id_]
+        ]):
+            verrors.add(schema_name, "name must be unique")
```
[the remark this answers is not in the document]
anodos325: Have to check for cases where api_key.create. In this case `['id', '!=', None]` Not the most intuitive, but paradigm exists in our code base.

src/middlewared/middlewared/plugins/auth.py
anodos325: Yes it would, but in a separate commit.

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -388,124 +383,511 @@ def get_token_for_shell_application(self, token_id, origin):
...
+            AUTH_TOKEN_PLAIN
+            token: authentication token string
+            login_options: dictionary with additional authentication options
+
+            OTP_TOKEN
+            otp_token: one-time password token. This is only permitted if
+            a previous auth.login_ex call responded with "OTP_REQUIRED".
+
+            login_options
```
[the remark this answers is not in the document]
anodos325: This is additional key in payload `login_options` other keys are mentioned above this.

src/middlewared/middlewared/plugins/api_key.py
```diff
@@ -22,161 +27,356 @@ class APIKeyModel(sa.Model):
...
+        verrors = ValidationErrors()
+        self._validate("api_key_create", data, verrors)
+        user = self.middleware.call_sync('user.query', [
+            ['username', '=', data['username']]
+        ])
+        if not user:
+            verrors.add('api_key_create', 'User does not exist.')
+
+        if user and not user[0]['roles']:
+            verrors.add('api_key_create', 'User lacks privilege role membership.')
```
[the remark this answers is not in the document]
anodos325: I wanted it logically separated from validation errors processing because we're shifting authorization check from being handled before method call to internally in method. This makes it harder for a helpful edit to accidentally bypass authorization checks.

bmeagherix (approved):

src/middlewared/middlewared/auth.py
```diff
@@ -38,12 +44,56 @@ def dump(self):
...
+        See comments in utils/auth.py for more information.
+        """
+        now = monotonic()
         self.user = user
+        self.assurance = assurance
         self.allowlist = Allowlist(user["privilege"]["allowlist"])
         self.is_user_session = True
+        self.login_at = datetime.now(UTC)
+        self.expiry = None
+        self.inactivity_timeout = None
```
bmeagherix, line 70: Is there a reason why these are not being initialized thusly:
```
        self.expiry = now + AA_LEVEL1.max_session_age
        self.inactivity_timeout = AA_LEVEL1.max_inactivity
```

src/middlewared/middlewared/alembic/versions/25.04/2024-10-08_18-48_user_api_key.py
```diff
@@ -0,0 +1,42 @@
+"""Convert to user-linked tokens
+
+Revision ID: 8ae49ac78d14
+Revises: 6dedf12c1035
```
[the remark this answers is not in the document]
anodos325: Indeed it is.  I just fixed.

src/middlewared/middlewared/auth.py
anodos325: Yes. This is also used with connections via Unix Domain Socket (in which case assurance is None).

PR #14587, opened 2024-09-27
add H30 to enclosure plugin

yocalebo:
The H30 is exactly the same as H10/H20 so all of the detection logic is in place. Simply need to add the model to the enum and everything else will just work.


william-gr (approved):

yocalebo:
backport

PR #14598, opened 2024-10-01
NAS-131420 / 25.04 / fix copy/paste error for r20/r20b slot mapping

yocalebo:
The slot mapping for the R20/R20B platforms were very clearly copy and pasted incorrectly because they have the same mapping as the R20A. QE has tested drive identify logic on an R20B and the lights are way off. This should fix it.


themylogin (approved):

yocalebo:
backport

PR #14601, opened 2024-10-01
NAS-131456 / 25.04 / Update user-group memberships using datastore

themylogin:
It seems that the original code for updating user-group relationships (that was never touched since it was written in 2017) contains an error:

https://github.com/truenas/middleware/blob/master/src/middlewared/middlewared/plugins/account.py#L1367

`gms` is an array of `account.bsdgroupmembership` objects, so `gm["id"]` has a Row ID type for `account.bsdgroupmembership` table. `groups` is an array of Row IDs for `account.bsdgroup` table. `gm['id'] not in groups` check has no sense, it compares Row IDs of different tables.

This is probably the root cause of a rare bug where the following code

```
from truenas_api_client import Client


c = Client()

while True:
    user = c.call('user.query', [['username', '=', 'smbuser']], {'get': True})

    pk = c.call('user.update', user['id'], {'ssh_password_enabled': True})
    pk = c.call('user.update', user['id'], {'groups': [43, 40, 90]})
    user = c.call('user.get_instance', pk)
    assert set(user['groups']) == set([43, 40, 90])
```

very rarely fails with assertion failure, `user["groups"]` being

```
    "groups": [
      90,
      43,
      90
    ],
```

Our `datastore` plugin has the ability to manage many-to-many relationships table, let's use it.

### DIFF
--- a/src/middlewared/middlewared/plugins/account.py
+++ b/src/middlewared/middlewared/plugins/account.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 import errno
 import glob
 import json
@@ -9,6 +10,8 @@ import wbclient
 from pathlib import Path
 from collections import defaultdict
 from contextlib import suppress
+
+from sqlalchemy.orm import relationship
 
 from dataclasses import asdict
 from middlewared.api import api_method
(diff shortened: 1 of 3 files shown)


themylogin:
http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/1097/

anodos325 (approved):

PR #14610, opened 2024-10-02
NAS-130009 / 25.04 / Use unique disk identifier for retrieving/maintaining disk stats

Qubad786:
## Problem

Netdata uses disk names to store disk statistics in it's database. We can configure it as such that it uses disk identifier to retrieve but that does not help because in the database they are stored with disk names which means after reboots or hardware changes, the data is not going to remain consistent if disk names changed.

## Solution

After much investigation we decided to write our own disk stats plugin which uses procfs similar to netdata to retreive disk statistics but in this case we are using disk identifiers to store/retrieve them in netdata database. The identifier being used to store the stats is similar to the logic we have in place to uniquely identify our own disks.

CI Run: http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/1119/


src/middlewared/middlewared/plugins/enclosure.py
```diff
@@ -10,12 +10,12 @@
 from middlewared.schema import Dict, Int, Str, accepts
 from middlewared.service import CallError, CRUDService, filterable, private
 from middlewared.service_exception import MatchNotFound
-from middlewared.plugins.disk_.enums import DISKS_TO_IGNORE
 import middlewared.sqlalchemy as sa
 from middlewared.utils import filter_list
 from middlewared.plugins.enclosure_.r30_drive_identify import set_slot_status as r30_set_slot_status
 from middlewared.plugins.enclosure_.fseries_drive_identify import set_slot_status as fseries_set_slot_status
 from middlewared.plugins.enclosure_.sysfs_disks import toggle_enclosure_slot_identifier
+from middlewared.utils.disks import DISKS_TO_IGNORE
```
mgrimesix, line 18: Not used.

src/middlewared/middlewared/utils/disks.py
```diff
@@ -1,44 +1,137 @@
...
+RE_IS_PART = re.compile(r'p\d{1,3}$')
 
-DISKS_TO_IGNORE: tuple = ('sr', 'md', 'dm-', 'loop', 'zd')
-NVME_TYPE: str = 'nvme'
-HDD_TYPE: str = 'hdd'
 
+def safe_retrieval(prop, key, default, as_int=False) -> typing.Any:
+    value = prop.get(key)
+    if value is not None:
+        if type(value) == bytes:
```
mgrimesix, line 15: flake8:  use 'is' not '==' for type check

src/middlewared/middlewared/utils/disk_stats.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def get_disk_stats(disk_identifier_mapping: dict | None = None) -> dict[str, dict]:
+    disk_identifier_mapping = disk_identifier_mapping or get_disks_with_identifiers()
+    available_disks = get_disk_names()
+    stats = {}
+    disk_stats_content = ''
+    with contextlib.suppress(IOError):
+        with open('/proc/diskstats', 'r') as f:
+            disk_stats_content = f.read()
```
yocalebo, line 18: Let's not read the entire contents of this file into memory. Instead we should iterate over it. (i.e.
```
with open('/proc/diskstats', 'r') as f:
    for line in f:
        parts = line.strip().split()
        if len(parts) < 14 or parts[2] not in available_disks:
            continue

yocalebo (changes requested):

mgrimesix (approved):
Approving, but might want to wait for Caleb's review.

yocalebo (approved):

PR #14614, opened 2024-10-03
NAS-131772 / 25.04 / Fix entire `core.get_jobs` crashing if one job fails to dump its result

themylogin:
This is especially painful in debug: if a job returns an invalid result, we lose the results of all other jobs.

### DIFF
--- a/src/middlewared/middlewared/job.py
+++ b/src/middlewared/middlewared/job.py
@@ -609,11 +609,16 @@ class Job:
                 'extra': extra,
             }
 
+        result_encoding_error = None
         if self.state == State.SUCCESS:
             if raw_result:
                 result = self.result
             else:
-                result = self.middleware.dump_result(self.method, self.result, False)
+                try:
+                    result = self.middleware.dump_result(self.method, self.result, False)
+                except Exception as e:
+                    result = None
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/job.py
```diff
@@ -628,6 +633,7 @@ def __encode__(self, raw_result=True):
             'logs_excerpt': self.logs_excerpt,
             'progress': self.progress,
             'result': result,
+            **({'result_encoding_error': result_encoding_error} if result_encoding_error is not None else {}),
```
yocalebo, line 636: This line is strange and I'm having problems groking it. Does this mean the `result_encoding_error` key only gets inserted into the object if `result_encoding_error` object is not None? Does it make sense to just do something like
```
'result': result,
'result_encoding_error': result_encoding_error,
...
```

yocalebo: Also curious why we're crashing?

themylogin: If the job result schema does not match the actual result, we'll fail to remove secrets and crash

anodos325 (approved):

themylogin:
time 2:00

PR #14616, opened 2024-10-03
NAS-131485 / 24.10.0 / Update default docker address pool

sonicaj:
This PR adds changes to update default docker address pool to use what docker uses but instead of having a size of `16` for each docker network, we use `24` so user can still get to consume at least 4096 docker networks this way.

Problem with earlier default was that users who were using apps in BETA and had 15 number of apps installed, each docker network was huge with a size of `16` which meant users could only create `15` apps with the default docker addr pool before it started consuming `192.168.0.0/16`. Then we made the change to use a smaller size per docker network and using base as `172.30/31.0.0/16` which was fine but the problem was users having 15 apps were already consuming this addr pool which resulted in docker exhausting all addr pools and refusing to start.

The current change defaults to using the default with what docker had but instead of having a very large network per docker network, we reduce the size to `24` so we can have more docker networks as it is unrealistic that we will have 65534 containers using a single docker network.

### DIFF
--- a/src/middlewared/middlewared/alembic/versions/24.10/2024-09-03_20-33_docker_addr_pool.py
+++ b/src/middlewared/middlewared/alembic/versions/24.10/2024-09-03_20-33_docker_addr_pool.py
@@ -22,7 +22,7 @@ def upgrade():
                 'address_pools',
                 sa.TEXT(),
                 nullable=False,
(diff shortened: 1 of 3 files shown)


sonicaj:
backport

yocalebo (approved):

PR #14620, opened 2024-10-04
NAS-131582 / 25.04 / Replace usage of hard-coded strings for the product type

mgrimesix:
### Problem
--------------------
The product types, 'SCALE' and 'SCALE_ENTERPRISE' are not defined anywhere.  We test for the product type in many places, each time using a hard-coded string. 
This presents two problems:  Increased possibility of typos and missing a single source of truth for the product type identifier.

### Solution
-------------------
Add a product type class, 'ProductType' to `system/product.py`.
Replace the hard-coded strings with the class variables.
Note: The enum class was not used solely because it required appending `.value` to the name.

Also fixed various minor flake8 issues.

### DIFF
--- a/src/middlewared/middlewared/plugins/system/product.py
+++ b/src/middlewared/middlewared/plugins/system/product.py
@@ -20,7 +20,7 @@ LICENSE_FILE_MODE = 0o600
 PRODUCT_NAME = 'TrueNAS'
 
 
-class ProductType():
+class ProductType:
     SCALE = 'SCALE'
     SCALE_ENTERPRISE = 'SCALE_ENTERPRISE'
 
--- a/src/middlewared/middlewared/plugins/system/product.py
+++ b/src/middlewared/middlewared/plugins/system/product.py
@@ -11,18 +11,13 @@ from licenselib.license import ContractType, Features, License
 from middlewared.plugins.truenas import EULA_PENDING_PATH
 from middlewared.schema import accepts, Bool, returns, Str
 from middlewared.service import no_authz_required, private, Service, ValidationError
-from middlewared.utils import sw_info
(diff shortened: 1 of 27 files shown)


src/middlewared/middlewared/plugins/system/product.py
```diff
@@ -20,6 +20,11 @@
 PRODUCT_NAME = 'TrueNAS'
 
 
+class ProductType():
```
yocalebo, line 23: This is a global mutable type. We can also reduce the size of this object considerably.

We can do 1 of 2 things:
Option 1:
```
from dataclasses import dataclass

@dataclass(slots=True, frozen=True)
class ProductType:
    SCALE: str = 'SCALE'
    SCALE_ENTERPRISE: str = 'SCALE_ENTERPRISE'
```

Option 2:
```
from enum import Enum

class ProductType(str, Enum):
    SCALE = 'SCALE'
    SCALE_ENTERPRISE = 'SCALE_ENTERPRISE'
```

For option number 1, we reduce memory size considerably and we make it immutable.

For option number 2, we reduce memory size to 0 however you need to change all the places we reference the enum to do something like this `ProductType.SCALE.value`. If you don't add the `.value` at the end and only use `ProductType.SCALE` it creates a _new_ instance of the enum class which is an expensive operation.

yocalebo (changes requested):

src/middlewared/middlewared/plugins/system/product.py
mgrimesix: I've got no particular preference.  Which option do _you_ prefer?

yocalebo (approved):
Very nice! Conditional approval once you've done your testing

PR #14634, opened 2024-10-09
Several OpenAPI schema fixes

rkojedzinszky:
https://github.com/oapi-codegen/oapi-codegen can generate client code with these patches.


yocalebo:
Always appreciate the PR but we're not going to include these changes. We're actively removing REST in favor of our websocket API.

PR #14635, opened 2024-10-09
NAS-131605 / 25.04 / Make sure NVRAM file of VM is preserved when vm name is changed

Qubad786:
## Problem
When a VM's name is changed, its NVRAM (non-volatile memory) is not being preserved, which causes some VMs to fail to boot after the name change.

## Solution
Ensure the NVRAM state is persisted even after the VM name is updated, allowing VMs to boot correctly post name change.

### DIFF
--- a/src/middlewared/middlewared/plugins/vm/vms.py
+++ b/src/middlewared/middlewared/plugins/vm/vms.py
@@ -15,7 +15,7 @@ from middlewared.service import CallError, CRUDService, item_method, job, privat
 from middlewared.validators import Range, UUID
 from middlewared.plugins.vm.numeric_set import parse_numeric_set, NumericSet
 
-from .utils import ACTIVE_STATES, get_default_status
+from .utils import ACTIVE_STATES, get_default_status, get_vm_nvram_file_name, SYSTEM_NVRAM_FOLDER_PATH
 from .vm_supervisor import VMSupervisorMixin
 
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/vm/utils.py
```diff
@@ -3,6 +3,8 @@
 
 ACTIVE_STATES = ['RUNNING', 'SUSPENDED']
 SYSTEM_NVRAM_FOLDER_PATH = '/data/subsystems/vm/nvram'
+LIBVIRT_QEMU_UID = 64055
+LIBVIRT_QEMU_GID = 64055
```
anodos325, line 7: We shouldn't be using IDs this high for service accounts. Please make ticket to shift below ID 1000.

src/middlewared/middlewared/plugins/vm/vms.py
```diff
@@ -377,6 +380,22 @@ async def do_update(self, id_, data):
         vm_data = await self.get_instance(id_)
         if new['name'] != old['name']:
             await self.middleware.run_in_thread(self._rename_domain, old, vm_data)
+            try:
+                new_path = os.path.join(SYSTEM_NVRAM_FOLDER_PATH, get_vm_nvram_file_name(new))
+                shutil.copy2(
+                    os.path.join(SYSTEM_NVRAM_FOLDER_PATH, get_vm_nvram_file_name(old)), new_path
+                )
+                os.chown(new_path, LIBVIRT_QEMU_UID, LIBVIRT_QEMU_GID)
```
anodos325, line 388: We should be doing blocking file IO in a thread.

src/middlewared/middlewared/plugins/vm/utils.py
Qubad786: Created one https://ixsystems.atlassian.net/browse/NAS-131695 ✅

src/middlewared/middlewared/plugins/vm/vms.py
```diff
@@ -377,6 +380,24 @@ async def do_update(self, id_, data):
         vm_data = await self.get_instance(id_)
         if new['name'] != old['name']:
             await self.middleware.run_in_thread(self._rename_domain, old, vm_data)
+            try:
```
yocalebo, line 383: I don't like creating 2 separate threads for small operations. Either convert the entirety of this method to being synchronous or add another method that does all blocking IO so that we can do it in 1 thread total

yocalebo (changes requested):

src/middlewared/middlewared/plugins/vm/vms.py
```diff
@@ -377,6 +380,24 @@ async def do_update(self, id_, data):
...
+                new_path = os.path.join(SYSTEM_NVRAM_FOLDER_PATH, get_vm_nvram_file_name(new))
+                await self.middleware.run_in_thread(
+                    shutil.copy2, os.path.join(SYSTEM_NVRAM_FOLDER_PATH, get_vm_nvram_file_name(old)), new_path
+                )
+                await self.middleware.run_in_thread(
+                    os.chown, new_path, LIBVIRT_QEMU_UID, LIBVIRT_QEMU_GID
```
anodos325, line 389: Perhaps we should rename instead?

yocalebo (approved):

anodos325 (approved):

PR #14639, opened 2024-10-09
NAS-131547 / 25.04 / Appropriately handle resilvering case with zpool status report

Qubad786:
## Problem

The current `zpool status` is incorrectly handling the "replacing" vdev type, which appears only during an ongoing resilvering process. This is causing a "path not found" key error, as it treats the replacing vdev as a regular vdev.

## Solution

Updated `zpool status` to properly handle and manage the "replacing" vdev type during resilvering, preventing key errors.


creatorcary (approved):
LGTM!

PR #14648, opened 2024-10-10
NAS-131715 / 25.04 / Add retry loop for kerberos start post domain join

anodos325:
This is continuation of fix in NAS-131697, which was determined to be insufficient for all edge cases with multi-dc setups.

During the domain join process temporarily hard-code the KDC used by MIT kerberos to the DC with which we're communicating. This is to prevent races on sysvol replication. Once we successfully join AD we need to remove this hard-coded configuration to allow AD failover. This commit adds a retry loop for trying to start keberos after the join attempt because there's risk that the DC we discover via DNS hasn't received the updated computer account information.


src/middlewared/middlewared/plugins/directoryservices_/activedirectory_join_mixin.py
```diff
@@ -43,6 +50,32 @@ def _ad_wait_wbclient(self) -> None:
...
+        After initial AD join we reconfigure kerberos to find KDC via DNS.
+        Unfortunately, depending on the AD environment it may take a significant
+        amount of time to replicate the new machine account to other domain
+        controllers. This means we have a retry loop on starting the kerberos
+        service.
+        """
+        tries = 0
+        while tries < MAX_KERBEROS_START_TRIES:
+            try:
+                self.middleware.call_sync('kerberos.start')
```
bmeagherix, line 64: missing return?

anodos325: Absolutely! Good eye.

bmeagherix (approved):

PR #14650, opened 2024-10-10
NAS-131317 / 25.04 / Include "Power On Hours Ago" in smartctl results

aiden3c:
I've added "Power On Hours Ago" as a key to be returned when getting smartctl test information. Refer to [this PR](https://github.com/truenas/webui/pull/10845) for further explanation as to what POH are and why we went with it for this ticket.

All tests are CI/unit tests.


src/middlewared/middlewared/api/v25_04_0/smartctl.py
```diff
@@ -10,6 +10,7 @@ class AtaSelfTest(BaseModel):
     status_verbose: str
     remaining: float
     lifetime: int
+    poh_ago: int # current lifetime - lifetime at test start, "power on hours ago"
```
themylogin, line 13: Can we use a complete name? `powered_on_hours_ago` is not a super long identifier (just 6 chars longer than `status_verbose`) and is self-explanatory. `poh` is a term that we'll need to look up in the documentation every time we see it.

src/middlewared/middlewared/api/v25_04_0/smartctl.py
```diff
@@ -10,6 +10,7 @@ class AtaSelfTest(BaseModel):
     status_verbose: str
     remaining: float
     lifetime: int
+    power_on_hours_ago: int # current lifetime - lifetime at test start, "power on hours ago"
```
themylogin, line 13: Now this comment isn't necessary :)
Also PEP8 should fail here, here must be two spaces between the code and the `#`

themylogin (approved):
Perfect, thank you!

PR #14664, opened 2024-10-11
NAS-131738 / 25.04 / add "DISABLED" option to pool.dataset.create snapdir

creatorcary:
Following the upstream ZFS change that adds the option to disable access to the snapshot directory altogether https://github.com/openzfs/zfs/pull/16587 and the corresponding change https://github.com/truenas/py-libzfs/pull/288 of the `snapdir` property's type in py-libzfs, it appears that the only change needed from middleware is a schema change for pool.dataset.create (and effectively for pool.dataset.update).

Note that the `snapdev` option is not affected.


anodos325 (approved):

anodos325:
Once you merge, please open corresponding UI ticket.

PR #14669, opened 2024-10-13
NAS-130280 / 25.04 / Make sure issuer subject is preserved when CSR is signed

sonicaj:
This PR adds changes to fix the case when an external CA is imported in TN and that CA is used to sign a CSR, the generated cert's issuer DN's componets are shuffled/out of order and do not match the order in the CA. With these changes, the original order of Relative Distinguished Names (RDNs) in the Subject field is preserved when generating a Certificate using an imported CA. By maintaining the sequence of RDNs, the CSRs signed by the Certificate Authority (CA) will reflect the same order of Issuer DN as in the CA's Subject DN ([RFC 3280](https://www.rfc-editor.org/rfc/rfc3280#section-4.1.2.4) on page 91).

### DIFF
--- a/src/middlewared/middlewared/pytest/unit/plugins/crypto/certs.py
+++ b/src/middlewared/middlewared/pytest/unit/plugins/crypto/certs.py
@@ -390,7 +390,8 @@ def test__generating_cert(generate_params, key_type, key_size, cert_info):
             }
         }
     }
-])
+], ids=['Test ca_sign_csr']
+)
 def test_ca_sign_csr(data):
     cert = sign_csr_with_ca(data)
     cert_data = x509.load_pem_x509_certificate(cert.encode('utf-8'), default_backend())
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/pytest/unit/plugins/crypto/certs.py
```diff
@@ -207,3 +211,192 @@ def test__generating_cert(generate_params, key_type, key_size, cert_info):
             cert_details[k].pop('SubjectKeyIdentifier')
 
         assert cert_info[k] == cert_details[k], cert_details
+
+
+@pytest.mark.parametrize('data', [
```
yocalebo, line 216: Please use `id=` kwarg to this parametrize decorator. Otherwise, the console from pytest will be polluted with the key contents.

yocalebo (changes requested):

yocalebo (approved):
Thanks!

PR #14682, opened 2024-10-14
NAS-131690 / 25.04 / Allow remote audit call to return exception if unresponsive.

mgrimesix:
### Problem
----
An `audit.query` to the remote controller returns an empty result if the remote controller is unable to respond.  This is incorrect behavior and results in a false interpretation of the query.

### Fix
----
Raise an exception when the remote controller does not respond to an `audit.query`.

The test involves generating an audit event on the remote node.  The event is an attempt to delete a built-in user.
Updated passing test can be found [here](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/1353/).

### DIFF
--- a/tests/api2/test_audit_basic.py
+++ b/tests/api2/test_audit_basic.py
@@ -123,12 +123,11 @@ def standby_audit_event():
     event = "user.delete"
     username = "backup"
     user = call('user.query', [["username", "=", username]], {"select": ["id"], "get": True})
-    try:
-        # This _should_ fail.  Assert if not!
+    # Generate an audit entry on the remote node
+    with pytest.raises(CallError):
         call('failover.call_remote', event, [user['id']])
-        assert True, "!!! Should not be allowed to delete a built-in user !!!"
-    except CallError:
-        yield {"event": event, "username": username}
+
+    yield {"event": event, "username": username}
 
 
 # =====================================================================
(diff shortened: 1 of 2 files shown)


tests/api2/test_audit_basic.py
```diff
@@ -115,29 +116,19 @@ def init_audit():
...
-            }],
-            {'raise_connect_error': False, 'timeout': 2, 'connect_timeout': 2}
-        )
-        yield name
-    finally:
-        if user_id is not None:
-            call('failover.call_remote', 'user.delete', [user_id])
+        # This _should_ fail.  Assert if not!
+        call('failover.call_remote', event, [user['id']])
+        assert True, "!!! Should not be allowed to delete a built-in user !!!"
```
themylogin, line 129: `assert True` will do nothing, what is this supposed to do?

tests/api2/test_audit_basic.py
```diff
@@ -324,47 +315,58 @@ def test_audit_timestamps(self, svc):
...
-
-        assert tries > 0, "Failed to get expected audit entry"
-        assert remote_audit_entry != []
-        params = remote_audit_entry[0]['event_data']['params'][0]
-        assert params['username'] == name
-
-    def test_audit_ha_export(self, standby_user):
+    @pytest.mark.parametrize('remote_available', [True, False])
+    def test_audit_ha_query(self, standby_audit_event, remote_available):
+        assert standby_audit_event, "Missing event data"
```
themylogin, line 320: A fixture that fails to set up will fail all tests that depend on it

tests/api2/test_audit_basic.py
mgrimesix: The goal is to generate an audit entry on the remote node.  To generate the entry I'm attempting to delete a built-in user. 
OK, let me refactor this.

tests/api2/test_audit_basic.py
mgrimesix: Refactoring.

themylogin (approved):
Excellent, thank you!

PR #14683, opened 2024-10-14
NAS-131090

bmeagherix:
#### Overview
SCST includes support for Fibre Channel targets using QLA2xxx HBAs.  This PR includes the middleware changes necessary to utilize the mechanism.

- ALUA is supported, using the same HA internal iSCSI targets as are used for iSCSI ALUA support,
- NPIV will be supported (not yet ready, hence WIP)

#### API summary
In CORE an `fcport` CRUD interface was available to perform the target to FC port mapping.  A similar CRUD API is presented here.

However, in CORE we could _reliably_ pair the ports on each node in a HA system simply by using the same moniker (e.g. `isp0`).  Since this is not quite true in SCALE we instead will have a DB table / API to wire the ports together based upon the slot information obtained from `dmidecode`.  The auto-wiring will use a default moniker of `fc0`, etc ... so that any migration from SCALE will be _forced_ to manually check that the A/B controllers port pairing are as expected.  The upshot of all of this is that another `fc.fc_host` API is implemented.

The `fc.fc_host` API will also include the NPIV setting for the FC Host port, which in CORE was a `System Tunable`.

### DIFF
--- a/src/middlewared/middlewared/plugins/fc/fc.py
+++ b/src/middlewared/middlewared/plugins/fc/fc.py
@@ -1,0 +1,153 @@
+import os.path
+import re
+import subprocess
+from functools import cache
+from pathlib import Path
+
+from middlewared.service import Service
+from middlewared.service_exception import CallError
+from .utils import dmi_pci_slot_info, wwn_as_colon_hex
+
+FIBRE_CHANNEL_SEARCH_STRING = 'Fibre Channel'
+QLA2XXX_KERNEL_MODULE = 'qla2xxx_scst'
+FIBRE_CHANNEL_VENDOR = 'QLogic'
+FC_HOST_PAT = re.compile('/sys/devices/(.*)/(?P<host>host.*)/fc_host/(?P=host)')
+
+
+class FCService(Service):
+
+    class Config:
+        private = True
+
+    async def capable(self):
+        """
+        Returns True if the system is licensed for FIBRECHANNEL and contains
+        one or more Fibre Channel cards.  False otherwise.
+        """
+        if await self.middleware.call('system.is_enterprise'):
+            if await self.middleware.call('system.feature_enabled', 'FIBRECHANNEL'):
+                if await self.middleware.call('fc.hba_present'):
+                    return True
+        return False
+
+    @cache
+    def hba_present(self):
+        """
+        Return True if a Qlogic Fibre Channel card is present.  False otherwise.
+        """
+        lspci_cmd = ['lspci']
+        ret = subprocess.run(lspci_cmd, capture_output=True)
+        if ret.returncode:
+            error = ret.stderr.decode() if ret.stderr else ret.stdout.decode()
+            if not error:
+                error = 'No error message reported'
+            self.logger.debug('Failed to execute command: lspci with error: %r', error)
+            raise CallError(f'Failed to determine serial number/product: {error}')
+        for line in ret.stdout.decode().split('\n'):
+            if FIBRE_CHANNEL_SEARCH_STRING not in line:
+                continue
+            if FIBRE_CHANNEL_VENDOR not in line:
+                continue
+            # We matched!
+            return True
+        return False
+
+    @cache
+    def fc_hosts(self):
+        result = []
+        if self.middleware.call_sync('fc.capable'):
+            self.__load_kernel_module()
+            slots = dmi_pci_slot_info()
+            with os.scandir('/sys/class/fc_host') as scan:
+                for i in filter(lambda x: x.is_symlink() and x.name.startswith('host'), scan):
+                    # Ensure realpath looks something like
+                    # '/sys/devices/pci0000:b2/0000:b2:00.0/0000:b3:00.0/host14/fc_host/host14'
+                    if srp := FC_HOST_PAT.search(os.path.realpath(i.path)):
+                        addr = srp[1]
+                        path = Path(i.path)
+                        # node_name and port_name must be transported as string to avoid JSON breakage
+                        entry = {
+                            'name': i.name,
+                            'path': i.path,
+                            'node_name': (path / 'node_name').read_text().strip(),
+                            'port_name': (path / 'port_name').read_text().strip(),
+                            'addr': addr,
+                        }
+                        # Maybe also get slot information from DMI (dmidecode)
+                        # First pull out the piece of PCI address that is also included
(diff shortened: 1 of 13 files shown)


src/middlewared/middlewared/plugins/fc/fc_host.py
```diff
@@ -0,0 +1,347 @@
...
+                    vpfilter = [['port', '~', f'^{data["alias"]}/[1-9][0-9]*$']]
+                    vports = [p['port'].split('/')[-1] for p in await self.middleware.call('fcport.query', vpfilter, {'select': ['port']})]
+                    vports = [int(p['port'].split('/')[-1]) for p in await self.middleware.call('fcport.query', vpfilter, {'select': ['port']})]
+                    for chan in vports:
+                        if chan > npiv:
+                            verrors.add(
+                                f'{schema_name}.npiv',
+                                f'Invalid npiv ({npiv}) supplied, {data["alias"]}/{chan} is currently mapped to a target'
+                            )
+                        break
```
anodos325, line 213: Does this need to be indented more?

bmeagherix: Yes!  Good catch.

src/middlewared/middlewared/plugins/fc/utils.py
```diff
@@ -0,0 +1,93 @@
...
+    elif wwpn_naa:
+        return [['port_name', '=', f'0x{wwpn_naa[4:]}']]
+    elif wwpn_b_naa:
+        return [['port_name', '=', f'0x{wwpn_b_naa[4:]}']]
+    else:
+        return []
+
+
+def is_fc_addr(addr):
+    return bool(HEX_COLON.match(addr) or NAA_PATTERN.match(addr))
```
anodos325, line 93: Do you think it would be helpful to create some basic tests for these parsers in `tests/unit`?

src/middlewared/middlewared/plugins/fc/fc.py
```diff
@@ -0,0 +1,142 @@
...
+            if FIBRE_CHANNEL_SEARCH_STRING not in line:
+                continue
+            if FIBRE_CHANNEL_VENDOR not in line:
+                continue
+            # We matched!
+            return True
+        return False
+
+    @filterable
+    def fc_hosts(self, filters, options):
```
anodos325, line 57: Should we have `api_method` with `QueryArgs`?

src/middlewared/middlewared/plugins/fc/utils.py
bmeagherix: Yes.

src/middlewared/middlewared/plugins/fc/fc.py
bmeagherix: Yes.

src/middlewared/middlewared/plugins/fc/fc_host.py
```diff
@@ -0,0 +1,378 @@
...
+            match node:
+                case 'A':
+                    node_a_fc_hosts = await self.middleware.call('fc.fc_hosts', physical_port_filter)
+                    node_b_fc_hosts = await self._get_remote_fc_hosts()
+                case 'B':
+                    node_a_fc_hosts = await self._get_remote_fc_hosts()
+                    node_b_fc_hosts = await self.middleware.call('fc.fc_hosts', physical_port_filter)
+                case _:
+                    raise CallError('Cannot configure FC until HA is configured')
+            # Match pairs by slow (includes PCI function)
```
anodos325, line 274: slot?

src/middlewared/middlewared/plugins/fc/fc_host.py
```diff
@@ -0,0 +1,378 @@
...
+            for fc_host in node_a_fc_hosts:
+                slot_2_fc_host_wwpn[fc_host['slot']]['A'] = str_to_naa(fc_host.get('port_name'))
+            for fc_host in node_b_fc_hosts:
+                slot_2_fc_host_wwpn[fc_host['slot']]['B'] = str_to_naa(fc_host.get('port_name'))
+            # Iterate over each slot and make sure the corresponding fc_host entry is preent
```
anodos325, line 279: present?

src/middlewared/middlewared/plugins/fc/fc_host.py
```diff
@@ -0,0 +1,378 @@
...
+            sorted_slots = sorted(list(slot_2_fc_host_wwpn.keys()))
+            result = True
+            for slot in sorted_slots:
+                wwpn = slot_2_fc_host_wwpn[slot].get('A')
+                wwpn_b = slot_2_fc_host_wwpn[slot].get('B')
+                if wwpn and wwpn_b:
+                    existing = await self.middleware.call('fc.fc_host.query', [
+                        ['OR', [['wwpn', '=', wwpn], ['wwpn_b', '=', wwpn_b]]]
```
anodos325, line 287: Rather than `OR` are we able to do `in` ?

anodos325:
```
[['wwpn', 'in', [wwpn, wwpn_b]]]
```

bmeagherix: I don't think so.  The keys are also different, not just the values.

src/middlewared/middlewared/plugins/fcport.py
```diff
@@ -1,9 +1,382 @@
...
+        data['wwpn_b'] = None
+        # We want to retrieve the WWPN associated with the port name
+        if VPORT_SEP_CHAR in data['port']:
+            # NPIV - Get the base FC Host
+            fc_host_alias, chan = data['port'].rsplit(VPORT_SEP_CHAR, 1)
+        else:
+            fc_host_alias = data['port']
+            chan = None
+        try:
+            fc_host_pair = await self.middleware.call('fc.fc_host.query', [['alias', '=', fc_host_alias]], {'get': True})
```
anodos325, line 324: Would it make sense to have a datastore `extend_context` calling `fc.fc_host.query`, then do filter_list here?

bmeagherix:
FWIW, currently writing more unit tests, but will need to wait for another PR to land/merge before rebasing this PR and committing those tests.

anodos325 (approved):

PR #14692, opened 2024-10-15
NAS-131570 / 25.04 / Fix pool deletion with active NFS shares.

mgrimesix:
Deleting a pool that holds the system dataset while NFS is active can cause a failure to delete the pool.

The fix for this is to temporarily stop NFS during the system dataset move operation.

This PR also includes:
- CI test for the pool delete condition (main topic)
- Better management of `/proc/fs/nfsd/nfsv4recoverydir` and a CI test
- Adding `wait_for_standby` routine in failover test utils
- Added `# noqa` to a few lines to avoid flake8 complaints

Passing HA CI NFS tests can be found [here](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/1452/).
Passing standalone CI NFS tests can be found [here](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/1453/)

### DIFF
--- a/src/middlewared/middlewared/plugins/nfs_/fs_attachment_delegate.py
+++ b/src/middlewared/middlewared/plugins/nfs_/fs_attachment_delegate.py
@@ -1,0 +1,18 @@
+from middlewared.common.attachment import LockableFSAttachmentDelegate
+
+from middlewared.plugins.nfs import SharingNFSService
+
+
+class NFSFSAttachmentDelegate(LockableFSAttachmentDelegate):
+    name = 'nfs'
+    title = 'NFS Share'
+    service = 'nfs'
+    service_class = SharingNFSService
+    resource_name = 'path'
+
+    async def restart_reload_services(self, attachments):
+        await self._service_change('nfs', 'reload')
+
+
+async def setup(middleware):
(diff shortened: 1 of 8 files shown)


src/middlewared/middlewared/plugins/nfs_/fs_attachment_delegate.py
```diff
@@ -0,0 +1,28 @@
...
+
+    async def restart_reload_services(self, attachments):
+        await self._service_change('nfs', 'reload')
+
+    async def delete(self, attachments):
+        ''' Delete NFS shares
+        The attachments is a list of shares exporting a pool that is being deleted
+        '''
+        for attachment in attachments:
+            await self.middleware.call('sharing.nfs.delete', attachment['id'])
```
bmeagherix, line 21: I _think_ we still need to include `remove_alert` inside the `for` loop.
```
        for attachment in attachments:
            await self.middleware.call('sharing.nfs.delete', attachment['id'])
            await self.remove_alert(attachment)
```

bmeagherix (changes requested):

mgrimesix:
The reproduction and fix are more slippery than first thought.   Changing to WIP for more investigation.

bmeagherix (approved):
LGTM

PR #14711, opened 2024-10-18
NAS-131863 / 25.04 / fix drive identify on fseries

yocalebo:
Reviewing OEM documentation for this platform showed that I missed a section when implementing this logic. There is a BMC Master role that must be toggled before lighting up drives. This implements that functionality so that drive identify works no matter what controller the command originates from.


bmeagherix (approved):

yocalebo:
This breaks all the things after physically going into DC and testing. Have let platform team know.

PR #14741, opened 2024-10-21
NAS-131714 / 25.04 / Add an endpoint which UI can use to conditionally show boolean flag for removing ix-volumes on app deletion

sonicaj:
This PR adds changes to expose an endpoint which tells if ix-volumes for an app exist or not so UI can conditionally show boolean flag when app is being deleted.

We want to have an ix-volume query/delete service in place for FT and the basics for that have been added in this ticket as well.

### DIFF
--- a/src/middlewared/middlewared/plugins/apps/ix_volumes.py
+++ b/src/middlewared/middlewared/plugins/apps/ix_volumes.py
@@ -18,8 +18,8 @@ class AppsIxVolumeService(Service):
     @filterable_returns(Dict(
         'ix-volumes_query',
         Str('app_name'),
-        Str('id'),
         Str('name'),
+        additional_attrs=True,
     ))
     async def query(self, filters, options):
         """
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/apps/ix_volumes.py
```diff
@@ -0,0 +1,59 @@
...
+        apps = collections.defaultdict(list)
+        for ds_name in filter(lambda d: d.count('/') > 3, map(lambda d: d['id'], datasets)):
+            name_split = ds_name.split('/', 4)
+            apps[name_split[3]].append(name_split[-1])
+
+        volumes = []
+        for app, app_volumes in apps.items():
+            for volume in app_volumes:
+                volumes.append({
+                    'id': volume,
```
Qubad786, line 46: I think `id` should not be used as it won't be unique and is misleading

Qubad786 (approved):

PR #14751, opened 2024-10-22
NAS-131910 / 25.04 / Fix non-fulladmin user getting `datetime_` instead of `datetime` on `alert.list`

creatorcary:
This fix https://github.com/truenas/middleware/pull/14712 introduced the new problem of non-full_admin users receiving `datetime_` instead of `datetime` on `alert.list`. This revealed the broader issue of our new-style API call results not respecting field aliases in serialization.

I've also cleaned up the type annotations in alert.py.

### DIFF
--- a/src/middlewared/middlewared/alert/base.py
+++ b/src/middlewared/middlewared/alert/base.py
@@ -2,7 +2,7 @@ from datetime import datetime, timedelta
 import enum
 import json
 import logging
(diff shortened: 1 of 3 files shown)


themylogin:
Can you please add a unit test for dumping by alias?

themylogin (approved):
Thank you!

PR #14752, opened 2024-10-22
NAS-130523 / 25.04 / Initial HA support for Containers

william-gr:


Qubad786 (approved):

src/middlewared/middlewared/plugins/failover_/event.py
```diff
@@ -992,6 +1002,27 @@ def stop_apps(self):
         else:
             logger.info('Docker service stopped gracefully')
 
+    def start_virt(self):
```
yocalebo, line 1005: Can we log when we call things in `start_virt` and `stop_virt`?

yocalebo (approved):

PR #14756, opened 2024-10-23
NAS-131944 / 25.04 / Convert rdma to new api method.

mgrimesix:
To support the rdma configuration setting in the UI the newish `rdma.capable_services` routine is required as a public method.  As such, we need to convert it from private to public.

_As part of that change_, the entire `rdma.py` service module was converted to the new api method.

Also fixed a couple flake8 issues.


yocalebo:
Ticket?

mgrimesix:
added jira ticket (thanks Caleb for the gentle reminder!)

PR #14757, opened 2024-10-23
NAS-131942 / 25.04 / User proper validators for query-filters and query-options

anodos325:
The transition to new api_method made query-options too restrictive and did not implement validator for query-filters. This commit plumbs in our existing middleware validators into these models.


src/middlewared/middlewared/api/v25_04_0/common.py
```diff
@@ -1,8 +1,18 @@
 from typing import Any
```
bmeagherix, line 1: `Any` no longer used.

bmeagherix (approved):

PR #14759, opened 2024-10-23
NAS-131974 / 25.04 / convert staticroute.* plugin to new api decorator

yocalebo:
Converts the staticroute CRUD endpoints to the new api paradigm.

### DIFF
--- a/src/middlewared/middlewared/api/v25_04_0/static_route.py
+++ b/src/middlewared/middlewared/api/v25_04_0/static_route.py
@@ -33,8 +33,8 @@ class StaticRouteCreateResult(BaseModel):
     result: StaticRouteEntry
 
 
-class StaticRouteUpdate(StaticRouteEntry, metaclass=ForUpdateMetaclass):
-    id: Excluded = excluded_field()
+class StaticRouteUpdate(StaticRouteCreate, metaclass=ForUpdateMetaclass):
+    pass
 
 
 class StaticRouteUpdateArgs(BaseModel):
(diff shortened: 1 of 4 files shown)


anodos325 (approved):

src/middlewared/middlewared/api/v25_04_0/static_route.py
```diff
@@ -0,0 +1,46 @@
...
+
+class StaticRouteEntry(BaseModel):
+    destination: IPvAnyAddress
+    gateway: IPvAnyAddress
+    destription: str = ""
+    id: int
+
+
+class StaticRouteUpdateArgs(BaseModel):
+    id: Excluded = excluded_field()
```
creatorcary, line 25: I believe the fields here should exactly match the parameters of the endpoint, so `id` --> `id_`

creatorcary (changes requested):

src/middlewared/middlewared/api/v25_04_0/static_route.py
```diff
@@ -0,0 +1,54 @@
...
+
+class StaticRouteCreateArgs(BaseModel):
+    data: StaticRouteCreate
+
+
+class StaticRouteCreateResult(BaseModel):
+    result: StaticRouteEntry
+
+
+class StaticRouteUpdate(StaticRouteEntry, metaclass=ForUpdateMetaclass):
```
themylogin, line 36: It's usually a good practice to inherit update schema from create schema. That's how we used to do this in the previous API and it worked well. The benefit in this specific case: no need to exclude `id` twice

yocalebo: Yeah, I find it weird to inherit from the create class. I fixed it though. ps. this is why I only like 1 positional argument for our public methods 😄 (instead of passing `id` as the 1st positional arg. if we didn't do that it would reduce the boilerplate for the new api paradigm)

creatorcary (approved):

PR #14760, opened 2024-10-23
NAS-131950 / 25.04 / Fix dictionary key names based on field aliases

anodos325:
We want pydantic to dump by alias to maintain backwards-compatibility with old API schema and prevent introduction of new bugs.

### DIFF
--- a/src/middlewared/middlewared/api/base/handler/accept.py
+++ b/src/middlewared/middlewared/api/base/handler/accept.py
@@ -72,4 +72,9 @@ def validate_model(model: type[BaseModel], data: dict, *, exclude_unset=False, e
 
         raise verrors from None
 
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/api/base/handler/accept.py
```diff
@@ -72,4 +72,9 @@ def validate_model(model: type[BaseModel], data: dict, *, exclude_unset=False, e
 
         raise verrors from None
 
-    return instance.model_dump(context={"expose_secrets": expose_secrets}, exclude_unset=exclude_unset, warnings=False)
+    return instance.model_dump(
+        context={"expose_secrets": expose_secrets},
+        exclude_unset=exclude_unset,i
```
yocalebo, line 77: typo? (,i)

PR #14761, opened 2024-10-23
NAS-131953 / 24.10.1 / Fix apps for HA in EE

sonicaj:
This commit fixes erroneous logging and apps functionality in HA on failover event for EE because a certain block code is not relevant for EE and that logic is already ingrained in docker start service endpoint.


src/middlewared/middlewared/plugins/failover_/event.py
```diff
@@ -983,16 +983,6 @@ def start_apps_impl(self):
             return
 
         logger.info('Going to initialize apps plugin as %r pool is configured for apps', pool)
```
yocalebo, line 985: Remove this extra log statement? Doesn't seem to be useful.

yocalebo (approved):

PR #14765, opened 2024-10-24
NAS-131945 / 25.04 / Remove pydantic warning

themylogin:


themylogin:
time 0:15

yocalebo:
Some context on why simply filtering this warning instead of fixing the root of the problem would be nice.

themylogin:
@yocalebo added it as the comment in the code

yocalebo (approved):
Brilliant! Thanks!

PR #14769, opened 2024-10-24
Convert iscsi.auth.* to new api

bmeagherix:
Converts the `iscsi.auth` CRUD endpoints to the new api paradigm.

### DIFF
--- a/src/middlewared/middlewared/api/base/types/iscsi.py
+++ b/src/middlewared/middlewared/api/base/types/iscsi.py
@@ -1,0 +1,5 @@
+from typing import Literal, TypeAlias
+
+__all__ = ["IscsiAuthType"]
+
+IscsiAuthType: TypeAlias = Literal['NONE', 'CHAP', 'CHAP_MUTUAL']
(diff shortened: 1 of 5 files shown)


src/middlewared/middlewared/plugins/iscsi_/auth.py
```diff
@@ -32,21 +33,7 @@ class Config:
         cli_namespace = 'sharing.iscsi.target.auth_credential'
         role_prefix = 'SHARING_ISCSI_AUTH'
```
creatorcary, line 35: `entry = IscsiAuthEntry` can go in the `Config` class

src/middlewared/middlewared/api/base/types/iscsi.py
```diff
@@ -0,0 +1,5 @@
+from typing import Literal, TypeAlias
+
+__all__ = ["IscsiAuthType"]
+
+IscsiAuthType: TypeAlias = Literal['NONE', 'CHAP', 'CHAP_MUTUAL']
```
creatorcary, line 5: Is this being used?

creatorcary (changes requested):

src/middlewared/middlewared/api/base/types/iscsi.py
bmeagherix: Not yet, but the plan is that it will be.

yocalebo:
@creatorcary ping

creatorcary (approved):

PR #14773, opened 2024-10-24
NAS-131863 / 25.04 / fix edge-case w/ drive ident on f-series

yocalebo:
This is an edge-case that we need to try and work-around until OEM can give definitive answer/fix.


yocalebo:
backport

anodos325 (approved):

PR #14777, opened 2024-10-25
NAS-131986 / 25.04 / Add integration test for apps

sonicaj:


src/middlewared/middlewared/test/integration/assets/apps.py
```diff
@@ -0,0 +1,13 @@
+import contextlib
+
+from middlewared.test.integration.utils import call
+
+
+@contextlib.contextmanager
+def create_app(app_name: str, create_payload: dict, delete_payload: dict | None = None):
```
themylogin, line 7: We usually test assets just `app`, so that the code looks good wrt to English grammar `with app(...)`, or is there any particular reason why here it should be different?

tests/api2/test_apps.py
```diff
@@ -0,0 +1,288 @@
...
+'''
+
+
+@pytest.fixture(scope='session')
+def docker_pool():
+    with another_pool() as pool:
+        yield pool['name']
+
+
+@pytest.mark.dependency(name='docker_setup')
```
themylogin, line 144: Should this be a fixture too? Then you won't need to have `depends=['docker_setup']` in all other tests. Also you won't need to have a hack-ish last test in this file, the fixture will teardown itself correctly.

themylogin (approved):

PR #14779, opened 2024-10-25
NAS-131903 / 24.10.1 / Properly handle case when local IP is not retrievable via application during REST call

Qubad786:
## Problem

For REST calls being made in EE, we are not able to retrieve local ip in our `websocket_local_ip` endpoint as that expects a websocket connection and any endpoint consuming this will fail when the endpoint is consumed via REST.

## Solution

After discussion with Caleb, we don't want to try and retrieve local ip for REST in EE and we should gracefully just handle it.


yocalebo (approved):

Qubad786:
backport

PR #14782, opened 2024-10-25
NAS-131804 / 25.04 / Show progress on cloud_backup.sync and restore

creatorcary:
Our TrueCloud Backup Tasks use restic to back up local data. The procedure we use to capture restic's output does not capture progress updates. By setting the `--json` flag we can receive continual status updates as JSON objects while the command runs. This allows us to monitor things like the percentage of completion, time remaining, and the status of files being accessed (new/updated/removed etc.) on `cloud_backup.sync` and `cloud_backup.restore`.

Tests: http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/1571/

https://github.com/user-attachments/assets/c5cf870d-f4c8-49ad-9362-556047b1a9e5


creatorcary:
ping @truenas/middleware @themylogin

src/middlewared/middlewared/plugins/cloud_backup/restic.py
```diff
@@ -37,38 +40,80 @@ async def run_restic(job, cmd, env, stdin=None):
...
 
 
-async def restic_check_progress(job, proc):
+async def restic_check_progress(job, proc, cmd: list[str]):
+    """Record progress of restic backup, restore, and forget commands.
+
+    Relevant documentation: https://restic.readthedocs.io/en/stable/075_scripting.html#json-output
+
+    """
+    if "forget" in cmd:
```
themylogin, line 73: We should make this an explicit flag, saying if we want to check progress or just want to copy the logs.

src/middlewared/middlewared/plugins/cloud_backup/restic.py
```diff
@@ -37,38 +40,80 @@ async def run_restic(job, cmd, env, stdin=None):
...
+                time_delta = str(timedelta(seconds=remaining))
+
+            progress_buffer.set_progress(
+                read["percent_done"] * 100,
+                (f"{time_delta} remaining\n" if time_delta else "") + action
+            )
+            continue
+
+        await job.logs_fd_write((json.dumps(read) + "\n").encode("utf-8", "ignore"))
+        if msg_type == "summary":
```
themylogin, line 101: We can use match/case syntax here

themylogin (changes requested):

src/middlewared/middlewared/plugins/cloud_backup/restic.py
creatorcary: We can, but we'll need to note that setting the flag on `restic forget` will cause this to fail since `forget` returns different JSON output than `backup` and `restore`.

themylogin: I understand, let's just not use methods like `"forget" in cmd` to determine what the output will be; let's use explicit flags instead

themylogin (approved):

creatorcary:
backport

PR #14783, opened 2024-10-25
NAS-130523 / 25.04 / Allow null for pool/bridge

william-gr:
UI team has complained they are not allowed to use `""` as value for pool/bridge and these are returned by pool_choices/bridge_choices.

choices cannot return `null` as keys in a dict, hence the need to support empty string.

### DIFF
--- a/src/middlewared/middlewared/api/v25_04_0/virt.py
+++ b/src/middlewared/middlewared/api/v25_04_0/virt.py
@@ -22,8 +22,8 @@ class VirtGlobalEntry(BaseModel):
 
 @single_argument_args('virt_global_update')
 class VirtGlobalUpdateArgs(BaseModel, metaclass=ForUpdateMetaclass):
-    pool: str | None = None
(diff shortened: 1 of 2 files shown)


Qubad786 (approved):
Choices should be able to return `None` as well?

william-gr:
@Qubad786 

```
    @api_method(VirtGlobalPoolChoicesArgs, VirtGlobalPoolChoicesResult, roles=['VIRT_GLOBAL_READ'])
    async def pool_choices(self):
        """ 
        Pool choices for virtualization purposes.
        """
        pools = {None: '[Disabled]'}
```

```
root@truenas:~# midclt call virt.global.pool_choices
{"null": "[Disabled]", "tank": "tank"}
```

Though I am discussing with Caleb how to move forward.

william-gr:
@Qubad786 this is what Caleb preference was. Let me know if this is also OK with you please.

Qubad786 (approved):

PR #14788, opened 2024-10-26
NAS-132011 / 25.04 / Add ability to allow method on remote node to be mocked

bmeagherix:
Add an optional remote parameter to the mock fixture, that will cause the mock to occur on the remote node in a HA pair.


anodos325 (approved):

anodos325:
Very nice and useful.

PR #14791, opened 2024-10-28
Modify config service to inject new style api models instead of accepts/returns

sonicaj:


src/middlewared/middlewared/service/config_service.py
```diff
@@ -29,17 +35,62 @@ def __new__(cls, name, bases, attrs):
...
-        klass.config = returns(config_entry)(klass.config)
+        config_model_name = f'{namespace.capitalize()}Config'
+
+        if klass._config.entry is not None or klass.ENTRY == NotImplementedError:
+            # FIXME: We do this temporarily to avoid breaking existing code
+            klass.ENTRY = None
+            if klass._config.entry is None:
+                klass._config.entry = create_model(
+                    f'{namespace.capitalize()}Entry',
+                    __base__=(AllowExtraBaseModel,),
```
themylogin, line 46: Why should extra fields be allowed?

src/middlewared/middlewared/service/config_service.py
```diff
@@ -29,17 +35,62 @@ def __new__(cls, name, bases, attrs):
...
-        config_entry.register = True
-        klass.config = returns(config_entry)(klass.config)
+        config_model_name = f'{namespace.capitalize()}Config'
+
+        if klass._config.entry is not None or klass.ENTRY == NotImplementedError:
+            # FIXME: We do this temporarily to avoid breaking existing code
+            klass.ENTRY = None
+            if klass._config.entry is None:
+                klass._config.entry = create_model(
+                    f'{namespace.capitalize()}Entry',
```
themylogin, line 45: Autogenerated classes should absolutely not be the case with the versioned API approach we have now. Everything must be declared explicitly even if it is repetitive

src/middlewared/middlewared/service/config_service.py
sonicaj: If entry is not defined, in that case by default we will allow any fields to be there. This is not what we want in the final stage of our api but when there is no entry defined i.e no schema in place - then that is what we would be going with to make sure the new service works.

src/middlewared/middlewared/service/config_service.py
sonicaj: Okay so after hashing it out with @themylogin, decided approach is to not touch the `config.update` method but apply it to `config` method for it's result unless overridden in the child class.

src/middlewared/middlewared/api/v25_04_0/docker.py
```diff
@@ -0,0 +1,59 @@
+import typing
+
+from pydantic import conint, IPvAnyInterface, field_validator
+
+from middlewared.api.base import (
+    BaseModel, Excluded, excluded_field, ForUpdateMetaclass, NonEmptyString, single_argument_args,
+)
+from middlewared.plugins.docker.state_utils import Status
```
themylogin, line 8: We can't do this since API must be immutable. If you change `Status` definition, API definition will change.

src/middlewared/middlewared/api/v25_04_0/docker_network.py
```diff
@@ -0,0 +1,19 @@
+from pydantic import ConfigDict
+
+from middlewared.api.base import BaseModel, NonEmptyString
+
+
+class DockerNetworkEntry(BaseModel):
+
+    model_config = ConfigDict(
```
themylogin, line 8: What extra fields are there?

sonicaj: Docker api provides a bunch of fields we are not really concerned about atm and hence extra has been allowed. If you think we should be restrictive, i can do that ?

themylogin: Thing is if we don't remove those fields, someone might start using them. Then the fields might disappear/change without notice, and things will break. My opinion is that we should either document those fields explicitly or remove them.

themylogin (approved):
Thank you!

src/middlewared/middlewared/plugins/docker/validation_utils.py
```diff
@@ -15,7 +15,11 @@ def validate_address_pools(system_ips: list[dict], user_specified_networks: list
     ])
     seen_networks = set()
     for index, user_network in enumerate(user_specified_networks):
-        base_network = ipaddress.ip_network(user_network['base'], False)
+        if isinstance(user_network['base'], ipaddress.IPv4Interface):
```
yocalebo, line 18: You don't need to do this check. The `ipaddress` modules accept an instance of themselves as an argument. I discovered this when I made changes converting the static route plugin (See [here](https://github.com/truenas/middleware/pull/14759/files#diff-c763d4b1445b671af3b54cbf8038617f8705014f593f4cda024cbd4286734d12R116))

yocalebo (changes requested):

src/middlewared/middlewared/plugins/docker/validation_utils.py
sonicaj: This is for the unit tests i have for the same endpoint, hence the explicit check

yocalebo (approved):

PR #14795, opened 2024-10-28
NAS-132045 / 25.04 / Convert NTP server plugin to api_method

anodos325:
This commit converts ntp.py to use new api_method as well as introducing and using a new `filterable_api_method` decorator for non-CRUD methods that were using legacy `filterable` decorator.

### DIFF
--- a/src/middlewared/middlewared/service/decorators.py
+++ b/src/middlewared/middlewared/service/decorators.py
@@ -60,6 +60,9 @@ def filterable_api_method(fn=None, /, *, roles=None, item=None, private=True):
         if item:
             returns = query_result(item)
         else:
+            if not private:
+                raise ValueError('Public methods may not use GenericQueryResult.')
+
             returns = GenericQueryResult
 
         return api_method(QueryArgs, returns, private=private, roles=roles)(fn)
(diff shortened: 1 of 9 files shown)


src/middlewared/middlewared/service/decorators.py
```diff
@@ -48,6 +51,26 @@ def filterable_internal(fn):
...
+def filterable_api_method(fn=None, /, *, roles=None, item=None, private=True):
+    def filterable_internal(fn):
+        fn._filterable = True
+        if hasattr(fn, 'wraps'):
+            fn.wraps._filterable = True
+
+        if item:
+            returns = query_result(item)
+        else:
+            returns = GenericQueryResult
```
themylogin, line 63: We should not allow this, we should introduce schemas for everything. Or is it impossible in this case?

anodos325: I envision this generally being used in private APIs. We have large amount of private usage of `filter_list` that can benefit from the normalization and validation. I don't see benefit to providing versioning for private APIs (it is in fact probably harmful in long-run).

I can make it so that `item` is required if non-private.

themylogin: Excellent idea, let’s do that!

creatorcary (approved):

PR #14796, opened 2024-10-28
Rename rdma.capable_services to rdma.capable_protocols.

mgrimesix:
The name for method `rdma.capable_services` is misleading in that what is returned is not related to `services`.  The return lists the supported 'protocols'.  Currently, this is only 'NFS'.

Changed the name of the method to `rdma.capable_protocols`.  The returns are unchanged.

Added a CI test.

### DIFF
--- a/tests/api2/test_300_nfs.py
+++ b/tests/api2/test_300_nfs.py
@@ -1734,6 +1734,35 @@ class TestNFSops:
             s = parse_server_config()
             assert s['mountd']['manage-gids'] == expected, str(s)
 
+    def test_rdma_config(self, start_nfs):
+        '''
+        Mock response from rdma.capable_protocols to confirm NFS over RDMA config setting
+        '''
+        assert start_nfs is True
+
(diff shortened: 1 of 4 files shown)


tests/api2/test_300_nfs.py
```diff
@@ -1734,6 +1734,35 @@ def test_manage_gids(self, start_nfs, state, expected):
...
+            ValidationError(
+                'nfs_update.rdma',
+                'This platform cannot support NFS over RDMA or is missing an RDMA capable NIC.',
+                22
+            )
+        ]
+
+        with mock("rdma.capable_protocols", return_value=['NFS']):
+            with nfs_config():
+                call("nfs.update", {"rdma": True})
```
bmeagherix, line 1760: Turn it off again below?

mgrimesix: That's the point of `with nfs_config()`.  It returns the config to original.

bmeagherix (approved):

PR #14797, opened 2024-10-29
NAS-132059 / 25.04 / Fix failing 2fa tests

yocalebo:
These are tripping the rate limit logic in middleware. There is already a function level fixture for clearing the rate limit using the root account, so I simply passed that fixture to the remaining tests.

### DIFF
--- a/tests/api2/test_twofactor_auth.py
+++ b/tests/api2/test_twofactor_auth.py
@@ -1,12 +1,7 @@
-#!/usr/bin/env python3
 import contextlib
 import errno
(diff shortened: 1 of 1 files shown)


yocalebo:
Passing tests are [here](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/1562/testReport/api2/test_twofactor_auth/)

themylogin (approved):

anodos325 (approved):

PR #14798, opened 2024-10-29
NAS-131785 / 25.04 / Make sure serial is specified for all VM disk devices

Qubad786:
## Problem

When disk devices are created in VMs, they don't have a serial present. Some types of disks may have an autogenerated one from libvirt but then we can potentially get persistence issues wrt serials and for some types, libvirt doesn't even add auto-generated ones.

## Solution

Generate a random serial number and assign it to VM disks making sure the serial persists across the device lifecycle.

### DIFF
--- a/src/middlewared/middlewared/alembic/versions/25.04/2024-10-29_12-43_vm_zvol_serial.py
+++ b/src/middlewared/middlewared/alembic/versions/25.04/2024-10-29_12-43_vm_zvol_serial.py
@@ -1,0 +1,48 @@
+"""
+Add serial to VM zvol disk devices
+
+Revision ID: a4ce4939b908
+Revises: dd6e581235b2
+Create Date: 2024-10-29 12:43:45.960915+00:00
+
+"""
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/alembic/versions/25.04/2024-10-29_12-43_vm_zvol_serial.py
```diff
@@ -0,0 +1,31 @@
...
+Revision ID: a4ce4939b908
+Revises: dd6e581235b2
+Create Date: 2024-10-29 12:43:45.960915+00:00
+
+"""
+import json
+
+from alembic import op
+
+from middlewared.utils.crypto import generate_string
```
themylogin, line 13: Please just copypaste the function here. We should not import anything in migrations code, they should be immutable.

themylogin (changes requested):

themylogin (approved):

PR #14803, opened 2024-10-29
NAS-130523 / 25.04 / Fix resetting incus and network ranges

william-gr:


src/middlewared/middlewared/plugins/virt/instance.py
```diff
@@ -36,9 +36,10 @@ async def query(self, filters, options):
         """
         Query all instances with `query-filters` and `query-options`.
         """
-        config = await self.middleware.call('virt.global.config')
-        if config['state'] != 'INITIALIZED':
-            return []
+        if not options.get('extra').get('skip_state'):
```
Qubad786, line 39: `options.get('extra').get()` seems kinda weird..iirc extra should always be there now

Qubad786 (approved):

src/middlewared/middlewared/plugins/virt/instance.py
william-gr: Indeed, thanks

PR #14804, opened 2024-10-29
NAS-132069 / 25.04 / Convert remaining auth plugin to api_method

anodos325:
This is mostly boiler-plate code with exception of the following:

1. default for match_origin in auth.generate_token to True as a general improvement to product security.

2. TrueNasNodeSessionManagerCredentials class was renamed to TruenasNodeSessionManagerCredentials to ensure that generated credentials class name is TRUENAS_NODE rather than TRUE_NAS_NODE.

3. auth.generate_token was changed from not_auth_required with separate internal check for session being authenticated to no_authz_required.

4. api_method is extended to support no_authz_required and no_auth_required kwargs.

### DIFF
--- a/src/middlewared/middlewared/api/base/decorator.py
+++ b/src/middlewared/middlewared/api/base/decorator.py
@@ -19,6 +19,8 @@ def api_method(
     rate_limit=True,
     roles: list[str] | None = None,
     private: bool = False,
+    authentication_required: bool = True,
+    authorization_required: bool = True,
 ):
     """
     Mark a `Service` class method as an API method.
@@ -40,6 +42,14 @@ def api_method(
     `roles` is a list of user roles that will gain access to this method.
 
     `private` is `True` when the method should not be exposed in the public API. By default, the method is public.
+
+    `authentication_required` is False when API endpoint does not require authentication. This should generally
+    *not* be set and requires appropriate review and approval to validate that its use complies with security
+    standards. This is incompatible with `roles`.
+
+    `authorization_required` is False API endpoint does not require authorization, but does require authentication.
+    This is incompatible with `roles`. Additional review will be required in order to validate that its use complies
+    with security standards.
     """
     if list(returns.model_fields.keys()) != ["result"]:
         raise TypeError("`returns` model must only have one field called `result`")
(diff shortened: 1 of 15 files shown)


src/middlewared/middlewared/plugins/auth.py
```diff
@@ -210,17 +215,7 @@ def __init__(self, *args, **kwargs):
...
-        'session',
-        Str('id'),
-        Bool('current'),
-        Bool('internal'),
-        Str('origin'),
-        Str('credentials'),
-        Dict('credentials_data', additional_attrs=True),
-        Datetime('created_at'),
-    ))
+    @filterable_api_method(item=AuthSessionEntry, private=False, roles=['AUTH_SESSIONS_READ'])
```
themylogin, line 218: Isn’t  `private=False` a default?

src/middlewared/middlewared/plugins/auth.py
```diff
@@ -306,13 +300,10 @@ async def terminate_other_sessions(self, app):
...
-    @no_auth_required
-    @accepts(
-        Int('ttl', default=600, null=True),
-        Dict('attrs', additional_attrs=True),
-        Bool('match_origin', default=False),
-    )
-    @returns(Str('token'))
+        return True
+
+    @no_authz_required
```
themylogin, line 305: Let’s add `@no_authz_required` functionality to `@api_method`

src/middlewared/middlewared/plugins/auth.py
anodos325: Actually, I got that backwards. `private=True` is default for `filterable_api_method`.

themylogin: That discrepancy with `@api_method` might cause a lot of confusion...

themylogin: That discrepancy with `@api_method` might cause a lot of confusion...

src/middlewared/middlewared/api/base/decorator.py
```diff
@@ -19,6 +19,8 @@ def api_method(
     rate_limit=True,
     roles: list[str] | None = None,
     private: bool = False,
+    no_auth_required: bool = False,
```
themylogin, line 22: Should we avoid double negation with `auth_required: bool = True`?

anodos325: That's not a bad idea.

src/middlewared/middlewared/plugins/auth.py
anodos325: OK. Will change.

themylogin (approved):
Excellent, thank you!

PR #14809, opened 2024-10-30
NAS-132097 / 25.04 / Add a migration to remove unused k8s table fields

sonicaj:


themylogin:
What will happen with 24.04 -> 25.04 upgrades?

sonicaj:
Nothing, these columns are not being used ?

yocalebo (approved):

themylogin:
I mean, were they never used?

sonicaj:
They were used when we had k8s but when we removed it, the table is not required anymore - we just want to know what pool was configured at the point and all other columns are redundant.

PR #14810, opened 2024-10-30
NAS-131911 / 25.04 / Allow removing ix-volumes of apps which were migrated from k8s

sonicaj:
This PR adds changes to allow removing ix-volumes of apps which were migrated over from k8s. When we migrated apps from k8s and they had ix-volumes, what we did was that we created a clone and promoted the clone under the new ix-apps dataset. So in order to delete these ix-volumes, we want to properly make sure all dependents are removed.

### DIFF
--- a/src/middlewared/middlewared/plugins/zfs_/dataset.py
+++ b/src/middlewared/middlewared/plugins/zfs_/dataset.py
@@ -191,11 +191,12 @@ class ZFSDatasetService(CRUDService):
     def do_delete(self, id_, options):
         force = options['force']
         recursive = options['recursive']
+        recursively_remove_dependents = options['recursively_remove_dependents']
 
         args = []
         if force:
             args += ['-f']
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/zfs_/dataset.py
```diff
@@ -194,7 +195,9 @@ def do_delete(self, id_, options):
         args = []
         if force:
             args += ['-f']
-        if recursive:
+        if options['recursively_remove_dependents']:
```
yocalebo, line 198: In this same method you're assigning variables based on the options that are set but this one option you're not doing the same. Let's be consistent and either
1. remove the `recursive` and `force` variables and check `options["recursive"]` etc
2. OR pull the `recursively_remove_dependents` into its own variable as well.

yocalebo (changes requested):

yocalebo (approved):

PR #14811, opened 2024-10-30
NAS-130523 / 25.04 / Archlinux has been failing to install netcat with 404 error

william-gr:


yocalebo (changes requested):
Let's add 2 global variables to this file so we don't have to update all these static strings everywhere. Something like
```
NAME: "centos"
IMAGE: "centos/9-Stream"

william-gr:
image

yocalebo (approved):
Nice!

PR #14812, opened 2024-10-30
NAS-132042 / 25.04 / Make all fields in query results non-required

themylogin:
All fields in the query result must be non-required since we can query subsets of fields.

### DIFF
--- a/src/middlewared/middlewared/api/base/model.py
+++ b/src/middlewared/middlewared/api/base/model.py
@@ -183,7 +183,7 @@ def single_argument_result(klass, klass_name=None):
         klass_name,
         __base__=(BaseModel,),
         __module__=inspect.getmodule(inspect.stack()[1][0]),
-        **{"result": Annotated[klass, Field()]},
+        result=Annotated[klass, Field()],
     )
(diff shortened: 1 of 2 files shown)


themylogin:
time 1:00

yocalebo (changes requested):
Please give a description of why we're doing this. (i.e. backwards compatible with legacy query behavior, etc, etc)

themylogin:
@yocalebo there is a comment there https://github.com/truenas/middleware/pull/14812/files#diff-456f5ca8165c53cf24278ed098d726e1fc3f61c1cc6f8993a5aa4817d76e3eaeR205 do you want me to elaborate it more?

yocalebo:
> @yocalebo there is a comment there https://github.com/truenas/middleware/pull/14812/files#diff-456f5ca8165c53cf24278ed098d726e1fc3f61c1cc6f8993a5aa4817d76e3eaeR205 do you want me to elaborate it more?

You can just copy and paste that same comment in the body of this PR. Idea is that someone can see the PR and get a general idea of why this happening.

yocalebo (approved):

PR #14835, opened 2024-10-31
NAS-131249 / 25.04 / Add middleware hooks to configure dedup_table* fields

Qubad786:
## Context

It was requested that we expose ability to prefetch/prune ddt entries and relevant changes have been added to accommodate that. Also it was requested that we preferably prefetch after pool has been imported and those changes have been accommodated as well on both pool import and failover.

### DIFF
--- a/src/middlewared/middlewared/plugins/failover_/event.py
+++ b/src/middlewared/middlewared/plugins/failover_/event.py
@@ -676,6 +676,9 @@ class FailoverEventsService(Service):
         self.run_call('directoryservices.setup')
         logger.info('Done starting background job for directoryservices.setup')
 
+        logger.info('Starting background job for prefetching DDT for zpools')
+        self.middleware.create_task(self.middleware.call('zfs.pool.ddt_prefetch_pools'))
+
         logger.info('Allowing network traffic.')
         fw_accept_job = self.run_call('failover.firewall.accept_all')
         fw_accept_job.wait_sync()
--- a/src/middlewared/middlewared/plugins/pool_/pool.py
+++ b/src/middlewared/middlewared/plugins/pool_/pool.py
@@ -80,6 +80,9 @@ class PoolService(CRUDService):
         Int('allocated', required=True, null=True),
         Int('free', required=True, null=True),
         Int('freeing', required=True, null=True),
+        Int('dedupcached', required=True, null=True),
+        Int('dedup_table_size', required=True, null=True),
+        Str('dedup_table_quota', required=True, null=True),
         Str('fragmentation', required=True, null=True),
         Str('size_str', required=True, null=True),
         Str('allocated_str', required=True, null=True),
@@ -159,6 +162,9 @@ class PoolService(CRUDService):
             'allocated_str': None,
             'free_str': None,
             'freeing_str': None,
+            'dedup_table_quota': None,
+            'dedup_table_size': None,
+            'dedupcached': None,
             'autotrim': {
                 'parsed': 'off',
                 'rawvalue': 'off',
@@ -196,6 +202,9 @@ class PoolService(CRUDService):
                 'free_str': info['properties']['free']['rawvalue'],
                 'freeing_str': info['properties']['freeing']['rawvalue'],
                 'autotrim': info['properties']['autotrim'],
+                'dedup_table_quota': info['properties']['dedup_table_quota']['parsed'],
+                'dedup_table_size': info['properties']['dedup_table_size']['parsed'],
+                'dedupcached': info['properties']['dedupcached']['parsed'],
             })
 
         return rv
@@ -408,6 +417,8 @@ class PoolService(CRUDService):
         'pool_create',
         Str('name', max_length=50, required=True),
         Bool('encryption', default=False),
+        Str('dedup_table_quota', default='AUTO', enum=['AUTO', None, 'CUSTOM'], null=True),
+        Int('dedup_table_quota_value', null=True, default=None, validators=[Range(min_=1)]),
         Str('deduplication', enum=[None, 'ON', 'VERIFY', 'OFF'], default=None, null=True),
         Str('checksum', enum=[None] + ZFS_CHECKSUM_CHOICES, default=None, null=True),
         Dict(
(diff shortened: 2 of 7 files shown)


src/middlewared/middlewared/plugins/failover_/event.py
```diff
@@ -644,6 +644,7 @@ def vrrp_master(self, job, fobj, ifname, event):
                     'However, other zpools imported so the failover process continued.'
                 )
         else:
+            self.middleware.create_task(self.middleware.call('zfs.pool.ddt_prefetch_pools'))
```
yocalebo, line 647: I don't think we should do this. I have no idea how long this takes or the performance implications on the zpools but not to fond of this change. We need @amotin and or @ixhamza, @usaleem-ix 's input.

amotin: We could run it asynchronously without waiting for completion.  May be even few seconds after pool import, so that other more important processes had time to start.  The prefetch may take some time, but without it on a pool with large DDT/BRT performance may suffer for hours until they get fully loaded on-demand.

src/middlewared/middlewared/plugins/zfs_/pool_actions.py
```diff
@@ -238,6 +239,43 @@ def import_pool(self, name_or_guid, properties, any_host, cachefile, new_name, i
...
+    def ddt_prefetch(self, pool_name):
+        with libzfs.ZFS() as zfs:
+            try:
+                zfs.get(pool_name).ddt_prefetch()
+            except libzfs.ZFSException as e:
+                raise CallError(str(e), e.code)
+
+    def ddt_prefetch_pools(self):
+        # We have this method so to avoid making excessive calls to process pool service
+        for pool_info in (self.middleware.call('zfs.pool.query_imported_fast')).values():
```
yocalebo, line 252: call_sync

src/middlewared/middlewared/plugins/zfs_/pool_actions.py
```diff
@@ -238,6 +239,43 @@ def import_pool(self, name_or_guid, properties, any_host, cachefile, new_name, i
...
+            try:
+                self.middleware.logger.info('Prefetching ddt table of %r pool', pool_info['name'])
+                self.ddt_prefetch(pool_info['name'])
```
yocalebo, line 255: This returns boot drive IIRC

src/middlewared/middlewared/plugins/zfs_/pool_actions.py
```diff
@@ -238,6 +239,43 @@ def import_pool(self, name_or_guid, properties, any_host, cachefile, new_name, i
...
+            except CallError as e:
+                self.logger.error('Failed to prefetch DDT for pool %r: %s', pool_info['name'], e)
+
+    @accepts(
```
yocalebo, line 259: This should use new API and I don't want more than 1 positional arg. Please make it accept a dictionary only

yocalebo (changes requested):

src/middlewared/middlewared/plugins/failover_/event.py
Qubad786: This is being run in the background and we are not waiting for it to complete execution or halt the failover event.

yocalebo: Yeah I know it's being run in background but I was talking about whether this task kicks off IO on the zpool that causes performance degradation.

Qubad786: Ah no, i meant to clarify to @amotin that it is running asynchronously and not blocking at this point. Based off that, please advise @amotin if you think we should let it be and keep doing it here?

src/middlewared/middlewared/plugins/pool_/pool.py
```diff
@@ -680,6 +697,30 @@ async def do_create(self, job, data):
         self.middleware.send_event('pool.query', 'ADDED', id=pool_id, fields=pool)
         return pool
 
+    @private
+    async def validate_dedup_table_quota(self, data, verrors, schema='pool_create'):
+        if data.get('dedup_table_quota') != 'CUSTOM' and data.get('dedup_table_quota_value') is not None:
```
yocalebo, line 702: Instead of doing `data.get` 4 times on the same variables. Why not `data.get` only 2 times and store in variables and then do checks?

src/middlewared/middlewared/plugins/failover_/event.py
yocalebo, line 647: This will fail on pools that have the root dataset encrypted. I'd rather move this call to further down in the failover event. I'd like to actually call this after we start the critical services (NFS, SMB and iSCSI)

yocalebo (changes requested):

src/middlewared/middlewared/plugins/failover_/event.py
Qubad786: We do at this point have root datasets unlocked already but yes i'll move it to after we have restarted critical services ✅

src/middlewared/middlewared/plugins/pool_/pool.py
```diff
@@ -680,6 +697,31 @@ async def do_create(self, job, data):
...
+                f'{schema}.dedup_table_quota',
+                'You must set Deduplication Table Quota to CUSTOM to specify a value.',
+            )
+        elif dedup_table_quota == 'CUSTOM' and dedup_table_quota_value is None:
+            verrors.add(
+                f'{schema}.dedup_table_quota_value',
+                'This field is required when Deduplication Table Quota is set to CUSTOM.',
+            )
+
+        if verrors or 'dedup_table_quota' not in data:
```
yocalebo, line 715: How can this be true?
```
if verrors or 'dedup_table_quota' not in data:
    return

if deduce_table_quota is None:
    return 'none'
```

The 2nd if branch can't be reached. It's also weird to still be checking if `dedup_table_quota not in data`

Qubad786: The first can be true for the update case when default value will not be populated. So we explicitly check if the key has been set or not because if it has been specified and there are no validation errors up till this point - then the value is safe to consume

yocalebo (approved):
It's still not reachable code but I'll approve so we can move on.

amotin (approved):
Have no objections as much as I can review python.  I would just like to mention that `dedupcached` property is more expensive than others, so it is not reported by default unless you request it explicitly, which should not be done routinely without a reason.

PR #14836, opened 2024-10-31
NAS-132150 / 25.04 / NAS-132150 - Fix instructions for location of Update Page

kmoore134:


yocalebo:
We should reconsider the text of these types of alerts. I think we should probably just point to online documentation? Otherwise, this will break anytime UI decides to change the location again.

kmoore134:
I'm fine with long term we come up with a better way to handle this. Ideally UI shows status on the page when there are updates and we don't need the "alert" at all. But for now this fixes for FT anyway.

yocalebo (approved):
Yep agree.

kmoore134:
backport

PR #14844, opened 2024-11-01
NAS-132164 / 25.04 / Fix `ReadonlyRootfsManager` not really making rootfs writeable

themylogin:
As per @usaleem-ix's explanation

> There was an issue in ZFS, datasets mounted during boot did not have correct permissions because mount command from busybox was used in initramfs to  mount the datasets. this mount command does mount syscall directly into the kernel, bypassing the mount helper program also used by mount from util-linux package, that is available when boot is completed. https://ixsystems.atlassian.net/browse/NAS-127825
> This issue was fixed recently. Now on SCALE, we have datasets like usr, etc and others now mounted with correct permissions. We set readonly=on for /usr, same goes for root (/). We do this from fhs.py in scale-build. These options were not being enforced previously, but after the fix for NAS-127825 this test case broke.

### DIFF
--- a/src/middlewared/middlewared/plugins/boot.py
+++ b/src/middlewared/middlewared/plugins/boot.py
@@ -209,22 +209,6 @@ class BootService(Service):
         )
         return interval
 
-    @asynccontextmanager
-    async def __toggle_rootfs_readwrite(self):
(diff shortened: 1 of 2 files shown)


themylogin:
time 2:00

usaleem-ix (approved):

PR #14855, opened 2024-11-02
NAS-132196 / 25.04 / Fix bugs in ACME DNS auth service and convert schemas to new api style

sonicaj:

1. Refactors some code to reflect new guidelines we have in place for defining pydantic models of services
2. Fixes some bugs with existing models with key mismatches and args handling
3. Convert authenticator schemas to use new api style pydantic models


src/middlewared/middlewared/api/base/types/base/path.py
```diff
@@ -0,0 +1,42 @@
+from pathlib import Path
```
yocalebo, line 1: This entire file can be removed and replaced with the built-in `FilePath` builtin pydantic annotation. Please use that instead.

yocalebo (changes requested):

creatorcary:
Why are we adding `validate_credentials` to our versioned api? (Are they public endpoints?)

sonicaj:
@creatorcary they are not middleware endpoints at all, it is just making use of the accepts/returns model with pure python as these are not services themselves, they will never get collected by middleware. @themylogin do you think we should have another util which `api_method` can call and that can hold this logic ?

themylogin:
@sonicaj it’s better to have all models be accepted immediately by API methods. The reason is: if models change between API definitions, who will convert them? Please, try to fit these models into the function calls themselves using a discriminator (that’s already done for new VMs, and that’s what I’ll soon do for alerts and cloud syncs).

themylogin (approved):
Excellent, thank you!

yocalebo (approved):

PR #14863, opened 2024-11-05
Nas 131955

bmeagherix:
Handle some exceptions so that we more fully tear down a JBOF, even if we can't currently communicate with it.

### DIFF
--- a/src/middlewared/middlewared/plugins/rdma/interface/crud.py
+++ b/src/middlewared/middlewared/plugins/rdma/interface/crud.py
@@ -97,9 +97,12 @@ class RDMAInterfaceService(CRUDService):
         data = await self.get_instance(id_)
 
         # Attempt to remove the live configuration
-        result = await self.middleware.call('rdma.interface.configure_interface', data['node'], data['ifname'], None)
-        if not result:
-            self.logger.warn("Failed to delete active RDMA interface configuration")
+        try:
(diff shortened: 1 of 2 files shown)


src/middlewared/middlewared/plugins/rdma/interface/crud.py
```diff
@@ -97,9 +97,12 @@ async def do_delete(self, id_):
         data = await self.get_instance(id_)
 
         # Attempt to remove the live configuration
-        result = await self.middleware.call('rdma.interface.configure_interface', data['node'], data['ifname'], None)
-        if not result:
-            self.logger.warn("Failed to delete active RDMA interface configuration")
+        try:
+            result = await self.middleware.call('rdma.interface.configure_interface', data['node'], data['ifname'], None)
+            if not result:
+                self.logger.warn("Failed to delete active RDMA interface configuration")
```
yocalebo, line 103: warn has been deprecated. Should use warning

src/middlewared/middlewared/plugins/rdma/interface/crud.py
```diff
@@ -97,9 +97,12 @@ async def do_delete(self, id_):
...
+        except Exception:
+            self.logger.error('Failed to remove live RDMA configuration', exc_info=True)
```
yocalebo, line 105: You can do self.logger.exception to remove the exc_info=True. Not required just fyi

yocalebo (approved):

PR #14865, opened 2024-11-05
NAS-131926 / 25.04 / add new boot.environment plugin

yocalebo:
The `bootenv.query` endpoint was taking an exorbitant amount of time to return basic information about the boot environments. For example, on a system with 10 boot environments, it was taking ~3 seconds to complete.
```
root@truenas[~]# zectl list -H | wc -l
10
root@truenas[~]# time midclt call bootenv.query > /dev/null
midclt call bootenv.query > /dev/null  0.12s user 0.02s system 4% cpu 2.806 total
```
With the changes implemented here, this now returns ~93% faster
```
root@truenas[~]# time midclt call boot.environment.query > /dev/null
midclt call boot.environment.query > /dev/null  0.11s user 0.02s system 64% cpu 0.206 total
```

It should be noted that I removed a bunch of weird space calculation logic that we were doing in the legacy plugin. I understand the sentiment but it is not needed and we can return the default zfs space used property. Finally, I've added tests to cover essentially 100% of the functionality implemented with this new plugin (passing tests [here](http://jenkins.eng.ixsystems.net:8080/job/tests/job/api_tests/1700/testReport/api2/test_boot_environments/)).

NOTE: Subsequent PRs will be made so that callers end up using the new plugin.


src/middlewared/middlewared/plugins/boot_/environments.py
```diff
@@ -0,0 +1,185 @@
...
+
+    @private
+    def zfs_get_props(self):
+        zget = "zfs get -r -d 1"
+        zget += " truenas:kernel_version,zectl:keep,used,creation"
+        bp_name = self.middleware.call_sync("boot.pool_name")
+        root = f"{bp_name}/ROOT"
+        json_flags = "-j --json-int"  # json format, raw numbers
+        try:
+            cp = subprocess.run(
```
themylogin, line 55: Should we start working on new py-libzfs-free plugin now? Generalizing code looks like a perfect first method for new ZFS service.

src/middlewared/middlewared/plugins/boot_/environments.py
```diff
@@ -0,0 +1,185 @@
...
+                " ".join([zget, root, json_flags]), shell=True, capture_output=True
+            )
+        except Exception:
+            self.logger.exception(
+                "Unexpected failure list zfs properties of boot dataset"
```
themylogin, line 60: We should log what the stderr is probably

src/middlewared/middlewared/plugins/boot_/environments.py
yocalebo: It's not a bad idea, however, we are going to start working on `py-libzfs2` (I've got a NEP that I can share with you). In `py-libzfs2`, we will be working with upstream to make `libzfs*` thread-safe and re-entrant. This would mean we could use native libzfs bindings inside coroutines safely. Anyways, we could potentially start writing our own separate package for doing things via zfs/zpool using `-j` flag but for right now, i'll keep this as-is.

src/middlewared/middlewared/plugins/boot_/environments.py
```diff
@@ -0,0 +1,189 @@
...
+)
+from middlewared.service import filterable_api_method, Service, private
+from middlewared.service_exception import CallError, ValidationError
+from middlewared.utils import filter_list
+from middlewared.utils.size import format_size
+
+
+def run_zectl_cmd(cmd):
+    try:
+        cmd.remove("zectl")
```
themylogin, line 27: Looks like a hack. Can we just do `subprocess.run(["zectl"] + cmd)` instead and not change `cmd` at all?

yocalebo: fixed

themylogin (approved):
Thank you!

PR #14869, opened 2024-11-05
NAS-132270 / 25.04 / Improve fcport.status when ALUA is disabled

bmeagherix:
This change will avoid a `KeyError` for NPIV configurations when ALUA is disabled.


mgrimesix (approved):
LGTM

PR #14879, opened 2024-11-06
NAS-132305 / 25.04 / Fix and improve tests/README.md

yocalebo:
It was pointed out to me that some have actually referenced this file (when they initially started) to get an idea of how to write/run tests. Let's improve this read me file with modern practices and bring it up to reality.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -10,16 +10,18 @@ In middleware/tests run the command bellow
 `pip3 install -r requirements.txt`
 
 ## Running the API test(s)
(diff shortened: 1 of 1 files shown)


themylogin:
Should we maybe put this to `docs/`?

yocalebo:
> Should we maybe put this to `docs/`?

It's not a bad idea, but I'll leave this here for now. Maybe I'll open another ticket so this gets moved to the `docs/` directory.

PR #14880, opened 2024-11-06
NAS-132306 / 25.04 / Fix --tests handling

bmeagherix:
The change that introduced `parse_test_name` broke `--tests` handling as the "api2" prefix was missing.

### DIFF
--- a/tests/runtest.py
+++ b/tests/runtest.py
@@ -312,11 +312,13 @@ def parse_test_name(test):
         return f"{filename}.py::{testname}"
     return test
 
+def parse_test_name_prefix_dir(test_name):
+    return f"api2/{parse_test_name(test_name)}"
 
 if tests:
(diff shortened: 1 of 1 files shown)


yocalebo (approved):

themylogin:
Does `--test` work as it was working before with this change?

bmeagherix:
> Does `--test` work as it was working before with this change?

Yes, I believe so.  For example, both of these work:
```
(middle_test) brian@debian12:~/truenas/middleware/tests$ ./runtest.py -vv --ip scale3 --interface enp0s8 --password qapassword -s --isns_ip 192.168.56.200 --test api2/test_mock_remote
(middle_test) brian@debian12:~/truenas/middleware/tests$ ./runtest.py -vv --ip scale3 --interface enp0s8 --password qapassword -s --isns_ip 192.168.56.200 --test test_mock_remote.py

```

PR #14881, opened 2024-11-06
NAS-132307 / 25.04 / remove accepts from CacheService plugin

yocalebo:
Our `api_method` decorator was designed (primarily) for public facing endpoints. However, for a private service like the `CacheService`, there is no reason to go through the boilerplate of adding all the arguments/returns. It's much easier and better (imo) to simply keep it marked private and remove the `@accepts` decorator. I've used builtin annotations to keep some semblance of what argument types are being passed to each method. This was easiest path forward without breaking backwards compatibility. There should be no change in functionality.

### DIFF
--- a/src/middlewared/middlewared/plugins/cache.py
+++ b/src/middlewared/middlewared/plugins/cache.py
@@ -1,81 +1,58 @@
-from middlewared.schema import Any, Str, Int, accepts
-from middlewared.service import Service, private
-
 from collections import namedtuple
-import time
+from collections.abc import Callable
+from time import monotonic
+from typing import Any
+
+from middlewared.service import Service
 
 
 class CacheService(Service):
-
     class Config:
         private = True
 
     def __init__(self, *args, **kwargs):
         super(CacheService, self).__init__(*args, **kwargs)
         self.__cache = {}
-        self.kv_tuple = namedtuple('Cache', ['value', 'timeout'])
+        self.kv_tuple = namedtuple("Cache", ["value", "timeout"])
 
-    @accepts(Str('key'))
-    def has_key(self, key):
-        """
-        Check if given `key` is in cache.
-        """
+    def has_key(self, key: str):
+        """Check if given `key` is in cache."""
         return key in self.__cache
 
-    @accepts(Str('key'))
-    def get(self, key):
+    def get(self, key: str):
         """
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/cache.py
```diff
@@ -1,81 +1,58 @@
...
 
+class CacheService(Service):
     class Config:
         private = True
 
     def __init__(self, *args, **kwargs):
         super(CacheService, self).__init__(*args, **kwargs)
         self.__cache = {}
-        self.kv_tuple = namedtuple('Cache', ['value', 'timeout'])
+        self.kv_tuple = namedtuple("Cache", ["value", "timeout"])
```
anodos325, line 16: Why are we changing quotes here?

yocalebo: Just standardized the file to double-quotes

src/middlewared/middlewared/plugins/cache.py
```diff
@@ -1,81 +1,58 @@
...
-        if cache is not None:
-            cache = cache.value
-
-        return cache
+    def pop(self, key: str):
+        """Removes and returns `key` from cache."""
+        try:
+            self.__cache.pop(key, None).value
+        except AttributeError:
+            pass
```
anodos325, line 44: Where's the return?

yocalebo: Fixed it

src/middlewared/middlewared/plugins/cache.py
```diff
@@ -1,81 +1,58 @@
...
-
-        if cache is not None:
-            cache = cache.value
-
-        return cache
+    def pop(self, key: str):
+        """Removes and returns `key` from cache."""
+        try:
+            return self.__cache.pop(key, None).value
+        except AttributeError:
```
anodos325, line 43: Would it be better to avoid AttributeError in first place?

yocalebo: It's quicker to catch known error then check for existence 🤷 but I can go back to original way of doing it.

yocalebo: fixed it

creatorcary (approved):

PR #14884, opened 2024-11-06
NAS-132312 / 25.04 / Validate cloud credentials using pydantic and discriminator

themylogin:
Current cloud credentials model is like this:

```
{
    "id": 1,
    "name": "test",
    "provider": "S3",
    "attributes": {"secret_access_key": "key"}
}
```

Pydantic discriminators can't work with this. They need the following structure:

```
{
    "id": 1,
    "name": "test",
    "provider": {"type": "S3", "secret_access_key": "key"}
}
```

This PR converts the API this way.

Any calls made by legacy clients will use the previous API (that is now defined as 24.10).


anodos325:
Should `type` be `provider` here so that field names are a little more consistent before/after?

themylogin:
@anodos325 sorry, I don't understand, consistent with what? I chose `type` because that's what in all the examples and that's what commonly used.

themylogin:
time 24:00

PR #14885, opened 2024-11-06
Fix `SystemServiceService` using `@accepts`

themylogin:
This causes
```
root@tnft-hv-snmp[~]# midclt call snmp.config
[EFAULT] Too many arguments (expected 0, found 1)
```
This method implementation is essentially the same than in its parent class (`ConfigService`), but wrapped in `@accepts()`. `ConfigService` works perfectly with new API, so we can just remove this implementation in order to resolve this issue.


themylogin:
time 0:10

yocalebo (approved):

mgrimesix (approved):

PR #14900, opened 2024-11-07
NAS-131723 / 25.04 / Add exclusion support for TrueCloud backup tasks

creatorcary:
Patterns can be given when creating a TrueCloud Backup task that will exclude matching directories and files from the backup.

https://restic.readthedocs.io/en/latest/040_backup.html#excluding-files


yocalebo:
@themylogin friendly ping

themylogin (approved):

PR #14901, opened 2024-11-08
NAS-132347 / 25.04 / Fix Fibre Channel ALUA

bmeagherix:
Avoid some fragility when toggling ALUA on.

Found that needed to **skip** the NPIV target in `scst.conf` when `standby_write_empty_config` was true, and add them afterwards.  Avoids a `scstadmin` load error on the (required) subsequent service reload.

Also, **always** write physical port targets to `scst.conf` as these will otherwise be auto-added by SCST - with a potentially clashing `rel_tgt_id`.  CI tests needed to be updated to take account of this change.

Finally, implemented some changes wrt ALUA failover.  Previously all targets were iSCSI, whereas now `iscsi.alua.become_active` and `iscsi.alua.removed_target_extent` need to take slightly different actions depending on the target mode.  This also necessitated some minor (private) API changes in `iscsi.scst`.

### DIFF
--- a/src/middlewared/middlewared/etc_files/scst.conf.mako
+++ b/src/middlewared/middlewared/etc_files/scst.conf.mako
@@ -148,6 +148,7 @@
     clustered_extents = set()
     active_extents = []
     standby_write_empty_config = False
+    skipped_wwpns = set()
     if failover_status == "MASTER":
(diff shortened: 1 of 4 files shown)


mgrimesix (approved):
LGTM

PR #14908, opened 2024-11-10
NAS-132402 / 25.04 / Refactor dtype key in VM device attrs

Qubad786:
## Context

In order to remove `accepts` from VM plugin, we need to move `dtype` key to inside of `vm_device.attributes` as that is how the new api style models will work with VM devices. However this change is extensive/involved and has been done separately before the actual removal of accepts decorator from VM plugin so that it is easier to review/manage the changes where accepts decorator is removed.


themylogin:
Can it be just `type` for consistency with other plugins? Or is this key already used in some schema?

Qubad786:
@themylogin we already have `type` key for DISPLAY device present already so it will clash

themylogin (approved):
Understood, thank you!

PR #14909, opened 2024-11-11
NAS-132248 / 25.04 / Avoid potentially unnecessary kerberos.start when activating AD

anodos325:
If we're in the process of joining active directory we will already have a kerberos ticket available and so can skip the step where we restart the kerberos service. This adds additional hardening against imcomplete sysvol replication.

### DIFF
--- a/src/middlewared/middlewared/plugins/directoryservices_/activedirectory_join_mixin.py
+++ b/src/middlewared/middlewared/plugins/directoryservices_/activedirectory_join_mixin.py
@@ -14,6 +14,7 @@ from middlewared.utils.directoryservices.ad_constants import (
 )
 from middlewared.utils.directoryservices.constants import DSType
 from middlewared.utils.directoryservices.krb5 import (
+    gss_dump_cred,
     gss_get_current_cred,
     kerberos_ticket,
 )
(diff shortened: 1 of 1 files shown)


src/middlewared/middlewared/plugins/directoryservices_/activedirectory_join_mixin.py
```diff
@@ -383,13 +398,21 @@ def _ad_join(self, job: Job, ds_type: DSType, domain: str):
...
+            })
+        except KRB5Error:
+            # Attempt to kinit against DC we were talking to failed and so we'll switch to generic
+            # kinit loop to wait for things to settle.
+            self.logger.debug('Initial attempt to kinit with new kerberos pricipal failed. Starting kinit loop.', exc_info=True)
+            job.set_progress(80, 'Waiting for active directory to replicate machine account changes.')
+            self._ad_wait_kerberos_start()
+        else:
+            self.middleware.call_sync('kerberos.wait_for_renewal')
+            self.middleware.call_sync('etc.generate', 'kerberos')
```
bmeagherix, line 416: Just wondering.  Should we set the job progress to 100 at the end of this function? (95 before the service update?)

bmeagherix (approved):
LGTM

src/middlewared/middlewared/plugins/directoryservices_/activedirectory_join_mixin.py
anodos325: At this point no, because the job is being wrapped around in 24.10. In future refactor work for 25.04, yes.

PR #14915, opened 2024-11-11
NAS-132418 / 25.04 / Fix result for device add/delete

william-gr:


src/middlewared/middlewared/api/v25_04_0/virt.py
```diff
@@ -240,7 +240,7 @@ class VirtInstanceDeviceAddArgs(BaseModel):
 
 
 class VirtInstanceDeviceAddResult(BaseModel):
-    result: dict
+    result: Literal[True]
```
yocalebo, line 243: Should it be `result: bool` ?

william-gr: Well, it won't ever return False, if there is an error it will be a CallError.
I think I have seen other places using Literal[True] but I can change to bool fi thats preferred

tests/api2/test_virt_002_instance.py
```diff
@@ -96,19 +96,21 @@ def test_virt_instance_device_add():
     assert ssh(f'incus list {INS1_NAME} -f json| jq ".[].status"').strip() == '"Running"'
     call('virt.instance.stop', INS1_NAME, {'force': True}, job=True)
 
-    call('virt.instance.device_add', INS1_NAME, {
+    res = call('virt.instance.device_add', INS1_NAME, {
```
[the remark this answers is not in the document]
william-gr: Done

src/middlewared/middlewared/api/v25_04_0/virt.py
yocalebo: Nope, if it raises an exception otherwise, this is fine.

yocalebo (approved):

PR #14920, opened 2024-11-12
NAS-132457 / 25.04 / Minor ACME logic refactors

sonicaj:
This PR adds some ACME logic refactors to allow us to call cert issuance logic directly minimizing existing dependencies.

### DIFF
--- a/src/middlewared/middlewared/plugins/acme_protocol_/client_utils.py
+++ b/src/middlewared/middlewared/plugins/acme_protocol_/client_utils.py
@@ -1,10 +1,27 @@
 import json
+import typing
 
 import josepy as jose
 from acme import client, messages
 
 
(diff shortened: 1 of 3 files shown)


src/middlewared/middlewared/plugins/acme_protocol_/client_utils.py
```diff
@@ -0,0 +1,50 @@
+import json
+
+import josepy as jose
+from acme import client, messages
+
+
+def get_acme_client_and_key(data: dict) -> tuple[client.ClientV2, jose.JWKRSA]:
+    """
+    Expected data dict should contain the following
```
themylogin, line 9: If interested, you can use `typing.TypedDict` to describe this schema more formally. It then can be re-used, IDE autocompletion will work better

themylogin (approved):

PR #14925, opened 2024-11-12
NAS-132462 / 25.04 / Fix schema annotations for new boot.environment plugin

yocalebo:
UI received the following traceback on certain endpoints
```
  File "/usr/lib/python3/dist-packages/truenas_api_client/legacy.py", line 416, in call
    return self.wait(c, callback=callback, job=job, timeout=timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/truenas_api_client/legacy.py", line 435, in wait
    raise ValidationErrors(c.extra)
truenas_api_client.exc.ValidationErrors: [EINVAL] result: Input should be a valid dictionary or instance of BootEnvironmentEntry
```
This is because the result annotation didn't match what was being returned. This fixes said issue and also updates the test to be more explicit about the returned objects.

### DIFF
--- a/src/middlewared/middlewared/plugins/boot_/environments.py
+++ b/src/middlewared/middlewared/plugins/boot_/environments.py
@@ -142,14 +142,14 @@ class BootEnvironmentService(Service):
             )
         else:
             run_zectl_cmd(["activate", data["id"]])
-            return self.query([["id", "=", data["id"]]])
+            return self.query([["id", "=", data["id"]]], {"get": True})
 
     @api_method(BootEnvironmentCloneArgs, BootEnvironmentCloneResult)
     def clone(self, data):
         be = self.validate_be("boot.environment.clone", data["id"])
         self.validate_be("boot.environment.clone", data["target"], should_exist=False)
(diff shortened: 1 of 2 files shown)


anodos325 (approved):

src/middlewared/middlewared/plugins/boot_/environments.py
```diff
@@ -142,14 +142,14 @@ def activate(self, data):
             )
         else:
             run_zectl_cmd(["activate", data["id"]])
-            return self.query([["id", "=", data["id"]]])
+            return self.query([["id", "=", data["id"]]], {"get": True})
```
themylogin, line 145: You can just use `get_instance`

yocalebo: it's actually `get_instance__sync` but yeah, I'll change it. Thanks

yocalebo: hmm, nope, I don't have access to that because of inherited class. I'll leave it as-is for now.

PR #14926, opened 2024-11-12
Do not validate the results of method calls made by a full administrator through the legacy WebSocket API

themylogin:
That's the intended behavior. It was broken recently by the new code that adapts the result to 24.10 schemas if they exist. Validation of the method call result should be performed lazily in that case.


themylogin:
time 1:30

yocalebo (approved):

PR #14932, opened 2024-11-12
NAS-132471 / 25.04 / Add virt.instance.device_update and fix destination for zvol

william-gr:


src/middlewared/middlewared/plugins/virt/instance_device.py
```diff
@@ -198,11 +199,20 @@ async def generate_device_name(self, device_names: list[str], device_type: str)
...
         return name
 
-    async def __validate_device(self, device, schema, verrors: ValidationErrors):
+    async def __validate_device(self, device, schema, verrors: ValidationErrors, instance: str = None):
         match device['dev_type']:
             case 'PROXY':
-                verror = await self.middleware.call('port.validate_port', schema, device['source_port'])
-                verrors.extend(verror)
+                # We want to make sure there are no other instances using that port
+                ports = await self.middleware.call('port.ports_mapping')
```
yocalebo, line 206: Do we need to always validate `source_port` even if it doesn't change?

william-gr: Good point, will fix

src/middlewared/middlewared/plugins/virt/instance_device.py
```diff
@@ -198,11 +199,23 @@ async def generate_device_name(self, device_names: list[str], device_type: str)
...
         match device['dev_type']:
             case 'PROXY':
-                verror = await self.middleware.call('port.validate_port', schema, device['source_port'])
-                verrors.extend(verror)
+                # Skip validation if we are updating and port has not changed
+                if old and old['source_port'] == device['source_port']:
+                    return
+                # We want to make sure there are no other instances using that port
+                ports = await self.middleware.call('port.ports_mapping')
+                if device['source_port'] in ports:
```
yocalebo, line 210: You can get rid of this if branch.

```
for attachment in ports.get(device['source_port'], {}).values():

yocalebo (approved):
Another nit but approved none-the-less

PR #14949, opened 2024-11-13
NAS-130918 / 25.04 / remove REST from test_426

yocalebo:


tests/api2/test_426_smb_vss.py
```diff
@@ -111,17 +104,8 @@ def check_previous_version_contents(path, contents, offset):
...
-        "share_type": "SMB"
-    }
-    results = POST("/pool/dataset/", payload)
-    assert results.status_code == 200, results.text
-    result = POST("/zfs/snapshot/", {
-        "dataset": ds,
-        "name": "init",
-    })
-    assert result.status_code == 200, results.text
+    assert call("pool.dataset.create", {"name": ds, "share_type": "SMB"})
```
themylogin, line 107: `assert` is redundant here

themylogin (approved):

creatorcary:
Are we still removing the numbers from file names/test names?

yocalebo:
> Are we still removing the numbers from file names/test names?

Good question but I'm too hesitant to do that with tests like this. I already learned import order matters in this script so I'd rather not shake up the tree too much.

PR #14951, opened 2024-11-14
NAS-132505 / 25.04 / Remove accepts style decorator from catalog plugin

Qubad786:
## Context

Removes accepts style decorator from catalog plugin and uses the new api method we have in place.


src/middlewared/middlewared/api/v25_04_0/catalog.py
```diff
@@ -0,0 +1,102 @@
...
+    latest_human_version: str = Field(description='Human-readable version of the app.')
+    last_update: str = Field(description='Timestamp of the last update in ISO format.')
+    name: str = Field(description='Name of the app.')
+    recommended: bool = Field(description='Indicates if the app is recommended.')
+    title: str = Field(description='Title of the app.')
+    maintainers: list[Maintainer] = Field(description='List of app maintainers.')
+    tags: list[str] = Field(description='Tags associated with the app.')
+    screenshots: list[HttpUrl] = Field(description='List of screenshot URLs.')
+    sources: list[HttpUrl] = Field(description='List of source URLs.')
+    icon_url: list[HttpUrl] = Field(None, description='URL of the app icon.')
```
themylogin, line 69: If it is a list, how can its default be `None`?

themylogin (commented):
Please rebase this on the latest master and run tests again, an important result validation change was merged

src/middlewared/middlewared/api/v25_04_0/catalog.py
```diff
@@ -0,0 +1,102 @@
...
+
+    # We do this because if we change anything in catalog.json, even older releases will
+    # get this new field and different roles will start breaking due to this
+    model_config = ConfigDict(extra='allow')
+
+
+@single_argument_args('catalog_apps_options')
+class CatalogAppsArgs(BaseModel):
+    cache: bool = Field(default=True)
```
creatorcary, line 78: If you're only setting the default, it's enough to just have `cache: bool = True`.

src/middlewared/middlewared/api/v25_04_0/catalog.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class Maintainer(BaseModel):
+    name: str
+    email: str
+    url: HttpUrl | None
+
+
+class CatalogAppInfo(BaseModel):
+    app_readme: str = Field(description='HTML content of the app README.')
```
creatorcary, line 51: `null=True` in the old schema implies that this should be `str | None`, same for several other fields in this model.

src/middlewared/middlewared/api/v25_04_0/catalog.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class Maintainer(BaseModel):
+    name: str
+    email: str
+    url: str | None
+
+
+class CatalogAppInfo(BaseModel):
+    app_readme: LongString | None = Field(description='HTML content of the app README.')
```
yocalebo, line 53: Okay, so you can get around all these `Field(` definitions by adding a doctoring underneath each variable. The docstrings will be rendered in the api docs automatically. i.e. something like
```
app_readme: LongString | None = None
"""HTML content of the app README."""
categories: list[str]
"""List of categories for the app."""

File: src/middlewared/middlewared/api/v25_04_0/catalog.py [hunk 5]
```diff
@@ -0,0 +1,121 @@
...
+    '''Title of the app.'''
+    maintainers: list[Maintainer]
+    '''List of app maintainers.'''
+    tags: list[str]
+    '''Tags associated with the app.'''
+    screenshots: list[str]
+    '''List of screenshot URLs.'''
+    sources: list[str]
+    '''List of source URLs.'''
+    icon_url: str | None = Field(None, description='URL of the app icon.')
```
yocalebo, line 88:
```suggestion
    icon_url: str | None = None
    '''URL of the app icon'''
```

yocalebo (approved):

themylogin (approved):

